PR #2468, opened 2018-12-18
Add CR in sanity check

BiNZGi:
Specified in https://stomp.github.io/stomp-specification-1.2.html#STOMP_Frames also carriage return (octet 13) is allowed:

> The frame starts with a command string terminated by an end-of-line (EOL), which consists of an OPTIONAL carriage return (octet 13) followed by a REQUIRED line feed (octet 10).

### DIFF
--- a/artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompDecoder.java
+++ b/artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompDecoder.java
@@ -527,7 +527,7 @@ public class StompDecoder {
       }
 
       // Sanity check
-      if (workingBuffer[pos - 1] != NEW_LINE) {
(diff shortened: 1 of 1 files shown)


jbertram:
Could you create a JIRA for this at https://issues.apache.org/jira/projects/ARTEMIS and also add a test to reproduce the failure you were seeing without the fix? Something simple in `org.apache.activemq.artemis.tests.integration.stomp.StompTest` should suffice.

jbertram:
After looking at this a bit more closely I think this change is incorrect. The optional carriage return (octet 13) was added in STOMP 1.2, but you have changed `org.apache.activemq.artemis.core.protocol.stomp.StompDecoder` which is responsible for STOMP 1.0 frames. There are different decoder implementations for STOMP 1.1 & 1.2, and from what I can tell the 1.2 decoder handles the EOL properly. In fact, the STOMP client used by the test suite sends `\r\n` EOLs by default when using STOMP 1.2. 

Unless you have a test demonstrating a problem I think this PR should be closed.

PR #2469, opened 2018-12-18
ARTEMIS-2221 avoid unnecessary Bindings instance creation

jbertram:
When trying to get the bindings for an address the getBindingsForAddress
method will create a Bindings instance if there are no bindings for the
address. This is unnecessary in most circumstances so use the
lookupBindingsForAddress method instead and check for null.


jbertram:
@clebertsuconic, done.

michaelandrepearce:
@jbertram dont worry managed to sort the conflict on merge.

PR #2290, opened 2018-09-04
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these [code quality badges](https://lgtm.com/projects/g/apache/activemq-artemis/ci/#badges) to your project. They will indicate a very high code quality to potential users and contributors.
You can also check the alerts discovered by [LGTM](https://lgtm.com/projects/g/apache/activemq-artemis).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

PR #2470, opened 2018-12-18
Fixes for alerts from lgtm.com 

jbertram:
After reviewing lgtm.com based on #2290 I decided to fix a handful of the warnings and errors they reported.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/PrintData.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/PrintData.java
@@ -216,63 +216,63 @@ public class PrintData extends DBOption {
 
          if (pgStore != null) {
             folder = pgStore.getFolder();
-         }
-         out.println("####################################################################################################");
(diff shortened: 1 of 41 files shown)


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/PrintData.java
```diff
@@ -216,63 +216,63 @@ private static void printPages(DescribeJournal describeJournal,
 
          if (pgStore != null) {
             folder = pgStore.getFolder();
```
clebertsuconic: There's a semantic change here, are you sure?

jbertram: The not-null check here on `pgStore` indicates that `pgStore` may, in fact, be null. If it is null then it will trigger a `NullPointerException` almost straight away.

clebertsuconic:
all lgtm, pun intended

If we could run the full testsuite just to be sure.

michaelandrepearce:
+1 from me, nice!

PR #2473, opened 2018-12-19
ARTEMIS-1058 Jars in web tmp dir locked on Windows

gaohoward:
Because Sun's URLClassLoader never closes it's jar file handles
Jetty doesn't cleanup is temp web dir after Artemis broker shut
down normally on Windows.

To work around this a new process is forked before broker VM
exits to clean up the tmp dir if they are not deleted. The
new process out lives the main VM so that those jar's handles
are released and the temp dir can be cleaned up.

### DIFF
--- a/artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
+++ b/artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
@@ -31,7 +31,6 @@ import org.apache.activemq.artemis.dto.AppDTO;
 import org.apache.activemq.artemis.dto.ComponentDTO;
 import org.apache.activemq.artemis.dto.WebServerDTO;
 import org.apache.activemq.artemis.utils.FileUtil;
-import org.apache.activemq.artemis.utils.TimeUtils;
 import org.eclipse.jetty.server.ConnectionFactory;
 import org.eclipse.jetty.server.Connector;
 import org.eclipse.jetty.server.HttpConfiguration;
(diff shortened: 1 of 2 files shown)


gaohoward:
manually tested on Windows 7 64bit and Windows 10 64bit.

jbertram:
Wasn't there a Jetty issue opened about this? If so, how was that resolved?

gaohoward:
Yes there is a issue opened. https://github.com/eclipse/jetty.project/issues/1425
that doesn't get fixed (PR didn't merged because the hacking into URLClassLoader brings concerns)
So that's why i came up with this work around.
(The issue still there with jetty latest releases).

artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
```diff
@@ -237,28 +236,39 @@ public void start() throws Exception {
...
+                  ActiveMQWebLogger.LOGGER.tmpFileNotDeleted(tmpdir);
+                  strBuilder.append(tmpdir);
+                  strBuilder.append(",");
+                  found = true;
                }
             }
          }
+
+         if (found) {
+            String artemisHome = System.getProperty("artemis.home");
```
clebertsuconic: can't you get the property from other means? Fileconfiguration for instance?

clebertsuconic: you're wrong actually,

WebServerComponent is getting the artemisInstance and artemisHome as parameters.

It is important you don't use the variable here as you may break embedded cases.

We use things like this on our testsuite as well. so, you can use a property to store what's passed from the configure method, and use the variable directly here instead.

gaohoward: ah ok. I'll use the parameter. thx.

PR #2474, opened 2018-12-20
[ARTEMIS-1536]: Incorrect Journal filesize calculation where specified size is lest that the block size when using AIO.

ehsavoie:
* If the specified file size is under the fs block size then the resulting file size is 0. Setting it to the block size in this case.

Jira: https://issues.apache.org/jira/browse/ARTEMIS-1536

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
@@ -161,16 +161,7 @@ public class JournalStorageManager extends AbstractJournalStorageManager {
       journalFF.setDatasync(config.isJournalDatasync());
 
 
-      int fileSize = config.getJournalFileSize();
-      // we need to correct the file size if its not a multiple of the alignement
-      int modulus = fileSize % journalFF.getAlignment();
-      if (modulus != 0) {
-         int difference = modulus;
-         int low = config.getJournalFileSize() - difference;
-         int high = low + journalFF.getAlignment();
-         fileSize = difference < journalFF.getAlignment() / 2 ? low : high;
-         ActiveMQServerLogger.LOGGER.invalidJournalFileSize(config.getJournalFileSize(), fileSize, journalFF.getAlignment());
-      }
+      int fileSize = fixJournalFileSize(config.getJournalFileSize(), journalFF.getAlignment());
       Journal localMessage = createMessageJournal(config, criticalErrorListener, fileSize);
 
       messageJournal = localMessage;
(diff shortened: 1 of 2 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -163,13 +163,17 @@ protected void init(Configuration config, IOCriticalErrorListener criticalErrorL
...
       int fileSize = config.getJournalFileSize();
       // we need to correct the file size if its not a multiple of the alignement
-      int modulus = fileSize % journalFF.getAlignment();
-      if (modulus != 0) {
-         int difference = modulus;
-         int low = config.getJournalFileSize() - difference;
-         int high = low + journalFF.getAlignment();
-         fileSize = difference < journalFF.getAlignment() / 2 ? low : high;
-         ActiveMQServerLogger.LOGGER.invalidJournalFileSize(config.getJournalFileSize(), fileSize, journalFF.getAlignment());
+      if (fileSize <= journalFF.getAlignment()) {
```
franz1981: If `alignment` is a power of 2 (and it should be right?) you can use:
```
   public static long align(final long value, final long alignment) {
      return (value + (alignment - 1)) & ~(alignment - 1);
   }
```
That is into `BytesUtils::align`

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -163,13 +163,17 @@ protected void init(Configuration config, IOCriticalErrorListener criticalErrorL
...
+         fileSize = journalFF.getAlignment();
```
clebertsuconic: I would actually throw an exception here. I wouldn't expect a block size that low. you can only have the header on the journal and nothing else if you specify fileSize == alignment.

ehsavoie: There is already such an exception when creating the journal (java.lang.IllegalArgumentException: File size cannot be less than 1024 bytes
    Caused by: java.lang.IllegalArgumentException: File size cannot be less than 1024 bytes") . Would setting to high be acceptable ?

clebertsuconic: Lets not complicate things... but a unit test would be nice :)

clebertsuconic: I will just mere it.. but if you can add an unit test...

ehsavoie: Done

clebertsuconic:
merged & cherry-picked into 2.6.x

PR #2475, opened 2018-12-20
ARTEMIS-2144 - tx begin failure in ra doesn't get cleaned up

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-2144


andytaylor:
don't merge just yet I want to run the whole test suite

andytaylor:
This is good to merge now, full test suite pass

PR #2478, opened 2018-12-24
ARTEMIS-2210 Fix PagingStore creation synchronization issue 

gaohoward:
In PagingManagerImpl#getPageStore() the operations on the map 'stores'
are not synchronzed and it's possible that more than one paging store is
created for one address.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingManagerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingManagerImpl.java
@@ -335,19 +335,30 @@ public final class PagingManagerImpl implements PagingManager {
    }
 
    /**
-    * stores is a ConcurrentHashMap, so we don't need to synchronize this method
+    * This method creates a new store if not exist.
     */
    @Override
    public PagingStore getPageStore(final SimpleString storeName) throws Exception {
       if (managementAddress != null && storeName.startsWith(managementAddress)) {
          return null;
       }
-      PagingStore store = stores.get(storeName);
 
+      PagingStore store = stores.get(storeName);
       if (store != null) {
          return store;
       }
-      return newStore(storeName);
+      //only if store is null we use computeIfAbsent
+      try {
+         return stores.computeIfAbsent(storeName, (s) -> {
+            try {
+               return newStore(s);
+            } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


gaohoward:
The fix is so obvious that I decide not to provide a test. :)

franz1981:
@gaohoward The synchronized block will block any threads attempting to both getting and creating a paging store, while a chm::get is lock-free ie won't block any threads and chm::computeIfAbsent is locked on per-segment basis (that means that depends on the number of segments and how the hashing of the keys will distribute the entries into the segment.

Putting a synchronized like this will prevent any threads to do anything on an hot-path, blocking each others: my advice is to keep chm::get lock-free (as it is now) and into `newStore(storeName)` use chm::computeIfAbsent to avoid duplicates creatings of `PagingStore` instances.
This should keep the scalability as it is now without allocating unnecessary instances. wdyt?

gaohoward:
@franz1981 ok, make sense. I'll change it. thx.

gaohoward:
@franz1981 I've update the branch, can you take a look? thx.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingManagerImpl.java
```diff
@@ -335,19 +335,25 @@ public void deletePageStore(final SimpleString storeName) throws Exception {
...
    @Override
    public PagingStore getPageStore(final SimpleString storeName) throws Exception {
       if (managementAddress != null && storeName.startsWith(managementAddress)) {
          return null;
       }
-      PagingStore store = stores.get(storeName);
 
-      if (store != null) {
-         return store;
+      try {
```
franz1981: I think that `stores.get` is better that will be used in the fast path as it was before: `chm:.get` is completly lock-free, while `chm::computeIfAbsent` will lock on segment level ie better to fallback to `chm::computeIfAbsent` only if `chm::get` return `null`

gaohoward: ok got it. I've updated the branch. Please take a look again. thx.

franz1981:
Looks fine to me buddy, just change the title, because no sync is happening anymore ;)

gaohoward:
@franz1981 Title changed. Thanks mate!

PR #2479, opened 2018-12-24
ARTEMIS-2211 Refactor ByteBuffer pooling, alignment and zeroing

franz1981:
Refactored thread local ByteBuffer pooling, alignment
and zeroing in order to avoid duplicate code and 
improve code coverage with tests.
In addition are being provided faster branchless
alignment operations and optional zeroing of
pooled ByteBuffers for both ASYNCIO and 
NIO/MAPPED journal types.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java
@@ -17,6 +17,8 @@
 package org.apache.activemq.artemis.utils;
 
 import java.nio.ByteBuffer;
+import java.nio.ReadOnlyBufferException;
+import java.util.Arrays;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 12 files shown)


franz1981:
I'm running the CI on it: I would like to provide some unit test, but I'm not sure it makes sense, considering that the original code were already tested end-to-end.

michaelandrepearce:
@franz1981 this is out my league on this, im not so familiar with journal bits/logic, you ok with me leaving this for @clebertsuconic ?

franz1981:
@michaelandrepearce @clebertsuconic I have added some unit tests quite strict, given that these classes are used almost everywhere :+1:

franz1981:
Sure! But given that are mostly math tricks if you want I can name the class with better name like Pow2Math, given that are math computation for power of 2 numbers

franz1981:
@michaelandrepearce @clebertsuconic We can talk about this one on the issue/forum too: the thing is that currently we have yet several parts on artemis where we rely on thread pool storage for direct `ByteBuffer`s:
1. write side of paging
2. write side of journal with NIO/MAPPED journal types are used

Considering that the context of execution in artemis is agnostic of the executing `Thread` and that the same task could move between executing `Thread`s, it could happen that after 1 minute of inactivity threads could die, they'll leak the thread pool storage related to them (until the first minor GC).
If a new burst of activity will create new threads and no minor GC has happened there risk is to have too many direct buffers alive that could lead the OOM killer of the OS to kill the first process that allocate native memory. 

For compaction, this has been addressed (by me) by allocating upfront the compaction task a single direct buffer and reusing it for any compaction sub-task, deallocating manually (like a C `free`) at the end: although it could seems a bad solution, it avoid leaking several MBs of direct memory if the compaction operations are handled by different threads each time. It would slow down compaction, but given that compaction is not meant to be an hot path like journal writes is best to have predictable memory footprint.

This same approach can't be used for the other 2 cases and I was thinking to provide a lock-free generic implementation of direct buffer pools to address it (as ASYNCIO does), but maybe there is a simpler and more effective solution: feel free to give your opinion ;)

franz1981:
@clebertsuconic @michaelandrepearce Please review it another time guys: I've included a change on how we perform zeroing of buffers, in order to provide (together with https://github.com/apache/activemq-artemis/pull/2522 changes) a further optimization on ASYNCIO ie avoid zeroing of the big buffer used by `TimedBuffer`.

michaelandrepearce:
@franz1981 has a full CI run occured on this? I noted earlier on you said you were going to run. also whats occuring with the possible leakage has that been addressed?

franz1981:
@michaelandrepearce @clebertsuconic Yep, the CI has run on it and it is ok :+1: 
I will use it right after for that optimization of both ASYNCIO and NIO to avoid zeroing of buffers while writing ;)

franz1981:
@michaelandrepearce 
> also whats occuring with the possible leakage has that been addressed

We have 2 leaks with NIO/MAPPED seq factories:

1. on compaction: it has been addressed not in this PR by NOT using `ByteBuffer` pooling ie it will allocate upfront and free the buffers when finished
2. on paging/while writing on the journal

The second one is due to thread-local `ByteBuffer` pooling + our thread pool + type of load: 

- we have burst of journal activity that will make N threads to be created 
- the burst of load will finish and more then 1 minute has passed: many threads got disposed, but the thread-pooled `ByteBuffer` will be cleaned up only if a GC will happen, but suppose it won't happen
- another burst of journal activity will make other N threads to be created
- we ends up having many `ByteBuffer` instances alive: old ones + new ones

michaelandrepearce:
@franz1981 cool, as per my original comment as this is journal area its a bit out my league so ill leave with clebert to make the final call and merge. but heres a +1 from me.

PR #2480, opened 2018-12-24
ARTEMIS-2212 Avoid using CLQ on ServerConsumerImpl

franz1981:
It would deliver a better performance for the most
common operations eg offer, poll, iterations, size.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
@@ -18,10 +18,12 @@ package org.apache.activemq.artemis.core.server.impl;
 
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.ArrayDeque;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
@@ -117,7 +119,10 @@ public class ServerConsumerImpl implements ServerConsumer, ReadyListener {
 
    @Override
    public String debug() {
-      return toString() + "::Delivering " + this.deliveringRefs.size();
+      String debug = toString() + "::Delivering ";
+      synchronized (lock) {
+         return debug + this.deliveringRefs.size();
+      }
    }
 
    /**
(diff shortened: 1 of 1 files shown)


franz1981:
@michaelandrepearce and @clebertsuconic  in the new here please take a look at this one :+1: 
I can change ArrayQueue with a LinkedList and I think that the benefits will be present due to avoiding a concurrent data structure, but we loose the added benefit of avoiding the garbage produced by the `Node` allocated.

michaelandrepearce:
Whats the concurrent performance on this. Do you have some stats?

Re sync lock blocks. Could be replaced with lock on this and or sync methods.

franz1981:
> Whats the concurrent performance on this. Do you have some stats?

Only allocations profiles and given that is in the hot path, we have a LOT of ConcurrentLinkedQueue::Node saved by this change.
The overall perf won't change by a bit (if not a tiny improvement), because it isn't a bottleneck AFAIK.

> Re sync lock blocks. Could be replaced with lock on this and or sync methods.

Agree at a first look, but I do not remember the reason why was used a separate lock TBH

michaelandrepearce:
If its not exposed and its used like doing lock on 'this". Then i can only assume it was the coders pattern at that time.

michaelandrepearce:
@franz1981, did you make the change to syncronized methods? As this PR is about performance im for getting the most we can in one ;)

franz1981:
@michaelandrepearce Not yet, need to chenge it and we just need the opinion from @clebertsuconic here, given that we I'm concerned about the max size that the array q can get and that will be retained for the whole life-cycle of the consumer :)

franz1981:
@michaelandrepearce I'm taking another look and I'm not 100% sure to replace the usage of the lock object with `this` or synchronized methods, because I see that the synchronization on `this` has a different semantic and uses, separated from `lock` eg 
```java
public HandleStatus handle(final MessageReference ref) throws Exception {
      // available credits can be set back to null with a flow control option.
      AtomicInteger checkInteger = availableCredits;
      if (callback != null && !callback.hasCredits(this) || checkInteger != null && checkInteger.get() <= 0) {
            logger.debug(this + " is busy for the lack of credits. Current credits = " +
                            availableCredits +
                            " Can't receive reference " +
                            ref);
         }

         return HandleStatus.BUSY;
      }

      synchronized (lock) {
```
`handle` is not synchronized, while other methods are, but a part of it `synchronized (lock)`, making the behaviour different if I replace `synchronized (lock)` with a `synchronized (this)` ie the risk is to be blocked by other method calls that currently are not blocking it

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -509,21 +518,16 @@ public synchronized void close(final boolean failed) throws Exception {
...
-
-      Iterator<MessageReference> iter = refs.iterator();
+      List<MessageReference> refs = cancelRefs(failed, false, null);
 
       Transaction tx = new TransactionImpl(storageManager);
 
-      while (iter.hasNext()) {
-         MessageReference ref = iter.next();
-
+      refs.forEach(ref -> {
```
clebertsuconic: is this atomic?

clebertsuconic: Ohhh... ok.. it's on the list from cancelRefs.. I confused as being the main list.

PR #2481, opened 2018-12-25
ARTEMIS-2213 don't expire critical component in the case of clock back drift

wy96f:
In our production cluster some brokers crashed. There is nothing unusual in the dump stack. After digging into code, we found component was incorrectly expired. When clock drifted back, left time was less than enter time. If the component was not entered in default 120000ms, it would be expired and server was halted.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/critical/CriticalMeasure.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/critical/CriticalMeasure.java
@@ -17,6 +17,7 @@
 
 package org.apache.activemq.artemis.utils.critical;
 
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 
 import org.jboss.logging.Logger;
@@ -31,9 +32,11 @@ public class CriticalMeasure {
    //uses updaters to avoid creates many AtomicLong instances
    static final AtomicLongFieldUpdater<CriticalMeasure> TIME_ENTER_UPDATER = AtomicLongFieldUpdater.newUpdater(CriticalMeasure.class, "timeEnter");
    static final AtomicLongFieldUpdater<CriticalMeasure> TIME_LEFT_UPDATER = AtomicLongFieldUpdater.newUpdater(CriticalMeasure.class, "timeLeft");
+   static final AtomicIntegerFieldUpdater<CriticalMeasure> PENDING_UPDATE = AtomicIntegerFieldUpdater.newUpdater(CriticalMeasure.class, "pending");
 
    private volatile long timeEnter;
    private volatile long timeLeft;
+   private volatile int pending;
 
    private final int id;
    private final CriticalComponent component;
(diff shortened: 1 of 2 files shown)


franz1981:
I know that System::nanoTime is based on monothonic clock of the CPU thast's not wall clock time ie can't go back.
System::currentTimeMillis ie UTC wall-clock can go backwards instead due to ntp alignment.
Can you please explain better whallt's happened?

michaelandrepearce:
I dont see how this handles clock drift if im honest. The test case doesnt actually cause a drift in nano.

Also Nanoseconds also are from an arbitary offset on the cpu, and afaik not from wall clock. So isnt subject to drift like millis are

michaelandrepearce:
@franz1981 haha snap. I had same thinking.

franz1981:
@michaelandrepearce thanks buddy you too!!!

wy96f:
We add a pending counter to count enterCritical calls. If left time is less than enter time and pending counter equals to 0, we don't expire component as all threads have called leaveCritical and it maybe caused by clock drift.
When a clock jumps backwards,  to avoid the edge case that  enterCritical is called during isExpired, we need to compare current time to latest timeEnter.

franz1981:
Nice investigation! Anyway,  CLOCK_MONOTONIC is not subject to NTP backward adjustment, but just incremental ones ie clock acceleration of any kind, but this would happen mostly for systems that shares a single common clock and I'm sure that is not the common case.
Re the openjdk code, I was already aware of cases where nanoTime was implemented using currentTimeMillis, but not that there was modern systems with such issues given that not just this critical analyser, but probably  anything that measure time in a Java program would be broken if it happens.
My concern is to add a XADD instruction in the hot path to support this fix given its cost.
Do you have verified that your system is not supporting nanoTime correctly?
Probably updating the kernel or just using something with hardware clock dedicated would help.

franz1981:
Ah, another consideration: nanoTime can actually go backward due to "starting in an unspecified point of time", but usually this need to be handled by not comparing directly values and just perform the difference, making use of numerical overflow to have correct values back, as reported by https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--.

wy96f:
@franz1981 thanks for the explanation. We have verified nanoTime is correctly supported. We observed the queue delivered messages and received no messages for abount 2mins, then broker is killed. Three brokers had been killed this week. That was weird. The dumped stacktrace is attached. Could you shed us some light? :)
[https://gist.github.com/wy96f/1a5c2f481e70d8dabde70f3a07eb57b7](url)

franz1981:
I will be able to look into that in the next days :) 
I can already ask you to collect time to safepoints/GC pauses if possible ie -XX:+PrintGCApplicationStoppedTime.
2 minutes seems a too long period TBH, but worth taking a look if you rely on MAPPED journal and/or paging a lot, given that major page faulting can causes long stall similar to very long full GC, but TBH nothing so long (2 minutes is a lot!).
As an additional suggestion you could run a Java program that just use one core and jmeasure the elapsed time between 2 consecutive nanoTime calls, recording in which wall-clock time a back-ward "drift" has happened to check if a broker shutdown has happened near the same time, makes sense?

wy96f:
@franz1981 thanks for your advice. We've enabled gc log and run the nano program. Any news will be reported:)

michaelandrepearce:
Can we close this for the interim?

wy96f:
> interim

Yes. The problem did't occur these days. We'll give more investigation.

michaelandrepearce:
@wy96f if you could close, only the person opening can close, or when we merge.

PR #2482, opened 2018-12-26
ARTEMIS-2214 Cache durable&deliveryTime in PagedReference

qihongxu:
We recently performed a test on artemis broker and found a severe performance issue.

When paged messages are being consumed, decrementMetrics in QueuePendingMessageMetrics will try to ‘getMessage’ to check whether they are durable or not. In this way queue will be locked for a long time because page may be GCed and need to be reload entirely. Other operations rely on queue will be blocked at this time, which cause a significant TPS drop. Detailed stacks are attached below.

This also happens when consumer is closed and messages are pushed back to the queue, artemis will check priority on return if these messages are paged.

To solve the issue, durable and priority need to be cached in PagedReference just like messageID, transactionID and so on. I have applied a patch to fix the issue. Any review is appreciated.

jira issue:https://issues.apache.org/jira/projects/ARTEMIS/issues/ARTEMIS-2214

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/PagedReferenceImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/PagedReferenceImpl.java
@@ -77,6 +77,12 @@ public class PagedReferenceImpl extends LinkedListImpl.Node<PagedReferenceImpl> 
 
    private Consumer<? super MessageReference> onDelivery;
 
+   //Durable field : 0 is false, 1 is true, -1 not defined
+   private static final byte IS_NOT_DURABLE = 0;
+   private static final byte IS_DURABLE = 1;
+   private static final byte UNDEFINED_IS_DURABLE = -1;
+   private byte durable = UNDEFINED_IS_DURABLE;
+
    @Override
    public Object getProtocolData() {
       return protocolData;
@@ -144,7 +150,8 @@ public class PagedReferenceImpl extends LinkedListImpl.Node<PagedReferenceImpl> 
          this.largeMessage = message.getMessage().isLargeMessage() ? IS_LARGE_MESSAGE : IS_NOT_LARGE_MESSAGE;
          this.transactionID = message.getTransactionID();
          this.messageID = message.getMessage().getMessageID();
-
+         this.durable = message.getMessage().isDurable() ? IS_DURABLE : IS_NOT_DURABLE;
+         this.deliveryTime = message.getMessage().getScheduledDeliveryTime();
          //pre-cache the message size so we don't have to reload the message later if it is GC'd
          getPersistentSize();
       } else {
@@ -152,6 +159,8 @@ public class PagedReferenceImpl extends LinkedListImpl.Node<PagedReferenceImpl> 
          this.transactionID = -2;
          this.messageID = -1;
          this.messageSize = -1;
+         this.durable = UNDEFINED_IS_DURABLE;
+         this.deliveryTime = UNDEFINED_DELIVERY_TIME;
       }
    }
 
@@ -385,6 +394,14 @@ public class PagedReferenceImpl extends LinkedListImpl.Node<PagedReferenceImpl> 
          }
       }
       return messageSize;
+   }
+
+   @Override
+   public boolean isDurable() {
+      if (durable == UNDEFINED_IS_DURABLE) {
+         durable = getMessage().isDurable() ? IS_DURABLE : IS_NOT_DURABLE;
+      }
+      return durable == IS_DURABLE;
    }
 
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/MessageReference.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/MessageReference.java
@@ -44,6 +44,8 @@ public interface MessageReference {
 
    long getMessageID();
 
+   boolean isDurable();
+
    SimpleString getLastValueProperty();
 
    /**
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LastValueQueue.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LastValueQueue.java
@@ -306,6 +306,11 @@ public class LastValueQueue extends QueueImpl {
       }
 
       @Override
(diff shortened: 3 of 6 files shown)


qihongxu:
[stacks.txt](https://github.com/apache/activemq-artemis/files/2709657/stacks.txt)

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/PagedReferenceImpl.java
```diff
@@ -74,6 +74,10 @@
 
    private long messageSize = -1;
 
+   private byte priority;
```
michaelandrepearce: default this to -1

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/PagedReferenceImpl.java
```diff
@@ -363,4 +371,20 @@ public long getPersistentSize() {
...
+   public byte getPriority() {
+      if (priority == -1) {
+         priority = getMessage().getPriority();
+      }
+      return priority;
+   }
+
+   @Override
+   public boolean isDurable() {
+      if (messageID < 0) {
```
michaelandrepearce: should use durable field or additional flag to see if its been cached, messageID maybe cached already but not durable.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/PagedReferenceImpl.java
```diff
@@ -74,6 +74,10 @@
 
    private long messageSize = -1;
 
+   private byte priority;
+
+   private boolean durable;
```
michaelandrepearce: need to default this not set somehow, possible use a byte to represent the boolean with 0 = false, 1 = true, -1 not set

qihongxu:
@michaelandrepearce Thanks for the review. Now I have changed the code at your suggestion.

qihongxu:
> Nice catch!!!
> If you care about checking perf of paging please try to profile the broker with https://github.com/jvm-profiling-tools/async-profiler using lock/cpu events too: it will allows to get profile data without affecting performance, helping to find issues like this one :)

Thanks for your advice! We used to record performance from producer/consumer side. Later on we will try to dig into broker side by using tools as recommended :)

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/PagedReferenceImpl.java
```diff
@@ -120,6 +128,8 @@ public PagedReferenceImpl(final PagePosition position,
          this.largeMessage = message.getMessage().isLargeMessage() ? IS_LARGE_MESSAGE : IS_NOT_LARGE_MESSAGE;
          this.transactionID = message.getTransactionID();
          this.messageID = message.getMessage().getMessageID();
+         this.priority = message.getMessage().getPriority();
```
wy96f: deliveryTime can be set in the constructor like transactionID , messageID , etc :)

qihongxu:
@michaelandrepearce 
There are some cases will perform lots of rollbacks in a short period of time. For example if we would like to upgrade our server while thousands of consumers are receiving message, the close of server causes massive rollbacks to original address, thus queue might blocked on reading GC'ed pages. Under this circumstance the upgrade will take more than 5-10 minutes(e.g 2000 consumers) and make a negative impact on downstream systems.

> deliveryTime can be set in the constructor like transactionID , messageID , etc :)

@wy96f 
Yes we do find similar block in PagedReferenceImpl::getScheduledDeliveryTime() since if deliveryTime is not set it will call getMessage() during rollback. 

To these two situations, detailed stacks are shown in attachment.

Considering priority only occupy one byte, it might be worthwhile to add it in PageRef to improve stability:) As for deliveryTime, since it is already in PageRef, we can simply add `this.deliveryTime = message.getMessage().getScheduledDeliveryTime();` in constructor to avoid block on rollback.

michaelandrepearce:
Im very cautious of optimising for one case
 Its very unusual for consumers to go away as it is good design in most mom a consumer is long lived.

So without it removed im -1. As i stated in an ideal world everything in paging would be off heap.

qihongxu:
> Im very cautious of optimising for one case
> Its very unusual for consumers to go away as it is good design in most mom a consumer is long lived. Like wise upgrade is a non normal event.
> 
> So without priority removed im -1 this. As i stated in an ideal world everything in paging would be off heap, we shouldnt just end up with every value in page ref, else we lose the reason for paging originals intent which is to remove the message from heap at penalty that it will be slower. If you want faster add more ram so more can be kept on heap.

Not adding priority is acceptable since it's rare. I will later remove the priority part.
And how about deliveryTime? It's already in PageRef and might need to be set in the constructor.

michaelandrepearce:
Yes you could add this to set in the section within the constructor where message != null, seems sensible.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/PagedReferenceImpl.java
```diff
@@ -120,14 +126,16 @@ public PagedReferenceImpl(final PagePosition position,
...
+         this.durable = message.getMessage().isDurable() ? IS_DURABLE : IS_NOT_DURABLE;
+         this.deliveryTime = message.getMessage().getScheduledDeliveryTime();
          //pre-cache the message size so we don't have to reload the message later if it is GC'd
          getPersistentSize();
       } else {
          this.largeMessage = UNDEFINED_IS_LARGE_MESSAGE;
          this.transactionID = -2;
          this.messageID = -1;
          this.messageSize = -1;
+         this.durable = UNDEFINED_IS_DURABLE;
```
michaelandrepearce: for completeness (its a nit) set deliveryTime to its undefined value here.

michaelandrepearce:
@qihongxu looks good to me now.

michaelandrepearce:
@franz1981 im away in another country without my main computer with my apache git ssh cert key. Could you merge this?

franz1981:
I will probably do it on the weekend or on Monday ;)

franz1981:
Please due to a change can you rebase this one?

clebertsuconic:
...merging

PR #2483, opened 2018-12-29
ARTEMIS-2215 largemessage have been consumed but not deleted

CNNJYB:
During the backup and live synchronization, the client consumes the largemessage, then the live crash(the performCachedLargeMessageDeletes method is not executed), after the live startup, the largemessages that have been consumed are not deleted from the disk.


artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -461,8 +462,7 @@ void deleteLargeMessageFile(final LargeServerMessage largeServerMessage) throws
          try {
             if (isReplicated() && replicator.isSynchronizing()) {
                synchronized (largeMessagesToDelete) {
```
franz1981: If it's now using CHM there is no need to sync on it

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
```diff
@@ -193,7 +193,7 @@ public static JournalContent getType(byte type) {
 
    protected final Map<SimpleString, PersistedAddressSetting> mapPersistedAddressSettings = new ConcurrentHashMap<>();
 
-   protected final Set<Long> largeMessagesToDelete = new HashSet<>();
+   protected final Map<Long, LargeServerMessage> largeMessagesToDelete = new ConcurrentHashMap<>();
```
franz1981: It is possible to use a primitive version of the map ie using primitive longs instead of boxed types

CNNJYB: @franz1981 modified, please review, Thanks.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
michaelandrepearce: @franz1981 looking at the code, looks like its just vanilla HM 

   protected final Map<Long, LargeServerMessage> largeMessagesToDelete = new HashMap<>();

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
```diff
@@ -193,7 +193,7 @@ public static JournalContent getType(byte type) {
 
    protected final Map<SimpleString, PersistedAddressSetting> mapPersistedAddressSettings = new ConcurrentHashMap<>();
 
-   protected final Set<Long> largeMessagesToDelete = new HashSet<>();
+   protected final Map<Long, LargeServerMessage> largeMessagesToDelete = new HashMap<>();
```
michaelandrepearce: @CNNJYB we have our own org.apache.activemq.artemis.utils.collections.LongConcurrentHashMap, this allows it to be concurrent safe, and also means it can be a primitive long.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -309,16 +309,17 @@ public void run() {
     */
    @Override
    protected void performCachedLargeMessageDeletes() {
-      for (Long largeMsgId : largeMessagesToDelete) {
-         SequentialFile msg = createFileForLargeMessage(largeMsgId, LargeMessageExtension.DURABLE);
+      for (LargeServerMessage largeServerMessage : largeMessagesToDelete.values()) {
```
michaelandrepearce: Calling values actually creates a new List, if you're iterating the objects, simply call using forEach method on the collection.

michaelandrepearce: If you wish the collection itself to be iterable, then please add this functionality to LongConcurrentHashMap implementation, it shouldn;t be too hard, as already it has a forEach method

CNNJYB: @michaelandrepearce update, please review, Thanks.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -727,7 +725,7 @@ private void checkAndCreateDir(final File dir, final boolean create) {
       List<Long> idList = new ArrayList<>();
       for (String filename : filenames) {
          Long id = getLargeMessageIdFromFilename(filename);
```
franz1981: you can just use a primitive `long` here

michaelandrepearce:
Im less familar with the journal stuff in general from what i can tell though it looks good, so will let @franz1981 give a thumbs up on the logic change, but from better use of collection libs +1 one me.

PR #2466, opened 2018-12-17
ARTEMIS-2206 The MQTT consumer reconnection caused the queue to not be cle…

onlyMIT:
### Test environment

1. Use 10,000 (9 thousand senders, 1 thousand consumers) MQTT connection on one server to test Artemis, set the 'cleanSession' property to true；
2. MQTT client: paho 1.2.0;
3. Server: CPU Cores:32, Mem:64G, SSD:250G, HDD:1T

### Issue

**Issue 1**
Artemis broker has the following exception log:
`[Thread-0 (ActiveMQ-remoting-threads-ActiveMQServerImpl::serverUUID=fb358579-feb3-11e8-bc7c-141877a7fd13-1409545055)] 17:27:59,035 WARN  [org.apache.activemq.artemis.utils.actors.OrderedExecutor] null: java.lang.NullPointerException
	at org.apache.activemq.artemis.core.protocol.mqtt.MQTTProtocolManager.isClientConnected(MQTTProtocolManager.java:182) [:]
	at org.apache.activemq.artemis.core.protocol.mqtt.MQTTConnectionManager.disconnect(MQTTConnectionManager.java:150) [:]
	at org.apache.activemq.artemis.core.protocol.mqtt.MQTTFailureListener.connectionFailed(MQTTFailureListener.java:37) [:]
	at org.apache.activemq.artemis.core.protocol.mqtt.MQTTConnection.fail(MQTTConnection.java:147) [:]
	at org.apache.activemq.artemis.core.remoting.server.impl.RemotingServiceImpl.issueFailure(RemotingServiceImpl.java:561) [:]
	at org.apache.activemq.artemis.core.remoting.server.impl.RemotingServiceImpl.connectionDestroyed(RemotingServiceImpl.java:542) [:]
	at org.apache.activemq.artemis.core.remoting.impl.netty.NettyAcceptor$Listener.connectionDestroyed(NettyAcceptor.java:858) [:]
	at org.apache.activemq.artemis.core.remoting.impl.netty.ActiveMQChannelHandler.lambda$channelInactive$0(ActiveMQChannelHandler.java:83) [:]
	at org.apache.activemq.artemis.utils.actors.OrderedExecutor.doTask(OrderedExecutor.java:42) [:]
	at org.apache.activemq.artemis.utils.actors.OrderedExecutor.doTask(OrderedExecutor.java:31) [:]
	at org.apache.activemq.artemis.utils.actors.ProcessorBase.executePendingTasks(ProcessorBase.java:66) [:]
[...]

PR #2485, opened 2018-12-29
ARTEMIS-2217 ‘MQTTSessionState’ in the ‘SESSIONS ConcurrentHashMap’ n…

onlyMIT:
…ever be removed

‘MQTTSessionState’ in the ‘SESSIONS ConcurrentHashMap’ should be removed when the conusmer （cleanSession is true） connection is closed


onlyMIT:
@jbertram  Nice! Really need more specific test

jbertram:
After looking at writing a test for this I thought the solution here was not very intuitive so I opened #2491 to address this problem (with a test).

jbertram:
@onlyMIT, can you close this PR since #2491 is in good shape now?

jbertram:
@onlyMIT, I put the `getSessionState` method back onto #2491, but #2466 will still have conflicts which you'll need to resolve once the merge on #2491 is done.

PR #2488, opened 2019-01-03
ARTEMIS-196 Implement Consumer Priority

michaelandrepearce:
Add consumer priority support
Includes refactor of consumer iterating in QueueImpl to its own logical class, to be able to implement.
Add OpenWire JMS Test - taken from ActiveMQ5
Add Core JMS Test
Add AMQP Test
Add Docs

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
@@ -16,6 +16,7 @@
  */
 package org.apache.activemq.artemis.core.protocol.core.impl.wireformat;
 
+import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
 import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
 import org.apache.activemq.artemis.api.core.SimpleString;
 
@@ -25,6 +26,8 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
 
    private SimpleString filterString;
 
+   private int priority;
+
    private boolean browseOnly;
 
    private boolean requiresResponse;
@@ -32,6 +35,7 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
    public SessionCreateConsumerMessage(final long id,
                                        final SimpleString queueName,
                                        final SimpleString filterString,
+                                       final int priority,
                                        final boolean browseOnly,
                                        final boolean requiresResponse) {
       super(SESS_CREATECONSUMER);
@@ -39,6 +43,7 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
       this.id = id;
       this.queueName = queueName;
       this.filterString = filterString;
+      this.priority = priority;
       this.browseOnly = browseOnly;
       this.requiresResponse = requiresResponse;
    }
@@ -52,6 +57,7 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
       StringBuffer buff = new StringBuffer(getParentString());
       buff.append(", queueName=" + queueName);
       buff.append(", filterString=" + filterString);
+      buff.append(", priority=" + priority);
       buff.append(", id=" + id);
       buff.append(", browseOnly=" + browseOnly);
       buff.append(", requiresResponse=" + requiresResponse);
@@ -65,6 +71,10 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
 
    public SimpleString getFilterString() {
       return filterString;
+   }
+
+   public int getPriority() {
+      return priority;
    }
 
    public boolean isBrowseOnly() {
@@ -84,6 +94,10 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
       this.filterString = filterString;
    }
 
+   public void setPriority(byte priority) {
+      this.priority = priority;
+   }
+
    public void setBrowseOnly(boolean browseOnly) {
       this.browseOnly = browseOnly;
    }
@@ -95,6 +109,7 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
       buffer.writeNullableSimpleString(filterString);
       buffer.writeBoolean(browseOnly);
       buffer.writeBoolean(requiresResponse);
+      buffer.writeInt(priority);
    }
 
    @Override
@@ -104,6 +119,11 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
       filterString = buffer.readNullableSimpleString();
       browseOnly = buffer.readBoolean();
       requiresResponse = buffer.readBoolean();
+      if (buffer.readableBytes() > 0) {
+         priority = buffer.readInt();
+      } else {
+         priority = ActiveMQDefaultConfiguration.getDefaultConsumerPriority();
+      }
    }
 
    @Override
@@ -113,6 +133,7 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
       result = prime * result + (browseOnly ? 1231 : 1237);
       result = prime * result + ((filterString == null) ? 0 : filterString.hashCode());
       result = prime * result + (int) (id ^ (id >>> 32));
+      result = prime * result + priority;
       result = prime * result + ((queueName == null) ? 0 : queueName.hashCode());
       result = prime * result + (requiresResponse ? 1231 : 1237);
       return result;
@@ -133,6 +154,8 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
          if (other.filterString != null)
             return false;
       } else if (!filterString.equals(other.filterString))
+         return false;
+      if (priority != other.priority)
          return false;
       if (id != other.id)
          return false;
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
@@ -16,11 +16,13 @@
  */
 package org.apache.activemq.artemis.protocol.amqp.broker;
 
+import java.util.Map;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
+import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
 import org.apache.activemq.artemis.api.core.ActiveMQAddressExistsException;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
 import org.apache.activemq.artemis.api.core.ActiveMQQueueExistsException;
@@ -83,6 +85,8 @@ import org.jboss.logging.Logger;
 public class AMQPSessionCallback implements SessionCallback {
 
    private static final Logger logger = Logger.getLogger(AMQPSessionCallback.class);
+
+   private static final Symbol PRIORITY = Symbol.getSymbol("priority");
 
    protected final IDGenerator consumerIDGenerator = new SimpleIDGenerator(0);
 
@@ -223,7 +227,9 @@ public class AMQPSessionCallback implements SessionCallback {
 
       filter = SelectorTranslator.convertToActiveMQFilterString(filter);
 
-      ServerConsumer consumer = serverSession.createConsumer(consumerID, queue, SimpleString.toSimpleString(filter), browserOnly, false, null);
+      int priority = getPriority(protonSender.getSender().getRemoteProperties());
+
+      ServerConsumer consumer = serverSession.createConsumer(consumerID, queue, SimpleString.toSimpleString(filter), priority, browserOnly, false, null);
 
       // AMQP handles its own flow control for when it's started
       consumer.setStarted(true);
@@ -231,6 +237,11 @@ public class AMQPSessionCallback implements SessionCallback {
       consumer.setProtocolContext(protonSender);
 
       return consumer;
+   }
+
+   private int getPriority(Map<Symbol, Object> properties) {
+      Integer value = properties == null ? null : (Integer) properties.get(PRIORITY);
+      return value == null ? ActiveMQDefaultConfiguration.getDefaultConsumerPriority() : value;
    }
 
    public void startSender(Object brokerConsumer) throws Exception {
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Consumer.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Consumer.java
@@ -21,7 +21,7 @@ import java.util.List;
 import org.apache.activemq.artemis.core.filter.Filter;
 import org.apache.activemq.artemis.spi.core.protocol.SessionCallback;
 
-public interface Consumer {
+public interface Consumer extends PriorityAware {
 
    /**
     *
@@ -80,4 +80,5 @@ public interface Consumer {
 
    /** an unique sequential ID for this consumer */
    long sequentialID();
+
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumers.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumers.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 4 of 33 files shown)


michaelandrepearce:
Ready for review. 

Ill look to merge within the week.

As noted by the Jira ticket, this is to add feature that is in ActiveMQ5 (http://activemq.apache.org/consumer-priority.html) so part of bringing feature parity, also note feature is in RabbitMQ also.

Think this might be the lowest Jira Ticket number i've worked on :)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Consumer.java
```diff
@@ -21,7 +21,7 @@
 import org.apache.activemq.artemis.core.filter.Filter;
 import org.apache.activemq.artemis.spi.core.protocol.SessionCallback;
 
-public interface Consumer {
+public interface Consumer extends PriorityAware {
```
franz1981: Just a design q: why using a specific `PriorityAware` interface?
I'm assuming that we can't have anymore a `Consumer` that doesn't provide a `default int getPriority()`

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumers.java
```diff
@@ -0,0 +1,34 @@
...
+package org.apache.activemq.artemis.core.server.impl;
+
+import org.apache.activemq.artemis.core.server.PriorityAware;
+
+import java.util.Collection;
+import java.util.Set;
+
+public interface QueueConsumers<T extends PriorityAware> extends Collection<T> {
+
+   Set<Integer> getPriorites();
```
franz1981: I know that's just a test method, so feel free to ignore me, but I would just use a byte[] here :P

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
```diff
@@ -0,0 +1,648 @@
...
+         if (midVal.level() > level)
+            low = mid + 1;
+         else if (midVal.level() < level)
+            high = mid - 1;
+         else
+            return midVal; //key found
+      }
+
+      if (createIfMissing) {
+         Level<T>[] newLevels = newLevelArrayInstance(current.length + 1);
```
franz1981: will pack this one into a separate method (if possible)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
```diff
@@ -0,0 +1,648 @@
...
+         }
+         newLevels[low] = new Level<T>(level);
+         setArray(newLevels);
+         return newLevels[low];
+      }
+      return null;
+   }
+
+   @Override
+   public synchronized boolean add(T t) {
```
franz1981: it could happen that many will access this one concurrently?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
```diff
@@ -0,0 +1,648 @@
...
+         }
+         levels = queueConsumers.getArray();
+         level = -1;
+         currentIterator = null;
+         return this;
+      }
+   }
+
+   /**
+    * This is represents a getPriority and is modeled on CopyOnArrayList.
```
franz1981: `{@link CopyOnWriteArrayList}`

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
```diff
@@ -0,0 +1,648 @@
...
+    *
+    * @param <E>
+    */
+   private static class Level<E> {
```
franz1981: I'm not a big fan of multiple layers of abstractions, but I admit that `Level` is sharing here a lot from `CopyOnWriteArratList`: we can't just reuse it somehow? It is due to the `setArray` call?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ResetableIterator.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Iterator;
+
+public interface ResetableIterator<T> extends Iterator<T> {
+
+   /**
+    * Resets the iterator so you can re-iterate over all elements.
+    *
+    * @return itself, this is just for convenience.
+    */
+   ResetableIterator<T> reset();
```
franz1981: We can extends `AutoCloseable` too and override `close`  in order to call `reset` by default too

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
```diff
@@ -25,20 +26,24 @@
 
    private SimpleString filterString;
 
+   private int priority;
```
franz1981: I will change it as `byte` if we are not planning to support priorities > 127 ie `Byte::MAX_VALUE`
For small messages it *could* save some space (TBH to be verified with JOL tool)

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
```diff
@@ -95,6 +109,7 @@ public void encodeRest(final ActiveMQBuffer buffer) {
       buffer.writeNullableSimpleString(filterString);
       buffer.writeBoolean(browseOnly);
       buffer.writeBoolean(requiresResponse);
+      buffer.writeInt(priority);
```
franz1981: I will write it as a byte, if we don't plan to support more then 127 priorities, consumer-side: but as I've said is a negligible save of space on the wire

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
```diff
@@ -233,6 +239,11 @@ public Object createSender(ProtonServerSenderContext protonSender,
       return consumer;
    }
 
+   private int getPriority(Map<Symbol, Object> properties) {
+      Integer value = properties == null ? null : (Integer) properties.get(PRIORITY);
```
franz1981: I would use a cast to `Number` and call `Number::byteValue`

franz1981:
I have taken a fast look and already written down some comment: it looks ok to me, but I need to look further in the `QueueConsumersImpl` details.
I'm happy about the nice abstraction over the consumers to simplify the message handling on `QueueImpl` and I suppose that keeping refactoring and this feature (consumer's priorities) separated isn't easy (nor sure it makes sense TBH), so I won't bother asking for it. well done! :+1:

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Consumer.java
michaelandrepearce: Correct, but to make testing of the QueueConsumersImpl logic easier and quite self contained this, its kept seperate, also it means if we wanted to reuse the new QueueConsumerImpl for other PriorityAware needs it makes it easier.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumers.java
michaelandrepearce: Priority is of Integer type, not byte.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: I tried....its not, or at least i couldnt ;)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: Yes, its following the lines on design of CopyOnWriteArrayList, but using syncronized methods over using Reentrant lock as concurrency expected is low.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: No we cant, as we need access to getArray() without a copy, which is package access only in CopyOnWriteArrayList.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2497,45 +2494,36 @@ private void deliver() {
...
                      handleMessageGroup(ref, consumer, groupConsumer, groupID);
                   }
 
+                  deliveriesInTransit.countUp();
+
+
+                  removeMessageReference(holder, ref);
+                  handledconsumer = consumer;
                   handled++;
+                  consumers.reset();
```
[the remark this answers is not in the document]
michaelandrepearce: That is not this intent, the intent in reset is just to move the iterator markers, its not a resources its purpose.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3080,45 +3053,20 @@ private boolean deliverDirect(final MessageReference ref) {
...
+         consumers.reset();
 
-         while (true) {
-            ConsumerHolder<? extends Consumer> holder;
-            if (redistributor == null) {
-               holder = consumerList.get(pos);
-            } else {
-               holder = redistributor;
-            }
+         while (consumers.hasNext() || redistributor != null) {
```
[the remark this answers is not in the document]
michaelandrepearce: Trying to keep to standard interfaces, this is the standard Iterator methods. Also at this point in the while we do not want to actually get the next, theres some timeouts and other checks needed to be done first, thus next is called a little later on.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
michaelandrepearce: We should support int, as AMQP uses can use -2^31 to 2^31-1. 

This changes nothing on the Message size, or any space concerns, this is Only on consumer creation.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
michaelandrepearce: we plan to support int.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
michaelandrepearce: Users in AMQP land will expect Integer.

Qpid -https://qpid.apache.org/releases/qpid-broker-j-7.0.6/book/Java-Broker-Runtime-Consumers.html#Java-Broker-Runtime-Consumers-Prioirty

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
franz1981: I understand it, but my point is more an idea to improve/simplify things: I understand that `CopyOnWriteArrayList` is the original starting point but here we are doing something similar, but different.
If we just use few methods of it I would instead just implement what we need, avoiding to test/provide behaviours for more complex things that we are not using from it. The same to be said from the concurrency behaviour: if we have this class designed to be used in a single writer fashing, making the code simpler (and faster most of the time), I would do this instead.
We are not implementing here any new CopyOnWriteArrayList, but a specific domain -specific class so I suppose that won't hurt to make it simpler (if possible)

michaelandrepearce:
@franz1981 alot of comments where why integer vs byte, whilst legacy ActiveMQ only supported 0-127 in open wire, many other brokers support integer for this feature, e.g. QPID for AMQP supports -2^31 to 2^31-1 like wise rabbitmq.

This is set on the consumer (its not per message) so size isn't an issue it makes sense to support the int and not constrain our selves un-neededly, making people migrating AMQP easier.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ResetableIterator.java
michaelandrepearce: -1 we are not closing the iterator, nor would this go in a try resources block., we are simply resetting the iterator so it marks the endpos = startpos, so we continue to round robin, as successfully handled a message.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: will do.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: All methods added we are using. Its not used in single write. This class essentially replaces the CopyOnWriteArrayList we used to hold the consumers, and supported only simple round robin before, and why most of its design is based on that.

franz1981:
@michaelandrepearce Yep, I have written it everytime while reading the code mate, sorry :)
Yes I do understand the bits related to byte vs int (and I agree) and related to `Iterator:.reset` too (i agree there too).
Just I'm concerned about reducing as much as possible the code complexity by having "inherited" a custom "CopyOnWriteArrayList".
I can't say if is better to simplify it as much as possible to get what we really need or just having this custom version of it as we are doing now. 
Using a final static `MethodHandle` to steal the array when we need it could be not such a bad solution too: I'm not a fan of `reflection-magic-like` things but I admit that it could avoid to copy an entire implementation just for 1 or 2 missing methods.

michaelandrepearce:
Reflection would not be good here, getArray is HOT path, its reason why i need it.

franz1981:
@michaelandrepearce Consider that a static final MethodHandle::invokeExact has performance similar to a direct call, not a reflective call http://mail.openjdk.java.net/pipermail/jigsaw-dev/2017-January/010894.html

michaelandrepearce:
@franz1981 is that java 8? I thought it was Java 9+

franz1981:
@michaelandrepearce It has been introduced into Java 8 to support the lambda as they are now, Java 9 has introduced VarHandle, that has a similar name and overlapping functionalities :P

michaelandrepearce:
@franz1981 can you check your private email, a few queries on MethodHandle ;)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: Done - resolving

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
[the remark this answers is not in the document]
michaelandrepearce: resolving as discussed else where

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ResetableIterator.java
michaelandrepearce: ill mark resolved then.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
michaelandrepearce: marking resolved.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: I assume thumbs up, means your query is resolved, ill mark this as resolved so i can keep track of open things i need to address.

franz1981:
Yep, it is a common scenario to use a security manager so I prefer to not force adding any exception on the security manger to allow just this calls to work

michaelandrepearce:
Cool so will leave as is then. thanks for the review!

michaelandrepearce:
Im going to close this for the other PR.

PR #2489, opened 2019-01-04
ARTEMIS-2220 Fix PageCursorStressTest::testSimpleCursorWithFilter NPE

franz1981:
FakeQueue is not correctly setting the queue on its PageSubscription,
leading to fail the test due to NPEs when PageSubscription::getQueue
is being used.


michaelandrepearce:
@franz1981 why not fix up FakeQueue to correctly set it....

franz1981:
@michaelandrepearce Already done! I was thinking exactly the same!

michaelandrepearce:
@franz1981 LGTM then :) merge...merge...merge

PR #2490, opened 2019-01-07
ARTEMIS-196 Implement Consumer Priority

michaelandrepearce:
@franz1981 an alternative so we don't have to have a copy of CopyOnWriteArrayList, it does mean on add or remove consumer we have to invoke toArray which causes a copy, but this is not on hot path, so i think we should be good, and avoids us having to clone a jvm class.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/MultiIterator.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/MultiIterator.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.utils.collections;
+
+import java.util.Iterator;
+
+/**
+ * Provides an Iterator that works over multiple underlying iterators.
+ *
+ * @param <T> type of the class of the iterator.
+ */
+public class MultiIterator<T> extends MultiIteratorBase<T, Iterator<T>> {
+
+   public MultiIterator(Iterator<T>[] iterators) {
+      super(iterators);
+   }
+}
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/MultiResettableIterator.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/MultiResettableIterator.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.utils.collections;
+
+/**
+ * Extends MultiIterator, adding the ability if the underlying iterators are resettable, then its self can reset.
+ * It achieves this by going back to the first iterator, and as moves to another iterator it resets it.
+ *
+ * @param <T> type of the class of the iterator.
+ */
+public class MultiResettableIterator<T> extends MultiIteratorBase<T, ResettableIterator<T>> implements ResettableIterator<T> {
+
+   public MultiResettableIterator(ResettableIterator<T>[] iterators) {
+      super(iterators);
+   }
+
+   @Override
+   protected void moveTo(int index) {
+      super.moveTo(index);
+      if (index > -1) {
+         get(index).reset();
+      }
+   }
+
+   @Override
+   public void reset() {
+      moveTo(-1);
+   }
+}
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityCollection.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityCollection.java
@@ -1,0 +1,324 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.utils.collections;
+
+import org.apache.activemq.artemis.core.PriorityAware;
+
+import java.lang.reflect.Array;
+import java.util.AbstractCollection;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Objects;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+/**
+ * This class's purpose is to hold the the different collections used for each priority level.
+ *
+ * A supplier is required to provide the underlying collection needed when a new priority level is seen,
+ * and the end behaviour is that of the underlying collection, e.g. if set add will follow set's add semantics,
+ * if list, then list semantics.
+ *
+ * Methods getArray, setArray MUST never be exposed, and all array modifications must go through these.
+ *
+ * @param <T> The type this class may hold, this is generic as can be anything that extends PriorityAware.
+ */
+public class PriorityCollection<T extends PriorityAware> extends AbstractCollection<T> {
+
+   private final Supplier<Collection<T>> supplier;
+   private volatile PriorityHolder<T>[] priorityHolders = newPrioritySetArrayInstance(0);
+   private volatile int size;
+
+   private void setArray(PriorityHolder<T>[] priorityHolders) {
+      this.priorityHolders = priorityHolders;
+   }
+
+   private PriorityHolder<T>[] getArray() {
+      return priorityHolders;
+   }
+
+
+   public PriorityCollection(Supplier<Collection<T>> supplier) {
+      this.supplier = supplier;
+   }
+
+   @SuppressWarnings("unchecked")
+   private static <T> PriorityHolder<T>[] newPrioritySetArrayInstance(int length) {
+      return (PriorityHolder<T>[]) Array.newInstance(PriorityHolder.class, length);
+   }
+
+   @Override
+   public int size() {
+      return size;
+   }
+
+   @Override
+   public boolean isEmpty() {
+      return size() == 0;
+   }
+
+   public Set<Integer> getPriorites() {
+      PriorityHolder<T>[] snapshot = getArray();
+      return Arrays.stream(snapshot).map(PriorityAware::getPriority).collect(Collectors.toSet());
+   }
+
+   @Override
+   public Iterator<T> iterator() {
+      Iterator<T>[] iterators = getIterators();
+      return new MultiIterator<>(iterators);
+   }
+
+   private Iterator<T>[] getIterators() {
+      PriorityHolder<T>[] snapshot = this.getArray();
+      int size = snapshot.length;
+      Iterator<T>[] iterators = newIteratorArrayInstance(size);
+      for (int i = 0; i < size; i++) {
+         iterators[i] = snapshot[i].getValues().iterator();
+      }
+      return iterators;
+   }
+
+   @SuppressWarnings("unchecked")
+   private static <T> Iterator<T>[] newIteratorArrayInstance(int length) {
+      return (Iterator<T>[]) Array.newInstance(Iterator.class, length);
+   }
+
+   public ResettableIterator<T> resettableIterator() {
+      return new MultiResettableIterator<T>(getResettableIterators());
+   }
+
+   private ResettableIterator<T>[] getResettableIterators() {
+      PriorityHolder<T>[] snapshot = this.getArray();
+      int size = snapshot.length;
+      ResettableIterator<T>[] iterators = newResettableIteratorArrayInstance(size);
+      for (int i = 0; i < size; i++) {
+         iterators[i] = ArrayResettableIterator.iterator(snapshot[i].getValues());
+      }
+      return iterators;
+   }
+
+   @SuppressWarnings("unchecked")
+   private static <T> ResettableIterator<T>[] newResettableIteratorArrayInstance(int length) {
+      return (ResettableIterator<T>[]) Array.newInstance(ResettableIterator.class, length);
+   }
+
+   @Override
+   public void forEach(Consumer<? super T> action) {
+      Objects.requireNonNull(action);
+      PriorityHolder<T>[] current = getArray();
+      int len = current.length;
+      for (int i = 0; i < len; ++i) {
+         current[i].getValues().forEach(action);
+      }
+   }
+
+   private Collection<T> getCollection(int priority, boolean createIfMissing) {
+      PriorityHolder<T>[] current = getArray();
+      int low = 0;
+      int high = current.length - 1;
+
+      while (low <= high) {
+         int mid = (low + high) / 2;
+         PriorityHolder<T> midVal = current[mid];
+
+         if (midVal.getPriority() > priority)
+            low = mid + 1;
+         else if (midVal.getPriority() < priority)
+            high = mid - 1;
+         else
+            return midVal.getValues(); //key found
+      }
+
+      if (createIfMissing) {
+         PriorityHolder<T>[] newArray = newPrioritySetArrayInstance(current.length + 1);
+         if (low > 0) {
+            System.arraycopy(current, 0, newArray, 0, low);
+         }
+         if (current.length - low > 0) {
+            System.arraycopy(current, low, newArray, low + 1, current.length - low);
+         }
+         newArray[low] = new PriorityHolder<T>(priority, supplier);
+         setArray(newArray);
+         return newArray[low].getValues();
+      }
+      return null;
+   }
+
+   @Override
+   public synchronized boolean add(T t) {
+      if (size() == Integer.MAX_VALUE) return false;
+      boolean result = addInternal(t);
+      calcSize();
+      return result;
+   }
+
+   private boolean addInternal(T t) {
+      if (t == null) return false;
+      Collection<T> priority = getCollection(t.getPriority(), true);
+      return priority.add(t);
+   }
+
+   @Override
+   public synchronized boolean remove(Object o) {
+      boolean result = removeInternal(o);
+      calcSize();
+      return result;
+   }
+
+   private boolean removeInternal(Object o) {
+      if (o instanceof PriorityAware) {
+         PriorityAware priorityAware = (PriorityAware) o;
+         Collection<T> priority = getCollection(priorityAware.getPriority(), false);
+         boolean result = priority != null && priority.remove(priorityAware);
+         if (priority != null && priority.isEmpty()) {
+            removeCollection(priorityAware.getPriority());
+         }
+         return result;
+      } else {
+         return false;
+      }
+   }
+
+   private Collection<T> removeCollection(int priority) {
+      PriorityHolder<T>[] current = getArray();
+      int len = current.length;
+      int low = 0;
+      int high = len - 1;
+
+      while (low <= high) {
+         int mid = (low + high) / 2;
+         PriorityHolder<T> midVal = current[mid];
+
+         if (midVal.getPriority() > priority)
+            low = mid + 1;
+         else if (midVal.getPriority() < priority)
+            high = mid - 1;
+         else {
+            PriorityHolder<T>[] newArray = newPrioritySetArrayInstance(len - 1);
+            System.arraycopy(current, 0, newArray, 0, mid);
+            System.arraycopy(current, mid + 1, newArray, mid, len - mid - 1);
+            setArray(newArray);
+            return midVal.getValues(); //key found
+         }
+      }
+      return null;
+   }
+
+   @Override
+   public boolean containsAll(Collection<?> c) {
+      Objects.requireNonNull(c);
+      for (Object e : c)
+         if (!contains(e))
+            return false;
+      return true;
+   }
+
+   @Override
+   public synchronized boolean addAll(Collection<? extends T> c) {
+      Objects.requireNonNull(c);
+      if (size() >= Integer.MAX_VALUE - c.size()) return false;
+      boolean modified = false;
+      for (T e : c)
+         if (addInternal(e))
+            modified = true;
+      calcSize();
+      return modified;
+   }
+
+   @Override
+   public synchronized boolean removeAll(Collection<?> c) {
+      Objects.requireNonNull(c);
+      boolean modified = false;
+      for (Object o : c) {
+         if (removeInternal(o)) {
+            modified = true;
+         }
+      }
+      calcSize();
+      return modified;
+   }
+
+   @Override
+   public synchronized boolean retainAll(Collection<?> c) {
+      Objects.requireNonNull(c);
+      boolean modified = false;
+      PriorityHolder<T>[] snapshot = getArray();
+      for (PriorityHolder<T> priorityHolder : snapshot) {
+         if (priorityHolder.getValues().retainAll(c)) {
+            modified = true;
+            if (priorityHolder.getValues().isEmpty()) {
+               removeCollection(priorityHolder.getPriority());
+            }
+         }
+      }
+      calcSize();
+      return modified;
+   }
+
+   @Override
+   public synchronized void clear() {
+      PriorityHolder<T>[] snapshot = getArray();
+      for (PriorityHolder<T> priorityHolder : snapshot) {
+         priorityHolder.getValues().clear();
+      }
+      calcSize();
+   }
+
+   @Override
+   public boolean contains(Object o) {
+      return o instanceof PriorityAware && contains((PriorityAware) o);
+   }
+
+   public boolean contains(PriorityAware priorityAware) {
+      if (priorityAware == null) return false;
+      Collection<T> prioritySet = getCollection(priorityAware.getPriority(), false);
+      return prioritySet != null && prioritySet.contains(priorityAware);
+   }
+
+   private void calcSize() {
+      PriorityHolder<T>[] current = getArray();
+      int size = 0;
+      for (PriorityHolder<T> priorityHolder : current) {
+         size += priorityHolder.getValues().size();
+      }
+      this.size = size;
+   }
+
+   public static class PriorityHolder<E> implements PriorityAware {
+
+      private final int priority;
+
+      private final Collection<E> values;
+
+      public PriorityHolder(int priority, Supplier<Collection<E>> supplier) {
+         this.priority = priority;
+         this.values = supplier.get();
+      }
+
+      @Override
+      public int getPriority() {
+         return priority;
+      }
+
+      public Collection<E> getValues() {
+         return values;
+      }
+   }
+}
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
@@ -16,14 +16,18 @@
  */
 package org.apache.activemq.artemis.core.protocol.core.impl.wireformat;
 
+import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
 import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
 import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.core.protocol.core.CoreRemotingConnection;
 
 public class SessionCreateConsumerMessage extends QueueAbstractPacket {
 
    private long id;
 
    private SimpleString filterString;
+
+   private int priority;
 
    private boolean browseOnly;
 
@@ -32,6 +36,7 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
    public SessionCreateConsumerMessage(final long id,
                                        final SimpleString queueName,
                                        final SimpleString filterString,
+                                       final int priority,
                                        final boolean browseOnly,
                                        final boolean requiresResponse) {
       super(SESS_CREATECONSUMER);
@@ -39,6 +44,7 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
       this.id = id;
       this.queueName = queueName;
       this.filterString = filterString;
+      this.priority = priority;
       this.browseOnly = browseOnly;
       this.requiresResponse = requiresResponse;
    }
@@ -55,6 +61,7 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
       buff.append(", id=" + id);
       buff.append(", browseOnly=" + browseOnly);
       buff.append(", requiresResponse=" + requiresResponse);
+      buff.append(", priority=" + priority);
       buff.append("]");
       return buff.toString();
    }
@@ -65,6 +72,10 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
 
    public SimpleString getFilterString() {
       return filterString;
+   }
+
+   public int getPriority() {
+      return priority;
    }
 
    public boolean isBrowseOnly() {
@@ -84,17 +95,25 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
       this.filterString = filterString;
    }
 
+   public void setPriority(byte priority) {
+      this.priority = priority;
+   }
+
    public void setBrowseOnly(boolean browseOnly) {
       this.browseOnly = browseOnly;
    }
 
    @Override
-   public void encodeRest(final ActiveMQBuffer buffer) {
+   public void encodeRest(final ActiveMQBuffer buffer, final CoreRemotingConnection coreRemotingConnection) {
       buffer.writeLong(id);
       buffer.writeSimpleString(queueName);
       buffer.writeNullableSimpleString(filterString);
       buffer.writeBoolean(browseOnly);
       buffer.writeBoolean(requiresResponse);
+      //Priority Support added in 2.7.0
+      if (coreRemotingConnection.isVersionSupportConsumerPriority()) {
+         buffer.writeInt(priority);
+      }
    }
 
    @Override
@@ -104,6 +123,12 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
       filterString = buffer.readNullableSimpleString();
       browseOnly = buffer.readBoolean();
       requiresResponse = buffer.readBoolean();
+      //Priority Support Added in 2.7.0
+      if (buffer.readableBytes() > 0) {
+         priority = buffer.readInt();
+      } else {
+         priority = ActiveMQDefaultConfiguration.getDefaultConsumerPriority();
+      }
    }
 
    @Override
@@ -113,6 +138,7 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
       result = prime * result + (browseOnly ? 1231 : 1237);
       result = prime * result + ((filterString == null) ? 0 : filterString.hashCode());
       result = prime * result + (int) (id ^ (id >>> 32));
+      result = prime * result + priority;
       result = prime * result + ((queueName == null) ? 0 : queueName.hashCode());
       result = prime * result + (requiresResponse ? 1231 : 1237);
       return result;
@@ -133,6 +159,8 @@ public class SessionCreateConsumerMessage extends QueueAbstractPacket {
          if (other.filterString != null)
             return false;
       } else if (!filterString.equals(other.filterString))
+         return false;
+      if (priority != other.priority)
          return false;
       if (id != other.id)
          return false;
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
@@ -16,8 +16,10 @@
  */
 package org.apache.activemq.artemis.protocol.amqp.broker;
 
+import java.util.Map;
 import java.util.concurrent.Executor;
 
+import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
 import org.apache.activemq.artemis.api.core.ActiveMQAddressExistsException;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
 import org.apache.activemq.artemis.api.core.ActiveMQQueueExistsException;
@@ -79,6 +81,8 @@ import org.jboss.logging.Logger;
 public class AMQPSessionCallback implements SessionCallback {
 
    private static final Logger logger = Logger.getLogger(AMQPSessionCallback.class);
+
+   private static final Symbol PRIORITY = Symbol.getSymbol("priority");
 
    protected final IDGenerator consumerIDGenerator = new SimpleIDGenerator(0);
 
@@ -199,7 +203,9 @@ public class AMQPSessionCallback implements SessionCallback {
 
       filter = SelectorTranslator.convertToActiveMQFilterString(filter);
 
-      ServerConsumer consumer = serverSession.createConsumer(consumerID, queue, SimpleString.toSimpleString(filter), browserOnly, false, null);
+      int priority = getPriority(protonSender.getSender().getRemoteProperties());
+
+      ServerConsumer consumer = serverSession.createConsumer(consumerID, queue, SimpleString.toSimpleString(filter), priority, browserOnly, false, null);
 
       // AMQP handles its own flow control for when it's started
       consumer.setStarted(true);
@@ -207,6 +213,11 @@ public class AMQPSessionCallback implements SessionCallback {
       consumer.setProtocolContext(protonSender);
 
       return consumer;
+   }
+
+   private int getPriority(Map<Symbol, Object> properties) {
+      Number value = properties == null ? null : (Number) properties.get(PRIORITY);
+      return value == null ? ActiveMQDefaultConfiguration.getDefaultConsumerPriority() : value.intValue();
    }
 
    public void startSender(Object brokerConsumer) throws Exception {
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
@@ -1,0 +1,127 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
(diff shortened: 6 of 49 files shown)


franz1981:
@michaelandrepearce Nice! Will take a look today or max tomorrow :+1:

michaelandrepearce:
@franz1981 just ignore class comments, theyre the originals still, ill need to change, but wanted to get to you quickly so you have chance to look over. If you think this is better ill make final tidyup bits, such as class comments and replace the real PR's branch.

michaelandrepearce:
@franz1981 did you get a chance to look, do you think this is better than original solution? 

Am keen to get this feature into the next release cut.

franz1981:
@mochaelandrepearce sadly haven't had much time today to look into it :( 
Tomorrow I have already scheduled some time in the morning to take a look into this

michaelandrepearce:
@franz1981 great thanks a million :)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/amq/QueueConsumerPriorityTest.java
```diff
@@ -0,0 +1,65 @@
...
+      ActiveMQQueue senderQueue = new ActiveMQQueue(queueName);
+
+      MessageProducer producer = senderSession.createProducer(senderQueue);
+
+      Message msg = senderSession.createTextMessage("test");
+      for (int i = 0; i < 1000; i++) {
+         producer.send(msg);
+         assertNotNull("null on iteration: " + i, highConsumer.receive(1000));
+      }
+      assertNull(lowConsumer.receive(2000));
```
gemmellr: Would a receiveNoWait (either in or outside the loop) like the other tests be nicer than burning 2 seconds? Slow tests is a key reason eventually noone wants to runs the tests :)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpReceiverPriorityTest.java
```diff
@@ -0,0 +1,90 @@
...
+      properties3.put(Symbol.getSymbol("priority"), 5);
+      AmqpReceiver receiver3 = session.createReceiver(getQueueName(), null, false, false, properties3);
+      receiver3.flow(100);
+
+      sendMessages(getQueueName(), 5);
+
+
+      for (int i = 0; i < 5; i++) {
+         AmqpMessage message1 = receiver1.receive(250, TimeUnit.MILLISECONDS);
+         AmqpMessage message2 = receiver2.receive(250, TimeUnit.MILLISECONDS);
```
gemmellr: Burning 250ms twice per loop seems excessive. There is a receiveNoWait that could be used for initial verification nothing arrived, and/or a small final timed wait could be done outside the loop afterwards. Alternatively, pullImmediate() would avoid unnecessary waiting.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpReceiverPriorityTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+      //Close the high priority receiver
+      receiver1.close();
+
+      sendMessages(getQueueName(), 5);
+
+      //Check messages now goto next priority receiver
+      for (int i = 0; i < 5; i++) {
+         AmqpMessage message2 = receiver2.receive(250, TimeUnit.MILLISECONDS);
+         AmqpMessage message3 = receiver3.receive(250, TimeUnit.MILLISECONDS);
```
gemmellr: As above.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpReceiverPriorityTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+   @Test(timeout = 30000)
+   public void testPriority() throws Exception {
+
+      AmqpClient client = createAmqpClient();
+      AmqpConnection connection = addConnection(client.connect());
+      AmqpSession session = connection.createSession();
+
+      Map<Symbol, Object> properties1 = new HashMap<>();
+      properties1.put(Symbol.getSymbol("priority"), 50);
```
gemmellr: I'd suggest creating consumers with priorities out of order (e.g highest in middle), so they arent simply registered in sequence, as otherwise a simple failure to round-robin delivery attempts (given every receiver has enough credit to receive all messages) might also lead to the expected result even without any priority handling consideration.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
```diff
@@ -52,6 +57,7 @@ public String toString() {
       StringBuffer buff = new StringBuffer(getParentString());
       buff.append(", queueName=" + queueName);
       buff.append(", filterString=" + filterString);
+      buff.append(", priority=" + priority);
```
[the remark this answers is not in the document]
michaelandrepearce: Makes sense

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
```diff
@@ -104,6 +119,11 @@ public void decodeRest(final ActiveMQBuffer buffer) {
       filterString = buffer.readNullableSimpleString();
       browseOnly = buffer.readBoolean();
       requiresResponse = buffer.readBoolean();
+      if (buffer.readableBytes() > 0) {
```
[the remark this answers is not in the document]
michaelandrepearce: This is typical pattern used for adding safely a new field that can be either nullable or defaultable. Used many times over.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpReceiverPriorityTest.java
michaelandrepearce: This is actually tested on the queueconsumerimpl test. But agree we can do same here

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/amq/QueueConsumerPriorityTest.java
michaelandrepearce: this is the original test from ActiveMQ5 i was trying to keep this test as much un-touched as possible to ensure behavior is the same.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/client/ConsumerPriorityTest.java
```diff
@@ -0,0 +1,298 @@
...
+import javax.jms.Session;
+import javax.jms.TextMessage;
+import javax.jms.Topic;
+
+/**
+ * Exclusive Test
+ */
+public class ConsumerPriorityTest extends JMSTestBase {
+
+   private SimpleString queueName = SimpleString.toSimpleString("jms.consumer.priority.queue");
```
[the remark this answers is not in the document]
michaelandrepearce: Nice i didnt know about that in the parent class. will change to use this..

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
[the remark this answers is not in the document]
michaelandrepearce: done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
```diff
@@ -233,6 +239,11 @@ public Object createSender(ProtonServerSenderContext protonSender,
       return consumer;
    }
 
+   private int getPriority(Map<Symbol, Object> properties) {
+      Integer value = properties == null ? null : (Integer) properties.get(PRIORITY);
```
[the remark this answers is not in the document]
michaelandrepearce: Have changed if you can recheck to make sure i understood.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/client/ConsumerPriorityTest.java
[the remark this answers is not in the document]
michaelandrepearce: now using getName ... again nice nice....

michaelandrepearce:
@gemmellr thanks for review, if you could recheck the AMQPSessionCallback for me, to make sure i understood you. 

As for the Openwire test case, this was a simple port over of the existing activemq5 test case as untouched as possible, i agree we could reduce the time but id rather (at least for this release) keep it the same, so we can be sure feature works for openwire same as activemq5.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
```diff
@@ -0,0 +1,124 @@
...
+   }
+
+   @Override
+   public T next() {
+      return currentIterator.next();
+   }
+
+   @Override
+   public QueueConsumers<T> reset() {
+      if (changedIterator != null) {
```
franz1981: I will just save `changedIterator` once and will perform a logic with a local value instead of volatile load twice

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
```diff
@@ -0,0 +1,124 @@
...
+         currentIterator = changedIterator;
+         changedIteratorFieldUpdater.compareAndSet(this, changedIterator, null);
```
franz1981: What happen if the cas will fail? we ends with a reset() that is not nulling `changedIterator`.
Given that we are just clearing the `changedIterator` to `null` a `lazySet(this, null) is enough.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
```diff
@@ -0,0 +1,124 @@
...
+      } else {
+         currentIterator.reset();
```
franz1981: I suppose that `reset` is safe to be called just by one thread at time, if not, it would be complex because `currentIterator` could be changed indipendently

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
```diff
@@ -0,0 +1,124 @@
...
+      }
+      return this;
+   }
+
+   @Override
+   public boolean add(T t) {
+      boolean result = consumers.add(t);
+      if (result) {
+         changedIteratorFieldUpdater.set(this, consumers.resettableIterator());
```
franz1981: `lazeSet` is enough for single-writer/single-threaded semantic

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
```diff
@@ -0,0 +1,124 @@
...
+      }
+      return result;
+   }
+
+   @Override
+   public boolean remove(T t) {
+      boolean result = consumers.remove(t);
+      if (result) {
+         changedIteratorFieldUpdater.set(this, consumers.resettableIterator());
```
franz1981: `lazeSet` is enough for single-writer/single-threaded semantic

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2409,14 +2410,10 @@ private void deliver() {
...
 
-      int size = 0;
-
-      int endPos = -1;
-
       int handled = 0;
 
       long timeout = System.currentTimeMillis() + DELIVERY_TIMEOUT;
-
+      consumers.reset();
```
franz1981: if any exception would be thrown before it is ok that `reset` won't be called? 
If not, better to wrap the whole logic with ìtry..finally` and `reset`

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2497,45 +2484,36 @@ private void deliver() {
...
                      handleMessageGroup(ref, consumer, groupConsumer, groupID);
                   }
 
+                  deliveriesInTransit.countUp();
+
+
+                  removeMessageReference(holder, ref);
+                  handledconsumer = consumer;
                   handled++;
+                  consumers.reset();
```
franz1981: if `removeMessageReference` would throw any exeption is fine to have `consumers.reset` not called?
if not, uses `try...finally`

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityCollection.java
```diff
@@ -0,0 +1,319 @@
...
+         }
+         newArray[low] = new PriorityHolder<T>(priority, supplier);
+         setArray(newArray);
+         return newArray[low].getValues();
+      }
+      return null;
+   }
+
+   @Override
+   public synchronized boolean add(T t) {
```
franz1981: Looking to the usage on `QueueImpl` it seems that is being always called into a `synchronized (this) {` block ie it doesn't need to be `synchronized` and same same for `remove`. 
If we allow `snapshotting` from a different thread is fine, but I would document it instead.
I know that's a less safe approach than putting `sychronized` all around, but I do believe that we would have to do the same on  `LinkedListImpl` for this same reason if we need to `protect` correct usage at any cost.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
```diff
@@ -113,6 +133,7 @@ public int hashCode() {
       result = prime * result + (browseOnly ? 1231 : 1237);
       result = prime * result + ((filterString == null) ? 0 : filterString.hashCode());
       result = prime * result + (int) (id ^ (id >>> 32));
+      result = prime * result + priority;
```
franz1981: I would uses `Integer::hashCode(priority)` that would `agitate` the value

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpReceiverPriorityTest.java
```diff
@@ -0,0 +1,144 @@
...
+      Map<Symbol, Object> properties3 = new HashMap<>();
+      properties3.put(Symbol.getSymbol("priority"), 10);
+      AmqpReceiver receiver3 = session.createReceiver(getQueueName(), null, false, false, properties3);
+      receiver3.flow(100);
+
+      sendMessages(getQueueName(), 5);
+
+
+      for (int i = 0; i < 5; i++) {
+         AmqpMessage message1 = receiver1.receiveNoWait();
```
gemmellr: I believe the test clients receiveNoWait only polls its local queue, so this might now likely sporadically fail due to racing the deliveries. I was only suggesting recieveNoWait as potential initial verification within the loop for non-receiving consumers, to be followed up or substituted by a more stringent final check.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityCollection.java
michaelandrepearce: This needs to be concurrent safe on modifications

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelandrepearce: Yes its fine. Would have no negative effect, and actually have same behaviour as old.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelandrepearce: Its fine. We actually only want to reset on a succesful handled nomatch or expired. Or before starting iterating

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: Agreed. But this is not hot path, and if anything would allow in future for consumers to be concurrently added or removed without the existing bit sync blocks there is today in queueimpl

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: Like any iterator and iterator should only be interacted by one thread at a time.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: If it fails it means whilst we were reseting and changing reference to the new iterator. a consumer was added or removed and another new iterator was created (as this can occur whilsy we are iterating or resetting.) And means simply whilst this reset updated its iterator, on next reset it needs to switch over its again. We wouldnt want to lazy set to null

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionCreateConsumerMessage.java
```diff
@@ -84,6 +94,10 @@ public void setFilterString(SimpleString filterString) {
       this.filterString = filterString;
    }
 
+   public void setPriority(byte priority) {
```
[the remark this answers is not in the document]
michaelandrepearce: Keeping inline with the rest of the fields and general approach. Id rather keep to status quo

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: Nice

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
franz1981: So it is a good reason to save using a `compareAndSet` that's not only expensive but give the false illusion that the code is thread-safe

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityCollection.java
```diff
@@ -0,0 +1,319 @@
...
+      if (priorityAware == null) return false;
+      Collection<T> prioritySet = getCollection(priorityAware.getPriority(), false);
+      return prioritySet != null && prioritySet.contains(priorityAware);
+   }
+
+   private void calcSize() {
+      PriorityHolder<T>[] current = getArray();
+      int size = 0;
+      for (PriorityHolder<T> priorityHolder : current) {
+         size += priorityHolder.getValues().size();
```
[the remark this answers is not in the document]
michaelandrepearce: If we have a case of the number of active consumers on a single queue is integer max value we will have many areas to address tbh! And that is some super extreme case.

We can reject on adding a new consumer if adding the new consumer would breach int max value, if we are really worried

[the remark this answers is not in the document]
michaelandrepearce: If we have a case of the number of active consumers on a single queue is integer max value we will have many areas to address tbh! And that is some super extreme case.

We can reject on adding a new consumer if adding the new consumer would breach int max value, if we are really worried

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityCollection.java
```diff
@@ -0,0 +1,319 @@
...
+   }
+
+   private Collection<T> removeCollection(int priority) {
+      PriorityHolder<T>[] current = getArray();
+      int len = current.length;
+      int low = 0;
+      int high = len - 1;
+
+      while (low <= high) {
+         int mid = (low + high) >>> 1;
```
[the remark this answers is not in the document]
michaelandrepearce: This is just coding pref for this stuff im happier with shift operations. Its how alot of the jvm classes do this.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/MultiResettableIterator.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package org.apache.activemq.artemis.utils.collections;
+
+/**
+ * Extends MultiIterator, adding the ability if the underlying iterators are resettable, then its self can reset.
+ * It achieves this by going back to the first iterator, and as moves to another iterator it resets it.
+ *
+ * @param <T> type of the class of the iterator.
+ */
+public class MultiResettableIterator<T> extends MultiIterator<T> implements ResettableIterator<T> {
```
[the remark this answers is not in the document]
michaelandrepearce: You could but this then makes using MutiIterator ugly as hell with little gain

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/MultiIterator.java
```diff
@@ -0,0 +1,78 @@
...
+package org.apache.activemq.artemis.utils.collections;
+
+import java.util.Iterator;
+
+/**
+ * Provides an Iterator that works over multiple underlying iterators.
+ *
+ * @param <T> type of the class of the iterator.
+ */
+public class MultiIterator<T> implements Iterator<T> {
```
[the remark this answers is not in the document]
michaelandrepearce: Yeah i had a version like that but it makes using simple MultiIterator ugly to use. Uneededly. And to extend if needed only needs a cast. In case of resetable.

michaelandrepearce:
@gemmellr ok will reduce back down the extra tests i added. To those you suggest. 

Also re activemq5. Fair enough, ill reduce / alter this then.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/MultiResettableIterator.java
```diff
@@ -0,0 +1,44 @@
...
+
+   public MultiResettableIterator(ResettableIterator<T>[] iterators) {
+      super(iterators);
+   }
+
+   @Override
+   protected void moveTo(int index) {
+      super.moveTo(index);
+      if (index > -1) {
+         ((ResettableIterator<T>) get(index)).reset();
```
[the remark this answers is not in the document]
michaelandrepearce: It could but changing MultiIterators generics then makes it ugly for any use of it. Id rather ensure usage of the classes is kept clean over their internals. And a single cast isnt a biggie imo for the benefit of a cleaner usage by the end using code that uses it.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/MultiIterator.java
```diff
@@ -0,0 +1,78 @@
...
+
+   private final Iterator<T>[] iterators;
+   int index = -1;
```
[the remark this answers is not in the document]
michaelandrepearce: Agreed

michaelandrepearce:
@franz1981 good point around adding further tests to the extracted out bits. Agree it will make everything more robust

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: The compare and set is because the new iterator could be created whilst we are mid resetting. The iterator is single thread usage only. Updating the collection should be thread safe and able to occur in parralell.

This is much like iterating copyonwriteareatlist. Iterator works on a snapshot view and this is single thread. Updating the collection itself can be concurrent and occur whilst someone else is iterating

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/MultiIterator.java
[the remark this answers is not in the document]
michaelandrepearce: OK, so to do this and keep the cleaner user interface what i can do is make an abstract base which has most of this logic in, and have the more ugly generics. And then make MultiIterator extend that which then keeps its cleaner generics, and like wise MultiResettableIterator. Ill update PR to see if this is good compromise, or if you'd prefer me to revert and simply keep the cast i had originally.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: see other comment

michaelandrepearce:
@gemmellr pushed changes based on latest comments, thanks.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityCollection.java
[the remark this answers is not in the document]
franz1981: I know but the reason is because some optimisations were not present in old JDK version and the classes aren't changed...

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: currentIterator is ONLY updated by reset, which should ONLY be called by the same threads operating on the ResettableIterator interface.

When add or remove of consumer occurs a new iterator is parked into the volatile changedIterator (using an atomic field updater), so the next reset can pick it up.

michaelandrepearce:
@franz1981 pushed following changes based on your comments that agree with, for others i have left comment for us to discuss. 

Changes:
1) Ensured total size of the priority collection can never exceed Integer.MAX_VALUE, by ensuring this on add, thus the edge case you were worried about of the calcSize being greater than int cannot occur. Note if someone has that many consumers, we probably want to have some discussions with them as they would be some power user!!! ;)

2)  Avoid double volatile read of changedIterator in reset method.

3) Removed need for a cast in MultiResettableIterator

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityCollection.java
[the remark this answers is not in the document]
michaelandrepearce: still just a coding preference, i see you even do bitwise alot when can be just written in standard math ;) ..... 

anyhow have changed to avoid disagreement.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: Thats fair comment. Ill add this.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/amq/QueueConsumerPriorityTest.java
michaelandrepearce: changed.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
michaelandrepearce: ill mark comment resolved, if you think it needs further docs or could be further abstracted to be more clearer let me know.

gemmellr:
@ Thanks. I'd still rather some of the tests weren't burning 250ms various times waiting to test something that can be verified 'immediately', but its a lot better.

I'm still not sure if you've seen this comment (github is hiding it as 'resolved') so I'll copy it here:

https://github.com/apache/activemq-artemis/pull/2490/files#r246398611
> Does this comment cover to the bit around the safety of always sending the new additional data even to old servers? I can't tell if its covered.
>
> I think it should at the very least be commented what/when the encoding+decoding handling behaviour changed so folks can understand the implications later without heading to find past commits.


which was related to part of an earlier comment https://github.com/apache/activemq-artemis/pull/2490/files#r245955337), mostly:
> Also, is the reverse case safe, does an older server failing to read the additional value (seemingly always sent now) have potential to lead to any issues on older servers, i.e how might the buffer continue to be used later if at all? Should the client omit the value for older servers? (Or does the presumed version change prevent the new client working with the old server anyway? I don't know how that stuff is handled, just commenting from reading the diff here).

michaelandrepearce:
@gemmellr so this is the same pattern used quite a bit, e.g. if you look at in master: ClusterTopologyChangeMessage_V2 it was used when adding backupGroupName, like wise you will see similar in other XXXMessage.

By the extra bits only being at the end, older servers simply read the old fields out it just doesnt read the extra byte. And having the readableBytes check means if the reverse where an older client wont have provided the extra bytes, then this is where we default it.

There is also a bunch on compatibility tests that @clebertsuconic made, that if there was a compatibility break it would be picked up.

gemmellr:
@michaelandrepearce I'd have made the same comment for those changes if reviewing them, as I would probably have to go digging through commits now if I wanted to understand the evolution there for subsequent changes, when at least a simple comment might have done the trick. Giving up on this point though, enough time wasted looking to make things clearer for later readers to avoid people needing to waste time :)

Its more the other side of the comment that primarily concerns me, that the new client always sends the extra detail to old servers despite knowing they cant use it. The fact the old server then doesnt read out the extra info in the buffer is what I find most concerning, as the data is going to be there in the buffer, and so how that buffer is later used by the old server becomes important following the changes even though it doesnt support the feature. The basic compatibility tests dont really ease that concern. It just seems simpler and safer to ensure ongoing compatibility isnt affected by not sending old servers new data we know they cant use, but could perhaps trip over in unexpected ways.

Looking at the existing code now rather than just the diff here, it seems the very first thing it does after decoding is use the current buffer index as a 'size', which then might not be accurate any more depending on how its being used or how it interacts with other areas later. I haven't traced further usage of the buffer or that size, but I personally wouldn't be comfortable with the new data being sent to old servers based on that alone. Who knows what things other areas and even older different versions of the code might do in the face of changing the buffer content.

michaelandrepearce:
@gemmellr as i stated, this approach has been used many times over, historically and is obviosuly proven, this is not something i invented, im just following the same working pattern.

gemmellr:
Sometimes tests don't pick up issues that may exist and we later find issues prompting more specific tests. You essentially don't need to rely on the compatibility test in cases like this though if you simply don't change the behaviour. Given the old servers are known not to support the feature, changing behaviour against them when not needed them doesn't seem beneficial at all, but could be detrimental. 

If an apparent issue on the first line executed after failing to read new-but-unusable-data sent to an old servers isn't enough reason for pause here, I'm don't know what would be.

michaelandrepearce:
@gemmellr a created a new SessionCreateConsumerMessage_V2 so hopefully this addresses this and added a version check on channel.

also addressed merge conflicts after recent merges to master from other prs.

gemmellr:
I'd have been content with an if and comment in the original class, but happy with the class if that is convention.

michaelandrepearce:
Oh, ok ill try do that.

michaelandrepearce:
Ok, I put back to original class, and added an if check to it, with comments.

Squashed everything also.

michaelandrepearce:
@franz1981 i know i still owe you those tests...

michaelandrepearce:
@franz1981 added the tests i owed you. Sorry took some time, had the weekend off.

michaelandrepearce:
@franz1981 good to go?

clebertsuconic:
Great work here!


I'm running testsuite just to make sure everything is covered.

clebertsuconic:
I tried pulling your branch, but the conflict that I thought I fixed caused a compilation error.


do you mind rebasing it please?

michaelandrepearce:
rebased

clebertsuconic:
one test failure at least: JMSMessageGroupsTest


The testsuite didn't complete.. but that one is failing for sure.

clebertsuconic:
I will spend some time understanding the failure and propose changes. I will communicate through here.. and if you find anything please let me know here as well.

michaelandrepearce:
Ill have a look

michaelandrepearce:
@clebertsuconic ok so i know whats going on, another feature i have implemented for 2.7.0 reset of message groups, the test i had created is a little over strict and tested exact current behaviour, rather than what is just needed, that when resetting a message group, the selection on the next consumer for that group would have been the next consumer that joined. Where as actually when you reset the group in reality you just open the oppurtunity for another consumer to take the group.

michaelandrepearce:
Sorted locally just extracting repeatble iterator logic as franz is keen on this

michaelandrepearce:
@clebertsuconic sorted,

just added ability in consumers to repeat (new repeatable iterator), and then if group, we tell it to repeat, that provides the original behavior, and test is passing locally.

clebertsuconic:
Current list of failures:


Test Name | Duration | Age
-- | -- | --
org.apache.activemq.artemis.tests.integration.client.AutogroupIdTest.testGroupIdAutomaticallySetMultipleProducers | 37 ms | 1
org.apache.activemq.artemis.tests.integration.client.MessageGroupingTest.testMultipleGroupingXACommit | 47 ms | 1
org.apache.activemq.artemis.tests.integration.client.MessageGroupingTest.testLoadBalanceGroups | 21 ms | 1
org.apache.activemq.artemis.tests.integration.client.MessageGroupingTest.testMultipleGroupingTXCommit | 0.25 sec | 1
org.apache.activemq.artemis.tests.integration.client.MessageGroupingTest.testMultipleGroupingConsumeHalf | 0.79 sec | 1
org.apache.activemq.artemis.tests.integration.client.MessageGroupingTest.testMultipleGroupingXARollback | 55 ms | 1
org.apache.activemq.artemis.tests.integration.client.MessageGroupingTest.testMultipleGrouping | 14 ms | 1
org.apache.activemq.artemis.tests.integration.client.MessageGroupingTest.testMultipleGroupingTXRollback | 0.28 sec | 1
org.apache.activemq.artemis.tests.integration.cluster.failover.ReplicatedMultipleServerFailoverExtraBackupsTest.testStartLiveFirst | 1 min 8 sec | 1
org.apache.activemq.artemis.tests.integration.jms.client.GroupingTest.testManyGroups | 6 sec | 1
org.apache.activemq.artemis.tests.integration.jms.client.GroupingTest.testManyGroups | 6 sec | 1
org.apache.activemq.artemis.tests.integration.jmx.JmxConnectionTest.testJmxConnection | 1.1 sec | 1
org.apache.activemq.artemis.tests.integration.journal.AIOJournalImplTest.testDoubleDelete | 53 ms | 1
org.apache.activemq.artemis.tests.integration.journal.AIOUnbuferedJournalImplTest.testDoubleDelete | 39 ms | 1
org.apache.activemq.artemis.tests.integration.journal.MappedJournalImplTest.testDoubleDelete | 39 ms | 1
org.apache.activemq.artemis.tests.integration.journal.MappedUnbuferedJournalImplTest.testDoubleDelete | 43 ms | 1
org.apache.activemq.artemis.tests.integration.journal.NIOJournalImplTest.testDoubleDelete | 36 ms | 1
org.apache.activemq.artemis.tests.integration.journal.NIONoBufferJournalImplTest.testDoubleDelete | 18 ms | 1
org.apache.activemq.artemis.tests.integration.mqtt.MqttAcknowledgementTest.testAcknowledgementQOS1 | 5.6 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTFQQNTest.testSendAndReceiveMQTTSpecial1 | 11 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTOpenwireTest.testWildcards | 3.2 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTSecurityCRLTest.crlNotRevokedTest | 5.2 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testSendAndReceiveAtLeastOnce | 5.3 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testManagementQueueMessagesAreAckd | 5.2 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testMQTTRetainQoS | 7.3 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testSubscribeMultipleTopics | 5 min 0 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testSendAndReceiveExactlyOnceWithInterceptors | 6 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testSendAtLeastOnceReceiveAtMostOnce | 5.4 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testNoMessageReceivedAfterUnsubscribeMQTT | 5.4 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testMQTTPathPatterns | 5.3 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testUniqueMessageIds | 5.3 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testSendAndReceiveRetainedMessages | 5.4 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testRetainedMessage | 5.9 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testResendMessageId | 5.3 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testClientConnectionFailure | 11 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testSendAtMostOnceReceiveExactlyOnce | 5.5 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testAnycastAddressWorksWithMQTT | 3.5 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testAnycastPrefixWorksWithMQTT | 2.4 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testReceiveMessageSentWhileOffline | 5.4 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testAmbiguousRoutingWithMQTT | 3.2 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testLinkRouteAmqpReceiveMQTT | 5.6 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testPacketIdGeneratorNonCleanSession | 2.1 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testWillMessageIsRetained | 11 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testSendAndReceiveAtMostOnce | 5.7 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testSendAtLeastOnceReceiveExactlyOnce | 7.9 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testDuplicateSubscriptions | 7.6 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testSendJMSReceiveMQTT | 11 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testCleanSessionForSubscriptions | 11 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testRetainedMessagesAreCorrectlyFormedAfterRestart | 7.7 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testUnsubscribeMQTT | 23 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testSendAndReceiveLargeMessages | 6.4 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testClientConnectionFailureSendsWillMessage | 15 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MQTTTest.testSendAndReceiveMQTT | 12 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MqttClusterRemoteSubscribeTest.unsubscribeRemoteQueue | 10 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MqttClusterRemoteSubscribeTest.unsubscribeExistingQueue | 10 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MqttClusterRemoteSubscribeTest.unsubscribeRemoteQueueMultipleSubscriptions | 8.7 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MqttClusterRemoteSubscribeTest.unsubscribeRemoteQueueWildCard | 10 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MqttClusterWildcardTest.loadBalanceRequests | 36 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.MqttClusterWildcardTest.wildcardsWithBroker1Disconnected | 18 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.PahoMQTTTest.testLotsOfClients | 39 sec | 1
org.apache.activemq.artemis.tests.integration.mqtt.imported.PahoMQTTTest.testSendAndReceiveMQTT | 1 min 2 sec | 1
org.apache.activemq.artemis.tests.integration.openwire.cluster.TemporaryQueueClusterTest.testClusteredQueue | 7.7 sec | 1
org.apache.activemq.artemis.tests.integration.openwire.cluster.TemporaryQueueClusterTest.testTemporaryQueue | 3.9 sec | 1
org.apache.activemq.artemis.tests.integration.paging.GlobalPagingTest.testDeletePhysicalPages[storeType=FILE, mapped=true] | 2.9 sec | 1
org.apache.activemq.artemis.tests.integration.paging.GlobalPagingTest.testReceiveImmediate[storeType=FILE, mapped=true] | 2.1 sec | 1
org.apache.activemq.artemis.tests.integration.paging.PageCountSyncOnNonTXTest.testSendNoTx | 57 sec | 1
org.apache.activemq.artemis.tests.integration.paging.PagingTest.testPurge[storeType=FILE, mapped=false] | 34 sec | 1
org.apache.activemq.artemis.tests.integration.plugin.MqttPluginTest.testSendAndReceiveMQTT | 15 sec | 1
org.apache.activemq.artemis.tests.integration.server.ExpiryRunnerTest.testExpireConsumeHalf | 2.5 sec | 1
org.apache.activemq.artemis.tests.integration.stomp.StompConnectionCleanupTest.sendSTOMPReceiveMQTT[ws+v10.stomp] | 11 sec | 1
org.apache.activemq.artemis.tests.integration.stomp.StompConnectionCleanupTest.sendSTOMPReceiveMQTT[tcp+v10.stomp] | 11 sec | 1
org.apache.activemq.artemis.tests.integration.stomp.StompConnectionCleanupTest.testMultiProtocolConsumers[tcp+v10.stomp] | 3.6 sec | 1
org.apache.activemq.artemis.tests.integration.stomp.StompTest.sendSTOMPReceiveMQTT[ws+v10.stomp] | 10 sec | 1
org.apache.activemq.artemis.tests.integration.stomp.StompTest.sendSTOMPReceiveMQTT[tcp+v10.stomp] | 10 sec | 1
org.apache.activemq.artemis.tests.integration.stomp.StompWebSocketMaxFrameTest.testStompSendReceiveWithMaxFramePayloadLength[ws+v10.stomp] | 14 sec | 1
org.apache.activemq.artemis.tests.integration.stomp.StompWebSocketMaxFrameTest.testStompSendReceiveWithMaxFramePayloadLength[ws+v11.stomp] | 13 sec | 1
org.apache.activemq.artemis.tests.integration.stomp.StompWebSocketMaxFrameTest.testStompSendReceiveWithMaxFramePayloadLength[ws+v12.stomp] | 15 sec | 1

michaelandrepearce:
@clebertsuconic can you recheck this against master, i know things like testDoubleDelete are failing currently on master without this PR.

michaelandrepearce:
@clebertsuconic could you look at this https://github.com/apache/activemq-artemis/commit/b3f0a87e0db4832585a0460633889e13ef4bcbf8#r31982567, all builds are failing since that was merged to master.....

michaelandrepearce:
@clebertsuconic fully rebased (including over the FQQN for producers stuff)

michaelandrepearce:
@clebertsuconic so as noted, the message groups ones is sorted, repeat needs to occur only if groupconsumer exist. 

It does seem MQTTTest and StompWebSocketMaxFrameTest is directly broken due to this, but i cant think to why, any ideas welcome, ill continue to look into

michaelandrepearce:
@clebertsuconic ive found the pesky thing, in ServerSessionImpl where i had added a new parameter to the createConsumer method, the old method signature where we pass though but default the priority in the new, i had missed forwarding on the credits value, thus the issue. 

Typical one liner causing lots of headache ;)

michaelandrepearce:
@clebertsuconic  hows this looking now, we good to merge???

clebertsuconic:
Let me run one final round of testsuite.  If all good will merge it tomorrow.

clebertsuconic:
It’s important to keep the testsuite clean.  As we should release soon.

clebertsuconic:
@michaelandrepearce I'm pretty sure I will be able to merge this tomorrow. Testsuite had some unrelated failures. I just want to double check.. if there's any issues I'm sure i will be able to figure out.


will let you know tomorrow. I got the results pretty late after I finished. leave it with me for tomorrow first thing please?

clebertsuconic:
I added one change on top of yours during merge... to remove unused variables.. 


merged!


you are the man!

michaelandrepearce:
@clebertsuconic some of those variables are used but accesed by atomic updaters. Have commented on your commit

PR #2491, opened 2019-01-08
ARTEMIS-2217 remove state on clean MQTT session disconnect

jbertram:


jbertram:
IMO this should replace #2485.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSession.java
```diff
@@ -117,14 +118,11 @@ boolean getStopped() {
...
+      return clean;
    }
 
-   void setIsClean(boolean isClean) throws Exception {
-      this.isClean = isClean;
-      if (isClean) {
-         clean();
-      }
+   void setClean(boolean clean) throws Exception {
+      this.clean = clean;
```
onlyMIT: It is necessary to call the "clean()" method to clean up old session information when creating a connection.
If it is not cleaned up, when the cleanSession of the last MQTT consumer is false, and the cleanSession of the connected MQTT consumer is true, the message in the old queue will be consumed, which is actually not allowed.  
I think this is why calling "clean()" in the "setIsClean(boolean isClean)" method

onlyMIT: It is necessary to call the "clean()" method to clean up old session information when creating a connection.
If it is not cleaned up, when the cleanSession of the last MQTT consumer is false, and the cleanSession of the connected MQTT consumer is true, the message in the old queue will be consumed, which is actually not allowed.
I think this is why calling "clean()" in the "setIsClean(boolean isClean)" method

onlyMIT:
@jbertram  I think I found out why your solution passed the test， your test was constructed on a wrong test code. the solution your provided, after I tested it with the revised test code, the test did not pass.
The test issue， i created a jira and opened a  [#2493](https://github.com/apache/activemq-artemis/pull/2493) to solve this test issue. 
we need it， calling "clean()" in the "setIsClean(boolean isClean)" method.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSession.java
jbertram: The expectation of a "setter" method is simply to _set_ a variable and nothing more. The additional logic in the `setClean` method is not intuitive which is why I removed it and put it in the `getSessionState` method.

jbertram:
The test code is correct. The state kept in an unclean session includes the subscriptions which is why the test doesn't create a subscription before trying to receive the message. If the broker doesn't properly clean the session state when the client connects with cleanSession=true then it will receive a message and the test will fail.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSession.java
onlyMIT: @jbertram in the getSessionState method.Only clear state，not call 'clean()' method. In fact, the queue is not cleaned up.
I use the code for the ‘paho’ test. The first consumer "cleanSession=false", using a different clientID to open a producer to send a message. Close the producer and consumer, use the same clientID and cleanSession = true" to open the second consumer and find that the consumer will consume the legacy message in the queue。So I suspect that there is a problem with the test code.
I am always looking for why my test results will consume the legacy messages in the queue, and your test results will not。
After seeing your information, I re-reviewed the code and found that the test code did not have any problems. What is causing my doubts is that because of your change, when cleanSession=true, only the MQTTSessionState is cleaned up, the queue still exists, and the legacy messages in the queue are consumed when resubscribing.
 Can close [#2493 ](https://github.com/apache/activemq-artemis/pull/2493) . I think you need to review your change。

onlyMIT: Look good to me!

PR #2492, opened 2019-01-08
ARTEMIS-2222 why the position remains unchanged if ignored is set to true

CNNJYB:
I am a bit confused about this, When CursorIterator:next is called during queue depage, if ignored is set to true, why the position remains unchanged. 		
if (!ignored) {
    position = message.getPosition();
}		
For example, the client sends some messages to the topic subscriber ta (this topic has two subscribers ta and tb), every time tb depage continuous PagePositions that ignored are set to true will be traversed again.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
@@ -1315,9 +1315,7 @@ public final class PageSubscriptionImpl implements PageSubscription {
                   }
                }
 
-               if (!ignored) {
-                  position = message.getPosition();
-               }
+               position = message.getPosition();
 
                if (valid) {
                   match = match(message.getMessage());


franz1981:
It is correct that the first commit is related to ARTEMIS-2144?

CNNJYB:
> It is correct that the first commit is related to ARTEMIS-2144?

Branch is not latest. Now updated.

michaelandrepearce:
This had a specific purpose so we ignore deleted or completed pages.

As per comment in code.

            // any deleted or complete page will be ignored on the moveNextPage, we will just keep going

CNNJYB:
> This had a specific purpose so we ignore deleted or completed pages, so we shouldnt be getting the message position.
> 
> As per comment in code.
> 
> ```
>         // any deleted or complete page will be ignored on the moveNextPage, we will just keep going
> ```
> Is there visible bug that needs addressing? We shouldnt just be changing code, without a specific testable bug we're trying to resolve.



> This had a specific purpose so we ignore deleted or completed pages, so we shouldnt be getting the message position.
> 
> As per comment in code.
> 
> ```
>         // any deleted or complete page will be ignored on the moveNextPage, we will just keep going
> ```
> Is there visible bug that needs addressing? We shouldnt just be changing code, without a specific testable bug we're trying to resolve.

If some pages have not been deleted (tb consumes these messages, ta has not consumed yet), CursorIterator:moveNext is called during queue depage. These messages will be traversed again. Especially during CursorProviderget:getPageCache, it takes time to readPage, if pages are not in the softCache collection.

michaelandrepearce:
Point being if this is a bug, please can there be a test that exposes the issues, so a) this can be reviewed b) we can validate the fix c) ensure there is no regression.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -1315,9 +1315,7 @@ private PagedReference moveNext() {
                   }
                }
 
-               if (!ignored) {
-                  position = message.getPosition();
-               }
+               position = message.getPosition();
```
clebertsuconic: I think this made sense at some point, but after a few fixes later it's harmless.

I don't think this needs a JIRA as there's no issue I think.

I will merge and keep the JIRA, but I will rename the commit from being a question into something more affirmative.

PR #2493, opened 2019-01-08
ARTEMIS-2223 when a new consumer is created, no subscription is called.

onlyMIT:
In the 'MQTTTest.testCleanSession()' test method, when a new consumer is created, no subscription is called.Consumers need to subscribe before they consume the news.


jbertram:
The test is correct. Creating a subscription before attempting to receive the message defeats the point of the test as the test is ensuring that the session is clean (i.e. no previous subscriptions exist) when reconnecting with the same client ID.

onlyMIT:
The test is correct.can close the jira and pull request

PR #2494, opened 2019-01-09
ARTEMIS-2224 Reduce contention on LivePageCacheImpl

franz1981:
It includes:

- **lock-free LivePageCache + tests**:
LivePageCacheImpl has been reimplemented to be
lock-free, multi-producer and multi-consumer
in any of its operations.
- **Avoid unnecessary page cache queries on ack TX**:
PageSubscriptionImpl::ackTx is already performing a counter update
using the message persistent size: the size can be reused on
PagePosition::setPersistentSize, avoiding to query the page cache just
to compute it.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/ConcurrentAppendOnlyChunkedList.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/ConcurrentAppendOnlyChunkedList.java
@@ -1,0 +1,286 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.utils.collections;
+
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+import java.util.function.IntFunction;
+
+/**
+ * This collection is a concurrent append-only list that grows in chunks.<br>
+ * It's safe to be used by many threads concurrently and has a max capacity of {@link Integer#MAX_VALUE}.
+ */
+public final class ConcurrentAppendOnlyChunkedList<T> {
+
+   private static final class AtomicChunk<T> extends AtomicReferenceArray<T> {
+
+      AtomicChunk<T> next = null;
+      final AtomicChunk<T> prev;
+      final int index;
+
+      AtomicChunk(int index, AtomicChunk<T> prev, int length) {
+         super(length);
+         this.index = index;
+         this.prev = prev;
+      }
+   }
+
+   private static final AtomicLongFieldUpdater<ConcurrentAppendOnlyChunkedList> LAST_INDEX_UPDATER = AtomicLongFieldUpdater.newUpdater(ConcurrentAppendOnlyChunkedList.class, "lastIndex");
+
+   private static final AtomicLongFieldUpdater<ConcurrentAppendOnlyChunkedList> CACHED_LAST_INDEX_UPDATER = AtomicLongFieldUpdater.newUpdater(ConcurrentAppendOnlyChunkedList.class, "cachedLastIndex");
+
+   private final int chunkSize;
+
+   private final int chunkMask;
+
+   private final int chunkSizeLog2;
+
+   private static final long RESIZING = -1;
+
+   private AtomicChunk<T> firstBuffer = null;
+
+   private AtomicChunk<T> lastBuffer = null;
+
+   //it is both the current index of the next element to be claimed and the current size of the collection
+   private volatile long lastIndex = 0;
+
+   //cached view of lastIndex used to avoid invalidating lastIndex while being updated by the appends
+
+   private volatile long cachedLastIndex = 0;
+
+   /**
+    * @throws IllegalArgumentException if {@code chunkSize} is <0 or not a power of 2
+    */
+   public ConcurrentAppendOnlyChunkedList(final int chunkSize) {
+      if (chunkSize <= 0) {
+         throw new IllegalArgumentException("chunkSize must be >0");
+      }
+      //IMPORTANT: to enable some nice optimizations on / and %, chunk size MUST BE a power of 2
+      if (Integer.bitCount(chunkSize) != 1) {
+         throw new IllegalArgumentException("chunkSize must be a power of 2");
+      }
+      this.chunkSize = chunkSize;
+      this.chunkMask = chunkSize - 1;
+      this.chunkSizeLog2 = Integer.numberOfTrailingZeros(chunkSize);
+   }
+
+   private long getValidLastIndex() {
+      while (true) {
+         final long lastIndex = this.lastIndex;
+         if (lastIndex == RESIZING) {
+            Thread.yield();
+            continue;
+         }
+         return lastIndex;
+      }
+   }
+
+   /**
+    * It returns the number of elements currently added.
+    */
+   public int size() {
+      return (int) getValidLastIndex();
+   }
+
+   /**
+    * It appends {@code elements} to the collection.
+    */
+   public void addAll(T[] elements) {
+      for (T e : elements) {
+         add(e);
+      }
+   }
+
+   /**
+    * Returns the element at the specified position in this collection or {@code null} if not found.
+    */
+   public T get(int index) {
+      if (index < 0) {
+         return null;
+      }
+      //it allow to perform less cache invalidations vs lastIndex if there are bursts of appends
+      long lastIndex = cachedLastIndex;
+      if (index >= lastIndex) {
+         lastIndex = getValidLastIndex();
+         //it is a element over the current size?
+         if (index >= lastIndex) {
+            return null;
+         }
+         //publish it for others readers
+         CACHED_LAST_INDEX_UPDATER.lazySet(this, lastIndex);
+      }
+      final AtomicChunk<T> buffer;
+      final int offset;
+      if (index >= chunkSize) {
+         offset = index & chunkMask;
+         //slow path is moved in a separate method
+         buffer = getChunkOf(index, lastIndex);
+      } else {
+         offset = index;
+         buffer = firstBuffer;
+      }
+      return pollElement(buffer, offset);
+   }
+
+   /**
+    * Implements a lock-free version of the optimization used on {@link java.util.LinkedList#get(int)} to speed up queries
+    * ie backward search of a node if needed.
+    */
+   private AtomicChunk<T> getChunkOf(final int index, final long lastIndex) {
+      final int chunkSizeLog2 = this.chunkSizeLog2;
+      //fast division by a power of 2
+      final int chunkIndex = index >> chunkSizeLog2;
+      //size is never allowed to be > Integer.MAX_VALUE
+      final int lastChunkIndex = (int) lastIndex >> chunkSizeLog2;
+      int chunkIndexes = chunkIndex;
+      AtomicChunk<T> buffer = null;
+      boolean forward = true;
+      int distanceFromLastChunkIndex = lastChunkIndex - chunkIndex;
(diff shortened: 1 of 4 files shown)


franz1981:
The effects of the contention on `LivePageCacheImpl` are more visible after applying https://github.com/apache/activemq-artemis/pull/2484

franz1981:
@michaelandrepearce @wy96f @qihongxu Please review and test :+1:

franz1981:
Most of the benefits on memory footprint are already explained on https://github.com/qihongxu/activemq-artemis/pull/1, while about the contention I have already built several contention graphs that shows that this implementation scale linearly with the number of producers on a topic, without any contention.
In addition, several JMH microbenchs (not attached here) shows that is always fastest (if single-threaded too) then the original one for any operations and in any condition.
The only missing bits is an end 2 end test after applying #2484

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/LivePageCacheImpl.java
```diff
@@ -28,15 +28,49 @@
 /**
  * This is the same as PageCache, however this is for the page that's being currently written.
  */
```
clebertsuconic: Isn't this mixing the collection implementation itself into the LivePageCache?

isn't there a way to implement this logic into its own structure? Like PageCache using a generic ChunkArray (a name I just came up here)?

I'm a bit concerned on maintaining the business side of this issue (that is the PageCache) with the speedy implementation of a collection.

michaelandrepearce: @franz1981 Id agree with @clebertsuconic here, a bit like what ive done with priority consumers, i ended up splitting out the collections logic, which has ended up making it cleaner, and easier to reason with. (and as youve marked out on its pr, more testable ;) )

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/LivePageCacheImpl.java
```diff
@@ -48,54 +82,228 @@ public long getPageId() {
...
+      //store release any previous write and "unblock" anyone waiting resizing to finish
+      PRODUCER_INDEX_UPDATER.lazySet(this, pIndex + 1);
+      return true;
    }
 
    @Override
-   public synchronized void close() {
+   public void close() {
       logger.tracef("Closing %s", this);
       this.isLive = false;
```
michaelandrepearce: now isLive is volatile and sync is remove, this should be updated by an atomic field updater? no?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/LivePageCacheImpl.java
franz1981: Makes sense :+1:

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/LivePageCacheImpl.java
```diff
@@ -48,54 +82,228 @@ public long getPageId() {
...
 
+   private static PagedMessage[] EMPTY_MSG = null;
+
+   private static PagedMessage[] noMessages() {
+      //it is a benign race: no need strong initializations here
+      PagedMessage[] empty = EMPTY_MSG;
+      if (empty != null) {
+         return empty;
+      } else {
+         empty = new PagedMessage[0];
```
michaelandrepearce: Why not simply make noMessages return a static empty array, e.g. make PagedMessage[0] - similar in nature to https://android.googlesource.com/platform/libcore/+/jb-mr2-release/luni/src/main/java/libcore/util/EmptyArray.java, and then saves instantiating every time.  

Also avoids a race condition, this current code, seems to have with lazy initialization, e.g. currently if two threads concurrently call noMessages its possible i get two PagedMessage[0] objects created.

franz1981: yep I have done it on purpose: is a benign race, because there is just a very low chance a new empty would be allocated several times, but I agree :+1: 
It will makes the code much easier

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/LivePageCacheImpl.java
franz1981: No needs,`isLive = false` is a volatile set exactly like using the updater

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/LivePageCacheImpl.java
```diff
@@ -48,54 +82,228 @@ public long getPageId() {
...
          return null;
       }
+      //it allow to perform less cache invalidations vs producerIndex if there are bursts of appends
+      long size = lastSeenProducerIndex;
+      if (messageNumber >= size) {
+         while ((size = producerIndex) == RESIZING) {
+            Thread.yield();
+         }
+         //it is a message over the current size?
+         if (messageNumber >= size) {
```
michaelandrepearce: surely if this occurs, theres some issue......

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/LivePageCacheImpl.java
```diff
@@ -48,54 +82,228 @@ public long getPageId() {
...
+      }
+      return empty;
+   }
+
    @Override
-   public synchronized PagedMessage[] getMessages() {
-      return messages.toArray(new PagedMessage[messages.size()]);
+   public PagedMessage[] getMessages() {
+      long currentSize;
+      while ((currentSize = producerIndex) == RESIZING) {
```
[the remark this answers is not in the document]
franz1981: good catch! :+1: While creating a more generic collection I will refactor the bits to avoid duplication when possible

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/LivePageCacheImpl.java
```diff
@@ -48,54 +82,228 @@ public long getPageId() {
...
+      return msg;
+   }
+
+   /**
+    * Implements a lock-free version of the optimization used on {@link java.util.LinkedList#get(int)} to speed up queries
+    * ie backward search of a node if needed.
+    */
+   private AtomicChunk<PagedMessage> jump(final int messageNumber, final long size) {
+      //fast division by a power of 2
+      final int jumps = messageNumber >> chunkSizeLog2;
```
michaelandrepearce: cough ;) ... for the same comment you left me .... :P

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/LivePageCacheImpl.java
michaelandrepearce: if this was a collection, this would almost be throwing illegal arguement.....

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/LivePageCacheImpl.java
clebertsuconic: This is a lot better!

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/ConcurrentAppendOnlyChunkedList.java
```diff
@@ -0,0 +1,291 @@
...
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+import java.util.function.IntFunction;
+
+/**
+ * This collection is a concurrent append-only list that grows in chunks.<br>
+ * It's safe to be used by many threads concurrently and has a max capacity of {@link Integer#MAX_VALUE}.
+ */
+public final class ConcurrentAppendOnlyChunkedList<T> {
```
clebertsuconic: This is a lot better!

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/LivePageCacheImpl.java
franz1981: Is not that easy here, this one is not a common collection (in theory, but we can get an agreement on the contract for sure): it is more an append only list that allows indexed queries like a map.
If I remember correctly it is similar to an [hashed array tree](https://en.wikipedia.org/wiki/Hashed_array_tree), where the top-level directory is a double linked list of "folders" (instead of an array, like the original implementation): indeed in the code there are chunkIndex (==key of top level directory) and offset (==key of leaf into a directory).

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/ConcurrentAppendOnlyChunkedList.java
franz1981: Thanks to you guys and one day I will be better in naming things too I swear :D

franz1981:
@qihongxu Both the changes have been merged ie if you want to perform some tests to check if is working as expected everything is on master!

franz1981:
> @franz1981 thx for recommending this tool, much easier to use than jfr!

I was sure you would have like it: is indeed "pragmatically better" for devs IMHO: just remember to enable debug non safepoints, but with JFR is the same, reading their docs!

@qihongxu 
> At the end of test the perf boosted to 30k tps

The current implementation (lock-free) of the live page cache suffer of the same performance limitation of a `LinkedList`: queries at the beginning and at the end of the list are faster ie `O(1)`, while in the middle are slower ie `O(n)`.
To make things faster we need to exploit a better access pattern and provide a cursor-like implementation on it to be used on `QueueImpl::depage`: it should improve things at the point of having ~`O(1)` `getMessage` cost if the cursor will be used with a sequential access pattern, like an iterator.
If you like the idea and it really seems to be important I can provide an implementaion for it ie 900M of page size and very small messages are not such a common use case TBH

qihongxu:
> 900M of page size and very small messages are not such a common use case TBH

Agreed. Use this case just to make it stay at livePageCache longer. In production we use default 10M page size, which means that under high concurrency it's not easy for livePageCache to be read&write at the same time.

In conclusion this patch works fine as described and performance benefits from it:)

qihongxu:
@franz1981 
> 1. the tps are dependent by the most demanding operation in the CPU flamegraph you've posted
>    ie `ConcurrentAppendOnlyList:.get(index)`

The flamegraph didn't show clearly without zoom in/out. On top of `ConcurrentAppendOnlyList:.get(index)` is 'getChunkOf(index, lastIndex)' since the page is huge and we only consumed messages from it.

While we let consumers and producers work together, and producer is much faster than producers the top of flamegrapgh changed to `getValidLastIndex()` due to the burst of messages i guess.

> If the same test (same configuration) is performed using just 1 consumer I'm expecting (best case scenario) that it would have (3k-6k)/200 tps

Probably higher than that. In other tests when using 200 threads it's broker's processing ability that becomes the bottleneck of tps. The reason of using '200' is to squeeze all power out of broker.  That is to say 1 consumer might perform better than total/200.

> I suppose that 2 can't be true, so it really depends on the number of cores/threads available to execute such queries to LivePageCache IF such queries are performed in parallel.

I agree. As for two assumptions, I have same ideas with you (1 is right and 2 might not). So if we use 400 threads the tps might not be higher than using 200. 

I could make a test on single and 400 consumers tomorrow using same config. What do you think or if you have better ideas/test cases towards our hypothesis? If so please let me know:)

franz1981:
@qihongxu thanks! I need to put a couple do logs to check how is th behaviour with several consumers and check if live page queries are happening with single-threaded semantic ie are just limited by the efficiency of the query and can't scale.

I'm surprised about getValidLastIndex(): it should be the bottleneck (if the paged message query is the caller) only when the cache is empty (no messages) or the index is == cache size. If there is a cache hit I expect that getChunkOf or the volatile get on the message slot to be costy.
About the test is a good idea: having 1/200/400 consumers can help to understand the scaling, but I suppose that your intuitions are right here.

qihongxu:
@franz1981 
Here's the result of single and 400 consumers.
With these settings applied: 
max-size-bytes = 950MB, 
page-size = 900MB, 
Produced 2,000,000 msgs in a new queue (make it a ~700MB page file as ‘livePage’ + 950MB in journal), then closed producer, started consumer.

**single consumer: avg 1k tps**
1 consumer 1k tps

**400 consumers: avg 5.3k tps**
400 consumer 5 4k tps

Their flamegraphes are similar like this.
image

As we can see `getChunkOf(index, lastIndex)` is most costy in both cases, and 400 consumers didn't perform better than 200. Also, we found that once most journal msgs have been consumed and all paged msgs loaded to memory, tps can rocket to 30k just like we have shown in other tests before.

PR #2496, opened 2019-01-11
ENTMQBR-2178 add Implementation-Version for bundle packaging

soul2zimate:
https://issues.apache.org/jira/browse/ARTEMIS-2227
Add Implementation-Version for bundle packaging MANIFEST.MF in artemis-protocols module

### DIFF
--- a/artemis-protocols/pom.xml
+++ b/artemis-protocols/pom.xml
@@ -41,4 +41,20 @@
       <module>artemis-mqtt-protocol</module>
    </modules>
 
+   <build>
+      <plugins>
+         <plugin>
+            <groupId>org.apache.felix</groupId>
+            <artifactId>maven-bundle-plugin</artifactId>
+            <version>3.0.0</version>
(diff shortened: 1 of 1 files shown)


jbertram:
Commit messages should reference ARTEMIS JIRAs, not JIRAs from external projects.

jbertram:
Also, for what it's worth, all the artifacts created from the `artemis-protocols` module don't have the `Implementation-Version` specified in their MANIFEST.MF not just the ones for `artemis-hornetq-protocol` and `artemis-hqclient-protocol`. This appears to be because the `packaging` for all of them is defined as `bundle`. Perhaps a better solution would be to modify the `pom.xml` of `artemis-protocols` to ensure all the protocol jars have the proper `Implementation-Version`?

soul2zimate:
Thanks, I have referred to ARTEMIS-2227 in commit message and made the change in artemis-protocols module parent pom.xml

PR #2497, opened 2019-01-14
ARTEMIS-2228 Check message size sent over management API

gaohoward:
Currently user can send an arbitrary size of messages via
management api (console for example) and this may break the
buffer size limit and cause the broker throw unexpected
exceptions. We need to put some check on the message size
over the management API and reject the message if it's too big.


clebertsuconic:
I agree with @michaelandrepearce here.. why not just convert it to large message.

I think the postOffice.route is already doing that? on that case we just need a test.

clebertsuconic:
@gaohoward I pulled your branch and fixed it here.

clebertsuconic:
@gaohoward  / @michaelandrepearce look into #2498  please?

it's a lot simpler there.

gaohoward:
@michaelandrepearce @clebertsuconic OK thanks guys, I agree with you. I think #2498 is fine.

clebertsuconic:
Can you close this one please?

PR #2498, opened 2019-01-14
ARTEMIS-2228 Large Messages over Management

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
```diff
@@ -1005,9 +1006,9 @@ public RemotingConnection getRemotingConnection() {
          ByteBuffer buffer = ByteBuffer.allocate(8);
          buffer.putLong(queue.getID());
          message.putBytesProperty(Message.HDR_ROUTE_TO_IDS, buffer.array());
-         postOffice.route(message, true);
+         server.getPostOffice().route(LargeServerMessageImpl.checkLargeMessage(message, server.getStorageManager()), true);
```
michaelandrepearce: Maybe rather than directly routing to the post office, should we not send via serversession? This would then sort the large message issue but also ensure the message is treated like others, including going via broker plugins etc

michaelandrepearce: Just FYI, the fix looks good, im just wondering if theres a better way, but if not then this looks good.

michaelandrepearce:
Just fyi, Re your failed pr build since some recent merges to master ive been getting more intermittent build failures more regularly. I dont think its your pr

PR #2504, opened 2019-01-18
NO-JIRA Check data != null during encode

michaelandrepearce:
Picked up by code analysis checks


michaelandrepearce:
data can be null, shown by check at top of method 

      int estimated = Math.max(1500, data != null ? data.capacity() + 1000 : 0);

but later reference to data is used without null check

michaelandrepearce:
@clebertsuconic just running some code analysis quickly if and fixing the issues highlighted that look genuine quickly.

PR #2509, opened 2019-01-18
NO-JIRA Fix Possible NPE in possibleLive.getA()

michaelandrepearce:
Variable possibleLive may be null, as signified by trace log above, also dedupe code.


gaohoward:
+1.

PR #2511, opened 2019-01-21
[ARTEMIS-2234]: Incorrect warning from LoggingConfigurationFileReloader

ehsavoie:
* returning the LogContextConfiguration if we have one.

Jira: https://issues.apache.org/jira/browse/ARTEMIS-2234


jbertram:
I went ahead and cleaned this up and merged it.

PR #2514, opened 2019-01-23
ARTEMIS-2236 Address Latency Impact caused by ARTEMIS-1451

michaelandrepearce:
Revert ARTEMIS-1451
Remove synchronization by simplifying original code and using atomics (addressing concerns in ARTEMIS-1451)

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ActiveMQThreadPoolExecutor.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ActiveMQThreadPoolExecutor.java
@@ -20,6 +20,7 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 /*
  * ActiveMQThreadPoolExecutor: a special ThreadPoolExecutor that combines
@@ -46,15 +47,13 @@ public class ActiveMQThreadPoolExecutor extends ThreadPoolExecutor {
 
       private ActiveMQThreadPoolExecutor executor = null;
 
-      // lock object to synchronize on
-      private final Object lock = new Object();
-
       // keep track of the difference between the number of idle threads and
       // the number of queued tasks. If the delta is > 0, we have more
       // idle threads than queued tasks and can add more tasks into the queue.
       // The delta is incremented if a thread becomes idle or if a task is taken from the queue.
       // The delta is decremented if a thread leaves idle state or if a task is added to the queue.
-      private int threadTaskDelta = 0;
+      private static final AtomicIntegerFieldUpdater<ThreadPoolQueue> DELTA_UPDATER = AtomicIntegerFieldUpdater.newUpdater(ThreadPoolQueue.class, "threadTaskDelta");
+      private volatile int threadTaskDelta = 0;
 
       public void setExecutor(ActiveMQThreadPoolExecutor executor) {
          this.executor = executor;
@@ -64,20 +63,15 @@ public class ActiveMQThreadPoolExecutor extends ThreadPoolExecutor {
       public boolean offer(Runnable runnable) {
          boolean retval = false;
 
-         // Need to lock for 2 reasons:
-         // 1. to safely handle poll timeouts
-         // 2. to protect the delta from parallel updates
-         synchronized (lock) {
-            if ((executor.getPoolSize() >= executor.getMaximumPoolSize()) || (threadTaskDelta > 0)) {
-               // A new task will be added to the queue if the maximum number of threads has been reached
-               // or if the delta is > 0, which means that there are enough idle threads.
+         if (threadTaskDelta > 0 || (executor.getPoolSize() >= executor.getMaximumPoolSize())) {
+            // A new task will be added to the queue if the maximum number of threads has been reached
+            // or if the delta is > 0, which means that there are enough idle threads.
 
-               retval = super.offer(runnable);
+            retval = super.offer(runnable);
 
-               // Only decrement the delta if the task has actually been added to the queue
-               if (retval)
-                  threadTaskDelta--;
-            }
+            // Only decrement the delta if the task has actually been added to the queue
+            if (retval)
+               DELTA_UPDATER.decrementAndGet(this);
          }
 
          return retval;
@@ -87,9 +81,8 @@ public class ActiveMQThreadPoolExecutor extends ThreadPoolExecutor {
       public Runnable take() throws InterruptedException {
          // Increment the delta as a thread becomes idle
          // by waiting for a task to take from the queue
-         synchronized (lock) {
-            threadTaskDelta++;
-         }
+         DELTA_UPDATER.incrementAndGet(this);
+
 
          Runnable runnable = null;
 
@@ -102,9 +95,7 @@ public class ActiveMQThreadPoolExecutor extends ThreadPoolExecutor {
             // (decremented by the thread and incremented by the taken task)
             // Only if no task had been taken, we have to decrement the delta.
             if (runnable == null) {
-               synchronized (lock) {
-                  threadTaskDelta--;
-               }
+               DELTA_UPDATER.decrementAndGet(this);
             }
          }
       }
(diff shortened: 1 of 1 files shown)


michaelandrepearce:
1451 removed synchronization in the class, by adding the overhead of creating a new thread until threadpool is full, this though means that instead of re-using an idle thread for new work, new threads are being created which is expensive, and the point of thread pooling.

michaelandrepearce:
To show this, using a broker with persistance disabled (to remove external disk perf factors)


Using CORE Producer -> Consume Latency Average per 10000 messages (20 runs)

CURRENT MASTER
RESULT: 292
RESULT: 187
RESULT: 151
RESULT: 124
RESULT: 109
RESULT: 148
RESULT: 121
RESULT: 127
RESULT: 104
RESULT: 201
RESULT: 118
RESULT: 116
RESULT: 115
RESULT: 153
RESULT: 126
RESULT: 127
RESULT: 107
RESULT: 128
RESULT: 121
RESULT: 126


WITH REVERTING AND APPLYING THIS FIX

RESULT: 249
RESULT: 173
RESULT: 150
RESULT: 120
RESULT: 127
RESULT: 124
RESULT: 140
RESULT: 122
RESULT: 106
RESULT: 100
RESULT: 96
RESULT: 104
RESULT: 103
RESULT: 96
RESULT: 92
RESULT: 94
RESULT: 100
RESULT: 91
RESULT: 101
RESULT: 100

michaelandrepearce:
As this is performance regression, request this makes 2.7.0 and also possible hot patched into 2.6.x stream.

franz1981:
Suggestion: i would use something more precise (with an higher spectrum of latencies) if possible to make things comparable: https://github.com/franz1981/artemis-load-generator
Change the core version you need and run it with:
```java -jar destination-bench.jar --bytes 100 --protocol artemis --url tcp://localhost:61616 ---out /tmp/latencies.txt --name q --iterations 100000 --runs 1 --warmup 100000 --wait 2```
It will give you the latencies over an all out throughout: then you need to choose a sustainable throughput eg 100000 and re-run the same test adding `--target 100000`, then you will have comparable latencies.

**Comment by franz1981** [comment]
@michaelandrepearce I agree and that's why now we don't have to do the same thing :)
Before saying that's a regression, please try to size the thread pool in the way I have suggested, at least.
Understanding the implication of an evident performance regressions help just to improve things and avoid in the future to accept such patches without proofs.

**Comment by michaelandrepearce** [comment]
@franz1981 we have been running investigations in a REAL environment going back an forth between versions on this for some time, even if we revert this class, then latency is improved.

**Comment by franz1981** [comment]
@michaelandrepearce I do understand but I believe that it could be dependent by a wrong configuration, given that the behaviour of the thread-pool is quite different from before: did you have tried to change the configuration of the thread-pool according to this new behaviour?

**Comment by michaelandrepearce** [comment]
@franz1981 its not, the new behaviour is simply it creates a new thread until the max pool size, this means even if an idle thread it creates a new thread, as you know thread creation is expensive, and reason for a thread pool.

**Comment by franz1981** [comment]
@michaelandrepearce 
Please quote on my comments where "it's not" is referring to, to help me understand

**Comment by michaelandrepearce** [comment]
@franz1981 its not down to sizing the thread pool.

**Comment by michaelandrepearce** [comment]
@franz1981 This is what the old varient tried to avoid, was creating new threads...

**Comment by franz1981** [comment]
@michaelandrepearce 
Ok, I need to understand this.
The new version distributes tasks across threads, but if the maxThread count is not reached it will create new ones. 
If any of those threads will remain idle for > 1 minute the thread is being interrupted and die.
If all threads are kept busy they will remain alive and will receive an equally distributed load of tasks.
Until now is correct?

**Comment by michaelandrepearce** [comment]
@franz1981 roughly correct

The original one before 1451, tried to re-use an idle thread if possible before creating a new one, albiet using some sync blocks.

**Comment by franz1981** [comment]
@michaelandrepearce 
Ok and the old thread-pool, assuming max thread count already reached, was equally distributing the load across threads?

**Comment by michaelandrepearce** [comment]
correct.

**Comment by franz1981** [comment]
Ok, if that's the problem I agree with you that need to be addressed: is not the behaviour of a thread-pool, that we expect to have, but it is mostly hitting when the system is got idle and suddenly start to process messages

**Comment by michaelandrepearce** [comment]
So the result i show provide the variance of 200,000 message as its split per 10,000

Just total run time is 15% faster:

MASTER = 2801
THIS = 2388

**Comment by franz1981** [comment]
Now that you've helped me to understand the issue I believe that's correct: let me take a look to the current impl, because there is another thing I have on my "bag" re this one :)
Well done to have found it, as you have said, that PR has stealthy entered into the code-base without having clue about the implications :O

**Comment by michaelandrepearce** [comment]
So all ive done, is revert that change of 1451 to put back semantics. 

Streamlined the code a tad, so then we can remove sync blocks, and can just make the updating of threadTaskDelta using an atomic updater rather than doing it inside a sync block. Which then re-addresses concern of 1451 which was sync blocks.

**Comment by franz1981** [comment]
Yep I see that there are a couple of things that could be addressed due to this + a possible improvement: I will propose the improvement separately, but I believe you will like it :)

File: artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ActiveMQThreadPoolExecutor.java [hunk 1]
```diff
@@ -29,33 +29,129 @@
...
+   // private constructor is needed to inject 'this' into the ThreadPoolQueue instance
+   private ActiveMQThreadPoolExecutor(int coreSize,
+                                      int maxSize,
+                                      long keep,
+                                      TimeUnit keepUnits,
+                                      ThreadPoolQueue myQueue,
+                                      ThreadFactory factory) {
+      super(coreSize, Integer.MAX_VALUE, keep, keepUnits, myQueue, factory);
+      maxPoolSize = maxSize;
+      myQueue.setExecutor(this);
```
franz1981: `myQueue.setExecutor` is bot protected by a so called safe-initialization: there are 2 ways to address it.
The correct one, but more complex is:
- declare `private volatile ActiveMQThreadPoolExecutor executor = null;` and perform a volatile set on `setExecutor`

Same should be for maxPoolSize, given that is an int, it risk to not be updated atomically, if changed

franz1981:
> I'm reverting but also removed the need for the sync blocks, and using atomic updater to the threadTaskDelta

Let me take a look to the original code and I can tell you :+1:

michaelandrepearce:
I can make it a two stage commit if it helps?

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ActiveMQThreadPoolExecutor.java
michaelandrepearce: This was original

franz1981:
@michaelandrepearce 
no no, bud don't worry: this was my ignorance :)
I've just look at the origianal version and I totally agree with you: the issues I've outlined exists (safe-construction etc etc), but was present in the original version too :O
Please go ahread and I will post you a branch I've done in the past on this :)

franz1981:
Yess and sorry for the "interrogation": given that's a core part you know that I'm just concerned to understand it :)

gemmellr:
Franz is the man for this type of thing so I'll leave him to the review, but I do think having a separate specific revert commit would have actually been a good idea given it looks to change so much, making it easier for someone looking at 'actual changes' from the original to do so easily. Essentially, in general I think if its worth actually reverting something, do that specifically.

michaelandrepearce:
@gemmellr @franz1981 i was in two minds as normally get asked to squash, but @gemmellr sold me. Ill split the commits

franz1981:
@michaelandrepearce 
> but @gemmellr sold me. Ill split the commits

I can say that's the right thing to do, @gemmellr has proposed something that worth to do :)

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ActiveMQThreadPoolExecutor.java
```diff
@@ -64,20 +63,15 @@ public void setExecutor(ActiveMQThreadPoolExecutor executor) {
...
          boolean retval = false;
 
-         // Need to lock for 2 reasons:
-         // 1. to safely handle poll timeouts
-         // 2. to protect the delta from parallel updates
-         synchronized (lock) {
-            if ((executor.getPoolSize() >= executor.getMaximumPoolSize()) || (threadTaskDelta > 0)) {
-               // A new task will be added to the queue if the maximum number of threads has been reached
-               // or if the delta is > 0, which means that there are enough idle threads.
+         if (threadTaskDelta > 0 || (executor.getPoolSize() >= executor.getMaximumPoolSize())) {
```
franz1981: Not a significan change, but why you've reverted the checks here?
The common path should be that you have all the threads busy up and running (so `executor.getPoolSize() >= executor.getMaximumPoolSize() = true`): it would avoid checking the delta...

michaelandrepearce: getPoolSize uses a lock underneath in the executor, so we possible can avoid that when we have idle threads.

if threadTaskDelta > 0 (which means we have idle) then we can offer, we only need to see if we are at max if we dont have idle.

franz1981: Ahhh is see that there is a lock there, that's why!

michaelandrepearce:
@franz1981 @gemmellr split the commits (there is now 2)

michaelandrepearce:
> Franz is the man

He is indeed!!

michaelandrepearce:
@franz1981  im getting very similar results in distribution, not sure what you are trying to discuss or if its relevant here? Apart from there is obviously some math rounding error in your test code as totals are 96% and 97%

ARTEMIS-2236
============================
2 % for consumer = 1
29 % for consumer = 2
2 % for consumer = 3
30 % for consumer = 4
2 % for consumer = 5
29 % for consumer = 6
2 % for consumer = 7


ARTEMIS-1451
============================
2 % for consumer = 1
29 % for consumer = 2
2 % for consumer = 3
30 % for consumer = 4
2 % for consumer = 5
30 % for consumer = 6
2 % for consumer = 7

franz1981:
Thanks to have verified, my laptop was giving weird results probably due to low core numbers :(

michaelandrepearce:
@franz1981 anything else do shout, it is appreciated to have the review :) as you know we enjoy the debates .. hehehe.

Can i leave with you to merge once you've got yourself happy. Once merged, ill then do a cherry pick to 2.6.x

clebertsuconic:
Oh shoot.. I don't know how I missed this.. I'm the one who merged it.. 

I wouldn't have merged it if I had fully understood the change.

I'm merging it now, cherry-picking it into 2.6.x

PR #2516, opened 2019-01-24
ARTEMIS-2192 fix listener for LegacyLDAPSecuritySettingPlugin

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LegacyLDAPSecuritySettingPlugin.java
```diff
@@ -333,14 +333,14 @@ private void processSearchResult(Map<String, Set<Role>> securityRoles,
...
-      if (rdns.size() != 3) {
+      if (rdns.size() < 3) {
          if (logger.isDebugEnabled()) {
             logger.debug("\tSkipping unexpected search result with " + rdns.size() + " RDNs.");
          }
          return;
       }
       // we can count on the RNDs being in order from right to left
-      Rdn rdn = rdns.get(0);
+      Rdn rdn = rdns.get(rdns.size() - 3);
```
michaelandrepearce: Im not entirely sure what this code is doing and what expected results or ldap setup must be followed, but i do know from LDAP its possible to get duplicate rdns, so what occurs if rdns is > 3 and a dn is duplicated

e.g.

uid=john.doe,ou=People,dc=example,dc=com

dc is twice here, what do you need? also would mean you may miss the dn you do want maybe the in this case, would be uid.

michaelandrepearce: tbh the older, hard coded rdns made a bit more sense to me, at least i know what rdn are being expected and looked for, and ordering (to handle duplicates in response) wouldnt be an issue.

jbertram: The problem with the hard-coded rdns is that the plugin won't work for anybody who doesn't use the same values. I don't see duplicates as being a problem with this code. It simply takes the first three rdns and interprets them as permission-type, destination-name, & destination-type. It doesn't care about the rdns past that.

PR #2517, opened 2019-01-24
[ARTEMIS-2171]: ThreadPoolExecutor leak under SM due to lack of privileged block.

ehsavoie:
* Ensuring that all threadPoolFactories are created under a privileged
block so the threads created are under the same AccessControlContext.

Jira: https://issues.apache.org/jira/browse/ARTEMIS-2171

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/DBOption.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/DBOption.java
@@ -26,6 +26,8 @@ import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.ThreadFactory;
 
 import io.airlift.airline.Option;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
 import org.apache.activemq.artemis.cli.commands.ActionContext;
 import org.apache.activemq.artemis.core.config.Configuration;
@@ -223,13 +225,17 @@ public class DBOption extends OptionalLocking {
       this.config = configuration;
       executor = Executors.newFixedThreadPool(5, ActiveMQThreadFactory.defaultThreadFactory());
       executorFactory = new OrderedExecutorFactory(executor);
-
-      scheduledExecutorService = new ScheduledThreadPoolExecutor(configuration.getScheduledThreadPoolMaxSize(), new ThreadFactory() {
+      scheduledExecutorService = new ScheduledThreadPoolExecutor(configuration.getScheduledThreadPoolMaxSize(), AccessController.doPrivileged(new PrivilegedAction<ThreadFactory>() {
          @Override
-         public Thread newThread(Runnable r) {
-            return new Thread(r);
+         public ThreadFactory run() {
+            return new ThreadFactory() {
+               @Override
+               public Thread newThread(Runnable r) {
+                  return new Thread(r);
+               }
+            };
          }
-      });
+      }));
 
       HierarchicalRepository<AddressSettings> addressSettingsRepository = new HierarchicalObjectRepository<>(config.getWildcardConfiguration());
       addressSettingsRepository.setDefault(new AddressSettings());
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
@@ -153,7 +153,12 @@ public abstract class ActiveMQScheduledComponent implements ActiveMQComponent, R
    }
 
    protected ActiveMQThreadFactory getThreadFactory() {
-      return new ActiveMQThreadFactory(this.getClass().getSimpleName() + "-scheduled-threads", false, getThisClassLoader());
+      return AccessController.doPrivileged(new PrivilegedAction<ActiveMQThreadFactory>() {
+         @Override
+         public ActiveMQThreadFactory run() {
+            return new ActiveMQThreadFactory(this.getClass().getSimpleName() + "-scheduled-threads", false, getThisClassLoader());
+         }
+      });
    }
 
    private ClassLoader getThisClassLoader() {
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/NetworkHealthCheck.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/NetworkHealthCheck.java
@@ -157,7 +157,12 @@ public class NetworkHealthCheck extends ActiveMQScheduledComponent {
 
    @Override
    protected ActiveMQThreadFactory getThreadFactory() {
-      return new ActiveMQThreadFactory("NetworkChecker", "Network-Checker-", false, getThisClassLoader());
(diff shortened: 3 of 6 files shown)


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/DBOption.java
```diff
@@ -223,13 +225,17 @@ protected void initializeJournal(Configuration configuration) throws Exception {
       this.config = configuration;
       executor = Executors.newFixedThreadPool(5, ActiveMQThreadFactory.defaultThreadFactory());
       executorFactory = new OrderedExecutorFactory(executor);
-
-      scheduledExecutorService = new ScheduledThreadPoolExecutor(configuration.getScheduledThreadPoolMaxSize(), new ThreadFactory() {
+      scheduledExecutorService = new ScheduledThreadPoolExecutor(configuration.getScheduledThreadPoolMaxSize(), AccessController.doPrivileged(new PrivilegedAction<ThreadFactory>() {
          @Override
-         public Thread newThread(Runnable r) {
-            return new Thread(r);
+         public ThreadFactory run() {
```
michaelandrepearce: I know this inst your change, but it highlights a bigger question, as to why is this not using ActiveMQThreadFactory?

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -153,7 +153,12 @@ public synchronized void start() {
    }
 
    protected ActiveMQThreadFactory getThreadFactory() {
-      return new ActiveMQThreadFactory(this.getClass().getSimpleName() + "-scheduled-threads", false, getThisClassLoader());
+      return AccessController.doPrivileged(new PrivilegedAction<ActiveMQThreadFactory>() {
+         @Override
+         public ActiveMQThreadFactory run() {
+            return new ActiveMQThreadFactory(this.getClass().getSimpleName() + "-scheduled-threads", false, getThisClassLoader());
```
michaelandrepearce: Threads created by ActiveMQThreadFactory should be already created by doPrivileged. how are we sure its not just that the thread issues you are seeing are maybe just where code isn't using ActiveMQThreadFactory

see

   @Override
   public Thread newThread(final Runnable command) {
      // create a thread in a privileged block if running with Security Manager
      if (acc != null) {
         return AccessController.doPrivileged(new ThreadCreateAction(command), acc);
      } else {
         return createThread(command);
      }
   }

artemis-jms-server/src/main/java/org/apache/activemq/artemis/jms/bridge/impl/JMSBridgeImpl.java
```diff
@@ -1598,25 +1598,30 @@ private static void copyProperties(final Message msg) throws JMSException {
...
-         public Thread newThread(Runnable r) {
-            final Thread thr = new Thread(group, r);
-            if (moduleTccl != null) {
-               AccessController.doPrivileged(new PrivilegedAction() {
-                  @Override
-                  public Object run() {
-                     thr.setContextClassLoader(moduleTccl);
-                     return null;
+         public ThreadFactory run() {
+            return new ThreadFactory() {
```
michaelandrepearce: I know this inst your change, but it highlights a bigger question, as to why is this not using ActiveMQThreadFactory?

As this creates threads with the captured context.

michaelandrepearce:
This **really** needs a test case, to a) validate the fix, and b) ensure no regression.

michaelandrepearce:
This really needs a test case

A) validate the fix (and validate any alternative proposals)
B) ensure no regression

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/NetworkHealthCheck.java
```diff
@@ -157,7 +157,12 @@ public NetworkHealthCheck parseURIList(String addressList) {
 
    @Override
    protected ActiveMQThreadFactory getThreadFactory() {
-      return new ActiveMQThreadFactory("NetworkChecker", "Network-Checker-", false, getThisClassLoader());
```
clebertsuconic: This is just creating a ThreadFactory.. is it really needed?

I would expect a security check around new Threads, but against a new ActiveMQThreadFactory? 

are you sure about it? just checking it's not a typo?

michaelandrepearce: I'd like to see a test case on this. I think the issue might be we have some ThreadFactories which are not using the ActiveMQThreadFactory which should imo ensure the privileged access, and actually we just need to fix those up.

ehsavoie: We run WildFly test suite with a security manager on (which is how this leak was found) BUT I don't have the runs where this happened so I don't know the symptoms :( for a test ensure that this is resolved.

michaelandrepearce: I mean we need a test in ActiveMQ Artemis.

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
ehsavoie: Then why do you still 'protect' ActiveMQThreadFactory like in https://github.com/apache/activemq-artemis/blob/master/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java#L350 maybe those blocks should go away too ?

michaelandrepearce: As i said if you can supply a test with this PR to recreate the issue. then we can validate what does or what does not need to be changed.

michaelandrepearce: Great stuff.

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/NetworkHealthCheck.java
michaelandrepearce: So we need this converted to a test, e.g. integration test case, with this PR which shows this, given a certain configuration and a way to automate the detection.

ehsavoie: Well it seems that the increase is stopping at 28 threads, so it is not leaking

ehsavoie: I don't see it on WildFly with 2.6.4 if I wait sufficiently for the polls to decrease

ehsavoie: I think the increase stops when the default size is reached so this doesn't look like an issue

michaelandrepearce: @ehsavoie can you close this PR then? And the Jira you opened.

michaelandrepearce: @ehsavoie please close this then.

PR #2520, opened 2019-01-24
ARTEMIS-2238 Enhancement to queueQuery on producer

clebertsuconic:


artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java
```diff
@@ -533,6 +496,52 @@ private void doSendx(ActiveMQDestination destination,
...
+            throw JMSExceptionHelper.convertFromActiveMQException(e);
+         }
+      }
+   }
+
+   private void checkQueue(ActiveMQDestination destination,
+                                SimpleString address,
+                                ClientSession clientSession,
+                                ClientSession.AddressQuery query) throws ActiveMQException, InvalidDestinationException {
+      if (destination.isQueue() && query.isAutoCreateQueues()) {
```
michaelandrepearce: should this not also, dare i say check: query.isAutoCreateAddresses()

michaelandrepearce: e.g. no point calling,          clientSession.createAddress(address, RoutingType.ANYCAST, true); 

if its now allowed....

clebertsuconic: I agree with you. although It wasn't part of my original change.. I was just fixing a bug

clebertsuconic: let me rewrite this

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java
```diff
@@ -533,6 +496,52 @@ private void doSendx(ActiveMQDestination destination,
...
 
+   private void checkDestination(ActiveMQDestination destination,
+                                 SimpleString address,
+                                 ClientSession clientSession) throws JMSException {
+      if (!connection.containsKnownDestination(address)) {
+         try {
+            ClientSession.AddressQuery query = clientSession.addressQuery(address);
+
+            if (!query.isExists()) {
+               checkQueue(destination, address, clientSession, query);
```
michaelandrepearce: this is missing added the destination (another bug introduced by the original change)

michaelandrepearce: e.g. if we are succesfull, then the address becomes known right?

clebertsuconic: it would always add the end, but I'm rewriting the logic now

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java
```diff
@@ -533,6 +496,58 @@ private void doSendx(ActiveMQDestination destination,
...
+
+      // TODO: What to do with FQQN
+      if (!connection.containsKnownDestination(address)) {
+         try {
+            ClientSession.AddressQuery addressQuery = clientSession.addressQuery(address);
+
+            boolean addressExists = addressQuery.isExists();
+            // first we check the address existence, and autoCreate it if allowed in case it does not exists
+
+            if (!addressExists && addressQuery.isAutoCreateAddresses()) {
```
michaelandrepearce: We should only create an address for a JMSQueue if we can create both the Address and Queue, else you could end up creating an address, but not the queue (old code had correct logic here)

clebertsuconic: I don't think it would be an issue.. but I have ammended it...
I will create a test..


let me run tests on it.. please don't merge this.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/QueueAutoCreationTest.java
```diff
@@ -133,6 +133,36 @@ public void testAutoCreateOnTopic() throws Exception {
       Assert.assertTrue(((ActiveMQConnection)connection).containsKnownDestination(addressName));
    }
 
+   @Test //(timeout = 30000)
+   // QueueAutoCreationTest was created to validate auto-creation of queues
+   // and this test was added to validate a regression: https://issues.apache.org/jira/browse/ARTEMIS-2238
+   public void testAutoCreateOnAddressOnly() throws Exception {
```
clebertsuconic: this is an incomplete test.. I will finish this before merging

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/QueueAutoCreationTest.java
```diff
@@ -114,7 +114,7 @@ public void testHugeString() throws Exception {
    }
 
 
-   @Test(timeout = 30000)
+   @Test //(timeout = 30000)
```
michaelandrepearce: ? is this now broker?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/QueueAutoCreationTest.java
michaelandrepearce: please address my other comments before merging

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java
```diff
@@ -533,6 +499,67 @@ private void doSendx(ActiveMQDestination destination,
...
+                     logger.debug("Address " + address + " was not created because we would not have permission to create queue");
+                  }
+                  // if it can't create the internal queue on JMS Queues, why bother creating the address, just mark it false now
+                  addressExists = false;
+               } else {
+                  RoutingType addressType = destination.isQueue() ? RoutingType.ANYCAST : RoutingType.MULTICAST;
+                  clientSession.createAddress(address, addressType, true);
+                  addressExists = true;
+               }
+            }
```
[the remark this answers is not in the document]
michaelandrepearce: or just simple use generic Destination as it was before even.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java
```diff
@@ -533,6 +499,67 @@ private void doSendx(ActiveMQDestination destination,
...
+                     } catch (ActiveMQQueueExistsException thatsOK) {
+                        // nothing to be done
+                     }
+                  } else {
+                     throw new InvalidDestinationException("Queue " + address + " does not exist");
+                  }
+               }
+            }
+
+            if (!addressExists) {
```
michaelandrepearce: -1 lets keep that exceptions are thrown from explicit points where checks are done and fail, so can be more explicit and also we can debug where the check fail occured by stack trace lines.

also general comment as this is JMS interface, we should use words like Queue, Topic or Destination

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
```diff
@@ -397,6 +380,55 @@ public MessageProducer createProducer(final Destination destination) throws JMSE
       }
    }
 
+   void checkDestination(ActiveMQDestination destination) throws JMSException {
```
franz1981: make it final to avoid overriding it (it shouldn't happen eh)

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
```diff
@@ -397,6 +380,55 @@ public MessageProducer createProducer(final Destination destination) throws JMSE
...
+      SimpleString address = destination.getSimpleAddress();
+      // TODO: What to do with FQQN
+      if (!connection.containsKnownDestination(address)) {
+         try {
```
franz1981: i will pack this slow path in a separate method to make the code more readable (perf are helped as well, but is not relevant)!

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/QueueAutoCreationTest.java
clebertsuconic: this is debug.. I will revert it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/QueueAutoCreationTest.java
clebertsuconic: @michaelandrepearce I thought by adding your commit they would be fixed? the only missing point was a test timeout that I'm reverting.

michaelandrepearce: Indeed they are i wrote this before that.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
clebertsuconic: TBH I hate final methods...

I have needed quite a few times to override them on test cases... so I only reserve the word final where there is a strong "business" case to not let it be overwritten.  but that's jut my opinion.

clebertsuconic:
@michaelandrepearce this is now ready for review and merge...

if you're happy with it.. let me know.. and I can then run a full testsuite and merge it if it all good



and just FYI, I applied one small change to your patch and ammended as yours (easier than asking you to change it)

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java
```diff
@@ -393,57 +395,18 @@ private void doSendx(ActiveMQDestination destination,
...
-               throw new UnsupportedOperationException("Where a default destination is specified " + "for the sender and a destination is " + "specified in the arguments to the send, " + "these destinations must be equal");
-            }
+      } else if (defaultDestination != null) {
+         if (!destination.equals(defaultDestination)) {
+            throw new UnsupportedOperationException("Where a default destination is specified " + "for the sender and a destination is " + "specified in the arguments to the send, " + "these destinations must be equal");
          }
+      } else {
+         session.checkDestination(destination);
 
          address = destination.getSimpleAddress();
```
michaelandrepearce: I just spotted i think this should be outside the if else block otherwise address will be null for default

clebertsuconic: Those lines are the reason why I wanted to run a full testsuite before merging this.

And I think you're right...



I'm adding a commit to be squashed into yours. I will squash it before merging it, but if you could review it please

clebertsuconic:
I pushed it but it still wrong.. hold on.. I will inform it here

clebertsuconic:
@michaelandrepearce now it's ready!

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java
michaelandrepearce: back on a proper pc,

i was meaning only:
         address = destination.getSimpleAddress();
 
if defaultdestination != null, then don't need to do session.checkDestination(destination), as we checked the destination already when we created the message producer

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
michaelandrepearce: +1 same here not the biggest fan of them either, i remember having on another project starting with the name K ;), such issue, and getting people to agree to make it non final was a nightmare.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
michaelandrepearce: future enhancement maybe?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/AutoCreateJmsDestinationTest.java
```diff
@@ -265,6 +283,98 @@ public void testAutoCreateOnReconnect() throws Exception {
       connection.close();
    }
 
+
+   @Test //(timeout = 30000)
```
michaelandrepearce: commented out timeout still....

michaelandrepearce:
+1

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
franz1981: Don't know: the higher level method is already packing everything in a separate method: why wait another PR just to do the same?

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
franz1981: I understand it and I agree, but depends on the visibility/importance of the method: this one if overridden really breaks badly pretty much everything.
I won't never use final for private methods, but for package private ie near-public and important ones, yes. 
As I've said is just style comment, I'm already happy with the PR :+1:

michaelandrepearce:
@clebertsuconic merged, thanks as always for the contribution.

PR #2521, opened 2019-01-24
ARTEMIS-2238 Enhancement to queueQuery on producer

michaelandrepearce:


michaelandrepearce:
@clebertsuconic little less of a complete refactor, but addresses the issues. thought id offer it up as an alternative.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java
```diff
@@ -409,33 +409,42 @@ private void doSendx(ActiveMQDestination destination,
...
-                           session.createTemporaryQueue(destination, RoutingType.ANYCAST, address, null, query);
+                        if (query.isAutoCreateQueues()) {
+                           if (destination.isTemporary()) {
+                              session.createTemporaryQueue(destination, RoutingType.ANYCAST, address, null, query);
+                           } else {
+                              session.createQueue(destination, RoutingType.ANYCAST, address, null, true, true, query);
+                           }
                         } else {
-                           session.createQueue(destination, RoutingType.ANYCAST, address, null, true, true, query);
+                           throw new InvalidDestinationException("JMSQueue " + address + " cannot be created, address exists but autoCreateQueues is " + query.isAutoCreateQueues());
```
clebertsuconic: I would prefer going with my PR. it's simpler after all if you look closely.

clebertsuconic: for example. another issue on the original code, it's never checking the destination on a defaultDestination.

small issue, but my change also addressed that.


I don't know what you think, but the code I wrote it's easier to read.

Although I understand your concern as this is the JMS layer, and TCK is applied here. But I think it's safe to make the change on this case. I would run the full testsuite,

clebertsuconic: hmmm.. there's another check on createProducer.

It's not worth the effort.. I will just take your version.. and keep the test I'm writing.

michaelandrepearce: im just about to send you pr

PR #2522, opened 2019-01-25
ARTEMIS-2239 Zero-copy NIO/MAPPED TimedBuffer

franz1981:
NIO/MAPPED journal types can use directly the buffer of TimedBuffer
to perform file writes, avoiding an expensive copy + zeroing.


artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/buffer/TimedBuffer.java
```diff
@@ -358,13 +358,7 @@ public boolean flushBatch() {
...
                }
 
-               final ByteBuffer bufferToFlush = bufferObserver.newBuffer(bufferSize, pos);
-               //bufferObserver::newBuffer doesn't necessary return a buffer with limit == pos or limit == bufferSize!!
-               bufferToFlush.limit(pos);
-               //perform memcpy under the hood due to the off heap buffer
-               buffer.getBytes(0, bufferToFlush);
-
-               bufferObserver.flushBuffer(bufferToFlush, pendingSync, callbacks);
+               bufferObserver.flushBuffer(buffer.byteBuf(), pendingSync, callbacks);
```
clebertsuconic: I wish it was that simple.

that's only viable if your write operation is synchronous.
That's probably ok with NIO or Mapped, but it's definitely not ok with libaio.

If you allow the inner buffer to be reused like that, TimedBuffer will be making changes to a buffer that's been submitted to the kernel, and you may endup with wrong data.

And that kind of race is hard to catch.

You can only that do this kind of thing on NIO or Mapped, assuming their buffers will be copied to the file right away.

franz1981: It has been solved on the abstract seq file: ASYNCIO is doing is exactly like before (with the copy), while NIO/MAPPED avoid the copy by using a specialized class

clebertsuconic: Hmmm.. but you moved the copy to LocalBufferObserver::flushBuffer.

What's the difference then, you're still making the copy.

clebertsuconic: I see.. you moved the previous logic to the SequentialFile.. nice one!!

What happens to Mapped though, from what I looked it still using a new buffer. is that what you wanted?

franz1981: By default `AbstractSequentialFile` cannot say if the underline file is performing async or sync writes, hence it just copy it, like it was done in the past on `TimedBuffer`, while NIO/MAPPED can assume that writes are synchronous and they will avoid the copy (if possilble, there are some contraints I've written on the code).

franz1981: > What happens to Mapped
It will use another wrapper I've written some time ago, because `MappedSequentialFile` isn't a child of 'AbstractSequentialFile`, here: https://github.com/apache/activemq-artemis/pull/2522/files#diff-6707e3363162061f97af51ef7f8d13a4R256

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/DelegateCallback.java
```diff
@@ -0,0 +1,71 @@
...
+   public void done() {
+      done(delegates);
+   }
+
+   @Override
+   public void onError(final int errorCode, final String errorMessage) {
+      onError(delegates, errorCode, errorMessage);
+   }
+
+   public static void done(Collection<IOCallback> delegates) {
```
michaelandrepearce: This is not specific to DelegateCallback, should this move to IOCallback

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/DelegateCallback.java
```diff
@@ -0,0 +1,71 @@
...
+      delegates.forEach(callback -> {
+         try {
+            callback.done();
+         } catch (Throwable e) {
+            ActiveMQJournalLogger.LOGGER.errorCompletingCallback(e);
+         }
+      });
+   }
+
+   public static void onError(Collection<IOCallback> delegates, int errorCode, final String errorMessage) {
```
michaelandrepearce: ditto as per other static method, thats takes IOCallback

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/AbstractSequentialFile.java
```diff
@@ -301,21 +269,19 @@ protected ByteBuffer newBuffer(int size, int limit) {
    protected class LocalBufferObserver implements TimedBufferObserver {
```
michaelandrepearce: Is it worth extracting this to its own class file.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/nio/NIOSequentialFile.java
```diff
@@ -51,6 +55,34 @@ public NIOSequentialFile(final SequentialFileFactory factory,
       this.maxIO = maxIO;
    }
 
+   @Override
+   protected TimedBufferObserver createTimedBufferObserver() {
+      return new LocalBufferObserver() {
```
michaelandrepearce: Make this its own class, so its not anonymous.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/DelegateCallback.java
franz1981: :+1:  Yep, good idea! And I could move `DelegateCallback` too to be near `IOCallback` too

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/AbstractSequentialFile.java
franz1981: It is not easy: it's `Local` to the owner class in the worst way: it use directly private files as `fileSize` that has a very tricky semantic around. I prefer to left it to stay where it is to reduce the amount of changes.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/nio/NIOSequentialFile.java
franz1981: I can do it, but (for this specific class) I can't see the benefits.
In a future implementation I would like to abstract away things like `AbstractSequentialFile::fileSize` and create 2 separate classes:
`AsyncLocalBufferObserver` and `SyncLocalBufferObserver`.
The former that perform the copy because it assymes asynchrounous writes, the latter not.
But at this point, given that the classes are packed into their owner seq file class probably I could just let them be where/as they are to reduce the code changes.

franz1981:
@michaelandrepearce 

> Have you got any stats to share on this, especially on running on standardish spec server? Im assuming that its quite a good improvement based on it removes copies :) , but be good for the record like....

Not yet: I can just say that on my box `fsync` is such heavy-weight that is difficult to measure the gain of this improvement. 
If I would use a `tmpfs` folder for the journal instead it would just show that zeroing (and copying) is just the major cost and the speedup is so dramatical that is just unrealistic.

@clebertsuconic @michaelandrepearce I will try to get a proper benchmark box to find out this, but nonetheless I'm expecting *at least* a better ability from NIO/MAPPED to maintain the buffer timeout: the next big improvement would be to run `fsync` on a separate "coalescing" thread (similarly to what libaio does on the blocked poll).

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/nio/NIOSequentialFile.java
michaelandrepearce: At least make it a nested class. Anonymous classes whilst not impossible to debug, just having it named makes stacks trace, heap dumps etc that little bit easier to quickly read and understand.

franz1981: Good point about stacks trace bud, will do it immediately!

michaelandrepearce:
Looking promising. @franz1981 with all the changes youve been working on for journals, im not sure what the merge order is. Can this just be merged OR do we need merge another PR first?

franz1981:
This one should be kinda independent from the rest so can be merged without worrying about any conflicts, thanks!!!

PR #2523, opened 2019-01-25
ARTEMIS-1977 ASYNCIO can reduce sys-calls to retrieve I/O events

franz1981:
On LibAIO is possible to retrieve the I/O completion
events without using io_getevents sys-calls by reading
the user-space ring buffer used by the kernel to store them.
This commit include another optimization to avoid
calling a method to obtain the buffers address, saving
safepoint polls, a method call and implicit instance 
checks performed.

### DIFF
--- a/artemis-native/src/main/c/org_apache_activemq_artemis_jlibaio_LibaioContext.c
+++ b/artemis-native/src/main/c/org_apache_activemq_artemis_jlibaio_LibaioContext.c
@@ -36,6 +36,11 @@
 #include "org_apache_activemq_artemis_jlibaio_LibaioContext.h"
 #include "exception_helper.h"
 
+//x86 has a strong memory model and there is no need of HW fences if just Write-Back (WB) memory is used
+#define mem_barrier() __asm__ __volatile__ ("":::"memory")
+#define read_barrier()	__asm__ __volatile__("":::"memory")
+#define store_barrier()	__asm__ __volatile__("":::"memory")
+
 struct io_control {
     io_context_t ioContext;
     struct io_event * events;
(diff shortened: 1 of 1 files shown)


franz1981:
@michaelandrepearce I suppose this one could be of some interest to you :)
@orpiske Please mate I need some help to compile this one for 32 bit, I'm not the good to configure cross-compilation on my box and I haven't an old box to do it...

franz1981:
It should be a nice step for a future optimization that would allow to use 1 core just to retrieve IO events in a busy loop (with some idle strategy, if necessary) to reduce dramatically latencies: DPDK uses a similar approach to improve dramatically networking on user space.

orpiske:
@franz1981 absolutely! I will be happy to help with this one. I have some 32 bits where we can work and test this. I'll take a look at this on Monday.

artemis-native/src/main/c/org_apache_activemq_artemis_jlibaio_LibaioContext.c
```diff
@@ -57,6 +57,92 @@ struct io_control {
...
 
+//These should be used to check if the user_io_getevents is supported, but almost every modern kernel support it
+#define AIO_RING_MAGIC	0xa10a10a1
+#define AIO_RING_INCOMPAT_FEATURES	0
+//memory order relaxed/acquire/release/seq_cst x86_64 fences implementations
+#define mem_barrier() __asm__ __volatile__ ("":::"memory")
+#define read_barrier()	__asm__ __volatile__("lfence":::"memory")
+#define store_barrier()	__asm__ __volatile__("sfence":::"memory")
+#define full_barrier()	__asm__ __volatile__("mfence":::"memory")
+
```
clebertsuconic: Is there any documentation about this?

franz1981:
About the memory barriers? Are the standard implementation used on the c++11 memory model and I have asked a colleague of mine that work on the Linux kernel, specifically on libAIO ie Paolo Bonzini on https://github.com/redhat-cip/fio/blob/master/engines/libaio.c

michaelandrepearce:
@franz1981 whats the compatibility of using the c++ c11 features, e.g. this change with centOS6 / rhel6 default install c support? Im cautious of introducing feature that could break on these platforms.

franz1981:
@michaelandrepearce 
> whats the compatibility of using the c++ c11 features

I don't know , that' s why I've implemented the barriers by myself, instead of using the ones provided by C11. Re the barriers I've implemented they work for any x86_64 platform (the version of Linux won't make any difference, because I'm not using OS features, but just arch-specific instructions). 
I could provide an implementation of x86 (32 bit) and we won't need any support (+ eventual incompatibilities) of C11.

franz1981:
On our CI (that's using a x86_64) it has passed without any issue :+1:

franz1981:
@michaelandrepearce 
> We will need 32 bit support. I remember there being someone in the community on a user thread being on 32 for valid reason.

Yep. that's why I've summoned @orpiske :) I will probablu need some help with CMake (that I'm not that use to it) in order to provide 2 different implementations for the barriers (x86 and x86_64): I'm used to write/read assembly but really bad to use build tools :P

franz1981:
@michaelandrepearce @clebertsuconic 
Today I will talk with some of the kernel guys that has written libAIO: I've taken a deep look to https://elixir.bootlin.com/linux/latest/source/fs/aio.c#L1148 and seems that for x86 I could avoid any hardware barriers.
`lfence,sfence` are hw barriers needed just when used among non-temporal stores ie that won't use the caches, while temporal x86 stores are guaranteed to not being reordered by the processor.
I need to cross-compile the 32 bit version accordly, many thanks to super- @orpiske  for his help!

franz1981:
@michaelandrepearce @clebertsuconic I have verified that no HW fences are needed so now worries about 32/64 bit :)
I've put some more comments and I'm waiting @orpiske to provide the missing bits to allow cross-compilation :+1:

michaelandrepearce:
@orpiske soo... want to come join us as a regular contributor ;) more the merrier....

@franz1981 & @orpiske this is great stuff!!! looking forward to it.

orpiske:
> @orpiske soo... want to come join us as a regular contributor ;) more the merrier....
> 
> @franz1981 & @orpiske this is great stuff!!! looking forward to it.

@michaelandrepearce Thanks. The bulk of work here was from @franz1981. I do plan to contribute more often, indeed :)

franz1981:
@michaelandrepearce 
> this is great stuff!!! looking forward to it.

The commit here is just mine :P
I've removed the part of Otavio for the resons below..
@clebertsuconic @michaelandrepearce 
I need your advice re this one indeed: I think that the changes of Otavio to enable cross-compilation and improve (a lot!!!) the compilation process deserve its own PR, wdyt?
This commit is a change in the mechanics of how we use libAIO and I'm not sure is ok to tied them together: the changes of Otavio are useful per se...

michaelandrepearce:
So if this pr doesnt support cross compilation with out the other then for me it cant be merged without the other. As such id have them togeather

franz1981:
This PR doesn't change anything specific for being 32/64 bit, but like any changes on the native layer, having cross-compilation would help (me or anyone) to provide the 32 bit binaries without pain.
IMHO it should be better to have the cross-compilation PR of Otavio merged first and I would use it to cross-compile this PR, but I prefer to have them separate if, for ANY reasons, one of them will need to be reverted, they weren't tied up together...

clebertsuconic:
I am merging this so I can cherry-pick using some git-magic into the new native project.

PR #2524, opened 2019-01-27
ARTEMIS-2240 ActiveMQThreadPoolExecutor should use LinkedTransferQueue

franz1981:
It contains 2 commits:

- just replacing LinkedBlockingQueue with LinkedTransferQueue
- using specific features of LinkedTransferQueue to simplify the thread-pool implementation

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ActiveMQThreadPoolExecutor.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ActiveMQThreadPoolExecutor.java
@@ -16,11 +16,12 @@
  */
 package org.apache.activemq.artemis.utils;
 
-import java.util.concurrent.LinkedBlockingQueue;
+import java.util.Objects;
+import java.util.concurrent.LinkedTransferQueue;
+import java.util.concurrent.RejectedExecutionHandler;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 /*
  * ActiveMQThreadPoolExecutor: a special ThreadPoolExecutor that combines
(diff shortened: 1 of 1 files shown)


franz1981:
On my box I'm getting the same performance of the master version for single producer/single consumer workloads and a better scalability with 8 producers/8 consumers (on different queues) for both the commits I've attached (the second one, in particular).
I'm putting this for review and for validation: if anyone has a better machine to validate its behaviour is wellcome.

A conservative approach will suggest to just use the first commit ie a simple q replacement, but I the second commit has more to offer given that would allow to use what this q offer out of the box and that seems very similar to what the thread pool was delivering in the implementation on master.

franz1981:
@michaelandrepearce @clebertsuconic If any of these commits will improve scalability (LinkedBlockingQueue is using a lock for any of its operations!) I will close this one :+1:

michaelandrepearce:
I am see a perf gain, in commit 1, i see if anything (its a micro second here or there) negative on commit 2.

I actually found something positive though, if you take the master version that we recently reverted as it kept adding new threads, vs trying to see if a thread was idle.

Change it to LinkedTransferQueue and then in the offer section, which it was just returning false and the core perf issue with it. And instead use your tryTransfer(runnable) which is available with switching to LinkedTransferQueue, this actually seems to be really good, as in it now will send a runnable to a idle thread, (e.g. tryTransfer) but also becomes very lock less, a simpler code, which i guess what it was trying to achieve.

sample:


```
/*
 * ActiveMQThreadPoolExecutor: a special ThreadPoolExecutor that combines
 * the benefits of a cached executor and a fixed size executor.
 * Similar to a cached executor, threads exceeding the core size are only created on demand,
 * and will be removed after idling for a specified keep time.
 * But in contrast to a standard cached executor, tasks are queued if the
 * maximum pool size if reached, instead of rejected.
 */
public class ActiveMQThreadPoolExecutor extends ThreadPoolExecutor {

   // Handler executed when a task is submitted and a new thread cannot be created (because maxSize was reached)
   // It queues the task on the executors's queue (using the add() method, see ThreadPoolQueue class below)
   private static final RejectedExecutionHandler QUEUE_EXECUTION_HANDLER = (r, e) -> {
      if (!e.isShutdown()) {
         e.getQueue().add(r);
      }
   };

   // A specialized LinkedBlockingQueue that takes new elements by calling add() but not offer()
   // This is to force the ThreadPoolExecutor to always create new threads and never queue
   private static class ThreadPoolQueue extends LinkedTransferQueue<Runnable> {

      @Override
      public boolean offer(Runnable runnable) {
         return tryTransfer(runnable);
      }

      @Override
      public boolean add(Runnable runnable) {
         return super.offer( runnable );
      }
   }

   public ActiveMQThreadPoolExecutor(int coreSize, int maxSize, long keep, TimeUnit keepUnits, ThreadFactory factory) {
      super( coreSize, maxSize, keep, keepUnits, new ThreadPoolQueue(), factory, QUEUE_EXECUTION_HANDLER );
   }
}


```

franz1981:
Thanks @michaelandrepearce to have looked into this, I will change it accordingly to the last comment (nice simple and effective, I like it!) and left the PR here for @clebertsuconic :)

michaelandrepearce:
@franz1981 a slight improvement, i spotted old master wasn't honoring the pre 2.4 original behaviour of on shutdown, rejection would then use default policy of the executor (AbortPolicy) which would throw rejection exception. a simple correction.

-->
 
```
/*
 * ActiveMQThreadPoolExecutor: a special ThreadPoolExecutor that combines
 * the benefits of a cached executor and a fixed size executor.
 * Similar to a cached executor, threads exceeding the core size are only created on demand,
 * and will be removed after idling for a specified keep time.
 * But in contrast to a standard cached executor, tasks are queued if the
 * maximum pool size if reached, instead of rejected.
 */
public class ActiveMQThreadPoolExecutor extends ThreadPoolExecutor {

   /**
    * The default rejected execution handler
    */
   private static final RejectedExecutionHandler defaultHandler = new AbortPolicy();

   // Handler executed when a task is submitted and a new thread cannot be created (because maxSize was reached)
   // It queues the task on the executors's queue (using the add() method, see ThreadPoolQueue class below)
   private static class QueueExecutionHandler implements RejectedExecutionHandler {

      private final RejectedExecutionHandler handler;

      private QueueExecutionHandler(RejectedExecutionHandler handler) {
         this.handler = handler;
      }

      @Override
      public void rejectedExecution(Runnable r, ThreadPoolExecutor executor) {
         if (executor.isShutdown() || !executor.getQueue().add(r)) {
            handler.rejectedExecution(r, executor);
         }
      }
   }


   // A specialized LinkedBlockingQueue that takes new elements by calling add() but not offer()
   // This is to force the ThreadPoolExecutor to always create new threads and never queue
   private static class ThreadPoolQueue extends LinkedTransferQueue<Runnable> {

      @Override
      public boolean offer(Runnable runnable) {
         return tryTransfer(runnable);
      }

      @Override
      public boolean add(Runnable runnable) {
         return super.offer( runnable );
      }
   }

   public ActiveMQThreadPoolExecutor(int coreSize, int maxSize, long keep, TimeUnit keepUnits, ThreadFactory factory) {
      this(coreSize, maxSize, keep, keepUnits, factory, defaultHandler);
   }

   public ActiveMQThreadPoolExecutor(int coreSize, int maxSize, long keep, TimeUnit keepUnits, ThreadFactory factory, RejectedExecutionHandler handler) {
      super(coreSize, maxSize, keep, keepUnits, new ThreadPoolQueue(), factory, new QueueExecutionHandler(handler));
   }
}

```

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ActiveMQThreadPoolExecutor.java
```diff
@@ -29,125 +29,57 @@
...
-      // The delta is incremented if a thread becomes idle or if a task is taken from the queue.
-      // The delta is decremented if a thread leaves idle state or if a task is added to the queue.
-      private static final AtomicIntegerFieldUpdater<ThreadPoolQueue> DELTA_UPDATER = AtomicIntegerFieldUpdater.newUpdater(ThreadPoolQueue.class, "threadTaskDelta");
-      private volatile int threadTaskDelta = 0;
+      private final RejectedExecutionHandler handler;
 
-      public void setExecutor(ActiveMQThreadPoolExecutor executor) {
-         this.executor = executor;
+      private QueueExecutionHandler(RejectedExecutionHandler handler) {
+         this.handler = handler;
```
michaelandrepearce: Add 

if (handler == null)
            throw new NullPointerException();

franz1981:
@michaelandrepearce nice catch!! 
you've already measured with your box how the last version is performing?
I've run the same bench I'm using and it seems really great: hopefully will scale a lot better too :+1:

franz1981:
@michaelandrepearce I'm very happy (I hope that perf are nice as well), because it makes much more simple to understand what the code does now

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ActiveMQThreadPoolExecutor.java
```diff
@@ -29,125 +30,58 @@
...
 
-      @Override
-      public Runnable take() throws InterruptedException {
-         // Increment the delta as a thread becomes idle
-         // by waiting for a task to take from the queue
-         DELTA_UPDATER.incrementAndGet(this);
-
-
-         Runnable runnable = null;
+   // A specialized LinkedBlockingQueue that takes new elements by calling add() but not offer()
```
michaelandrepearce: Need to change this, sorry in my example i was doing quickly, so comments etc left unchanged.

Maybe

----------------------

A specialized LinkedTransderQueue that takes new elements on offer, by trying to transfer if if idle thread by calling tryTransfer(), calling add() will force the queue to take a new element.

This is to force the ThreadPoolExecutor to offer successfully when an idle thread exists, but otherwise create new threads and never queue.
----------------------

franz1981: I don't know bud, just looking at the doc of `tryTransfert` is very clear (for a concurrent q doc obviously :P).
But If you feel that adding some doc could help I will add some for sure :+1:

michaelandrepearce: Well then we remove the current doc at least. The doc atm != the code

franz1981: True: just a qq...have you tried this new impl too?
I've just run a simple load and it seems the best of the 3, but you know...I'm not quite trusty of my box :)

michaelandrepearce: I did but i didnt capture results. It was a bit better than your commit 1, for the use case i was checking for. If you need results, ill have to try do later in the week, busy week again.

franz1981: @michaelandrepearce No worries Michael: I can ask @clebertsuconic super-charged machine as well to run a test with it :P
The primary concerns/motivation of this PR are:

- better scalability (ReentrantLock of LinkedBlockingQueue makes me super-sad!!) without sacrificing low number, but fast pairs connected
- reduction of code complexity

michaelandrepearce: @franz1981 

test is run 20 times without restart of the process, and the provided times are us average for each run of 100,000, was tested on a stock dell r630, with persistance disabled to eliminate disk io from impacting results.

commit 1)

175
165
66
65
62
60
98
98
45
45
50
50
45
52
44
44
46
47
44

latest version)

165
161
69
70
52
51
104
47
45
45
53
53
51
44
42
42
43
44
45

I think this looks good to merge.

PR #2525, opened 2019-01-28
ARTEMIS-2241 Support direct deliver for InVMAcceptors

michaelandrepearce:
Push isDirectDeliver method from netty impl, to the Connection interface
Add support to InVMConnection for isDirectDeliver flag and ability to set via config, defaulting to false, to keep current default behavior. 
Extend DirectDeliverTest to check InVM as well.


michaelandrepearce:
@franz1981 is that thumbs up a signal you're happy for this to merge?

franz1981:
@michaelandrepearce It seems good to me! I wanted the opinion of @clebertsuconic too :)
If possible (just a personal way to manage these things) I always prefer to have at least 2 reviewers for any PR...Just to have 3 different people that could form a balanced quorum :)

PR #2527, opened 2019-01-28
NO-JIRA add logging to LegacyLDAPSecuritySettingPlugin

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LegacyLDAPSecuritySettingPlugin.java
```diff
@@ -303,6 +303,13 @@ private LegacyLDAPSecuritySettingPlugin populateSecurityRoles() {
 
       securityRoles = new HashMap<>();
       try {
+         if (logger.isDebugEnabled()) {
+            logger.debug("Performing LDAP search: " + destinationBase);
```
clebertsuconic: I don't mean to be annoying, but when I deal with big loggers.. it's a lot easier if this was a single logger.debug call


easier to identify on the logger output... and if you use customer environments even more.  A lot of our users are using logger aggregators now.. so it would be a lot easier with a single call here.

jbertram: I had the same thought actually, but I wasn't sure if I was being too picky. I'll change it.

PR #2528, opened 2019-01-29
ARTEMIS-2226 last consumer connection should close the previous consu…

onlyMIT:
If use MQTT protocol, multiple consumers using the same clientId in the cluster, the last consumer connection should close the previous consumer connection! 
There are two solutions to this issue:

> 1. Add a new cluster notification when the client’s connections connect to Artemis, close existing connections that use the same clientId when processing cluster-notification. This solution is the best way to address the MQTT agreement.But I don't think it's necessary to solve this problem by sending a cluster-notification every time client connect.

> 2.The solution is this RP used, close existing connections that use the same clientId when processing the CONSUMER_CREATE  cluster-notification. No additional cluster-notification added. Only when a consumer using the same clientId consumes the same topic will it trigger a close connection.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ManagementHelper.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ManagementHelper.java
@@ -84,6 +84,10 @@ public final class ManagementHelper {
 
    public static final SimpleString HDR_MESSAGE_ID = new SimpleString("_AMQ_Message_ID");
 
+   public static final SimpleString HDR_PROTOCOL_NAME = new SimpleString("_AMQ_Protocol_Name");
+
+   public static final SimpleString HDR_CLIENT_ID = new SimpleString("_AMQ_Client_ID");
+
    // Attributes ----------------------------------------------------
 
    // Static --------------------------------------------------------
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
@@ -162,14 +162,7 @@ public class MQTTConnectionManager {
    }
 
    private MQTTSessionState getSessionState(String clientId) {
-      /* [MQTT-3.1.2-4] Attach an existing session if one exists otherwise create a new one. */
-      MQTTSessionState state = MQTTSession.SESSIONS.get(clientId);
-      if (state == null) {
-         state = new MQTTSessionState(clientId);
-         MQTTSession.SESSIONS.put(clientId, state);
-      }
-
-      return state;
+      return session.getProtocolManager().getSessionState(clientId);
    }
 
    private String validateClientId(String clientId, boolean cleanSession) {
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
@@ -18,9 +18,9 @@ package org.apache.activemq.artemis.core.protocol.mqtt;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
@@ -30,6 +30,9 @@ import io.netty.handler.codec.mqtt.MqttEncoder;
 import io.netty.handler.codec.mqtt.MqttMessage;
 import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
 import org.apache.activemq.artemis.api.core.BaseInterceptor;
+import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.api.core.management.CoreNotificationType;
+import org.apache.activemq.artemis.api.core.management.ManagementHelper;
 import org.apache.activemq.artemis.core.remoting.impl.netty.NettyServerConnection;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
 import org.apache.activemq.artemis.core.server.management.Notification;
@@ -40,11 +43,12 @@ import org.apache.activemq.artemis.spi.core.protocol.ProtocolManagerFactory;
 import org.apache.activemq.artemis.spi.core.protocol.RemotingConnection;
 import org.apache.activemq.artemis.spi.core.remoting.Acceptor;
 import org.apache.activemq.artemis.spi.core.remoting.Connection;
+import org.apache.activemq.artemis.utils.collections.TypedProperties;
 
 /**
  * MQTTProtocolManager
  */
-class MQTTProtocolManager extends AbstractProtocolManager<MqttMessage, MQTTInterceptor, MQTTConnection> implements NotificationListener {
+public class MQTTProtocolManager extends AbstractProtocolManager<MqttMessage, MQTTInterceptor, MQTTConnection> implements NotificationListener {
 
    private static final List<String> websocketRegistryNames = Arrays.asList("mqtt", "mqttv3.1");
 
@@ -55,18 +59,53 @@ class MQTTProtocolManager extends AbstractProtocolManager<MqttMessage, MQTTInter
    private final List<MQTTInterceptor> outgoingInterceptors = new ArrayList<>();
 
    //TODO Read in a list of existing client IDs from stored Sessions.
-   private Map<String, MQTTConnection> connectedClients = new ConcurrentHashMap<>();
+   private final Map<String, MQTTConnection> connectedClients;
+   private final Map<String, MQTTSessionState> sessionStates;
 
    MQTTProtocolManager(ActiveMQServer server,
+                       Map<String, MQTTConnection> connectedClients,
+                       Map<String, MQTTSessionState> sessionStates,
                        List<BaseInterceptor> incomingInterceptors,
                        List<BaseInterceptor> outgoingInterceptors) {
       this.server = server;
+      this.connectedClients = connectedClients;
+      this.sessionStates = sessionStates;
       this.updateInterceptors(incomingInterceptors, outgoingInterceptors);
+      server.getManagementService().addNotificationListener(this);
    }
 
    @Override
    public void onNotification(Notification notification) {
-      // TODO handle notifications
+      if (!(notification.getType() instanceof CoreNotificationType))
+         return;
+
+      CoreNotificationType type = (CoreNotificationType) notification.getType();
+      if (type != CoreNotificationType.SESSION_CREATED)
+         return;
+
+      TypedProperties props = notification.getProperties();
+
+      SimpleString protocolName = props.getSimpleStringProperty(ManagementHelper.HDR_PROTOCOL_NAME);
+
+      //Only process SESSION_CREATED notifications for the MQTT protocol
+      if (protocolName == null || !protocolName.toString().equals(MQTTProtocolManagerFactory.MQTT_PROTOCOL_NAME))
+         return;
+
+      int distance = props.getIntProperty(ManagementHelper.HDR_DISTANCE);
+
+      //distance > 0 means only processing notifications which are received from other nodes in the cluster
+      if (distance > 0) {
+         String clientId = props.getSimpleStringProperty(ManagementHelper.HDR_CLIENT_ID).toString();
+         /*
+          * If there is a connection in the node with the same clientId as the value of the "_AMQ_Client_ID" attribute
+          * in the SESSION_CREATED notification, you need to close this connection.
+          * Avoid consumers with the same client ID in the cluster appearing at different nodes at the same time.
+          */
+         MQTTConnection mqttConnection = connectedClients.get(clientId);
+         if (mqttConnection != null) {
+            mqttConnection.destroy();
+         }
+      }
    }
 
    @Override
@@ -194,5 +233,18 @@ class MQTTProtocolManager extends AbstractProtocolManager<MqttMessage, MQTTInter
     */
    public MQTTConnection addConnectedClient(String clientId, MQTTConnection connection) {
       return connectedClients.put(clientId, connection);
+   }
+
+   public MQTTSessionState getSessionState(String clientId) {
+      /* [MQTT-3.1.2-4] Attach an existing session if one exists otherwise create a new one. */
+      return sessionStates.computeIfAbsent(clientId, MQTTSessionState::new);
+   }
+
+   public MQTTSessionState removeSessionState(String clientId) {
+      return sessionStates.remove(clientId);
+   }
+
+   public Map<String, MQTTSessionState> getSessionStates() {
+      return new HashMap<>(sessionStates);
    }
 }
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSession.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSession.java
@@ -17,10 +17,7 @@
 
 package org.apache.activemq.artemis.core.protocol.mqtt;
 
-import java.util.HashMap;
-import java.util.Map;
 import java.util.UUID;
-import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.activemq.artemis.core.config.WildcardConfiguration;
 import org.apache.activemq.artemis.core.message.impl.CoreMessageObjectPools;
@@ -29,8 +26,6 @@ import org.apache.activemq.artemis.core.server.impl.ServerSessionImpl;
 import org.apache.activemq.artemis.spi.core.protocol.SessionCallback;
 
 public class MQTTSession {
-
-   static Map<String, MQTTSessionState> SESSIONS = new ConcurrentHashMap<>();
 
    private final String id = UUID.randomUUID().toString();
 
@@ -108,7 +103,7 @@ public class MQTTSession {
 
          if (isClean()) {
             clean();
-            SESSIONS.remove(connection.getClientID());
+            protocolManager.removeSessionState(connection.getClientID());
          }
       }
       stopped = true;
@@ -201,7 +196,4 @@ public class MQTTSession {
       return coreMessageObjectPools;
    }
 
-   public static Map<String, MQTTSessionState> getSessions() {
-      return new HashMap<>(SESSIONS);
-   }
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
@@ -59,6 +59,7 @@ import org.apache.activemq.artemis.api.core.management.AddressControl;
 import org.apache.activemq.artemis.api.core.management.BridgeControl;
 import org.apache.activemq.artemis.api.core.management.CoreNotificationType;
 import org.apache.activemq.artemis.api.core.management.DivertControl;
+import org.apache.activemq.artemis.api.core.management.ManagementHelper;
 import org.apache.activemq.artemis.api.core.management.Parameter;
 import org.apache.activemq.artemis.api.core.management.QueueControl;
 import org.apache.activemq.artemis.core.client.impl.Topology;
@@ -2954,7 +2955,16 @@ public class ActiveMQServerControlImpl extends AbstractControl implements Active
       if (!(notification.getType() instanceof CoreNotificationType))
          return;
       CoreNotificationType type = (CoreNotificationType) notification.getType();
-      TypedProperties prop = notification.getProperties();
+      if (type == CoreNotificationType.SESSION_CREATED) {
+         TypedProperties props = notification.getProperties();
+         /*
+          * If the SESSION_CREATED notification is received from another node in the cluster, no broadcast call is made.
+          * To keep the original logic to avoid calling the broadcast multiple times for the same SESSION_CREATED notification in the cluster.
+          */
+         if (props.getIntProperty(ManagementHelper.HDR_DISTANCE) > 0) {
+            return;
+         }
+      }
 
       this.broadcaster.sendNotification(new Notification(type.toString(), this, notifSeq.incrementAndGet(), notification.toString()));
    }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionBridge.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionBridge.java
@@ -233,6 +233,8 @@ public class ClusterConnectionBridge extends BridgeImpl {
                                                    " AND " +
                                                    ManagementHelper.HDR_NOTIFICATION_TYPE +
                                                    " IN ('" +
+                                                   CoreNotificationType.SESSION_CREATED +
+                                                   "','" +
                                                    CoreNotificationType.BINDING_ADDED +
                                                    "','" +
                                                    CoreNotificationType.BINDING_REMOVED +
@@ -252,6 +254,8 @@ public class ClusterConnectionBridge extends BridgeImpl {
                                                    flowRecord.getMaxHops() +
                                                    " AND (" +
                                                    createSelectorFromAddress(appendIgnoresToFilter(flowRecord.getAddress())) +
+                                                   ") AND (" +
+                                                   createPermissiveManagementNotificationToFilter() +
                                                    ")");
 
          sessionConsumer.createTemporaryQueue(managementNotificationAddress, notifQueueName, filter);
@@ -351,10 +355,20 @@ public class ClusterConnectionBridge extends BridgeImpl {
       }
       filterString += "!" + storeAndForwardPrefix;
       filterString += ",!" + managementAddress;
-      filterString += ",!" + managementNotificationAddress;
       return filterString;
    }
 
+   /**
+    * Create a filter rule,in addition to SESSION_CREATED notifications, all other notifications using managementNotificationAddress
+    * as the routing address will be filtered.
+    * @return
+    */
+   private String createPermissiveManagementNotificationToFilter() {
+      StringBuilder filterBuilder = new StringBuilder(ManagementHelper.HDR_NOTIFICATION_TYPE).append(" = '")
+              .append(CoreNotificationType.SESSION_CREATED).append("' OR (").append(ManagementHelper.HDR_ADDRESS)
+              .append(" NOT LIKE '").append(managementNotificationAddress).append("%')");
+      return filterBuilder.toString();
+   }
 
    @Override
    protected void nodeUP(TopologyMember member, boolean last) {
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
@@ -272,6 +272,8 @@ public class ServerSessionImpl implements ServerSession, FailureListener {
       if (!xa) {
          tx = newTransaction();
       }
+      //When the ServerSessionImpl initialization is complete, need to create and send a SESSION_CREATED notification.
+      sendSessionNotification(CoreNotificationType.SESSION_CREATED);
    }
 
    // ServerSession implementation ---------------------------------------------------------------------------
@@ -422,11 +424,26 @@ public class ServerSessionImpl implements ServerSession, FailureListener {
          }
 
          closed = true;
+         //When the ServerSessionImpl is closed, need to create and send a SESSION_CLOSED notification.
+         sendSessionNotification(CoreNotificationType.SESSION_CLOSED);
 
          if (server.hasBrokerSessionPlugins()) {
             server.callBrokerSessionPlugins(plugin -> plugin.afterCloseSession(this, failed));
          }
       }
+   }
+
+   private void sendSessionNotification(final CoreNotificationType type) throws Exception {
+      final TypedProperties props = new TypedProperties();
+      props.putSimpleStringProperty(ManagementHelper.HDR_CONNECTION_NAME, SimpleString.toSimpleString(this.getConnectionID().toString()));
+      props.putSimpleStringProperty(ManagementHelper.HDR_USER, SimpleString.toSimpleString(this.getUsername()));
+      props.putSimpleStringProperty(ManagementHelper.HDR_SESSION_NAME, SimpleString.toSimpleString(this.getName()));
+
+      props.putSimpleStringProperty(ManagementHelper.HDR_CLIENT_ID, SimpleString.toSimpleString(this.remotingConnection.getClientID()));
+      props.putSimpleStringProperty(ManagementHelper.HDR_PROTOCOL_NAME, SimpleString.toSimpleString(this.remotingConnection.getProtocolName()));
+      props.putSimpleStringProperty(ManagementHelper.HDR_ADDRESS, managementService.getManagementNotificationAddress());
+      props.putIntProperty(ManagementHelper.HDR_DISTANCE, 0);
+      managementService.sendNotification(new Notification(null, type, props));
    }
 
    private void securityCheck(SimpleString address, CheckType checkType, SecurityAuth auth) throws Exception {
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManagerFactory.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManagerFactory.java
@@ -19,6 +19,7 @@ package org.apache.activemq.artemis.core.protocol.mqtt;
 
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.activemq.artemis.api.core.BaseInterceptor;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
@@ -37,12 +38,15 @@ public class MQTTProtocolManagerFactory extends AbstractProtocolManagerFactory<M
 
    private static final String[] SUPPORTED_PROTOCOLS = {MQTT_PROTOCOL_NAME};
 
+   private final Map<String, MQTTConnection> connectedClients  = new ConcurrentHashMap<>();
+   private final Map<String, MQTTSessionState> sessionStates = new ConcurrentHashMap<>();
+
    @Override
    public ProtocolManager createProtocolManager(ActiveMQServer server,
                                                 final Map<String, Object> parameters,
                                                 List<BaseInterceptor> incomingInterceptors,
                                                 List<BaseInterceptor> outgoingInterceptors) throws Exception {
-      return BeanSupport.setData(new MQTTProtocolManager(server, incomingInterceptors, outgoingInterceptors), parameters);
+      return BeanSupport.setData(new MQTTProtocolManager(server, connectedClients, sessionStates, incomingInterceptors, outgoingInterceptors), parameters);
    }
 
    @Override
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/AbstractAcceptor.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/AbstractAcceptor.java
@@ -17,6 +17,7 @@
 
 package org.apache.activemq.artemis.core.remoting.impl;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
@@ -41,6 +42,10 @@ public abstract class AbstractAcceptor implements Acceptor {
       for (ProtocolManager manager : protocolMap.values()) {
          manager.updateInterceptors(incomingInterceptors, outgoingInterceptors);
       }
+   }
+
+   public Map<String, ProtocolManager> getProtocolMap() {
+      return Collections.unmodifiableMap(protocolMap);
    }
 
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionImpl.java
@@ -1078,6 +1078,10 @@ public final class ClusterConnectionImpl implements ClusterConnection, AfterConn
                doUnProposalReceived(message);
                break;
             }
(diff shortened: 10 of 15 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -343,6 +347,18 @@ public void onNotification(final Notification notification) {
                   if (binding != null) {
                      // We have a local queue
                      Queue queue = (Queue) binding.getBindable();
+                     //If the client ID represents a client already connected to the server then the server MUST disconnect the existing client.
+                     //Avoid consumers with the same client ID in the cluster appearing at different nodes at the same time
+                     SimpleString protocolName = props.getSimpleStringProperty(ManagementHelper.HDR_PROTOCOL_Name);
```
franz1981: I'm not that familiar with the MQTT protocol, but I can give you an idea: instead of using a SimpleString property, that is costly to be compared/transferred/encoded/decoded why not define a single byte (or short to be more future-proof) property? 
It would be less human-readable, but more cheaper for a machine: in addition, a Short/Byte tends to be pooled on the ranges of -127<->128 AFAIK or something similar, so only the property name + HashMap entry will produce some garbage.

onlyMIT: Thanks, very nice advice. i will done

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -343,6 +347,18 @@ public void onNotification(final Notification notification) {
...
+                     if (protocolName != null && protocolName.equals(ProtocolName.MQTT_PROTOCOL_NAME.getProtocolName())) {
+                        String clientId = props.getSimpleStringProperty(ManagementHelper.HDR_CLIENT_ID).toString();
```
michaelandrepearce: This is protocol specific logic. It is not suitable to have this in the core modules.Any mqtt logic MUST be in the protocol module only.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -540,6 +541,12 @@ public ServerConsumer createConsumer(final long consumerID,
             props.putSimpleStringProperty(ManagementHelper.HDR_FILTERSTRING, filterString);
          }
 
+         String protocolName = remotingConnection.getProtocolName();
+         if (protocolName != null && protocolName.equals(ProtocolName.MQTT_PROTOCOL_NAME.getProtocolName().toString())) {
```
michaelandrepearce: This is protocol specific logic. This MUST not ever live in core modules. Protocol specific logic can ONLY reside in its protocol module

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ManagementHelper.java
```diff
@@ -84,6 +84,10 @@
 
    public static final SimpleString HDR_MESSAGE_ID = new SimpleString("_AMQ_Message_ID");
 
+   public static final SimpleString HDR_PROTOCOL_Name = new SimpleString("_AMQ_ProtocolName");
```
michaelandrepearce: static final field should be all upper case. HDR_PROTOCOL_NAME
key should be '_' separated _AMQ_Protocol_Name

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/ProtocolName.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.core.protocol;
+
+import org.apache.activemq.artemis.api.core.SimpleString;
+
+public enum ProtocolName {
+
+   MQTT_PROTOCOL_NAME(SimpleString.toSimpleString("MQTT")),
```
michaelandrepearce: -1, Core should not know about specific protocols.

onlyMIT: Got it thanks.  Since the processing of cluster-notification is in the Core, and the Core does not depend on the specific protocol code. I will think about how to adjust the code structure and then submit a version. Maybe need a listener or a callback

onlyMIT:
@michaelandrepearce @franz1981  This PR has updated.  
Because of core modules  should not know about specific protocols,  can't map protocol to byte in CORE module，so protocol name still use SimpleString property in cluster-notification

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
```diff
@@ -62,11 +73,44 @@
...
+
+         Binding binding = server.getPostOffice().getBinding(queueName);
+         if (binding != null) {
+            Queue queue = (Queue) binding.getBindable();
+            String clientId = props.getSimpleStringProperty(ManagementHelper.HDR_CLIENT_ID).toString();
+            //If the client ID represents a client already connected to the server then the server MUST disconnect the existing client.
+            //Avoid consumers with the same client ID in the cluster appearing at different nodes at the same time
+            Collection<Consumer> consumersSet = queue.getConsumers();
+            for (Consumer consumer : consumersSet) {
+               ServerConsumerImpl serverConsumer = (ServerConsumerImpl) consumer;
```
michaelandrepearce: Cast to interface not concrete class.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
```diff
@@ -62,11 +73,44 @@
...
+               if (clientId.equals(serverConsumer.getConnectionClientID()))
+                  serverConsumer.getRemotingConnection().destroy();
```
michaelandrepearce: The new getRemotingConnection isn't needed.
 You should simply be closing the consumer, this will then mean all bits around closing a consumer are safely done, and also other notifications sent.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -1544,4 +1545,8 @@ public String getConnectionLocalAddress() {
    public String getConnectionRemoteAddress() {
       return this.session.getRemotingConnection().getTransportConnection().getRemoteAddress();
    }
+
+   public RemotingConnection getRemotingConnection() {
```
michaelandrepearce: Shouldnt be needed, you should just be acting on the consumer, closing the consumer via its close methods if needed.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -540,6 +540,14 @@ public ServerConsumer createConsumer(final long consumerID,
             props.putSimpleStringProperty(ManagementHelper.HDR_FILTERSTRING, filterString);
          }
 
+         String protocolName = remotingConnection.getProtocolName();
+         if (protocolName != null)
+            props.putSimpleStringProperty(ManagementHelper.HDR_PROTOCOL_NAME, SimpleString.toSimpleString(protocolName));
+
+         String clientId = remotingConnection.getClientID();
+         if (clientId != null)
```
michaelandrepearce: curlys on the if please

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -540,6 +540,14 @@ public ServerConsumer createConsumer(final long consumerID,
             props.putSimpleStringProperty(ManagementHelper.HDR_FILTERSTRING, filterString);
          }
 
+         String protocolName = remotingConnection.getProtocolName();
+         if (protocolName != null)
```
michaelandrepearce: curlys on the if please

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
```diff
@@ -62,11 +73,44 @@
...
+      if (distance > 0) {
+         SimpleString queueName = props.getSimpleStringProperty(ManagementHelper.HDR_ROUTING_NAME);
+
+         Binding binding = server.getPostOffice().getBinding(queueName);
+         if (binding != null) {
+            Queue queue = (Queue) binding.getBindable();
+            String clientId = props.getSimpleStringProperty(ManagementHelper.HDR_CLIENT_ID).toString();
+            //If the client ID represents a client already connected to the server then the server MUST disconnect the existing client.
+            //Avoid consumers with the same client ID in the cluster appearing at different nodes at the same time
+            Collection<Consumer> consumersSet = queue.getConsumers();
```
michaelandrepearce: Might be worth adding a new method on queue that takes an additional predicate parameter, this way can pass in a predicate of 
clientId.equals(serverConsumer.getConnectionClientID()
and return only the consumers matching, this will avoid expensive copying of all consumers if the queue has high consumer count, and repetitive iterating over those.

michaelandrepearce: e.g. adding something like

```
   @Override
   public Set<Consumer> getConsumers(Predicate<Consumer> consumerPredicate) {
      Set<Consumer> consumersSet = new HashSet<>();
      for (ConsumerHolder<? extends Consumer> consumerHolder : consumers) {
         if (consumerPredicate.test(consumerHolder.consumer)) {
            consumersSet.add(consumerHolder.consumer);
         }
      }
      return consumersSet;
   }
```

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
michaelandrepearce: Also should check instanceof, as by contract, it may not be a ServerConsumer

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSession.java
```diff
@@ -30,10 +30,12 @@
 
 public class MQTTSession {
 
-   static Map<String, MQTTSessionState> SESSIONS = new ConcurrentHashMap<>();
+   private static Map<String, MQTTSessionState> SESSIONS = new ConcurrentHashMap<>();
```
michaelandrepearce: make it final too?

michaelandrepearce: Looking at this im bit worried that this is a static in first place, e.g. its possible to have multiple embedded brokers, in that case each broker shouldnt share any state, by being static means they would.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
onlyMIT: I also disagree with putting this method into the `ServerConsumer` interface. Using instanceof is a good suggestion.

michaelandrepearce: what method? if youre talking re remoting connection you shouldnt be even have added that new method, you simply want to close previous consumers, you therefor should be calling close method on the consumer.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
onlyMIT: Our goal is to avoid simultaneous occurrence of multiple MQTT connections with the same clientId, not to prevent MQTT consumers with the same clientId to subscribe the same topic.

described in the [two solutions](https://github.com/apache/activemq-artemis/pull/2528#issue-248449219) .  Adding a new cluster notification when the client’s connections connect to Artemis,
it is the best way to address the MQTT agreement, but the cost is too great. So it put the logic in CREATE_CONSUMER cluster-notification.

The following code description, before each processing of the mqtt connection request in artemis, check whether there is a connection with the same clientId in the node, and if so, close it.
Similarly, in the cluster, we also want to avoid the connection of the same clientId. Executing this logic only in CREATE_CONSUMER is just that I think it can reduce resource consumption and avoid problems, although this is not the way to match the MQTT specification.
```
private String validateClientId(String clientId, boolean cleanSession) {
      if (clientId == null || clientId.isEmpty()) {
         // [MQTT-3.1.3-7] [MQTT-3.1.3-6] If client does not specify a client ID and clean session is set to 1 create it.
         if (cleanSession) {
            clientId = UUID.randomUUID().toString();
         } else {
            // [MQTT-3.1.3-8] Return ID rejected and disconnect if clean session = false and client id is null
            return null;
         }
      } else {
         MQTTConnection connection = session.getProtocolManager().addConnectedClient(clientId, session.getConnection());

         if (connection != null) {
            // [MQTT-3.1.4-2] If the client ID represents a client already connected to the server then the server MUST disconnect the existing client
            connection.disconnect(false);
         }
      }
      return clientId;
   }
```

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
onlyMIT: got it

michaelandrepearce:
@onlyMIT to help with addressing the comments i sent a pr to your branch with a number of proposed changes

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSession.java
michaelandrepearce: i sent a possible solution to address this, in a pr to your branch.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
michaelandrepearce: i sent this in pr to your branch

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
michaelandrepearce: But your current logic is simply find all consumers with matching clientId and kill beneath it, their connection. You may as well simply close down the consumer

michaelandrepearce: Why is:

Adding a new cluster notification when the client’s connections connect to Artemis,
it is the best way to address the MQTT agreement, but the cost is too great

Whats the cost? To me if you want the behaviour on connection behaviour this truely would be better, rather than working on consumer

michaelandrepearce: we already have a connection notification, just need to add the generic extra data as you were going to on the consumer

   private void sendConnectionNotification(final RemotingConnection connection, final CoreNotificationType type) {
      final ManagementService managementService = getManagementService();

      if (managementService != null && sendConnectionNotifications) {
         try {
            final TypedProperties props = new TypedProperties();
            props.putSimpleStringProperty(ManagementHelper.HDR_CONNECTION_NAME, SimpleString.toSimpleString(connection.getID().toString()));
            props.putSimpleStringProperty(ManagementHelper.HDR_REMOTE_ADDRESS, SimpleString.toSimpleString(connection.getRemoteAddress()));
            props.putSimpleStringProperty(ManagementHelper.HDR_CLIENT_ID, SimpleString.toSimpleString(connection.getClientID()));
            props.putSimpleStringProperty(ManagementHelper.HDR_PROTOCOL_NAME, SimpleString.toSimpleString(connection.getProtocolName()));

            managementService.sendNotification(new Notification(null, type, props));
         } catch (Exception e) {
            logger.warn("Error sending notification: " + type, e.getMessage(), e);
         }
      }
   }

and then act just at the connection level.

onlyMIT: This code seems to be easily misunderstood, I am trying to find out if there is a better way to handle this logic.

onlyMIT: Thanks for your answer，i not find the connection notification，if we had we can use it.

onlyMIT:
@michaelandrepearce thanks to your review, I will address  the comments.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
michaelandrepearce: @onlyMIT but re your top level options:

> Add a new cluster notification when the client’s connections connect to Artemis, close existing connections that use the same clientId when processing cluster-notification. This solution is the best way to address the MQTT agreement.But I don't think it's necessary to solve this problem by sending a cluster-notification every time client connect.

There already is a cluster-notification sent every time a client connects if using NotificationActiveMQServerPlugin, so if anything simply we could just make that more built in, like consumer one is, and simply make it enable the  flag sendConnectionNotifications, if MQTT protocol is used (keeping it as is today disabling for those who have not enabled MQTT)

onlyMIT: got it thanks

michaelandrepearce: Looking some more when an connection with MQTT is created the MQTTProtocolHandler is called, so actually may not even need notification could just hook into that logic and flow

michaelandrepearce: e.g.

from
```
@Override
   public ConnectionEntry createConnectionEntry(Acceptor acceptorUsed, Connection connection) {
      try {
         MQTTConnection mqttConnection = new MQTTConnection(connection);
         ConnectionEntry entry = new ConnectionEntry(mqttConnection, null, System.currentTimeMillis(), MQTTUtil.DEFAULT_KEEP_ALIVE_FREQUENCY);

         NettyServerConnection nettyConnection = ((NettyServerConnection) connection);
         MQTTProtocolHandler protocolHandler = nettyConnection.getChannel().pipeline().get(MQTTProtocolHandler.class);
         protocolHandler.setConnection(mqttConnection, entry);

         return entry;
      } catch (Exception e) {
         log.error(e);
         return null;
      }
   }
```

to something like: (replacing the code added with the logic you think is needed)
```
@Override
   public ConnectionEntry createConnectionEntry(Acceptor acceptorUsed, Connection connection) {
      try {
         MQTTConnection mqttConnection = new MQTTConnection(connection);
         ConnectionEntry entry = new ConnectionEntry(mqttConnection, null, System.currentTimeMillis(), MQTTUtil.DEFAULT_KEEP_ALIVE_FREQUENCY);

         NettyServerConnection nettyConnection = ((NettyServerConnection) connection);
         MQTTProtocolHandler protocolHandler = nettyConnection.getChannel().pipeline().get(MQTTProtocolHandler.class);
         protocolHandler.setConnection(mqttConnection, entry);

         String clientID = connection.getProtocolConnection().getClientID();
         MQTTConnection existingMQTTConnection = connectedClients.get(clientID);
         if (existingMQTTConnection != null) {
            existingMQTTConnection.getTransportConnection().getProtocolConnection().destroy();
         }
         return entry;
      } catch (Exception e) {
         log.error(e);
         return null;
      }
   }
```

onlyMIT: My goal is to close connections that have the same clientId in different nodes.The following code already done the logic which your mentioned . 
```
private String validateClientId(String clientId, boolean cleanSession) {
      if (clientId == null || clientId.isEmpty()) {
         // [MQTT-3.1.3-7] [MQTT-3.1.3-6] If client does not specify a client ID and clean session is set to 1 create it.
         if (cleanSession) {
            clientId = UUID.randomUUID().toString();
         } else {
            // [MQTT-3.1.3-8] Return ID rejected and disconnect if clean session = false and client id is null
            return null;
         }
      } else {
         MQTTConnection connection = session.getProtocolManager().addConnectedClient(clientId, session.getConnection());

         if (connection != null) {
            // [MQTT-3.1.4-2] If the client ID represents a client already connected to the server then the server MUST disconnect the existing client
            connection.disconnect(false);
         }
      }
      return clientId;
   }
```
I think I am almost revised, I will update after the test is completed.

onlyMIT: And clientId is null is this place:
```
 @Override
   public ConnectionEntry createConnectionEntry(Acceptor acceptorUsed, Connection connection) {
      try {
         //New code, to kill of existingConnection if same client id.
         String clientID = connection.getProtocolConnection().getClientID();
         MQTTConnection existingMQTTConnection = connectedClients.get(clientID); 
```

clientId is set after the `createConnectionEntry `method is called.  In `MQTTConnectionManager.connect method`
```
   synchronized void connect(String cId,
                             String username,
                             byte[] passwordInBytes,
                             boolean will,
                             byte[] willMessage,
                             String willTopic,
                             boolean willRetain,
                             int willQosLevel,
                             boolean cleanSession) throws Exception {
      String clientId = validateClientId(cId, cleanSession);
      if (clientId == null) {
         session.getProtocolHandler().sendConnack(MqttConnectReturnCode.CONNECTION_REFUSED_IDENTIFIER_REJECTED);
         session.getProtocolHandler().disconnect(true);
         return;
      }

      String password = passwordInBytes == null ? null : new String(passwordInBytes, CharsetUtil.UTF_8);
      session.getConnection().setClientID(clientId);
```

onlyMIT:
@michaelandrepearce  This PR has updated.  `SESSIONS` is still in MQTTSession,  do not move to  `MQTTProtocolHandler ` . Because SESSIONS should be shared by multiple Acceptors. 
```
public class MQTTSession {

    static Map<String, MQTTSessionState> SESSIONS = new ConcurrentHashMap<>();
   private static Map<String, MQTTSessionState> SESSIONS = new ConcurrentHashMap<>();
```

michaelandrepearce:
There is the factory if thats the case where a map could be passed into each handler during favtory crreate. What we cant have though is statics for this tho

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/server/management/NotificationService.java
```diff
@@ -37,4 +37,8 @@
 
    void removeNotificationListener(NotificationListener listener);
 
+   default void callNotificationListener(Notification notification) {
```
michaelandrepearce: This is purpose of sendNotification, any new thing should integrate with the existing way things work.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
```diff
@@ -633,6 +633,20 @@ public boolean isStarted() {
       return started;
    }
 
+   @Override
+   public void callNotificationListener(Notification notification) {
```
michaelandrepearce: This is repeating logic of sendNotification. Please just use that.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
```diff
@@ -163,13 +163,7 @@ synchronized void disconnect(boolean failure) {
...
    private MQTTSessionState getSessionState(String clientId) {
       /* [MQTT-3.1.2-4] Attach an existing session if one exists otherwise create a new one. */
-      MQTTSessionState state = MQTTSession.SESSIONS.get(clientId);
-      if (state == null) {
-         state = new MQTTSessionState(clientId);
-         MQTTSession.SESSIONS.put(clientId, state);
-      }
-
-      return state;
+      return MQTTSession.getSessionState(clientId, session.getServer().getIdentity());
```
michaelandrepearce: -1 Ive raised this already, simply shouldnt be having static map holding state.

onlyMIT:
@michaelandrepearce I have addressed the comments your sent a pr to my branch. Change the clusterConsumer queue’s filter. Because it will filter CONNECTION_CONNECTED notification.

artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/protocol/RemotingConnection.java
```diff
@@ -245,6 +245,13 @@
     */
    String getTransportLocalAddress();
 
+   /**
+    * Called when all connection related operations are completed.
+    */
+   default void connected() throws Exception {
```
michaelandrepearce: As this introduces a notification about connections created, should ensure that this is always called, and not dependent on a protocol logic.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnection.java
```diff
@@ -216,8 +220,10 @@ public void flush() {
    public void bufferReceived(Object connectionID, ActiveMQBuffer buffer) {
    }
 
-   public void setConnected(boolean connected) {
-      this.connected = connected;
+   @Override
+   public void connected() throws Exception {
+      remotingService.sendConnectedNotification(this);
```
michaelandrepearce: This should be responsibility of core system to sending notification so connected notification is generic.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/CoreNotificationType.java
```diff
@@ -47,7 +47,8 @@
    SESSION_CREATED(26),
    SESSION_CLOSED(27),
    MESSAGE_DELIVERED(28),
-   MESSAGE_EXPIRED(29);
+   MESSAGE_EXPIRED(29),
+   CONNECTION_CONNECTED(30);
```
michaelandrepearce: because this is in core, this must be implemented in a generic way for ALL protocols.

michaelandrepearce: Going through and running this, and running with notification plugin, i notice every time this occurs, a SESSION_CREATED is also sent. 

digging deeper and looking going through the logic of the connect in MQTTConnectionManager it seems this actually connected in the MQTTConnectionManager maps over to CORE as a server session (e.g. connect == create a serversession), which means we already get a SESSION_CREATED notification with that plugin, so is it actually we want to ensure SESSION_CREATED notificaions are not reliant on plugin, e.g. bring that in... 

populating.
session.getRemotingConnection().getClientID()

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnection.java
michaelandrepearce: see other comments, ive dug deeper on this and seems actually the act of mqtt connected == core server session created.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/CoreNotificationType.java
michaelandrepearce: So this ties in with the fact you want to ensure a single connected session (its not actually at connection level, its actually at the session level).

onlyMIT: I am here to make a unified reply to your comments.

In fact, this is only the mqtt protocol need, other protocols do not.Either way, it doesn't look too good in CORE, it just tries to make it look reasonable and more readable.
```
   /**
    * Called when all connection related operations are completed.
    */
   default void connected() throws Exception {
```
I use `connected` as default not a interface method，Because we need to send cluster-notifation when the mqtt protocol connection is initialized, But other protocols don't need this method to do anything, use a empty default methods I think is appropriate.And all protocol connections have this connection completion status, and other protocols have the possibility to use this method. As for the logic that is only needed for MQTT to add a listener or callback, I think it is over-designed.

I also find SESSION_CREATED  and CONNECTION_CREATED, but i can not use them, because we need clientId, and clientId read from client message. The clientId is read from the client's connection message. The logic for reading the clientId cannot be placed in the CORE code， when create a connection or create a session.

For using callbacks or listeners, I still prefer to use the default method connected. Unless you have a better design to let me change my mind. 
the clientId is read from the client message, it seems that the message can only be sent after the protocol connection has been processed. ( MQTT protocol neet to process the CONNECT request packet, some protocols do not have a CONNECT request packet)

michaelandrepearce: re

"The clientId is read from the client's connection message. The logic for reading the clientId cannot be placed in the CORE code， when create a connection or create a session."

the clientid is set on the connection BEFORE createServerSession as such it will be available when the SESSION_CREATED notification is created, 

MQTTConnectionManager lines 78, 79
```
 session.getConnection().setClientID(clientId);
 ServerSessionImpl serverSession = createServerSession(username, password);
```
just need to populate the extra field which would be available at the time via session.getRemotingConnection().getClientID() .

onlyMIT:
```
if (hasBrokerSessionPlugins()) {
    callBrokerSessionPlugins(plugin -> plugin.afterCreateSession(session));
}
```
SESSION_CREATED will only be sent if BrokerSessionPlugins is turned on in the configuration file.
Also the notifications of CONNECTION_CREATE and SESSIONTION_CREATE will be filtered and will not be sent to other nodes in the cluster.You can see `ClusterConnectionBridge.setupNotificationConsumer()` method, where the cluster-connection filter is set .
This is why I want to modify the filtering rules, CONNECTION_CREATE, SESSIONTION_CREATE and other connection-related notifications, only can use the `ManagementNotificationAddress` address to route messages.

onlyMIT:
```
private void handleNotificationMessage(ClientMessage message, CoreNotificationType ntype) throws Exception {
         // TODO - optimised this by just passing int in header - but filter needs to be extended to support IN with

         switch (ntype) {
            case BINDING_ADDED: {
               doBindingAdded(message);

               break;
            }
            case BINDING_REMOVED: {
               doBindingRemoved(message);

               break;
            }
            case CONSUMER_CREATED: {
               doConsumerCreated(message);

               break;
            }
            case CONSUMER_CLOSED: {
               doConsumerClosed(message);

               break;
            }
            case PROPOSAL: {
               doProposalReceived(message);

               break;
            }
            case PROPOSAL_RESPONSE: {
               doProposalResponseReceived(message);
               break;
            }
            case UNPROPOSAL: {
               doUnProposalReceived(message);
               break;
            }
            default: {
               throw ActiveMQMessageBundle.BUNDLE.invalidType(ntype);
            }
         }
      }
```

Only these notifications will pass between cluster nodes

michaelandrepearce:
As ive already said you have two options either make mqtt to ensure plugin js enabled or simply make this notification more perm as like consumer one is.

onlyMIT:
@michaelandrepearce As described in my comment.Even if use a plugin, we can't use notifications in the plugin like SESSION_CREATED，it can't achieve our goal, modify the SESSION_CREATED filter? I think we need to think carefully, and if we modify their filtering rules, we can't affect the existing logic. We need to conduct comprehensive testing and evaluation.I think the plugin should be where the optional code is placed, and the logic that is must required is not suitable for the plugin.

Like the first version I submitted, the logic was placed in CONSUMER_CREATE, the readability was too poor and not exactly matched with the MQTT specification; 
This version the logic was added by adding CONNECTED_CREATED. It seems that not all protocols are used, and it is not the most Good plan.
 If you want to use SESSION_CREATED, CONNECTION_CREATED,  must force the MQTT protocol to open all plugins? and modify the filter so that the SESSION_CREATED notification can communicate between the clusters. This is the risk of making the original notification rules undefined，and need to modify their filtering rules, I am not sure if modifying the filtering rules will introduce other issues.

onlyMIT:
Now let's solve it step by step. First, we need to determine whether to use the existing notification or add a new notification to handle this logic. If I comment, it seems that the existing notification is not suitable.
In the second step, if we use the new notification, we need to determine how to make it more reasonable.

michaelandrepearce:
First of all mqtt connect = create session in core. 

I dont want to introduce anything thats protocol specific. 

Im happy with current create session being a top level supported notification as this is agnostic and will cause little extra traffic compared with create consumer which already  is.

michaelandrepearce:
I am off now for 48 hours (pto).

michaelandrepearce:
As noted i sent an unpolished pr to your branch last week, to enhance and use SESSION_CREATED

As noted connected == create session in core, so im against constructing some new terminology within core notifications, here.

onlyMIT:
@michaelandrepearce I was on vacation last week,  started this week.

onlyMIT:
@michaelandrepearce move logic to SESSION_CREATE notification

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -2954,7 +2955,13 @@ public void onNotification(org.apache.activemq.artemis.core.server.management.No
       if (!(notification.getType() instanceof CoreNotificationType))
          return;
       CoreNotificationType type = (CoreNotificationType) notification.getType();
-      TypedProperties prop = notification.getProperties();
+      if (type == CoreNotificationType.SESSION_CREATED) {
+         TypedProperties props = notification.getProperties();
+         //SESSION_CREATED notification should not be broadcast twice
+         if (props.getIntProperty(ManagementHelper.HDR_DISTANCE) > 0) {
```
michaelandrepearce: Why would this logic specific to just SESSION_CREATED, would this not be generic to all?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionBridge.java
```diff
@@ -351,10 +355,15 @@ private String appendIgnoresToFilter(String filterString) {
       }
       filterString += "!" + storeAndForwardPrefix;
       filterString += ",!" + managementAddress;
-      filterString += ",!" + managementNotificationAddress;
       return filterString;
    }
 
+   private String createPermissiveManagementNotificationToFilter() {
+      StringBuilder filterBuilder = new StringBuilder(ManagementHelper.HDR_NOTIFICATION_TYPE).append(" = '")
```
michaelandrepearce: Whats this trying to achieve?

michaelandrepearce:
@onlyMIT looking better, i have some questions, which im sure all makes sense if explained, but probably a sign a little java doc might be beneficial, and would help future understanding when this PR is closed.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
michaelandrepearce: Also isnt cyclic between nodes in cluster already avoided, because of filter in the cluster connection bridge where it does a HDR_DISTANCE < flow.maxHops.

onlyMIT: To avoid calling the broadcast twice.The original logical SESSION_CREATED notification will not be transmitted between cluster nodes, and it is now required to transfer between cluster nodes. In order to maintain consistency with the original logic, this code to prevent the node receiving the SESSION_CREATED notification from making a broadcast call again.
Keep the original logic for other notification messages not judged

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionBridge.java
onlyMIT: The inter-cluster notification message needs to specify the routing address to route the message. Different from the notifications of CONSUMER_CREATE and BINDING_ADD, the routing address (the address created or consumed by itself) has been determined before the notification is sent. When sending a SESSION_CREATED notification, We can't know which address to use, so I chose to use the `ManagementManagementAddress`, which is consumed by each node in the cluster-connection.
But the notification message using the `ManagementNotificationAddress` address is filtered by the filter, so I modified the filter and allow SESSION_CREATED notification to be routed to other nodes in the cluster.

onlyMIT: Although it is possible to add a dedicated address to the SESSION_CREATED notification when the cluster-connection is created, it is not necessary to do so.

onlyMIT:
Add java doc

michaelandrepearce:
@onlyMIT i would like some of others in the group to review this as well.

onlyMIT:
@michaelandrepearce Ok, thank you very much for your help.

michaelandrepearce:
@jbertram @clebertsuconic and anyone else. As no one else is commenting /reviewing, and to me this looks ok, Ill look to merge this in a few days if no further review comment from anyone.

clebertsuconic:
@michaelandrepearce the commit message should have been edited before merged.

no big deal.. just mentioning it.

PR #2529, opened 2019-01-30
Refactor the native library build code

orpiske:
This refactoring changes the native library build code to allow cross-compilation and brings the following changes:

- Fix the link target to the one that was actually found by find_library

- Replace add definitions with standard CMake C flags handling options. Among
other things, this allow greater control over which compiler flags are used
and makes it easier to turn on/off debug flags by manipulating the
CMAKE_USER_C_FLAGS variable as documented in the README

- Updates the CMake scripts to allow cross-compilation of the artemis
native library into x86 32-bit code when building on a x86 64-bit system

- Updates the library documentation

- Defines C99 as the C standard in use so that the native library code can be
compiled in older GCC versions

- Renames the readme file so it can be better visualized on the web

These are loosely related to the changes I did on #2523 to help @franz1981 cross-compile the code.


artemis-native/README.md
```diff
@@ -0,0 +1,87 @@
...
+```cmake -DCMAKE_USER_C_FLAGS="-fomit-frame-pointer" -DCMAKE_VERBOSE_MAKEFILE=On .```
+
+Compiling with debug options:
+```cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_VERBOSE_MAKEFILE=On .```
+
+Cross-compilation:
+```cmake -DCMAKE_VERBOSE_MAKEFILE=On -DCMAKE_USER_C_FLAGS="-m32" -DARTEMIS_CROSS_COMPILE=On -DARTEMIS_CROSS_COMPILE_ROOT_PATH=/usr/lib .```
+
+Cross-compilation with debugging symbols:
+```cmake -DCMAKE_VERBOSE_MAKEFILE=On -DCMAKE_USER_C_FLAGS="-m32" -DARTEMIS_CROSS_COMPILE=On -DARTEMIS_CROSS_COMPILE_ROOT_PATH=/usr/lib .```
```
clebertsuconic: We will have to make this manually every time we compile?

clebertsuconic: Do we need to install any libraries to allow 32 bits compilation? any new dependency?

orpiske: You only need the 32-bit version of the same libraries as you need for 64-bit. I added some details on the readme, but the summary is that on a 64-bit Fedora/RHEL you only need to install: glibc-devel.i686 and libaio-devel.i686.

orpiske: > We will have to make this manually every time we compile?

You can still compile it with the `compile-native.sh` script. The changes retain the current behaviour, but allow you to cross-compile if you decide to do so.

clebertsuconic: I just proposed we make this a separate github repository. Look at my DISCUSS thread on the activeMQ dev list.

orpiske: Sure. I will share some of my thoughts there.

michaelandrepearce: :( i dont see any thread

clebertsuconic: I sent an email to dev@activemq...

"[DISCUSS] ActiveMQ Artemis Native as a separated project"

michaelandrepearce: Email system sent it to junk. Have found it. I think this is a good idea. And actually should do it more and foster the idea more with extensions etc.

PR #2530, opened 2019-01-30
Native library fixes

orpiske:
These are small fixes to the native library that:

1) prevent a NULL pointer dereference. Although rare, it is still possible that a call to malloc fails and returns null. In that case, it could potentially bring down the whole process. It replaces the first malloc call on exceptionMessage with a call to asprintf which performs the allocation of a string large enough to hold both the err and result string plus the NULL pointer. 

2) Handles failure to initialize the events structure

3) Fix a cleanup error caused by trying to deallocate stack memory with free

4) Ensure proper memory cleanup of IOCB structure on error


orpiske:
Also, when testing the change on the events structure, I think it unveils another issue. Since libaioContext is allocated on the stack and initialized by io_queue_init, calling free on it does not seem to be the correct way to perform the cleanup. I believe we should be calling io_queue_release instead.

orpiske:
As a matter of fact, io_queue_release [does seem to be](https://github.com/linux-test-project/ltp/blob/058b2e55825344ab9da0567f94ef535749041ac9/testcases/kernel/io/aio/aio01/aio01.c) how the kernel tests cleanup the context.

PR #2533, opened 2019-01-31
ARTEMIS-2244 checkDepage method placed outside CRITICAL_DELIVER avoid critical analyzer timeout

CNNJYB:
We found server crash becauseof critical analyzer timeout, thread dump can be seen in attachment.	
Suppose that there is a topic t with two subscriber ta and tb. Some messages were routed to subscriber ta, not to tb. When a consumer that subscribe tb is created and send ConsumerCredits to server, ServerConsumerImpl will call promptDelivery method, then forceDelivery() -> messageQueue.deliverAsync() -> deliverRunner will executed in the pageSubscription's executor(step1), then checkDepage(step2) -> pageIterator.hasNext(synchronized method) -> next -> moveNext, moveNext method will iterate through queue until find a matching message. At this time(step1), DeliverRunner call deliver method -> checkDepage -> pageIterator.hasNext, this step blocked on CursorIterator which was locked by step2, then critical analyzer timeout.
thread dump can be seen in attachment.
[threaddump.txt](https://github.com/apache/activemq-artemis/files/2820006/threaddump.txt)

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -2409,7 +2409,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
     * This method will deliver as many messages as possible until all consumers are busy or there
     * are no more matching or available messages.
     */
-   private void deliver() {
+   private boolean deliver() {
       if (logger.isDebugEnabled()) {
          logger.debug(this + " doing deliver. messageReferences=" + messageReferences.size());
       }
@@ -2430,7 +2430,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
             deliverAsync();
 
-            return;
+            return false;
          }
 
          if (System.currentTimeMillis() > timeout) {
@@ -2440,7 +2440,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
             deliverAsync();
 
-            return;
+            return false;
          }
 
          MessageReference ref;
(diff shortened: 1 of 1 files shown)


michaelandrepearce:
-1 critical analyzer was implemented specifically to pick up disk and other issues, that are on critical path but not performant.

clebertsuconic:
do you have a thread dump for when the issue happened?

clebertsuconic:
Just saw it...


it seems that the Critical Analyzer did its job and kicked the server of after a real dead lock? and which you are fixing it here?

clebertsuconic:
This is actually nice!!!! I thought I wrote the Critical Analyzer in vain and it seems the Critical Analyzer kicked the server after a real dead lock.. !!!

clebertsuconic:
@CNNJYB what version did you use on this thread dump? 2.6.3 ?

I'm not sure the real issue is addressed here???

clebertsuconic:
sorry for my monolog here.. I see the issue... on your case you had getNext moving the cursor while checking... 


The fix is valid..

Although I would prefer either keep the Critical Analyzer call around the checkDepage...

or create a new CriticalAnalyzer check. I will do the later.

michaelandrepearce:
To me it seems we shouldnt have change critical analyslzer at all. Its done its job. Its detected on a critical path the is something blocking and it shouldnt be. The real fix is to fix the block, not remove the safety check

clebertsuconic:
I have amended the change before merging (extra commit)

Basically the operation was moved outside of a lock. 

And I added a new critical analyzer path to still have it on the new path created.

wy96f:
> To me it seems we shouldnt have change critical analyslzer at all. Its done its job. Its detected on a critical path the is something blocking and it shouldnt be. The real fix is to fix the block, not remove the safety check.
> 
> It was added for this exact purpose!
Agreed. To fix the blocking problem, could we add a flag indicating checkDepage() is in progress? If in process, we could return directly, no need to block in iterator.hasNext() which potentially causes triggering critical analyzing.

PR #2534, opened 2019-01-31
ARTEMIS-1058 Jars in web tmp dir locked on Windows

gaohoward:
This is the second commit. It improves the windows service
configuration file so that the service can be shutdown
gracefully (clean shutdown) and allow for a chance to
clean up the web tmp dir.

### DIFF
--- a/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service.xml
+++ b/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service.xml
@@ -33,39 +33,58 @@
    <logmode>roll</logmode>
 
    <executable>%JAVA_HOME%\bin\java.exe</executable>
-   <argument>-Xbootclasspath/a:%ARTEMIS_HOME%\lib\${logmanager}</argument>
-   <argument>-XX:+UseParallelGC</argument>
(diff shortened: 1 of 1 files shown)


clebertsuconic:
The problem is only on windows service?

On this case, we could revert ae00423034b6905564f6729f681dc8e29c86422e after this commit?

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service.xml
```diff
@@ -33,39 +33,58 @@
    <logmode>roll</logmode>
 
    <executable>%JAVA_HOME%\bin\java.exe</executable>
```
clebertsuconic: don't you need to change this to startexecutable?

gaohoward:
@clebertsuconic  The commit ae00423 was valid. However on Windows when running artemis as a service,  stopping the service simply kills the process, So the shutdown hook won't get called and the clean up code never get executed. This commit just fixes this. 
According to this doc: https://github.com/kohsuke/winsw/blob/master/doc/xmlConfigFile.md
The <executable> is used for starting a service, not <startexecutable>.

clebertsuconic:
@gaohoward Please check #2537

PR #2536, opened 2019-02-04
ARTEMIS-2245 Implement Docker images

franz1981:
It provides both CentOS and Ubuntu Docker images for the Artemis broker.


clebertsuconic:
this looks good, but I think we should add docs.

franz1981:
I will sent a doc commit today :+1: thanks for the review mate!

orpiske:
This looks very cool @franz1981! 

A few suggestions:
1. I'd probably arrange the ARTEMIS_VERSION so that it can be both a variable and an argument to the dockerfile. That way, you are not forced to change the Dockerfile for every release. 

```
ARG ARTEMIS_VERSION
ENV ARTEMIS_VERSION ${ARTEMIS_VERSION:-2.7.0-SNAPSHOT}
LABEL ARTEMIS_VERSION=${ARTEMIS_VERSION}
```

2. I think it would be good to label the Centos and Ubuntu images accordingly. 
```LABEL ARTEMIS_BASE_SYSTEM=ubuntu```

Another thing you could do is consolidate the 2 files in one, so you could possibly reuse the container build code post the part that you setup the initial system. But I consider that a matter of personal taste and nothing that could give immediate gains to the image IMHO.

**Comment by clebertsuconic** [comment]
What if we moved the whole docker under the binary distribution? it would be a lot cleaner instead of requiring a new distribution just for docker.

That way users would just be able to turn their binary into docker at any point.


I just sent a PR to your branch: https://github.com/franz1981/activemq-artemis/pull/11

**Comment by orpiske** [comment]
> Do we need that label? Can't we just remove it?

Yes, you could remove it altogether. There's no harm in having it there, though. One benefit, IMHO, is that if you are pulling the images by a different tag other than the version (ie.: ```docker pull apache/artemis:latest```), it makes it very easy to find the actual version that maps to 'latest' by just inspecting the container (with ```docker inspect```).

**Comment by clebertsuconic** [comment]
on a second though.. having the docker as part of the binary distribution is not such a good idea.


I have updated the PR on your Branch Franz.

**Comment by michaelandrepearce** [comment]
I quite like the idea of a ready to run docker image that we distribute. Much more than an example

**Comment by franz1981** [comment]
Let me take some time to think about the best approach here: I see that the current PR is very dev-friendly (given that a compiled distribution IS the base to create a running docker image), but I would prefer something more user-friendly and self contained eg that download or receive from an argument the URL/zip to be used to build the image.
That would allow the docker file to be self-contained and somehow indipendent from the compiled code...

**Comment by michaelandrepearce** [comment]
For me the ultimate would be an image in docker hub, just as we push java artifacts to maven central. Its about making things as easy to consume as possible

**Comment by clebertsuconic** [comment]
That part is a bit more complex as Apache infra is still in early days for this. 

What we are trying to do here is to provide the docker files so this could be built anywhere else.

**Comment by clebertsuconic** [comment]
Even if we decided to deploy it as part of apache account, we still need the docker files.. which is this step here.

**Comment by michaelandrepearce** [comment]
Another option then maybe like what occurs for some other apache projects (e.g. tomcat and httpd, 
), is host it under docker-library and maintain it there, as "official"

https://github.com/docker-library/tomcat
https://github.com/docker-library/httpd

**Comment by michaelandrepearce** [comment]
Another option is to have a dedicated submodule specifically for kubernetes, e.g. like Apache Ignite.

https://github.com/apache/ignite/tree/master/modules/kubernetes

**Comment by michaelandrepearce** [comment]
For me, docker/kubernetes support is becoming quite a defacto, so think it warrants being a bit more top level thing for us to have, than just being some example. 

Just to iterate I think what @franz1981 has done is really great.

File: artemis-docker/Dockerfile-centos [hunk 1]
```diff
@@ -0,0 +1,71 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# ActiveMQ Artemis
+
+FROM jboss/base-jdk:8
```
michaelandrepearce: Any chance for an alpine one too? pretty pretty please.

https://github.com/docker-library/openjdk/tree/922344e3283ef2c87cb17cc04ded9cef5e13a1e5/8/jdk/alpine

cecilphillip:
@michaelandrepearce looks like there an image available here https://github.com/vromero/activemq-artemis-docker

PR #2538, opened 2019-02-04
NO-JIRA fix race condition in QueueQueryTest

jbertram:
My previous commit to fix this was amended which nullified the fix.


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/QueueQueryTest.java
```diff
@@ -161,14 +161,14 @@ public void testQueueQueryOnAutoCreatedQueueWithFQQN() throws Exception {
          QueueQueryResult queueQueryResult = server.queueQuery(fqqn);
          assertEquals(queueName, queueQueryResult.getName());
          assertEquals(addressName, queueQueryResult.getAddress());
-         Wait.assertEquals(1, server.queueQuery(fqqn)::getMessageCount);
+         Wait.assertEquals(1, () -> server.queueQuery(fqqn).getMessageCount());
```
clebertsuconic: Better syntax:

Wait.assertEquals(1,  server.queueQuery(fqqn)::getMessageCount);

PR #2539, opened 2019-02-06
ARTEMIS-2248 don't create sslEngine w/sniHost in NettyConnector

jbertram:


jbertram:
@roddiekieley please review

PR #2541, opened 2019-02-06
ARTEMIS-2246 clarify docs for defaults; change max-disk-usage default

jbertram:


artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -749,10 +749,10 @@
             </xsd:annotation>
          </xsd:element>
 
-         <xsd:element name="max-disk-usage" type="xsd:int" default="90" maxOccurs="1" minOccurs="0">
+         <xsd:element name="max-disk-usage" type="xsd:int" default="100" maxOccurs="1" minOccurs="0">
```
clebertsuconic: I understand 90 may be too low for developers.. but isn't 100 too high?

We should perhaps use it at 99%? Some systems may become non responsive at 100% disk utilization.

clebertsuconic: Why you are disabling by default anyway? I would expect the system to block on a disk full, before it got full.

I think there's a good value on having the system protected against disk full, out of box. We could increase the default to 99 and users can still disable if they would rather take the risk.

michaelandrepearce: Agreed i think 90% was a sensible default.

jbertram: I think you guys should read the [JIRA](https://issues.apache.org/jira/browse/ARTEMIS-2246). Changing the XSD does *not* change the actual default in the code. The default in the code is _already_ `100`. This is also noted in the documentation. The XSD change is just to avoid confusion between the schema and the docs. Lastly, the out-of-the-box broker.xml sets `max-disk-usage` to `90`.

clebertsuconic: My issue is changing the out of box broker.xml.  We should keep it as is.  If increasing to 95 may be ok.  

I’m -1 on disabling it.

jbertram: I'm not sure what you mean, @clebertsuconic. The out of the box broker.xml is *not* being changed. Nothing is being disabled. As stated previously, the schema change is just to avoid confusion between the schema and the docs. Changing the schema does not change the out-of-the-box broker.xml.

michaelandrepearce: Maybe be better to change doc and code default to that in the schema? Having a max at 100 once disk is full its normally to late. I think its better we have safe defaults

clebertsuconic: I agree... lets move the default to 90.

@jbertram WDYT?

clebertsuconic: although.. I just remember one issue... my laptop wouldn't be able to run tests when I had my family photos installed. :)

Devs use a lot of disk.. sometimes close to max.

michaelandrepearce: Dev vs Prod issue. Prod wins for me....

jbertram: Done.

clebertsuconic: I will merge it..but I will add a PR to have a system property to change the value to -1, and have the maven settings setting it to 100.

Some of my CI boxes are close to 90... so this would break things as well.

jbertram:
This was replaced by #2544.

PR #2542, opened 2019-02-07
ARTEMIS-2081 listConfiguredQueues returns only queues created by config

michaelandrepearce:
Move configuration managed in persister before consumer dispatch so then the fix can be applied in 2.6.x branch safely.


michaelandrepearce:
@jbertram this is to move this change in 2.7.0 to before new consumer dispatch, so then it would be safe to apply the original pr into 2.6.x

PR #2543, opened 2019-02-07
ARTEMIS-2081 listConfiguredQueues returns only queues created by config

michaelandrepearce:
Extend test case to reproduce problem of client created queues being incorrectly removed on simple reload of config.
Add a flag/field to the queues created by configuration/broker.xml so we can correctly filter only queues created/managed by config.
Update listConfiguredQueues to use the new queue flag

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/QueueControl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/QueueControl.java
@@ -225,6 +225,12 @@ public interface QueueControl {
    /**
     *
     */
+   @Attribute(desc = "is this queue managed by configuration (broker.xml)")
+   boolean isConfigurationManaged();
+
+   /**
+    *
+    */
    @Attribute(desc = "If the queue should route exclusively to one consumer")
    boolean isExclusive();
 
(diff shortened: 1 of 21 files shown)


michaelandrepearce:
@jbertram ended up having to handle a number of conflicts when cherry picking, but think it's sorted.

clebertsuconic:
wouldn't be best this on 2.7.0 only?

jbertram:
Even though it's adding data to the queue encoding I see this as bug fix for the feature added in [ARTEMIS-1235](https://issues.apache.org/jira/browse/ARTEMIS-1235). That feature is not working as designed (i.e. removing queues it shouldn't) which IMO should be fixed.

michaelandrepearce:
@jbertram @clebertsuconic agreed this is a bug fix, i meant to port this to 2.6.x back last year but went off radar.

FYI rat failure exists in 2.6.x already seems one of the recent merges introduced it. https://travis-ci.org/apache/activemq-artemis/builds/489811029?utm_source=github_status&utm_medium=notification

michaelandrepearce:
Seems like commit on 2.6.x branch:
https://github.com/apache/activemq-artemis/commit/3ee6699a0ea576975b78f0248fe061d26296805b

introduced two empty java files,

https://github.com/apache/activemq-artemis/blob/3ee6699a0ea576975b78f0248fe061d26296805b/artemis-web/src/test/java/org/apache/activemq/cli/test/WebServerCLITest.java

https://github.com/apache/activemq-artemis/blob/3ee6699a0ea576975b78f0248fe061d26296805b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/failover/LiveCrashOnBackupSyncTest.java

jbertram:
I removed those 2 empty files and pushed your commit directly.

PR #2546, opened 2019-02-08
ARTEMIS-2249 - update cpp example to use correct client libs

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-2249


andytaylor:
This must be env related, I only changed an example

clebertsuconic:
I will merge this today.

PR #2287, opened 2018-09-03
ARTEMIS-2069 Backup doesn't activate after shared store is reconnected

TomasHofman:
https://issues.apache.org/jira/browse/ARTEMIS-2069

Fix tries to prevent a server activation thread from terminating when FileLockNodeManager.tryLock() throws an IOException, e.g. because temporarily inaccessible NFS directory.
The node manager will repeat tryLock() call every two seconds.
WARN message with stack trace will be printed on first failure, DEBUG messages will be printed on recurring failures.

PR #2547, opened 2019-02-08
Patched with live lock evaluation

emagiz:
When using the shared store the live server can loose the lock on the journal but does not notice it. This can happen when a shared file system is being used like in AWS where we use EFS.

This can cause problems when the live server regains the network file system connection and just continues to process messages. At some point the live or the backup quits because it notices changes on the filesystems which it did not do itself.

We were able to prevent the occurence by tweaking EFS connection settings so it does not occur anymore in our setup.

For artemis we would like to show our change maybe someone can review the change and see if it can be improved and adapted in artemis.

Patch is here for master:
https://github.com/emagiz/activemq-artemis/commit/788adfbd3e5a54c63eed0810b7377641684b6fe1.patch

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
@@ -20,6 +20,13 @@ import java.io.File;
 import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileLock;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.activemq.artemis.api.core.ActiveMQIllegalStateException;
 import org.apache.activemq.artemis.api.core.SimpleString;
@@ -49,7 +56,9 @@ public class FileLockNodeManager extends NodeManager {
 
    private static final byte NOT_STARTED = 'N';
 
-   private FileLock liveLock;
+   private static final int LOCK_MONITOR_TIMEOUT_MILLIES = 2000;
+
+   private volatile FileLock liveLock;
 
    private FileLock backupLock;
 
@@ -57,13 +66,18 @@ public class FileLockNodeManager extends NodeManager {
 
    protected boolean interrupted = false;
 
-   public FileLockNodeManager(final File directory, boolean replicatedBackup) {
+   private ScheduledExecutorService scheduledPool;
+
+   public FileLockNodeManager(final File directory, boolean replicatedBackup, ScheduledExecutorService scheduledPool) {
       super(replicatedBackup, directory);
+      this.scheduledPool = scheduledPool;
    }
 
-   public FileLockNodeManager(final File directory, boolean replicatedBackup, long lockAcquisitionTimeout) {
+   public FileLockNodeManager(final File directory, boolean replicatedBackup, long lockAcquisitionTimeout,
+         ScheduledExecutorService scheduledPool) {
       super(replicatedBackup, directory);
 
+      this.scheduledPool = scheduledPool;
       this.lockAcquisitionTimeout = lockAcquisitionTimeout;
    }
 
@@ -174,6 +188,7 @@ public class FileLockNodeManager extends NodeManager {
          public void activationComplete() {
             try {
                setLive();
+               startLockMonitoring();
             } catch (Exception e) {
                ActiveMQServerLogger.LOGGER.warn(e.getMessage(), e);
             }
@@ -183,6 +198,7 @@ public class FileLockNodeManager extends NodeManager {
 
    @Override
    public void pauseLiveServer() throws Exception {
+      stopLockMonitoring();
       setPaused();
       if (liveLock != null) {
          liveLock.release();
@@ -191,6 +207,7 @@ public class FileLockNodeManager extends NodeManager {
 
    @Override
    public void crashLiveServer() throws Exception {
+      stopLockMonitoring();
       if (liveLock != null) {
          liveLock.release();
          liveLock = null;
@@ -318,7 +335,8 @@ public class FileLockNodeManager extends NodeManager {
          }
       }
 
-      // todo this is here because sometimes channel.lock throws a resource deadlock exception but trylock works,
+      // todo this is here because sometimes channel.lock throws a resource deadlock
+      // exception but trylock works,
       // need to investigate further and review
       FileLock lock;
       do {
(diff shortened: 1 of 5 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -338,5 +345,104 @@ protected FileLock lock(final long lockPosition) throws Exception {
       while (lock == null);
       return lock;
    }
+   
+   private void startLockMonitoring()	{
+	   logger.debug("Starting the lock monitor");
+	   Thread monitorThread = new Thread(new MonitorLock());
```
michaelandrepearce: We have thread executors. It is important it must use these. Also this solution uses up a thread permanently which goes agaisnt some of the thread modelling. Look maybe to use schedule thread executor

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -338,5 +345,104 @@ protected FileLock lock(final long lockPosition) throws Exception {
...
+				lostLock = true;
+			}
+			
+			if (lostLock)	{
+				logger.warn("Lost the lock according to the monitor, notifying listeners");
+				notifyLostLock();
+			}
+			
+			try {
+				Thread.sleep(LOCK_MONITOR_TIMEOUT_MILLIES);
```
michaelandrepearce: Use a schedule executor. Instead of sleeping threads blocking and using up a thread

michaelandrepearce:
Left some comments around using the common executor pools and better thread use.

As like anything test case as well please

michaelandrepearce:
Also could you look at:

https://github.com/apache/activemq-artemis/pull/2287

Are these to solutions dealing with same issue

michaelandrepearce:
Nudge are you looking to address the comments?

emagiz:
Yes, I will. I will address your comments and I checked the other pull request which addresses a different issue. It will take some time for me to write the unit tests but I can probably borrow some from the other pull request.
The other pull request addresses the bailing out of activation where this pull request addresses the bailing out while running and loosing the lock.

emagiz:
It's taking some time because we are also looking into large message support using amqp where we get two different issues with broken consumers and journal-buffer-size crashes of the message broker. So we also want to mitigate those issues but it might be a bridge to far for us now because of lacking knowledge in our team. https://issues.apache.org/jira/projects/ARTEMIS/issues/ARTEMIS-1975?filter=allopenissues and 
AMQ222010: Critical IO Error, shutting down the server. file=NIOSequentialFile /opt/emagiz/instances/emagiz-ipaas.jms-server.01/data/artemis/journal/activemq-data-5.amq, message=Can't write records bigger than the bufferSize(501760) on the journal
when using amqp which I didn't find on jira.

michaelandrepearce:
@emagiz if you havent seen on dev list, we are looking to release 2.7.0 within the week, how are you progressing? Do you think your reworked solution will be ready with time for review and further feedback?

emagiz:
Will be working on it today and afterwards until it is done. Comments should be done quickly because there are few and they are not hard to implement. 
Unit test I have experience with but not in this project and with these classes so will take longer, I do not know how to estimate how much time I will need on these. There are already a lot of unit tests so I can probably get started quickly by looking at those.

clebertsuconic:
rebase your branch please? squash your commits, and fix checkstyle?


I'm leaving this commit towards 2.7..1 now.. I'm already tagging it.

michaelandrepearce:
@emagiz looking much improved, on top of the bits clebert mentioned, could we have a test case for this as well please.

emagiz:
Yes, I'm adding that right now and looking into byteman to do so (new to that, but looks effective).
I saw a commit fixing one of the issues mentioned above related to calculation of message size now including headers, great to see that.
Someone else on our side is gathering more information (by testing with different setups) around the amqp (qpid) consumers not handling large messages anymore (5mb messages) which will be our next focus point after we have the test cases for this one.

jbertram:
@emagiz, can you resolve the conflict in FileLockNodeManager?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -49,21 +56,40 @@
...
 
    protected long lockAcquisitionTimeout = -1;
 
    protected boolean interrupted = false;
 
+   private ScheduledExecutorService scheduledPool;
+
    public FileLockNodeManager(final File directory, boolean replicatedBackup) {
       super(replicatedBackup, directory);
+      this.scheduledPool = new ScheduledThreadPoolExecutor(1);
```
michaelandrepearce: Should use the global threadpool for scheduled tasks not its own

emagiz: I left this one for some test case(s) which I didn't want to change. But I can change the test cases to use the new constructor. So unless the reason changes your mind I will be changing the test cases to take care of creating this executor and passing it to the FileLockNodeManager.

tests/extra-tests/src/test/java/org/apache/activemq/artemis/tests/extras/byteman/FileLockMonitorTest.java
```diff
@@ -0,0 +1,131 @@
...
+      lostLock = false;
+   }
+
+   @Test
+   @BMRules(rules = {
+         @BMRule(name = "lock is invalid", targetClass = "sun.nio.ch.FileLockImpl", targetMethod = "isValid", action = "return false;") })
+   public void testLockMonitorInvalid() throws Exception {
+      lostLock = false;
+      startServer();
+      Thread.sleep(5000);
```
wy96f: use a latch, a wait, etc. instead of sleep.

tests/extra-tests/src/test/java/org/apache/activemq/artemis/tests/extras/byteman/FileLockMonitorTest.java
```diff
@@ -0,0 +1,131 @@
...
+      executor.shutdown();
+   }
+
+   @Test
+   @BMRules(rules = {
+         @BMRule(name = "lock is invalid", targetClass = "org.apache.activemq.artemis.core.server.impl.FileLockNodeManager", targetMethod = "getState", action = "throw new java.io.IOException(\"EFS is disconnected\");") })
+   public void testLockMonitorIOException() throws Exception {
+      lostLock = false;
+      startServer();
+      Thread.sleep(5000);
```
wy96f: use a latch, a wait, etc. instead of sleep.

tests/extra-tests/src/test/java/org/apache/activemq/artemis/tests/extras/byteman/FileLockMonitorTest.java
```diff
@@ -0,0 +1,131 @@
...
+      }
+      nodeManager.crashLiveServer();
+      executor.shutdown();
+   }
+
+   @Test
+   public void testLockMonitorHasCorrectLockAndState() throws Exception {
+      lostLock = false;
+      startServer();
+      Thread.sleep(5000);
```
wy96f: use a latch, a wait, etc. instead of sleep.

tests/extra-tests/src/test/java/org/apache/activemq/artemis/tests/extras/byteman/FileLockMonitorTest.java
```diff
@@ -0,0 +1,131 @@
...
+      executor.shutdown();
+   }
+
+   @Test
+   @BMRules(rules = {
+         @BMRule(name = "lock is invalid", targetClass = "org.apache.activemq.artemis.core.server.impl.FileLockNodeManager", targetMethod = "getState", action = "return 70;") })
+   public void testLockMonitorHasLockWrongState() throws Exception {
+      lostLock = false;
+      startServer();
+      Thread.sleep(5000);
```
wy96f: use a latch, a wait, etc. instead of sleep.

jbertram:
This is superseded by #2901.

PR #2548, opened 2019-02-11
ARTEMIS-2118 Enhanced Message Groups Support

michaelandrepearce:
Support using group buckets on a queue for better local group scaling
Support disabling message groups on a queue
Support rebalancing groups when a consumer is added.


billpoole-mi:
I have a couple of questions regarding the "support rebalancing groups when a consumer is added" aspect of this change:

- When a message group X is rebalanced from consumer A to consumer B, how will consumer A be informed that it is no longer assigned message group X and shouldn't expect any further messages in that message group? Will Artemis send the consumer a message, perhaps with JMSXGroupSeq = -1, or perhaps some kind of special header like JMSXGroupLastForConsumer?
- Can consumer A release possession of the messages in message group X that consumer A is still processing (i.e. hasn't yet acknowledged) by acknowledging those messages as "released" (as opposed to accepted/rejected), such that those messages are then resent by Artemis to consumer B?

michaelandrepearce:
@billpoole-mi No there wont be any special message.

The intent here is for apps where you want evenly load balanced groups but where apps can handle a slight change in which groups they are handling, e.g. this is very usefull where a stateless app that simply needs to process and enrich data, and produces onwards. Any data relating to the group can be loaded lazily.

RE: JMSXGroupLastForConsumer Because an consumers can be added at anypoint, when the group may not have a message to dispatch onwards it would not be possible it would not be possible to add.

This said if you have the need, you can though listen for notification CONSUMER_ADDED on notifications address, and as such could then if you know that queue is using that feature, could know rebalance will occur.

billpoole-mi:
There may be a workable workflow for transferring ownership of a message group, where that workflow is cooperation between Artemis and the existing consumer assigned for that message group.

What if a message group is flagged for rebalancing when the need arises. Then, the next message to arrive in that message group is sent to the existing consumer with a special header that informs the consumer that the message group is being reassigned. That consumer is then expected to "release" all messages it is working on in that message group back to Artemis, which then sends them to the newly assigned consumer for that message group.

If no further message turns up, then it's not really a load balancing issue.

What do you think?

michaelandrepearce:
@billpoole-mi as i said, if/when this merges, you're more than welcome to contribute further enhancements you want.

billpoole-mi:
@michaelandrepearce no worries. I really appreciate all the work you guys are doing and just wanted to check whether something like I proposed would be seen by the team as valuable/worthwhile or not.

michaelandrepearce:
So yes i think if you find a way to elegantly achieve that without a performance impact i personally would be open to reviewing it. I like the idea.

Another further enhancement could be to add header when a group is assigned to a consumer, for signal of firstmessage in group also, i think there is a jira in backlog for that one already.

billpoole-mi:
Okay great! You are correct regarding a header for when a group is assigned to a consumer. That's the JMSXGroupFirstForConsumer header in ActiveMQ 5.x, described [here](http://activemq.apache.org/message-groups.html).

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/BucketMap.java
```diff
@@ -0,0 +1,163 @@
...
+      return buckets.isEmpty();
+   }
+
+   @Override
+   public boolean containsKey(Object key) {
+      return buckets.containsKey(getBucket(key));
+   }
+
+   private int getBucket(Object key) {
+      return (key.hashCode() & Integer.MAX_VALUE) % bucketCount;
```
franz1981: Why masking it with Integer.MAX_VALUE?
The % op could be replaced by an & (bucketCount-1) with power of 2 bucketCount

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/IntHashMap.java
```diff
@@ -0,0 +1,878 @@
...
+      this(initialCapacity, loadFactor, true);
+   }
+
+   /**
+    * Construct a new map allowing a configuration for initial capacity and load factor.
+    * @param initialCapacity      for the backing array
+    * @param loadFactor         limit for resizing on puts
+    * @param shouldAvoidAllocation should allocation be avoided by caching iterators and map entries.
+    */
+   public IntHashMap(
```
franz1981: If it is taken from agrona I would reference it somehow to both give proper credits and track eventual issues on it...

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3347,6 +3346,16 @@ public void onError(int errorCode, String errorMessage) {
 
    }
 
+   public static Map<SimpleString, Consumer> groupMap(int groupBuckets) {
+      if (groupBuckets == -1) {
+         return new HashMap<>();
```
franz1981: You can use the IntObjMap here as well

michaelandrepearce: No we cant the key is simplestring

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/BucketMap.java
michaelandrepearce: So its always positive

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/IntHashMap.java
michaelandrepearce: I actually dont need this class, just realised Netty has one, so can use it.

michaelandrepearce:
@franz1981 

as noted i found Netty has an int primitive hashmap implementation using open addressing implementation so can use that, so removed IntHashMap to avoid maintenance and testing burden.

added tests for NoOpMap, BucketMap as well as docs.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/BucketMap.java
```diff
@@ -0,0 +1,177 @@
...
+ * The number of buckets is provided at construction.
+ *
+ * Its initial use, is in QueueImpl, where we want to bucket message groups to consumers.
+ *
+ * @param <K> the key type.
+ * @param <V> the value type.
+ */
+public class BucketMap<K, V> implements Map<K, V> {
+
+   private final IntObjectMap<V> buckets = new IntObjectHashMap<>();
```
franz1981: I'm not sure if I have already left this comment but % is really an heavy operation to be done on hot path so I suggest to use the power of 2 + mask trick

franz1981: And..idea..given that you are % on bucket count is you could have collisions. Maybe you don't need a map here but just an array bucketCount sized...

michaelandrepearce: Collisions are 100% expected we are bucketing. E.g if bucket size is two and the key was integer with keys 1 and 3 i actually expect them to collide and goto the same bucket.

Re map vs array, i was toying on the same thing, downside with array is up front will need to allocate the memory for that array even if not all buckets used. Benefit of using underlying map is it will grow as needed.

franz1981: Not sure...consider that the map will allocate both the int and obj arrays multiplied by the load factor. And we will pay the rehashing too..a fixed size thing will be more "direct" (less pointer chasing) and simple (just obj array, never resized). 
Is just a personal opinion eh, I'm anyway happy about it.

franz1981: I will have some time on the next week and I will look better into this. ATM I have just my phone :)

michaelandrepearce: Ive actually found a small issue in Netty's implementation...they don't support values().itertator().remove(), annoyingly theres no reason either, the underlying iterator they facade would support it .... grrrrr

michaelandrepearce: @franz1981 issue with array is imaging someone on their broker (that someone being me) wants to ensure group buckets by default across the broker to say 20k, as such sets address setting match = #.

Then for every queue will create a 20k array, even if the producer on the address doesnt set message groups.

franz1981: I could implement it on Netty bud, np :)

michaelandrepearce: @franz1981 just because of that netty thing, i may have to have our own version again.

What im thinking of doing, is so we dont hold back this feature and it makes the 2.7.0 cut, is clone in the netty varient with a fix to our collections (they're apache license so no license issues in doing this, ill just add an attribution to the java doc), and then if netty fix it, then we can remove it again at some fututre point.

michaelandrepearce:
@franz1981 

Hopefully fixed the checkstyle, netty used 4 space tabs we use 3, lol, i feel like an episode of "silicone valley" now.

Anyhow once PR build completes and goes green are you happy?

Also i sent an PR to netty, so fingers crossed we don't need to have this clone/fork too long.
https://github.com/netty/netty/pull/8866

michaelandrepearce:
@franz1981 the needing to copy a IntObjectHashmap, still didnt fit right with me. 
As such have had a sit and think.

Have done a small refactor that means we can still use IntObjectHashmap without needing to clone it over.

Essentially in queueimpl for message groups, we only use, 
put
get
remove
clear
size 
iterate over the collection so we can for remove consumers. 

As such rather than using map interface i made a new Interface named 'MessageGroups' with just this subset, as such makes implementations that are not backed by a plain forward map a little easier as less methods to contend with, e.g. i can still use IntObjectHashMap but i don't have to make it fully implement Map<SimpleString, Consumer>.

michaelandrepearce:
@franz1981 and now because of the refactor and simpler methods we need to support i think i can easily implement the buckets one with a single array ... its like a win win .. just implementing the array bit now for you.

michaelandrepearce:
@franz1981 done, now using a single array (thats lazy initiated)

michaelandrepearce:
@franz1981 if you're happy could you merge please?

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/SimpleString.java
```diff
@@ -454,7 +454,16 @@ public boolean contains(final char c) {
     * @return the concatenated SimpleString
     */
    public SimpleString concat(final String toAdd) {
-      return concat(new SimpleString(toAdd));
+      int len = toAdd.length();
```
franz1981: This concat method need test coverage IMO

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/BucketMessageGroups.java
```diff
@@ -0,0 +1,164 @@
...
+   private int getBucket(SimpleString key) {
+      Object bucketKey = key;
+      if (key.startsWith(_AMQ_GROUP_BUCKET_INT_KEY)) {
+         bucketKey = retrieveBucketIntFromKey(key);
+      }
+      return getHashBucket(bucketKey, bucketCount);
+   }
+
+   private static int getHashBucket(final Object key, final int bucketCount) {
+      return (key.hashCode() & Integer.MAX_VALUE) % bucketCount;
```
franz1981: % is a quite heavy operation: masking with a power of 2 bucketCount would be better if possible.
This is suppose to be on the hot path?

michaelandrepearce: Ive responded to this same comment several times yes. That only works if buckets is power of two and we have to honour the bucket count set.

The actual performance compared to everything is tiny. Note we actually have mod in journal code fyi.

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/SimpleString.java
michaelandrepearce: I assume as existing code its already covered

michaelandrepearce:
Ive run what i can of the CI and integration tests
 Tbh i have never managed to run a full one. 

Its one of my biggest bugs it seems you need voodo magic. 

Would be nice if we had a once a night full ci run on master rather than relying on cleberts special box.

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/SimpleString.java
michaelandrepearce: It basically avoiding a new object and a byte array being created and thrown away uneededly. If its contentious i can revert, as not actually needed for this feature, just something i noticed when looking at object creation when profiling during dev of the feature.

michaelandrepearce: there is also explicit tests for concat in
org.apache.activemq.artemis.tests.util.SimpleStringTest.testConcat

michaelandrepearce:
Answered your queries franz, Ill look to merge on monday if no further comments.

clebertsuconic:
Can I have a day to review this? Say tomorrow?

franz1981:
I've asked the review of Clebert or Justin too: please wait until their "yes".
I could have missed something so better 4 eyes then just 2...

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/BucketMessageGroups.java
clebertsuconic: Did you implement this from scratch or did you base it on the ActiveMQ5 bucket?

We had an outstanding task to implement it based on AMQ5 version.

clebertsuconic: Ohh.. you hash the GroupID... got it!

I feel dumb that I don't understand why the bitwise operation.

clebertsuconic: that's just stripping out negatives.. got it!

clebertsuconic: @jbertram pointed me this is the same on AMQ5

clebertsuconic:
@michaelandrepearce man.. you put quite a lot of work here...  nice one!!!!!


What I liked here is that you configured the number of groups at queue level, during creation.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/BucketMessageGroups.java
michaelandrepearce: Yes its similar to amq5.

clebertsuconic:
@michaelandrepearce everything looks good.

Let me just run testsuite one time to be sure.. and i will merge it if everything good.

testsuite takes 3 hours on my box... I should be able to merge this either tognitght or early morning if no test issues. 

I don't foresee any major issues from what I looked.. but I wanted to make sure before merging a big PR like this.

clebertsuconic:
tests are good.. .merged

PR #2549, opened 2019-02-13
ARTEMIS-2251 Large messages might not be deleted when server crashed

qihongxu:
When deleting large messages, artemis will use storePendingLargeMessage to insert a temporary record in journal for reload, in case server crashed and large messages stayed forever. But in storePendingLargeMessage that appendAddRecord inserts records asynchronously. In this way there are potential risks that tasks in executor get lost due to server crash, which may lead to undeletable large messages. To solve this problem a Boolean is added to storePendingLargeMessage so that it will be forced to use SimpleWaitIOCallback in delete situation.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
@@ -432,7 +432,7 @@ public class JournalStorageManager extends AbstractJournalStorageManager {
       journalFF.releaseBuffer(buffer);
    }
 
-   public long storePendingLargeMessage(final long messageID, long recordID) throws Exception {
+   public long storePendingLargeMessage(final long messageID, long recordID, boolean wait) throws Exception {
       readLock();
       try {
          if (recordID == LargeServerMessage.NO_PENDING_ID) {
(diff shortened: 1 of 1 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -432,7 +432,7 @@ public void freeDirectBuffer(ByteBuffer buffer) {
       journalFF.releaseBuffer(buffer);
    }
 
-   public long storePendingLargeMessage(final long messageID, long recordID) throws Exception {
```
clebertsuconic: I'm not sure this is a good idea, this is introducing a blocking operation.

clebertsuconic: which will block any netty processing.

clebertsuconic: although delete itself is already blocking.. so I think it's fine.

michaelandrepearce: There is diff between a memory block and an waiting on disk io.

Can this not be async with a callback with a response later?

clebertsuconic: I am sending an alternate PR.

clebertsuconic: @qihongxu I fixed it without a blocking operation through a different PR. Please let me know if you see any issues.

Thanks

@michaelandrepearce  ^^

PR #2550, opened 2019-02-14
NO-JIRA Fix a "Computation of average could overflow" SpotBugs warning

jdanekrh:
https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#im-computation-of-average-could-overflow-im-average-computation-could-overflow

CC @michaelandrepearce


michaelandrepearce:
Great find. Thanks. Will merge

PR #2553, opened 2019-02-17
NO-JIRA Fix a "Inefficient Map Iterator" SpotBugs warning

jdanekrh:
https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#wmi-inefficient-use-of-keyset-iterator-instead-of-entryset-iterator-wmi-wrong-map-iterator

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AbstractControl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AbstractControl.java
@@ -106,8 +106,8 @@ public abstract class AbstractControl extends StandardMBean {
       try {
          CoreMessage message = new CoreMessage(storageManager.generateID(), 50);
          if (headers != null) {
-            for (String header : headers.keySet()) {
-               message.putStringProperty(new SimpleString(header), new SimpleString(headers.get(header)));
+            for (Map.Entry<String, String> header : headers.entrySet()) {
+               message.putStringProperty(new SimpleString(header.getKey()), new SimpleString(headers.get(header.getValue())));
             }
          }
          message.setType((byte) type);


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AbstractControl.java
```diff
@@ -106,8 +106,8 @@ protected String sendMessage(SimpleString address,
       try {
          CoreMessage message = new CoreMessage(storageManager.generateID(), 50);
          if (headers != null) {
-            for (String header : headers.keySet()) {
-               message.putStringProperty(new SimpleString(header), new SimpleString(headers.get(header)));
+            for (Map.Entry<String, String> header : headers.entrySet()) {
+               message.putStringProperty(new SimpleString(header.getKey()), new SimpleString(headers.get(header.getValue())));
```
jbertram: The point of the SpotBugs warning is to avoid calling `Map.get(key)`. However, this fix doesn't actually do that and even introduced [a bug](https://issues.apache.org/jira/browse/ARTEMIS-2332).

clebertsuconic: this should been:

```java
               message.putStringProperty(new SimpleString(header.getKey()), new SimpleString(header.getValue()));
```


is that what you're talking about?

How did you realize it?

jbertram: I realized it was broken when [ARTEMIS-2332](https://issues.apache.org/jira/browse/ARTEMIS-2332) was opened and I took a look to see what the problem might be. I've got a fix for it, but not yet a test. Will send a PR when it's all ready.

michaelandrepearce: Do you even need "new SimpleString" on the values? The underlying message will convert to simplestring if needed.

clebertsuconic: @michaelandrepearce  / @jbertram / @franz1981 from now on.. any "automated" changes like this will have a big flag for me.

It's easy to make changes while you're not even testing them.

PR #2557, opened 2019-02-20
Core client cleanups

orpiske:
These are a couple of relatively safe cleanups for the Core Client code for review. If adequate/accepted, I'll squash and combine them.


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
```diff
@@ -333,7 +333,7 @@ private ServerLocatorImpl(final Topology topology,
 
       this.discoveryGroupConfiguration = discoveryGroupConfiguration;
 
-      this.initialConnectors = transportConfigs != null ? transportConfigs : null;
```
clebertsuconic: nice catch. This used to be     
```
this.initialConnectors = transportConfigs != null ? transportConfigs : new TransportConfiguration[] {};
```

back in 2012, still at hornetq.

PR #2558, opened 2019-02-21
ARTEMIS-2257 Race condition when calling shutdownGracefully in SharedEventLoopGroup

wy96f:
While thread a shutdowns shared event loop, between scheduling a task and set the shutdown reference thread b calls getInstance() and retrieves ready shutting down instance. Add synchronized block to fix the problem.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/SharedEventLoopGroup.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/SharedEventLoopGroup.java
@@ -84,30 +84,32 @@ public class SharedEventLoopGroup extends DelegatingEventLoopGroup {
 
    @Override
    public Future<?> shutdownGracefully(final long l, final long l2, final TimeUnit timeUnit) {
-      if (channelFactoryCount.decrementAndGet() == 0) {
-         shutdown.compareAndSet(null, next().scheduleAtFixedRate(new Runnable() {
-            @Override
-            public void run() {
-               synchronized (SharedEventLoopGroup.class) {
-                  if (shutdown.get() != null) {
-                     Future<?> future = SharedEventLoopGroup.super.shutdownGracefully(l, l2, timeUnit);
-                     future.addListener(new FutureListener<Object>() {
-                        @Override
-                        public void operationComplete(Future<Object> future) throws Exception {
-                           if (future.isSuccess()) {
-                              terminationPromise.setSuccess(null);
-                           } else {
-                              terminationPromise.setFailure(future.cause());
+      synchronized (SharedEventLoopGroup.class) {
+         if (channelFactoryCount.decrementAndGet() == 0) {
+            shutdown.compareAndSet(null, next().scheduleAtFixedRate(new Runnable() {
+               @Override
+               public void run() {
+                  synchronized (SharedEventLoopGroup.class) {
+                     if (shutdown.get() != null) {
+                        Future<?> future = SharedEventLoopGroup.super.shutdownGracefully(l, l2, timeUnit);
+                        future.addListener(new FutureListener<Object>() {
+                           @Override
+                           public void operationComplete(Future<Object> future) throws Exception {
+                              if (future.isSuccess()) {
+                                 terminationPromise.setSuccess(null);
+                              } else {
+                                 terminationPromise.setFailure(future.cause());
+                              }
                            }
-                        }
-                     });
-                     instance = null;
+                        });
+                        instance = null;
+                     }
                   }
                }
-            }
 
-         }, 10, 10, TimeUnit.SECONDS));
+            }, 10, 10, TimeUnit.SECONDS));
+         }
+         return terminationPromise;
       }
-      return terminationPromise;
    }
 }
--- a/artemis-core-client/src/test/java/org/apache/activemq/artemis/core/remoting/impl/netty/SharedEventLoopGroupTest.java
+++ b/artemis-core-client/src/test/java/org/apache/activemq/artemis/core/remoting/impl/netty/SharedEventLoopGroupTest.java
@@ -1,0 +1,97 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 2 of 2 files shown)


clebertsuconic:
@wy96f did you see my comment on the JIRA?

Synchronized SharedEventLoopGroup is not the bug, it's the fix you applied to fix the bug.

I don't know in what condition the issue happened. 

Can you add more detail on how it happened? a test would be great if possible... if you think a test is almost impossible due to an issue lets discuss and see if there's a way around.

Fixes without tests should be an exception... not the rule.

wy96f:
> @wy96f did you see my comment on the JIRA?
> 
> Synchronized SharedEventLoopGroup is not the bug, it's the fix you applied to fix the bug.
> 
> I don't know in what condition the issue happened.
> 
> Can you add more detail on how it happened? a test would be great if possible... if you think a test is almost impossible due to an issue lets discuss and see if there's a way around.
> 
> Fixes without tests should be an exception... not the rule.

Sorry for my ambiguousness. We use jms client. We will close ActiveMqConnection if in idle for 5 minutes and recreate connection if some requests come in. These days we found exceptions occasionally as follows:

2019-02-24 08:23:44.294 [] [ThreadPoolAsynchronousProcessor-thread-2] WARN  io.netty.channel.AbstractChannel -Force-closing a channel whose registration task was not accepted by an event loop: [id: 0xab32cb4c]
java.util.concurrent.RejectedExecutionException: event executor terminated
	at io.netty.util.concurrent.SingleThreadEventExecutor.reject(SingleThreadEventExecutor.java:821)
	at io.netty.util.concurrent.SingleThreadEventExecutor.offerTask(SingleThreadEventExecutor.java:327)
	at io.netty.util.concurrent.SingleThreadEventExecutor.addTask(SingleThreadEventExecutor.java:320)
	at io.netty.util.concurrent.SingleThreadEventExecutor.execute(SingleThreadEventExecutor.java:746)
	at io.netty.channel.AbstractChannel$AbstractUnsafe.register(AbstractChannel.java:479)
	at io.netty.channel.SingleThreadEventLoop.register(SingleThreadEventLoop.java:80)
	at io.netty.channel.SingleThreadEventLoop.register(SingleThreadEventLoop.java:74)
	at io.netty.channel.MultithreadEventLoopGroup.register(MultithreadEventLoopGroup.java:86)
	at org.apache.activemq.artemis.core.remoting.impl.netty.DelegatingEventLoopGroup.register(DelegatingEventLoopGroup.java:46)
	at io.netty.bootstrap.AbstractBootstrap.initAndRegister(AbstractBootstrap.java:332)
	at io.netty.bootstrap.Bootstrap.doResolveAndConnect(Bootstrap.java:163)
	at io.netty.bootstrap.Bootstrap.connect(Bootstrap.java:145)
	at org.apache.activemq.artemis.core.remoting.impl.netty.NettyConnector.createConnection(NettyConnector.java:616)
	at org.apache.activemq.artemis.core.client.impl.ClientSessionFactoryImpl.openTransportConnection(ClientSessionFactoryImpl.java:1004)
	at org.apache.activemq.artemis.core.client.impl.ClientSessionFactoryImpl.createTransportConnection(ClientSessionFactoryImpl.java:1040)
	at org.apache.activemq.artemis.core.client.impl.ClientSessionFactoryImpl.establishNewConnection(ClientSessionFactoryImpl.java:1232)
	at org.apache.activemq.artemis.core.client.impl.ClientSessionFactoryImpl.getConnection(ClientSessionFactoryImpl.java:862)
	at org.apache.activemq.artemis.core.client.impl.ClientSessionFactoryImpl.getConnectionWithRetry(ClientSessionFactoryImpl.java:775)
	at org.apache.activemq.artemis.core.client.impl.ClientSessionFactoryImpl.connect(ClientSessionFactoryImpl.java:221)
	at org.apache.activemq.artemis.core.client.impl.ServerLocatorImpl.createSessionFactory(ServerLocatorImpl.java:643)
	at org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory.createConnectionInternal(ActiveMQConnectionFactory.java:748)
	at org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory.createConnection(ActiveMQConnectionFactory.java:248)

Before the stack trace some logs were printed showing idle connection is closed.
Our code is like this:
        serverLocator = ActiveMQClient.createServerLocatorWithoutHA(transportConfiguration);
        serverLocator.setUseGlobalPools(true);
        serverLocator.setRetryInterval(2000);
        serverLocator.setRetryIntervalMultiplier(2.0);
        serverLocator.setMaxRetryInterval(10000);
        serverLocator.setReconnectAttempts(5);
        serverLocator.setCallFailoverTimeout(0);
        serverLocator.setCallTimeout(10000);
        serverLocator.setProducerWindowSize(-1);
        serverLocator.setMinLargeMessageSize(1024 * 1024);
        int windowSize = slowConsumer ? 1 : 200 * 1024;
        serverLocator.setConsumerWindowSize(windowSize);
        final ActiveMQConnection amqConnection = (ActiveMQConnection) factory.createConnection(username, password);

Then we investigated the SharedEventLoopGroup code. There was some race conditions:
Thread t1 closes connection and shutdowns the SharedEventLoopGroup. When t1 decrement channelFactoryCount to 0, thread t2 is creating connection and getting the instance. After that t1 schedules a shutdown task and cas shutdown reference. After 10 secs instance used by t1 will shutdown.

michaelandrepearce:
Whats occuring on this on? Im keen we get bug fix prs closed before 2.7 cut.

clebertsuconic:
I was looking on this... I'm a bit puzzled on what is being fixed.. there are no tests.

clebertsuconic:
-1 on this. When you shutdown the client. the executors should go down gracefully. 


I would need some example on what's going on.. and fix the real issue.

Whatever happened this does not seem the proper fix.

clebertsuconic:
if you provide me a test. .I can change my mind

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/SharedEventLoopGroup.java
```diff
@@ -84,30 +84,32 @@ public ThreadFactory run() {
...
-                  if (shutdown.get() != null) {
-                     Future<?> future = SharedEventLoopGroup.super.shutdownGracefully(l, l2, timeUnit);
-                     future.addListener(new FutureListener<Object>() {
-                        @Override
-                        public void operationComplete(Future<Object> future) throws Exception {
-                           if (future.isSuccess()) {
-                              terminationPromise.setSuccess(null);
-                           } else {
-                              terminationPromise.setFailure(future.cause());
+      synchronized (SharedEventLoopGroup.class) {
```
clebertsuconic: if you applied the synchronize here to this instead of .class I would be ok with merging this without a test. 

A synchornize on a .class here.. without a test.. I don't know what implications it would have.

Can you tell me why the synchornize on the class was needed?

michaelandrepearce: Thats a good point there is risk especially where you may run two seperate clients to different brokers or have multiple embedded brokers. Sync on class or static sharing isnt always safe

clebertsuconic:
I see you added a test, but you still haven't addressed my question about synchronized on .class.

Why is that needed?

clebertsuconic:
I see the issue... getInstance applies on synchrnoized class, and shutdown.getAndSet is executed in there.

This is just acting on the scheduling.. not the actual waiting. 


Anything SharedEventLoopGroup is a factory for the entire VM, so the synchronized should be applied on class. 


(That was what I was looking for)

PR #2559, opened 2019-02-21
ARTEMIS-2258 The FileLockNodeManager directory should be configurable

franz1981:


franz1981:
I'm doing some local tests to see if it works as expected, but the configuration bits are already covered by tests.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -608,6 +608,14 @@
             </xsd:annotation>
          </xsd:element>
 
+         <xsd:element name="node-manager-lock-directory" type="xsd:string" maxOccurs="1" minOccurs="0">
```
clebertsuconic: Can't we call it just lock-folder, or lock-directory? Similar to the name you used on a test?

michaelandrepearce:
Whislt i see tests for the config, is there any way to make an automated test to check what this is trying to fix?

franz1981:
Nice one, I can modify the test used for shared store HA.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
franz1981: I wasn't sure to do it because we have already a lock folder for the CLI lock file and there is a pending issue to make that configurable..I don't want to confuse people...

michaelandrepearce: Yeah thats going to be confusing agreed. Im already confused whats the diff ;)

clebertsuconic: Can’t we just use the same folder? Do we need separate entries for lock and node manager lock ?

franz1981: I suppose yes, because are used for different purposes:

- cli lock is used by cli commands to avoid executing in parallel with a specific broker instance
- node manager lock is used to coordinate different broker instances

I don't think is a good idea to mix their configurations

michaelandrepearce: Would you not want cli locked in a distributed way also? E.g. if you had a shared san master slave you wouldnt want to run in parallel the cli on slave and master

franz1981: > Would you not want cli locked in a distributed way also?

I suppose not: the cli commands are meant to run just on a single specific instance AFAIK.
https://issues.apache.org/jira/browse/ARTEMIS-528 is the reference JIRA for the cli lock change: IMHO I would left these 2 configurations as separated given that are meant to be changed (if they are) for very different reasons/purposes.

clebertsuconic: I would expect to be the same folder. i wouldn't want to run the same on backup and live.

franz1981: > I would expect to be the same folder

Do you mean for both the cli lock and the HA shared store lock?

clebertsuconic: The only reason I didn’t reuse the same file on the CLI was the richer file.  It made it easier to use a new one. 

So it’s the same functionality. 

We can chat online about it.

clebertsuconic: Hmmm.  Let me think about it.

franz1981: The file used for HA contains the state of the current live instance and the nodeId so I won't feeling very good to mix the purposes, but I understand that the one used for the node manager could be used as the cli lock too

franz1981:
@michaelandrepearce I've added support for the `NettyFailoverTest` that's using the node managers for every mechanics with the added benefit that `FileLockNodeManager` has a much improved coverage now, given that the test wasn't originally used to test it.

PR #2561, opened 2019-02-22
ARTEMIS-2259 Client session not exist if reattach timeout

wy96f:
When client session reattach on new connection or session recreation timeout, an exception is thrown causing session recreation and consumer recreation don't complete. Although client connection failover completes, client session actually doesn't work bcs there is no corresponding session channel on server side. We close connection if session failover fails to fix the problem.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
@@ -771,7 +771,11 @@ public class ClientSessionFactoryImpl implements ClientSessionFactoryInternal, C
       ((CoreRemotingConnection) connection).syncIDGeneratorSequence(((CoreRemotingConnection) oldConnection).getIDGeneratorSequence());
 
       for (ClientSessionInternal session : sessionsToFailover) {
-         session.handleFailover(connection, cause);
+         if (!session.handleFailover(connection, cause)) {
+            connection.destroy();
+            this.connection = null;
+            return;
+         }
       }
    }
 
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
@@ -1350,10 +1350,12 @@ public final class ClientSessionImpl implements ClientSessionInternal, FailureLi
    // Needs to be synchronized to prevent issues with occurring concurrently with close()
 
    @Override
-   public void handleFailover(final RemotingConnection backupConnection, ActiveMQException cause) {
+   public boolean handleFailover(final RemotingConnection backupConnection, ActiveMQException cause) {
+      boolean suc = true;
+
       synchronized (this) {
          if (closed) {
-            return;
+            return true;
          }
 
          boolean resetCreditManager = false;
@@ -1426,6 +1428,7 @@ public final class ClientSessionImpl implements ClientSessionInternal, FailureLi
             }
          } catch (Throwable t) {
             ActiveMQClientLogger.LOGGER.failedToHandleFailover(t);
+            suc = false;
          } finally {
             sessionContext.releaseCommunications();
          }
@@ -1447,6 +1450,8 @@ public final class ClientSessionImpl implements ClientSessionInternal, FailureLi
       }
 
       sessionContext.resetMetadata(metaDataToSend);
+
+      return suc;
 
    }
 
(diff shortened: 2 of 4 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
```diff
@@ -771,7 +771,11 @@ private void reconnectSessions(final RemotingConnection oldConnection,
       ((CoreRemotingConnection) connection).syncIDGeneratorSequence(((CoreRemotingConnection) oldConnection).getIDGeneratorSequence());
 
       for (ClientSessionInternal session : sessionsToFailover) {
-         session.handleFailover(connection, cause);
+         if (session.handleFailover(connection, cause)) {
+            connection.destroy();
```
clebertsuconic: Can you add a test for such condition?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
```diff
@@ -1350,10 +1350,12 @@ public void preHandleFailover(RemotingConnection connection) {
    // Needs to be synchronized to prevent issues with occurring concurrently with close()
 
    @Override
-   public void handleFailover(final RemotingConnection backupConnection, ActiveMQException cause) {
+   public boolean handleFailover(final RemotingConnection backupConnection, ActiveMQException cause) {
```
clebertsuconic: true usually means success. this is a bit confusing. Can you make it opposite.

meaning handleFailover would return true if succeeded.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
wy96f: We'll add test to simulate the scenario where reattach timeout. Just not sure the proper way to fix the problem:)

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
wy96f: This makes sense:) We'll modify it along with the test code.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
clebertsuconic: Well.. .the thing is... I"m not sure what was the problem to be able to tell you what is the fix. a test would help.

If you can't provide a test.. for something rare happening in production, it would be nice some more details.. like logs.. stack traces.... 


i couldn't figure out the issue by just the synchrnoized implementation.

wy96f: The synchrnoized implementation? Do you mean #2558 ?  This issue is easy to simulate:)

clebertsuconic: I actually confused with this issue.. but yeah.. on both cases.. a test case... or more detailed information would help.


Think about this: another user is having a similar issue, if you only provide the fix and not what happened they will never be able to pinpoint this is the fix they need. for both cases.

michaelandrepearce: @clebertsuconic i think i have seen something similar in our production environment, unfortunatly because its prod they restarted the client app too quickly before details were gathered.

@wy96f as clebert notes, would be great to have a test case then i could also tell if it is what ive also seen.

wy96f:
@clebertsuconic @michaelandrepearce Hi, we have added test. Please review it :)

clebertsuconic:
I don't think this is correct...

if you can't reconnect for any reason, the failover code should loop retrying as many retries it can do.

On your case here it will give up right away.????

clebertsuconic:
Why reattach would fail? I'm just trying to determine if the proper change would be to retry the logic.

clebertsuconic:
ok, I see this as a palliative in case the reattach failed. although such thing should not happen.


I'm still curious about why such thing happened on your case.

clebertsuconic:
@wy96f I will merge this, but please provide more information on how reattach failed please.

wy96f:
@cle

> @wy96f I will merge this, but please provide more information on how reattach failed please.

This was caused by large scale network infrastructure outrage in our company. We had assumptions either tcp connection was successfully established but reattch packet was lost bcs of network instability  or large incoming traffic caused server gc taking too long to respond.

PR #2563, opened 2019-02-23
NO-JIRA Document reloading when using modularised broker.xml

michaelandrepearce:


clebertsuconic:
@michaelandrepearce if you're merging your own commit on a small thing, there's no need to send a  PR.. just send the commit directly.


Unless you want to make sure it's not breaking the build through the PR CI.

PR #2564, opened 2019-02-23
ARTEMIS-2183 Useless statement in public synchronized List

orpiske:
Removes the seemingly unnecessary call to addAll to self.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
@@ -190,9 +190,7 @@ public class RefsOperation extends TransactionOperationAbstract {
 
    @Override
    public synchronized List<MessageReference> getRelatedMessageReferences() {
-      List<MessageReference> listRet = new LinkedList<>();
-      listRet.addAll(listRet);
-      return listRet;
+      return refsToAck;
    }
 
    @Override


orpiske:
> I would assume this is a bug and actually is meant to be populated by an internal field

Maybe it should be combining either refsToAck or pagedMessagesToPostACK ... or, maybe, even both?

franz1981:
Take a look at the history for it ans maybe it has been changed from something that was correct before!

orpiske:
> Take a look at the history for it ans maybe it has been changed from something that was correct before!

It seems that it was already this way when this code was introduced, [around October 2014](https://github.com/apache/activemq-artemis/commit/177e6820b541e0a71b952eebf503a4d2235910c5#diff-321554fd18f6c7a3a382a7d50ce7a1abR191). Maybe we would need to go even farther ahead in the past to figure out.

Looking at the original HornetQ code, it seems that it used to [return the refsToAck](https://github.com/hornetq/hornetq/blob/HornetQ_2_2_25_Final/src/main/org/hornetq/core/server/impl/QueueImpl.java#L2712).

orpiske:
@franz1981 @michaelandrepearce thanks for the insights and ideas! 

In light of that, I changed it to return the refsToAck as it may be a more suitable fix.

franz1981:
I think that the original idea were to return a copy of it, that's why the method were synchronized..
But the method was used??

michaelandrepearce:
Id agree with franz for it not to be causing issue you got to think is this even being used. Is it and the calling code become redundant. And if not there must be a test case

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
```diff
@@ -190,9 +190,7 @@ private void decrementRefCount(MessageReference refmsg) {
 
    @Override
    public synchronized List<MessageReference> getRelatedMessageReferences() {
-      List<MessageReference> listRet = new LinkedList<>();
```
clebertsuconic: this is dangerous... The RefsOperation will clear the list upon completion.
The callers could be using the list for other factors...

You really need the isolation here.


Can you close this PR? This is meant to be.

clebertsuconic: I'm not sure this is a good idea. I think this needs to be isolated.

at least that would break synchronization on the list.. a method listing JSon while the transaction still being appeneded.. you could get consequences almost impossible to test.


Can you close this PR pretty please?


I appreciate you know the codebase well... but this was meant to be

clebertsuconic:
-1000

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
michaelandrepearce: I think the point here is the code is addall of the list it just created. This seems very odd. 

Id agree an internal list should just be exposed. But the current logic just looks crazy wrong. Youd be better to return Collections.emptylist!

orpiske: Sure thing @clebertsuconic, I appreciate the explanation and I'll close that one.

orpiske: Yeah, the original commit was just replacing the addAll with a return of an empty list. I modified it to return refsAcks. 

Given that there can be unintended consequences here, I am closing the PR and letting those of you with more experience on the code base to decide what may be the more appropriate solution in this case.

Nonetheless, I agree with @michaelandrepearce, that there is something strange with the original code that the PR intended to fix.

clebertsuconic: I guess I don't understand the issue then... /me back to looking into this.

clebertsuconic: OHHHHHH... I see now...


I was reading the method as it was supposed to be written:


```java

      List<MessageReference> listRet = new LinkedList<>();
      listRet.addAll(refsToAck);
      return listRet;
```


@orpiske  this is how this method was supposed to be written ^^

clebertsuconic: when I read addAll, my brain was tricking me and I was reading refsToAck...


if we make that change... it will be ok...


I guess this needs to validate if empty... or null and be a proper change.

orpiske: @clebertsuconic thanks for taking the time to go through the code and checking how it should be. I will reopen this PR and do the change as instructed!

PR #2566, opened 2019-02-25
ARTEMIS-2183 Useless statement in public synchronized List

orpiske:
Sending this PR again based feedback from @clebertsuconic. 

Ensures that the returned list returns the refsToAck list instead of trying to add all items of itself.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
@@ -191,7 +191,11 @@ public class RefsOperation extends TransactionOperationAbstract {
    @Override
    public synchronized List<MessageReference> getRelatedMessageReferences() {
       List<MessageReference> listRet = new LinkedList<>();
-      listRet.addAll(listRet);
+
+      if (refsToAck != null && !refsToAck.isEmpty()) {
+         listRet.addAll(refsToAck);
+      }
+
       return listRet;
    }
 


orpiske:
> So the only other thing here. Is this obviosuly has gone unnoticed for ages. To avoid this in future a test, ensuring its functionality (what ever that is)

So, I dig a bit further to find out where it was used. This is part of the management API and is exposed both via Hawtio and Jolokia. 

> 
> "Transaction heuristic operations
> 
> In case of a server crash, when the server restarts, it it possible that some transaction requires manual 
> intervention. The listPreparedTransactions() method lists the transactions which are in the prepared states (the transactions are represented as opaque Base64 Strings.) To commit or rollback a given prepared transaction, the commitPreparedTransaction() or rollbackPreparedTransaction() method can be used to resolve heuristic transactions. Heuristically completed transactions can be listed using the listHeuristicCommittedTransactions() and listHeuristicRolledBackTransactions methods."

In my case here, I called it like this: 

> curl -u admin:admin "http://localhost:8161/console/jolokia/exec/org.apache.activemq.artemis:broker=%220.0.0.0%22/listPreparedTransactionDetailsAsJSON()"

And like this: 

> curl -u admin:admin "http://localhost:8161/console/jolokia/exec/org.apache.activemq.artemis:broker=%220.0.0.0%22/listPreparedTransactionDetailsAsHTML()"                                                  [21:07:49]
> {"request":{"mbean":"org.apache.activemq.artemis:broker=\"0.0.0.0\"","type":"exec","operation":"listPreparedTransactionDetailsAsHTML()"},"value":"<h3>*** Prepared Transaction Details ***<\/h3><p>No entry.<\/p>","timestamp":1551125316,"status":200}

So, it looks like no one has ever been in a situation that used this functionality.

orpiske:
> So the only other thing here. Is this obviosuly has gone unnoticed for ages. To avoid this in future a test, ensuring its functionality (what ever that is)

I originally thought it was related to listPreparedTransactionDetailsAsJSON or listPreparedTransactionDetailsAsHTML, but it does not seem to be the case.

clebertsuconic:
@orpiske / @michaelandrepearce there are no tests for listPreparedTransactionsDetail on consumers.

I added one, and it turns out it was broken.


thanks a lot @orpiske !!!! you rock!!!

clebertsuconic:
#2567 supersedes this.

when merged, it will close this one.

clebertsuconic:
can you close this? I sent a statement to close it but for some reason it did not work.


already merged with another PR.

PR #2568, opened 2019-02-26
ARTEMIS-2262: Correlate management response messages with the request

k-wall:
This change correlates management response messages with the message request that caused them.  It uses either the Correlation ID pattern (if the request message carried a correlation id), or Message ID pattern (if the request carried a message ID but no correlation id).    If the request carried neither, no correlation is performed.

This change would cause a user-visible change to existing users of the management API over messaging.  It is probably unlikely to be breaking for most valid uses, but a switch might be considered for caution's sake.  Feedback welcome.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
@@ -88,7 +88,6 @@ import org.apache.activemq.artemis.core.server.management.NotificationListener;
 import org.apache.activemq.artemis.core.settings.HierarchicalRepository;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
 import org.apache.activemq.artemis.core.transaction.ResourceManager;
-import org.apache.activemq.artemis.reader.MessageUtil;
 import org.apache.activemq.artemis.spi.core.remoting.Acceptor;
 import org.apache.activemq.artemis.utils.collections.ConcurrentHashSet;
 import org.apache.activemq.artemis.utils.collections.TypedProperties;
@@ -393,9 +392,9 @@ public class ManagementServiceImpl implements ManagementService {
       reply.setType(Message.TEXT_TYPE);
       reply.setReplyTo(message.getReplyTo());
 
-      String correlationId = getCorrelationIdentity(message);
-      if (correlationId != null) {
-         MessageUtil.setJMSCorrelationID(reply, correlationId);
+      Object correlationID = getCorrelationIdentity(message);
+      if (correlationID != null) {
+         reply.setCorrelationID(correlationID);
       }
 
       String resourceName = message.getStringProperty(ManagementHelper.HDR_RESOURCE_NAME);
@@ -796,9 +795,12 @@ public class ManagementServiceImpl implements ManagementService {
     * @param request
     * @return correlation identify
     */
-   private String getCorrelationIdentity(final Message request) {
-      String correlationId = MessageUtil.getJMSCorrelationID(request);
+   private Object getCorrelationIdentity(final Message request) {
+      Object correlationId = request.getCorrelationID();
       if (correlationId == null) {
+         // CoreMessage#getUserId returns UUID, so to implement this part a alternative API that returned object. This part of the
+         // change is a nice to have for my point of view. I suggested it for completeness.  The application could
+         // always supply unique correl ids on the request and achieve the same effect.  I'd be happy to drop this part.
          Object underlying = request.getUserID() != null ? request.getUserID() : request.getStringProperty(NATIVE_MESSAGE_ID);
          correlationId = underlying == null ? null : String.valueOf(underlying);
       }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
@@ -16,6 +16,8 @@
  */
 package org.apache.activemq.artemis.core.server.management.impl;
 
+import static org.apache.activemq.artemis.api.core.FilterConstants.NATIVE_MESSAGE_ID;
+
 import javax.management.InstanceNotFoundException;
 import javax.management.MBeanRegistrationException;
 import javax.management.MBeanServer;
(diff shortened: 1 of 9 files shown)


clebertsuconic:
There's a checkstyle issue on your change. (easy to fix)


Can you provide a use case for why this is needed? is that for enMass?

Just wanted to add some context.


It's probably best if you post the usecase on the JIRA.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
```diff
@@ -390,6 +393,11 @@ public ICoreMessage handleMessage(Message message) throws Exception {
       reply.setType(Message.TEXT_TYPE);
       reply.setReplyTo(message.getReplyTo());
 
+      String correlationId = getCorrelationIdentity(message);
+      if (correlationId != null) {
+         MessageUtil.setJMSCorrelationID(reply, correlationId);
```
michaelandrepearce: Can we avoid using JMS in core. You should set correlation using core api

michaelandrepearce: e.g. what would occur for other protocols that in future we may want to support management service, that may not match JMS. 

It better to introduce correlationId get and set at message api level, e.g. in AMQP it would set in properties, for CORE could make it map to a _AMQ property. etc.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
```diff
@@ -781,5 +789,22 @@ private Object invokeOperation(final String resourceName,
...
+   /**
+    * Correlate management responses using the Correlation ID Pattern, if the request supplied a correlation id,
+    * or fallback to the Message ID Pattern providing the request had a message id.
+
+    * @param request
+    * @return correlation identify
+    */
+   private String getCorrelationIdentity(final Message request)
+   {
+      String correlationId = MessageUtil.getJMSCorrelationID(request);
```
michaelandrepearce: ditto same comment lets avoid using JMS inside core.

k-wall:
> There's a checkstyle issue on your change. (easy to fix)
> 
> Can you provide a use case for why this is needed? is that for enMass?

The context for this request is captured by https://issues.apache.org/jira/browse/ARTEMIS-2261.  In EnMasse's case, we programmatically control the Broker using Management over AMQP.  We currently rely on AMQP's dynamic node feature to create the reply-to queues.  Management responses are currently being lost when the Broker is full (global-max-size exceeded) because messages on the dynamically created nodes are subjected to paging which fails when the Broker is full (`[errorType=ADDRESS_FULL message=AMQ229102: Address "54111434-5686-4dbb-8a94-dd0182dbe7eb" is full.]`)

Switching to non-dynamically created response queue named named in such a way to be excluded from the store (`org.apache.activemq.artemis.core.paging.impl.PagingManagerImpl#getPageStore`) is possible but the life cycling is tricky.   We have many applications collaborating in the management of the Broker and each currently needs a separate response queue in order to avoid cross-talk.  Lifecycling of these response queues ourselves is tricky.  You need to consider cases of abnormal shutdown etc.

If this enhancement existed, we could share a single static management response queue amongst all the applications.  Correlation of requests and responses would avoid the cross-talk between applications or application instances.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
k-wall: @michaelandrepearce thanks for the feedback.   I'm inexperience in the Artemis code base but  think I understand what you mean.  I'll rework the patch once @clebertsuconic has had chance to comment on merits or otherwise of the use-case.

clebertsuconic:
please squash the previous commit on the first..

Also did you look into Michael's consideration on using JMS Layer on server's? We should at least have a discussion about it.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
clebertsuconic: @k-wall I never doubt the usecase.. I just wanted to have it documented for other users when they look on the JIRA / feature request.

one thing I wonder if this warrants documentation changes or not.

k-wall:
> please squash the previous commit on the first..
> 
> Also did you look into Michael's consideration on using JMS Layer on server's? We should at least have a discussion about it.

I'll address @michaelandrepearce's comments.

clebertsuconic:
@k-wall one way would be to add a method setCorrelation on Message. 

There, the default implementation could use the JMS Bits... then later on on AMQP you can make any more specific usage.

k-wall:
@clebertsuconic @michaelandrepearce I'm struggling to find the right way to refactor by moving responsibility for  `#setCorrelationID(Object)` and  `#getCorrelationID()` to Message.  The difficulties I'm having come from the refactoring of `CoreAmqpConverter` and `AmpqCoreConverter` and their interactions with `ServerJMSMessage`.  The fact that the conversion goes through JMS model (Message#JMSCorrelationID() and Message#JMSCorrelationIDAsBytes is making things awkward.  Indeed, I see a couple of places where the conversion side steps the ServerJMSMessage completely.  I don't currently see the right way to proceed.

Could you elaborate a bit more on how you'd approach this change in those areas?

michaelandrepearce:
@k-wall for CoreAmqpConverter and AmpqCoreConverter if the method is now in core, for translating correlation id from core to amqp and back, you would remove the JMS translation mapping that was there, and actually map/translate accross  using the new core api methods.

clebertsuconic:
this seems ready...???

clebertsuconic:
I will merge this... if there's anything to be amended here, it can be a separate commit

k-wall:
Glad to see the feature accepted.  If the coding can be improved along the lines @michaelandrepearce suggested great.  I don't have the depth of code base experience to be able to make the deeper changes efficiently.

The second PR did leave a question in the comments in the code.  You'll probably want to remove this.  See `ManagementServiceImpl#getCorrelationIdentity`.   As I said there, there was a difficultly presented by `CoreMessage#getUserId`.   For my use-case, correlation by *correlation-id* is sufficient so the correlation by *management id* could be removed if you prefer.   If you decide to do that, the support tests ManagementServiceImplTest#testCorrelateResponseByMessageID  and AmqpManagementTest#testCorrelationByMessageID* need to be removed.

PR #2569, opened 2019-02-27
 ARTEMIS-2263 Support queue level auto-delete configuration

michaelandrepearce:
Add ability to configure when creating auto created queues at the queue level
Add support for configuring message count check
Add test cases


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1591,38 +1571,9 @@ public void run() {
...
-            }
-         }
-
-         for (Queue queue : queues) {
-            int consumerCount = queue.getConsumerCount();
-            long messageCount = queue.getMessageCount();
-            boolean autoCreated = queue.isAutoCreated();
-            long consumerRemovedTimestamp =  queue.getConsumerRemovedTimestamp();
-
-            if (!queue.isInternalQueue() && autoCreated && messageCount == 0 && consumerCount == 0 && consumerRemovedTimestamp != -1) {
```
michaelandrepearce: This was duplicated logic with QueueManager so have deduped.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1591,38 +1571,9 @@ public void run() {
 
       @Override
       public void run() {
-         Map<SimpleString, Binding> nameMap = addressManager.getBindings();
-
-         List<Queue> queues = new ArrayList<>();
```
michaelandrepearce: This was duplicate logic with expiry reaper, so simply extracted to a method

clebertsuconic:
+1000 merging it

PR #2571, opened 2019-02-28
Added Javadoc reference to the documentation

tofflos:
I would have preferred the link to point to https://activemq.apache.org/artemis/docs/2.7.0/configuring-transports.html but I'm not super-familiar with Javadoc and there seems to be some extra work involved in getting Javadocs to reflect the Maven project version automatically.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/Configuration.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/Configuration.java
@@ -357,6 +357,7 @@ public interface Configuration {
     * @param uri  the URI of the acceptor
     * @return this
     * @throws Exception in case of Parsing errors on the URI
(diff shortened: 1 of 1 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/config/Configuration.java
```diff
@@ -357,6 +357,7 @@
     * @param uri  the URI of the acceptor
     * @return this
     * @throws Exception in case of Parsing errors on the URI
+    * @see <a href="https://activemq.apache.org/artemis/docs/latest/configuring-transports.html">Configuring the Transport</a>
```
michaelandrepearce: IMO This isnt a good idea, as location of docs url can change in time, and once code is tagged and release we cannot go back and change the links in previously released code.

clebertsuconic: I will usee the github address instead.

PR #2573, opened 2019-03-02
ARTEMIS-2265 Support Federated Queues and Addresses

michaelandrepearce:
Implement Federated Queue (builds on recent consumer priority)
Implement Federated Address (builds on recent queue level auto-delete)
Add Functional Tests
Add Failure Tests (remote and local shutdowns)
Add Documentation


michaelandrepearce:
@clebertsuconic as mentioned, here is the federation stuff, sorry wasnt sooner, just a manic week.

clebertsuconic:
I will review this in detail tomorrow.. leave it with me please.

clebertsuconic:
I looked into this.. and I need a training class on the new feature :) this is great stuff indeed...

the only thing I think this needs are examples.. (that would actually help me understand)... and perhaps a video call with whoever wants to participate (perhaps we could send an open invite).

michaelandrepearce:
Ok i can do a call tonight at 21:00 GMT. Ill send a link here later today.

michaelandrepearce:
https://meet.lync.com/iggroup/michael.pearce/CJG399AB

michaelandrepearce:
@clebertsuconic if you can confirm you saw this

cshannon:
@michaelandrepearce - I won't be able to make the call but thanks for starting work on this feature. This should be quite a useful feature and is something I've been looking at myself recently. 

I've been trying to figure out the best way to migrate over a various 5.x broker setups that use a network of brokers over to Artemis and in some cases clustering didn't quite seem like the right fit for various reasons.  A couple examples are as high latency as you pointed out and also the fact that clustering setup is very coupled to other brokers and requires brokers be configured at start up time and can't be dynamically done at runtime which is a problem.

I haven't taken a look at the full PR yet to see what features exist but ultimately this is something I would want to more or less have feature parity with network of brokers in 5.x to make migration easier.  (things such as duplex connectors, forwarding based on consumer demand, virtual destination consumer demand support (demand based on divert creation in artemis ) etc would be nice to have)

michaelandrepearce:
@cshannon indeed this is entirely where this hopes to go. Atm its single way, as like network brokers achieved first, doing duplex would be next steps.

clebertsuconic:
It's in 4 minutes from now.. are you still available?

michaelandrepearce:
@clebertsuconic im on the call already

michaelandrepearce:
@clebertsuconic examples done, but do not merge still please, as just need to fix up the docs on the examples.

clebertsuconic:
@michaelandrepearce ok.. let me know when this is ready.. this is really good stuff.

gtully:
This looks great. bravo!
From my perspective it needs two capabilities that would make it a "better bridge" than the 5.x networkConnector.

1) limiting forwarding to consumer demand or credit. Using amqp as the transport would facilitate this to some extent. The flow could be bound to the number of local consumers and their rate. Ideally we would only "forward/bridge" what we think we can consume.
This is preferable to flowing and flowing back! Having some variation on this, over and above the consumer priority would be a huge win over 5.x.
Consumer priority is fine when all consumers are fast, however, doing real work in on message will often be slower than forwarding, such that federation becomes a fast consumer. Federation needs to be, in some way limited, to local consumption

2) supporting "at most once forwarding". This is a trade off for sure because something needs to be persisted and typically the application can deal with a duplicate. amqp link recovery would possibly provide an angle here (if it were implemented).
Core bridge has a sliding window of sequence ids which has value. I think this is one for some future consideration. It is less important than 1 :-)

michaelandrepearce:
Was great to have you on the call and get future feature enhancements for this and the feedback. 

We discussed clebert will look at moving some amqp code after 2.7 this would the  enable us to add support for amqp easier in future version.

Once this is merged i think its important we track these not on here but in jira. Ill do this later once merged.

michaelandrepearce:
@clebertsuconic examples docs done. will leave with you now.

cshannon:
+1 for some of Gary's suggestions, they sound like a good idea as future enhancements.  Maybe we can look at making the transport configurable, etc.

I am very willing to help out with enhancements as this feature is very exciting.  I had actually been experimenting with enhancing clustering the past few months trying to add features or tweak it a bit but I have always thrown it away as I was never happy with how it turned out.  Supporting dynamic bridges like 5.x seems like a great solution because often my thoughts were "ok how can I make clustering act more like 5.x " :)

What I found, at least for my use case coming from a 5.x setup, is that a clustering setup makes the most sense from the standpoint of high availability and failover as the brokers can be configured to act as a master/slave setup and being tightly coupled on startup and configuration is fine as you want the brokers to act as a pair, especially in the replication case. This setup isn't really ever dynamic so there's no issue like there would be with establishing bridges dynamically.

So thanks again @michaelandrepearce for getting this started.  This is really going to be a great feature and will I think really help the adoption of people from 5.x who are used to bridging brokers together in the 5.x way and want to continue that without setting up clustering. (Such as myself :) )

cshannon:
@michaelandrepearce - I forgot to ask but is there a reason that this only supports multicast and not anycast at the moment other than this just being the first version and it hasn't been done yet? I am looking at this feature more in detail and supporting anycast should probably be one of the first enhancements.

michaelandrepearce:
@cshannon the issue with anycast right now is we dispatch to one queue only. So not sure how that would work or behaviours we'd want. As such thats why i did multicast only for now. E.g. would you want it to still dispatch to just one queue and we have some form of queue priorty , or would you want a duplicate delivered to the remote address? A bit of discussion is probably needed there, it might be wed like both even....

cshannon:
@michaelandrepearce - So what you are describing seems backwards to me. 

 With multicast being used (ie for topic subscriptions and durable subscriptions) a separate queue is created for each consumer.  So wouldn't that mean we would have duplicates being sent over the bridge as it looks like each queue is replicating individually?  In 5.x we have conduit subscriptions which are on by default so that duplicates are not sent but it can be turned off.  I would think this could be solved by only sending a message once per address and not by queue.

On the other hand with anycast (the typical jms queue case) there is only 1 queue created even though there are multiple consumers.  So this seems like not an issue as we would just be replicating the 1 queue for the address.

michaelandrepearce:
Yes for jms queues i would just use queue federation.




Get Outlook for Android







On Thu, Mar 14, 2019 at 12:45 PM +0000, "Christopher L. Shannon" <notifications@github.com> wrote:












@michaelandrepearce - So what you are describing seems backwards to me.


With multicast being used (ie for topic subscriptions and durable subscriptions) a separate queue is created for each consumer.  So wouldn't that mean we would have duplicates being sent over the bridge as it looks like each queue is replicating individually?  In 5.x we have conduit subscriptions which are on by default so that duplicates are not sent but it can be turned off.  I would think this could be solved by only sending a message once per address and not by queue.


On the other hand with anycast (the typical jms queue case) there is only 1 queue created even though there are multiple consumers.  So this seems like not an issue as we would just be replicating the 1 queue for the address.

michaelandrepearce:
@cshannon so for JMS Topics, yes you look at Address Federation with multicast addresses if you want everyone to get everything everywhere. For JMS Queue you would look just at Queue Federation only. 


Remember the model in Artemis is queues and address's and whilst for JMS it you have for Anycast a 1 to 1 mapping, its actually possible in artemis to have many queues under a single address, just this doesnt map to JMS. 

This is why i only did for now Multicast at address level.

clebertsuconic:
this conversation would be great on the dev list! :)

cshannon:
Fair enough, queue federation would probably work ok for the queue case. 

@clebertsuconic - I am going to spend some time just playing around with things and then can start some more discussions later on the dev list about improvements and enhancements.

michaelandrepearce:
TBH. we need to just create JIRA's for all the enhancements. and then discussion per enhancement in those JIRA's. Action is on me to make those, ill get to that next week.

codybum:
Our transition to Artemis was on hold until federation came along, so this is a very exciting development.  We are making heavy use of 5.x network connectors and I hope this development can help us move forward. 

A few questions:
1.) Can federation (queue or topic) currently be used like a 5.x network connector, where we can dynamically (after broker has started) specific a remote participant via static address?  
2.) Can the described connection be used bidirectionally?

For instance, say you have two independent brokers, broker A and broker B, with public unicast ip address.  Broker A and B are not configured on startup to be aware of each other and neither have access to a multicast network.  Broker A is behind a firewall and can initiate IP connections to Broker B, but Broker B can't initiate connections to Broker A.

Using a 5.x NetworkConnector a bidirectional connection can be made from Broker A to Broker B bridging the queues and topics on both brokers.  Is this or will this be possible using the existing federation implementation?

jerhat:
I second @codybum 's comment, a bidirectional connection (equivalent of `duplex=true` in 5.x) would  help us migrate to Artemis. Is this feature still planned? Is there another way we can connect to a firewall'ed instance?

It might be a good idea to make a note in the [roadmap page](https://activemq.apache.org/activemq-artemis-roadmap.html).

michaelandrepearce:
You're welcome to contribute, please remember this is an apache community project as such driven by contributions.

michaelandrepearce:
See https://github.com/apache/activemq-artemis/pull/2892

michaelandrepearce:
As well look for downstream section in docs for both queue and address

https://activemq.apache.org/components/artemis/documentation/latest/federation-queue.html

https://activemq.apache.org/components/artemis/documentation/latest/federation-address.html

PR #2575, opened 2019-03-05
ARTEMIS-2269 Using karaf.etc for config location

MrEasy:
Karaf config should use predefined etc-dir instead of hard-coded etc

### DIFF
--- a/artemis-features/src/main/resources/org.apache.activemq.artemis.cfg
+++ b/artemis-features/src/main/resources/org.apache.activemq.artemis.cfg
@@ -1,4 +1,4 @@
-config=file:etc/artemis.xml
+config=file:${karaf.etc}/artemis.xml
 name=local
 domain=karaf
 rolePrincipalClass=org.apache.karaf.jaas.boot.principal.RolePrincipal


clebertsuconic:
how karaf.etc is translated by default?

MrEasy:
> how karaf.etc is translated by default?

karaf.etc (along with its variants karaf.data, karaf.home, karaf.base and in latest version also karaf.log) are set as system properties and available as replacements in any config file.

You can verify via
`karaf@root()> system:property | grep karaf.etc`
`karaf.etc=/home/neubauer/apache-karaf-4.2.3/etc`

michaelandrepearce:
So this def needs a default please

MrEasy:
> So this def needs a default please

With def you mean a default like in this construct?
`${karaf.etc:etc}`
This unfortunately does not work in the .cfg file - will resolve to empty string.

It **does** work though in the artemis.xml and that would actually have been my next planned request to make the directories for journal-files etc. independent of the run-directory and put it under karaf.data like so:
`<journal-directory>${karaf.data:.}/artemis/local/journal</journal-directory>`
however this does not to result in the expected directories then. It seems Artemis itself is uilding the sub-dirs depending upon whether the path is absolute or relative, so with karaf.data above being unset, this would result in `data/artemis/local/artemis/local/journal` which is not really what is desired.

michaelandrepearce:
So what i would do next is work out how to make that supported. Im sure it will be possible.

Having a config without a default especially when one exists today is a bit of a show stopper for me, as people maybe relying on that default today, and would be a change in behaviour for them.

clebertsuconic:
the default is specified by karaf, right? that's just like using ant properties.. etc?

I will merge this.

PR #2577, opened 2019-03-11
ARTEMIS-1604 Artemis deadlock using MQTT Protocol

franz1981:
Direct and async deliveries lock QueueImpl::this and
ServerConsumerImpl::this in different order causing deadlock:
has been introduced a deliverLock to prevent both type of delivers
to concurrently happen, making irrelevant the lock ordering.

### DIFF
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionCallback.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionCallback.java
@@ -37,6 +37,11 @@ public class MQTTSessionCallback implements SessionCallback {
    }
 
    @Override
+   public boolean supportsDirectDelivery() {
+      return false;
+   }
+
+   @Override
    public boolean isWritable(ReadyListener callback, Object protocolContext) {
       return connection.isWritable(callback);
    }
--- a/artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
+++ b/artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
@@ -80,6 +80,11 @@ public class StompSession implements SessionCallback {
    }
 
    @Override
+   public boolean supportsDirectDelivery() {
+      return false;
+   }
+
+   @Override
    public boolean isWritable(ReadyListener callback, Object protocolContext) {
       return connection.isWritable(callback);
    }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -253,7 +253,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
    private volatile boolean directDeliver = true;
 
-   private volatile boolean supportsDirectDeliver = true;
+   private volatile boolean supportsDirectDeliver = false;
 
    private AddressSettingsRepositoryListener addressSettingsRepositoryListener;
 
@@ -1078,8 +1078,12 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
                throw ActiveMQMessageBundle.BUNDLE.maxConsumerLimitReachedForQueue(address, name);
             }
 
-            if (!consumer.supportsDirectDelivery()) {
-               this.supportsDirectDeliver = false;
+            if (consumers.isEmpty()) {
+               this.supportsDirectDeliver = consumer.supportsDirectDelivery();
+            } else {
+               if (!consumer.supportsDirectDelivery()) {
+                  this.supportsDirectDeliver = false;
+               }
             }
 
             cancelRedistributor();
@@ -1158,6 +1162,9 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    }
 
    private boolean checkConsumerDirectDeliver() {
+      if (consumers.isEmpty()) {
+         return false;
+      }
       boolean supports = true;
       for (ConsumerHolder consumerCheck : consumers) {
          if (!consumerCheck.consumer.supportsDirectDelivery()) {
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -39,6 +39,7 @@ import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
@@ -209,6 +210,9 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    private final AtomicInteger scheduledRunners = new AtomicInteger(0);
 
    private final Runnable deliverRunner = new DeliverRunner();
+
+   //This lock is used to prevent deadlocks between direct and async deliveries
+   private final ReentrantLock deliverLock = new ReentrantLock();
 
    private volatile boolean depagePending = false;
 
@@ -881,7 +885,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
             return;
          }
 
-         if (supportsDirectDeliver && !directDeliver && direct && System.currentTimeMillis() - lastDirectDeliveryCheck > CHECK_QUEUE_SIZE_PERIOD) {
+         if (direct && supportsDirectDeliver && !directDeliver && System.currentTimeMillis() - lastDirectDeliveryCheck > CHECK_QUEUE_SIZE_PERIOD) {
             if (logger.isTraceEnabled()) {
                logger.trace("Checking to re-enable direct deliver on queue " + this.getName());
             }
@@ -3069,57 +3073,74 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
     * This method delivers the reference on the callers thread - this can give us better latency in the case there is nothing in the queue
     */
    private boolean deliverDirect(final MessageReference ref) {
-      synchronized (this) {
-         if (!supportsDirectDeliver) {
-            return false;
-         }
-         if (paused || !canDispatch() && redistributor == null) {
-            return false;
-         }
-
-         if (checkExpired(ref)) {
-            return true;
-         }
-
-         consumers.reset();
-
-         while (consumers.hasNext() || redistributor != null) {
-
-            ConsumerHolder<? extends Consumer> holder = redistributor == null ? consumers.next() : redistributor;
-            Consumer consumer = holder.consumer;
-
-            final SimpleString groupID = extractGroupID(ref);
-            Consumer groupConsumer = getGroupConsumer(groupID);
-
-            if (groupConsumer != null) {
-               consumer = groupConsumer;
+      //The order to enter the deliverLock re QueueImpl::this lock is very important:
+      //- acquire deliverLock::lock
+      //- acquire QueueImpl::this lock
+      //DeliverRunner::run is doing the same to avoid deadlocks.
+      //Without deliverLock, a directDeliver happening while a DeliverRunner::run
+      //could cause a deadlock.
+      //Both DeliverRunner::run and deliverDirect could trigger a ServerConsumerImpl::individualAcknowledge:
+      //- deliverDirect first acquire QueueImpl::this, then ServerConsumerImpl::this
+      //- DeliverRunner::run first acquire ServerConsumerImpl::this then QueueImpl::this
+      if (!deliverLock.tryLock()) {
+         logger.tracef("Cannot perform a directDelivery because there is a running async deliver");
+         return false;
+      }
+      try {
+         synchronized (this) {
+            if (!supportsDirectDeliver) {
+               return false;
+            }
+            if (paused || !canDispatch() && redistributor == null) {
+               return false;
             }
 
-            HandleStatus status = handle(ref, consumer);
-
-            if (status == HandleStatus.HANDLED) {
-
-               if (redistributor == null) {
-                  handleMessageGroup(ref, consumer, groupConsumer, groupID);
-               }
-
-               messagesAdded.incrementAndGet();
-
-               deliveriesInTransit.countUp();
-               proceedDeliver(consumer, ref);
-               consumers.reset();
+            if (checkExpired(ref)) {
                return true;
             }
 
-            if (redistributor != null || groupConsumer != null) {
-               break;
-            }
-         }
+            consumers.reset();
 
-         if (logger.isTraceEnabled()) {
-            logger.tracef("Queue " + getName() + " is out of direct delivery as no consumers handled a delivery");
+            while (consumers.hasNext() || redistributor != null) {
+
+               ConsumerHolder<? extends Consumer> holder = redistributor == null ? consumers.next() : redistributor;
+               Consumer consumer = holder.consumer;
+
+               final SimpleString groupID = extractGroupID(ref);
+               Consumer groupConsumer = getGroupConsumer(groupID);
+
+               if (groupConsumer != null) {
+                  consumer = groupConsumer;
+               }
+
+               HandleStatus status = handle(ref, consumer);
+
+               if (status == HandleStatus.HANDLED) {
+
+                  if (redistributor == null) {
+                     handleMessageGroup(ref, consumer, groupConsumer, groupID);
+                  }
+
+                  messagesAdded.incrementAndGet();
+
+                  deliveriesInTransit.countUp();
+                  proceedDeliver(consumer, ref);
+                  consumers.reset();
+                  return true;
+               }
+
+               if (redistributor != null || groupConsumer != null) {
+                  break;
+               }
+            }
+
+            if (logger.isTraceEnabled()) {
+               logger.tracef("Queue " + getName() + " is out of direct delivery as no consumers handled a delivery");
+            }
+            return false;
          }
-         return false;
+      } finally {
+         deliverLock.unlock();
       }
    }
 
@@ -3464,8 +3485,11 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
             enterCritical(CRITICAL_DELIVER);
             boolean needCheckDepage = false;
             try {
-               synchronized (QueueImpl.this.deliverRunner) {
+               deliverLock.lock();
+               try {
                   needCheckDepage = deliver();
+               } finally {
+                  deliverLock.unlock();
                }
             } finally {
                leaveCritical(CRITICAL_DELIVER);


franz1981:
I'm working to add a reproducer (not easy TBH, but possible and useful), but the referenced issue 
already report a stack trace that make easy to understand what is happening.
IMHO Core and AMQP are more likely to be exposed to the same issue, but I need to check it first to be sure.

franz1981:
@clebertsuconic I have found another issue, probably the most important one!!
Both Stomp and MQTT are supporting direct delivery, so I have changed that one...
on qq, Stomp will support direct delivery or not?

franz1981:
@clebertsuconic So it is fine that I have overridden supportDirectDelivery in order to return false?

michaelandrepearce:
Surprised that disabling direct here is needed as using netty mqtt protocol

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -210,6 +211,9 @@
 
    private final Runnable deliverRunner = new DeliverRunner();
 
+   //This lock is used to prevent deadlocks between direct and async deliveries
+   private final ReentrantLock deliverLock = new ReentrantLock();
```
michaelandrepearce: Ouch really??? Seriouslly!!

franz1981: That was happening because MQTT callback wasn't implemented to support direct delivery, but was wrongly  reporting to use it. 
Anyway using this lock will prevent direct deliveries to happen concurrently with an async deliveries, that is a way to enforce a correct behaviour anyway.
Not sure if thanks to this MQTT could enable direct deliveries, probably not.

michaelandrepearce: Point is we now have two different lock approaches in this class. Recipe for disaster

franz1981: Not sure: I have made explicit an implicit requirement. And now the order of locking is ensured just into QueueImpl , while before it was enforced in different classes (callbacks, transactions and queue).

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -249,7 +253,7 @@
 
    private volatile boolean directDeliver = true;
 
-   private volatile boolean supportsDirectDeliver = true;
+   private volatile boolean supportsDirectDeliver = false;
```
michaelandrepearce: this should be default true.

franz1981:
Re the performance impacts I cannot see any evident ones, because tryLock will cost just a volatile load + a compareAndSwap (bad we have already many of them all around). But we can measure it and see if there are other not evident changes that couple impact..

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
franz1981: @clebertsuconic has proposed this change, to avoid having a small window in which a protocol that isn't suportting directDelivery could attempt to do one due to a race while adding its consumers

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3069,57 +3080,74 @@ private void move(final Transaction originalTX,
...
     */
    private boolean deliverDirect(final MessageReference ref) {
-      synchronized (this) {
-         if (!supportsDirectDeliver) {
-            return false;
-         }
-         if (paused || !canDispatch() && redistributor == null) {
-            return false;
-         }
+      //The order to enter the deliverLock re QueueImpl::this lock is very important:
```
michaelandrepearce: rather than this, why not have the original method renamed and existing method with the lock delegating then to the renamed (but original) method. This way keeps the lock cleaner.

franz1981: I'm not sure, I don't want the original method to be called without using the proper lock order and having them in one place will help to avoid misusing: that's personal...
If not, we can split the method and add an assertion to ensure that the deliveryLock needs to be held by the current Thread too: it will avoid future misuses.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelandrepearce: fair enough.

franz1981: Yep, I see that I this PR the only relevant commit is the one that force MQTT and Stomp to avoid directDelivery, the rest are just improvements to avoid bad behaviours to happen..

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelandrepearce: Re: mqtt i don't see why it can't have direct deliveries if CORE can, either we have a locking issue in code or we don't.

michaelandrepearce:
I think having reproducer code here is very important. You're fixing something down in CORE layer, but if there is a bug there, it would affect all protocols. And then other changes shouldn't be needed.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelandrepearce: Its private, its no different to someone calling deliver() not within the runner.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelandrepearce: So i don't get why MQTT can have direct delivery here still, it used to, all you've done is found a bug it seems down in Core queue, which if thats the case, it will be a bug for all protocols, not just MQTT or STOMP.

franz1981:
I have already run the whole CI having a stable test outcomes (and slightly better 
 results but I know that is mostly luck).
I agree about the reproducers in general, but I'm not sure about mqtt being easy for direct delivery: according to Cleb it shouldn't be ready AFAIK, but I could be wrong. I need to dig into the mqtt code to check why.

artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
```diff
@@ -79,6 +79,11 @@
       this.consumerCredits = ConfigurationHelper.getIntProperty(TransportConstants.STOMP_CONSUMERS_CREDIT, TransportConstants.STOMP_DEFAULT_CONSUMERS_CREDIT, connection.getAcceptorUsed().getConfiguration());
    }
 
+   @Override
+   public boolean supportsDirectDelivery() {
```
michaelandrepearce: if its a deadlock issue in queue, this shouldn't be needed. We should avoid using a sledge hammer by disabling features. fix the deadlock not just disable stuff.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionCallback.java
```diff
@@ -36,6 +36,11 @@ public MQTTSessionCallback(MQTTSession session, MQTTConnection connection) throw
       this.connection = connection;
    }
 
+   @Override
+   public boolean supportsDirectDelivery() {
```
michaelandrepearce: if its a deadlock issue in queue, this shouldn't be needed. We should avoid using a sledge hammer by disabling features. fix the deadlock not just disable stuff.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
franz1981: In theory it could happen on core too but it depends on "unlucky" timings: for core anyway I need to check the ordering of locking ServerConsumerImpl.this and QueueImpl.this to be sure of it.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionCallback.java
franz1981: According to @clebertsuconic is not that way: there are protocols that doesn't support direct deliver and I seems that mqtt is one of them.

artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
franz1981: Same of mqtt

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionCallback.java
michaelandrepearce: Still the point is you're fixing a deadlock in this PR, just fix the deadlock, don't make other changes. If we have deadlock in QueueImpl, then its irregardless of these other changes.

artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
michaelandrepearce: and same response.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionCallback.java
franz1981: I could split the pr in several ones but they are tightly related, but yes, can be done for sure

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
clebertsuconic: I understand this is settled, but let me explain with my words, to make sure you understand.. and I'm not wrong.

The reason we disable direct deliver is certain protocols won't have the locking support to deal with direct delivery, and they may dead lock.

So, with MQTT what I believe it could happen is.. supports= true, now you add a consumer, it becomes false.

If an user, out of an anti-pattern creates and deletes consumers rapidly while sending messages this could lead to races on delivering.

This is just for that kind of anti-pattern.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionCallback.java
clebertsuconic: I don't think MQTT was supposed to have direct delivery support. There's conversion on messages.

michaelandrepearce: @clebertsuconic i don't see why that impacts if it can direct deliver or not, direct deliver in my understanding is if it can use the producer thread, to avoid needing to context switch for perf, as long as transform is non blocking

I iterate though this bug if i understand the change and franz, is theres deadlock in queueimpl. NOT the protocols.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelandrepearce: so the deadlock isn't in MQTT from my understanding, its in QueueImpl. Its just MQTT was the protocol that the person using was on. The bug is there not in MQTT.

clebertsuconic: My understanding was using direct delivery on a protocol that can't handle it.

clebertsuconic: I'm not sure though.. this issue was reported Jan last year.. I thought I had worked on it after then. I didn't know it was still an issue until Franz worked on it today.


I didn't close the original JIRA before because the user never reported what version he was working on... 

So, I guess it's important to find a test replicating the issue, as we may be fixing something that was already fixed.

michaelandrepearce: @clebertsuconic if thats the case, then all the changes franz has done in QueueImpl are NOT needed. But looking through it does seem the issue IS in QueueImpl not MQTT.

michaelandrepearce: > So, I guess it's important to find a test replicating the issue, as we may be fixing something that was already fixed.

Agreed that is very important!!

FrederikP: > I'm not sure though.. this issue was reported Jan last year.. I thought I had worked on it after then. I didn't know it was still an issue until Franz worked on it today.
> 
> I didn't close the original JIRA before because the user never reported what version he was working on...
> 
> So, I guess it's important to find a test replicating the issue, as we may be fixing something that was already fixed.

Hi,
I'm the guy who uploaded the thread dump to the jira issue last friday. Just want to make sure it's clear that we experienced this dead lock with version 2.6.4. Maybe I should've opened a new issue because it might have nothing to do with mqtt and it might be a different issue than the one the original reporter encountered a year ago with an older version.

franz1981: @michaelandrepearce @clebertsuconic 

According to @FrederikP it is an actual issue and with MQTT + debugging I've been able to reproduce it with ease: I just need to do the same using a test (and probably ByteMan or ByteBuddy to make it happen).
Re the why it can be fixed is not ease: according to what Clebert said I believe him that there are protocols that do not support direct delivery and MQTT is reporting (on the code) to support it. 
The simplest fix here is contained in one of the commit of this PR: just making it to report that it doesn't support direct delivery.

Thanks to the way MQTT works I have found that direct deliveries and async deliveries could happen concurrently (and TBH it doesn't make sense, from many point of views) AND that if the locks of ServerConsumerImpl and QueueImpl are not acquired in the correct order it could lead to a deadlock. 
Given that async and direct deliveries shouldn't happen concurrently (are mutually exclusive, but this wasn't never enforced on code level!), fixing it will make irrelevent the order of locks.

Now I'm checking if Core is exposed to the same issue.

michaelandrepearce: I think we need to have a reproducer test case here before making any assumptions or move forwards with reviewing.

clebertsuconic:
@franz1981  / @michaelandrepearce let me check if we can make it single lock. I will do some work on it.

franz1981:
@clebertsuconic This implementation is already using a single lock: in the origianl version we were using the DeliveryRunner instance instrinsic lock. I've not verified yet if the issue is hitting core anyway...

michaelandrepearce:
@clebertsuconic im less concerned about detail right now, im more concerned with we are we making/ needing changes. Either the issue is in MQTT or the issue is in QueueImpl, and we should only be making changes in that place NOT both. I think i understand the issue after looking at code changes, and think the issues is just in QueueImpl. But to be sure we need a test right.

clebertsuconic:
@franz1981 did you confirm the issue still hits the user? TL-DR; if you mentioned that before?I picked this issue an year ago when it was opened.. and there was little detail on what version they used.. and we had changes previously. So, I always assumed the user was using an older version.

I'm missing some context on why you picked this one up. Did you have other user internally at work complaining about this?

clebertsuconic:
just realized @FrederikP comments... 

do you have an idea on what could be a test reproducing your behavior? Are you closing / opening consumers too fast or anything like that? it's Stomp, right?

franz1981:
@clebertsuconic I see that with MQTT the issue is real and present in the last version on master, but I'm able to reproduce it just using debug ATM: the thing is that MQTT is wrongly reporting to be able to perform direct deliveries and if I override the callback to avoid it, everything should work fine.
The other commits added on this PR are just to enforce a correct behaviour of direct deliveries

FrederikP:
> just realized @FrederikP comments...
> 
> do you have an idea on what could be a test reproducing your behavior? Are you closing / opening consumers too fast or anything like that? it's Stomp, right?

nope, no idea how to reproduce it. But @franz1981 already seems to be able to reproduce it locally using debug as he stated in his comment.

I'm still going to summarize the usage of our artemis cluster because it might help: We took a three machine cluster online on friday and a couple of hours later one of the nodes ran into this dead lock. There are many mqtt and some stomp connections sending data and mostly stomp and some jms connections retrieving the data. The sending mqtt connections are under our control and should behave well. One of the stomp senders that is not under our control sends one item and then disconnects and opens a new connection for another one and so on (we are trying to get him to improve this). This actually might have triggered this dead lock, but I won't be able to verify.

After the node crashed we had to roll back to the old activemq (non-artemis) for production. The issue didn't come up on our staging environment.

If you have any other specific questions about our setup, let me know.

michaelandrepearce:
@franz1981 if you revert the protocol changes, and ONLY have the QueueImpl fix, is the issue resolved? e.g. no deadlock.

clebertsuconic:
The real issue is, as I said, MQTT was not supposed to direct deliver:


It's holding a lock eventually back to the queue here:


        at org.apache.activemq.artemis.core.server.impl.ServerConsumerImpl.individualAcknowledge(ServerConsumerImpl.java:886)
        -  blocked on org.apache.activemq.artemis.core.server.impl.ServerConsumerImpl@115d5ec3
        at org.apache.activemq.artemis.core.server.impl.ServerSessionImpl.individualAcknowledge(ServerSessionImpl.java:825)
        at org.apache.activemq.artemis.core.protocol.mqtt.MQTTPublishManager.sendMessage(MQTTPublishManager.java:127)
        at org.apache.activemq.artemis.core.protocol.mqtt.MQTTSessionCallback.sendMessage(MQTTSessionCallback.java:50)
        at org.apache.activemq.artemis.core.server.impl.ServerConsumerImpl.deliverStandardMessage(ServerConsumerImpl.java:1089)
        at org.apache.activemq.artemis.core.server.impl.ServerConsumerImpl.proceedDeliver(ServerConsumerImpl.java:460)
        at org.apache.activemq.artemis.core.server.impl.QueueImpl.proceedDeliver(QueueImpl.java:2757)
        at org.apache.activemq.artemis.core.server.impl.QueueImpl.deliverDirect(QueueImpl.java:2742)
        -  locked org.apache.activemq.artemis.core.server.impl.QueueImpl@4309e0af
        at org.apache.activemq.artemis.core.server.impl.QueueImpl.addTail(QueueImpl.java:683)




which would certainly dead lock.


@michaelandrepearce  this is similar to how AMQP would deadlock when we were using the older proton model. MQTT shouldn't do direct deliver.

artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
clebertsuconic: the issue is that the protocol is holding locks on both places (consumer and queue), hence it cannot support direct deliver.

clebertsuconic:
@FrederikP a work around, set directDeliver=false on the acceptors...

clebertsuconic:
@michaelandrepearce I want to merge this.. I can't produce a test for this.. and I know for a fact MQTT was not supposed to direct deliver...

I'm adding a test to validate direct-deliver = false on MQTT as part of this PR.

FrederikP:
> @FrederikP a work around, set directDeliver=false on the acceptors...

Thanks for the suggestion. We tried that and ran into another deadlock. To avoid confusion I opened an issue here: https://issues.apache.org/jira/browse/ARTEMIS-2276

PR #2579, opened 2019-03-12
ARTEMIS-2273 Adding Audit Log

gaohoward:
The Audit log allows user to log some important actions,
such as ones performed via management APIs or clients,
like queue management, sending messages, etc.
The log tries to record who (the user if any) doing what
(like deleting a queue) with arguments (if any) and timestamps.

By default the audit log is disabled. Through configuration can
be easily turned on.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
@@ -88,6 +88,7 @@ import org.apache.activemq.artemis.core.transaction.Transaction.State;
 import org.apache.activemq.artemis.core.transaction.TransactionOperationAbstract;
 import org.apache.activemq.artemis.core.transaction.TransactionPropertyIndexes;
 import org.apache.activemq.artemis.core.transaction.impl.TransactionImpl;
+import org.apache.activemq.artemis.logs.AuditLogger;
 import org.apache.activemq.artemis.spi.core.protocol.RemotingConnection;
 import org.apache.activemq.artemis.spi.core.protocol.SessionCallback;
 import org.apache.activemq.artemis.utils.CompositeAddress;
@@ -486,6 +487,9 @@ public class ServerSessionImpl implements ServerSession, FailureListener {
                                         final boolean browseOnly,
                                         final boolean supportLargeMessage,
                                         final Integer credits) throws Exception {
+      if (AuditLogger.isEnabled()) {
+         AuditLogger.createCoreConsumer(this, getUsername(), consumerID, queueName, filterString, priority, browseOnly, supportLargeMessage, credits);
+      }
       final SimpleString unPrefixedQueueName = removePrefix(queueName);
 
       Binding binding = postOffice.getBinding(unPrefixedQueueName);
@@ -632,6 +636,11 @@ public class ServerSessionImpl implements ServerSession, FailureListener {
                             final long autoDeleteDelay,
                             final long autoDeleteMessageCount,
                             final boolean autoCreated) throws Exception {
+      if (AuditLogger.isEnabled()) {
+         AuditLogger.createQueue(this, getUsername(), addressInfo, name, filterString, temporary, durable, maxConsumers, purgeOnNoConsumers,
+                  exclusive, groupRebalance, groupBuckets, lastValue, lastValueKey, nonDestructive, consumersBeforeDispatch,
+                  delayBeforeDispatch, autoDelete, autoDeleteDelay, autoDeleteMessageCount, autoCreated);
+      }
       final SimpleString unPrefixedName = removePrefix(name);
 
       AddressInfo art = getAddressAndRoutingType(addressInfo);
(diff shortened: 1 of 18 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -1516,6 +1541,9 @@ public synchronized RoutingStatus send(Transaction tx,
                                           final boolean direct,
                                           boolean noAutoCreateQueue,
                                           RoutingContext routingContext) throws Exception {
+      if (AuditLogger.isEnabled()) {
+         AuditLogger.sendMessage(this, getUsername(), tx, messageParameter, direct, noAutoCreateQueue, routingContext);
```
clebertsuconic: Shouldn't this be done on PostOfficeImpl::send?

Then the same would apply to all the protocols.


Currently your implementation would ignore anything coming from Federation, MQTT AMQP or Stomp, since those will call PostOfficeImpl::send directly.

michaelandrepearce: If there is audit on this. I think there needs to be some audit level seperation. E.g. users may want to audit management actions but not want to audit log every message going through the broker

michaelandrepearce: I am -1 it for this reason to avoid this just being merged without this being addressed.

michaelandrepearce: Just to add i like the idea, just think there needs to be finer control on what to audit especially hot paths like send message and other bits.

michaelandrepearce: Please see:  org.apache.activemq.artemis.core.server.plugin.impl.LoggingActiveMQServerPlugin

It be good to extend this, rather than another implementation.

gaohoward: > Shouldn't this be done on PostOfficeImpl::send?
> 
> Then the same would apply to all the protocols.
> 
> Currently your implementation would ignore anything coming from Federation, MQTT AMQP or Stomp, since those will call PostOfficeImpl::send directly.

The send() is auditing the message coming into the broker. So I think it can cover other protocols. right? (currently the audit doesn't record the messages going out but it logs the consumer creations).

gaohoward: Hi @michaelandrepearce Just had a look at the plugin class my impression is it's a big too complicated for audit log. But let me think about it more. Also I agree the idea to fine control the logging on 'hot' paths so to give user a choice to turn off them particularly.
Thanks

michaelandrepearce: Whats complicated about it? You define plugin in broker xml and voila... 

The point being this is exactly the use case for plugins

gaohoward: the log format is different. and the logging plugin serves a different purpose that auditing. (actually in amq5 they are totally different impls). and it brings in extra config.
for hot path (actually so far there is only one sending message) I can use log level to easily control it.
So I'd for the moment don't touch the logging plugin. 
wdyt?

andytaylor:
I think as long as its a users choice its fine to have it on or off(default), if the user has it on they accept the hit.

michaelandrepearce:
@andytaylor point here is to make it more usable. E.g. audit is important for lots of people but what hit they take and need to audit is different. Especially hot path.

gaohoward:
I updated the branch to allow user to control the logging of hot path (i.e. sending message here).
If the logger level is INFO, no hot path logging. if set to DEBUG the hot path is enabled.

clebertsuconic:
I don’t think using debug versus info is a good choice on this case. 


You should make a new logger for messaging logger.  


Or use broker plugin and enable / disable based on that plugin.

michaelandrepearce:
Agree with clebert here. 

I wasn't meaning just seperate send message, it was just one example on why you want to toggle what gets audit. You want to give flexibility to toggle different areas just like the plugin is doing.

Still -1 atm only to avoid this being merged until bits are addressed, im 100% +1 the idea and where its going.

clebertsuconic:
I don’t think it’s feasible to control every single point.  I think we should separate messages from everything else. 


Otherwise too many options on confit also make it difficult to admin.  


And that’s the same even if you used plugin.

michaelandrepearce:
Agreed. I think the level the plugin has is about the right level. So if that was similar here that be best.

gaohoward:
OK guys. I'll make the hot path logging as a separate logger.  
Maybe we can let the basics working and improve later.
thanks guys!

gaohoward:
Just updated the branch. Now we have 2 loggers. one is base logger and the other is message logger (for hot path logging i.e. sending messages)

michaelandrepearce:
Turning my frown upside down. +1 thanks for addressing.

michaelandrepearce:
@clebertsuconic @gaohoward since this change, i can still build from command line, but its causing a massive issue within my IDE as its causing generated classes in the parent target folder, which don't obviously then have access to logging libs (as declared in submodule), and thus meaning i cannot debug unit tests within my IDE. 

If i revert this change its all fine.

Can you have a look at please, as its kinda made working and debugging really hard now, as i keep having to revert this commit to work locally within the IDE. Im using IntelliJ fyi,

michaelandrepearce:
fyi, i think its something to do with javadoc plugin causing it, other modules that declare logging have a profile for javadoc that seems to be the thing stopping it from causing issue for those loggers

michaelandrepearce:
Have sent possible fix on PR 2654

clebertsuconic:
@michaelandrepearce I am working with IDEA fine.

Some times I have to git clean -xdf

And rebuild my entire workspace..

But nothing related to this change.

PR #2581, opened 2019-03-13
ARTEMIS-2274 sending a close to JournalBufferSize message cause Critical IO Error

CNNJYB:
we set the values of MinLargeMessageSize and JournalBufferSize to 1Mb and set JournalFileSize value to 10Mb, then we send a message less than 1Mb but close to 1Mb. We have the following errors :
java.lang.IllegalStateException: Can't write records bigger than the bufferSize(1048576) on the journal: java.lang.IllegalStateException: Can't write records bigger than the bufferSize(1048576) on the journal
when ActiveMQBuffer encode, which will write some specified fields, such as fileID, recordType, compactCount, etc. This result in buffer greater than 1Mb. TimedBuffer checksize and find it`s bigger than bufferSize before write buffer.


clebertsuconic:
You shouldn't use min-large-message-size that close to the buffer size. that's a recipe for a disaster?

the PR should be sent for fixes. this is kind of a bug report?

clebertsuconic:
I have ammended your test, and added the fix on top of this.

PR #2582, opened 2019-03-13
Revert "ARTEMIS-2240 ActiveMQThreadPoolExecutor should use LinkedTransferQueue"

franz1981:
This reverts commit ea29483449eb74d4bc8ee703e1161c0d67dc77b4

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ActiveMQThreadPoolExecutor.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ActiveMQThreadPoolExecutor.java
@@ -16,12 +16,11 @@
  */
 package org.apache.activemq.artemis.utils;
 
-import java.util.Objects;
-import java.util.concurrent.LinkedTransferQueue;
-import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 /*
  * ActiveMQThreadPoolExecutor: a special ThreadPoolExecutor that combines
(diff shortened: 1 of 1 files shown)


franz1981:
Nudge nudge @michaelandrepearce I have performed some single-producer single-consumer tests with 100 bytes messages and no persistence (core protocol) and I have found several interesting things:

1. the new thread pool theoretically should scale way better then the original: there are no more locks associated to the tail/head of the linked blocking task queue 
2. the new thread pool try to reuse dormient and existing threads, but it means that with few producers/consumers then avaialble threads the cost to wakeup them will add a huge latency hit (the percentiles show it for small burst) :(

This is a regression if compared with other commits:
- 7dfa0fe7f43012f2f4ee5715049878019fe0bef7
- 71d549434591715c1f02359000e424e4d7dfbfda
- https://github.com/apache/activemq-artemis/tree/2.6.4

I suggest anyone that has enough cores to perform tests on these commits and check the latencies/throughput, because although on my box the throughput is not that different, the performance regression is evident on latency side.
Shame on me that I have trusted that a lock-free q would have performed better in any case: the cost to scale better is just too high and I think is better to work after the release to improve the thread pool.

michaelandrepearce:
The commits you reference are other features a bit confused why your referencing them

franz1981:
@michaelandrepearce 
I have added them for a specific reason :)
The 2.6.4 tag was my original baseline, while 7dfa0fe7f43012f2f4ee5715049878019fe0bef7 was my original bet about the reason of the regression and 71d549434591715c1f02359000e424e4d7dfbfda was the commit right before it (that I have used to validate that  7dfa0fe7f43012f2f4ee5715049878019fe0bef7 wasn't guilty of anything).

I have already tried the commit that has introduced the new thread pool too and reverting it and master + this PR too: my box always report the new thread pool as the source of the latency regression.

michaelandrepearce:
So where is this putting us back to. My original ? Or the version we had perf issue with?

franz1981:
I've reverted the change in order to use the first revert you did on the thread pool: I know that is twisted

clebertsuconic:
please, use the complete commit on the this Reverts to..

3 reasons: the report I wrote only takes full commit IDs. Second: I don't want to change it as I don't have time;)

3rd.. pretty please?

michaelandrepearce:
Ok merge it.

clebertsuconic:
I will merge it, but referencing the whole commit ID... leave it with me

franz1981:
Guys if you have better machines then mine don't you want to give a shot?

franz1981:
@clebertsuconic Fixed the commit hash message

PR #2583, opened 2019-03-13
ARTEMIS-2229 turns out a valid bug. So restore the fix commit.

clebertsuconic:
this was reverted previously on commit f4436a9f72afeb8117ecd6e377df8198eeec9aa3.

It seems it is ok after further changes into FQQN, so this is being reapplied


clebertsuconic:
@jbertram  this was previously reverted. I'm reapplying it here, but please if you could carefully evaluate these changes? I'm concerned especially about the ProtonHandler change around addressing as it seems to be doing a bit more than FQQN. If you could evaluate it please?

jbertram:
@clebertsuconic this looks fine to me.

clebertsuconic:
Ok.. I'm running tests...

@howardgao I should merge this tomorrow morning.

PR #2587, opened 2019-03-20
ARTEMIS-2280 Add new test for moving message to FQQN

zschwarz:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/QueueControlTest.java
```diff
@@ -1329,6 +1329,64 @@ public void testMoveMessage() throws Exception {
...
+    *    <address name="ErrorQueue">
+    *             <anycast>
+    *                <queue name="ErrorQueue" />
+    *             </anycast>
+    *          </address>
+    *          <address name="parent.addr.1">
+    *             <multicast>
+    *                <queue name="child.queue.1" />
+    *                <queue name="child.queue.2" />
+    *                <queue name="child.queue.3" />
```
gaohoward: Base on the test, this queue is never created/used. Should be removed.

gaohoward: The rest looks good to me.

zschwarz: Thank you for review, I removed the unused queue from comment.

PR #2590, opened 2019-03-23
ARTEMIS-2282 QueueImpl::addTail of a management message could throw NPE

franz1981:
Performing direct deliveries of management messages could enter
a code path on QueueImpl::addTail with a NULL pageIterator: performing
a null check will avoid it to throw NPE.


michaelandrepearce:
Test case

franz1981:
Sure, just need to understand how: the mechanics isn't easy to replicate looking at the logic.
The guilty code path can be reached only if the last delivery check has happened more then CHECK_QUEUE_SIZE_PERIOD ie 1 second after a failed direct deliver (ie directDeliver = false).
Let me think how to make it happen :+1:

franz1981:
@michaelandrepearce @k-wall I've added a test (not one of my most brilliant one I admit): it is always reproducing the issue, without needing any weird concurrent/non-deterministic behaviours :+1:

michaelandrepearce:
Thanks @franz1981 lgtm now

PR #2591, opened 2019-03-26
ARTEMIS-2283 Bad WARN message "AMQ222211: Storage is back to stable now"

franz1981:
LocalMonitor::under on PagingManagerImpl won't log anymore with a
warning message if the producers got unblocked


franz1981:
@ehsavoie I've changed my mind: I have noticed that we already have some test to cover things like log messages to I think tht is the right place to put them :+1:

jbertram:
@franz1981, there's a checkstyle violation:

```
[ERROR] src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingManagerImpl.java:[20,8] (imports) UnusedImports: Unused import - java.util.Iterator.
```

franz1981:
Argh!

PR #2592, opened 2019-03-26
ARTEMIS-2285 Sorting by user column in connections view not working

sebthom:
The value returned by ConnectionsView.getField(con, "users") currently returns a Set which does not implement the Comparable interface thus ActiveMQAbstractView#getComparator() always returns a comparator which evaluates to 0.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/ConnectionView.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/ConnectionView.java
@@ -18,9 +18,9 @@ package org.apache.activemq.artemis.core.management.impl.view;
 
 import javax.json.JsonObjectBuilder;
 import java.util.Date;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.TreeSet;
 
 import org.apache.activemq.artemis.api.core.client.ClientSession;
 import org.apache.activemq.artemis.core.management.impl.view.predicate.ConnectionFilterPredicate;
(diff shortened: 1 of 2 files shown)


gaohoward:
looking good. Could you add a unit test?

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/ConnectionView.java
```diff
@@ -83,12 +83,12 @@ public Object getField(RemotingConnection connection, String fieldName) {
...
             return connection.getRemoteAddress();
          case "users":
-            Set<String> users = new HashSet<>();
+            Set<String> users = new TreeSet<>();
             for (ServerSession session : sessions) {
                String username = session.getUsername() == null ? "" : session.getUsername();
                users.add(username);
             }
-            return users;
+            return StringUtil.joinStringList(users, ",");
```
michaelandrepearce: Whilst i understand treeset to give ordering by insert, the return shouldn't change the type, e.g just return users still. Or im missing something.... i guess a test may help to describe/show why its needed.

sebthom:
I will try to provide a test case. stay tuned :)

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/ConnectionView.java
sebthom: The returned value is a set of users of a single connection. to sort the connections by their user(s), these lists need to be comparable with each other. however treeset does not implement the comparable interface, so you cannot sort multiple treesets by there values. thus the PR now returns a comparable/sortable string representation of the set. does that make sense?

sebthom: The returned value is a set of users of a single connection. to sort the connections by their user(s), these lists need to be comparable with each other. however treeset does not implement the comparable interface, so you cannot sort multiple treesets by there values. thus the PR now returns a comparable/sortable string representation of the set. does that make sense?

sebthom:
@gaohoward @michaelandrepearce  I added a test as requested.

michaelandrepearce:
@sebthom thanks for this great contributions. merged.

PR #2593, opened 2019-03-27
ARTEMIS-2264 Added test that receive all messages instead of remove

franz1981:
It improves the test coverage of the same use case ie "PurgeOnNoConsumers prevent removal of messages with replication", but receiving all the messages instead of removing them.


michaelandrepearce:
Lgtm

PR #2597, opened 2019-03-29
ARTEMIS-2291 ClassCastException on JMSBridgeImpl::stop leaks bridge resources

franz1981:


franz1981:
Not sure if is possible to provide a test for this, but I'll try :+1:

PR #2598, opened 2019-03-29
ARTEMIS-2290 JMSBridgeImpl::stop is failing when called from FailureHandler

franz1981:


franz1981:
I'm writing a test for it :+1:

michaelandrepearce:
> I'm writing a test for it 

:)

franz1981:
@clebertsuconic @michaelandrepearce I need to understand a thing first: I see that the bridge stop could happen from any thread (FailureHandler or any caller thread): JMSBridgeImpl::stop is supposed to wait a completed racing stop happening from another thread?

michaelandrepearce:
@franz1981 i honestly dont know. We had some early issues with the jmsbridge so ended up where we needed to bridge different jms providers with a remote app made with spring integration.

franz1981:
I understand this and I would say that logic and scalability of JMSBridge could be improved for sure (it uses 3 threads for bridge!!)...
Not sure I would do it in this PR to avoid introducing too many changes...

franz1981:
I will add a test fo it in a separate PR: it seems not easy to be reproduced :+1:

PR #2602, opened 2019-04-02
ARTEMIS-2292 - Fix queue level auto-delete for topic subscription queue

michaelandrepearce:


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1582,7 +1582,7 @@ public void run() {
       public void run() {
          for (Queue queue : getLocalQueues()) {
             if (!queue.isInternalQueue() && QueueManagerImpl.isAutoDelete(queue) && QueueManagerImpl.consumerCountCheck(queue) && QueueManagerImpl.delayCheck(queue) && QueueManagerImpl.messageCountCheck(queue)) {
-               QueueManagerImpl.deleteAutoCreatedQueue(server, queue);
+               QueueManagerImpl.deleteAutoDeleteQueue(server, queue);
```
clebertsuconic: I'm confused.. is that a typo on the name?

michaelandrepearce: its because the name gives meaning it only affected auto created queue, where thats not true. you want delete on any queue thats auto delete.

clebertsuconic: Do you mind if I rename this to performAutoDeleteQueue?

I can amend it myself if you agree.

michaelandrepearce: sure. just wanted better naming that was all. happy with your suggestion.

jbertram:
Could this issue be addressed a bit more simply by just marking the queues backing JMS topic subscriptions as "auto-created" so the existing logic will delete them when necessary?

clebertsuconic:
@jbertram topic subscriptions are not really auto-created, the client performs a create...

auto-create has other side effects, right? like deleting it when no consumers and no messages? that would easily break the TCK.

michaelandrepearce:
Correct @clebertsuconic , also that would mean a change in behaviour for existing users, anything we do we dont want change of behaviour unless someone explicitly sets right? or we do a major version bump, which we want to avoid.

jbertram:
Fair enough.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
clebertsuconic: ok, I amended it during merge with a temporary branch (used scripts ./scripts/checkout-branch.sh, amended, and ./scripts/merge-branch.sh)

it will appear on the git log as if you did it. (it's just better than coming back and forth asking you to do it).

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -3160,6 +3160,14 @@
                </xsd:annotation>
             </xsd:element>
 
+            <xsd:element name="auto-delete-created-queues" type="xsd:boolean" default="true" maxOccurs="1" minOccurs="0">
```
clebertsuconic: @michaelandrepearce I had a question from @michalxo on this...

Wan't the default here supposed to be "false" ?

PR #2603, opened 2019-04-02
ARTEMIS-2292 - Fix queue level auto-delete for topic subscription queue

michaelandrepearce:
Additional bits missed


michaelandrepearce:
@clebertsuconic just realized i missed a commit in squashed version in the prev pr you just merged. sorry.

PR #2605, opened 2019-04-03
ARTEMIS-2292 Fix default on XSD for auto-delete-created-queues

clebertsuconic:


clebertsuconic:
@michaelandrepearce can you check this please?

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -3160,7 +3160,7 @@
                </xsd:annotation>
             </xsd:element>
 
-            <xsd:element name="auto-delete-created-queues" type="xsd:boolean" default="true" maxOccurs="1" minOccurs="0">
```
clebertsuconic: @michaelandrepearce I'm making this change.. as @michalxo asked me about this.

michaelandrepearce: Yup good spot. Just quickly i spotted something late last night and fixed directly in master. Just in case you are cherry picking

PR #2608, opened 2019-04-04
NO-JIRA Adding Compatibility tests on Large Message for HornetQ

clebertsuconic:


clebertsuconic:
I was working on a test to validate an user's issue, and I had to write a test to prove it was a non issue.

Since there's no such thing as enough tests.. and I'm adding such tests to the codebase.

PR #2609, opened 2019-04-04
ARTEMIS-2294 implement duplicate detection for AMQP

jbertram:
I'm not entirely sure this should be an application property. Will someone with more expertise with AMQP weigh in? @tabish121? Should it be a delivery annotation?

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
@@ -971,7 +971,7 @@ public class AMQPMessage extends RefCountMessage {
 
    @Override
    public Object getDuplicateProperty() {
-      return null;
+      return getObjectProperty(org.apache.activemq.artemis.api.core.Message.HDR_DUPLICATE_DETECTION_ID);
    }
 
    @Override
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpSenderTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpSenderTest.java
@@ -20,6 +20,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.apache.activemq.artemis.api.core.Message;
 import org.apache.activemq.artemis.core.server.Queue;
 import org.apache.activemq.artemis.protocol.amqp.proton.AmqpSupport;
 import org.apache.activemq.artemis.tests.integration.IntegrationTestLogger;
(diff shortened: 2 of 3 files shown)


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -971,7 +971,7 @@ public Object getAMQPUserID() {
 
    @Override
    public Object getDuplicateProperty() {
-      return null;
+      return getObjectProperty(org.apache.activemq.artemis.api.core.Message.HDR_DUPLICATE_DETECTION_ID);
```
michaelandrepearce: Surely this should not be Artemis specific but some amqp value

jbertram: I wasn't aware that there was an AMQP-specific duplicate detection property/annotation. Do you have documentation on that?

tabish121:
@jbertram Given there's no requirements laid out in the issue I can't say for sure other than to say that if you want it to be set via an AMQP JMS client then I suppose application property is the only viable option.  Although that supposes that the message ID can't work for you.  

Given the application is going to have to manage these things I'm not entirely clear why it is being implemented in the broker to start with.

jbertram:
The duplicate detection is done by the broker as that is the only reliable place for the duplicate to be detected. See the [Artemis documentation ](https://activemq.apache.org/artemis/docs/latest/duplicate-detection.html)for more info on the use-cases. I don't think the message ID would suffice as that isn't under the application's control.

gemmellr:
I'd guess that Tim probably meant that doing it in the broker still requires the application to generate unique details per message (as it would if deduping entirely on its own), but even with the broker de-duping the production based on such property there are still cases consumption (+ production) duplicates can occur, so folks concerned with duplicates likely still need to take steps to handle that themselves to some extent regardless.

Service Bus dedupes based on the AMQP message-id field as far as I know. Uniquely identifying a message is the reason that field exists. Its use doesn't align that well with JMS applications though since JMS mandates the client sets the message-id on send. Arguably calling send multiple times is in fact 'sending a different message' so thats not particularly unreasonable...but back to application-level identification for de-duplication usage either way.

For consistency among different clients it would probably be nice if the broker supported its artemis-specific property, but similarly it might be nice to support using the message-id field in the intended manner. Using either will require decoding the section its in ('properties' for message-id, 'application properties' for the obvious) if it wasn't already done so, which could add some overhead. Perhaps an enabling toggle would be an idea.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
michaelandrepearce: I dont know if its amqp specific. I just know they do it by message id, which is a more agnostic to vendor field.

jbertram: Nothing here precludes Service Bus from continuing to de-dupe by message ID is there?

jbertram:
To the point about application requirements in this use-case, there are definitely requirements for the application if it wants to handle duplicates both for sending and receiving. That's always been true (as far as I know) and this PR doesn't change anything in that regard.

This PR is aimed at supporting the same duplicate detection for AMQP that we have for core. That's all. Additional functionality to de-duplicate AMQP messages based on message ID is beyond the scope of this PR, although that seems like a nice feature to have.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
michaelandrepearce: nope. It was more a query if there was any form of standard, i assumed maybe there was based on azure having same feature. But if there isn't and its vendor specific there. Then for now we just have to be vendor specific to our keys too. +1 this PR

michaelandrepearce:
+1

PR #2610, opened 2019-04-05
ARTEMIS-2295 fix IBM JDK schema parsing

clebertsuconic:


artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -1669,12 +1669,24 @@
    </xsd:complexType>
 
    <xsd:complexType name="federationTransformerType">
-      <xsd:complexContent>
-         <xsd:extension base="transformerType">
-            <xsd:attribute name="name" type="xsd:ID" use="required" />
-            <xsd:attributeGroup ref="xml:specialAttrs"/>
-         </xsd:extension>
-      </xsd:complexContent>
+      <xsd:sequence>
```
clebertsuconic: @michaelandrepearce I will merge this... but if you have any other better idea... I banged my head for a whole day into IBM JDK!!! I'm giving up.. and just going for this simple solution.

clebertsuconic: look at the JIRA for more information. Basically IBM JDK is complaining for no absolute reason.

I'm doing a copy & paste instead of the inheritance.

not as cool.. but I don't know how to solve the IBM JDK

michaelandrepearce: Its fine. It was valid xsd thats a nasty ibm bug, but you change doesnt break anything just causes us to duplicate schema bits. So im happy

PR #2611, opened 2019-04-09
ARTEMIS-2297 Avoiding Split Brains during replication catch up when n…

clebertsuconic:
…o quorum is established


clebertsuconic:
I'm running a complete testsuite. if everything is good in 2 hours and 30 minutes i will merge it tomorrow.

PR #2615, opened 2019-04-10
ARTEMIS-1230 Added artemis-bom

jmax01:
Added artemis-bom and removed <version>${project.version}</version>
tags for most org.apache.activemq:artemis* and
org.apache.activemq.rest:artemis* (jar, war, test-jar) dependencies.
org.apache.activemq.examples.*:* and org.apache.activemq.tests.*:*
dependencies were not addressed as it might be best
to handle them separately

### DIFF
--- a/artemis-bom/pom.xml
+++ b/artemis-bom/pom.xml
@@ -1,0 +1,737 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+   <modelVersion>4.0.0</modelVersion>
+   <parent>
+      <groupId>org.apache</groupId>
+      <artifactId>apache</artifactId>
+      <version>18</version>
+      <relativePath>org.apache:apache</relativePath>
+   </parent>
+   <!-- Since artemis-pom declares a dependencyManagement section our parent must be org.apache:apache and we must redeclare artemis project management details. -->
+   <groupId>org.apache.activemq</groupId>
+   <artifactId>artemis-bom</artifactId>
+   <packaging>pom</packaging>
+   <version>2.8.0-SNAPSHOT</version>
+
+   <name>ActiveMQ Artemis Bill of Materials</name>
+   <url>http://apache.org/activemq</url>
+
+   <prerequisites>
+      <maven>3.1.0</maven>
+   </prerequisites>
+
+   <properties>
+       <!-- base url for site deployment.  See distribution management for full url.  Override this in settings.xml for staging -->
+      <staging.siteURL>scp://people.apache.org/x1/www/activemq.apache.org</staging.siteURL>
+
+      <ActiveMQ-Version>${project.version}(${activemq.version.incrementingVersion})</ActiveMQ-Version>
+      <activemq-artemis-native-version>1.0.0</activemq-artemis-native-version>
+      <activemq.basedir>${project.basedir}</activemq.basedir>
+      <activemq.version.incrementingVersion>130,129,128,127,126,125,124,123,122</activemq.version.incrementingVersion>
+      <activemq.version.majorVersion>1</activemq.version.majorVersion>
+      <activemq.version.microVersion>0</activemq.version.microVersion>
+      <activemq.version.minorVersion>0</activemq.version.minorVersion>
+      <activemq.version.versionName>${project.version}</activemq.version.versionName>
+      <activemq.version.versionTag>${project.version}</activemq.version.versionTag>
+     
+      <maven.assembly.plugin.version>2.4</maven.assembly.plugin.version>
+      <owasp.version>1.4.3</owasp.version>
+      <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+      <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+      <skipLicenseCheck>true</skipLicenseCheck>
+      <skipOWASP>true</skipOWASP>
+      <skipStyleCheck>true</skipStyleCheck>
+   </properties>
+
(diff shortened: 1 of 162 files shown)


clebertsuconic:
Was there a particular reason other than make it more organized?


I mean: was there a particular error that you faced that lead you to do this?

jmax01:
@clebertsuconic 
Boms minimize (or eliminate altogether) the risk of having mismatched versions of a project's artifacts. 
They also allow the sharing of versions without relying on the _Maven_ pom hierarchy.

We are using _Artemis_ embedded with _Spring Boot_. _Spring Boot_ declares 10 _Artemis_ dependencies in its [bom](https://github.com/spring-projects/spring-boot/blob/v2.0.9.RELEASE/spring-boot-project/spring-boot-dependencies/pom.xml), we require two additional dependencies.  If we wish to vary our version of _Artemis_ from the _Spring Boot_ version (which we [do ](https://issues.apache.org/jira/browse/ARTEMIS-2206?jql=project%20%3D%20ARTEMIS%20AND%20fixVersion%20%3D%202.7.0%20AND%20summary%20~%20%22mqtt%22%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC)) we then have to duplicate all the _Spring_ declared dependencies or else we risk having inconsistent versions of _Artemis_ libraries in our classpath.  

Further we have non-_Spring Boot_ libraries that use _Artemis_ so again we have to explicitly declare each dependency we reference. 

It should be noted that _Spring Boot's_ bom contains 408 _non-spring-framework_ dependencies and _Activemq_ (both projects) make up over 9% of the total entries.

For the record though I am not a fan of how _Spring Boot_ handles their dependencies and would not recommend it for _Artemis_.  _Spring Boot_ mixes Spring and non-Spring dependencies together.  In my experience boms for large projects should be decomposed in the following manner: 

- _External Project Bom_:  The members of the project that will be published externally.  This is the _artemis-bom_  created in this pull request
- _External Third Party Bom_:  The third party dependencies that are used by the items declared in the _External Project Bom_
- _Composite External Bom_:  The composite of the other _External Boms_
- _Internal Project Bom (Optional)_:  The members of the project that will not be published externally. This would contain the _org.apache.activemq.tests_ modules 
- _Internal Third Party Bom (Optional)_:  The third party dependencies that are used by the items declared in the _Internal Project Bom_. This would contain any third party dependencies reference by _org.apache.activemq.tests_ modules.
- _Composite Internal Bom  (Optional)_:  The composite of all the other bom projects. This is really only needed when projects cannot share a _parent_. If modules share the same parent then the boms can be added to the parent's `dependencyManagement` section.

Finally providing a bom (or boms) has become pretty standard for larger projects (spring, jackson, netty, jetty, log4j2, junit, infinispan, jersey, wildfly, etc)

> Was there a particular reason other than make it more organized?
> 
> I mean: was there a particular error that you faced that lead you to do this?

michaelandrepearce:
@jmax01 on that note then in doing this instead of declaring a large list of the same third parties submodules why doesnt this just add their boms instead as further clean up.

jmax01:
@michaelandrepearce Can you clarify the _this_  to which you are referring? 

But if you mean why I didn't I clean up artemis's third party dependencies, it was because that would fall outside of the scope of the originally submitted issue.

> @jmax01 on that note then in doing this instead of declaring a large list of the same third parties submodules why doesnt this just add their boms instead as further clean up. As you give example with netty is a clear case.
> 
> Also i dont see why we would make some modules special. either we do it one way or another. Not two diff approaches.

michaelandrepearce:
This = the pr/change your proposing.

Either everything follows the bom way or everything follows the existing. Its not good to have two approaches that change depending. It makes things actually worse.

I quite like the bom approach. But i think it needs to be more complete if we were to change artemis where it would be done for all. 

I dont want a half way house personally. Its all in or all out.

jmax01:
I agree, but when submitting PR's I attempt to keep the focus as narrow as possible as that is what most projects prefer. 

If it is agreed among the maintainers that they want to go all in I am more than will to submit another PR with those changes. 

I think we would need another issue as the scope is a bit different than the original.

> This = the pr/change your proposing.
> 
> Either everything follows the bom way or everything follows the existing. Its not good to have two approaches that change depending. It makes things actually worse.
> 
> I quite like the bom approach. But i think it needs to be more complete if we were to change artemis where it would be done for all.
> 
> I dont want a half way house personally. Its all in or all out.

michaelandrepearce:
So i think maybe it should be discussed on dev list to get the lay of the land. Im for it, but it needs to be discussed and have concensus as it is changing the way people are used to with the project quite a bit.

Just to be clear i am against having a half and half approach.

jmax01:
I will create another pr with the entire solution

clebertsuconic:
@jmax01 I'm all for it... 

however.. when submitting big changes.. it's nice to have a [DISCUSS] thread on the dev-list.

if you could also create the thread on the dev-list.. PR discussions are not sent to the dev-list any longer, and having a discussion there would probably be a good idea.

michaelandrepearce:
Has a [DISCUSS] thread been sent? I haven't seen one....

clebertsuconic:
@michaelandrepearce not yet.. I was waiting if @jmax01 wanted to start it first

michaelandrepearce:
@clebertsuconic sorry, yes the question was meant for @jmax01

jmax01:
> @michaelandrepearce not yet.. I was waiting if @jmax01 wanted to start it first

Haven't had a chance yet.  Spent the weekend looking at the scope of what might be changed.  Wrote a couple of utilities generate some data:

- Identified versions explicitly declared (not property driven) in `<dependency>` or `<dependencyManagement>` sections  (around 63)
- Identified version inconsistencies 
- Identified version property convention mismatches (<slf4j-version>1.7.21</slf4j-version, vs<slf4j.version>1.7.21</slf4j.version>)
- Identified projects that do not have explicit referenced dependencies declared (`mvn dependency:analyze -Dverbose=true -DoutputXML=true`)
- Found that `org.apache.activemq.examples.broker:netty-openssl` project version  is `2.7.0-SNAPSHOT` and not `2.8.0-SNAPSHOT`

Once a have the issue set determined I will start the thread.

clebertsuconic:
@jmax01 I would start with the DISCUSS now... that gives people time to suggest changes ahead... make suggestions.. or simply accept the changes you are making.

You don't need to have the whole solution ready for the start of the thread.

jmax01:
@clebertsuconic  will do.

jmax01:
Team, sorry for the delay.

The encrypted drive on my primary machine decided to no longer decrypt and I have been busy getting my environment restored.

I will try to get the [DISCUSS] thread created as soon as I am able.

jbertram:
@jmax01, it's been awhile since you last commented. I don't recall seeing a [DISCUSS] thread. Did you ever start one? If not, will you?

michaelandrepearce:
I think this PR can close, if it remains inactive and no discuss thread is sent in the next few weeks. Just to keep PR backlog only to those that are actively worked on.

sebthom:
I would like to see such an artemis-bom. Helps to always resolve the correct transitive dependencies when using embedded Artemis in Maven project.

michaelandrepearce:
I'm closing this, the issues haven't been resolved, nor discussion thread. Please re-open once issues are resolved and a discussion thread with positive outcome has occurred.

jbertram:
I was looking at reviving this idea with an updated PR based essentially on the work already done by @jmax01. However, I don't understand your objection, @michaelandrepearce. Can you elaborate on the all-in vs. all-out issue you raised? Isn't this BOM just giving folks who want it the option of using and everyone else can continue doing business as normal?

PR #2616, opened 2019-04-11
ARTEMIS-2301 Minor change on Epoll and kQueue.isAvailable

clebertsuconic:
just checking for NoClassDefFound and return false

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/CheckDependencies.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/CheckDependencies.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.activemq.artemis.core.remoting.impl.netty;
+
+import io.netty.channel.epoll.Epoll;
+import io.netty.channel.kqueue.KQueue;
+import org.apache.activemq.artemis.core.client.ActiveMQClientLogger;
+import org.apache.activemq.artemis.utils.Env;
+import org.jboss.logging.Logger;
+
+/**
+ * This class will check for Epoll or KQueue is available, and return false in case of NoClassDefFoundError
+ * it could be improved to check for other cases eventually.
+ */
+public class CheckDependencies {
+
+   private static final Logger logger = Logger.getLogger(CheckDependencies.class);
+
+   public static final boolean isEpollAvailable() {
+      try {
+         return Env.isLinuxOs() && Epoll.isAvailable();
+      } catch (Throwable e)  {
+         ActiveMQClientLogger.LOGGER.unableToCheckKQueueAvailability(e);
+         return false;
+      }
+   }
+
+   public static final boolean isKQueueAvailable() {
+      try {
+         return Env.isMacOs() && KQueue.isAvailable();
+      } catch (Throwable e) {
+         ActiveMQClientLogger.LOGGER.unableToCheckKQueueAvailability(e);
+         return false;
+      }
+   }
+}
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/Env.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/Env.java
@@ -59,6 +59,10 @@ public final class Env {
     */
    private static boolean testEnv = false;
 
+   private static final String OS = System.getProperty("os.name").toLowerCase();
+   private static final boolean IS_LINUX = OS.startsWith("linux");
+   private static final boolean IS_MAC = OS.startsWith("mac");
+
    private Env() {
 
    }
@@ -77,6 +81,14 @@ public final class Env {
 
    public static void setTestEnv(boolean testEnv) {
       Env.testEnv = testEnv;
+   }
+
+   public static boolean isLinuxOs() {
+      return IS_LINUX == true;
+   }
+
+   public static boolean isMacOs() {
+      return IS_MAC == true;
    }
 
 }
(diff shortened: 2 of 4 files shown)


michaelandrepearce:
So this is bundled with code clebert what scenario is this not coming in? It would be as bad as not having netty on the cp

clebertsuconic:
@michaelandrepearce the main issue is we removed KQueue from the classpath during runtime, and the server would not start.

clebertsuconic:
@michaelandrepearce  on the case kqueue is removed only from runtime. so, the check above would work.

michaelandrepearce:
But thats like removing netty on the cp. Surely if you want kqueue deactivated and use nio then just use the settings. I dont think we should be advicating removal of bundled jars

michaelandrepearce:
Also to note that if in my system i am expecting kqueue and imagine i have managed to somehow break my dependency tree i would want it to epically fail like it does. Just as i expect any other dependency needed if missing i would expect it to crash and fail fast. 

I wouldnt want it to silently carry on and in hope i spot a log line to notice.

michaelandrepearce:
Just a note if you really want to remove it from the cp and you dont want kqueue then set useKQueue to false and it wont go into that code path already.

clebertsuconic:
@michaelandrepearce there are no semantic changes on the commit I made here.. it's just... *If* happened. just capturing the exception, logging.debug, and returning false. that's all.

The real reason is that I have a rebuild of netty without KQueue available at runtime. this is checking if KQueue is available or not.

michaelandrepearce:
There is before it would throw a exception and fail to start. But now it wouldnt and would continue thats a big change.

michaelandrepearce:
To add to that the fact it threw exception has caught early already an osx client that missing this dependency when they expected it to be there. But mucked up their dependcies and thankfully it was noticed because it failed to start. Obviosuly if it was intentional they would have disabled kqueue using the flag. But it wasnt.

clebertsuconic:
I strongly disagree here.. it should back to Servlet.. log a warn if you like. and back to NioServerSocketChannel, no reason to stop the server from starting.

michaelandrepearce:
My question is why would you remove the libs from the classpath? If you dont want kqueue disable it via the property.

clebertsuconic:
My answer is above, because I have a rebuild with netty that does not include kQueue on it. in a certain cenario where, on a linux machine epoll is not available, then it would try kqueue, and it would fail. I wanted to keep the code aligned with a fork my company has without having to maintain separate branches.

and I believe it would still valid to perform the check on the class without any other side effects.

michaelandrepearce:
So in my company we rely on the jars being there and expect failure if a provided jar of the apache project is missing.

As i said there is a way to disable it, if you wish to do some custom build.

clebertsuconic:
@michaelandrepearce the server starts even with this... the acceptor won't be found. 

I will add a log.warn if the class is not found, and same semantics will be preserved.

michaelandrepearce:
Also btw im not sure if you realise but by creating an incompelete(eg not all modules built) build of netty. You then leave users an issue where they may have have other projects in their apps then you leave them with a verison of netty that wont be able to resolve all dependencies because they depend on the modules being there. Just a heads up. As ive experienced that. It causes either choosing to override a custom builds version or forcing users to build the missing netty modules them selves. Its pretty nasty not fully building netty tbh.

clebertsuconic:
I don't have the environment to build and support KQueue on my downstream build.

clebertsuconic:
I don't understand what is your concern? All I will do is to log a warn, and keep the server running if any CFNE happens, The server already starts.. a log.warn would raise any concerns.

michaelandrepearce:
Lets talk in about an hour on slack

michaelandrepearce:
Pinging you on slack. If youre there. Ill hang about for 30 mins

michaelandrepearce:
@clebertsuconic see here: https://github.com/apache/activemq-artemis/commit/0bc55100592453cbfd2fe1359dcd8d96b347d58b#diff-82f39a6c64e759a782ec63b8ed103700

used to have our own Epoll and Queue class's that we did pre-checks on.

michaelandrepearce:
we removed it in https://github.com/apache/activemq-artemis/commit/b53b3acdb42e67fc596535523ba393d76bcff2e7#diff-ecf6968445bd8f176790fbf90c4f090b

so its been like this for a while....

Anyhow revert some of this commit to get back those original checks.

michaelandrepearce:
I take it all back...you're correct we did used to catch and log. As such entirely my bad.

+1

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/CheckDependencies.java
```diff
@@ -0,0 +1,52 @@
...
+         return Epoll.isAvailable();
+      } catch (java.lang.NoClassDefFoundError e) {
+         ActiveMQClientLogger.LOGGER.unableToCheckKQueueAvailability(e);
+         return false;
+      }
+   }
+
+   public static final boolean isKQueueAvailable() {
+      try {
+         return Env.isMacOs() && KQueue.isAvailable();
```
michaelandrepearce: is the Env.isMacOs() needed? as you pointed out to me. I agree i totally was wrong earlier.
 it would anyhow as raised it will just throw a NCDFE.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/CheckDependencies.java
```diff
@@ -0,0 +1,52 @@
...
+      } catch (java.lang.NoClassDefFoundError e) {
+         logger.debug(e.getMessage(), e);
```
michaelandrepearce: is debugger needed, as logger one line below.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/CheckDependencies.java
clebertsuconic: It would avoid checking for KQueue if not even needed.
In that sense you would see a NCDFE only if needed.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/CheckDependencies.java
clebertsuconic: removed.. forgot it in the process.

clebertsuconic:
@michaelandrepearce let me know if it's ok to merge it... I'm looking to do it tomorrow.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/CheckDependencies.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+public class CheckDependencies {
+
+   private static final Logger logger = Logger.getLogger(CheckDependencies.class);
+
+   public static final boolean isEpollAvailable() {
+      try {
+         return Env.isLinuxOs() && Epoll.isAvailable();
+      } catch (Throwable e)  {
+         ActiveMQClientLogger.LOGGER.unableToCheckKQueueAvailability(e);
```
franz1981: ActiveMQClientLogger.LOGGER.unableToCheckEpollAvailability(e)?

michaelandrepearce: did this get fixed on merge?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/CheckDependencies.java
slovdahl: A little bit late to the party, but shouldn't check also evaluate to `true` if we're running on e.g. FreeBSD?

michaelandrepearce: Yes it should @clebertsuconic the old netty one did do this

PR #2621, opened 2019-04-16
NO-JIRA Fix small ActiveMQServerControl api break with 2.6.4

michaelandrepearce:
In adding auto-delete queue level feature, its been noticed as some feature bits were added during hot fix branch, that there's api break with the 2.6.x hotfix branch.

This addresses that by fixing this in 2.7.x


michaelandrepearce:
@clebertsuconic another small find, nothing major just adding a method to ActiveMQServerControl to fix an issue where features were cherry picked into 2.6.x branch. Just makes 2.7.x / master compatible with the 2.6.x branch for these exposed api methods.

PR #2623, opened 2019-04-16
ARTEMIS-2309 TempQueueCleanerUpper instances are leaking

jbertram:
The changes from ARTEMIS-2189 mean that
o.a.a.a.c.s.i.ServerSessionImpl#deleteQueue
is no longer called from the same ServerSessionImpl instance that
created it which means that TempQueueCleanerUpper instances will leak.
To resolve the leak the tracking Map should be static (i.e. shared among
all the ServerSessionImpl instances) so it doesn't matter which session
creates or deletes the temporary queue.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
@@ -291,6 +291,10 @@ public class ServerSessionImpl implements ServerSession, FailureListener {
       this.closeables.add(closeable);
    }
 
+   public Map<SimpleString, TempQueueCleanerUpper> getTempQueueCleanUppers() {
+      return tempQueueCleannerUppers;
+   }
+
    @Override
    public Executor getSessionExecutor() {
       return sessionExecutor;
--- a/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQDestination.java
+++ b/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQDestination.java
@@ -421,18 +421,30 @@ public class ActiveMQDestination extends JNDIStorable implements Destination, Se
 
    public void delete() throws JMSException {
       if (session != null) {
-         /**
-          * The status of the session used to create the temporary destination is uncertain, but the JMS spec states
-          * that the lifetime of the temporary destination is tied to the connection so even if the originating
-          * session is closed the temporary destination should still be deleted. Therefore, just create a new one
-          * and close it after the temporary destination is deleted. This is necessary because the Core API is
-          * predicated on having a Core ClientSession which is encapsulated by the JMS session implementation.
-          */
-         try (ActiveMQSession sessionToUse = (ActiveMQSession) session.getConnection().createSession()) {
+         boolean openedHere = false;
+         ActiveMQSession sessionToUse = session;
+
+         if (session.getCoreSession().isClosed()) {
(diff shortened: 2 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -153,7 +153,7 @@
 
    protected volatile boolean started = false;
 
-   protected final Map<SimpleString, TempQueueCleanerUpper> tempQueueCleannerUppers = new HashMap<>();
+   protected static final Map<SimpleString, TempQueueCleanerUpper> tempQueueCleannerUppers = new HashMap<>();
```
clebertsuconic: this does not work well here. 

A HashMap on ServerSession is fine, as it represents a single thread context of the client.

however, if you make it static. .that's an issue.


Still, i am not sure i would make it static here, perhaps store such thing on the Server or other controller?

michaelandrepearce: What occurs where multiple embedded brokers in same jvm? 
Always super cautious of state in statics.

michaelandrepearce: As temp dest are tied to the connection should maybe the map live there and shared on all sessions on thay connection. Thus removing the static but also being correct with regards to being able to share between sessions on same connection

jbertram: @michaelandrepearce, that's good idea. I'll look into that...

michaelandrepearce: Also if being shared between sessions on a connection need to handle concurrency. Maybe simply use concurrenthashmap

clebertsuconic: I have an alternative approach...

clebertsuconic: I'm sending a PR now

clebertsuconic: @jbertram I sent a PR to your branch

clebertsuconic: The issue is that the session is temp is always creating a new session to delete the queue.

I'm reusing the original session on my patch, and only opening a new one if it happened to be closed. that's a minimal change here.

michaelandrepearce: I put comment on your pr to justins.

Essentially what about other protocols...

jbertram: @clebertsuconic, I'm using your change, but I squashed the commits. I hope that's ok.

@michaelandrepearce, I don't think there's going to be any issue on the other protocols because the JMS client should be the only one that was creating a new session every time when deleting a temporary queue.

michaelandrepearce: fair enough, i guess im just worried temp queues being fixed (the prev pr that caused this) for core all well and good, but we miss the other protocols and then find a more central solution for all would have been better...

anyhow +1 this for the immediate.

PR #2624, opened 2019-04-16
alternative fix to #2623

clebertsuconic:


clebertsuconic:
They both deal with it on the client AFIK.

michaelandrepearce:
As both are jms as well, and the test cases are jms is it worth running same tests for those to be sure?

clebertsuconic:
Not in the context of this fix.

This fix is about a Leak on TempQueueCleanerUpper, which is only used by the CoreProtocol, more specificially on ServerSessionImpl::createQueue and deleteQueue. In both cases they only have calls from Packets received on the core protocol.


If you think Temp Queues may suffer in other protocols, that's a different issue, and it would be on a different PR.

michaelandrepearce:
fair enough, i guess im just worried temp queues being fixed for core all well and good, but we miss the other protocols and then find a more central solution for all would have been better....

anyhow +1 this for the immediate.

PR #2625, opened 2019-04-17
ARTEMIS-2306 Make group first off by default, unless configured

michaelandrepearce:


michaelandrepearce:
@gemmellr @clebertsuconic 
Disables feature by default.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSMessageGroupsTest.java
```diff
@@ -53,6 +55,13 @@ protected String getConfiguredProtocols() {
       return "AMQP,OPENWIRE,CORE";
    }
 
+   @Override
+   protected void configureAddressPolicy(ActiveMQServer server) {
+      super.configureAddressPolicy(server);
+      AddressSettings addressSettings = server.getConfiguration().getAddressesSettings().get("#");
```
gemmellr: This seems like it enables the feature for all the groups tests. It would be nice to enable it only for those tests aimed at testing that functionality. Tests can/should use test specific addresses, so could it be configured only for those used by the specific tests verifying it works?

A negative-test that its otherwise disabled might also be nice, there doesnt look to be one here or in the Core tests.

michaelandrepearce: Thats because all the tests are testing that functionality currently. re negative test case, though what are you testing, as you wont have a key to look for / test if it is present or not.

gemmellr: Ah, that is unfortunate re: all tests testing it, given its not on by default that would mean none of the tests actually cover what most users might see by default.

In a negative test case, what you described sounds good. Beyond that I'd maybe check that I get only the properties that I sent (which might be none).

michaelandrepearce: Ok have made it so we have a specific address-setting where we turn it ON.
And re-factored tests, so we now 

Test with non-prefixed address where it should be OFF by default, and check the property isnt there. 
And then a separate test where we use an address which is prefixed and matches the ON address setting, and test the property is there for the first message, and not for others.

michaelandrepearce: Just now fixed a checkstyle i introduced whoops...

michaelandrepearce: @gemmellr forgot to say thanks for the review.

michaelandrepearce:
@clebertsuconic  i've addressed all Robbies concerns, and added further tests on his feedback. Im about to finish for the day, anything else?

clebertsuconic:
I will merge onde I’m back from lunch :)

michaelandrepearce:
Cool. Have a great easter weekend :)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSMessageGroupsTest.java
gemmellr: np

PR #1607, opened 2017-10-24
ARTEMIS-1478 - ActiveMQ Artemis to Kafka Bridge

michaelandrepearce:
Initial Kafka Bridge for Apache ActiveMQ Artemis to Apache Kafka, using Serivce Connector interface
Support to handle Core or AMQP based protocol over kafka.
Ensured for Core TextMessage can map to StringSerializer for a consumer
Ensure for Core ByteMessage can map to ByteArraySerializer for a consumer 
Kafka Serdes to support kafka consumers to consumer Core or AMQP back to CoreMessage, ProtonMessage, or JMSMessage
Added Documentation
Add integration tests

PR #2626, opened 2019-04-17
ARTEMIS-2308 Support exporting metrics to Prometheus

jbertram:
This commit adds an integrated implementation of the Prometheus JMX
Exporter (https://github.com/prometheus/jmx_exporter).

An integrated approach was preferrable to running the JMX exporter as a
Java Agent (which is the standard usage pattern) for several reasons.

The JMX Exporter Java Agent starts its own HTTP server, but we already
have an HTTP server available via embedded Jetty. This creates an
additional port to manage (which can be a significant hurdle in some
environments) as well as additional security attack vector. Securing
this web server via SSL requires setting the various Java SSL system
properties which may have undesirable consequences for other components
in the JVM. Also, the various keystore passwords cannot be masked when
set via system properties. Masking passwords is a requirement in many
environments. All these problems go away if the embedded Jetty HTTP
server is used.

The JMX Exporter Java Agent has its own YAML-based configuration file.
For usability's sake it is preferrable to avoid the proliferation of
configuration across different files and formats. All the same
configuration elements available via the individual YAML file are
available via broker.xml.

In general, configuring the Java Agent is a bit clumsy as it requires
command-line changes (e.g. modifying artemis.profile). An integrated
approach avoids these changes.

The RBAC security enforced on the MBean server makes using the Java
Agent difficult without modifications to the way the MBean server is
built/configured. Using an integrated approach reduces those changes
to a few lines.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
@@ -245,6 +245,9 @@ public class Create extends InputAbstract {
    @Option(name = "--mapped", description = "Sets the journal as mapped.")
    private boolean mapped;
 
+   @Option(name = "--prometheus", description = "Enables Prometheus support.")
+   private boolean prometheus;
+
    // this is used by the setupJournalType method
    private JournalType journalType;
 
@@ -576,6 +579,12 @@ public class Create extends InputAbstract {
       }
 
       filters.put("${journal.settings}", journalType.name());
+
+      if (prometheus) {
+         filters.put("${prometheus}", " prometheusServletEnabled=\"true\"");
+      } else {
+         filters.put("${prometheus}", "");
+      }
 
       if (sslKey != null) {
          filters.put("${web.protocol}", "https");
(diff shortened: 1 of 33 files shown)


michaelandrepearce:
I have left my comments in Jira. We should be exposing metrics generically for any metric system to integrate with via pluggable api.  We shouldnt be tying to one specific. Many users have different monitoring platforms we should agnostic to that.



-1 for such tight coupling.

michaelandrepearce:
One option would be to look at something that’s much more agnostic like https://opencensus.io

michaelandrepearce:
also just to note: https://github.com/apache/activemq-artemis/pull/1607 outcome of this and subsequent discussion threads was:

> broker should provide the pieces to enable integration, but that the integration plugins themselves were kept maintained outside of the broker.

michaelandrepearce:
Another approach could be like with istio mixer, with adapter interface

https://istio.io/docs/concepts/policies-and-telemetry/

This way can support any number of adapters to support anyones preffered metric collection technology, 

Prometheus, Influx, DataDog, CloudWatch, Circonus, Azure Monitor, StatsD, WaveFront etc. etc. the list can be quite long.

clebertsuconic:
That’s different from that other example. 

If you take extreme measures like this for every bit then a lot of integration would live outside. 

Prometheus is becoming a standard in monitoring. And having it integrated by default is a very important feature. 

I have offered to review options for the Kafka Brkdge in tbe past.  But the issue here is completely different.

michaelandrepearce:
@clebertsuconic Prometheus really isnt becoming standard. This is why Itsio abastracts this, reality is every org has preferred infra backends, and the number of tech opensource and closed is quite a sea still...

clebertsuconic:
We could make it abstract. And add Prometheus as a one of the implementations. It would be nice to have something out of the box.  Enabled during server creation. 

Using the argument of external repo could be used against everything.  Web server for instance. 

I don’t really think it applies the same thing. 

We should move the discussion to the dev list.  It’s bexomong out of the scope of code review now.

michaelandrepearce:
As i noted, i think an approach similar to Istio would be quite pragmatic...

PR #2627, opened 2019-04-17
ARTEMIS-2311 Dealing with Protocol conversions and JMSReplyTo

clebertsuconic:


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
```diff
@@ -106,11 +106,12 @@
    private final CoreMessageObjectPools coreMessageObjectPools;
 
    public CoreMessage(final CoreMessageObjectPools coreMessageObjectPools) {
+      new Exception("new coreMessage").printStackTrace();
```
michaelandrepearce: ?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
```diff
@@ -106,11 +106,12 @@
...
       this.coreMessageObjectPools = coreMessageObjectPools;
    }
 
    public CoreMessage() {
-      this.coreMessageObjectPools = null;
+      this((CoreMessageObjectPools)null);
```
michaelandrepearce: ?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
clebertsuconic: geez.. removing it

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
clebertsuconic: this is just using the same constructor, that is doing the same thing

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
clebertsuconic: that's a debug trick that leaked on the PR :)

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
clebertsuconic: I will remove it.. as it's not related to anything I needed here.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
```diff
@@ -530,4 +543,17 @@ private static byte destinationType(Destination destination) {
...
+      if (destination.startsWith(ActiveMQDestination.QUEUE_QUALIFIED_PREFIX_SIMPLE))
+         return QUEUE_TYPE;
+      if (destination.startsWith(ActiveMQDestination.TEMP_QUEUE_QUALIFED_PREFIX_SIMPLE))
+         return TEMP_QUEUE_TYPE;
+      if (destination.startsWith(ActiveMQDestination.TOPIC_QUALIFIED_PREFIX_SIMPLE))
+         return TOPIC_TYPE;
+      if (destination.startsWith(ActiveMQDestination.TEMP_TOPIC_QUALIFED_PREFIX_SIMPLE))
+         return TEMP_TOPIC_TYPE;
+
+      throw new IllegalArgumentException("Unknown Destination Type passed to JMS Transformer.");
```
michaelandrepearce: This will break stuff not everyone does jms. It’s valid to send without a jms type.

clebertsuconic: It's only used for JMSReplyTo property. but I will return Queue (as it does now) just in case.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -220,11 +221,11 @@ public static ICoreMessage toCore(AMQPMessage message, CoreMessageObjectPools co
          }
       }
 
-      result.getInnerMessage().setReplyTo(message.getReplyTo());
       result.getInnerMessage().setDurable(message.isDurable());
       result.getInnerMessage().setPriority(message.getPriority());
       result.getInnerMessage().setAddress(message.getAddressSimpleString());
       result.getInnerMessage().setRoutingType(message.getRoutingType());
```
michaelandrepearce: isn’t this needed to mark it

clebertsuconic: this was removing whatever I had set before causing part of the issue. This is previously set on the converter, and this was removing the property.

clebertsuconic: and the test was asserting the case.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -324,7 +325,23 @@ private static ServerJMSMessage processProperties(ServerJMSMessage jms, Properti
...
+                  org.apache.activemq.artemis.reader.MessageUtil.setJMSReplyTo(jms.getInnerMessage(), ActiveMQDestination.QUEUE_QUALIFIED_PREFIX + properties.getReplyTo());
+                  break;
+               case AMQPMessageSupport.TEMP_QUEUE_TYPE:
+                  org.apache.activemq.artemis.reader.MessageUtil.setJMSReplyTo(jms.getInnerMessage(), ActiveMQDestination.TEMP_QUEUE_QUALIFED_PREFIX + properties.getReplyTo());
+                  break;
+               case AMQPMessageSupport.TOPIC_TYPE:
+                  org.apache.activemq.artemis.reader.MessageUtil.setJMSReplyTo(jms.getInnerMessage(), ActiveMQDestination.TOPIC_QUALIFIED_PREFIX + properties.getReplyTo());
+                  break;
+               case AMQPMessageSupport.TEMP_TOPIC_TYPE:
+                  org.apache.activemq.artemis.reader.MessageUtil.setJMSReplyTo(jms.getInnerMessage(), ActiveMQDestination.TEMP_TOPIC_QUALIFED_PREFIX + properties.getReplyTo());
```
michaelandrepearce: What about no match it’s valid to not have extra jms type info just it won’t map to specific type

clebertsuconic: That's a Message Annotation. all valid values are in the range.

I will set Queue on default though, but it shouldn't happen ever

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -196,7 +200,18 @@ private OpenWireMessageConverter() {
...
+         } else
+         if (replyTo instanceof TemporaryTopic) {
+            MessageUtil.setJMSReplyTo(coreMessage, org.apache.activemq.artemis.jms.client.ActiveMQDestination.TEMP_TOPIC_QUALIFED_PREFIX + (((TemporaryTopic) replyTo).getTopicName()));
+         } else
+         if (replyTo instanceof Queue) {
+            MessageUtil.setJMSReplyTo(coreMessage, org.apache.activemq.artemis.jms.client.ActiveMQDestination.QUEUE_QUALIFIED_PREFIX + (((Queue) replyTo).getQueueName()));
+         } else
+         if (replyTo instanceof Topic) {
+            MessageUtil.setJMSReplyTo(coreMessage, org.apache.activemq.artemis.jms.client.ActiveMQDestination.TOPIC_QUALIFIED_PREFIX + (((Topic) replyTo).getTopicName()));
+         }
```
michaelandrepearce: What happens if no match .... e.g. Else?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -220,11 +224,11 @@ public static ICoreMessage toCore(AMQPMessage message, CoreMessageObjectPools co
          }
       }
 
-      result.getInnerMessage().setReplyTo(message.getReplyTo());
       result.getInnerMessage().setDurable(message.isDurable());
       result.getInnerMessage().setPriority(message.getPriority());
       result.getInnerMessage().setAddress(message.getAddressSimpleString());
       result.getInnerMessage().setRoutingType(message.getRoutingType());
+      //result.getInnerMessage().messageChanged();
```
michaelandrepearce: .?

clebertsuconic: debug forgotten. nice catch!

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -220,7 +224,6 @@ public static ICoreMessage toCore(AMQPMessage message, CoreMessageObjectPools co
          }
       }
 
-      result.getInnerMessage().setReplyTo(message.getReplyTo());
```
michaelandrepearce: I think I added this due to a bug. Can you check history by selection, I would have added test case and check it’s passing.

clebertsuconic:
It's all caught up with you now. real time review.. pair programming :)


I'm calling the day, I may do more stuff tomorrow if you found anything else

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
clebertsuconic: The previous set was definitely buggy. that's why you added this... but this was always setting it to Queue.

I'm doing a more through check now.

clebertsuconic: I will run a full testsuite on this.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -324,7 +327,26 @@ private static ServerJMSMessage processProperties(ServerJMSMessage jms, Properti
...
+               case AMQPMessageSupport.TEMP_QUEUE_TYPE:
+                  org.apache.activemq.artemis.reader.MessageUtil.setJMSReplyTo(jms.getInnerMessage(), ActiveMQDestination.TEMP_QUEUE_QUALIFED_PREFIX + properties.getReplyTo());
+                  break;
+               case AMQPMessageSupport.TOPIC_TYPE:
+                  org.apache.activemq.artemis.reader.MessageUtil.setJMSReplyTo(jms.getInnerMessage(), ActiveMQDestination.TOPIC_QUALIFIED_PREFIX + properties.getReplyTo());
+                  break;
+               case AMQPMessageSupport.TEMP_TOPIC_TYPE:
+                  org.apache.activemq.artemis.reader.MessageUtil.setJMSReplyTo(jms.getInnerMessage(), ActiveMQDestination.TEMP_TOPIC_QUALIFED_PREFIX + properties.getReplyTo());
+                  break;
+               default: // in theory this should never happen
```
tabish121: Why would this never happen in theory, AMQP messages can have message annotations and also use a reply to value and NOT be from a JMS type client.

michaelandrepearce: Agree here.

michaelandrepearce: Worth adding some non jms tests

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
```diff
@@ -355,6 +359,15 @@ public static AMQPMessage fromCore(ICoreMessage coreMessage) throws Exception {
       }
    }
 
+   private static String removePrefix(String address) {
```
michaelandrepearce: Isn’t this logic somewhere in a util or static already? Looks familiar

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
```diff
@@ -530,4 +543,17 @@ private static byte destinationType(Destination destination) {
       throw new IllegalArgumentException("Unknown Destination Type passed to JMS Transformer.");
```
michaelandrepearce: I know it’s not your change but this looks like a nasty bug waiting to occur. Where a non jms client is being used

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
clebertsuconic: @tabish121 if the value is not set here, it will use QUEUE-TYPE, from line 330:

```java
            Byte value = (Byte)(annotations != null ? annotations.getValue().get(AMQPMessageSupport.JMS_REPLY_TO_TYPE_MSG_ANNOTATION) : AMQPMessageSupport.QUEUE_TYPE);
```

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
```diff
@@ -173,9 +174,12 @@ public static AMQPMessage fromCore(ICoreMessage coreMessage) throws Exception {
          properties.setTo(toAddress(destination));
          maMap.put(JMS_DEST_TYPE_MSG_ANNOTATION, destinationType(destination));
       }
-      Destination replyTo = message.getJMSReplyTo();
```
michaelandrepearce: Why remove this seemed a cleaner way to encapsulate

clebertsuconic: This was using a fake (incomplete) JMS Layer only available on Server, that only had ActiveMQDestination as base, and in the end it was just to the the property.

clebertsuconic: you would need to implement the other 4 types of classes on the JMS Layer, to get to the same result in the end.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
clebertsuconic: I'm removing the note

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: There are only 4 types of JMS Queues.. all of them part of destinationType check

This is not used on JmsReply Any longer.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: why not? would make it cleaner. Out of query why did we create a JMS representation of amqp qpid bits, when actually already qpid-jms has done this. Then it would make JMS to JMS much cleaner, as you would parse it through the same mapping logic to handle any foreign JMS message when you do a JMS produce.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: The only bug that there is here, is JMS_DEST_TYPE_MSG_ANNOTATION, destinationType will always be a fixed Type, or Queue.

clebertsuconic: I will deal with it on a separate PR:
https://issues.apache.org/jira/browse/ARTEMIS-2312
I marked as BLOCKER, so I won't release without fixing it.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -196,7 +200,21 @@ private OpenWireMessageConverter() {
 
       final ActiveMQDestination replyTo = messageSend.getReplyTo();
       if (replyTo != null) {
-         putMsgReplyTo(replyTo, marshaller, coreMessage);
+         if (replyTo instanceof TemporaryQueue) {
+            MessageUtil.setJMSReplyTo(coreMessage, org.apache.activemq.artemis.jms.client.ActiveMQDestination.TEMP_QUEUE_QUALIFED_PREFIX + (((TemporaryQueue) replyTo).getQueueName()));
```
michaelandrepearce: Same query i put on the AMQP one but also relevant i think here, but why dont we use the same logic when you transform a JMS Message for foreign Messages on producer send (which you have to handle as part of TCK)

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: yes but what if NOT jms.....

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
clebertsuconic: What Query you referring to? This is about conversions.. I'm not sure how the TCK applies here.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: I looked for it and couldn't find it. I can check tomorrow what the client is doing here.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: @clebertsuconic unresolving - as this is the other query im referring to from the openwire class.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
michaelandrepearce: you resolved it, just unresolved it.

clebertsuconic:
There are failing tests with NPE on the conversion already. I'll finish this tomorrow and add a few non JMS tests with AMQP client.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
michaelandrepearce: https://github.com/apache/activemq-artemis/blob/4ca3a971baeb97623384ddc17893f7ebaf8cace1/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java#L416

michaelandrepearce: and for doing it for OpenWire, we could levy its implementation, and like wise for AMQP can use qpid-jms varient.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -90,13 +91,16 @@
...
 import io.netty.buffer.PooledByteBufAllocator;
+import org.jboss.logging.Logger;
 
 /**
  *  This class was created just to separate concerns on AMQPConverter.
  *  For better organization of the code.
  * */
 public class AmqpCoreConverter {
 
+   private static final Logger log = Logger.getLogger(AmqpCoreConverter.class);
```
michaelandrepearce: I dont see this being used?

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
clebertsuconic: This is dealing directly with the properties here, I'm not going to implement ServerQueue, ServerTopic, ServerTemporaryTopic, ServerTemporaryQueue just to set the property on this case and still not solve it completely.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: not possible. This is just setting the properties directly.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
michaelandrepearce: sounds good

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: even so, it would be duplication. Those JMS Client classes are not available at the server's. It would be duplication anyways.

This is just at the conversion side for this property.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
clebertsuconic: it would mean duplciation regardless. The client JMS implementation is not available at the server's. This is only needed when converting this property between protocols.

michaelandrepearce: Just to be clear this is more a general query, and question for longer term maintenance of this stuff... its not needed to be addressed as part of this PR, just this PR raises the question.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
clebertsuconic: I was going to use, then removed. you're right

clebertsuconic: resolving it

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: PrefixUtil

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: @michaelandrepearce there's currently an expectation that if you're sending AMQP to be converted, the message has to be in a certain format anyways. It's not always possible to convert everything between protocols.


If NOT JMS, send AMQP, receive AMQP, and don't deal with conversions.


This specific bit wouldn't be a problem though.


Also: non JMS Conversions are currently a problem regardless of what I'm doing here. it's beyond the scope.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: ServerSessionImpl.removePrefix invokes it, you could probably invoke it via the same server session, that way it would pick up the same config as server session already has done for you.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: Fair enough to default it then, as long as we dont throw IllegalArgumentException i think that the main gotcha here.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: fair enough as was a query more than anything.

michaelandrepearce:
@clebertsuconic have resolved all the comments which were my queries, but you answered and subsequently made sense.

michaelandrepearce:
oops hit wrong button meant to just comment there, not close your pr, just re-opened. my bad

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -324,7 +324,29 @@ private static ServerJMSMessage processProperties(ServerJMSMessage jms, Properti
             jms.setJMSType(properties.getSubject());
          }
          if (properties.getReplyTo() != null) {
-            jms.setJMSReplyTo(new ServerDestination(properties.getReplyTo()));
+            Byte value = (Byte)(annotations != null ? annotations.getValue().get(AMQPMessageSupport.JMS_REPLY_TO_TYPE_MSG_ANNOTATION) : AMQPMessageSupport.QUEUE_TYPE);
```
michaelandrepearce: Cast to Number, and then call .toByteValue()

clebertsuconic: Guess What: MessageTransformationTest::testComplexQpidJMSMessageEncodeDecode was failing because of this! :)... fixing it!

clebertsuconic: and this was the only failure on the testsuite.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: I had to create a new method though.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
clebertsuconic: everything passing ok, I will resolve this!

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
clebertsuconic: I was mistaken, I thought the JMS Client was not available at the server.

The reason we created ServerMessage was because ActiveMQMessage needed a ClientMessage and we have a ServerMessage here.


I am removing ServerDestination and using the regular ActiveMQDestination. I think the changese will become cleaner for this.


BTW: I thought this was going to e a simple bug fix.

tabish121:
@clebertsuconic it would be good to see some additional tests added for the AMQP client case (not JMS) for reply-to being used, specifically with 

1. No Message Annotations
2. Message Annotations but not the JMS destination type one.
3. For completeness also provide annotations with JMS destination type.

clebertsuconic:
@tabish121 agreed.. i will do that on monday.
I won't be able to finish it before the weekend. I have addressed all requests except to the test you suggested. But I won't be able to finish it today.

clebertsuconic:
Every piece here is handled now.
I'm going to merge this.. if anything is needed beyond this, lets discuss online.. and we can follow up with another PR.


I'm intending to release soon including this, so I'm merging it.

PR #2628, opened 2019-04-18
ARTEMIS-2313 Accumulation in HierarchicalObjectRepository cache

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -2133,6 +2133,8 @@ public void destroyQueue(final SimpleString queueName,
          }
       }
 
+      getSecurityRepository().removeCacheMatch(queueName.toString());
```
clebertsuconic: Please do not merge this yet.. I'm looking at some details.

clebertsuconic: There is something wrong. We clear the entire cache every time a queue is removed.

This is actually being called twice.. on destroyQueue and on PostOffice.removeBinding.

clebertsuconic: @jbertram ^^^

PR #2629, opened 2019-04-19
ARTEMIS-2314 broken compat with old core JMS FQQN consumer

jbertram:
When auto-creation is off then older clients consuming messages from an
FQQN won't work. This commit fixes that problem and adds a compatibility
test to verify.

### DIFF
--- a/tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/FQQNConsumerTest.java
+++ b/tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/FQQNConsumerTest.java
@@ -1,0 +1,79 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.activemq.artemis.tests.compatibility;
+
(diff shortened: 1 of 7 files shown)


tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/FQQNConsumerTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+   // this will ensure that all tests in this class are run twice,
+   // once with "true" passed to the class' constructor and once with "false"
+   @Parameterized.Parameters(name = "server={0}, producer={1}, consumer={2}")
+   public static Collection getParameters() {
+      // we don't need every single version ever released..
+      // if we keep testing current one against 2.4 and 1.4.. we are sure the wire and API won't change over time
+      List<Object[]> combinations = new ArrayList<>();
+
+      combinations.add(new Object[]{SNAPSHOT, TWO_SIX_THREE, TWO_SIX_THREE});
```
clebertsuconic: why not test against older versions?


At least there's a copy & paste thing here, as the comment says you only need to test against 2.4 and 1.4.

clebertsuconic: You should also test against older versions, at least on 2.0 versions

jbertram: I added some additional tests for 2.x clients.

clebertsuconic: I am sending a PR to your branch, the issue is using SNAPSHOT against any older version, including 2.6.3.

I am not sure you need to test 2.6.3, testing a combinatory of 2.4.0, SNAPSHOT, 2.10. 2.0 should be enough, unless there was anything specific that was changed into 2.6.x and that would warrant a speicif test.

clebertsuconic: any time you use 2.6.3, 2.4.0, 2.2.0 as the server, and SNAPSHOT as a client, on any combination

jbertram: That's expected since the client is producing to an FQQN as well as consuming. Support for sending messages to an FQQN was added in 2.7.0.

PR #2633, opened 2019-04-23
ARTEMIS-2317 Avoid long TTSP caused by Page::read using mmap read

franz1981:
It implements Page::read using chunked reading of SequentialFile
instead of using memory mapped files causing long time to safepoint
issues.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
@@ -16,27 +16,20 @@
  */
 package org.apache.activemq.artemis.core.paging.impl;
 
-import java.io.File;
-import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.nio.MappedByteBuffer;
-import java.nio.channels.FileChannel;
-import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import io.netty.buffer.Unpooled;
-import io.netty.util.internal.PlatformDependent;
 import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
+import org.apache.activemq.artemis.api.core.ActiveMQBuffers;
 import org.apache.activemq.artemis.api.core.ICoreMessage;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.buffers.impl.ChannelBufferWrapper;
 import org.apache.activemq.artemis.core.io.SequentialFile;
 import org.apache.activemq.artemis.core.io.SequentialFileFactory;
-import org.apache.activemq.artemis.core.io.mapped.MappedSequentialFileFactory;
-import org.apache.activemq.artemis.core.io.nio.NIOSequentialFileFactory;
 import org.apache.activemq.artemis.core.paging.PagedMessage;
 import org.apache.activemq.artemis.core.paging.cursor.LivePageCache;
 import org.apache.activemq.artemis.core.paging.cursor.PageSubscriptionCounter;
@@ -45,6 +38,7 @@ import org.apache.activemq.artemis.core.server.ActiveMQMessageBundle;
 import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
 import org.apache.activemq.artemis.core.server.LargeServerMessage;
 import org.apache.activemq.artemis.utils.DataConstants;
+import org.apache.activemq.artemis.utils.Env;
 import org.apache.activemq.artemis.utils.collections.ConcurrentHashSet;
 import org.jboss.logging.Logger;
 
@@ -87,8 +81,6 @@ public final class Page implements Comparable<Page> {
     */
    private Set<PageSubscriptionCounter> pendingCounters;
 
-   private boolean canBeMapped;
-
    public Page(final SimpleString storeName,
                final StorageManager storageManager,
                final SequentialFileFactory factory,
@@ -99,8 +91,6 @@ public final class Page implements Comparable<Page> {
       fileFactory = factory;
       this.storageManager = storageManager;
       this.storeName = storeName;
-      this.canBeMapped = fileFactory instanceof NIOSequentialFileFactory || fileFactory instanceof MappedSequentialFileFactory;
-      //pooled buffers to avoid allocations on hot paths
    }
 
    public int getPageId() {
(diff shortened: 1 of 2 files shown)


franz1981:
@clebertsuconic Please don't merge it yet:I need reviews and some help from you and @wy96f and @qihongxu :+1

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
```diff
@@ -120,105 +110,121 @@ public void setLiveCache(LivePageCache pageCache) {
...
-                     // Sanity Check: This would only happen if there is a bug on decode or any internal code, as
-                     // this
-                     // constraint was already checked
-                     throw new IllegalStateException("Internal error, it wasn't possible to locate END_BYTE " + b);
-                  }
+   private List<PagedMessage> readFromSequentialFile(StorageManager storage) throws Exception {
+      final List<PagedMessage> messages = new ArrayList<>();
+      final int fileSize = (int) file.size();
+      file.position(0);
+      final int minimumMessagePersistentSize = DataConstants.SIZE_INT + 2;
```
clebertsuconic: This is a constant. Just declare it as a private static final MINIMUM_MESSAGE_PERSISTENT_SIZE on the class, and stop passing arguments. For simplicity.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
```diff
@@ -120,105 +110,121 @@ public void setLiveCache(LivePageCache pageCache) {
...
+      int position = 0;
+      ByteBuffer fileBuffer = null;
+      try {
+         while ((fileSize - position) >= minimumMessagePersistentSize) {
+            if (fileBuffer == null) {
+               fileBuffer = fileFactory.newBuffer(Math.min((fileSize - position), Env.osPageSize()));
```
clebertsuconic: I wouldn't declare a buffer here. Make the instantiantion inside readIntoFileBuffer

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
```diff
@@ -120,105 +110,121 @@ public void setLiveCache(LivePageCache pageCache) {
...
-                  }
+   private List<PagedMessage> readFromSequentialFile(StorageManager storage) throws Exception {
+      final List<PagedMessage> messages = new ArrayList<>();
+      final int fileSize = (int) file.size();
+      file.position(0);
+      final int minimumMessagePersistentSize = DataConstants.SIZE_INT + 2;
+      int position = 0;
+      ByteBuffer fileBuffer = null;
+      try {
+         while ((fileSize - position) >= minimumMessagePersistentSize) {
```
clebertsuconic: What was wrong with the previous reading logic, where we would just check for START_BYTE?

This might be right, but did you do enough tests on this to make sure this is ok?

I'm afraid of changing logic here on a years old tested code for no reason. If you found a reason for this I would like to know.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
```diff
@@ -120,105 +110,121 @@ public void setLiveCache(LivePageCache pageCache) {
...
+                  fileBuffer.position(endPosition + 1);
+                  assert fileBuffer.get(endPosition) == Page.END_BYTE : "decoding cannot change end byte";
                   msg.initMessage(storage);
                   if (logger.isTraceEnabled()) {
-                     logger.trace("Reading message " + msg + " on pageId=" + this.pageId + " for address=" + storeName);
+                     logger.tracef("Reading message %s on pageId=%d for address=%s", msg, pageId, storeName);
                   }
-                  readMessages++;
                   messages.add(msg);
+                  position += minimumMessagePersistentSize + encodedSize;
```
clebertsuconic: I am a bit concerned about the new reading buffer logic.

You were supposed to fix the barrier into how mmap is reading, but you're introducing a new reading logic in the mix. It's a bit hard to review it here.

I have spent months testing this before the first release of paging was out. I'm just afraid about the amount of testing done into the new logic.

clebertsuconic:
so, in summary, I'm not sure what is the TTSP fix, and what is the logic change you made in the mix here.

clebertsuconic:
Can  you simplify the change somehow it only fix the TTSP. If you still feel strong about the loop logic change, can you make it a separate commit? 

This way would be better to eventually bisect eventual issues.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
franz1981: But I'm checking START_BYTE as before, the logic on the single message hasnìt changed at all.
In the new logic I'm moving a window of ByteBuffer on the file regions in order to be able to reduce I/O and buffer copies/allocations as much as possible.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
franz1981: I understand but the point of this PR is to maintain a similar speed of mmap without incurring into the time-to-safepoint issue; that's why the logic has changed.

The original implementation (pre-mmap) was instantiating for each file a giant direct `ByteBuffer`, impacting on the overall native memory footprint and causing a long context switch while making the kernel fill that buffer from its OS (file) page cache.
It has been solved using mmap, to avoid allocations and copies, but `MappedByteBuffer::get(byte)` could pause the JVM for long time: that's why I'm addressing both the issues at once: memory footprint and long time to safepoints (to save memory).

franz1981: > I have spent months testing this before the first release of paging was out. I'm just afraid about the amount of testing done into the new logic.

I'm testing the logic with the many tests we have on paging and I'm sure I will get the same robustness as before :+1:

franz1981:
> Can you simplify the change somehow it only fix the TTSP. If you still feel strong about the loop logic change, can you make it a separate commit?

Sadly the loop logic using `FileChanne::readl` is the solution of the issue: so I don't think it can be splitted, but I will cover as much as I can in order to make it robust as it was before.
Re the logic I will make it more clear with comments, in order to be future proof :+1:

franz1981:
The CI has shown good results: is basically on par of master :+1: 
Now my concern is just to measure/improve performance.
The current concerns I've are:

- reducing the number of new ChannelBufferWrapper (but maybe escape analysis won't allocate them)
- find the correct read chunk size to maximize disk

I remember  @wy96f and @qihongxu were interested into paging performance, so maybe can take a look

clebertsuconic:
@franz1981 I tested this, and this is a huge improvement over paging.

I tested a variation of 4k, 100K, 2M on the buffer and did not make a difference, I was always able to receive 500K messages in 42 seconds on my laptop, each message sizing 12K.

Using the current version this was always taking more time than I could afford on waiting.. I did not even bother about measuring it.

michaelandrepearce:
I haven't had chance to test it, but based on @clebertsuconic's results posted here, thats really good, great stuff @franz1981

wy96f:
> @franz1981 I tested this, and this is a huge improvement over paging.
> 
> I tested a variation of 4k, 100K, 2M on the buffer and did not make a difference, I was always able to receive 500K messages in 42 seconds on my laptop, each message sizing 12K.
> 
> Using the current version this was always taking more time than I could afford on waiting.. I did not even bother about measuring it.

What did you mean buffer size? And what was the page size in your test?

qihongxu:
@franz1981 Thanks for contribution to page performance issues!
I have deployed a test as soon as I saw this thread.

Our configuration:
4C/8G/500G SSD single broker as usual
200 threads of consumers to squeeze out all potentials
Most options in broker.xml remained default except for changing journal-type to MAPPED

Then we tested its performance in two versions (one with this patch and the other not)
Producers and consumers ran separately with each message sizing 4k. TPS is as below:

  | Producer | Consumer
-- | -- | --
Without patch | 13k | 15k
With patch | 13k | 14.8k

It seemed that consumer tps in two versions were very close.

We also tried raising page size from 10M to 100M, raising message size form 4k to 12k, etc. But none of these made a significant difference in tps between two versions.

Flame graph for version with patch is as below:
image

@franz1981 Would you please give some hints on the differences between this and @clebertsuconic 's result? Is there any configuration that we missed in tests?

franz1981:
@qihongxu I'm getting similar results to yours :) 
This pr together with   https://github.com/apache/activemq-artemis/pull/2646 should make numbers on par with the original version, but with an important difference: smaller time to safe point.
If you enable -XX:+PrintGCApplicationStoppedTime, with the patched version we won't have long pauses anymore they could lead the broker to became unresponsive. 
It should happen with slow disk and/or OS page cache that won't contain the requested OS page (ie causing major page fault).
I would say a very stressing scenario, not easy to test..
If you have any means to bench it in a similar scenario (with https://github.com/apache/activemq-artemis/pull/2646) would be awesome!
Many thanks for taking a look at it!

franz1981:
@qihongxu just curious: which value are you configured for page-max-concurrent-io?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
```diff
@@ -120,105 +110,133 @@ public void setLiveCache(LivePageCache pageCache) {
          throw ActiveMQMessageBundle.BUNDLE.invalidPageIO();
       }
 
-      final List<PagedMessage> messages = new ArrayList<>();
-
       size.lazySet((int) file.size());
 
-      if (this.canBeMapped) {
-         readFromMapped(storage, messages);
```
wy96f: The original readFromSequentialFile() allocates file size memory and may cause long TTSP. 
But readFromMapped() only calls mmap64() and gets starting address of the mapping which should not cause long TTSP. Why we not reserve mmap file read here?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
```diff
@@ -99,8 +91,6 @@ public Page(final SimpleString storeName,
       fileFactory = factory;
       this.storageManager = storageManager;
       this.storeName = storeName;
-      this.canBeMapped = fileFactory instanceof NIOSequentialFileFactory || fileFactory instanceof MappedSequentialFileFactory;
```
wy96f: For the original version, If we use NIOSequentialFileFactory, why we use mapped file read instead? Shouldn't it be consistent with the fileFactory type?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
franz1981: The long TTSP issue of the mapped version wasn't caused during the cheap mmap64, but while reading the MappedByteBuffer, on each new OS page that wasn't mapped onto the OS page cache.
The original version were having other issues eg memory footprint...but not long TTSP.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
franz1981: In the original version it was a "cheap trick" to use NIO for write side, but mapped on read side

qihongxu:
@franz1981 Thanks for your reply! I have updated test results as we changed message size from 4k to 12k and 50k (keep msg_size between os_pagesize and largeMsg_threshold). It started to show a 10-20% differences in tps between two versions' consumer.


  | Msg size 4k | Msg size 12k | Msg size 50k
-- | -- | -- | --
Without patch | 15.0k | 7.0k | 1.7k
With patch | 14.5k | 8.7k | 2.0k


> @qihongxu just curious: which value have you configured for page-max-concurrent-io?

We just let it remain default to 5.

franz1981:
@qihongxu that's super interesting :O
TBH I was expecting the patched version to be on par with very small messages instead, because the most of the cost (just guessing eh) was on encoding side, while with big buffers I was expecting that the zero-copy approach of mmap was unbeatable: it would be interesting to check where is the cost of the non-patched version with such big messages too.
I suppose that small messages are the common case so, do you have some test result with smaller messages (and including the GC optimizations on https://github.com/apache/activemq-artemis/pull/2646)?

Re 
> We just let it remain default to 5.

Why not making it totally unbounded and let the SSD (iotop -o) tell you if I/O is maximized?
If GC will occur, I'm expecting that Artemis PageCache will be cleaned up of several cached pages and limiting it just to 5 concurrent requests means could be quite limiting with very fast ssd and many cores servers...
https://github.com/apache/activemq-artemis/pull/2645 could be a game changes with many concurrent page read requests :)

franz1981:
@clebertsuconic @wy96f @qihongxu 
I'm aware that this patch introduce a (small) perf regression in some cases (due to the added unecessary copy), but I don't know ATM any other ways to avoid the long TTSP issues: a "simple" (bust costy in term of dev/test time) way to improve the situation would be to expose a JNA `mlock` call to allow *at least* that while reading a OS page the OS will swap out randomly OS pages due to other threads interacting with the OS page cache eg the MAPPED journal, that tends to reuse the same files, if compaction is working as expected.
It won't solve the TTSP issue, but it would improve the situation on busy servers, leaving just major/minor page faults to be the other causes for long TTSP.

clebertsuconic:
>  	Msg size 4k	Msg size 12k	Msg size 50k
> Without patch	15.0k	7.0k	1.7k
> With patch	14.5k	8.7k	2.0k


Are you sure there is a regression here?

on these small amounts on my experience, if you repeat the same test within itself you may get one run faster or shorter depending on many factors (such as Hotspot, OS changing priorities... etc).

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
wy96f: Got it, thanks for the elaboration :)

michaelandrepearce:
I agree. It looks like perf is on par to me. That said performance never should be at the cost of correctness or stability. As such even if a slight drop. For me this is a good pr as its ensuring stability.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
franz1981: I forgot to answer you on

> Good job. By the way, How did you find the issue?

Using clustering we have measured such long time to safepoint pauses that was causing the broker to become unresponsive so I've investigated on them using an approach similar to https://www.mail-archive.com/mechanical-sympathy@googlegroups.com/msg01122.html.
I have found that `MappedByteBuffer::get` was making the CPU to get descheduled for several seconds, making the whole JVM to freeze. 
The filesystem used was XFS so the reason behind such behaviour wasn't dependent by any ext4 journalling daemon, but due to major page faults + I/O activity to fill the OS page cache.
As I've said: assuming infinite RAM and the fastest disk ever with XFS, this issue is very unlikely to happen...

PR #2634, opened 2019-04-23
ARTEMIS-2316 AMQPMessage missing ApplicationProperties from custom transformer

ldebello:
This PR is to solve the issue https://issues.apache.org/jira/browse/ARTEMIS-2316

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
@@ -1223,7 +1223,9 @@ public class AMQPMessage extends RefCountMessage {
 
    @Override
    public Object removeProperty(String key) {
-      return getApplicationPropertiesMap(false).remove(key);
+      Object removed = getApplicationPropertiesMap(false).remove(key);
+      messageChanged();
+      return removed;
    }
 
    @Override
@@ -1395,60 +1397,70 @@ public class AMQPMessage extends RefCountMessage {
    @Override
    public org.apache.activemq.artemis.api.core.Message putBooleanProperty(String key, boolean value) {
       getApplicationPropertiesMap(true).put(key, Boolean.valueOf(value));
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putByteProperty(String key, byte value) {
       getApplicationPropertiesMap(true).put(key, Byte.valueOf(value));
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putBytesProperty(String key, byte[] value) {
       getApplicationPropertiesMap(true).put(key, value);
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putShortProperty(String key, short value) {
       getApplicationPropertiesMap(true).put(key, Short.valueOf(value));
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putCharProperty(String key, char value) {
       getApplicationPropertiesMap(true).put(key, Character.valueOf(value));
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putIntProperty(String key, int value) {
       getApplicationPropertiesMap(true).put(key, Integer.valueOf(value));
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putLongProperty(String key, long value) {
       getApplicationPropertiesMap(true).put(key, Long.valueOf(value));
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putFloatProperty(String key, float value) {
       getApplicationPropertiesMap(true).put(key, Float.valueOf(value));
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putDoubleProperty(String key, double value) {
       getApplicationPropertiesMap(true).put(key, Double.valueOf(value));
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putBooleanProperty(SimpleString key, boolean value) {
       getApplicationPropertiesMap(true).put(key.toString(), Boolean.valueOf(value));
+      messageChanged();
       return this;
    }
 
@@ -1495,12 +1507,14 @@ public class AMQPMessage extends RefCountMessage {
    @Override
    public org.apache.activemq.artemis.api.core.Message putStringProperty(String key, String value) {
       getApplicationPropertiesMap(true).put(key, value);
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putObjectProperty(String key, Object value) throws ActiveMQPropertyConversionException {
       getApplicationPropertiesMap(true).put(key, value);
+      messageChanged();
       return this;
    }
 
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/BridgeApplicationPropertiesTransformer.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/BridgeApplicationPropertiesTransformer.java
@@ -1,0 +1,32 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.tests.integration.amqp;
+
+import org.apache.activemq.artemis.api.core.Message;
+import org.apache.activemq.artemis.core.server.transformer.Transformer;
+
+public class BridgeApplicationPropertiesTransformer implements Transformer {
+
+   @Override
+   public Message transform(final Message message) {
+
+      message.putStringProperty("C", "3");
+      message.putStringProperty("D", "4");
+
+      return message;
+   }
+}
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/DivertApplicationPropertiesTransformer.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/DivertApplicationPropertiesTransformer.java
@@ -1,0 +1,34 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.tests.integration.amqp;
+
+import org.apache.activemq.artemis.api.core.Message;
+import org.apache.activemq.artemis.core.server.transformer.Transformer;
+
+public class DivertApplicationPropertiesTransformer implements Transformer {
+
+   public static final String TRX_ID = "trxId";
+
+   @Override
+   public Message transform(final Message message) {
+
+      message.putStringProperty("A", "1");
+      message.putStringProperty("B", "2");
+
+      return message;
+   }
+}
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpBridgeApplicationProperties.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpBridgeApplicationProperties.java
@@ -1,0 +1,121 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.tests.integration.amqp;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.apache.activemq.artemis.api.core.RoutingType;
+import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.api.core.client.ActiveMQClient;
+import org.apache.activemq.artemis.api.core.client.ClientConsumer;
+import org.apache.activemq.artemis.api.core.client.ClientMessage;
+import org.apache.activemq.artemis.api.core.client.ClientSession;
+import org.apache.activemq.artemis.api.core.client.ClientSessionFactory;
+import org.apache.activemq.artemis.api.core.client.ServerLocator;
+import org.apache.activemq.artemis.core.config.BridgeConfiguration;
+import org.apache.activemq.artemis.core.config.DivertConfiguration;
+import org.apache.activemq.artemis.core.config.TransformerConfiguration;
+import org.apache.activemq.artemis.core.server.ActiveMQServer;
+import org.apache.activemq.artemis.core.server.ComponentConfigurationRoutingType;
+import org.junit.Before;
+import org.junit.Test;
+
+public class AmqpBridgeApplicationProperties extends AmqpClientTestSupport {
+
+   private ActiveMQServer server0;
+   private ActiveMQServer server1;
+
+   private SimpleString customNotificationQueue;
+   private SimpleString frameworkNotificationsQueue;
+   private SimpleString bridgeNotificationsQueue;
+   private SimpleString notificationsQueue;
+
+   private String getServer0URL() {
+      return "tcp://localhost:61616";
+   }
+
+   private String getServer1URL() {
+      return "tcp://localhost:61617";
+   }
+
+   @Override
+   public URI getBrokerAmqpConnectionURI() {
+      try {
+         return new URI(getServer0URL());
+      } catch (URISyntaxException e) {
+         throw new RuntimeException(e);
+      }
+   }
+
+   @Override
+   @Before
+   public void setUp() throws Exception {
+      super.setUp();
+      server0 = createServer(false, createBasicConfig());
+      server1 = createServer(false, createBasicConfig());
+
+      server0.getConfiguration().addAcceptorConfiguration("acceptor", getServer0URL());
+      server0.getConfiguration().addConnectorConfiguration("notification-broker", getServer1URL());
+      server1.getConfiguration().addAcceptorConfiguration("acceptor", getServer1URL());
+
+      DivertConfiguration customNotificationsDivert = new DivertConfiguration().setName("custom-notifications-divert").setAddress("*.Provider.*.Agent.*.CustomNotification").setForwardingAddress("FrameworkNotifications").setExclusive(true).setTransformerConfiguration(new TransformerConfiguration(DivertApplicationPropertiesTransformer.class.getCanonicalName()));
+      DivertConfiguration frameworkNotificationsDivert = new DivertConfiguration().setName("framework-notifications-divert").setAddress("BridgeNotifications").setForwardingAddress("Notifications").setRoutingType(ComponentConfigurationRoutingType.MULTICAST).setExclusive(true);
+
+      server0.getConfiguration().addDivertConfiguration(customNotificationsDivert);
+      server1.getConfiguration().addDivertConfiguration(frameworkNotificationsDivert);
+
+      customNotificationQueue = SimpleString.toSimpleString("*.Provider.*.Agent.*.CustomNotification");
+      frameworkNotificationsQueue = SimpleString.toSimpleString("FrameworkNotifications");
+      bridgeNotificationsQueue = SimpleString.toSimpleString("BridgeNotifications");
+      notificationsQueue = SimpleString.toSimpleString("Notifications");
+
+      server0.start();
+      server1.start();
+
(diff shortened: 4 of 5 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/DivertImpl.java
```diff
@@ -122,6 +122,7 @@ public void route(final Message message, final RoutingContext context) throws Ex
 
          if (transformer != null) {
             copy = transformer.transform(copy);
+            copy.reencode();
```
michaelandrepearce: This should be responsibility of the transformer code.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
```diff
@@ -559,6 +559,7 @@ protected Message beforeForward(final Message message, final SimpleString forwar
                                " as transformedMessage");
             }
          }
+         transformedMessage.reencode();
```
michaelandrepearce: This is responsibility of transformer code.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/BridgeApplicationPropertiesTransformer.java
```diff
@@ -0,0 +1,32 @@
...
+
+public class BridgeApplicationPropertiesTransformer implements Transformer {
+
+   @Override
+   public Message transform(final Message message) {
+
+      message.putStringProperty("C", "3");
+      message.putStringProperty("D", "4");
+
+      return message;
```
michaelandrepearce: This test should call .re-encode if it is altering the message and wishes the original message to re-encode. Typically a copy would be made also when doing a transform, to avoid modifying the underlying original message, which could be shared by several ref's

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/DivertApplicationPropertiesTransformer.java
```diff
@@ -0,0 +1,34 @@
...
+
+public class DivertApplicationPropertiesTransformer implements Transformer {
+
+   public static final String TRX_ID = "trxId";
+
+   @Override
+   public Message transform(final Message message) {
+
+      message.putStringProperty("A", "1");
+      message.putStringProperty("B", "2");
```
michaelandrepearce: This test should call .reencode() it is responsibility of transformer,  if it is altering the message and wishes the original message to re-encode. Typically a copy would be made also when doing a transform, to avoid modifying the underlying original message, which could be shared by several ref's

michaelandrepearce:
@ldebello since https://issues.apache.org/jira/browse/ARTEMIS-2316 it is responsibility for a transformer if it is modifying the message to call reencode. This is because not all transformers may modify the message, and only the particular transformer impl knows if it is modifying or not. Like wise its also good practice if you are transforming to copy, to avoid modifying the underlying message, which can be shared by many message ref's on different queues.

ldebello:
@michaelandrepearce I totally understand that but that is a regression regarding 2.6.3 where this behaviour is out of the box I can fix my use case adding the reencode, but it sound weird

ldebello:
Also, I am not an expert of using transform but at least for me if you are using a transformer is because you will modify the message, so it seems so many use cases will have to add this reencode? and let me change my previous statement from "sound weird" to "sound wrong to me"

michaelandrepearce:
Actually some transformers don't e.g. where a transformer is being used complex  filter e.g. check for something and return null message if message shouldn't be further processed.

ldebello:
@michaelandrepearce I was not aware of that usecase, my code does not check for null so that is an issue, but I don't how you handle regressions in this case, the original behaviour was reencode and that was change in 2.6.4, it was a side effect of an improvement in performance but the behaviour was there. Imaging our use case moving from 2.6.3 to 2.6.4 and getting this error nobody expects this kind of changes in a patch release. Also if you see the test AddHeadersTransformer it is adding properties but not calling reencode() so at least we should document this requirement in docs and in the code as well

michaelandrepearce:
So a reencode is quite heavy, you want to avoid doing this over and over again.  Thus some of the work you link to was addressing. I think possibly the issue here is you're calling putXProperty (replace x with Int Long String etc method varients) but its not marking the message as changed, if you see CoreMessage it invokes messageChanged for each of those methods, but AMQPMEssage doesnt. This maybe a better solution....

michaelandrepearce:
e.g.


```
   @Override
   public org.apache.activemq.artemis.api.core.Message putStringProperty(String key, String value) {
      getApplicationPropertiesMap(true).put(key, value);
      return this;
   }
```

to 

```
   @Override
   public org.apache.activemq.artemis.api.core.Message putStringProperty(String key, String value) {
      getApplicationPropertiesMap(true).put(key, value);
      messageChanged();
      return this;
   }
```

michaelandrepearce:
That said the application properties in AMQP are not meant to be modified, so you're breaking spec in doing that.

ldebello:
At first I tried that solution but for some reason fails, because the application properties are not in the proton message so doing `if (applicationProperties != null) getProtonMessage().setApplicationProperties(applicationProperties);` works, but if you include several properties you are doing this several times, that is the reason why I prefer to do this once after leaving the transformer, also removing should be handle and giving the fact that was more error prone I choose the error way. but could be a solution, modifying the proton message as well.

michaelandrepearce:
getApplicationPropertiesMap(true) is meant to create if absent, you shouldn't need extra check.

michaelandrepearce:
It should be noted here, application properties in AMQP are not meant to be modified, so you're breaking spec in doing that. This is fine for your custom transformer to do that but then you need to handle that in your transformer, especially as there could be introduced concurrency issues if modifying the underlying shared message...

ldebello:
The application properties already exist because the message bring one properties from the client, I did the change and it didnt work if you only call messageChanged(), I would try to check it again but I did that test

michaelandrepearce:
```

   @SuppressWarnings("unchecked")
   private Map<String, Object> getApplicationPropertiesMap(boolean createIfAbsent) {
      ApplicationProperties appMap = lazyDecodeApplicationProperties();
      Map<String, Object> map = null;

      if (appMap != null) {
         map = appMap.getValue();
      }

      if (map == null) {
         if (createIfAbsent) {
            map = new HashMap<>();
            this.applicationProperties = new ApplicationProperties(map);
         } else {
            map = Collections.EMPTY_MAP;
         }
      }

      return map;
   }

```

ldebello:
> It should be noted here, application properties in AMQP are not meant to be modified, so you're breaking spec in doing that. This is fine for your custom transformer to do that but then you need to handle that in your transformer, especially as there could be introduced concurrency issues (thus safer to make copy first) if modifying the underlying shared message... which i would be cautious of adding to main broker.

I think that should explicit in docs and also in the tests and as I mentioned the test that you can find in the code is not doing that AddHeadersTransformer. also breaking that in a patch is like something unexpected, so I dont agree with custom transformer responsability at least for 2.6.4 it does not make sense.

michaelandrepearce:
So AddHeadersTransformer was made for working on CoreMessage.

ldebello:
> ```
> 
>    @SuppressWarnings("unchecked")
>    private Map<String, Object> getApplicationPropertiesMap(boolean createIfAbsent) {
>       ApplicationProperties appMap = lazyDecodeApplicationProperties();
>       Map<String, Object> map = null;
> 
>       if (appMap != null) {
>          map = appMap.getValue();
>       }
> 
>       if (map == null) {
>          if (createIfAbsent) {
>             map = new HashMap<>();
>             this.applicationProperties = new ApplicationProperties(map);
>          } else {
>             map = Collections.EMPTY_MAP;
>          }
>       }
> 
>       return map;
>    }
> ```

That code is from master the code in 2.6.x or 2.6.4 does not have that logic, if that fix the issue for messageChanged we should backport that change to 2.6.x

michaelandrepearce:
So we're now on 2.7.0 branch for fixes, and about to release 2.8.0, any change will be in newer versions

ldebello:
I suppose you would release 2.6.5 to fix a regression, isnt it? I mean if you define that is transformer responsability to call reencode for 2.7.0 ok , I don't agree but at least is a minor change (for me is more than that), but leaving 2.6.4 broken is a kind of unexpected.

michaelandrepearce:
I doubt anyone will be releasing any further 2.6.x, i certainly i have no intent personally to run a release on itunless there's a CVC.  Its like any 2.5.x or 2.4.x branch, its unlikely to have any further releases unless CVC.

You're welcome to maintain you're own fork and builds obviously.

ldebello:
ok, I would try to call the messageChanged() to see if that works in master.

ldebello:
@michaelandrepearce I have just checked the messageChanged() in master and it works because the logic is different from 2.6.x, I would do the change in the method and update this PR

michaelandrepearce:
PR should be against master.

michaelandrepearce:
@ldebello you need to close this pr, and open a new one against master.

ldebello:
@michaelandrepearce the PR was update and base branch change

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1395,128 +1397,162 @@ public SimpleString getSimpleStringProperty(String key) throws ActiveMQPropertyC
...
       getApplicationPropertiesMap(true).put(key.toString(), Boolean.valueOf(value));
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putByteProperty(SimpleString key, byte value) {
-      return putByteProperty(key.toString(), value);
+      org.apache.activemq.artemis.api.core.Message message = putByteProperty(key.toString(), value);
+      messageChanged();
```
michaelandrepearce: this isnt needed as the delegated method you have added messageChanged() to already, same for a number of methods.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1395,128 +1397,162 @@ public SimpleString getSimpleStringProperty(String key) throws ActiveMQPropertyC
...
+      return message;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putBytesProperty(SimpleString key, byte[] value) {
-      return putBytesProperty(key.toString(), value);
+      org.apache.activemq.artemis.api.core.Message message = putBytesProperty(key.toString(), value);
+      messageChanged();
```
michaelandrepearce: ditto

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1395,128 +1397,162 @@ public SimpleString getSimpleStringProperty(String key) throws ActiveMQPropertyC
...
+      org.apache.activemq.artemis.api.core.Message message = putShortProperty(key.toString(), value);
+      messageChanged();
+      return message;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putCharProperty(SimpleString key, char value) {
-      return putCharProperty(key.toString(), value);
+      org.apache.activemq.artemis.api.core.Message message = putCharProperty(key.toString(), value);
+      messageChanged();
```
michaelandrepearce: ditto

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1395,128 +1397,162 @@ public SimpleString getSimpleStringProperty(String key) throws ActiveMQPropertyC
...
+      return message;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putIntProperty(SimpleString key, int value) {
-      return putIntProperty(key.toString(), value);
+      org.apache.activemq.artemis.api.core.Message message = putIntProperty(key.toString(), value);
+      messageChanged();
```
michaelandrepearce: ditto

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1395,128 +1397,162 @@ public SimpleString getSimpleStringProperty(String key) throws ActiveMQPropertyC
...
+      return message;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putLongProperty(SimpleString key, long value) {
-      return putLongProperty(key.toString(), value);
+      org.apache.activemq.artemis.api.core.Message message = putLongProperty(key.toString(), value);
+      messageChanged();
```
michaelandrepearce: ditto.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1395,128 +1397,162 @@ public SimpleString getSimpleStringProperty(String key) throws ActiveMQPropertyC
...
+      org.apache.activemq.artemis.api.core.Message message = putFloatProperty(key.toString(), value);
+      messageChanged();
+      return message;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putDoubleProperty(SimpleString key, double value) {
-      return putDoubleProperty(key.toString(), value);
+      org.apache.activemq.artemis.api.core.Message message = putDoubleProperty(key.toString(), value);
+      messageChanged();
```
michaelandrepearce: ditto

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1395,128 +1397,162 @@ public SimpleString getSimpleStringProperty(String key) throws ActiveMQPropertyC
...
+      org.apache.activemq.artemis.api.core.Message message = putLongProperty(key.toString(), value);
+      messageChanged();
+      return message;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putFloatProperty(SimpleString key, float value) {
-      return putFloatProperty(key.toString(), value);
+      org.apache.activemq.artemis.api.core.Message message = putFloatProperty(key.toString(), value);
+      messageChanged();
```
michaelandrepearce: ditto

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1395,128 +1397,162 @@ public SimpleString getSimpleStringProperty(String key) throws ActiveMQPropertyC
...
       getApplicationPropertiesMap(true).put(key, value);
+      messageChanged();
       return this;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putObjectProperty(SimpleString key, Object value) throws ActiveMQPropertyConversionException {
-      return putObjectProperty(key.toString(), value);
+      org.apache.activemq.artemis.api.core.Message message = putObjectProperty(key.toString(), value);
+      messageChanged();
```
michaelandrepearce: ditto

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1395,128 +1397,162 @@ public SimpleString getSimpleStringProperty(String key) throws ActiveMQPropertyC
...
+      return message;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putStringProperty(SimpleString key, SimpleString value) {
-      return putStringProperty(key.toString(), value.toString());
+      org.apache.activemq.artemis.api.core.Message message = putStringProperty(key.toString(), value.toString());
+      messageChanged();
```
michaelandrepearce: ditto

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1395,128 +1397,162 @@ public SimpleString getSimpleStringProperty(String key) throws ActiveMQPropertyC
...
+      return message;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putStringProperty(SimpleString key, String value) {
-      return putStringProperty(key.toString(), value);
+      org.apache.activemq.artemis.api.core.Message message = putStringProperty(key.toString(), value);
+      messageChanged();
```
michaelandrepearce: ditto

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1526,7 +1562,9 @@ public SimpleString getLastValueProperty() {
 
    @Override
    public org.apache.activemq.artemis.api.core.Message setLastValueProperty(SimpleString lastValueName) {
-      return putStringProperty(HDR_LAST_VALUE_NAME, lastValueName);
+      org.apache.activemq.artemis.api.core.Message message = putStringProperty(HDR_LAST_VALUE_NAME, lastValueName);
+      messageChanged();
```
michaelandrepearce: ditto

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1395,128 +1397,162 @@ public SimpleString getSimpleStringProperty(String key) throws ActiveMQPropertyC
...
+      org.apache.activemq.artemis.api.core.Message message = putBytesProperty(key.toString(), value);
+      messageChanged();
+      return message;
    }
 
    @Override
    public org.apache.activemq.artemis.api.core.Message putShortProperty(SimpleString key, short value) {
-      return putShortProperty(key.toString(), value);
+      org.apache.activemq.artemis.api.core.Message message = putShortProperty(key.toString(), value);
+      messageChanged();
```
michaelandrepearce: ditto

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
ldebello: Done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
ldebello: Done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
ldebello: Done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
ldebello: Done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
ldebello: Done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
ldebello: Done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
ldebello: Done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
ldebello: Done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
ldebello: Done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
ldebello: Done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
ldebello: Done

ldebello:
@michaelandrepearce all comments done, thanks for catch that :), let me know if another change is necessary.

michaelandrepearce:
@ldebello please squash your commits.

ldebello:
@michaelandrepearce done

gemmellr:
This feels like it might be the wrong change.

To my knowledge the reencode method is expected to be used explicitly in cases like this, and thats actually the main reason it exists at all and was added originally.

The commit pointed to as the earlier change in behaviour for 2.6.4 was clearly the correct change to make for that issue, if anything it would seem the original behaviour was incorrect/unexpected. Changing 2.8.0 to differ from 2.7.0 and 2.6.4 (doing whats expected) just to restore a 2.6.3 behaviour that perhaps shouldn't have been there isnt seeming all that obvious to me. It would also now become inconsistent amongst itself regarding what happens with other modifications to the message.

tabish121:
In the changes that were done previously the modified calls were intentionally left out as they lead to silently allowing AMQP specification breaking changes happen on the message instead of forcing the user to manually re-encode once they've done something they mostly likely shouldn't have in the first place.

ldebello:
I partially agree with both, because I think the behaviour should be consistent at least during a patch release, although there was a error, some kind of "bug feature" because the fix which generate the fix was about performance, I think it was a collateral effect and not a real fix for that specific use case.
Having said that if the idea is to force the use to call reencode if they modify the message, I think that should be changed from 2.7.0 onwards but only reverting this commit does not solve all the cases, because if you send a message, inject properties, does not call reencode and consume the message from the same broker the properties are there, which is weird, this issue arise when the bridge has to serialize the message and I suppose if you have too many message and pagination start working maybe the "issue" is present as well.

gemmellr:
The original referenced commit was actually made to resolve a message corruption issue present in 2.6.3, per the commit message and JIRA it was made against. The commit does also reference perf admittedly, but I expect that was simply because it seemed obvious there was waste overhead in cases from repeated encoding due to the flag not having been reset when it should have been.

clebertsuconic:
@gemmellr / @ldebello I think we should revert this change. The reencode method was meant for this.

gemmellr:
I think thats probably wise for now, at least for 2.8.0 which was originally meant to be under vote before this was raised. If a change is done then taking some more time on it seems like a good idea vs dropping it in just before release.

clebertsuconic:
I'm reverting the fix on AMQPMessage, but keeping the test, after doing some organization (name changes, moving classes), and calling the reencode method.

ldebello:
Great np, I will include reencode when upgrading to new version :)

PR #2635, opened 2019-04-23
NO-JIRA Fix issue where bridge is modifying underlying message

michaelandrepearce:


clebertsuconic:
I believe this will have issues with the ClusteredBridge. I can run the whole testsuite before merging this. did you have any time to do a full run?

clebertsuconic:
@michaelandrepearce You gave up on this one?

michaelandrepearce:
Ran out of time, and havent created a unit. Issue is real, basically bridge modifies underlying so if multiple queues and one queue has bridge, the bridge is modifying the shared message, meaning messages consumed on the other queue are modified.

PR #2636, opened 2019-04-23
ARTEMIS-2296 PgResultSet.updateBlob not implemented

jbertram:
There is no test for this as we don't have a way to embed/test Postgres.
It will need to be verified externally. However, given the exception the
fix looks solid.


michaelandrepearce:
Have you looked at

https://github.com/opentable/otj-pg-embedded

Or
https://www.testcontainers.org/modules/databases/postgres/

As a way to run postgres for unit testing

rvsoni:
Hi, 

I don't have exact steps to reproduce of this case, I have saved PostgreSQL DB offline to test this case.

it does something related to pagination on the queue on server startup and throws an exception mention in ticket ARTEMIS-2296.

Note: I Reported this ticket and submitted patch too.
Thanks
Ravi

clebertsuconic:
I will merge this, but it would be nice to add a JIRA to add tests.

PR #2639, opened 2019-04-24
ARTEMIS-2318 HornetQ Protocol Manager should copy the messages before…

clebertsuconic:
… replacing message properties


artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
```diff
@@ -355,6 +355,8 @@ public void referenceOriginalMessage(final Message original, String originalQueu
 
    @Override
    public Message copy() {
+      new Exception("hmmmmm").printStackTrace();
+      System.exit(-1);
```
franz1981: @clebertsuconic System.exit(-1)?

PR #2640, opened 2019-04-25
ARTEMIS-2310 support system prop sub in xincludes

jbertram:
Historically the broker has read the XML configuration file as a String,
substituted system properties, and then parsed that String into an XML
document. However, this method won't substitute system properties in the
files which are imported via xinclude. In order to substitue system
properties in xincluded files the substitution needs to be performed
after the file is parsed into an XML document. This commit implements
that change and refactors the XMLUtil class a bit to eliminate redundant
code, obsolete comments, etc.

### DIFF
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationTest.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationTest.java
@@ -64,6 +64,13 @@ import org.junit.Test;
 
 public class FileConfigurationTest extends ConfigurationImplTest {
 
+   static {
+      System.setProperty("a2Prop", "a2");
+      System.setProperty("falseProp", "false");
+      System.setProperty("trueProp", "true");
+      System.setProperty("ninetyTwoProp", "92");
+   }
+
    protected String getConfigurationName() {
       return "ConfigurationTest-full-config.xml";
    }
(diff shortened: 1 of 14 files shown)


artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationTest.java
```diff
@@ -64,6 +64,13 @@
 
 public class FileConfigurationTest extends ConfigurationImplTest {
 
+   static {
```
clebertsuconic: Please, do not leave stuff after tests.

There's a rule on ActiveMQTestBae to cleanup properties, and that's probably why your test failed here.

So, you should use a BeforeClass and AFterClass to cleanup such thing.


I'm adding here and squashing the change as yours.

clebertsuconic: I am adding this code:


```java
   @BeforeClass
   public static void setupProperties() {
      System.setProperty("a2Prop", "a2");
      System.setProperty("falseProp", "false");
      System.setProperty("trueProp", "true");
      System.setProperty("ninetyTwoProp", "92");
   }

   @AfterClass
   public static void clearProperties() {
      System.clearProperty("a2Prop");
      System.clearProperty("falseProp");
      System.clearProperty("trueProp");
      System.clearProperty("ninetyTwoProp");
   }

```

clebertsuconic: @jbertram I added a checkstyle issue during my merge. My next PR will fix it

PR #2645, opened 2019-04-26
ARTEMIS-2321 Paging GC improvements

franz1981:
It contains several improvements:

- PageCursorProviderImpl can use primitive maps
- PageCache doesn't need a Page reference
- Removed unnecessary volatile/Atomic operations and fields
- PageCursorInfo caches number of msgs to save Page::read

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/SoftValueLongObjectHashMap.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/SoftValueLongObjectHashMap.java
@@ -21,26 +21,26 @@ import java.lang.ref.SoftReference;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Comparator;
-import java.util.HashMap;
-import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeSet;
-import java.util.concurrent.atomic.AtomicLong;
 
+import io.netty.util.collection.LongObjectHashMap;
+import io.netty.util.collection.LongObjectMap;
 import org.jboss.logging.Logger;
 
-public class SoftValueHashMap<K, V extends SoftValueHashMap.ValueCache> implements Map<K, V> {
+public class SoftValueLongObjectHashMap<V extends SoftValueLongObjectHashMap.ValueCache> implements LongObjectMap<V> {
 
-   private static final Logger logger = Logger.getLogger(SoftValueHashMap.class);
+   private static final Logger logger = Logger.getLogger(SoftValueLongObjectHashMap.class);
 
    // The soft references that are already good.
    // too bad there's no way to override the queue method on ReferenceQueue, so I wouldn't need this
    private final ReferenceQueue<V> refQueue = new ReferenceQueue<>();
 
-   private final Map<K, AggregatedSoftReference> mapDelegate = new HashMap<>();
+   private final LongObjectMap<AggregatedSoftReference<V>> mapDelegate = new LongObjectHashMap<>();
 
-   private final AtomicLong usedCounter = new AtomicLong(0);
+   private long usedCounter = 0;
 
    private int maxElements;
 
@@ -57,7 +57,7 @@ public class SoftValueHashMap<K, V extends SoftValueHashMap.ValueCache> implemen
 
    // Constructors --------------------------------------------------
 
-   public SoftValueHashMap(final int maxElements) {
+   public SoftValueLongObjectHashMap(final int maxElements) {
       this.maxElements = maxElements;
    }
 
@@ -96,8 +96,7 @@ public class SoftValueHashMap<K, V extends SoftValueHashMap.ValueCache> implemen
     */
    @Override
    public boolean containsKey(final Object key) {
-      processQueue();
-      return mapDelegate.containsKey(key);
+      return containsKey(objectToKey(key));
    }
 
    /**
@@ -107,7 +106,7 @@ public class SoftValueHashMap<K, V extends SoftValueHashMap.ValueCache> implemen
    @Override
    public boolean containsValue(final Object value) {
       processQueue();
-      for (AggregatedSoftReference valueIter : mapDelegate.values()) {
+      for (AggregatedSoftReference<V> valueIter : mapDelegate.values()) {
          V valueElement = valueIter.get();
          if (valueElement != null && value.equals(valueElement)) {
             return true;
@@ -117,16 +116,31 @@ public class SoftValueHashMap<K, V extends SoftValueHashMap.ValueCache> implemen
       return false;
    }
 
+   private static long objectToKey(Object key) {
+      return ((Long) key).longValue();
+   }
+
    /**
     * @param key
     * @see java.util.Map#get(java.lang.Object)
     */
    @Override
    public V get(final Object key) {
+      return get(objectToKey(key));
+   }
+
+   @Override
+   public V put(Long key, V value) {
+      return put(objectToKey(key), value);
+   }
+
+   @Override
+   public V get(long key) {
       processQueue();
-      AggregatedSoftReference value = mapDelegate.get(key);
+      AggregatedSoftReference<V> value = mapDelegate.get(key);
       if (value != null) {
-         value.used();
+         usedCounter++;
+         value.used(usedCounter);
          return value.get();
       } else {
          return null;
@@ -139,14 +153,26 @@ public class SoftValueHashMap<K, V extends SoftValueHashMap.ValueCache> implemen
     * @see java.util.Map#put(java.lang.Object, java.lang.Object)
     */
    @Override
-   public V put(final K key, final V value) {
+   public V put(final long key, final V value) {
       processQueue();
-      AggregatedSoftReference newRef = createReference(key, value);
-      AggregatedSoftReference oldRef = mapDelegate.put(key, newRef);
+      AggregatedSoftReference<V> newRef = createReference(key, value);
+      AggregatedSoftReference<V> oldRef = mapDelegate.put(key, newRef);
       checkCacheSize();
-      newRef.used();
+      usedCounter++;
+      newRef.used(usedCounter);
       if (oldRef != null) {
          return oldRef.get();
+      } else {
+         return null;
+      }
+   }
+
+   @Override
+   public V remove(long key) {
+      processQueue();
+      AggregatedSoftReference<V> ref = mapDelegate.remove(key);
+      if (ref != null) {
+         return ref.get();
       } else {
          return null;
       }
@@ -156,7 +182,7 @@ public class SoftValueHashMap<K, V extends SoftValueHashMap.ValueCache> implemen
       if (maxElements > 0 && mapDelegate.size() > maxElements) {
          TreeSet<AggregatedSoftReference> usedReferences = new TreeSet<>(new ComparatorAgregated());
 
-         for (AggregatedSoftReference ref : mapDelegate.values()) {
+         for (AggregatedSoftReference<V> ref : mapDelegate.values()) {
             V v = ref.get();
 
             if (v != null && !v.isLive()) {
@@ -169,7 +195,7 @@ public class SoftValueHashMap<K, V extends SoftValueHashMap.ValueCache> implemen
                Object removed = mapDelegate.remove(ref.key);
 
                if (logger.isTraceEnabled()) {
-                  logger.trace("Removing " + removed + " with id = " + ref.key + " from SoftValueHashMap");
+                  logger.trace("Removing " + removed + " with id = " + ref.key + " from SoftValueLongObjectHashMap");
                }
 
                if (mapDelegate.size() <= maxElements) {
@@ -210,13 +236,7 @@ public class SoftValueHashMap<K, V extends SoftValueHashMap.ValueCache> implemen
     */
    @Override
    public V remove(final Object key) {
-      processQueue();
-      AggregatedSoftReference ref = mapDelegate.remove(key);
-      if (ref != null) {
-         return ref.get();
-      } else {
-         return null;
-      }
+      return remove(objectToKey(key));
    }
 
    /**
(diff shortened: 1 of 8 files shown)


franz1981:
The most relevant one from a functional point of view is the non-blocking Page::read on page cache.
I've maintained a separate commit for each improvement to make it easier to review them separately.
Although not really a bug I've recently looked into many stacktraces showing many threads blocked on soft cache due to slow `Page::read`, causing the broker to be unresponsive and suicide itself right after:
this change (together with https://github.com/apache/activemq-artemis/pull/2633) should make the broker less prone to became unresponsive under a storm of page cache misses.

franz1981:
@wy96f Please take a look :+1:

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
```diff
@@ -132,43 +137,57 @@ public PagedReference newReference(final PagePosition pos,
...
+               inProgressReadPage = readPage;
+               inProgressReadPages.put(pageId, readPage);
             }
          }
-
-         return cache;
+         if (createPage) {
+            return readPage(pageId, page, cache, inProgressReadPage);
+         } else {
+            return inProgressReadPage.get(10, TimeUnit.SECONDS);
```
clebertsuconic: this fixed timeout here is bad. I really like the logic...

However:

i - it should be configurable somehow. (it would be nice if there is already another value that would semantic apply in our configuration, but I don't think that will be the case here).
ii - a Timeout should be considered an IOException, and CriticalIOListener should be called on this case.

clebertsuconic:
Shouldn't these commits be squashed?


My only issue here is with an IOException and a Timeout on reading.

franz1981:
@clebertsuconic I've added detection for slow Page::read, but I'm not sure I'm using correcty the `CriticalAnalyzer`: let me know wdyt mate :+1:

franz1981:
I'm sure I need to add a test to cover slow page reading for both an awaiting thread and a performing one

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/SoftValueLongObjectHashMap.java
```diff
@@ -21,26 +21,26 @@
...
 
-   private static final Logger logger = Logger.getLogger(SoftValueHashMap.class);
+   private static final Logger logger = Logger.getLogger(SoftValueLongObjectHashMap.class);
 
    // The soft references that are already good.
    // too bad there's no way to override the queue method on ReferenceQueue, so I wouldn't need this
    private final ReferenceQueue<V> refQueue = new ReferenceQueue<>();
 
-   private final Map<K, AggregatedSoftReference> mapDelegate = new HashMap<>();
+   private final LongObjectMap<AggregatedSoftReference<V>> mapDelegate = new LongObjectHashMap<>();
```
franz1981: @michaelandrepearce Do you think it could be a problem to use it directly? I remember you've recently working with some primitive map of Netty...

franz1981:
@qihongxu @wf96q it should improve paging scalability over the same address (with topic for sure) a lot, especially with small heaps (and GC is cleaning up many requested PageCaches)

franz1981:
@clebertsuconic I have added a test using Mockito, but I was planning to add some more to cover some weird concurrent corner case too

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/SoftValueLongObjectHashMap.java
michaelandrepearce: Why do you need to use it directly? They implement map interface. And the add primitive methods that will be invoked at runtime, you dont need to make the field the concrete impl. This is the trick all primitive impl use.

michaelandrepearce:
Added comment inline

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/SoftValueLongObjectHashMap.java
franz1981: But I wanted to avoid the boxing/unboxing cost: the page cache (that uses this soft value map) is sensible to GC cycles and the more is your allocation rate, the more likely you will have a minor GC, making the cache to clean some of its entries...
Avoiding boxing/unboxing is just a way to let the entries stays much longer and makes this map more effective.
TBH I don't know (the details I mean) why we are using a soft value map here: I would have used a LRU map instead with a fixed size of entries instead like https://github.com/real-logic/agrona/blob/master/agrona/src/main/java/org/agrona/collections/IntLruCache.java : need to ask @clebertsuconic about it...

michaelandrepearce: It dossnt box and unbox. Thats the trick they use. By implementing the same interface method but with primitive param at runtime the primitive version is invoked.

franz1981: Probably I've misunderstood your comment Michael, sorry.
I have understood that if you:
```
Map<Long, V> map;
...
V put(Long k, V value) {
    return map.put(k, value);
}

```
In this case a caller that will invoke put(...) is forced to provide a Long, instead of a primitive `long`...no? Or am I missing something?

michaelandrepearce: SoftValue is so it can gc. But if it doesnt has to can keep as much in memory as possible. Rather than a strict bound

michaelandrepearce: Youre missing something. You missing the wonder of the jvm at runtime. It like you have a method in interface that takes object. Your impl has to support that but it can implement the same name with a tighter typing and that will be invoked instead.

franz1981: That's clear to me, but let me explain better what I mean...
If we have:
```
private final LongObjectMap<AggregatedSoftReference<V>> mapDelegate
```
and externally we provide:
```
//(1)
V put(long k, V value) {
    return mapDelegate.put(k, value);
}

//(2)
V put(Long k, V value) {
    return mapDelegate.put(k, value);
}
```
If caller will call (1), the javac will hide a boxing (long -> Long) to allow calling `map.put(k, value)` given that 
mapDelegate has been declared as `Map<Long,...`.

michaelandrepearce: No it doesnt

franz1981: Try to debug it and you will tell me bud :)
Or look at the bytecode ;)

michaelandrepearce: So i think im missing something here, if i have two methods, the tighter typed one will get invoked.

franz1981: @michaelandrepearce 
Try to put a debug print on `Long::valueOf(long)` using:
```
   private static final Map<Long, Object> primitiveMap = new LongObjectHashMap<>();

   public static void main(String[] args) {
      final Object value = new Object();
      for (long l = 0 ; l<10000;l++) {
         final long x = l^l;
         primitiveMap.put(x, value);
      }
   }
```

michaelandrepearce: Ive realised what ive done, ive missread 

"LongObjectMap" as "LongObjectHashMap" ... you using LongObjectMap as field value not LongObjectHashMap.

michaelandrepearce: youre all good.

michaelandrepearce:
LGTM

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
```diff
@@ -132,43 +156,88 @@ public PagedReference newReference(final PagePosition pos,
...
             }
          }
+         //We need to create a new criticalComponent each time, because
+         //each read page action can happen from a different thread
+         final CriticalComponent readPageActionComponent;
+         if (criticalAnalyzer == null) {
+            readPageActionComponent = null;
+         } else {
+            final SimpleString address = pagingStore.getAddress();
+            readPageActionComponent = new CriticalComponentImpl(criticalAnalyzer, PAGE_CURSOR_PROVIDER_CRITICAL_PATHS) {
```
wy96f: Do we need to cache the CriticalComponentImpl here to reduce memory footprint?

franz1981: I have no idea, maybe yes, but I don't want to make this code more complex then necessary: consider that on each `Page::read` request we already create 1 `CompleableFuture` and each completed read will produce several Paged messages as well: saving just 1 instance of CriticalComponentImpl maybe is not a super-win...
I'm more concerned from a functional point of view instead:  in this version I'm triggering the critical analyzer in both cases ie on Page::read and on who is waiting for it. 
Probably I can just leave the critical analyzer to monitor Page:.read and just logging otherwise.. @clebertsuconic wdyt?

clebertsuconic: I would expect the actual component to implement CriticalComponent

clebertsuconic: @franz1981 the implementation here is wrong I think..

you should have PageCursorProviderImpl extending CriticalComponentImpl.

Add it upon construction.. (check remove points when the address is removed)

and just call enter / leave...


like all the other implementations.

franz1981: I wasn't sure about it because you can have several concurrent enterCritical() calls on the same component,while I want the critical analyser to report which pageId read was too slow... 
Wdyt?

clebertsuconic: It wasn't engineered for that.. it would be too much information.

if the Critical Analyzer kicks in (remember, it wasn't supposed to), the relevant information is the stack trace and the methods where you have a possible deadlock.

franz1981: I was thinking to let Page extends CriticalComponentImpl and wrap the read with enter/exit..leaving the CompleteableFuture without it (it is not needed we monitor page read), given that we don't need to monitor both (will leave just the timeout logs).
I know that is not a perfect fit for the critical analyser ..but it could work?

clebertsuconic: @franz1981 you are reinventing things on the CriticalAnalyzer. I wouldn't go that route. Just stick to what's already there.

The CriticalAnalyzer is already complicated enough :)

The CriticalAnalyzer was intended to capture deadlocks. upon a critical call you call enter / leave on a finalize call. I strongly recommend against anything different.

it could work? I don't know.. you're using it in a different way than originally written, and I would not have enough time to test it.

clebertsuconic: The original design expected the component on the list as long as it lives. you are for instance adding and removing it like crazy on each write. I don't have an idea of what would be the implications. Definitely not being used as intended.

franz1981: Got it , but given they concurrent behaviour of such component if you have 1 slow page read and all the others fast enough, the slowest one won't be detected..
I have already a test in this PR to detect slow page read without concurrenct requests: I will create a new one just to check if my guess is correct, thanks!

clebertsuconic: @franz1981 yes, you're right... it won't work with multiple Threads.. I don't know what to do on that case.

perhaps we need multiple Components.. but I would on that case use a ThreadLocal or some other cache. I don't know.

clebertsuconic: Perhaps implement something on the PageCursor where isExpired would return false, and have individual measures using some sort of thread local?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
```diff
@@ -132,52 +146,99 @@ public PagedReference newReference(final PagePosition pos,
...
+                            pageId, pagingStore.getAddress(), elapsedMillis);
+            }
+         }
          page.open();
-
+         final long startedReadPage = System.nanoTime();
          List<PagedMessage> pgdMessages = page.read(storageManager);
+         final long elapsedReadPage = System.nanoTime() - startedReadPage;
+         if (elapsedReadPage > PAGE_READ_TIMEOUT_NS) {
+            logger.warnf("Page::read for pageNr=%d on cursor %s tooks %d ms to read %d bytes", pageId,
```
franz1981: @clebertsuconic @jbertram 
In order to not modify `Page::read` and introduce a whole new "failure" state (half pages read with critical IO error raised) 
I've preferred to add some more detailed logs in order to detect weird behaviours, long wait or `page-max-concurrent-io` misconfigurations.
I suppose that adding a proper "slowness disk/JVM suicide" deserve its own PR.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
clebertsuconic: I thought you were going to fix this? Is there any extra commits you did not push?

franz1981: It is already outdated: I've added proper logging around this method and the behaviour is the same as before

franz1981:
I've moved 2 commits re scalability of paging cache from here into https://github.com/apache/activemq-artemis/pull/2667

PR #2646, opened 2019-04-26
ARTEMIS-2317 Reuse file buffer wrapper instances to reduce allocations

franz1981:
While measuring throughput of Page::read introduced by https://github.com/apache/activemq-artemis/pull/2633 I'm not getting the same performance of the original mmap read due to the increased allocation rate: reusing the file buffer wrapper makes the new Page::read on par (but without long TTSP issue) of the original version.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
@@ -22,9 +22,8 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
-import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
-import org.apache.activemq.artemis.api.core.ActiveMQBuffers;
 import org.apache.activemq.artemis.api.core.ICoreMessage;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.buffers.impl.ChannelBufferWrapper;
@@ -119,12 +118,6 @@ public final class Page implements Comparable<Page> {
       return messages;
    }
 
-   private static void decodeInto(ByteBuffer fileBuffer, int encodedSize, PagedMessageImpl msg) {
-      final ActiveMQBuffer wrappedBuffer = ActiveMQBuffers.wrappedBuffer(fileBuffer);
-      wrappedBuffer.writerIndex(encodedSize);
-      msg.decode(wrappedBuffer);
-   }
-
    private ByteBuffer allocateAndReadIntoFileBuffer(ByteBuffer fileBuffer, int requiredBytes) throws Exception {
       final ByteBuffer newFileBuffer = fileFactory.newBuffer(Math.max(requiredBytes, MIN_CHUNK_SIZE));
       newFileBuffer.put(fileBuffer);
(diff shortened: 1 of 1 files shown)


franz1981:
The speed of `Page::read` has been measured by a handrolled test added on `org.apache.activemq.artemis.tests.unit.core.paging.impl.PageTest`:
```java
   @Test
   public void testReadSpeedWithNIO() throws Exception {
      recreateDirectory(getTestDir());
      NIOSequentialFileFactory factory = new NIOSequentialFileFactory(getTestDirfile(), 1);
      SequentialFile file = factory.createSequentialFile("00010.page");

      Page impl = new Page(new SimpleString("something"), new NullStorageManager(), factory, file, 10);

      Assert.assertEquals(10, impl.getPageId());

      impl.open();

      Assert.assertEquals(1, factory.listFiles("page").size());

      SimpleString simpleDestination = new SimpleString("Test");

      final int numberOfElements = 10 * 1024 * 1024 / 100;

      addPageElements(simpleDestination, impl, numberOfElements);

      final int fileSize = impl.getSize();

      System.out.println("fileSize = " + fileSize);

      impl.sync();
      impl.close();

      file = factory.createSequentialFile("00010.page");

      file.open();

      impl = new Page(new SimpleString("something"), new NullStorageManager(), factory, file, 10);

      final int tests = 10;
      final int times = 100;
      NullStorageManager manager = new NullStorageManager();

      for( int t = 0; t<tests;t++) {

         long totalTime = 0;

         for (int m = 0; m<times;m++) {
            final long start = System.nanoTime();
            List<PagedMessage> msgs = impl.read(manager);
            final long elapsed = System.nanoTime() - start;
            totalTime+=elapsed;
            Assert.assertEquals(numberOfElements, msgs.size());

            Assert.assertEquals(numberOfElements, impl.getNumberOfMessages());

            for (int i = 0; i < msgs.size(); i++) {
               Assert.assertEquals(simpleDestination, msgs.get(i).getMessage().getAddressSimpleString());
            }
         }

         final long totalBytes = fileSize * times;
         final long megaPerSec = (totalBytes * 1000L)/ totalTime;
         System.out.println(megaPerSec + " MB/sec");
         System.gc();
      }


      impl.delete(null);

      Assert.assertEquals(0, factory.listFiles(".page").size());
   }
```
Although it is a naive one, this test stress the best case scenario of `Page::read` ie when a file page is already into the OS page cache.

franz1981:
@wy96f Please take a look :+1:

franz1981:
@clebertsuconic Please don't merge it yet: I'm working on a test failure: I will add both test and final solution first

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
```diff
@@ -119,12 +117,6 @@ public void setLiveCache(LivePageCache pageCache) {
       return messages;
    }
 
-   private static void decodeInto(ByteBuffer fileBuffer, int encodedSize, PagedMessageImpl msg) {
-      final ActiveMQBuffer wrappedBuffer = ActiveMQBuffers.wrappedBuffer(fileBuffer);
```
clebertsuconic: This was wrapping each message basically?

I see.. that's a nice improvement

franz1981: Yep! Now I just need to fix something on the wrapper indexes and it should be fine...I didn't expect to allocate so much garbage just for few instances :O

franz1981: Ok, I've just fixed the missing bits (due to how `ByteBuf` work)

franz1981:
I've addressed the test failure, now it should be ok

@wy96f thanks for the review! Let me know if now the performance is much similar to the original version :+1

PR #2647, opened 2019-04-26
ARTEMIS-2323 NettyTransport should also send requests with void promises

franz1981:


tests/artemis-test-support/src/main/java/org/apache/activemq/transport/amqp/client/AmqpConnection.java
```diff
@@ -295,31 +294,6 @@ public void run() {
       return session;
    }
 
-   //----- Access to low level IO for specific test cases -------------------//
-
-   public void sendRawBytes(final byte[] rawData) throws Exception {
```
clebertsuconic: You removed this because it was dead code?

franz1981: Yep!

tests/artemis-test-support/src/main/java/org/apache/activemq/transport/netty/NettyWSTransport.java
```diff
@@ -80,16 +80,13 @@ public NettyWSTransport(NettyTransportListener listener, URI remoteLocation, Net
...
 
    @Override
-   public ChannelFuture send(ByteBuf output) throws IOException {
-      checkConnected();
-      int length = output.readableBytes();
-      if (length == 0) {
-         return null;
-      }
-
-      LOG.trace("Attempted write of: {} bytes", length);
```
michaelandrepearce: Should keep trace logging in new methods for future debugging of issues in live systems

tests/artemis-test-support/src/main/java/org/apache/activemq/transport/netty/NettyTcpTransport.java
```diff
@@ -222,17 +226,34 @@ public ByteBuf allocateSendBuffer(int size) throws IOException {
...
+   }
+
+   @Override
+   public ChannelFuture send(ByteBuf output) throws IOException {
+      return writeAndFlush(output, channel.newPromise(), identity());
+   }
+
+   @Override
+   public void sendVoidPromise(ByteBuf output) throws IOException {
+      writeAndFlush(output, channel.voidPromise(), identity());
```
clebertsuconic: Are you doing this because you saw a high number of propmises being created by the default method?

Looking at send (the one that's using channel.newPromise()) this is only used now in stomp, and even there it should be using sendVoidPromise i think.


I'm not sure I understand what's the point of creating a promise if you're not using it. I thought it was the case.

So, shouldn't all the sends be done without a promise (or with a void promise, which is like a static instance / singleton)?

tests/artemis-test-support/src/main/java/org/apache/activemq/transport/netty/NettyTransport.java
```diff
@@ -38,6 +38,8 @@
 
    ByteBuf allocateSendBuffer(int size) throws IOException;
 
+   void sendVoidPromise(ByteBuf output) throws IOException;
```
clebertsuconic: I'm trying to understand why we would need a send with and without promise.


It seems to be all we need is to change send to be with a void promise.

please advise?

franz1981: Probably MQTT or Stomp (i don't remember) are using the version that need a proper callback while AMQP doesn't need it

tests/artemis-test-support/src/main/java/org/apache/activemq/transport/netty/NettyTcpTransport.java
franz1981: Yes we can always use void primise as long as we don't need a callback: if the callback is used void promise won't work

clebertsuconic: I just looked further.. the only place this is being used is on client APIs on testing.

franz1981: nice, so in this case I could just make 1 call without callbacks and no promises. If, in the future, we will need it again we will add a new method

clebertsuconic: lets just leave it.. otherwise you will have to rewrite those client testers

tests/artemis-test-support/src/main/java/org/apache/activemq/transport/netty/NettyWSTransport.java
clebertsuconic: I don't think this one is very useful...

I will merge it.

PR #2649, opened 2019-04-26
NO-JIRA Fix test

franz1981:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/paging/GlobalPagingTest.java
```diff
@@ -334,7 +334,7 @@ public void testManagementMessageRequestCannotFailAfterFailedDirectDeliver() thr
                   final MessageReference reference = MessageReference.Factory.createReference(session.createMessage(false), queue);
                   //it will cause QueueImpl::directDeliver -> false
                   queue.addHead(reference, false);
-                  Assert.assertSame(reference, queue.removeReferenceWithID(reference.getMessageID()));
+                  queue.removeReferenceWithID(reference.getMessageID());
```
clebertsuconic: Can you assertNotNull instead instead of just ignore the result?

franz1981: But it can be null, is just mater of timing: the remove itself is not necessary to the test TBH (I've written that test)

clebertsuconic: @franz1981 make it not racy then:

```java
Wait.assertNotNull(() -> queue.removReferenceWithID(reference.getMessageID())
```


It seems to me you could even use Wait.assertSame

clebertsuconic:
@franz1981 I changed the test, and pushed using this PR as a reference

PR #2650, opened 2019-04-26
 ARTEMIS-2320 Multiple math-related fixes

jdanekrh:


artemis-commons/src/test/java/org/apache/activemq/artemis/utils/collections/ConcurrentLongHashSetTest.java
```diff
@@ -25,10 +25,10 @@
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
-import java.util.Random;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
+import java.util.concurrent.ThreadLocalRandom;
```
michaelandrepearce: No please dont

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/mapped/MappedSequentialFile.java
```diff
@@ -110,7 +110,7 @@ public void open(int maxIO, boolean useExecutor) throws IOException {
    @Override
    public boolean fits(int size) {
       checkIsOpen();
-      final long newPosition = this.mappedFile.position() + size;
+      final long newPosition = (long) this.mappedFile.position() + size;
       final boolean hasRemaining = newPosition <= this.mappedFile.length();
```
michaelandrepearce: No need for casting. Is just code clutter.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1398,9 +1398,9 @@ public long getMessageCount() {
       if (pageSubscription != null) {
          // messageReferences will have depaged messages which we need to discount from the counter as they are
          // counted on the pageSubscription as well
-         return pendingMetrics.getMessageCount() + getScheduledCount() + getDeliveringCount() + pageSubscription.getMessageCount();
+         return (long) pendingMetrics.getMessageCount() + getScheduledCount() + getDeliveringCount() + pageSubscription.getMessageCount();
```
michaelandrepearce: No need for casting. Just causing code clutter

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1419,9 +1419,9 @@ public long getPersistentSize() {
    public long getDurableMessageCount() {
       if (isDurable()) {
          if (pageSubscription != null) {
-            return pendingMetrics.getDurableMessageCount() + getDurableScheduledCount() + getDurableDeliveringCount() + pageSubscription.getMessageCount();
+            return (long) pendingMetrics.getDurableMessageCount() + getDurableScheduledCount() + getDurableDeliveringCount() + pageSubscription.getMessageCount();
```
michaelandrepearce: Ditto

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -1783,7 +1783,7 @@ protected final void receiveMessages(ClientConsumer consumer,
       journal.load(committedRecords, preparedTransactions, null, false);
 
       for (RecordInfo info : committedRecords) {
-         Integer ikey = new Integer(info.getUserRecordType());
+         Integer ikey = (int) info.getUserRecordType();
```
michaelandrepearce: Correct way would actually be to call intValue

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/mapped/MappedSequentialFile.java
jdanekrh: It may change semantics, depending on how large the variables are. You are adding together two ints, then assigning to long. Without the cast, you can get overflow on the int range, before you assign to that long. (Which is what the IntLongMath warning is about).

jdanekrh:
```
jshell> Integer.MAX_VALUE + Integer.MAX_VALUE
$2 ==> -2

jshell> (long) Integer.MAX_VALUE + Integer.MAX_VALUE
$3 ==> 4294967294
```

artemis-commons/src/test/java/org/apache/activemq/artemis/utils/collections/ConcurrentLongHashSetTest.java
jdanekrh: Thanks, I'll report it there https://github.com/apache/bookkeeper/issues/2077

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/mapped/MappedSequentialFile.java
michaelandrepearce: Ah yes i forgot about that. Good stuff

PR #2651, opened 2019-04-26
ARTEMIS-2320 Fix ClassCanBeStatic errorprone warning

jdanekrh:
Inner class is non-static but does not reference enclosing class
    (see https://errorprone.info/bugpattern/ClassCanBeStatic)

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/Consumer.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/Consumer.java
@@ -125,7 +125,7 @@ public class Consumer extends DestAbstract {
       }
    }
 
-   private class SerialiserMessageListener implements MessageListener {
+   private static class SerialiserMessageListener implements MessageListener {
 
       private MessageSerializer messageSerializer;
 
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/xml/XMLMessageImporter.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/xml/XMLMessageImporter.java
@@ -315,7 +315,7 @@ public class XMLMessageImporter {
       void processBodyBytes(byte[] bytes) throws IOException;
    }
 
-   public class MessageInfo {
+   public static class MessageInfo {
       public long id;
       public List<String> queues;
       public Message message;
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodec.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodec.java
@@ -108,7 +108,7 @@ public class DefaultSensitiveStringCodec implements SensitiveDataCodec<String> {
       return algorithm.verify(inputValue, storedValue);
    }
 
-   private abstract class CodecAlgorithm {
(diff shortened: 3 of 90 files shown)


michaelandrepearce:
If classes can be made static then actually they should be in their own class file.

clebertsuconic:
@michaelandrepearce I don't think that's a good idea. static classes can be useful to avoid clutter, when the logic only makes sense in the context of that class.

static classes can be in their own class file.. should is a strong word on this IMHO.


Like the other day I moved an independent class as a static class file, as the class only made sense in the context of the test.


Having said that though, I don'w think we should abuse static classes... I'm just saying that in some cases the class doesn't make sense as an independent class.

michaelandrepearce:
I agree about tone. And thats why i did not say MUST , and instead used SHOULD which means ideally but if not sensible or possible then it doesnt have to. 

What seems to be the case here is a blanket of changing of classes but without looking at what can be split out properly, only keeping those as nested classes where sensible. As in not a single class in this PR has become a class in its own right. I find that hard to believe

michaelandrepearce:
Just one example of classes that should be their own class is the algo implementations.

jdanekrh:
I'm not going to pull out any static nested classes into their own files. If this should get done, somebody more familiar with the codebase needs to do it. Possibly in a subsequent PR.

I was thinking, we could configure the ClassCanBeStatic warning to be an error. Then all subsequently created nested classes that can be static would _have_ to be static, or not be nested.

michaelandrepearce:
So then i suggest leave the warning, warnings are there to suggest reviewing and refactoring code, its not just about finding the sticky plaster to make the warning go away. 

Its like code coverage, if you have a class that has low coverage, the solution isn't just to add tests quickly for all your getters and setters, which would get coverage up, but really you should add better coverage to real code. 

The point being, don't hide an issue with a sticky plaster, go fix it properly, or leave the warning there, so it keeps getting noticed....

jdanekrh:
The purpose of the warning isn't really to point out style issues about class hierarchy, but a potential (tiny) performance issue, which is fixed by just adding `static` everywhere.

> A non-static inner class has a reference to its outer class, and access to the outer class' fields and methods. That class reference makes the inner class larger and could cause the outer class instance to live in memory longer than necessary. https://rules.sonarsource.com/java/tag/performance/RSPEC-2694

Another thing, does anybody ever read those warnings? They don't show in IntelliJ (they don't show for me) and they get lost in mvn output...

I don't consider this PR very important, and it was easy to produce (IntelliJ autofix), so I don't mind closing it and wait for the eventual comprehensive refactoring.

michaelandrepearce:
As it stands, this pr which is a refactor needs a few iterations imo. Not everything should be inner classes. Thats not to say all has to be top level either.

clebertsuconic:
Another reason to not merge this... is... automatic changes can introduce failures.

You're not testing line by line on these changes.. it's a recipe for a failure.

jdanekrh:
I have to concede the point about

> Another reason to not merge this... is... automatic changes can introduce failures.

In this case, it can be argued the change is safe: automatic changes are less likely to introduce failures than manual changes of this scope; class static is checked at compile time. If it compiles, then it is ok (although if asked whether I'd be willing to bet someones life on it, I'd ponder about reflection, get nowhere, and quickly back off from that claim).

It is hard to tell how much safety net do I have in the form of unit tests already written.

These "fixes" just do not seem to be worth the potential trouble it can cause down the line.

PR #2652, opened 2019-04-27
 ARTEMIS-2320 Assorted small code fixes, mostly from IntelliJ autofixes

jdanekrh:


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -592,7 +592,6 @@ public Object run(ActionContext context) throws Exception {
          filters.put("${extra.web.attributes}", "");
       }
       filters.put("${fsync}", String.valueOf(!noJournalSync));
-      filters.put("${user}", System.getProperty("user.name", ""));
```
jdanekrh: was rewritten on line 607

tests/jms-tests/src/test/java/org/apache/activemq/artemis/jms/tests/tools/container/InVMInitialContextFactory.java
```diff
@@ -104,10 +100,6 @@ public Context getInitialContext(final Hashtable<?, ?> environment) throws Namin
       }
    }
 
-   public static void reset() {
-      InVMInitialContextFactory.initialContexts = new HashMap<>();
```
michaelandrepearce: This is a public method we shouldnt be removing.

tests/jms-tests/src/test/java/org/apache/activemq/artemis/jms/tests/tools/container/InVMInitialContextFactory.java
```diff
@@ -31,11 +31,7 @@
 
    // Static --------------------------------------------------------
 
-   private static Map<Integer, Context> initialContexts;
-
-   static {
-      InVMInitialContextFactory.reset();
-   }
+   private static final Map<Integer, Context> initialContexts = new HashMap<>();
```
michaelandrepearce: This was not final because it was resettable.

tests/jms-tests/src/test/java/org/apache/activemq/artemis/jms/tests/tools/container/InVMInitialContextFactory.java
jdanekrh: Nobody seems to be calling that, and it is in the tests part of the codebase. If we want to keep it, then maybe calling `initialContexts.clear()` (under whatever synchronization that is possibly required) would be sufficient.

If it needs further thought, I can drop it from the PR.

tests/jms-tests/src/test/java/org/apache/activemq/artemis/jms/tests/tools/container/InVMInitialContextFactory.java
jdanekrh: Yes, but nobody was actually resetting it, in the codebase.

michaelandrepearce: Fair enough.

tests/jms-tests/src/test/java/org/apache/activemq/artemis/jms/tests/tools/container/InVMInitialContextFactory.java
michaelandrepearce: As long as no ones using this in extended tests then fair enough. Concern is just where people may use and extend it in their own projects. If maybe cross check with wildfly project (known big user) to make sure. Then im happy

PR #2655, opened 2019-04-30
ARTEMIS-2327 ExceptionListener invoked when connection level exception

michaelandrepearce:
Add tests
Add fix - if timeout occurs on sending packet, calls same code that is invoked if timeout occurs on during ping aligning logic, and ensuring JMS connection exception listener gets invoked to inform the client logic to react. Such libs like Spring JMS reply on exception listeners to react to connection level errors.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
@@ -451,7 +451,9 @@ public final class ChannelImpl implements Channel {
             }
 
             if (response == null) {
-               throw ActiveMQClientMessageBundle.BUNDLE.timedOutSendingPacket(connection.getBlockingCallTimeout(), packet.getType());
+               ActiveMQException e = ActiveMQClientMessageBundle.BUNDLE.timedOutSendingPacket(connection.getBlockingCallTimeout(), packet.getType());
+               connection.fail(e);
+               throw e;
             }
 
             if (response.getType() == PacketImpl.EXCEPTION) {
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/remoting/ReconnectTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/remoting/ReconnectTest.java
@@ -330,9 +330,13 @@ public class ReconnectTest extends ActiveMQTestBase {
       server.start();
       // imitate session reattach timeout
       Interceptor reattachInterceptor = new Interceptor() {
+
+         boolean reattached;
+
          @Override
          public boolean intercept(Packet packet, RemotingConnection connection) throws ActiveMQException {
-            if (packet.getType() == PacketImpl.REATTACH_SESSION) {
+            if (!reattached && packet.getType() == PacketImpl.REATTACH_SESSION) {
+               reattached = true;
                return false;
             } else {
                return true;
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/connection/ExceptionListenerForConnectionTimedOutExceptionTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/connection/ExceptionListenerForConnectionTimedOutExceptionTest.java
@@ -1,0 +1,173 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.tests.integration.jms.connection;
+
+import java.util.concurrent.atomic.AtomicReference;
+import javax.jms.Connection;
+import javax.jms.JMSException;
+import javax.jms.MessageConsumer;
+import javax.jms.MessageProducer;
+import javax.jms.Queue;
+import javax.jms.Session;
+import javax.jms.TextMessage;
+import org.apache.activemq.artemis.api.core.ActiveMQConnectionTimedOutException;
+import org.apache.activemq.artemis.api.core.ActiveMQException;
+import org.apache.activemq.artemis.api.core.Interceptor;
+import org.apache.activemq.artemis.core.protocol.core.Packet;
+import org.apache.activemq.artemis.core.protocol.core.impl.PacketImpl;
+import org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory;
+import org.apache.activemq.artemis.spi.core.protocol.RemotingConnection;
+import org.apache.activemq.artemis.tests.util.JMSTestBase;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ExceptionListenerForConnectionTimedOutExceptionTest extends JMSTestBase {
+
+   private Queue queue;
+
+   @Before
+   @Override
+   public void setUp() throws Exception {
+      super.setUp();
+
+      queue = createQueue("TestQueue");
+   }
+
+   @Test(timeout = 60000)
+   public void testOnAcknowledge() throws Exception {
+      Connection sendConnection = null;
+      Connection connection = null;
+      AtomicReference<JMSException> exceptionOnConnection = new AtomicReference<>();
+
+      try {
+         ((ActiveMQConnectionFactory) cf).setOutgoingInterceptorList(OutBoundPacketCapture.class.getName());
+         ((ActiveMQConnectionFactory) cf).setIncomingInterceptorList(SessAcknowledgeCauseResponseTimeout.class.getName());
+         ((ActiveMQConnectionFactory) cf).setBlockOnAcknowledge(true);
+
+         sendConnection = cf.createConnection();
+
+         final Session sendSession = sendConnection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+         final MessageProducer producer = sendSession.createProducer(queue);
+
+         TextMessage message = sendSession.createTextMessage();
+
+         message.setText("Message");
+
+         producer.send(message);
+
+         connection = cf.createConnection();
+         connection.start();
+         connection.setExceptionListener(exceptionOnConnection::set);
+         final Session consumerSession = connection.createSession(false, Session.CLIENT_ACKNOWLEDGE);
+         final MessageConsumer messageConsumer = consumerSession.createConsumer(queue);
+
+         TextMessage message1 = (TextMessage) messageConsumer.receive(1000);
+
+         assertEquals("Message", message1.getText());
+
+         message1.acknowledge();
+
+         fail("JMSException expected");
+
+      } catch (JMSException e) {
+         assertTrue(e.getCause() instanceof ActiveMQConnectionTimedOutException);
+         //Ensure JMS Connection ExceptionListener was also invoked
+         assertNotNull(exceptionOnConnection.get());
+         assertTrue(exceptionOnConnection.get().getCause() instanceof ActiveMQConnectionTimedOutException);
+      } finally {
+         if (connection != null) {
+            connection.close();
+         }
+         if (sendConnection != null) {
+            sendConnection.close();
+         }
+      }
+   }
+
+   @Test(timeout = 60000)
+   public void testOnSend() throws Exception {
+      Connection sendConnection = null;
+      Connection connection = null;
+      AtomicReference<JMSException> exceptionOnConnection = new AtomicReference<>();
+
+      try {
+         ((ActiveMQConnectionFactory) cf).setOutgoingInterceptorList(OutBoundPacketCapture.class.getName());
+         ((ActiveMQConnectionFactory) cf).setIncomingInterceptorList(SessSendCauseResponseTimeout.class.getName());
+
+         sendConnection = cf.createConnection();
+         sendConnection.setExceptionListener(exceptionOnConnection::set);
+         final Session sendSession = sendConnection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+         final MessageProducer producer = sendSession.createProducer(queue);
+
+         TextMessage message = sendSession.createTextMessage();
+
+         message.setText("Message");
+
+         producer.send(message);
+
+         fail("JMSException expected");
+
+      } catch (JMSException e) {
+         assertTrue(e.getCause() instanceof ActiveMQConnectionTimedOutException);
+         //Ensure JMS Connection ExceptionListener was also invoked
+         assertNotNull(exceptionOnConnection.get());
+         assertTrue(exceptionOnConnection.get().getCause() instanceof ActiveMQConnectionTimedOutException);
+
+      } finally {
+         if (connection != null) {
+            connection.close();
+         }
+         if (sendConnection != null) {
+            sendConnection.close();
+         }
+      }
+   }
+
+
+   static Packet lastPacketSent;
+
+   public static class OutBoundPacketCapture implements Interceptor {
+
+      @Override
+      public boolean intercept(Packet packet, RemotingConnection connection) throws ActiveMQException {
+         lastPacketSent = packet;
+         return true;
+      }
+   }
+
+   public static class SessAcknowledgeCauseResponseTimeout implements Interceptor {
+
+      @Override
+      public boolean intercept(Packet packet, RemotingConnection connection) throws ActiveMQException {
+         if (lastPacketSent.getType() == PacketImpl.SESS_ACKNOWLEDGE && packet.getType() == PacketImpl.NULL_RESPONSE) {
+            return false;
+         }
+         return true;
+      }
+   }
+
+   public static class SessSendCauseResponseTimeout implements Interceptor {
+
+      @Override
+      public boolean intercept(Packet packet, RemotingConnection connection) throws ActiveMQException {
+         if (lastPacketSent.getType() == PacketImpl.SESS_SEND && packet.getType() == PacketImpl.NULL_RESPONSE) {
+            return false;
+         }
+         return true;
+      }
+   }
+}


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
```diff
@@ -451,7 +451,9 @@ public Packet sendBlocking(final Packet packet,
             }
 
             if (response == null) {
-               throw ActiveMQClientMessageBundle.BUNDLE.timedOutSendingPacket(connection.getBlockingCallTimeout(), packet.getType());
+               ActiveMQException e = ActiveMQClientMessageBundle.BUNDLE.timedOutSendingPacket(connection.getBlockingCallTimeout(), packet.getType());
+               connection.fail(e);
```
clebertsuconic: LGTM: I would just run the whole testsuite to make sure. I can run it here and merge it ok.

ok with you?

clebertsuconic: I'm more concerned on FailoverTests... if the connection.failure here would cause any further issues.

michaelandrepearce: So i ran a few of them in the normal int tests (didnt run any extra tests tho). Happy for you to run more if you like

gemmellr:
The JIRA notes that "the exception is returned to calling code". I'm not clear if it means through to the application code or not. I bring it up as if the exception is being explicitly reported elsewhere, e.g a send method throwing, its generally not meant to be reported to the ExceptionListener according to the JMS spec.

michaelandrepearce:
Yes it does. But if its a connection exception it should also make it to the connection exception listener. Typically frameworks such as spring rely on that.

gemmellr:
It perhaps depends what the exception is, and what reaction is being taken to it (e.g killing the connection, or considering it already dead), but the specs fairly clear about what it wants overall:

> "If an exception is thrown on a JMS call it, by definition, must not be delivered to an ExceptionListener (in other words, ExceptionListener is not for the purpose of monitoring all exceptions thrown by a connection)"

michaelandrepearce:
So as i said, the code is doing this for others, and in this case because spring wasnt notified on the exception listener it doesnt let go of what became a stale connection, when this specific issue occurred. 

As per here, it is meant to be notified of a critical connection exception async, and in this side case it wasn't
https://docs.oracle.com/javaee/6/api/javax/jms/ExceptionListener.html

gemmellr:
Something else doing something isnt necessarily a reason to copy it. The async is in reference to the application not having already been told synchronously via a method call, which as above the spec doc is clear on. 

As I said, it perhaps depends what the exception is, and what reaction is being taken to it. If this exception always represents the connections ultimate failure and isnt in response to a particular method call then telling the exception listener does seem reasonable. If it doesnt, e.g its failing over, or its only a single methods failure and everythign else is fine, then it often might not be appropriate.

In particular the async ExceptionListener is good for things only using MessageListeners, which will never find out otherwise since they dont call methods [if theres no delivery happening]. Spring uses synchronous receive calls (by default at least) so it should already discover failures that result in the connection being considered broken, without the ExceptionListener. If it doesnt that might point to another issue.

michaelandrepearce:
re:

If this exception always represents the connections ultimate failure and isnt in response to a particular method call then telling the exception listener does seem reasonable

yes, if failover was occuring that would have been caught further up in the checks (notice method about 10ish lines up, waitOnFailover), and from experience of the issue, it seems to be that way once it hits here the connection is no longer recoverable..

gemmellr:
Didnt see most of your edits when replying earlier, but my reply essentially fits anyway. I havent looked at and am not familiar with the client code outside the context of this diff. I'm just saying 'always call the exception listener' for connection exceptions seems out of line with the specs general desire, except perhaps if those exceptions are always classed as connection-ending failures, or if the surrounding code such as the failover stuff is intercepting and handling that, in which case ok.

michaelandrepearce:
As i said, failover is handled before it gets to this, so we know best efforts to failover etc have taken place, so if it gets here, its terminal, you need to inform listeners so they can react.

gemmellr:
Missed another edit:
"ExceptionListeners is used by caching and pooling cf wrappers (e.g. springs CachingConnectionFactory), as such they're only access to if issue with exception is by exception listener, so its not just MessageListeners as a use case here."

As wrappers they can have access to all exceptions thrown anywhere if they want it. They might choose not to bother of course.

michaelandrepearce:
Thats why exception listener is there, its been like that for as many years as i can remember, all frameworks that pool or cache connections im aware of not just spring, rely heavily on the exception listener and being notified by the client.

gemmellr:
> As i said, failover is handled before it gets to this, so we know best efforts to failover etc have taken place, so if it gets here, its terminal, you need to inform listeners so they can react.

I was talking about any failover starting to happen in response to this 'connection ending exception', not any checks happening before it.

gemmellr:
If there is no such response, and this is the end, then sure notify the exception listener.

michaelandrepearce:
And thats what we're doing. Calling the same connection.fail(e) method that other blocks and code paths do.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
clebertsuconic: @michaelandrepearce I ran the 2 hours:40 min test with on a jenkins server I have:
```
mvn test -Ptests
```

apparently there are tests hunging because of this.


I couldn't figure out what test is it.. as the build generate zillions of:

```
med out after waiting 2,000 ms for response when sending packet 32 [code=CONNECTION_TIMEDOUT]
[main] 09:19:18,204 WARN  [org.apache.activemq.artemis.core.client] AMQ212037: Connection failure has been detected: AMQ219014: Timed out after waiting 2,000 ms for response when sending packet 32 [code=CONNECTION_TIMEDOUT]
2019-05-01 09:19:18,204 WARN  [org.apache.activemq.artemis.core.client] AMQ212037: Connection failure has been detected: AMQ219014: Timed out after waiting 2,000 ms for response when sending packet 32 [code=CONNECTION_TIMEDOUT]
2019-05-01 09:19:18,204 WARN  [org.apache.activemq.artemis.core.client] AMQ212037: Connection failure has been detected: AMQ219014: Timed out after waiting 2,000 ms for response when sending packet 32 [code=CONNECTION_TIMEDOUT]
[main] 09:19:20,210 WARN  [org.apache.activemq.artemis.core.client] AMQ212037: Connection failure has been detected: AMQ219014: Timed out after waiting 2,000 ms for response when sending packet 32 [code=CONNECTION_TIMEDOUT]
2019-05-01 09:19:20,210 WARN  [org.apache.activemq.artemis.core.client] AMQ212037: Connection failure has been detected: AMQ219014: Timed out after waiting 2,000 ms for response when sending packet 32 [code=CONNECTION_TIMEDOUT]
2019-05-01 09:19:20,210 WARN  [org.apache.activemq.artemis.core.client] AMQ212037: Connection failure has been detected: AMQ219014: Timed out after waiting 2,000 ms for response when sending packet 32 [code=CONNECTION_TIMEDOUT]
[main] 09:19:22,216 WARN  [org.apache.activemq.artemis.core.client] AMQ212037: Connection failure has been detected: AMQ219014: Timed out after waiting 2,000 ms for response when sending packet 32 [code=CONNECTION_TIMEDOUT]
2019-05-01 09:19:22,216 WARN  [org.apache.activemq.artemis.core.client] AMQ212037: Connection failure has been detected: AMQ219014: 
```


I am taking a break today.. I can take a look on more details later today / tomorrow morning.

clebertsuconic: I bet it will be nettyFailoverTest. I have a Deja Voux on this. I believe I had tried once to add such code and the reconnection loop failed.

clebertsuconic: It is not NettyFailoverTest. ran it manually.

I will have to read the logs later today / tomorrow morning.

Keep this PR open please until we figure out?

clebertsuconic: @michaelandrepearce org.apache.activemq.artemis.tests.integration.remoting.ReconnectTestis::testReattach failing.

It's probably a test issue, since the test was meant to disconnect the inner socket and play with a reattachment.

michaelandrepearce: @clebertsuconic ill keep pr open, if we can't resolve obviously we cannot merge, but then possibly will need to find another way to fix this.

michaelandrepearce: @clebertsuconic have just pushed a small change to that test, originally it would just repeat the same over and over, where as test is just expecting it ever to be called once, so made the interceptor in the test to adjust the behaviour just the once using a bool

michaelandrepearce: @clebertsuconic if you look at and let me know what you think.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/remoting/ReconnectTest.java
```diff
@@ -330,9 +330,13 @@ public void testReattachTimeout() throws Exception {
       server.start();
       // imitate session reattach timeout
       Interceptor reattachInterceptor = new Interceptor() {
+
+         boolean reattached;
+
          @Override
          public boolean intercept(Packet packet, RemotingConnection connection) throws ActiveMQException {
-            if (packet.getType() == PacketImpl.REATTACH_SESSION) {
+            if (!reattached && packet.getType() == PacketImpl.REATTACH_SESSION) {
```
michaelandrepearce: @clebertsuconic re-attach is meant to occur just the once, so making the imitation done via interceptor do that.

jbertram:
I'm not sure where I was when this PR was originally sent, but I came across these changes while investigating another issue. I'm considering reverting these changes for a handful reasons:

1. As Robbie pointed out originally, this doesn't fit with the JMS specification which clearly indicates that an `ExceptionListener` should *only* be invoked when the exception has nowhere else to go. The main use-case here is when a client implements `MessageListener` (which is invoked asynchronously by the underlying implementation) but still wants to know when its connection dies. Given that, in this situation, the client itself isn't invoking e.g. `javax.jms.Consumer#receive()` it has no way to receive the exception other than through the `ExceptionListener`. I'll quote (again) the salient part section 6.1.7 from the JMS 2 specification:

    > The exceptions delivered to `ExceptionListener` are those that have no other place to be reported. If an exception is thrown on a JMS call it, by definition, must not be delivered to an `ExceptionListener` (in other words, **`ExceptionListener` is not for the purpose of monitoring all exceptions thrown by a connection**). [emphasis mine]

   As far as I can tell by looking through the code-base `sendBlocking` is only ever called from a JMS client application in response to a specific JMS API call, and it is from these calls that the exception must be thrown to the caller and *not* to the `ExceptionListener`.

2. I'm not convinced by the counter-argument that Spring et al. rely on this. The specification clearly says that the `ExceptionListener` is not a way to monitor all connection exceptions. Therefore, well-written frameworks/wrappers should have other ways to detect or be notified of connection problems. If the wrappers don't implement such functionality I see that as their problem, not ours. We shouldn't violate the JMS specification to suit their deficiencies.
3. The "ping" functionality was mentioned as the example to follow here. However, I don't think that comparison is apt in this case. The pinging functionality is an implementation specific behavior to monitor the validity of the connection. If a ping fails then by definition the connection has failed. This functionality executes without any action from the client application. Therefore, a ping failure has *nowhere else* to be reported but in an `ExceptionListener`. This is categorically different from a call to `sendBlocking()` which is the result of an explicit JMS API call from the client application.
4. The underlying issue here is that the client did not receive a response from the broker for a single "packet" within the specified timeout. That's all. There are a number of reasons this situation might arise. Perhaps the broker is under heavy load and simply didn't have time to respond. Perhaps the network is slow for some reason. Perhaps the timeout is simply too low for the use-case. The connection is *not* necessarily dead. It's plausible that the client might, in fact, catch the `JMSException` and retry the failed operation or eventually make another call. The comment on [`o.a.a.a.c.p.c.i.ChannelImpl#sendBlocking(Packet, int, byte)`](https://github.com/apache/activemq-artemis/blob/master/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java#L365) even notes this. If the connection actually is dead then the underlying pinger/monitor should detect that.
5. This change gives the impression to users that exceptions like this which get thrown to a caller should also invoke the `ExceptionListener` which is incorrect per the specification.
6. I looked through the client code-base for uses of `o.a.a.a.s.c.p.RemotingConnection#fail(o.a.a.a.a.c.ActiveMQException)` aside from the one added via this PR, and I only found 3 others which were all truly asynchronous use-cases where the exception had nowhere else to go.

michaelandrepearce:
So if exception occurs on blocking send a connection exception is meant to be raised if it is terminal. I dont see in jms spec it stating otherwise

michaelandrepearce:
Otherwise without that frameworks like spring etc would be unable to react to this. I would ask this is discussed and consult in the spring group the framework is too pervasivrly used in java ecosystem. Jms spec here is sometimes not specific enough and here its unfortunately open a little to interpretation two ways, one which spring has taken is that it has coded and relied of for years...

michaelandrepearce:
If a JMS provider detects a serious problem with a Connection object, it informs the Connection object's ExceptionListener, if one has been registered. It does this by calling the listener's onException method, passing it a JMSException argument describing the problem.

An exception listener allows a client to be notified of a problem asynchronously. Some connections only consume messages, so they would have no other way to learn that their connection has failed.

A JMS provider should attempt to resolve connection problems itself before it notifies the client of them.

michaelandrepearce:
https://docs.oracle.com/javaee/6/api/javax/jms/ExceptionListener.html

michaelandrepearce:
There is no mention what so ever that it should not be called if a connection level exception is raised else where.

michaelandrepearce:
I  wont argue on this as i dont work for the company where this caused a significant prod outage due to lack of it, and other brokers do send the exception and no issues with spring. Im sure that company may have issue if that problem regresses, i believe the org you work for supports them still so that will be something you should be aware of.

clebertsuconic:
When I last looked this was useful for the case when a connection issue caused an exception. 

Back then there was a case where a disconnect was silently failing. 


But we should only call the exception listener on that case. 

What is the case where this is failing ?

jbertram:
> So if exception occurs on blocking send a connection exception is meant to be raised if it is terminal. I dont see in jms spec it stating otherwise

As noted previously, a failure to get a response from a blocking call is not necessarily a fatal or terminal connection error. If, for example, a JMS client is attempting to send a persistent message and the blocking call fails to receive a response then the caller will receive a `javax.jms.JMSException` as described in the [API documentation](https://docs.oracle.com/javaee/7/api/javax/jms/MessageProducer.html#send-javax.jms.Destination-javax.jms.Message-). Since the `javax.jms.JMSException` can be reported to the caller then it should not, according to the specification, be sent to any `ExceptionListener`.

> Otherwise without that frameworks like spring etc would be unable to react to this.

As noted in #2 of my previous comment, the specification clearly states that the "`ExceptionListener` is not for the purpose of monitoring all exceptions thrown by a connection." Therefore, if frameworks like Spring are relying on the `ExceptionListener` to monitor the connection for all exceptions then it would seem they are improperly implemented. I'm not familiar enough with Spring to say whether or not that's true about it.

> Jms spec here is sometimes not specific enough and here its unfortunately open a little to interpretation two ways, one which spring has taken is that it has coded and relied of for years...

I'm not familiar enough with Spring to say authoritatively how it is or isn't implemented so I'll leave that bit alone. 

However, the JMS specification seems crystal clear on these points:

1. Exceptions that can be reported elsewhere *must not* go to the `ExceptionListener`.
2. Therefore, the `ExceptionListener` should not be relied upon to monitor for all connection exceptions.

For clarity's sake I'll quote a larger chunk of section 6.1.7 of the JMS 2 specification which essentially includes the portions that both you and I have quoted together:

> If a JMS provider detects a problem with a connection, it will inform the connection’s `ExceptionListener`, if one has been registered. To retrieve an `ExceptionListener`, the JMS provider calls the connection’s `getExceptionListener()` method. This method returns the `ExceptionListener` for the connection. If no `ExceptionListener` is registered, the value `null` is returned. The connection can then use the listener by calling the listener’s `onException()` method, passing it a `JMSException` describing the problem.
>
>This allows a client to be asynchronously notified of a problem. Some connections only consume messages, so they would have no other way to learn their connection has failed.
>
>...
>
>A JMS provider should attempt to resolve connection problems itself prior to notifying the client of them.
>
>The exceptions delivered to `ExceptionListener` are those that have no other place to be reported. If an exception is thrown on a JMS call it, by definition, must not be delivered to an `ExceptionListener` (in other words, `ExceptionListener` is not for the purpose of monitoring all exceptions thrown by a connection).

In summary, if a JMS provider detects a problem with a connection it will inform the connection's `ExceptionListener` of the exception if that exception has no other place to be reported. This allows a client to be asynchronously notified of a problem when no exception can be synchronously delivered to the client from a JMS API invocation. The `ExceptionListener` is not for the purpose of monitoring all exceptions thrown by a connection.

> There is no mention what so ever that it should not be called if a connection level exception is raised else where.

In my estimation, that is exactly what the JMS spec means when it says, "The exceptions delivered to `ExceptionListener` are those that have no other place to be reported. If an exception is thrown on a JMS call it, by definition, must not be delivered to an `ExceptionListener`"

For what it's worth, I didn't mean to imply that there wasn't a legitimate problem at the root of this issue. My claim is simply that this is the wrong way to fix it.

michaelandrepearce:
So all i would say to revert this an alternative fix should be supplied to fix that original isshe. Not just a plain revert else you will cause an immediate regression. One that caused a prod outage.

michaelandrepearce:
I do not see in the detail any mention of what you summarise as

> if that exception has no other place to be reported.

I assume also other brokers and spring also didnt read it the way you do. As noted i suggest a discussion with spring community should be had to agree a interpretation and then agree a way forward. Spring as i mentioned is too pervasive used just to go break it.

michaelandrepearce:
For what its worth, i dont have access to original case tickets (as i moved on) but i believe the route issue was a producer never recovered and it was later found it was because the connection had failed to reconnect itself but spring didnt invalidate the connection, im sure you probably have access to the original case/issue details and logs though.

jbertram:
> I do not see in the detail any mention of what you summarise as, "if that exception has no other place to be reported."

I'm not sure why this information is absent from [the JavaDoc for `ExceptionListener`](https://docs.oracle.com/javaee/7/api/javax/jms/ExceptionListener.html). This seems like an oversight since the expected semantics are clearly outlined in the specification. To quote once more:

> The exceptions delivered to `ExceptionListener` are those that have no other place to be reported. If an exception is thrown on a JMS call it, by definition, must not be delivered to an `ExceptionListener` (in other words, `ExceptionListener` is not for the purpose of monitoring all exceptions thrown by a connection).

To be clear, this was not added in JMS 2. The same language exists in section 4.3.8 of the JMS 1.1 *and* 1.0.2 specifications.

Take a look at [the assertions for the JMS 2 TCK](https://htmlpreview.github.io/?https://github.com/eclipse-ee4j/jakartaee-tck/blob/master/internal/docs/jms/JMSSpecAssertions.html). Search for "JMS:SPEC:109". This is the specific TCK assertion dealing with this particular semantic, and it quotes the corresponding bit of the JMS specification (which both Robbie and I also quoted - which is absent from the `ExceptionListener` JavaDoc). It's clearly called out here, but also note that the "testable" column is marked as `false`. My guess is that it's just not possible to reliably inject failures into any arbitrary JMS implementation in order to test this. However, it's worth noting that several other very important bits of the spec also have "testable" set to `false` (e.g. JMS:SPEC:116, JMS:SPEC:133, JMS:SPEC:135, etc.) - ones that you wouldn't expect implementers to just ignore. For what it's worth the same JMS:SPEC:109 assertion is in the JMS 1.1 TCK, but I couldn't find a public copy of that document to share.

> I assume also other brokers and spring also didnt read it the way you do.

Other brokers may implement things differently. Spring may rely on the `ExceptionListener` for all connection exceptions. If that's the case then I think they're running afoul of the specification, and that's bad for compatibility among other things.

I want to be pragmatic about this so my plan is to make the particular semantic you introduced configurable via a URL parameter, but the behavior will be disabled by default.

michaelandrepearce:
> I want to be pragmatic about this so my plan is to make the particular semantic you introduced configurable via a URL parameter, but the behavior will be disabled by default.

This sounds a good compromise. I assume the param would also be settable on the cf programatically as well like other url pram settings.

PR #2656, opened 2019-05-01
ARTEMIS-2328 Routing after empty addresses could lead to invalid mess…

clebertsuconic:
…ages

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -934,7 +934,7 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
       }
 
       if (logger.isTraceEnabled()) {
-         logger.trace("Message after routed=" + message);
+         logger.trace("Message after routed=" + message + "\n" + context.toString());
       }
 
       try {
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/RoutingContext.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/RoutingContext.java
@@ -83,5 +83,4 @@ public interface RoutingContext {
    boolean isReusable(Message message, int version);
 
 
-
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RoutingContextImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RoutingContextImpl.java
@@ -16,6 +16,8 @@
  */
 package org.apache.activemq.artemis.core.server.impl;
 
+import java.io.PrintWriter;
+import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -30,8 +32,11 @@ import org.apache.activemq.artemis.core.server.RouteContextList;
 import org.apache.activemq.artemis.core.server.RoutingContext;
 import org.apache.activemq.artemis.api.core.RoutingType;
 import org.apache.activemq.artemis.core.transaction.Transaction;
+import org.jboss.logging.Logger;
 
 public final class RoutingContextImpl implements RoutingContext {
+
+   private static final Logger logger  = Logger.getLogger(RoutingContextImpl.class);
 
    // The pair here is Durable and NonDurable
    private final Map<SimpleString, RouteContextList> map = new HashMap<>();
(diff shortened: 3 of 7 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -907,8 +907,7 @@ public RoutingStatus route(final Message message,
       AddressInfo addressInfo = addressManager.getAddressInfo(address);
 
       if (bindingMove != null) {
-         context.clear();
-         context.setReusable(false);
+         context.clear().setReusable(false);
```
michaelandrepearce: really is that needed? having on two lines helps when running debuggers

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/RoutingContext.java
```diff
@@ -82,6 +82,9 @@
 
    boolean isReusable(Message message, int version);
 
+   /** Provides human readable information about the routing, for tracing and debug */
+   String debug();
```
michaelandrepearce: Isn't that just a toString method then?

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
clebertsuconic: not needed, I'm just keeping how this is called everywhere else. I don't see it much different.. I was debugging it the whole day.

I can take it out from the diff.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RoutingContextImpl.java
```diff
@@ -128,6 +133,29 @@ public void addQueue(final SimpleString address, final Queue queue) {
       queueCount++;
    }
 
+   @Override
+   public String debug() {
```
michaelandrepearce: this looks very much like a toString method

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/RoutingContext.java
clebertsuconic: I wanted to keep it separate TBH.

message.toString() may include routing, and that will include too much information.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RoutingContextImpl.java
clebertsuconic: I want to keep it separate..  and call it explicitly from trace methods only.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/RoutingContext.java
michaelandrepearce: If thats the case does it need to be a method in interface? If during some custom debug, you can just cast the type to access the method. Not sure we' want it a public advertised method.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
clebertsuconic: I reverted these two lines. no big deal.

will mark this section as resolved.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/RoutingContext.java
clebertsuconic: fair enough. I will take it out of the interface.

clebertsuconic: it would look horrible though. on PostOfficeImpl, after routed I want to include the debug information on that case.

This would require a case, and test if RoutingContext implements RoutingContextImpl.

clebertsuconic: I think it's better to keep it.

clebertsuconic: know what.. I will just make it toString()

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RoutingContextImpl.java
clebertsuconic: solved

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/RoutingContext.java
clebertsuconic: solved

michaelandrepearce:
@clebertsuconic lgtm

whilst you're online could you look at https://github.com/apache/activemq-artemis/pull/2655 its a few line change just to get exception to exception listeners for spring jms.

clebertsuconic:
@michaelandrepearce thanks.. i will merge it

PR #2661, opened 2019-05-07
ARTEMIS-2319 Upgrading jboss-logging dependencies

ehsavoie:
* Upgrading versions
* Adding wildfly-common dependency as jboss-logmanager now depends on it
for simple common operations such as getting hostname or process id
* Updating bootclasspath with wildfly-common

Jira: https://issues.apache.org/jira/browse/ARTEMIS-2319

### DIFF
--- a/artemis-cdi-client/pom.xml
+++ b/artemis-cdi-client/pom.xml
@@ -56,6 +56,10 @@
          <artifactId>jboss-logmanager</artifactId>
       </dependency>
       <dependency>
+          <groupId>org.wildfly.common</groupId>
+          <artifactId>wildfly-common</artifactId>
+      </dependency>
+      <dependency>
          <groupId>org.apache.activemq</groupId>
          <artifactId>artemis-core-client</artifactId>
          <version>${project.version}</version>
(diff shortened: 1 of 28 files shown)


michaelandrepearce:
@clebertsuconic , i see you merged this wildfly is LPGL https://github.com/wildfly/wildfly/blob/master/LICENSE.txt, this isn't a cat A license, can we please revert this.

jbertram:
From what I can tell, Wildfly Common is ASL 2.0. See https://github.com/wildfly/wildfly-common/blob/master/LICENSE.

michaelandrepearce:
As per noted, the general wildfly project is LGPL, so when code is added or moved there is risk of LGPL contamination.

As there already is unfortunately common has code in it the LGPL headers.

 https://github.com/wildfly/wildfly-common/blob/1.5/src/test/java/org/wildfly/common/string/CompositeCharSequenceTestCase.java

It is best we revert this, and moving forwards a solution not needing a wildfly dep, will be better, else we run risk of code within their code base getting contaminated with LGPL.

clebertsuconic:
@ehsavoie Can you either revert this on a Pull Request, or request an updated license on wildflycommons?

The project is licensed as ASL, but I believe these headers are there for a mistake.

gemmellr:
For later reference, if anyone tries to follow along later, related discussion on dev list (around the artifact being used not being affected):
https://lists.apache.org/thread.html/ea2a8875d362158720a121c8ae2b6845bdb27614715cf48dae3000be@%3Cdev.activemq.apache.org%3E

ehsavoie:
https://github.com/wildfly/wildfly-common/pull/46  the PR for wildfly-common for fixing the missing / wrong license headers

PR #2666, opened 2019-05-13
ARTEMIS-2336 Use zero copy to replicate journal/page/large message file

wy96f:
Use sendfile to replicate files during backup sever bootstrap.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
@@ -16,6 +16,8 @@
  */
 package org.apache.activemq.artemis.core.protocol.core.impl;
 
+import java.io.RandomAccessFile;
+import java.nio.channels.FileChannel;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -25,6 +27,7 @@ import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
+import io.netty.channel.ChannelFutureListener;
 import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
 import org.apache.activemq.artemis.api.core.ActiveMQInterruptedException;
@@ -274,6 +277,60 @@ public final class ChannelImpl implements Channel {
       }
    }
 
+   private ActiveMQBuffer beforeSend(final Packet packet, final int reconnectID) {
+      packet.setChannelID(id);
+
+      if (responseAsyncCache != null && packet.isRequiresResponse() && packet.isResponseAsync()) {
+         packet.setCorrelationID(responseAsyncCache.nextCorrelationID());
+      }
+
+      if (logger.isTraceEnabled()) {
+         logger.trace("RemotingConnectionID=" + (connection == null ? "NULL" : connection.getID()) + " Sending packet nonblocking " + packet + " on channelID=" + id);
+      }
+
+      ActiveMQBuffer buffer = packet.encode(connection);
+
+      lock.lock();
+
+      try {
+         if (failingOver) {
+            waitForFailOver("RemotingConnectionID=" + (connection == null ? "NULL" : connection.getID()) + " timed-out waiting for fail-over condition on non-blocking send");
+         }
+
+         // Sanity check
+         if (transferring) {
+            throw ActiveMQClientMessageBundle.BUNDLE.cannotSendPacketDuringFailover();
+         }
+
+         if (resendCache != null && packet.isRequiresConfirmations()) {
+            addResendPacket(packet);
+         }
+
+      } finally {
+         lock.unlock();
+      }
+
+      if (logger.isTraceEnabled()) {
+         logger.trace("RemotingConnectionID=" + (connection == null ? "NULL" : connection.getID()) + " Writing buffer for channelID=" + id);
+      }
+
+      checkReconnectID(reconnectID);
+
+      //We do this outside the lock as ResponseCache is threadsafe and allows responses to come in,
+      //As the send could block if the response cache cannot add, preventing responses to be handled.
+      if (responseAsyncCache != null && packet.isRequiresResponse() && packet.isResponseAsync()) {
+         while (!responseAsyncCache.add(packet)) {
+            try {
+               Thread.sleep(1);
+            } catch (Exception e) {
+               // Ignore
+            }
+         }
+      }
+
+      return buffer;
+   }
+
    // This must never called by more than one thread concurrently
    private boolean send(final Packet packet, final int reconnectID, final boolean flush, final boolean batch) {
       if (invokeInterceptors(packet, interceptors, connection) != null) {
@@ -281,60 +338,43 @@ public final class ChannelImpl implements Channel {
       }
 
       synchronized (sendLock) {
-         packet.setChannelID(id);
-
-         if (responseAsyncCache != null && packet.isRequiresResponse() && packet.isResponseAsync()) {
-            packet.setCorrelationID(responseAsyncCache.nextCorrelationID());
-         }
-
-         if (logger.isTraceEnabled()) {
-            logger.trace("RemotingConnectionID=" + (connection == null ? "NULL" : connection.getID()) + " Sending packet nonblocking " + packet + " on channelID=" + id);
-         }
-
-         ActiveMQBuffer buffer = packet.encode(connection);
-
-         lock.lock();
-
-         try {
-            if (failingOver) {
-               waitForFailOver("RemotingConnectionID=" + (connection == null ? "NULL" : connection.getID()) + " timed-out waiting for fail-over condition on non-blocking send");
-            }
-
-            // Sanity check
-            if (transferring) {
-               throw ActiveMQClientMessageBundle.BUNDLE.cannotSendPacketDuringFailover();
-            }
-
-            if (resendCache != null && packet.isRequiresConfirmations()) {
-               addResendPacket(packet);
-            }
-
-         } finally {
-            lock.unlock();
-         }
-
-         if (logger.isTraceEnabled()) {
-            logger.trace("RemotingConnectionID=" + (connection == null ? "NULL" : connection.getID()) + " Writing buffer for channelID=" + id);
-         }
-
-         checkReconnectID(reconnectID);
-
-         //We do this outside the lock as ResponseCache is threadsafe and allows responses to come in,
-         //As the send could block if the response cache cannot add, preventing responses to be handled.
-         if (responseAsyncCache != null && packet.isRequiresResponse() && packet.isResponseAsync()) {
-            while (!responseAsyncCache.add(packet)) {
-               try {
-                  Thread.sleep(1);
-               } catch (Exception e) {
-                  // Ignore
-               }
-            }
-         }
+         ActiveMQBuffer buffer = beforeSend(packet, reconnectID);
 
          // The actual send must be outside the lock, or with OIO transport, the write can block if the tcp
          // buffer is full, preventing any incoming buffers being handled and blocking failover
          try {
             connection.getTransportConnection().write(buffer, flush, batch);
+         } catch (Throwable t) {
+            //If runtime exception, we must remove from the cache to avoid filling up the cache causing it to be full.
+            //The client would get still know about this as the exception bubbles up the call stack instead.
+            if (responseAsyncCache != null && packet.isRequiresResponse() && packet.isResponseAsync()) {
+               responseAsyncCache.remove(packet.getCorrelationID());
+            }
+            throw t;
+         }
+         return true;
+      }
+   }
+
+   @Override
+   public boolean send(Packet packet,
+                       RandomAccessFile raf,
+                       FileChannel fileChannel,
+                       long offset,
+                       int dataSize,
+                       Callback callback) {
+      if (invokeInterceptors(packet, interceptors, connection) != null) {
+         return false;
+      }
+
+      synchronized (sendLock) {
+         ActiveMQBuffer buffer = beforeSend(packet, -1);
+
+         // The actual send must be outside the lock, or with OIO transport, the write can block if the tcp
+         // buffer is full, preventing any incoming buffers being handled and blocking failover
+         try {
+            connection.getTransportConnection().write(buffer);
+            connection.getTransportConnection().write(raf, fileChannel, offset, dataSize, callback == null ? null : (ChannelFutureListener) future -> callback.done(future == null || future.isSuccess()));
          } catch (Throwable t) {
             //If runtime exception, we must remove from the cache to avoid filling up the cache causing it to be full.
             //The client would get still know about this as the exception bubbles up the call stack instead.
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
@@ -16,7 +16,10 @@
  */
 package org.apache.activemq.artemis.core.remoting.impl.netty;
 
+import java.io.IOException;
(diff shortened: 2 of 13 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationSyncFileMessage.java
```diff
@@ -197,6 +232,15 @@ public void decodeRest(final ActiveMQBuffer buffer) {
       }
    }
 
+   private void readFile(ByteBuffer buffer) {
```
franz1981: You can create directly the FileChannel without using FileInputStream

franz1981:
I'm yet reviewing it, but do you have already some numbers to show how it performs in comparison with the original version? Great work!

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
```diff
@@ -345,21 +351,41 @@ private boolean canWrite(final int requiredCapacity) {
...
-      final int readableBytes = buffer.readableBytes();
+      //no need to lock because the Netty's channel is thread-safe
+      //and the order of write is ensured by the order of the write calls
+      final Channel channel = this.channel;
+
+      int readableBytes = buffer.readableBytes();
+      Object fileObject = null;
+      if (buffer instanceof FileDelegatedChannelBufferWrapper) {
+         readableBytes += ((FileDelegatedChannelBufferWrapper)buffer).getDataSize();
+         fileObject = getFileObject((FileDelegatedChannelBufferWrapper)buffer);
```
franz1981: on each NettyConnection::write it is supposed to open a different RandomAccessFile (with SSL)  or a new FileChannel?
It will happen just once for each File or several times?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
```diff
@@ -395,19 +425,18 @@ private static void flushAndWait(final Channel channel, final ChannelPromise pro
...
-         return channel.write(bytes, promise);
-      } else if (nextBatchSize == batchLimit) {
-         return channel.writeAndFlush(bytes, promise);
+      }
+      if (fileObject == null) {
+         return nextBatchSize == batchLimit ? channel.writeAndFlush(bytes, promise) : channel.write(bytes, promise);
       } else {
-         //let netty use its write batching ability
-         return channel.write(bytes, promise);
+         channel.write(bytes);
```
franz1981: The ChannelFuture associated with this write is lost here

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
```diff
@@ -395,19 +425,18 @@ private static void flushAndWait(final Channel channel, final ChannelPromise pro
...
+         return nextBatchSize == batchLimit ? channel.writeAndFlush(fileObject, promise) : channel.write(fileObject, promise);
```
franz1981: Try to check if it makes sense to batch writes if we are using ChunkedFile/DefaultFileRegion

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationSyncFileMessage.java
```diff
@@ -150,30 +161,54 @@ public void encodeRest(final ActiveMQBuffer buffer) {
...
+         }
+         buffer = createPacket(connection, bufferSize);
+
+         encodeHeader(buffer);
+
+         encodeRest(buffer, connection);
+
+         encodeSize(buffer, expectedEncodeSize());
+
+         return fileId != -1 && dataSize > 0 ? new FileDelegatedChannelBufferWrapper(file, offset, dataSize, buffer) : buffer;
```
franz1981: It implies that the file is being read lazily by Netty to apply the zero-copy transfert, but it means also that we need to be sure that the file won't be moved or deleted util the zero-copy transfert will finish

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationSyncFileMessage.java
```diff
@@ -197,6 +232,15 @@ public void decodeRest(final ActiveMQBuffer buffer) {
       }
    }
 
+   private void readFile(ByteBuffer buffer) {
+      try (FileInputStream fis = new FileInputStream(file); FileChannel channel = fis.getChannel()) {
+         channel.position(offset);
+         channel.read(buffer);
```
franz1981: For this purpose https://docs.oracle.com/javase/6/docs/api/java/nio/channels/FileChannel.html#read(java.nio.ByteBuffer, long) it more efficient. But is ok to open/close every time the FileChannel just to read a chunk of it?

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationSyncFileMessage.java
wy96f: The file is opend before sending and closed after finished. Transfer won't be affected if the file is removed or deleted bcs it's referenced?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
wy96f: Emm, i think we just need ChannelFuture of the last write(i.e. the file transfer here)?

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationSyncFileMessage.java
wy96f: We only get here for the InVM connection where we use in our test, so it doesn't matter with the performance? If needed, we can pass another parameter into constructor like fileChannel?

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationSyncFileMessage.java
wy96f: Could you elaborate this? I thought FileChannel can only be created by FileInputStream/FileOutputStream/RandomAccessFile.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
wy96f: Good point. I'll work on it:)

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationSyncFileMessage.java
franz1981: FileChannel::open

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
franz1981: No idea, depends..if the first one will fail and the second one succeed is not good...Take a look if is possible to create a composed future or (even better)  a single object that encapsulate both the things we want to write in sequence as a whole

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationSyncFileMessage.java
franz1981: > We only get here for the InVM connection where we use in our test, so it doesn't matter with the performance

Not sure about it, EAP and other products based on Artemis use InVM connections...
maybe not with replication @jbertram @clebertsuconic any ideas?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
clebertsuconic: I would rather create a new method to be called from ReplicationManager.sendLargeFile instead of adding an instanceof here.

That is the caller should decide the semantic here, and a proper method to be created.

I would really try to avoid breaking encapsulation here. Not much for a nit-pick, but because this is hot code, meaning everything goes through here. I would rather create a new method for the optimization than to make everything else go through this for an exceptional case.



a question I have is: Was this really an issue for you during a production env? this was using a PooledBuffer, and it was only happening during catch up or large message transfer.

wy96f:
@franz1981 @clebertsuconic  Thanks for the review. I've modified as you commented.  There is no issue in our production env using buffer send:) We just want to squeeze performance(sometimes tens of GB files needs to be transferred although it's only happening during catchup). Also in a few months, we'll use 10-gigabit cards for our cluster which maybe make cathup more fast by using sendfile.  We deployed the test and saw that the change saved a visible amount of CPU. 
Flamegraph of master:
Flamegraph-orig

Flamegraph of the PR:
Flamegraph-zero

As we can see, at least 29% of the total time(21% FileChannelImpl::read and 8% PacketImpl::encode) is saved.

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationSyncFileMessage.java
michaelandrepearce: InVM is very important for performance. There is a use case at a company i know heavily reliant on InVM perf

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/invm/InVMConnection.java
```diff
@@ -237,6 +240,22 @@ public void run() {
...
 
+   @Override
+   public void write(ActiveMQBuffer buffer,
+                     RandomAccessFile raf,
+                     FileChannel fileChannel,
+                     long offset,
+                     int dataSize,
+                     boolean closeChannel) {
+      try {
+         if (closeChannel)
```
franz1981: This method is correct that's empty?
Anyway will use a proper logger here

artemis-server/src/main/java/org/apache/activemq/artemis/core/replication/ReplicationManager.java
```diff
@@ -551,49 +580,52 @@ private void sendLargeFile(AbstractJournalStorageManager.JournalContent content,
...
+               flushReplicationStream(action);
+            }
+            if (toSend == 0 || maxBytesToSend == 0)
+               break;
          }
          flushReplicationStream(action);
+
+      } catch (Exception e) {
+         if (fileChannel != null)
+            fileChannel.close();
```
franz1981: And raf? 
Maybe would be better to mark the expected last packet and make netty able to register a listener and let it close it on deallocate (after a successful/failed write)?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
```diff
@@ -385,6 +402,34 @@ public final void write(ActiveMQBuffer buffer,
...
+      }
+
+      //no need to lock because the Netty's channel is thread-safe
+      //and the order of write is ensured by the order of the write calls
+      final Channel channel = this.channel;
+      final ByteBuf bytes = buffer.byteBuf();
+      assert readableBytes >= 0;
+      channel.write(bytes);
+      channel.writeAndFlush(getFileObject(raf, fileChannel, offset, dataSize)).addListener(future -> {
+         if ((closeChannel || !future.isSuccess()) && fileChannel.isOpen()) {
```
franz1981: I've proposed a different solution on caller side to make this more efficient and effective

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
```diff
@@ -385,6 +402,34 @@ public final void write(ActiveMQBuffer buffer,
       }
    }
 
+   @Override
+   public void write(ActiveMQBuffer buffer,
```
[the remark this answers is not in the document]
wy96f: Do you mean we call write twice, first write buffer api, second new write file api in channel::send()? This way we will call netty channel::writeAndFlush twice, right?

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/invm/InVMConnection.java
wy96f: In ReplicationSyncFileMessage::encode, the file is read into buffer for non netty connection, so we just need to send buffer.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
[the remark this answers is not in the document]
franz1981: Depends if we are using batching or not: it depends on the writeable bytes on the channel

artemis-server/src/main/java/org/apache/activemq/artemis/core/replication/ReplicationManager.java
wy96f: You mean on caller side we add a callback where is called in netty channelFutureListener after write?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
```diff
@@ -274,67 +276,103 @@ private void waitForFailOver(String timeoutMsg) {
...
-            if (resendCache != null && packet.isRequiresConfirmations()) {
-               addResendPacket(packet);
+      //We do this outside the lock as ResponseCache is threadsafe and allows responses to come in,
+      //As the send could block if the response cache cannot add, preventing responses to be handled.
+      if (responseAsyncCache != null && packet.isRequiresResponse() && packet.isResponseAsync()) {
+         while (!responseAsyncCache.add(packet)) {
+            try {
+               Thread.sleep(1);
+            } catch (Exception e) {
+               // Ignore
```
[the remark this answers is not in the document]
wy96f: It seems responseAsyncCache::add always returns true, so the sleep code doesn't make sense?

artemis-server/src/main/java/org/apache/activemq/artemis/core/replication/ReplicationManager.java
franz1981: Yes, but just the supposed last packet need that listener, while the other writes can just use a void promise instead

wy96f:
@franz1981 Thanks for the great suggestions. All should be done:)

franz1981:
@wy96f Just a qq: `DefaultFileRegion` is supported by windows/MacOS too?

wy96f:
> @wy96f Just a qq: `DefaultFileRegion` is supported by windows/MacOS too?

It should work on both. I tried it successfully in windows. Wikipedia notes "macOS supports zero-copy through the FreeBSD portion of the kernel", should be sendfile system call i guess.

franz1981:
@wy96f 
I'm more concerned about Windows TBH: I don't know if Netty supports zero copy on it...
let me take a look :+1:

michaelandrepearce:
Where are we with this id like to merge it? It seems to run fine locally on windows. Any other concerns?

jbertram:
@franz1981, is this ready to merge?

franz1981:
@jbertram I would like to profile the improvement, but it seems ok to me bud, can be merged!

PR #2667, opened 2019-05-13
ARTEMIS-2321 Non-blocking Page::read on page cache

franz1981:
It includes just the scalability improvements of https://github.com/apache/activemq-artemis/pull/2645.
It is addressing the case where N different pages need to be read concurrently from `PageCursorProviderImpl`: in the original version the `Page::read` would be serialized and will wait each others to happen to make the cache available again, while this change allow to read concurrently different pages without blocking the whole cache.

### DIFF
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImplTest.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImplTest.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.core.paging.cursor.impl;
+
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.activemq.artemis.core.paging.PagingStore;
+import org.apache.activemq.artemis.core.paging.cursor.PageCache;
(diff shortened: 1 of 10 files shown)


franz1981:
@clebertsuconic I've separated the scalability fix from the GC improvements re paging in this PR.

pom.xml
```diff
@@ -1624,6 +1624,8 @@
                   <exclude>activemq-artemis-native/src/main/c/org_apache_activemq_artemis_jlibaio_LibaioContext.h</exclude>
                   <exclude>**/dependency-reduced-pom.xml</exclude>
 
+                  <!-- Mockito -->
+                  <exclude>**/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker</exclude>
```
clebertsuconic: what is this? (simple question. .no review intended here)

artemis-server/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
clebertsuconic: is this change accidental?

franz1981: Nope, has to be done with mockito if you want to override final class behaviours

pom.xml
franz1981: This file is being used to allow Mockito to use a special new engine that make possible to override method calls of final classes

artemis-server/src/test/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImplTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class PageCursorProviderImplTest {
+
+   @Test(timeout = 30_000)
+   public void shouldAllowsConcurrentPageReads() throws Exception {
```
clebertsuconic: just nit-picking the name of the test.. shouldAllowConcurrentPageReads() 

(should *allow*....)

franz1981: uh right..and I see that there are conflicts...
let me me fix them...

franz1981: done! :+1:

PR #2671, opened 2019-05-16
ARTEMIS-2344 return security errors for unauthorized anonymous sasl

brusdev:
When user attempts unauthorized anonymous sasl the broker can return an
error of 'failed' instead of the security error that is expected in
these cases.


tabish121:
Certainly seems worthy of a JIRA to me.

clebertsuconic:
Yeah.. you need a JIrA

PR #2675, opened 2019-05-21
ARTEMIS-2347 JournalStorageManager::stopReplication can deadlock while stopping

franz1981:
AbstractJournalStorageManager::performCachedLargeMessageDeletes
must enforce acquisition of manager write lock (as documented)
to avoid unlucky racing calls of stopReplication while stopping
to deadlock.

### DIFF
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManagerTest.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManagerTest.java
@@ -15,134 +15,153 @@
  */
 package org.apache.activemq.artemis.core.persistence.impl.journal;
 
+import java.util.ArrayList;
 import java.util.Collection;
-import java.util.List;
-import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Future;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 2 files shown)


artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManagerTest.java
```diff
@@ -145,4 +166,73 @@ public void testFixJournalFileSize() {
       Assert.assertEquals(4096, manager.fixJournalFileSize(4098, 4096));
       Assert.assertEquals(8192, manager.fixJournalFileSize(8192, 4096));
    }
+
+   @Test(timeout = 20_000)
```
clebertsuconic: This test is hard coded to libaio.

Please, either make it runnable in other platforms, or add an Assume call to ignore the test in case of libaio:

 org.junit.Assume.assumeTrue("Test case needs AIO to run", AIOSequentialFileFactory.isSupported());


It's probably best if you move the new test to a new test class, and add this on the @BeforeClass:


```java

   @BeforeClass
   public static void hasAIO() {
      org.junit.Assume.assumeTrue("Test case needs AIO to run", AIOSequentialFileFactory.isSupported());
   }
```


An example of a class playing with Assume is AIOImportExportTest.


But you also have the option to make it runnable with NIO. But if not possible, please add the Assume call here.

franz1981: :+1: Thanks to have taken a look bud, please check if now is ok (I don't have a Mac/Win to test it!)

clebertsuconic:
The use of theory here is really odd!

If the test failed on NIO, there's no output on my IDE. it simply says.. failure.. but not why.. and what method.


In the future, it would be better to use @Parameter... at least the test would show up as a separate test.

I will make a change while merging it this afternoon.

franz1981:
> The use of theory here is really odd!

I can refactor into our common parametric form too: I've tried to make the jun it theory more verbose, but it seems a limit of this tool :(

PR #2676, opened 2019-05-22
ARTEMIS-2349 CLI FQQN name parsing is incorrect

gaohoward:
Fixed the wrong returned values in those methods.
Also move those methods into a util class to make
testing easy.


gaohoward:
I got the fqqn format wrong. The original is correct. I'll reject the jira and close this PR.

PR #2679, opened 2019-05-23
ARTEMIS-2352 - Add the ability to reject messages without a validated…

andytaylor:
… user set

https://issues.apache.org/jira/browse/ARTEMIS-2352

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/JMSXUserIDPlugin.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/JMSXUserIDPlugin.java
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.tests.integration.security;
+
+import org.apache.activemq.artemis.api.core.ActiveMQException;
+import org.apache.activemq.artemis.api.core.Message;
+import org.apache.activemq.artemis.core.server.ServerSession;
+import org.apache.activemq.artemis.core.server.plugin.ActiveMQServerPlugin;
+import org.apache.activemq.artemis.core.transaction.Transaction;
+
+import java.util.Map;
+
+public class JMSXUserIDPlugin implements ActiveMQServerPlugin {
+
+   private static String POPULATE_VALIDATED_USER = "POPULATE_VALIDATED_USER";
+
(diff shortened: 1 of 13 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/JMSXUserIDPlugin.java
```diff
@@ -0,0 +1,58 @@
...
+   @Override
+   public void init(Map<String, String> properties) {
+      populateValidatedUser = properties.getOrDefault(POPULATE_VALIDATED_USER, null);
+   }
+
+   @Override
+   public void beforeSend(ServerSession session, Transaction tx, Message message, boolean direct, boolean noAutoCreateQueue) throws ActiveMQException {
+      if (populateValidatedUser != null && !message.containsProperty(Message.HDR_VALIDATED_USER)) {
+         message.messageChanged();
+         message.putStringProperty(Message.HDR_VALIDATED_USER, populateValidatedUser);
```
clebertsuconic: this will not work on AMQ, you know that, right?
you would need to call message.reencode.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/RejectValidatedUserTest.java
```diff
@@ -0,0 +1,90 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+
+import javax.jms.Connection;
+import javax.jms.Message;
+import javax.jms.MessageProducer;
+import javax.jms.Queue;
+import javax.jms.Session;
+
+public class RejectValidatedUserTest  extends ActiveMQTestBase {
```
clebertsuconic: I think you would need to add AMQP tests at least here.

clebertsuconic:
There is a real failure as part of this. look on the PR check

andytaylor:
Let me fix the test

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/JMSXUserIDPlugin.java
andytaylor: can you explain what i need to change?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/JMSXUserIDPlugin.java
```diff
@@ -0,0 +1,58 @@
...
+    */
+   @Override
+   public void init(Map<String, String> properties) {
+      populateValidatedUser = properties.getOrDefault(POPULATE_VALIDATED_USER, null);
+   }
+
+   @Override
+   public void beforeSend(ServerSession session, Transaction tx, Message message, boolean direct, boolean noAutoCreateQueue) throws ActiveMQException {
+      if (populateValidatedUser != null && !message.containsProperty(Message.HDR_VALIDATED_USER)) {
+         message.messageChanged();
```
clebertsuconic: You don't need to call messageChanged... message.putStringProperty will do it for you

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/JMSXUserIDPlugin.java
clebertsuconic: you need to call message.reencode() right after this.


This is actually subject of a discussion upstream. We used to re-encode automatically, after a discussion about AMQP specification we are not supposed to change the body of an AMQP message, however we have the method reencode for cases where we want to bypass such rule.


I think we should add a strict toggle on the server, and do it automatically unless the strict toggle is turned on. on which case we wouldn't allow modifications at all.

PR #2680, opened 2019-05-23
ARTEMIS-2346 - make timestamp human readable in console

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-2346


jbertram:
This was already merged via #2674.

PR #2681, opened 2019-05-26
ARTEMIS-2308 Support exporting metrics

jbertram:
Based on the conversation from #2626 I decided to pursue Micrometer integration which required a completely different approach than the one I took before. There is now no direct integration with Prometheus. The broker uses Micrometer APIs and a plugin structure was added to support specific integrations. A Prometheus plugin is available at https://github.com/jbertram/artemis-prometheus-metrics-plugin.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
@@ -23,6 +23,8 @@ import java.util.Map;
  * An AddressControl is used to manage an address.
  */
 public interface AddressControl {
+   String ROUTED_MESSAGE_COUNT_DESCRIPTION = "number of messages routed to one or more bindings";
+   String UNROUTED_MESSAGE_COUNT_DESCRIPTION = "number of messages not routed to any bindings";
 
    /**
     * Returns the managed address.
@@ -101,19 +103,19 @@ public interface AddressControl {
    @Attribute(desc = "names of all bindings (both queues and diverts) bound to this address")
    String[] getBindingNames() throws Exception;
 
-   @Attribute(desc = "number of messages added to all the queues for this address")
+   @Attribute(desc = "number of messages currently in all queues bound to this address (includes scheduled, paged, and in-delivery messages)")
    long getMessageCount();
 
    /**
     * Returns the number of messages routed to one or more bindings
     */
-   @Attribute(desc = "number of messages routed to one or more bindings")
+   @Attribute(desc = ROUTED_MESSAGE_COUNT_DESCRIPTION)
    long getRoutedMessageCount();
 
    /**
     * Returns the number of messages not routed to any bindings
     */
-   @Attribute(desc = "number of messages not routed to any bindings")
+   @Attribute(desc = UNROUTED_MESSAGE_COUNT_DESCRIPTION)
    long getUnRoutedMessageCount();
 
 
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/AddressInfo.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/AddressInfo.java
@@ -16,14 +16,18 @@
  */
 package org.apache.activemq.artemis.core.server.impl;
 
-import org.apache.activemq.artemis.api.core.RoutingType;
-import org.apache.activemq.artemis.api.core.SimpleString;
-import org.apache.activemq.artemis.utils.CompositeAddress;
-import org.apache.activemq.artemis.utils.PrefixUtil;
-
 import java.util.EnumSet;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+
+import org.apache.activemq.artemis.api.core.RoutingType;
+import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.api.core.management.AddressControl;
+import org.apache.activemq.artemis.api.core.management.ResourceNames;
+import org.apache.activemq.artemis.core.server.metrics.AddressMetricNames;
+import org.apache.activemq.artemis.core.server.metrics.MetricsManager;
+import org.apache.activemq.artemis.utils.CompositeAddress;
+import org.apache.activemq.artemis.utils.PrefixUtil;
 
 public class AddressInfo {
 
@@ -191,4 +195,18 @@ public class AddressInfo {
       return unRoutedMessageCountUpdater.get(this);
    }
 
+   public void registerMeters(MetricsManager metricsManager) {
+      if (metricsManager != null) {
+         metricsManager.registerAddressGauge(name.toString(), builder -> {
+            builder.register(AddressMetricNames.ROUTED_MESSAGE_COUNT, this, metrics -> Double.valueOf(getRoutedMessageCount()), AddressControl.ROUTED_MESSAGE_COUNT_DESCRIPTION);
+            builder.register(AddressMetricNames.UNROUTED_MESSAGE_COUNT, this, metrics -> Double.valueOf(getUnRoutedMessageCount()), AddressControl.UNROUTED_MESSAGE_COUNT_DESCRIPTION);
+         });
+      }
+   }
+
+   public void unregisterMeters(MetricsManager metricsManager) {
+      if (metricsManager != null) {
+         metricsManager.remove(ResourceNames.ADDRESS + name);
+      }
+   }
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -50,6 +50,9 @@ import org.apache.activemq.artemis.api.core.RoutingType;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.api.core.management.CoreNotificationType;
 import org.apache.activemq.artemis.api.core.management.ManagementHelper;
+import org.apache.activemq.artemis.api.core.management.QueueControl;
+import org.apache.activemq.artemis.api.core.management.ResourceNames;
+import org.apache.activemq.artemis.core.PriorityAware;
 import org.apache.activemq.artemis.core.filter.Filter;
 import org.apache.activemq.artemis.core.io.IOCallback;
 import org.apache.activemq.artemis.core.paging.cursor.PagePosition;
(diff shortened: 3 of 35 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/MetricsManager.java
```diff
@@ -0,0 +1,125 @@
...
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Metrics;
+import io.micrometer.core.instrument.binder.jvm.JvmMemoryMetrics;
+import org.apache.activemq.artemis.api.core.management.ResourceNames;
+import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
+
+public class MetricsManager {
```
franz1981: That's a huge comment; API change proposal...
```java
   @FunctionalInterface
   public interface MetricGaugeBuilder {
      void register(String metricName, Object state, ToDoubleFunction f, String description);
   }

   public void registerQueueGauge(String address, String queue, Consumer<MetricGaugeBuilder> builder) {
      final List<Meter> newMeters = new ArrayList<>();
      builder.accept((metricName, state, f, description) -> {
         Meter meter = Gauge
            .builder("artemis." + metricName, state, f)
            .tag("broker", brokerName)
            .tag("address", address)
            .tag("queue", queue)
            .description(description)
            .register(meterRegistry);
         newMeters.add(meter);
      });
      final String resource = ResourceNames.QUEUE + queue;
      this.meters.compute(resource, (s, meters) -> {
         if (meters == null) {
            meters = new ArrayList<>();
         }
         meters.addAll(newMeters);
         return meters;
      });
   }

   public void registerAddressGauge(String address, Consumer<MetricGaugeBuilder> builder) {
      final List<Meter> newMeters = new ArrayList<>();
      builder.accept((metricName, state, f, description) -> {
         Meter meter = Gauge
            .builder("artemis." + metricName, state, f)
            .tag("broker", brokerName)
            .tag("address", address)
            .description(description)
            .register(meterRegistry);
         newMeters.add(meter);
      });
      final String resource = ResourceNames.ADDRESS + address;
      this.meters.compute(resource, (s, meters) -> {
         if (meters == null) {
            meters = new ArrayList<>();
         }
         meters.addAll(newMeters);
         return meters;
      });
   }

   public void registerBrokerGauge(Consumer<MetricGaugeBuilder> builder) {
      final List<Meter> newMeters = new ArrayList<>();
      builder.accept((metricName, state, f, description) -> {
         Meter meter = Gauge
            .builder("artemis." + metricName, state, f)
            .tag("broker", brokerName)
            .description(description)
            .register(meterRegistry);
         newMeters.add(meter);
      });
      final String resource = ResourceNames.BROKER + "." + brokerName;
      this.meters.compute(resource, (s, meters) -> {
         if (meters == null) {
            meters = new ArrayList<>();
         }
         meters.addAll(newMeters);
         return meters;
      });
   }
```

That could be used with:
```java
if (server != null && server.getMetricsManager() != null) {
         server.getMetricsManager().registerQueueGauge(queueName, addressName, builder -> {
            builder.register("message.count", pendingMetrics, metrics -> Double.valueOf(pendingMetrics.getMessageCount()), QueueControl.MESSAGE_COUNT_DESCRIPTION);
            builder.register(//...other useful metrics
```

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
```diff
@@ -101,19 +103,19 @@
    @Attribute(desc = "names of all bindings (both queues and diverts) bound to this address")
    String[] getBindingNames() throws Exception;
 
-   @Attribute(desc = "number of messages added to all the queues for this address")
+   @Attribute(desc = "number of messages currently in all queues bound to this address (includes scheduled, paged, and in-delivery messages)")
```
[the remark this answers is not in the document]
jbertram: I didn't refactor the description here because I'm not actually exported this metric since it will be easy to aggregate all the message counts of all the queues on the address in whatever system the metrics are exported to.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3852,6 +3859,41 @@ void scheduleSlowConsumerReaper(AddressSettings settings) {
       }
    }
 
+   private void registerMeters() {
+      String addressName = address.toString();
+      String queueName = name.toString();
+
+      if (server != null && server.getMetricsManager() != null) {
+         MetricsManager metricsManager = server.getMetricsManager();
+         metricsManager.registerQueueGauge("message.count", addressName, queueName, pendingMetrics, metrics -> Double.valueOf(pendingMetrics.getMessageCount()), QueueControl.MESSAGE_COUNT_DESCRIPTION);
```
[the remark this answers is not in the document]
jbertram: I'm not entirely sure what you mean. Do you mean declaring something like:

`static final String MESSAGE_COUNT = "message.count";`

And then using `MESSAGE_COUNT` in the method invocation?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/MetricsManager.java
```diff
@@ -0,0 +1,125 @@
...
+
+   private String brokerName;
+
+   private MeterRegistry meterRegistry;
+
+   private Map<String, List<Meter>> meters = new HashMap<>();
```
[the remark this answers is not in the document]
jbertram: Agreed. I'll change that.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/MetricsManager.java
```diff
@@ -0,0 +1,125 @@
...
+      meterRegistry = metricsPlugin.getRegistry();
+      Metrics.globalRegistry.add(meterRegistry);
+      new JvmMemoryMetrics().bindTo(meterRegistry);
+   }
+
+   public MeterRegistry getMeterRegistry() {
+      return meterRegistry;
+   }
+
+   public void registerQueueGauge(String metricName, String address, String queue, Object object, ToDoubleFunction f, String description) {
```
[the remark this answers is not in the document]
jbertram: This isn't possible because a `QueuePendingMessageMetrics` isn't always passed into the method (e.g. for all the queue metrics that aren't tracked by the `QueuePendingMessageMetrics` object).

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/MetricsManager.java
jbertram: Maybe it's just me, but I think the existing implementation is easier to read.  :)

Is there a particular benefit to your implementation? I'm assuming it has better performance for some esoteric reason. :)

franz1981: Maybe :P It has a better behaviour under load, because it won't call many times put on the concurrent hash map, just once...

docs/user-manual/en/metrics.md
```diff
@@ -0,0 +1,104 @@
+# Metrics
+
+Apache ActiveMQ Artemis can export metrics to a variety of monitoring systems
+via the [Micrometer](https://micrometer.io/) vendor-neutral application metrics
+facade.
+
+Important runtime metrics have been instrumented via the Micrometer API, and
+all a user needs to do is implement `org.apache.activemq.artemis.core.server.metrics.ActiveMQMetricsPlugin`
+in order to instantiate and configure a `io.micrometer.core.instrument.MeterRegistry`
+implementation. Many implementations of `MeterRegistr` are available from
```
michaelandrepearce: sentance seems to end abrubtly i assume you meant to put a link to the micrometer web page with a list of these?

michaelandrepearce:
Great stuff. +1 on this from me. Nice abstraction from implementations.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -1012,6 +1012,33 @@
             </xsd:complexType>
          </xsd:element>
 
+         <xsd:element name="metrics-plugin" maxOccurs="1" minOccurs="0">
```
michaelandrepearce: Is the test version of artemis-configuration.xsd updated also, to keep them in sync.

clebertsuconic:
It would be nice now to have the components repo, and an option to include it with the installation.
Great stuff...


is this ready to merge?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/AddressInfo.java
```diff
@@ -191,4 +194,16 @@ public long getUnRoutedMessageCount() {
...
 
+   public void registerMeters(MetricsManager metricsManager) {
+      if (metricsManager != null) {
+         metricsManager.registerAddressGauge("routed.message.count", name.toString(), this, metrics -> Double.valueOf(getRoutedMessageCount()), AddressControl.ROUTED_MESSAGE_COUNT_DESCRIPTION);
+         metricsManager.registerAddressGauge("unrouted.message.count", name.toString(), this, metrics -> Double.valueOf(getUnRoutedMessageCount()), AddressControl.ROUTED_MESSAGE_COUNT_DESCRIPTION);
+      }
+   }
+
+   public void unregisterMeters(MetricsManager metricsManager) {
+      if (metricsManager != null) {
```
franz1981: Given that `metricsManager` is volatile, it should be called in a local variable and if not null, used to call its method

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/MetricsManager.java
```diff
@@ -0,0 +1,125 @@
...
+      }
+      Meter meter = Gauge
+         .builder("artemis." + metricName, object, f)
+         .tag("broker", brokerName)
+         .tag("address", address)
+         .tag("queue", queue)
+         .description(description)
+         .register(meterRegistry);
+
+      meters.add(meter);
```
franz1981: This risk to be called by different threads: take a look to the implementation I've posted above (that is using the compute method, that is thread safe for a specific key).

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/AddressInfo.java
franz1981: Ups sorry, reading it from the web I haven't noticed from where `metricsManager` was coming

jbertram:
I just pushed the latest changes:

- Fixed documentation (Mike)
- Fixed XSD (Mike)
- Addressed thread-safety issues (Franz)
- Added static finals for metric names (Franz)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/MetricsManager.java
franz1981: And it is more safe concurrently, given that the overall configuration is consider an atomic operation

clebertsuconic:
@jbertram there still work to be improved, mainly when we get the repository, but this is good as is for now. We can send more commits later.

docs/user-manual/en/metrics.md
```diff
@@ -0,0 +1,105 @@
...
+**Broker**
+
+- connection.count
+- total.connection.count
+- address.memory.usage
+
+**Address**
+
+- routed.message.count
+- unrouted.message.count
```
michaelandrepearce: @jbertram what about paging size?

PR #2682, opened 2019-05-28
 ARTEMIS-2355: Marking message as changed after setting routing type, because it is set after divert

ldebello:
In this specific use case the Divert sets the routing type to Multicast but the message is not marked as modified so the encode is not done again and the message does not match the routing type in the target queue when there is cluster redistribution.

One possible alternative is not mark the message as changed in the setRoutingType and instead doing it from the divert and only do the mark if the routing type does not match and the message was really changed

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
@@ -1121,6 +1121,7 @@ public class AMQPMessage extends RefCountMessage {
       } else {
          setMessageAnnotation(AMQPMessageSupport.ROUTING_TYPE, routingType.getType());
       }
+      messageChanged();
       return this;
    }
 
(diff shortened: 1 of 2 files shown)


clebertsuconic:
I'm making a few modifications.. the issue is that the Bridge had made a copy of the message, but it should call changed at the copy.

Same with the Bridge as it should have made a copy over the Bridge.


the message source should be kept untouched. and this was an issue on the Bridge, which I'm fixing here.

jbertram:
I'm taking a look at this as well as I'm not sure the fix is even necessary at this point. Please wait to merge.

jbertram:
Ignore my previous comment. Proceed @clebertsuconic.

ldebello:
@clebertsuconic I saw the message copied by the Bridge it sounds weird to me call the messageChanged in a recently copied message or you mean marking the messagedChanged after setting the routing type in the message copied?

clebertsuconic:
I am closing this as this is superceeded by #2686

clebertsuconic:
@ldebello copy is creating a new version of the message.

messageChanged is what will allow re-encoding of anything that was changed. On this case the messageChanged is to allow the serialization of the routingType and address after the copy.

clebertsuconic:
Which was a valid issue BTW: Thanks for reporting it with the test. The only thing that I did not want to allow was automatic changes. Due to AMQP constraints we need to explicitly have a method to force re-encoding of the message by either calling reencode of messageChanged.

PR #2683, opened 2019-05-29
NO-JIRA typo on logger

wy96f:


artemis-server/src/main/java/org/apache/activemq/artemis/core/replication/ReplicationManager.java
```diff
@@ -613,7 +613,7 @@ private void flushReplicationStream(FlushAction action) throws Exception {
       action.reset();
       replicationStream.execute(action);
       if (!action.await(this.timeout, TimeUnit.MILLISECONDS)) {
```
clebertsuconic: I think the typo was the other way around? actiona.await(initialReplicationSynctimeout?)

wy96f: The docs says initial-replication-sync-timeout means "the amount of time the replicating server will wait at the completion of the initial replication process for the replica to acknowledge it has received all the necessary data. The default is 30,000 milliseconds. **Note:** during this interval any journal related operations will be blocked". So it's the time waiting for the last ack and i assume "timeout" should be used in the middle of replication process?

michaelandrepearce: @clebertsuconic @wy96f  whats the situation on this one? is the PR valid?

clebertsuconic: @wy96f / @michaelandrepearce from what I looked, the correct fix would be:

```java
     if (!action.await(this.initialReplicationTimeout, TimeUnit.MILLISECONDS))
```

clebertsuconic: the typo is not on the logger, it's on the await.


otherwise initialReplicationTimeout is not used anywhere and it could go away.

PR #2686, opened 2019-05-29
ARTEMIS-2361 Bridge should make a copy of messages

clebertsuconic:
Same as Divert

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
@@ -550,7 +550,15 @@ public class BridgeImpl implements Bridge, SessionFailureListener, SendAcknowled
    }
 
    /* Hook for processing message before forwarding */
-   protected Message beforeForward(final Message message, final SimpleString forwardingAddress) {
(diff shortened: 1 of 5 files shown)


ldebello:
@clebertsuconic I know you close https://github.com/apache/activemq-artemis/pull/2682 because this ticket will solve it, but the changes here seems against the idea of leaving the responsibility to transform to reencode the message when adding property (https://github.com/apache/activemq-artemis/pull/2634). Because as far as understand marking the message as changed in the divert will reencode it in the future (https://github.com/apache/activemq-artemis/pull/2686/files#diff-565f29f5e7c3e5234faff2352404eedbR129). I am not totally sure but based on the changes I think the properties added by a transformer in the divert will work out of the box but properties added in a transform in a bridge will not work, so that could be misleading.

clebertsuconic:
@ldebello The Divert makes a copy of the message before anyway... I want to avoid a case where the broker changes the message unadvertadely .  AMQP has a few specs that would be broken if I let that happen.

the way to go is to call messageChanged on the copy. and this is working fine, right?


I went further and looked at the Bridge which was also changing the root message instead of making a copy.

ldebello:
For my the change is ok, I am just point that now we can remove reencode from transformer because the message changed will do the trick, I copied your changes to my branch and include some application properties in the transformer without doing reencode and that works, also I remove the reencode from the test included in the PR mentioned above and that also works

PR #2687, opened 2019-05-30
ARTEMIS-2359 Upgrade to Guava 24.1

brusdev:
CVE-2018-10237 guava: Unbounded memory allocation in AtomicDoubleArray
and CompoundOrdering classes allow remote attackers to cause a denial
of service.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
       <commons.beanutils.version>1.9.3</commons.beanutils.version>
       <commons.collections.version>3.2.2</commons.collections.version>
       <fuse.mqtt.client.version>1.14</fuse.mqtt.client.version>
-      <guava.version>19.0</guava.version>
+      <guava.version>24.1-jre</guava.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -800,6 +800,12 @@
             <scope>provided</scope>
             <!-- License: Apache 2.0 -->
          </dependency>
+         <dependency>
```
clebertsuconic: Guava is only used by airlift.. Do we really need to add the dependency globally?

clebertsuconic:
can you rebase please?

pom.xml
brusdev: I see your point but airlift/airline is added as global dependency and Guava is used by artemis-cli (XmlDataExporterUtil), by integrations-tests (MethodCalledVerifier) by airlift/airline and by hawtio/web.
Airlift/airline is added as global dependency, as artemis-cli dependency and as artemis-distribution dependency.
Hawtio/web is added as artemis-console dependency.

brusdev:
rabased

PR #2689, opened 2019-05-30
NO-JIRA Test Improvements

clebertsuconic:


artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -174,6 +175,9 @@
    @Rule
    public CleanupSystemPropertiesRule propertiesRule = new CleanupSystemPropertiesRule();
 
+   @Rule
+   public PortCheckRule portCheckRule = new PortCheckRule(61616, 61617, 61618);
```
michaelpearce-gain: Do we need tests actually bound to sepecific ports, failing if not free? would it not be better to simply find some free ports and use those? Ive seen this in other projects, its quite an elegant way of ensuring you dont get port clash issues that fail builds un-neededly. (and allows concurrent builds in a shared env)

michaelpearce-gain: e.g. you could look at https://docs.spring.io/spring/docs/4.0.5.RELEASE/javadoc-api/org/springframework/util/SocketUtils.html#findAvailableTcpPort--

or even more rough:
        try (ServerSocket socket = new ServerSocket(0)) {
            socket.setReuseAddress(true);
            return socket.getLocalPort();
        }

clebertsuconic: @michaelandrepearce 61616 is the broker default port, right? We already have such validations on the testsuite. Refactoring to find a free port would be a way too heavy refactoring that I'm not ready to do. (thousands of tests).

Besides it's sane to have such rules anyways.

This is to validate if the broker is not leaked. I was dealing with a specific failure on my CI here.

clebertsuconic:
this is now ready to be merged (as long as the testsuite is passing).

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
clebertsuconic: BTW: I'm only checking for the default port now (this was just for a test I was doing.. the final version is only checking for the default port).

PR #2691, opened 2019-06-01
ARTEMIS-2364 collision avoidance for redelivery

jbertram:
This is a feature from 5.x implemented via
https://issues.apache.org/jira/browse/AMQ-747.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
@@ -166,6 +166,8 @@ public final class FileConfigurationParser extends XMLConfigurationUtil {
 
    private static final String REDELIVERY_DELAY_MULTIPLIER_NODE_NAME = "redelivery-delay-multiplier";
 
+   private static final String REDELIVERY_COLLISION_AVOIDANCE_FACTOR_NODE_NAME = "redelivery-collision-avoidance-factor";
+
    private static final String MAX_REDELIVERY_DELAY_NODE_NAME = "max-redelivery-delay";
 
    private static final String MAX_DELIVERY_ATTEMPTS = "max-delivery-attempts";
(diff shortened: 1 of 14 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3458,9 +3460,15 @@ private long calculateRedeliveryDelay(final AddressSettings addressSettings, fin
       long redeliveryDelay = addressSettings.getRedeliveryDelay();
       long maxRedeliveryDelay = addressSettings.getMaxRedeliveryDelay();
       double redeliveryMultiplier = addressSettings.getRedeliveryMultiplier();
+      double collisionAvoidanceFactor = addressSettings.getRedeliveryCollisionAvoidanceFactor();
 
       int tmpDeliveryCount = deliveryCount > 0 ? deliveryCount - 1 : 0;
       long delay = (long) (redeliveryDelay * (Math.pow(redeliveryMultiplier, tmpDeliveryCount)));
+      if (collisionAvoidanceFactor > 0) {
+         Random random = RandomUtil.getRandom();
```
franz1981: RandomUtil return a Thread local random?

jbertram: I'm not sure what you mean. It returns a `static final` instance of `java.util.Random`. Is that what you're asking about?

franz1981: I mean this https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ThreadLocalRandom.html
Random is synchronized on so it would block any concurrent calls

jbertram: Fair enough. I've pushed the change.

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -1037,6 +1039,11 @@ public SecuritySettingPlugin run() {
             addressSettings.setRedeliveryDelay(XMLUtil.parseLong(child));
          } else if (REDELIVERY_DELAY_MULTIPLIER_NODE_NAME.equalsIgnoreCase(name)) {
             addressSettings.setRedeliveryMultiplier(XMLUtil.parseDouble(child));
+         } else if (REDELIVERY_COLLISION_AVOIDANCE_FACTOR_NODE_NAME.equalsIgnoreCase(name)) {
+            double redeliveryCollisionAvoidanceFactor = XMLUtil.parseDouble(child);
+            Validators.GE_ZERO.validate(REDELIVERY_COLLISION_AVOIDANCE_FACTOR_NODE_NAME, redeliveryCollisionAvoidanceFactor);
+            Validators.LE_ONE.validate(REDELIVERY_COLLISION_AVOIDANCE_FACTOR_NODE_NAME, redeliveryCollisionAvoidanceFactor);
+            addressSettings.setRedeliveryCollisionAvoidanceFactor(redeliveryCollisionAvoidanceFactor);
```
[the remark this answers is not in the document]
jbertram: For continuity with 5.x I think the configuration should use `redelivery-collision-avoidance-factor` since 5.x uses `collisionAvoidanceFactor`. That said, this is really just a matter of opinion and not strict correctness. It could be implemented any number of ways.

PR #2694, opened 2019-06-05
ARTEMIS-2371 Message with huge header shuts broker down

brusdev:
Add max record size check before adding a record to prevent that the
broker shuts down, when there is one really large header sent with the
message.

### DIFF
--- a/artemis-journal/src/main/java/org/apache/activemq/artemis/journal/ActiveMQJournalBundle.java
+++ b/artemis-journal/src/main/java/org/apache/activemq/artemis/journal/ActiveMQJournalBundle.java
@@ -47,4 +47,7 @@ public interface ActiveMQJournalBundle {
 
    @Message(id = 149004, value = "unable to open file")
    String unableToOpenFile();
+
+   @Message(id = 149005, value = "Message of {0} bytes is bigger than the max record size of {1} bytes. You should try to move large application properties to the message body.", format = Message.Format.MESSAGE_FORMAT)
+   ActiveMQIOErrorException recordLargerThanStoreMax(long recordSize, long maxRecordSize);
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
@@ -63,6 +63,7 @@ import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
 import org.apache.activemq.artemis.core.server.JournalType;
 import org.apache.activemq.artemis.core.server.LargeServerMessage;
 import org.apache.activemq.artemis.core.server.files.FileStoreMonitor;
+import org.apache.activemq.artemis.journal.ActiveMQJournalBundle;
 import org.apache.activemq.artemis.utils.ExecutorFactory;
 import org.apache.activemq.artemis.utils.critical.CriticalAnalyzer;
 import org.jboss.logging.Logger;
@@ -547,6 +548,17 @@ public class JournalStorageManager extends AbstractJournalStorageManager {
          largeMessage.copyHeadersAndProperties(message);
 
          largeMessage.setMessageID(id);
+
+
+         // Check durable large massage size before to allocate resources if it can't be stored
+         if (largeMessage.isDurable()) {
+            final long maxRecordSize = getMaxRecordSize();
+            final int messageEncodeSize = largeMessage.getEncodeSize();
+
+            if (messageEncodeSize > maxRecordSize) {
+               throw ActiveMQJournalBundle.BUNDLE.recordLargerThanStoreMax(messageEncodeSize, maxRecordSize);
+            }
+         }
 
          // We do this here to avoid a case where the replication gets a list without this file
          // to avoid a race
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpLargeMessageTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpLargeMessageTest.java
@@ -16,6 +16,7 @@
  */
 package org.apache.activemq.artemis.tests.integration.amqp;
 
+import java.io.IOException;
 import java.nio.charset.StandardCharsets;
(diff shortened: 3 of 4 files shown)


franz1981:
I'm not totally sure about it, but we could put a check even before in the process chain ie
```java
   private static Message asLargeMessage(Message message, StorageManager storageManager) throws Exception {
      ICoreMessage coreMessage = message.toCore();
      //in order to avoid headers
      if (message.isDurable()) {
         final int estimatedHeaderSize = coreMessage.getHeadersAndPropertiesEncodeSize();
         if (estimatedHeaderSize > storageManager.getMaxRecordSize()) {
            throw new ActiveMQLargeMessageException("The large message has a too big header size");
         }
      }
      LargeServerMessage lsm = storageManager.createLargeMessage(storageManager.generateID(), coreMessage);
      ActiveMQBuffer buffer = coreMessage.getReadOnlyBodyBuffer();
      final int readableBytes = buffer.readableBytes();
      lsm.addBytes(buffer);
      lsm.releaseResources();
      lsm.putLongProperty(Message.HDR_LARGE_BODY_SIZE, readableBytes);
      return lsm;
   }
```
Not sure if `isDurable` and `getHeadersAndPropertiesEncodeSize` are the correct methods , but the idea is to avoid that regardless the protocol used a large message won't be created/replicated at all if the journal is not able to persist it

michaelandrepearce:
Tbh this check should be occuring at the protocol layers. Before any processing. Fail fast.

clebertsuconic:
@michaelandrepearce / @franz1981 I think this is fine as is... the max-size would be checked within the same thread as the record is being generated instead of the writer executor.

Only thing I think we need is a logger for the exception and a proper exception. not part of this change but it would be nice since this is an expected exception now.

clebertsuconic:
@franz1981 you're suggesting adding the check on asLargeMessageSize? but the buffer size is only known at the journal. I would probably just move the check to the main thread instead of the executor. As @brusdev is doing here.

franz1981:
@clebertsuconic but the point here is that the large message shouldn't be created at all given that is the header that make it too large, not the body size.

brusdev:
@clebertsuconic I changed the exception with a more specific one and I checked that the exception was already logged by ProtonServerReceiverContext.actualDelivery.
@franz1981 I added a check to avoid to to allocate resources for a large message if it can't be stored, this check compares large message encoded size (without body) with the max record size (for JournalStorageManager is the minimum between file size and buffer size).
@michaelandrepearce I need to estimate header size at protocol layer to make a check and i think that this evaluations is too heavy.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -548,6 +549,17 @@ public LargeServerMessage createLargeMessage(final long id, final Message messag
 
          largeMessage.setMessageID(id);
 
+
+         // Check durable large massage size before to allocate resources if it can't be stored
+         if (largeMessage.isDurable()) {
+            final long maxRecordSize = getMaxRecordSize();
+            final int messageEncodeSize = largeMessage.getEncodeSize();
```
franz1981: You need to check the header and properties encoded size (that doesn't cost anything): this check will check size including header + trailer + body altogether.

brusdev: I see your point and I understand how to check the header and properties encoded size without costs from your example. In that point large message contains only header and properties of the source message and in addition it is precisely the record that is added to the journal. I think the last point is important because it let the check to work even if in the future other elements will be added to the large message.

franz1981: I see that you've put the checks before `validateFile` too, well done! :+1:

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpLargeMessageTest.java
```diff
@@ -265,6 +266,51 @@ public void testSendAMQPReceiveAMQPViaJMSBytes() throws Exception {
...
+         }
+         message.setApplicationProperty("str", buffer.toString());
+         message.setDurable(true);
+
+         try {
+            sender.send(message);
+            fail();
+         } catch (IOException e) {
+            Assert.assertTrue(e.getCause() instanceof JMSException);
+            Assert.assertTrue(e.getMessage().contains("larger than the store maxRecordSize"));
```
clebertsuconic: Please, assert for the Exception Code here.

on this case... contains("AMQ149005")

you will understand why based on my next comment

artemis-journal/src/main/java/org/apache/activemq/artemis/journal/ActiveMQJournalBundle.java
```diff
@@ -47,4 +47,7 @@
 
    @Message(id = 149004, value = "unable to open file")
    String unableToOpenFile();
+
+   @Message(id = 149005, value = "Record (size={0}) larger than the store maxRecordSize({1})", format = Message.Format.MESSAGE_FORMAT)
```
clebertsuconic: please, use a more user friendly message... something that would make sense to an user...

"Your message is storing too much data as part of properties (maxSize=maxSize). Store large data as part of the MessageBody"

Something like that.. I dunno.. this is just a suggestion.


this is actually the reason I asked you to change the test.. if you ever change the english grammar, translations.. whateverthe test will be broken. Test based on the code.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpLargeMessageTest.java
brusdev: I changed the check to test Exception Code.

artemis-journal/src/main/java/org/apache/activemq/artemis/journal/ActiveMQJournalBundle.java
brusdev: The new message of the exception is: "Message of {0} bytes is bigger than the max record size of {1} bytes. You should try to move large application properties to the message body."

PR #2695, opened 2019-06-05
ARTEMIS-2372 Filtering on Message Annotations

clebertsuconic:


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1263,7 +1263,10 @@ public Long getLongProperty(String key) throws ActiveMQPropertyConversionExcepti
 
    @Override
    public Object getObjectProperty(String key) {
-      if (key.equals(MessageUtil.TYPE_HEADER_NAME.toString())) {
+
+      if (key.startsWith(AMQPMessageSupport.JMS_AMQP_MESSAGE_ANNOTATION_PREFIX)) {
```
tabish121: This doesn't seem right at all, as the specification reserves all Annotations Symbol and UnsignedLong keys for its own use but the broker seems to be populating annotations that violate that.  Any client or other intermediary would be within its rights (technically required) to close the connection upon receiving a message with these broker specific message annotations as per the spec: 

←
3.2.10 Annotations -  The annotations type is a map where the keys are restricted to be of type symbol or of type ulong. All ulong keys, and all symbolic keys except those beginning with "x-" are reserved. Keys beginning with "x-opt-" MUST be ignored if not understood. On receiving an annotation key which is not understood, and which does not begin with "x-opt", the receiving AMQP container MUST detach the link with a not-implemented error.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpExpiredMessageTest.java
```diff
@@ -108,9 +109,16 @@ public void testExpiryThroughTTL() throws Exception {
       connection = addConnection(client.connect());
       session = connection.createSession();
 
-      AmqpReceiver receiverDLQ = session.createReceiver(getDeadLetterAddress());
+      AmqpReceiver receiverDLQ = session.createReceiver(getDeadLetterAddress(), "_AMQ_ORIG_ADDRESS='" + getQueueName() + "'");
```
tabish121: Instead of disturbing an existing test to validate new functionality you should really be writing new tests that cover the feature in depth with both positive and negative validation tests to ensure that future changes don't break and that the feature works as intended.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
clebertsuconic: @tabish121 what UnsignedLong keys? This is getObjectProperty().

This is about filtering. The issue is when moving a message to DLQ, the copy will apply a message annotation. 

Previously with AMQ5, users would be able to filter by JMS_AMQP_MA_ORIG_QUEUE. This is just bringing that possibility.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpExpiredMessageTest.java
clebertsuconic: You mean, copy & paste the test, and do the same? What for? it's the same test being done.

that will just add more time (a few seconds) to the testsuite for no reason.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
clebertsuconic: @tabish121 getObjectProperty is only used on the server's side for filtering. This is just leverating what's already implemented through setAnnotation, and Message.referenceOriginalMessage

After the message is copied to DLQ or ExpiryQueue, this will allow what's in there to be filtered.

tabish121: You are putting illegal message annotations into the message from what I can tell, and the test seems to validate that.  The message that is received has annotations which are of the form ({_AMQ_ORIG_QUEUE=testExpiryThroughTTL, _AMQ_ORIG_ADDRESS=testExpiryThroughTTL, _AMQ_ORIG_MESSAGE_ID=69, _AMQ_ACTUAL_EXPIRY=1559769271365}) which is illegal according the spec section I posted and a receiver is technically required to close the connection if it sees them.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpExpiredMessageTest.java
tabish121: I mean write a real set of tests that validate that it applied to a given property and not others etc.  It really isn't that hard and given how long the broker tests take I'm pretty sure nobody will break out the torches and pitchforks.  Although given what the broker is doing is creating illegal message content maybe the tests should check for that not happening instead.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
clebertsuconic: @tabish121 these message annotations are already there and it is not part of this change.

These annotations are already taking place since 2017. If these are not valid, this should be the subject of further changes.. and nothing to do with this specific change.

clebertsuconic: @tabish121 when moving to DLQ or Expiry, the message is copied..

we could (perhaps should) instead of using annotations here (I agree it's wrong), use a regular property.

After all when moving the message to DLQ or Expiry we are moving a copy of the message, and not the original message. (you could see it that way.. although that's one line of argument)


Or perhaps we could fix it properly, and use proper annotations. but that will break compatibility with previous AMQ5, as the message is always converted from OpenWire into AMQP. (AMQ5 never preserves the AMQP message on that sense)

tabish121: > we could (perhaps should) instead of using annotations here (I agree it's wrong), use a regular property.

You could use an annotation for this if you wanted, although one that wasn't illegal according to the spec would be my suggestion (I know crazy spec purists).  

> After all when moving the message to DLQ or Expiry we are moving a copy of the message, and not the original message. (you could see it that way.. although that's one line of argument)

Whether it is copied or not doesn't really come into play here as the annotation being used is illegal.  By supporting this selector feature you are just adding a finer point to the thing that the broker is doing that it really shouldn't be.  The broker could for instance store the value in an annotation named 'x-opt-original-destination' or some such and that would be legal.  It would however not be accessible to a client that doesn't have direct access to annotations but only to application properties.

> Or perhaps we could fix it properly, and use proper annotations. but that will break compatibility with previous AMQ5, as the message is always converted from OpenWire into AMQP. (AMQ5 never preserves the AMQP message on that sense)

I'm unclear how ActiveMQ 5 or openwire come into this as 5.x doesn't add the original destination set on expired and DLQ'd messages when it converts them to AMQP to the best of my knowledge.

clebertsuconic: I was actually asked to do this because AMQ5 is supporting this, according to my user.

clebertsuconic: @tabish121 what I meant in regard to the copy was to use regular application properties on the copy, rather than an annotation.

tabish121: I can only find one sneaky way to apply a selector that would allow you to match on the original destination when a message is in the DLQ when using an AMQP consumer with 5.x.  If the user is talking about an OpenWire consumer then that is a different thing entirely and you cannot just say the two are equivalent.  

Doing something more akin to what 5.x does is probably actually an easier to explain solution than the way you've done here as it doesn't involve knowledge of message annotations or even protocol for the most part (if tweaked a bit).  

In 5.x as it stands now you can create a consumer using Qpid JMS that is subscribed to the DLQ and provide a selector along the lines of "JMSDestination='queue://myOriginalQueue'" and that would match if the message had expired off the queue "myOriginalQueue".  This works because the property expression applied via the selector first checks if the message has an original destination stored in it and uses that instead of the current destination to match on the selector value.  If Artemis was to do something similarly clever then you wouldn't need to require the user to provide some arbitrary value like _AMQ_SOMETHING_OR_OTHER their code remains vendor and protocol neutral.

clebertsuconic: Ok, i will use extraProperties and do something about that...

Just so you know this PR was about exposing the properties.. the use of messageAnnotations for this wasn't introduced here.


But i will fix it as part of this PR.


i will close it for now as I won't have time to finish it before 1 week. I may reopen it later.

clebertsuconic:
This PR has been reworked to not change any semantics.. other than looking at the annotations on filter.

gemmellr:
I dont think this should have been added before https://issues.apache.org/jira/browse/ARTEMIS-2740 is addressed. I know that JIRA is new, but I see the issue was raised in comments here when this PR was raised last year, and the PR closed, but now reopened/updated and added anyway - with the tests still using annotations that are reserved. It seems unwise to add a new functionality that relies on/encourages this behaviour, given fixing that would then necessarily break any uses of this new functionality.

It also seems that now any use of filters where the key value isnt found in application properties is going to decode every annotation now. Unlikely to be great for performance, and especially not nice for folks who aren't even looking to filter on any annotations, which is likely to be most values for most people, and certainly every existing user up to this point given it wasn't even possible before.

The filter extensions spec in the AMQP TC (https://www.oasis-open.org/committees/document.php?document_id=66227) is currently proposing using "m.<annotation-key>" as a means of requesting filtering on the message-annotations map within an SQL based filter string. It might be worth aligning to that, as it would at least allow avoiding decoding annotations when folks have not asked for them to be filtered.

clebertsuconic:
Before this the Jira had added an invalid annotation.   Recently there is no annotation changes and it’s just exposing the Jira. 



I understand your concern and will address before the release.

clebertsuconic:
```
"It also seems that now any use of filters where the key value isnt found in application properties is going to decode every annotation now. Unlikely to be great for performance, and especially not nice for folks who aren't even looking to filter on any annotations, which is likely to be most values for most people, and certainly every existing user up to this point given it wasn't even possible before."
```

@gemmellr: if you use expressions, that's an expected cost. no issues in there.

gemmellr:
I disagree its an expected cost. Existing uses may now do more work and go slower than before, for no gain. Plus if they haven't specifically asked to filter on an annotation, which again no existing use could be doing, then it shouldnt really be looking at them anyway, cost or not. It may even lead to incorrect/unexpected behaviour by doing so all of a sudden.

clebertsuconic:
You may disagree.. but it's working as designed.. if users ever use Expressions, we are expected to hit a decode on the message.. nothing to be changed there.

clebertsuconic:
I got users asking for the feature, and you are basically opposing the feature is implemented.

It's such a simple addition. I don't see a reason to remove a filter, or even add a toggle to enable it or not.

gemmellr:
I didn't suggest not supporting filtering annotations, or adding a toggle to enable it or not.

I do suggest:
- Not causing performance or correctness changes for existing users who are definitely not using the functionality, by having it decode and inspect things all of a sudden that it shouldnt be looking (as it hasnt been asked to, and it wasnt possible before).
- Not encouraging [further] violating the core AMQP spec and thus preventing interop with peers that might enforce what it says, when there are simple routes not to.
- Aligning with the "m." prefix the AMQP TC draft filter spec currently suggests for this functionality.
- In concert for these avoiding later causing breakage in the usage of these new users you aim satisfy, should the broker ever decide to actually adhere the specs later rather than doing it now before the functionality even exists.

clebertsuconic:
it's not possible to implement filteres over annotations without parsing the annotations.. simply not possible!

if you use a filter on a property that is not being found anywhere.. the server will try to find it on annotations.

users usually don't set filters with empty properties.. so your concern here is really really a corner and not practical case IMO.


What I'm going to do is to use x-opt in any property on the server.. and when parsing for a property, it will then look for the property on annotations.. if still not found it will convert the _AMQ_ into a x-opt property. 


I will send a PR.. but it's not possible to do this with 0 computational cost for someone using filters.

tabish121:
I don't think Robbie ever implied that the application of a message annotations filter should or could be a zero computational cost operation.  What is being argued here (and rightly so) is that filtering on an AMQP message annotation should be an opt in filtering operation and not a by product of the broker walking into the message annotations section any time a filter doesn't hit on an application property in the AMQP message when it applies the filters. 

Currently with Artemis users can't filter on message annotations in AMQP messages, if you want to give them that option then you should implement it following the recommendations from the AMQP Core TC draft specification that the filter should be an explicit "m." prefix on the message annotation property the user wishes to filter upon (m.x-opt-original-destination=foo) etc.

clebertsuconic:
@tabish121 you could argue the same over application properties, and make users to use properties.


Users are not going to set a property filter into something that's already null... 

This is not amqp filter, this is a broker functionality over JMS. and we don't only have AMQP as the protocol. The same filter would have to work across other protocols in case a client changed the protocol.

clebertsuconic:
@gemmellr / @tabish121 I really don't want to introduce m. p. or any other properties on filters.. as this is a broker funcationality that's independent from the protocols.


I actually wasn't part of the discussion. I think it's a bit strong position from the protocol implementation dictating how a broker supposedly protocool agnostic should implement filters.

tabish121:
I'm going to -1 this commit unit a proper solution is found then as this is a rather serious technical issue that needs to be worked out before any future release can provide this sort of functionality.

clebertsuconic:
I can check for prefixes on a property if it matches the standard for message annotations..

But I can't implement m.,p. as this would break current functionality. (I actually the AMQP group should consider against such prefixes imposition).

gemmellr:
> Users are not going to set a property filter into something that's already null...

They can though. I'm pretty sure I have actually even seen it done previously. There are also various operations/behaviours that check/depend on null values explicitly in particular ways.

> This is not amqp filter, this is a broker functionality over JMS.

Specifically about adding filtering AMQP message annotations by AMQP consumers that sent a selector using an AMQP filter though, so it seems relevant.

> I think it's a bit strong position from the protocol specification dictating how a broker supposedly protocool agnostic should implement filters.

There is no dictation or imposition from the AMQP spec/TC that Artemis must use "m.<key>". Its a completely new and entirely optional SQL-like filter (and so not actually the filter being used by the broker or JMS client currently, though no reason the syntax cant be supported by the broker) that would allow for AMQP clients (which are mostly not JMS clients) to be able to filter the AMQP messages sections in a consistent way against any AMQP server that implements the filter. Nothing says Artemis needs to support it. It is a broker-agnostic filter, but clearly not necessarily protocol-agnostic because since it is an AMQP filter aimed at AMQP consumers filtering on sections of AMQP messages. (Aside, not really important : there are also other non-SQL-like filters for achieving similar or more specific results. Since you mentioned it, there is also a "a." prefix for the application-properties section for the SQL-like filter, though that one was made optional since its the traditional section to filter on). 

Nothing stops Artemis having other filters (as it already does) and using other syntax(es) to achieve the result of filtering AMQP message annotations of course. I just strongly think whatever it is shouldn't encourage AMQP messaging use violating the AMQP specs reserved annotations, or result in sudden unwittingly filtering on things (i.e. annotations) the user hasnt expressly asked to filter given that could actually change/break the behaviour of their systems. More so given its likely rarely going to be the case that the filter identifier is actually an annotation, certainly currently since its never been possible. I was merely suggesting aligning to an existing proposal on the specific subject matter.

michaelandrepearce:
I have to agree with Cleberts earlier comment amqp 1.0 was not meant to care if a sender or receiver is a broker, its agnostic. As such it really shouldnt be enforcing any broker specifics, by doing so it no longer is an agnostic protocol.

That said and it seems we have two camps in general here, those that are wanting strict conformist and thats important for their use case and those who need broker features that technically break strict conformance but for some users actually have business value for them such this filter ability in this pr and in other prs where sometimes to have a broker feature requires mutating (adding or amending) some amqp sections. 

Could i suggest we have a broker flag in config that is called maybe "amqpConformance" where its set to STRICT or RELAXED, and going forwards when hit these issue simply use the flag to still allow it should end broker users which to enable the RELAXED operation mode.

gemmellr:
> I have to agree with Cleberts earlier comment amqp 1.0 was not meant to care if a sender or receiver is a broker, its agnostic. As such it really shouldnt be enforcing any broker specifics, by doing so it no longer is an agnostic protocol.

There is no caring whether its a broker or anything else, so its not being broker-specific. It does define that certain parts of its standard message format are considered immutable such that recipients receive what senders actually sent in those parts, which they could then verify if desired. It also defines multiple other sections that are mutable should things want their messages mutated by intermediaries for reasons.

Being a component/platform agnostic protocol doesnt necessarily mean its an instant universal drop in replacement for all existing different protocols and behaviours. A particular mechanism underneath may differ from other protocols, but is known, i.e everything has a shared knowledge of how certain things are defined such as the format of the messages and armed with this anything/anyone can implement the same knowing they can then interoperate with other components doing the same.  Going outwith such existing definitions can obviously make that interop bit a problem.

> Could i suggest we have a broker flag in config that is called maybe "amqpConformance" where its set to STRICT or RELAXED, and going forwards when hit these issue simply use the flag to still allow it should end broker users wish to enable the RELAXED operation mode thats their prerogative.

I think having a single global flag of this kind would be a very bad idea, far worse than having any individual one. Having any is annoying, but if they do exist at all they should at least be very specific to a particular case, minimising their effect. Using any such behaviour shouldnt require changing any/all the other behaviours too 'just because'.

clebertsuconic:
@gemmellr / @tabish121 it's not even possible to use m. or p. as a name, since those are not considered JMS expressions, and qpid-jms will throw errors and not complete operations.

gemmellr:
You can quote it to get round that. Though that syntax is intended for that filters string, which is not necessarily what the application sets, but what the client might.

Again, Artemis could even have its own syntax for denoting filtering annotations. Maybe there is precedence for one in the codebase already.

A JMS application might not often be expected to filter on AMQP annotations either.

clebertsuconic:
@gemmellr I will only parse annotations when the user is asking for anything starting with "x-opt" on the filter.

The user will be able to use "hyphenated_props:x-opt-ORIG-ADDRESS='my-address'" as an example of a filter.

clebertsuconic:
@gemmellr currently the qpid-jms client will not let me parse m. ... it's actually against JMS Specs.


i can't use quotes, as the parser will of course use it as a quote.


The only possibility I have so far that would work is use a dolar sign... m$

or do what I told you.. only parse annotations when including x-opt...

gemmellr:
When I said quoting I meant "\\"quoted.key\\"=value". I works since the regular single quotes are something else.

Something like that or the "hyphenated_props:" thing above (not familiar with that against-JMS-spec workaround, Qpid JMS probably wont like that) would still be necessary for anything with x-opt- in it given It'd otherwise be saying x minus opt minus something.

Saying it is only possibly an annotation if it starts x-opt is a bit better than the earlier case (of saying any key could be an annotation, but only if it wasn't an application property). Though it still basically has the same problem of not specifically trying to denote you want to filter an annotation, is just a bit less likely to clash with an application property (particular for JMS clients, obviously).

clebertsuconic:
PR #3129 is addressing the concerns here.

i couldn't use quotes on the key names.. but I was able to use . if using hyphanated_props: (at least for now). 

Please.. transfer any discussion now to the new PR, but everything here should been addressed.

PR #2696, opened 2019-06-07
ARTEMIS-2374 JournalStorageManager::addBytesToLargeMessage leaks ByteBuffer

franz1981:


artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManagerTest.java
```diff
@@ -70,6 +77,11 @@
    private static ExecutorService ioExecutor;
    private static ExecutorService testExecutor;
 
+   @Before
+   public void init() {
+      disableCheckThread();
```
clebertsuconic: please, don't do this.

Can you rather fix the thread leakage by shutting down the pool, which was the issue on this test?

franz1981: there is no leak: I'm avoiding to re-create 3/4 threads for each of the test method...

PR #2697, opened 2019-06-10
ARTEMIS-2376 Remove guava from artemis-common feature

graben:


graben:
Hi @clebertsuconic,
may you have a look on this?
Thx Benjamin

jbonofre:
I can take a look.

graben:
@jbonofre Merci :)

clebertsuconic:
@jbonofre  / @graben  this is an area that my co-worker, @brusdev  recently touched

brusdev:
@graben why do you remove guava bundle dependency from artemis-common?

brusdev:
@graben why do you remove guava bundle dependency from artemis-common?

graben:
@brusdev : Because it isn't used/needed anymore and therefore reduces maintenance in custom Karaf distribution if several guava versions are bundled. Actual used version 18 is quite old and has known security issues!

jbonofre:
If Guava is not used, it makes sense to remove it (I will double check). Furthermore, Guava 18 is old and contains CVE issue.

clebertsuconic:
I can't merge it right away as my workspace is messed up.. but I will do it shortly.

PR #2698, opened 2019-06-10
ARTEMIS-2367 Adding rule to avoid files leaking on the main folder

clebertsuconic:
We had to fix a few cases where data folder was being created outside of the ./target output.
This is just to avoid re-ocurrence of that.


clebertsuconic:
lots of failures (some tests are still leaving data behind). 

Let me fix them as part of this PR.

clebertsuconic:
this is now ready!

clebertsuconic:
@jbertram since you worked on this task before, can you merge this one please?

jbertram:
Nice one!

PR #2703, opened 2019-06-15
ARTEMIS-2380 Fix delivering message in the case of consume close

wy96f:
When consumer is closed and transaction is not committed or rollbacked, the msg acked is not showed by delivering message but counted by delivering count.

We add closed consumer if any message is acked but not committed or rollbacked, and remove closed consumer if transaction is finished. In this way, delivering messages can be properly retained.

During calling hasCosnumer and getConsumerId, there is some race condition that emptyConsumerId is called causing following getConsumerId throws IllegalStateException. Also optimize it without extra effort.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -80,6 +80,7 @@ import org.apache.activemq.artemis.core.server.QueueFactory;
 import org.apache.activemq.artemis.core.server.RoutingContext;
 import org.apache.activemq.artemis.core.server.ScheduledDeliveryHandler;
 import org.apache.activemq.artemis.core.server.ServerConsumer;
+import org.apache.activemq.artemis.core.server.ServerSession;
 import org.apache.activemq.artemis.core.server.cluster.RemoteQueueBinding;
 import org.apache.activemq.artemis.core.server.cluster.impl.Redistributor;
 import org.apache.activemq.artemis.core.server.management.ManagementService;
@@ -101,6 +102,7 @@ import org.apache.activemq.artemis.utils.Env;
 import org.apache.activemq.artemis.utils.ReferenceCounter;
 import org.apache.activemq.artemis.utils.ReusableLatch;
 import org.apache.activemq.artemis.utils.actors.ArtemisExecutor;
+import org.apache.activemq.artemis.utils.collections.ConcurrentHashSet;
 import org.apache.activemq.artemis.utils.collections.LinkedListIterator;
 import org.apache.activemq.artemis.utils.collections.PriorityLinkedList;
 import org.apache.activemq.artemis.utils.collections.PriorityLinkedListImpl;
@@ -318,6 +320,8 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
     * to guarantee ordering will be always be correct
     */
    private final Object directDeliveryGuard = new Object();
+
+   private final ConcurrentHashSet<String> lingerSessionIds = new ConcurrentHashSet<>();
 
    public String debug() {
       StringWriter str = new StringWriter();
@@ -1259,6 +1263,16 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    }
 
    @Override
+   public void addLingerSession(String sessionId) {
+      lingerSessionIds.add(sessionId);
+   }
+
+   @Override
+   public void removeLingerSession(String sessionId) {
+      lingerSessionIds.remove(sessionId);
+   }
+
+   @Override
    public void removeConsumer(final Consumer consumer) {
 
       enterCritical(CRITICAL_CONSUMER);
@@ -1583,6 +1597,15 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
             mapReturn.put(holder.consumer.toManagementString(), msgs);
          }
       }
+
+      for (String lingerSessionId : lingerSessionIds) {
+         ServerSession serverSession = server.getSessionByID(lingerSessionId);
+         List<MessageReference> refs = serverSession == null ? null : serverSession.getInTxLingerMessages();
+         if (refs != null && !refs.isEmpty()) {
+            mapReturn.put(serverSession.toManagementString(), refs);
+         }
+      }
+
       return mapReturn;
    }
 
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
@@ -52,6 +52,8 @@ public class RefsOperation extends TransactionOperationAbstract {
     */
    protected boolean ignoreRedeliveryCheck = false;
 
+   private String lingerSessionId = null;
+
    public RefsOperation(Queue queue, AckReason reason, StorageManager storageManager) {
       this.queue = queue;
       this.reason = reason;
@@ -97,6 +99,8 @@ public class RefsOperation extends TransactionOperationAbstract {
       List<MessageReference> ackedRefs = new ArrayList<>();
 
       for (MessageReference ref : refsToAck) {
+         clearLingerRef(ref);
+
          ref.emptyConsumerID();
 
          if (logger.isTraceEnabled()) {
@@ -175,8 +179,10 @@ public class RefsOperation extends TransactionOperationAbstract {
    @Override
    public void afterCommit(final Transaction tx) {
       for (MessageReference ref : refsToAck) {
+         clearLingerRef(ref);
+
          synchronized (ref.getQueue()) {
-            queue.postAcknowledge(ref, reason);
+            ref.getQueue().postAcknowledge(ref, reason);
          }
       }
 
@@ -187,6 +193,12 @@ public class RefsOperation extends TransactionOperationAbstract {
                decrementRefCount(refmsg);
             }
          }
+      }
+   }
+
+   private void clearLingerRef(MessageReference ref) {
+      if (!ref.hasConsumerId() && lingerSessionId != null) {
+         ref.getQueue().removeLingerSession(lingerSessionId);
       }
    }
 
@@ -228,4 +240,18 @@ public class RefsOperation extends TransactionOperationAbstract {
       return refsToAck;
    }
 
+   public synchronized List<MessageReference> getLingerMessages() {
+      List<MessageReference> list = new LinkedList<>();
+      for (MessageReference ref : refsToAck) {
+         if (!ref.hasConsumerId() && lingerSessionId != null) {
+            list.add(ref);
+         }
+      }
+
+      return list;
+   }
+
+   public void setLingerSession(String lingerSessionId) {
+      this.lingerSessionId = lingerSessionId;
+   }
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Queue.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Queue.java
@@ -129,6 +129,10 @@ public interface Queue extends Bindable,CriticalComponent {
 
    void addConsumer(Consumer consumer) throws Exception;
 
+   void addLingerSession(String sessionId);
+
+   void removeLingerSession(String sessionId);
+
    void removeConsumer(Consumer consumer);
 
    int getConsumerCount();
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ServerSession.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ServerSession.java
@@ -429,6 +429,10 @@ public interface ServerSession extends SecurityAuth {
 
    List<MessageReference> getInTXMessagesForConsumer(long consumerId);
 
+   List<MessageReference> getInTxLingerMessages();
+
+   void addLingerConsumer(ServerConsumer consumer);
+
    String getValidatedUser();
 
    SimpleString getMatchingQueue(SimpleString address, RoutingType routingType) throws Exception;
@@ -490,4 +494,6 @@ public interface ServerSession extends SecurityAuth {
    int getProducerCount();
 
    int getDefaultConsumerWindowSize(SimpleString address);
+
+   String toManagementString();
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
@@ -18,8 +18,8 @@ package org.apache.activemq.artemis.core.server.impl;
 
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
-import java.util.ArrayList;
 import java.util.ArrayDeque;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedList;
@@ -571,6 +571,8 @@ public class ServerConsumerImpl implements ServerConsumer, ReadyListener {
 
       tx.rollback();
 
+      addLingerRefs();
+
       if (!browseOnly) {
          TypedProperties props = new TypedProperties();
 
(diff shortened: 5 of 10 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1175,6 +1179,17 @@ public void addConsumer(final Consumer consumer) throws Exception {
 
    }
 
+   @Override
+   public void addLingerRef(Consumer consumer, MessageReference lingerRef) throws Exception {
+      lingerRef.emptyConsumerID();
+      lingerRefMap.put(lingerRef, consumer);
```
franz1981: Beware putting `MessageReference` into a `Map`: `MessageReferenceImpl::hashCode` relies on `getMessage().hashCode()` and `CoreMessage::hashCode` will use the native hash code (like `java.lang.System#identityHashCode`).
It is known that it would break biased locking optimization on it (see https://blogs.oracle.com/dave/biased-locking-in-hotspot for more detail): given that `CoreMessage` heavily relies on `synchronized` operation, I would suggest to measure the perf impact of it first :+1:

michaelandrepearce: On that note, need to ensure order is preserved hashmap looses any ordering

wy96f: @franz1981 
> Beware putting `MessageReference` into a `Map`: `MessageReferenceImpl::hashCode` relies on `getMessage().hashCode()` and `CoreMessage::hashCode` will use the native hash code (like `java.lang.System#identityHashCode`).
> It is known that it would break biased locking optimization on it (see https://blogs.oracle.com/dave/biased-locking-in-hotspot for more detail): given that `CoreMessage` heavily relies on `synchronized` operation, I would suggest to measure the perf impact of it first
Thanks, franz. I'll have a look and measure the perf.

> On that note, need to ensure order is preserved hashmap looses any ordering
Good point. BTW, as i read from JMS spec, session is not thread safe, so closing consumer and committing session from which consumer is created can't be from multiple threads, correct? If not thread safe, the issue referred in this PR should be happening rarely? Just not sure whether it's deserved to fix this trivial problem by adding some tricky and fragile code.

michaelandrepearce: Remember Artemis is mutliprotocol. E.g. it supports more than just JMS

michaelandrepearce: Original order of message when rollback occurs needs to preserved so that theyre replayed the same order again

clebertsuconic:
This needs to be rebased... 
Can this wait until after 2.10 is tagged?

wy96f:
> This needs to be rebased...
> Can this wait until after 2.10 is tagged?

I'm still working on it, don't merge :)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelandrepearce: Can use the new sorted methods instead of head.

wy96f:
Added synchronize block to avoid race condition between consumer::close and session::commit/rollback, also keeping message ordering.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -319,6 +321,8 @@
     */
    private final Object directDeliveryGuard = new Object();
 
+   private final CopyOnWriteArrayList<Pair<Consumer, List<MessageReference>>> lingerConsumerRefs = new CopyOnWriteArrayList<>();
```
michaelandrepearce: Is this needed surely we can hold the list within the consumer removing need for this altogeather

michaelandrepearce:
No need for sync. Theres a new addsorted method that will put back in order so if you call that instead of head problem gone

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/plugin/impl/NotificationActiveMQServerPlugin.java
```diff
@@ -138,7 +138,10 @@ public void messageExpired(MessageReference message,
                props.putSimpleStringProperty(ManagementHelper.HDR_ROUTING_NAME, message.getQueue().getName());
                props.putLongProperty(ManagementHelper.HDR_MESSAGE_ID, message.getMessageID());
                if (message.hasConsumerId()) {
-                  props.putLongProperty(ManagementHelper.HDR_CONSUMER_NAME, message.getConsumerId());
+                  long consumerId = message.getConsumerId();
+                  if (consumerId >= 0) {
```
michaelandrepearce: Redudent check message.hasConsumerId which does same check 2 lines above

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/MessageReferenceImpl.java
```diff
@@ -70,9 +70,7 @@ public int compare(MessageReference o1, MessageReference o2) {
 
    private final Queue queue;
 
-   private long consumerID;
-
-   private boolean hasConsumerID = false;
+   private long consumerID = -1L;
```
michaelandrepearce: Are we 100% sure consumerid cannot ever be -1

michaelandrepearce: Im not sure how safe this optimisation is especially if someone implements custom id generators

michaelandrepearce: BatchingIDGenerator looks as if it can loop round into - numbers, this is used by mqtt publish manager via storage handler

wy96f:
> No need for sync. Theres a new addsorted method that will put back in order so if you call that instead of head problem gone

The sync is to make sure consumer not added in QueueImpl while transaction is committing or rollbacking. Without sync block, there is chance that consumer would be added in QueueImpl and never be removed bcs afterCommit/afterRollback has been called before it.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/MessageReferenceImpl.java
wy96f: I'm not worried about our own generators since they all start from zero and it would take long time to loop round especially used for consumer id. Furthermore, number wrapping should not be allowed bcs this would resulting in the consumer id clashing. But for custom id generators, it's not safe bcs they maybe start from some negative numbers.

The purpose of this optimization is to prevent RefsOperation::getListOnConsumer from throwing exception. There is some chance that ref.emptyConsumerId() will be called Between ref.hasConsumerId() and ref.getConsumerId() == consumerID, which would throw IllegalStateException. I don't feel strong about this optimization since getListOnConsumer is triggered only by management service which would not affect core logic even if race conditions happen. I can remove it and fix it later when better idea emerges.

michaelandrepearce: Simpler change to address that is to set the has field after the id field. If we are to ensure no negative then we should ensure all if genetators down wrap into negative. Currently only the simple one enforces that. Others like the batcher do not

michaelandrepearce:
This goes back to note the linger messages can be tracked at consumer level not queue though.

wy96f:
> This goes back to note the linger messages can be tracked at consumer level or even the server session. not queue though. E.g. if the issue is consumer is closed but session not committed, then it seems more logical that its the sessions responsibility to track and to have the logic within the session rollback/commit logic.
> Having it at the queue level i can just see is prone to leaks like we have had other places which by putting the logic and ownership into the correct place resolved the leak and also cleaned up.

Yes, we can track linger messages at consumer or session level. Linger consumer should also be tracked in queue, otherwise i'm not sure how to get linger messages without going through consumer while calling QueueImpl::getDeliveringMessages. Moreover in our case, we need to know which consumer is closed with these linger messages by which we can locate ip address where problem occurs.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/MessageReferenceImpl.java
wy96f: For current code, when hasConsumerId returns true, thread1 will call getConsumerId. At this moment if another thread calls emptyConsumerId just before thread1 calling, thread1 would throw exception. Setting the has field after the id field can't address this.

michaelandrepearce:
You may have multiple consumers on a single ip... relying on ip doesnt sound safe way to link to a single consumer

michaelandrepearce:
The more i look at this i think you should track these in the session. Not consumer, else we are not letting go of the closed consumer. Once consumer is closed it should be able to gc'd should be treated as no longer existing. How to link it would be by tracking uncommitted messages in session and tracking the session in the queue, this way also if multiple consumers in a session are closed then only tracking one thing. And is more closely mapping the actual behaviour

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/MessageReferenceImpl.java
michaelandrepearce: Would a simpler option be simply make this then Long object?

wy96f:
> You may have multiple consumers on a single ip... relying on ip doesnt sound safe way to link to a single consumer

Once we know ip, we can let users take thread dumps of their process. Most likely something hangs in business logic, resulting in session not committed, this way they can quickly fix the issue.

wy96f:
> The more i look at this i think you should track these in the session. Not consumer, else we are not letting go of the closed consumer. Once consumer is closed it should be able to gc'd should be treated as no longer existing. How to link it would be by tracking uncommitted messages in session and tracking the session in the queue, this way also if multiple consumers in a session are closed then only tracking one thing. And is more closely mapping the actual behaviour

Thanks, will try.

wy96f:
Refactored the code and reverted consumerId change

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
```diff
@@ -175,6 +179,8 @@ protected void rollbackRedelivery(Transaction tx, MessageReference ref, long tim
    @Override
    public void afterCommit(final Transaction tx) {
       for (MessageReference ref : refsToAck) {
+         clearLingerRef(ref);
```
michaelandrepearce: You dont need to do this inside the iterator. You have the operation queue and only need be done one

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
```diff
@@ -190,6 +196,12 @@ public void afterCommit(final Transaction tx) {
       }
    }
 
+   private void clearLingerRef(MessageReference ref) {
```
michaelandrepearce: No need for message ref. Just use the refop queue

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -319,6 +321,8 @@
     */
    private final Object directDeliveryGuard = new Object();
 
+   private final ConcurrentHashSet<String> lingerSessionIds = new ConcurrentHashSet<>();
```
michaelandrepearce: How are sure we dont leak here if sessions are closed

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1567,6 +1581,10 @@ public long getDurableScheduledSize() {
       return scheduledDeliveryHandler.getScheduledReferences();
    }
 
+   private String getSessionManagementString(ServerSession serverSession) {
```
michaelandrepearce: ?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
wy96f: Refs maybe from different queues, E.g. one session creates two consumers, one subscribing queue1, the other subscribing queue2.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
wy96f: Closing ServerSession will roll back transaction. And I add synchronized transaction block to avoid adding session ids when transaction is rollbacking/committing. This is what ServerSessionImpl::addLingerConsumer does(synchronizing access by this and afterCommit/afterRollback).

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
wy96f: I want to know the client ip that closes consumer but not committing session, so I added server session info to return.

michaelandrepearce: It forces string concatenation here though un-neededly.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
michaelandrepearce: The ref operation holds the queue its for.

wy96f: I think it just hold the queue which the first acked message belongs to. Following acked messages in this transaction maybe from other queues since a session might create consumers subscribing to different queues.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
wy96f: I copied code from Consumer::toManagementString. Do you mean using StringBuilder? It sounds not a hot path here.

michaelandrepearce: More the issue is that its used as a key later in code. Key should really be based off a unique id. Not some formatted string

wy96f: Session name is a unique id and is used as a key in ActiveMQServerImpl::sessions map.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
michaelandrepearce: Then there would be seperate ref ops in the tx

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelandrepearce: We should just be using that, not some new formatted key

michaelandrepearce: Formatting should be either in the control or owned by session so that its uniform anywhere its used

michaelandrepearce: See consumer.toManagementString()

michaelandrepearce: Checking some other code that displays management string representation exists this should follow suit so its uniform.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
wy96f: There is only one ref ops for each tx. I've added a test to validate it.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelandrepearce: Understood

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
michaelandrepearce: I stand corrected :) nice test :)

michaelandrepearce:
+1 lgtm. Great work on this. I would merge this, but as youre a new committer :) do you want to try? This way also check everything is setup for you going forwards

PR #2708, opened 2019-06-18
ARTEMIS-2386 - use internal session for management queues for QOS2

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-2386


franz1981:
The failure doesn't seem related to this PR , I'm gonna merge it!

PR #2710, opened 2019-06-18
ARTEMIS-1825 Live-backup topology not correctly displayed on console

gaohoward:
Second commit. The topology shown on backup's console is not yet
correct. This fixes the problem.


michalxo:
This looks good to me. I have tested on 3 pairs in HA configuration with failback "on" on slaves (so master became master again). This works fine with 2 minor issues:
1) I have seen "undefined" slave object on 1 slave node (after killing 2nd master) (bringing all masters up again removed this "undefined" node)
2) Killing of slaves does not propagate into topology (?), so after killing all slaves, they are still "visible" as active slaves in diagram, which is imo a bug.

Else, this PR addresses original issue - diagram not displaying topology properly. Slaves are in red, masters in green as expected even after promotion/demotion.

gaohoward:
@michalxo Many thanks for testing it out.

PR #2712, opened 2019-06-18
Allow setting JMS Properties when sending messages via REST.

twyszomirski:
Allows setting JMS properties on a message posted via REST interface. The property is transferred as a HTTP header having name prefixed with "jms.property.". On successful arrival on topic/queue resource the prefix is stripped and property set as JMS property. This allows filtering on custom fields and using features requiring JMS properties (i.e JMSXGroupID).


michaelandrepearce:
The broker is agnostic of JMS. We specifically stripping JMS specifics this is so the broker can be protocol agnostic. Maybe implement this in a way to set anyheader

twyszomirski:
@michaelandrepearce Ok I get the "agnostic" argument, so assuming I drop or replace the word "jms" (for example instead "jms.property." I use "message property.") are we fine? Or should I rewrite the whole feature? You already use http headers to pass the content-type and I'm reluctant to change that, BUT I need to somehow put custom values into message properties to use them in selectors (also I need to be able to set JMSXGroupID in order to use artemis grouping feature).

PR #2714, opened 2019-06-18
ARTEMIS-2385 Log header for rejecting message with too large header

brusdev:
Use warning level to add in the log the ID and debug level to add in the
log the message header of rejecting message with too large header.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
@@ -556,6 +556,12 @@ public class JournalStorageManager extends AbstractJournalStorageManager {
             final int messageEncodeSize = largeMessage.getEncodeSize();
 
             if (messageEncodeSize > maxRecordSize) {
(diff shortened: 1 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -556,6 +556,12 @@ public LargeServerMessage createLargeMessage(final long id, final Message messag
             final int messageEncodeSize = largeMessage.getEncodeSize();
 
             if (messageEncodeSize > maxRecordSize) {
+               if (logger.isDebugEnabled()) {
+                  logger.debug("Message header too large for " + largeMessage);
+               } else {
+                  logger.warn("Message with id " + largeMessage.getMessageID() + " has a header too large. Set log level to debug to log the message header.");
```
clebertsuconic: two things:

- the logger.warn should always be called, no matter what.
- Can you use a Logger with an ID here?

jbertram:
@brusdev, can you rebase this? There's a conflict since merging another PR just now.

PR #2716, opened 2019-06-19
Allow setting message properties when sending messages via REST.

twyszomirski:
Allows setting message properties on a message posted via REST interface. The property is transferred as a HTTP header having name prefixed with "mesage.property.". On successful arrival on address resource the prefix is stripped and property set as message property. This allows filtering on custom fields and using other broker features requiring message properties set (i.e AMQ_GROUP_ID).

### DIFF
--- a/artemis-rest/src/main/java/org/apache/activemq/artemis/rest/HttpHeaderProperty.java
+++ b/artemis-rest/src/main/java/org/apache/activemq/artemis/rest/HttpHeaderProperty.java
@@ -19,9 +19,11 @@ package org.apache.activemq.artemis.rest;
 public class HttpHeaderProperty {
 
    public static final String CONTENT_TYPE = "http_content$type";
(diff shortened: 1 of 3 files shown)


jbertram:
This needs a "New Feature" JIRA in https://issues.apache.org/jira/browse/ARTEMIS and the commit message needs to reference the JIRA. See [here](https://github.com/apache/activemq-artemis/blob/master/docs/hacking-guide/en/maintainers.md#commit-messages).

michaelandrepearce:
@jbertram this work seems good, and it be a shame to lose, i will look to create a jira amend the commit and merge if no objection

jbertram:
@michaelandrepearce, fair enough.

twyszomirski:
@michaelandrepearce @jbertram Nice, thanks guys.

michaelandrepearce:
https://issues.apache.org/jira/browse/ARTEMIS-2646

PR #2717, opened 2019-06-19
ARTEMIS-2354 Improve compatibillity of tests with JDK 11

franz1981:


jbertram:
Travis was messed up so I ran the PR build locally.

PR #2719, opened 2019-06-20
ARTEMIS-2358 - fixing issue with last commit

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-2358


franz1981:
I'm looking into testing with some DBs to check if is fine :+1:

franz1981:
It works fine for me :+1 I'm gonna merge it!

PR #2724, opened 2019-06-20
ARTEMIS-2391 static address may be auto-deleted

jbertram:


michaelandrepearce:
Will look at this next week. Just quickly in laymans terms whats the noticed issue trying to stop or prevent. Realise it might be evident when i review next week, as im sure tests added probably define it.

jbertram:
@michaelandrepearce, if you statically define an address (e.g. in broker.xml or via core API) and that address has no bindings it may be inadvertently removed by `org.apache.activemq.artemis.core.postoffice.impl.PostOfficeImpl.AddressQueueReaper`.

michaelandrepearce:
@jbertram thanks, make sense. +1 (though you've already merged it)

toolkit:
@michaelandrepearce - IG saw this issue in 7.3. 
We were advised to set address-queue-scan-period to -1 as a temporary workaround.

michaelandrepearce:
@toolkit i heard from @henock friday evening

toolkit:
A question on this. If addressQueueReaperPeriod is set to -1, how will an existing addressQueueReaperRunnable be stopped?

toolkit:
@jbertram - one question on this. It looks like  addressQueueReaperRunnable.stop() is only called if this object is not null, but also only if addressQueueReaperPeriod is > 0.

Assuming that the broker XML configuration file is only modified, and the broker is not restarted, then does this mean that the  addressQueueReaperRunnable cannot be disabled by setting it to -1 ?

jbertram:
@toolkit, that's correct. The broker will need to be restarted in order to stop the ` addressQueueReaperRunnable` and prevent it from starting again. You'll note in [the documentation](http://activemq.apache.org/components/artemis/documentation/latest/config-reload.html) that only a handful of settings actually get reloaded when changed via `broker.xml` at runtime.

PR #2725, opened 2019-06-21
ARTEMIS-2390 JMSMessageID header can be null when messages are cross-protocol

gaohoward:
If a jms client (be it openwire, amqp, or core jms) receives a message that
is from a different protocol, the JMSMessageID maybe null when the
jms client expects it.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
@@ -55,6 +55,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.UUID;
 
 import javax.jms.DeliveryMode;
 import javax.jms.JMSException;
@@ -67,6 +68,7 @@ import org.apache.activemq.artemis.protocol.amqp.converter.jms.ServerJMSMessage;
 import org.apache.activemq.artemis.protocol.amqp.converter.jms.ServerJMSStreamMessage;
 import org.apache.activemq.artemis.protocol.amqp.util.NettyWritable;
 import org.apache.activemq.artemis.protocol.amqp.util.TLSEncode;
+import org.apache.activemq.artemis.utils.UUIDGenerator;
 import org.apache.activemq.artemis.utils.collections.TypedProperties;
 import org.apache.qpid.proton.amqp.Binary;
 import org.apache.qpid.proton.amqp.Decimal128;
@@ -311,7 +313,15 @@ public class AmqpCoreConverter {
       if (properties != null) {
          if (properties.getMessageId() != null) {
             jms.setJMSMessageID(AMQPMessageIdHelper.INSTANCE.toMessageIdString(properties.getMessageId()));
+            //core jms clients get JMSMessageID from UserID which is a UUID object
+            if (properties.getMessageId() instanceof UUID) {
+               //AMQP's message ID can be a UUID, keep it
+               jms.getInnerMessage().setUserID(UUIDGenerator.getInstance().fromJavaUUID((UUID) properties.getMessageId()));
+            } else {
+               jms.getInnerMessage().setUserID(UUIDGenerator.getInstance().generateUUID());
+            }
          }
+
          Binary userId = properties.getUserId();
          if (userId != null) {
             jms.setStringProperty("JMSXUserID", new String(userId.getArray(), userId.getArrayOffset(), userId.getLength(), StandardCharsets.UTF_8));
@@ -345,6 +355,7 @@ public class AmqpCoreConverter {
             }
          }
          Object correlationID = properties.getCorrelationId();
+
          if (correlationID != null) {
             try {
                jms.getInnerMessage().setCorrelationID(AMQPMessageIdHelper.INSTANCE.toCorrelationIdString(correlationID));
@@ -374,7 +385,6 @@ public class AmqpCoreConverter {
             jms.setJMSExpiration(properties.getAbsoluteExpiryTime().getTime());
          }
       }
-
       return jms;
    }
 
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
@@ -159,6 +159,10 @@ public class CoreAmqpConverter {
          } catch (ActiveMQAMQPIllegalStateException e) {
             properties.setMessageId(messageId);
          }
+      } else {
+         if (message.getInnerMessage().getUserID() != null) {
+            properties.setMessageId("ID:" + message.getInnerMessage().getUserID().toString());
+         }
       }
       Destination destination = message.getJMSDestination();
       if (destination != null) {
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/MessageIDMultiProtocolTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/MessageIDMultiProtocolTest.java
@@ -1,0 +1,146 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.activemq.artemis.tests.integration.crossprotocol;
+
+import org.apache.activemq.artemis.api.core.RoutingType;
+import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.tests.integration.openwire.OpenWireTestBase;
+import org.apache.activemq.artemis.tests.util.Wait;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import javax.jms.Connection;
+import javax.jms.ConnectionFactory;
+import javax.jms.Destination;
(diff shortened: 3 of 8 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/MessageIDMultiProtocolTest.java
```diff
@@ -0,0 +1,138 @@
...
+   public void testMessageIDNotNullCorrelationIDPreserved() throws Throwable {
+      Connection senderConn = senderCF.createConnection();
+      Connection consumerConn = consumerCF.createConnection();
+      consumerConn.setClientID("consumer");
+
+      try (Session senderSession = senderConn.createSession(false, Session.AUTO_ACKNOWLEDGE)) {
+         Queue senderDestination = senderSession.createQueue(queueName.toString());
+         MessageProducer senderProducer = senderSession.createProducer(senderDestination);
+         Message sentMessage = senderSession.createMessage();
+         sentMessage.setJMSCorrelationID("ID:MessageIDCorrelationId");
```
jdanekrh: No need to set correlation id here, because nothing is asserted about it, and checking that it is preserved would be a different test (and I realize I am the one who introduced this to the first version of the test).

jdanekrh: Ok, there is the correlation id assert in the test later. Ok, so probably lets leave it here?

gaohoward: +agreed. I think the test is good. :)

jbertram:
LGTM

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -345,6 +346,12 @@ private static ServerJMSMessage processProperties(ServerJMSMessage jms, Properti
             }
          }
          Object correlationID = properties.getCorrelationId();
+
+         if (properties.getMessageId() != null) {
+            //core jms clients get JMSMessageID from UserID which is a UUID object
+            jms.getInnerMessage().setUserID(UUIDGenerator.getInstance().generateUUID());
```
tabish121: If the Core bits truly use UUID here then one thing you could do is attempt to preserve the original ID by checking if it is a UUID as that is one of the legal Message ID types:
http://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-message-id-uuid

By preserving a true UUID as message ID you'd also be reducing the cost of the conversion by avoiding string operations to create the UUID string and also to encode the string to UTF8 for on the wire vs sending a couple longs and a byte to represent the value.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
```diff
@@ -159,6 +160,12 @@ public static AMQPMessage fromCore(ICoreMessage coreMessage) throws Exception {
          } catch (ActiveMQAMQPIllegalStateException e) {
             properties.setMessageId(messageId);
          }
+      } else {
+         if (message.getInnerMessage().getUserID() != null) {
+            properties.setMessageId("ID:" + message.getInnerMessage().getUserID().toString());
+         } else {
+            properties.setMessageId("ID: " + UUIDGenerator.getInstance().generateStringUUID());
```
tabish121: But JMS does allow a message to **not** have a message ID using the disableMessageId option on the producer.  Not all the clients might support this option but it they did then generating a new ID seems to go against the grain of this issue which is to meet expectations albeit in this case in the negative.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
gaohoward: Looks like proton-j maps to java UUID, but the core message using it's own UUID type.
If we preserve the original uuid, we need to convert the type. I'll take a look at the code and what need to be changed.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
gaohoward: Yes that's true. If the core message doesn't have a userID, it should be kept as null.
I'll remove the line.

gaohoward:
@tabish121 I updated the branch. Can you take a look? thx.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -345,6 +347,17 @@ private static ServerJMSMessage processProperties(ServerJMSMessage jms, Properti
             }
          }
          Object correlationID = properties.getCorrelationId();
+
+         if (properties.getMessageId() != null) {
```
gemmellr: This seems like it should move up beside the other message id handling at the top of the method. Its strange this happens much later, right after getting the correlation-id value and before the actual subsequent use of that correlation-id value.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/MessageIDMultiProtocolTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+            Assert.assertEquals(sentMessage.getJMSCorrelationID(), receivedMessage.getJMSCorrelationID());
+
+            String messageId = receivedMessage.getJMSMessageID();
+            Assert.assertNotNull(messageId);
+
+            Assert.assertTrue(messageId.startsWith("ID:"));
+
+            if (protocolConsumer.equals(protocolSender)) {
+               //only same protocol we guarantee the same JMSMessageID
```
gemmellr: Is this still true? It seemed like with the changes at least some other cases might now align following the changes? If so, those should perhaps also be verified specifically, to prevent regression.

tabish121:
@gaohoward Changes look ok to me.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
gaohoward: Yes your are right, it should be moved to the top. I'll do it. thx

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/MessageIDMultiProtocolTest.java
gaohoward: Yes that's still true. Can you give some cases where it need to be verified? I don't quite get it. thx.

jdanekrh:
Typo in commit message, should be ARTEMIS-2390, not AETEMIS...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
jdanekrh: There should be a test for that. Lets say, https://invent.kde.org/snippets/280

The results of that test on this PR at this moment are

```
x [sender=OPENWIRE,consumer=<anything>]
x [sender=CORE,consumer=OPENWIRE]
✓ [sender=CORE,consumer=CORE]
✓ [sender=CORE,consumer=AMQP]
x [sender=AMQP,consumer=OPENWIRE]
✓ [sender=AMQP,consumer=CORE]
✓ [sender=AMQP,consumer=AMQP]
```

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/MessageIDMultiProtocolTest.java
jdanekrh: It is actually aligned when going CORE -> AMQP (but not the other way). In the CORE -> AMQP cases, I observed aligned IDs in the form `ID:e5b36859-973d-11e9-8ac1-0050b6aa100e`. When going the other way, there was something like
```
org.junit.ComparisonFailure: 
Expected :ID:7a995d3f-f285-47f2-849c-fba9ce252139:1:1:1-1
Actual   :ID:e5fdb7e1-973d-11e9-8ac1-0050b6aa100e
```
I wouldn't myself update the test to enforce the CORE->AMQP equivalence status quo.

gemmellr: Yes, Core->AMQP was one that looked like it could work.

gaohoward: I see the point. The changes brings another benefit for [core]-[amqp]. I'll  improve the test a bit.
thx

gaohoward:
Done.

jdanekrh:
Does the preservation of null messageID work in openwire? When I tried it last with this PR,

using test https://invent.kde.org/snippets/280

I got these results

```
x [sender=OPENWIRE,consumer=<anything>]
x [sender=CORE,consumer=OPENWIRE]
✓ [sender=CORE,consumer=CORE]
✓ [sender=CORE,consumer=AMQP]
x [sender=AMQP,consumer=OPENWIRE]
✓ [sender=AMQP,consumer=CORE]
✓ [sender=AMQP,consumer=AMQP]
```

michaelpearce-gain:
@jdanekrh thats a worry on openwire front, sounds almost blocking.

gaohoward:
hmm, the Openwire sender seems doesn't responds to the 'disable' call. That doesn't break the jms spec thou. Anyway the openwire uses MessageID internallly so if it's null there will be Exceptions thrown (which is true with or without this PR). I don't quite understand this background, need to look deeper.

gaohoward:
Guys, I believe this is the old openwire behavior that the producer always produces a messageID regardless of the disableMessageID flag. 
My suggestion is to keep it as is for this PR.

tabish121:
@gaohoward Openwire JMS client opts not to honor the disable message ID flag which is allowable by spec.  The client expects that incoming messages will have a message ID and has not been tested without one so my advice would be to always ensure that cross protocol conversions ensure that outgoing messages get a JMS Message ID set in the openwire message to avoid breaking any existing users.

clebertsuconic:
@gaohoward I am merging this, I'm just adding a small optimization where I'm adding a method longToBytes on ByteUtil and avoid an intermediate ByteBuffer to generate the byte array.

PR #2726, opened 2019-06-21
ARTEMIS-2392 Enable remove on cancel policy for scheduled pool

brusdev:
By default, such a cancelled task is not automatically removed from the
work queue until its delay elapses. It may cause unbounded retention of
cancelled tasks. To avoid this, set remove on cancel policy to true.

### DIFF
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImplTest.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImplTest.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.core.server.impl;
+
+import java.lang.ref.WeakReference;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.activemq.artemis.core.server.ActiveMQServer;
+import org.apache.activemq.artemis.tests.util.ActiveMQTestBase;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ActiveMQServerImplTest extends ActiveMQTestBase {
+
+   @Test
+   public void testScheduledPoolGC() throws Exception {
+      ActiveMQServer server = createServer(false);
+
+      server.start();
+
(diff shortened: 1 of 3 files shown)


michaelandrepearce:
Test please. So that the use case needed is understood and ensure no future regression.

brusdev:
@michaelandrepearce I added the test testScheduledPoolGC.

artemis-server/src/test/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImplTest.java
```diff
@@ -0,0 +1,56 @@
...
+      {
+         Runnable scheduledRunnable = new Runnable() { @Override public void run() { } };
+         scheduledRunnableRef = new WeakReference<>(scheduledRunnable);
+
+         ScheduledFuture scheduledFuture = server.getScheduledPool().schedule(scheduledRunnable, 5000, TimeUnit.MILLISECONDS);
+
+         scheduledFuture.cancel(true);
+      }
+
+      Assert.assertNotEquals(null, scheduledRunnableRef.get());
```
franz1981: GC is not precise (when things got out of scope/not referenced they got collected immediately) and depend on the type of GC chosen: probably you need to issue (full) GC and right after check the weak reference, but it is not reliable enough. We have other tests that do something similar using an ad-hoc call to GC, will point you them on Monday

clebertsuconic:
ForceGC uses a loop and forceGC to make the forced gc


You will need to call system.gc and a loop to make the proper assertion.

brusdev:
@clebertsuconic I changed the test.

artemis-server/src/test/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImplTest.java
```diff
@@ -0,0 +1,70 @@
...
+         Assert.assertTrue(scheduledFuture.cancel(true));
+         Assert.assertTrue(scheduledFuture.isCancelled());
+
+         Assert.assertNotEquals(null, scheduledRunnableRef.get());
+      }
+
+      long timeout = System.currentTimeMillis() + 5000;
+
+      while (timeout > System.currentTimeMillis() && scheduledRunnableRef.get() != null) {
+         System.gc();
```
clebertsuconic: We have a call on the ActiveMQTestBAse:: forceGC

artemis-server/src/test/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImplTest.java
```diff
@@ -0,0 +1,63 @@
...
+      WeakReference<Runnable> scheduledRunnableRef = new WeakReference<>(scheduledRunnable);
+
+      ScheduledExecutorService scheduledPool = server.getScheduledPool();
+      ScheduledFuture scheduledFuture = scheduledPool.schedule(scheduledRunnable, 5000, TimeUnit.MILLISECONDS);
+
+      Assert.assertFalse(scheduledFuture.isCancelled());
+      Assert.assertTrue(scheduledFuture.cancel(true));
+      Assert.assertTrue(scheduledFuture.isCancelled());
+
+      Assert.assertNotEquals(null, scheduledRunnableRef.get());
```
wy96f: `Assert.assertNotEquals(null, scheduledRunnableRef.get());` should be put before 
 `Assert.assertTrue(scheduledFuture.cancel(true));`. GC might occur right at this moment and cause reference object cleared, resulting in test failure.

brusdev: The scheduledRunnableRef referent shouldn't be cleared until the local variable scheduledRunnable won't be set null.

wy96f: you are right, i missed that :)

PR #2727, opened 2019-06-21
ARTEMIS-2394 Improve scheduling sync for AMQPConnectionContext

brusdev:
Remove synchronized blocks using an AtomicReference.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
@@ -22,10 +22,13 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Future;
+import java.util.concurrent.FutureTask;
 import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.UnaryOperator;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.channel.EventLoop;
@@ -35,6 +38,7 @@ import org.apache.activemq.artemis.protocol.amqp.broker.AMQPConnectionCallback;
 import org.apache.activemq.artemis.protocol.amqp.broker.AMQPSessionCallback;
 import org.apache.activemq.artemis.protocol.amqp.broker.ProtonProtocolManager;
 import org.apache.activemq.artemis.protocol.amqp.exceptions.ActiveMQAMQPException;
+import org.apache.activemq.artemis.protocol.amqp.logger.ActiveMQAMQPProtocolLogger;
 import org.apache.activemq.artemis.protocol.amqp.proton.handler.EventHandler;
 import org.apache.activemq.artemis.protocol.amqp.proton.handler.ExecutorNettyAdapter;
 import org.apache.activemq.artemis.protocol.amqp.proton.handler.ExtCapability;
@@ -71,6 +75,7 @@ public class AMQPConnectionContext extends ProtonInitializable implements EventH
 
    public static final Symbol CONNECTION_OPEN_FAILED = Symbol.valueOf("amqp:connection-establishment-failed");
    public static final String AMQP_CONTAINER_ID = "amqp-container-id";
+   private static final FutureTask<Void> VOID_FUTURE = new FutureTask<>(() -> { }, null);
 
    protected final ProtonHandler handler;
 
@@ -87,9 +92,8 @@ public class AMQPConnectionContext extends ProtonInitializable implements EventH
 
    private final boolean useCoreSubscriptionNaming;
 
-   private boolean isSchedulingCancelled;
-   private ScheduledFuture scheduledFuture;
-   private final Object schedulingLock = new Object();
+   private final ScheduleOperator scheduleOp = new ScheduleOperator(new ScheduleRunnable());
+   private final AtomicReference<Future<?>> scheduledFutureRef = new AtomicReference(VOID_FUTURE);
 
    public AMQPConnectionContext(ProtonProtocolManager protocolManager,
                                 AMQPConnectionCallback connectionSP,
@@ -117,8 +121,6 @@ public class AMQPConnectionContext extends ProtonInitializable implements EventH
          this.connectionProperties.putAll(connectionProperties);
       }
 
-      this.scheduledFuture = null;
-      this.isSchedulingCancelled = false;
       this.scheduledPool = scheduledPool;
       connectionCallback.setConnection(this);
       EventLoop nettyExecutor;
@@ -191,17 +193,13 @@ public class AMQPConnectionContext extends ProtonInitializable implements EventH
    }
 
    public void close(ErrorCondition errorCondition) {
-      synchronized (schedulingLock) {
-         isSchedulingCancelled = true;
+      Future<?> scheduledFuture = scheduledFutureRef.getAndSet(null);
 
-         if (scheduledPool != null && scheduledPool instanceof ThreadPoolExecutor &&
-            scheduledFuture != null && scheduledFuture instanceof Runnable) {
-            if (!((ThreadPoolExecutor) scheduledPool).remove((Runnable) scheduledFuture) &&
-               !scheduledFuture.isCancelled() && !scheduledFuture.isDone()) {
-               log.warn("Scheduled task can't be removed from scheduledPool.");
-            } else {
-               scheduledFuture = null;
-            }
+      if (scheduledPool instanceof ThreadPoolExecutor && scheduledFuture != null &&
+         scheduledFuture != VOID_FUTURE && scheduledFuture instanceof Runnable) {
+         if (!((ThreadPoolExecutor) scheduledPool).remove((Runnable) scheduledFuture) &&
+            !scheduledFuture.isCancelled() && !scheduledFuture.isDone()) {
+            ActiveMQAMQPProtocolLogger.LOGGER.cantRemovingScheduledTask();
          }
       }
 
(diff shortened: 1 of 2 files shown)


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -87,9 +88,7 @@
 
    private final boolean useCoreSubscriptionNaming;
 
-   private boolean isSchedulingCancelled;
-   private ScheduledFuture scheduledFuture;
-   private final Object schedulingLock = new Object();
+   private AtomicReference scheduledFutureRef;
```
franz1981: make it final and it would be better to use the not raw version by using <>. You can always create a singleton instance to be used to do your logic

michaelandrepearce:
Any stats on the improvement?

michaelandrepearce:
forgot to say +1 with removing sync blocks, the question is more a query not a reason not to merge.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
brusdev: I made it final. How could I use a singleton instance in this case?

brusdev:
I executed testLeakAfterClose 5000 times with CPU profiling before and after the changes but I am not very satisfied with the results:
- Before AMQPConnectionContext.close() CPU Time: 661ms
- After AMQPConnectionContext.close() CPU Time: 864ms

franz1981:
I won't trust manually handrolled benchmarks or common sampling profilers: I suggest instead to use JMH (for benchmarking) and async-profiler as a profiler instead.
There are many optimizations that the JVM can (and will) do that doesn't make results reliable in this case..
Will explain about it on Monday mate :)

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -436,16 +434,18 @@ public void onRemoteOpen(Connection connection) throws Exception {
       public void run() {
          Long rescheduleAt = handler.tick(false);
 
-         synchronized (schedulingLock) {
-            if (!isSchedulingCancelled) {
+         scheduledFutureRef.getAndUpdate(value -> {
+            if (value != null) {
                if (rescheduleAt == null) {
```
michaelandrepearce: this can be moved outside the function, as if rescheduleAt == null, then you dont need to make a getAndUpdate method call, but just a get, and then the function class can be one per parent object, instead of creating one per call.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -416,12 +411,15 @@ public void onRemoteOpen(Connection connection) throws Exception {
        * */
       if (connection.getRemoteProperties() == null || !connection.getRemoteProperties().containsKey(CONNECTION_OPEN_FAILED)) {
          long nextKeepAliveTime = handler.tick(true);
-         synchronized (schedulingLock) {
-            if (nextKeepAliveTime != 0 && scheduledPool != null && !isSchedulingCancelled) {
-               scheduledFuture = scheduledPool.schedule(new ScheduleRunnable(), (nextKeepAliveTime - TimeUnit.NANOSECONDS.toMillis(System.nanoTime())), TimeUnit.MILLISECONDS);
+
+         scheduledFutureRef.getAndUpdate(value -> {
+            if (value != null && nextKeepAliveTime != 0 && scheduledPool != null) {
```
michaelandrepearce: nextKeepAliveTime != 0 && scheduledPool != null can be pulled outside the atomic function call.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -87,9 +89,8 @@
 
    private final boolean useCoreSubscriptionNaming;
 
-   private boolean isSchedulingCancelled;
-   private ScheduledFuture scheduledFuture;
-   private final Object schedulingLock = new Object();
+   private static final FutureTask<Void> voidFuture = new FutureTask<>(() -> { }, null);
```
michaelandrepearce: VOID_FUTURE as this is a constant.

michaelandrepearce: on that note what is its purpose? surely can just set the reference to null

brusdev: ScheduledFutureRef is used for 2 purpose: keeping the last scheduledFuture and the scheduling status, so I need tree states for scheduledFutureRef:
- VOID_FUTURE before onRemoteOpen / NOT SCHEDULED / SCHEDULING ALLOWED
- NULL after close / SCHEDULING NOT ALLOWED
- ELSE after onRemoteOpen and before close / SCHEDULED / SCHEDULING ALLOWED

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
brusdev: In the scenario in which close is executed before or concurrently onRemoteOpen, the scheduling must be prevented. Checking the current value of scheduledFutureRef is useful to avoid scheduling after close because it is used for 2 purpose: keeping the last scheduledFuture and the scheduling status.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
brusdev: I completely agree I will pull the "nextKeepAliveTime != 0 && scheduledPool != null" out outside.

brusdev:
I changed the commit with @franz1981 and @michaelandrepearce suggestions.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
franz1981: I suppose that @michaelandrepearce means to rename the variable from `voidFuture` to `VOID_FUTURE`

michaelandrepearce: no i was literally meaning why the need for this, just set reference to null, i dont see the need for three states, two states is all thats needed looking at code, and the state changes.

my first comment was just it should be capitlised, still not changed, but that was before i looked at its use and now question its need at all.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -191,17 +190,13 @@ public void flush() {
...
-            } else {
-               scheduledFuture = null;
-            }
+      Future scheduledFuture = scheduledFutureRef.getAndSet(null);
+
+      if (scheduledPool != null && scheduledPool instanceof ThreadPoolExecutor &&
+         scheduledFuture != null && scheduledFuture != voidFuture && scheduledFuture instanceof Runnable) {
+         if (!((ThreadPoolExecutor) scheduledPool).remove((Runnable) scheduledFuture) &&
+            !scheduledFuture.isCancelled() && !scheduledFuture.isDone()) {
+            log.warn("Scheduled task can't be removed from scheduledPool.");
```
michaelandrepearce: parameterised logger needed

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
michaelandrepearce: but rescheduled at, is calculated once before it hits into the getAndUpdate its doesnt change, either it will be null or not, before the atomic, and there for the logic causing can come outside the function and we can then make the function have not outside value purer functional.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
michaelandrepearce: still waiting on this change.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
michaelandrepearce: e.g.
```
         Long rescheduleAt = handler.tick(false);
         if (rescheduleAt == null) {
            scheduledFutureRef.getAndUpdate(value -> value == null ? null : scheduledPool.schedule(scheduleRunnable, 10, TimeUnit.MILLISECONDS));
         } else if (rescheduleAt != 0) {
            scheduledFutureRef.getAndUpdate(value -> value == null ? null : scheduledPool.schedule(scheduleRunnable, rescheduleAt - TimeUnit.NANOSECONDS.toMillis(System.nanoTime()), TimeUnit.MILLISECONDS));
         }
```

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
michaelandrepearce:
```
      if (scheduledPool != null && (connection.getRemoteProperties() == null || !connection.getRemoteProperties().containsKey(CONNECTION_OPEN_FAILED))) {
         long nextKeepAliveTime = handler.tick(true);
         if (nextKeepAliveTime != 0) {
            scheduledFutureRef.getAndUpdate(value -> {
               if (value != null) {
                  return scheduledPool.schedule(new ScheduleRunnable(), (nextKeepAliveTime - TimeUnit.NANOSECONDS.toMillis(System.nanoTime())), TimeUnit.MILLISECONDS);
               }

               return value;
            });
         }
      }
```

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
michaelandrepearce: ok, think ive spotted what you're trying to achieve with the VOID_FUTURE. still part about capitilised naming for constant remains.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -191,17 +190,13 @@ public void flush() {
...
-            if (!((ThreadPoolExecutor) scheduledPool).remove((Runnable) scheduledFuture) &&
-               !scheduledFuture.isCancelled() && !scheduledFuture.isDone()) {
-               log.warn("Scheduled task can't be removed from scheduledPool.");
-            } else {
-               scheduledFuture = null;
-            }
+      Future scheduledFuture = scheduledFutureRef.getAndSet(null);
+
+      if (scheduledPool != null && scheduledPool instanceof ThreadPoolExecutor &&
+         scheduledFuture != null && scheduledFuture != voidFuture && scheduledFuture instanceof Runnable) {
```
michaelandrepearce: scheduledFuture != null is redundent, due to scheduledFuture instanceOf runnable check.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -416,42 +414,57 @@ public void onRemoteOpen(Connection connection) throws Exception {
...
          long nextKeepAliveTime = handler.tick(true);
-         synchronized (schedulingLock) {
-            if (nextKeepAliveTime != 0 && scheduledPool != null && !isSchedulingCancelled) {
-               scheduledFuture = scheduledPool.schedule(new ScheduleRunnable(), (nextKeepAliveTime - TimeUnit.NANOSECONDS.toMillis(System.nanoTime())), TimeUnit.MILLISECONDS);
-            }
+
+         if (nextKeepAliveTime != 0 && scheduledPool != null) {
+            scheduleOp.setDelay(nextKeepAliveTime - TimeUnit.NANOSECONDS.toMillis(System.nanoTime()));
+
+            scheduledFutureRef.getAndUpdate(scheduleOp);
```
franz1981: there is some specific return value from getOrUpdate that you expect to find? if yes, maybe you can put an assert there to help spot issues on tests...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -87,9 +92,8 @@
 
    private final boolean useCoreSubscriptionNaming;
 
-   private boolean isSchedulingCancelled;
-   private ScheduledFuture scheduledFuture;
-   private final Object schedulingLock = new Object();
+   private final ScheduleOperator scheduleOp = new ScheduleOperator(new ScheduleRunnable());
+   private final AtomicReference<Future> scheduledFutureRef = new AtomicReference(VOID_FUTURE);
```
franz1981: Please use Future<?> , do not use the raw type if possible...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -416,42 +414,57 @@ public void onRemoteOpen(Connection connection) throws Exception {
...
+         if (nextKeepAliveTime != 0 && scheduledPool != null) {
+            scheduleOp.setDelay(nextKeepAliveTime - TimeUnit.NANOSECONDS.toMillis(System.nanoTime()));
+
+            scheduledFutureRef.getAndUpdate(scheduleOp);
          }
       }
    }
 
-   class TickerRunnable implements Runnable {
+   class ScheduleOperator implements UnaryOperator<Future> {
```
franz1981: Future<?> here as well

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
brusdev: The return value depends on current value so I don't know how to predict it without access to scheduledFutureRef.

PR #2728, opened 2019-06-21
ARTEMIS-2395 empty address.txt causes NPE

jbertram:


jbertram:
Don't merge yet. I'm adding a test.

PR #2730, opened 2019-06-24
ARTEMIS-2399 Fix performance degradation when there are a lot of subscribers

wy96f:
We noticed that there was a significant drop in performance when entering page mode in the case of multiple subscribers.

### **Environment**:
broker 2.9.0
cpu: 4 cores, memory: 8G, disk: ssd 500G
broker.xml:
         `<thread-pool-max-size>60</thread-pool-max-size>
         <address-setting match="#">
                 <max-size-bytes>51Mb</max-size-bytes>
                 <page-size-bytes>50Mb</page-size-bytes>
                 <page-max-cache-size>1</page-max-cache-size>
                 <address-full-policy>PAGE</address-full-policy>
         </address-setting> 
         <message-expiry-scan-period>-1</message-expiry-scan-period>`
         
### **Test steps**:
We created a topic and 100 queues bound to it. We ran our GrinderRunner test in our inner test infra cluster with 500 threads producing messages(200-500 bytes size) and 560 threads, each one picked a random queue to subscribe. The test showed performance is bad: 13000 msg/s sent and 5000 msg/s received.
producer tps and latency:
orig_producer
consumer tps and latency:
orig_consumer

### **Analysis**:
There were two root causes:
1. Usually 1000 messages are delivered at once, sometimes less if consumers are busy. Then depage() is called to fill the queue with number of messages delivered in last step. To read these messages we need to read the whole page. It's acceptable if the page is in the softValueCache. But in the case of multiple subscribers with different cursor position, they read there own page file, put it in the cache and evict other pages. At the later time, when they need to read it again, they found it not in cache and have to read the whole page from disk which spends much time and puts pressure on disk.
2. For multiple subscribers to the same address, just one executor is responsible for delivering which means at the same moment only one queue is delivering. Thus the queue maybe stalled for a long time. We get queueMemorySize messages into memory, and when we deliver these after a long time, we probably need to query message and read page file again.

### **Solution**:
1. We add a new cache called PageIndexCache which stores the message number and file position. This cache is built when we first read the page and put it into softValueCache. If the page is evicted from softValueCache later, we'll use PageIndexCache to read message. In this way, we don't need to read the whole page for just a few messages.
2. In most cases, one depage round is followed by at most MAX_SCHEDULED_RUNNERS deliver round. Thus we just need to read MAX_DELIVERIES_IN_LOOP * MAX_SCHEDULED_RUNNERS messages. This reduces the possibility of requering the reference case the Garbage Collection removes it and keeps enough messages to deliver each time.

With this pr we get the result: 11000 msg/s sent and 16500 msg/s received.
producer tps and latency:
pr1_producer
consumer tps and latency:
pr1_consumer

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
@@ -19,6 +19,7 @@ package org.apache.activemq.artemis.core.paging.cursor.impl;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.TreeMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
@@ -74,6 +75,8 @@ public class PageCursorProviderImpl implements PageCursorProvider {
 
    private final SoftValueLongObjectHashMap<PageCache> softCache;
 
+   private final LongObjectHashMap<PageCache> indexCacheMap;
+
    private final LongObjectHashMap<CompletableFuture<PageCache>> inProgressReadPages;
 
    private final ConcurrentLongHashMap<PageSubscription> activeCursors = new ConcurrentLongHashMap<>();
@@ -99,6 +102,7 @@ public class PageCursorProviderImpl implements PageCursorProvider {
       this.storageManager = storageManager;
       this.executor = executor;
       this.softCache = new SoftValueLongObjectHashMap<>(maxCacheSize);
+      this.indexCacheMap = new LongObjectHashMap<>();
       this.inProgressReadPages = new LongObjectHashMap<>();
    }
 
@@ -145,6 +149,22 @@ public class PageCursorProviderImpl implements PageCursorProvider {
 
    @Override
    public PageCache getPageCache(final long pageId) {
+      PageCache pageCache = getCacheFromSoftCache(pageId, false);
+      if (pageCache == null) {
+         synchronized (softCache) {
+            pageCache = softCache.get(pageId);
+            if (pageCache == null) {
+               pageCache = indexCacheMap.get(pageId);
+            }
+         }
+      }
+      if (pageCache == null) {
+         pageCache = getCacheFromSoftCache(pageId, true);
+      }
+      return pageCache;
+   }
+
+   private PageCache getCacheFromSoftCache(final long pageId, boolean read) {
       try {
          if (pageId > pagingStore.getCurrentWritingPage()) {
             return null;
@@ -152,6 +172,7 @@ public class PageCursorProviderImpl implements PageCursorProvider {
          boolean createPage = false;
          CompletableFuture<PageCache> inProgressReadPage;
          PageCache cache;
+         PageIndexCacheImpl indexCache = null;
          Page page = null;
          synchronized (softCache) {
             cache = softCache.get(pageId);
@@ -163,8 +184,12 @@ public class PageCursorProviderImpl implements PageCursorProvider {
             }
             inProgressReadPage = inProgressReadPages.get(pageId);
             if (inProgressReadPage == null) {
+               if (!read) {
+                  return null;
+               }
                final CompletableFuture<PageCache> readPage = new CompletableFuture<>();
                cache = createPageCache(pageId);
+               indexCache = createPageIndexCache(pageId);
                page = pagingStore.createPage((int) pageId);
                createPage = true;
                inProgressReadPage = readPage;
@@ -172,7 +197,7 @@ public class PageCursorProviderImpl implements PageCursorProvider {
             }
          }
          if (createPage) {
-            return readPage(pageId, page, cache, inProgressReadPage);
+            return readPage(pageId, page, cache, indexCache, inProgressReadPage);
          } else {
             final long startedWait = System.nanoTime();
             while (true) {
@@ -194,6 +219,7 @@ public class PageCursorProviderImpl implements PageCursorProvider {
    private PageCache readPage(long pageId,
                               Page page,
                               PageCache cache,
+                              PageIndexCacheImpl indexCache,
                               CompletableFuture<PageCache> inProgressReadPage) throws Exception {
       logger.tracef("adding pageCache pageNr=%d into cursor = %s", pageId, this.pagingStore.getAddress());
       boolean acquiredPageReadPermission = false;
@@ -209,7 +235,16 @@ public class PageCursorProviderImpl implements PageCursorProvider {
          }
          page.open();
          final long startedReadPage = System.nanoTime();
-         List<PagedMessage> pgdMessages = page.read(storageManager);
+         List<PagedMessage> pgdMessages;
+         if (indexCache == null) {
+            pgdMessages = page.read(storageManager);
+         } else {
+            TreeMap<Integer, Integer> messageNumberToOffset = new TreeMap<>();
+            pgdMessages = page.read(storageManager, messageNumberToOffset);
+            indexCache.setMessageIndex(messageNumberToOffset);
+            indexCache.setNumberOfMessages(pgdMessages.size());
+            indexCache.setPage(page);
+         }
          final long elapsedReadPage = System.nanoTime() - startedReadPage;
          if (elapsedReadPage > PAGE_READ_TIMEOUT_NS) {
             logger.warnf("Page::read for pageNr=%d on cursor %s tooks %d ms to read %d bytes", pageId,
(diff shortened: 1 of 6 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
```diff
@@ -74,6 +75,8 @@
 
    private final SoftValueLongObjectHashMap<PageCache> softCache;
 
+   private final LongObjectHashMap<PageCache> indexCacheMap;
```
clebertsuconic: Wouldn't this break SoftReferences cleanup? meaning a page will be leaving forever.

I would use a SoftValueLongObjectHashMap here as well. In case memory is low, we want PageCache to be releasable. And this would make hard references on pageCache.

<<this is a question>>

franz1981:
@wy96f 5000 msg/sec seems to me too similar to the default value for page-max-concurrent-io ie "The max number of concurrent reads allowed on paging. Default=5"
Could you check the results again by increasing the default to the total number of consumers?
To avoid being backpressured by the `Semaphore` when the pages are being eviced?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
franz1981: @clebertsuconic They have implemented a specialized `PageCache` that just hold the offset into the page file to read the data (kafka does something similar), not the full fate `PageCache` we are used to reference with Soft references, but I agree with Clebert that it has some GC costs (both due to the TreeMap and the Integer usage).
I think this could improved by using something primitive or (even better) by storing the offset in the page file in some paged message reference instead, wdyt?

clebertsuconic:
I don't consider this a fix (as in a bug fix). it's rather an improvement (nice one). As the previous version was performing as designed. I would tell users to manage their subscriptions, and paging was meant as a palleative.. keep your messages in memory (not paged).. blabblablabal...

but this is a game changer for paging.


I will need to carefully look at the GC implications though.. but this is really promising.

franz1981:
> I will need to carefully look at the GC implications though.. but this is really promising.

If we manage to put the fileoffset into each Paged reference we don't need the page cache but we can just read directly from the files. If the file regions are *hot*, the OS will make it very performant to read it, basically like reading from the RAM. :P

wy96f:
> @wy96f 5000 msg/sec sounds to me too similar to the default value for page-max-concurrent-io ie "The max number of concurrent reads allowed on paging. Default=5"
> Could you check the results again by increasing the default to the total number of consumers?
> To avoid being backpressured by the `Semaphore` when the pages are being evicted?

We adjusted page-max-concurrent-io to 560, the result was similar.
producer tps and latency:
orig_560_producer

consumer tps and latency:
orig_560_consumer

franz1981:
@wy96f thanks to have tried! Did you have considered the other comments? Related to putting the file offset into the pages message reference? This would avoid to creating the Map<Integer,Integer>

wy96f:
The page index cache would reside until the page is deleted when 

> @wy96f thanks to have tried! Did you have considered the other comments? Related to putting the file offset into the pages message reference? This would avoid to creating the Map<Integer,Integer>

For the created page reference, we can record the file offset which can be used to seek and read file if message is removed by gc. But for the message not yet depaged or not committed in a page transaction, how do we read message and create reference just according to page number and message number?

wy96f:
@clebertsuconic We have considered memory usage by page index cache.  Out first thought was the usage by index cache is not as much as the pageCursorInfo used by the PageSubscription given the cache wouldn't be removed until all the subscribers have consumed this page where pageCursorfInfo will be removed too. So it's worth although not cheap. But in the case that there are some subscribers not working, the index cache would grow indefinitely while entries in pageCursorInfo are clean up. To make this problem less severe, we can increase the index interval byte size to reduce index cache memory or build a mechanism to evict the index cache?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
```diff
@@ -312,6 +451,10 @@ public void close() throws Exception {
     * While reading the cache we don't need (and shouldn't inform the backup
     */
    public synchronized void close(boolean sendEvent) throws Exception {
+      if (readFileBuffer != null) {
+         fileFactory.releaseBuffer(readFileBuffer);
```
franz1981: Currently the NIO pool is thread local, so I'm not sure is a good idea to do this on different threads (very likely).
We could implement a non-thread local pool or (better), do not use the pool but directly direct ByteBuffer instances obtained from the factory

michaelandrepearce:
@wy96f i would go with an index cache that would evict. This would be most scalable imo.

franz1981:
> For the created page reference, we can record the file offset which can be used to seek and read file if message is removed by gc. But for the message not yet depaged or not committed in a page transaction, how do we read message and create reference just according to page number and message number?

I need to look better how paging works to answer on this :)
Looking just at your question I can tell that we need something like this: https://github.com/real-logic/agrona/blob/master/agrona/src/main/java/org/agrona/collections/BiInt2ObjectMap.java
If we care about memory footprint, but I believe there must be a way to make it simpler too

michaelandrepearce:
@wy96f is the tool you are using "GrinderRunner" available so i could run on different setups, to validate the perf result is not just for benefit one use case/setup.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
```diff
@@ -209,7 +235,16 @@ private PageCache readPage(long pageId,
          }
          page.open();
          final long startedReadPage = System.nanoTime();
-         List<PagedMessage> pgdMessages = page.read(storageManager);
+         List<PagedMessage> pgdMessages;
+         if (indexCache == null) {
+            pgdMessages = page.read(storageManager);
+         } else {
+            TreeMap<Integer, Integer> messageNumberToOffset = new TreeMap<>();
```
michaelandrepearce: If this has to stay, Can we use a primitive map variant.

michaelandrepearce: Or even could this be simply an array of int, where array size will be max messageid - min messageid, and then simply access the offset by int[messageid - min messageid]

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageIndexCacheImpl.java
```diff
@@ -0,0 +1,108 @@
...
+import org.jboss.logging.Logger;
+
+public class PageIndexCacheImpl implements PageCache {
+   private static final Logger logger = Logger.getLogger(PageIndexCacheImpl.class);
+
+   private final long pageId;
+   private Page page;
+
+   // message number to file position map
+   private TreeMap<Integer, Integer> messageIndex;
```
michaelandrepearce: ditto, could this be primitive map, or even simply an array of ints, using the array idx as the key.

clebertsuconic:
We are going through so many scholar back and forths.. and I'm lost now...

I don't understand why we need two page caches now.. one on softCache, one on indexCacheMap

The idea of the softCache was to let PageCache go, and be reconstructed if memory is really low.


If you are adding a permanent cache, why not remove the soft Cache? Why do you intend to break the soft references here?


A SoftCache is only released if memory is really low.. so why break that rule?

clebertsuconic:
@wy96f it would be nice to bring these discussions to the Dev List .. this is really great stuff to be discussed there.


The only thing I'm concerned is the breaking of SoftReferences here... we either remove SoftReferences altogether, or we keep them the way they are now... it seems we are in midle ground.

clebertsuconic:
I think I'm starting to understand.. you created a new class called PageIndexCacheImpl.. and that new attribute is for the new version.


so, why not just change PageCacheImpl, and keep it on the soft cache. What would be the trouble on reconstructing (re-reading) the cache in case it's gone?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
clebertsuconic: And.. is there a way to encapsulate this logic on the PageIndexCacheImpl? just for code organization?

clebertsuconic:
Any chance we could chat about this? we could arrange a meeting open to the community, we could chat on slack.. dunno!?

michaelandrepearce:
I really like @franz1981 idea of even putting the offset into the pagedreference, personally.

wy96f:
Hi, all, I've brought a discussion to dev list [http://activemq.2283324.n4.nabble.com/Improve-paging-performance-when-there-are-lots-of-subscribers-td4751333.html](url)

wy96f:
#2750

PR #2731, opened 2019-06-25
[ARTEMIS-2401]: Implement the Pause method for a Topic

ehsavoie:
Adding support to pause a topic

Jira: https://issues.apache.org/jira/browse/ARTEMIS-2401


jbertram:
Since pausing an address is essentially just pausing all of the bound queues (i.e. not the actual consumers) I think the method names should reflect that. Therefore, I recommend using `pause()`, `resume()`, & `isPaused()`. The `AddressControl` should also expose `paused` so users can determine whether or not an address is paused administratively.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AddressControlImpl.java
```diff
@@ -373,6 +371,58 @@ public String sendMessage(final Map<String, String> headers,
...
+      clearIO();
+      try {
+         if (!addressInfo.isPaused()) {
+            addressInfo.pauseConsumers();
+            Bindings bindings = server.getPostOffice().lookupBindingsForAddress(addressInfo.getName());
+            if (bindings != null) {
+               for (Binding binding : bindings.getBindings()) {
+                  if (binding instanceof QueueBinding) {
+                     QueueControl coreQueueControl = (QueueControl) managementService.getResource(ResourceNames.QUEUE + binding.getUniqueName().toString());
+                     coreQueueControl.pause();
```
brusdev: I have a little concern here and in resumeConsumers. In a case with many bindings one of related queues could be deleted during the iteration and coreQueueControl could be null. Adding a check on coreQueueControl before to call coreQueueControl.pause should avoid to interrupt the iteration for a NullPointerException.

ehsavoie:
Persistence added
Commit message updated
Methods renamed

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
```diff
@@ -135,4 +135,35 @@ String sendMessage(@Parameter(name = "headers", desc = "The headers to add to th
...
+   /**
+    * Pauses all the consumers of this address.Messages are no longer delivered to all its queues consumers.
+    * Newly added consumers will be paused too until resume is called.
+    * @throws java.lang.Exception
+    */
+   @Operation(desc = "Pauses the queues bound to this address", impact = MBeanOperationInfo.ACTION)
+   void pause(@Parameter(name = "persist", desc = "if true, the pause state will be persisted.") boolean persist) throws Exception;
+
+   /**
+    * Resume all the consumers of this address.Messages are no longer delivered to all its queues consumers.
```
michaelandrepearce: Can you update the doc too please. It should state it pauses the queues, not consumers.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
```diff
@@ -135,4 +135,35 @@ String sendMessage(@Parameter(name = "headers", desc = "The headers to add to th
...
+   /**
+    * Pauses all the consumers of this address.Messages are no longer delivered to all its queues consumers.
+    * Newly added consumers will be paused too until resume is called.
+    * @throws java.lang.Exception
+    */
+   @Operation(desc = "Pauses the queues bound to this address", impact = MBeanOperationInfo.ACTION)
+   void pause() throws Exception;
+
+   /**
+    * Pauses all the consumers of this address.Messages are no longer delivered to all its queues consumers.
```
michaelandrepearce: Can you update the doc too please. It should state it pauses the queues, not consumers.

michaelandrepearce:
As this is a new feature can there be some docs added so users are aware or it and can understand.

ehsavoie:
@michaelandrepearce where does this doc live ?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/TopicPauseTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.activemq.artemis.api.core.management.AddressControl;
+import org.apache.activemq.artemis.api.core.management.ResourceNames;
+import org.apache.activemq.artemis.tests.util.JMSTestBase;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TopicPauseTest extends JMSTestBase {
+
+   @Test
+   public void testPauseQueue() throws Exception {
```
michaelandrepearce: Isnt this test, testing pausing an address

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
```diff
@@ -1481,6 +1509,7 @@ public JournalLoadInformation loadBindingJournal(final List<QueueBindingInfo> qu
          } else if (rec == JournalRecordIds.ADDRESS_BINDING_RECORD) {
             PersistentAddressBindingEncoding bindingEncoding = newAddressBindingEncoding(id, buffer);
             addressBindingInfos.add(bindingEncoding);
+            mapAddressBindings.put(id, bindingEncoding);
          } else if (rec == JournalRecordIds.GROUP_RECORD) {
```
michaelandrepearce: Is the a compatibility test to ensure broker upgrade downgrade there is no issues with bindings. E.g. old broker can read new binding files ok, and new broker version can read old generated binding files.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
```diff
@@ -135,4 +135,34 @@ String sendMessage(@Parameter(name = "headers", desc = "The headers to add to th
                       @Parameter(name = "durable", desc = "Whether the message is durable") boolean durable,
                       @Parameter(name = "user", desc = "The user to authenticate with") String user,
                       @Parameter(name = "password", desc = "The users password to authenticate with") String password) throws Exception;
+
+   /**
+    * Pauses all the consumers of this address.Messages are no longer delivered to all its queues consumers.
```
michaelandrepearce: Ditto

ehsavoie: sorry

michaelandrepearce: Same issue as other comments you just fixed

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
ehsavoie: Is there such a test already existing somewhere ?

michaelandrepearce: Yes theres a whole compatibility test suite.

michaelandrepearce: @clebertsuconic will be best to chat to

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AddressControlImpl.java
```diff
@@ -373,6 +372,86 @@ public String sendMessage(final Map<String, String> headers,
...
+      }
+      checkStarted();
+
+      clearIO();
+      try {
+         if (!addressInfo.isPaused()) {
+            addressInfo.pause();
+            if (persist) {
+               server.getStorageManager().storeAddressStatus(addressInfo.getId(), AddressStatus.PAUSED);
+            }
```
jbertram: I think this logic should be moved into the `AddressInfo#pause` method so users of `AddressInfo` get the persistence functionality without having to use the control object. Also, there needs to be a way to detect multiple invocations of `pause` so that PAUSED records don't just keep getting added over and over as that will cause the bindings journal size to increase over time.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AddressControlImpl.java
```diff
@@ -373,6 +372,86 @@ public String sendMessage(final Map<String, String> headers,
...
+            Bindings bindings = server.getPostOffice().lookupBindingsForAddress(addressInfo.getName());
+            if (bindings != null) {
+               for (Binding binding : bindings.getBindings()) {
+                  if (binding instanceof QueueBinding) {
+                     QueueControl coreQueueControl = (QueueControl) managementService.getResource(ResourceNames.QUEUE + binding.getUniqueName().toString());
```
jbertram: Instead of going through the `managementService` you can just call `binding.getQueue().pause(boolean)`. Be sure you pass the `persist` option through to the `Queue`.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AddressControlImpl.java
```diff
@@ -373,6 +372,86 @@ public String sendMessage(final Map<String, String> headers,
...
+         AuditLogger.resume(addressInfo);
+      }
+      checkStarted();
+
+      clearIO();
+      try {
+         if (addressInfo.isPaused()) {
+            addressInfo.resume();
+            if (persist) {
+               server.getStorageManager().storeAddressStatus(addressInfo.getId(), AddressStatus.RUNNING);
```
jbertram: This should be pushed into `AddressInfo` as described above. Also, resuming the address should simply delete the existing PAUSED record (if one exists) otherwise the bindings journal will grow unbounded with these records.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/AddressStatus.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.core.persistence;
+
+public enum AddressStatus {
```
jbertram: I'd recommend re-using `QueueStatus` and renaming it `AddressQueueStatus` since the two objects are essentially identical to each other.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/codec/AddressStatusEncoding.java
```diff
@@ -0,0 +1,86 @@
...
+
+   private AddressStatus status;
+
+   private long addressId;
+
+   private long id;
+
+   public AddressStatusEncoding(long addressId, AddressStatus status) {
+      this.status = status;
+      this.addressId = id;
```
jbertram: This assignment seems incorrect. It should almost certainly be `this.addressId = addressId;`.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/codec/AddressStatusEncoding.java
```diff
@@ -0,0 +1,86 @@
...
+import org.apache.activemq.artemis.core.persistence.AddressStatus;
+import org.apache.activemq.artemis.utils.DataConstants;
+
+public class AddressStatusEncoding implements EncodingSupport {
+
+   private AddressStatus status;
+
+   private long addressId;
+
+   private long id;
```
jbertram: I don't think you need this variable since it's not persisted.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
```diff
@@ -430,6 +430,10 @@ void slowConsumerDetected(String sessionID,
    @Message(id = 221080, value = "Deploying address {0} supporting {1}", format = Message.Format.MESSAGE_FORMAT)
    void deployAddress(String addressName, String routingTypes);
 
+   @LogMessage(level = Logger.Level.INFO)
+   @Message(id = 221081, value = "There is no queue with ID {0}, deleting record {1}", format = Message.Format.MESSAGE_FORMAT)
```
jbertram: Log message references "queue" when it should say "address".

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/AddressInfo.java
```diff
@@ -136,6 +138,18 @@ public void setBindingRemovedTimestamp(long bindingRemovedTimestamp) {
       this.bindingRemovedTimestamp = bindingRemovedTimestamp;
    }
 
+   public void pause() {
```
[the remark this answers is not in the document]
jbertram: IMO `pause()` and `resume()` should be `synchronized` like they are on `QueueImpl`.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AddressControlImpl.java
ehsavoie: ok

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AddressControlImpl.java
ehsavoie: That means that the storageManager will be in AddressInfo  right ?

jbertram: Yes. That appears to be the most straight-forward solution.

ehsavoie: the storageManager and the postOffice are not always available where AddressInfo are created. Hum are we sure about this storage feature ?

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AddressControlImpl.java
clebertsuconic: Moving the method to AddressInfo would make us introduce serious refactoring into Addressinfo
AddressNfo would on that case have a similar power to what Queue does. 

That is, we would need to bring the Server, StorageManager and other functionality that wouldn't belong to AddressInfo, unless we make it similar to Queue.. (we could even rename the class to Address)...

Why don't we keep this on AddressController for now.. with a // TODO block to move this to AddressInfo, and make proper refactoring on the Address as a separate task?

We would need to at least change the contructor to contain the Server (you can find everything else after the ACtiveMQServer).

clebertsuconic: A simple solution would be to create a setter on Addressinfo, and we call the setter right before we call the pause method.

We could throw an IllegalSTateException if the server is not informed. So the API would be... setServer(server);addressInfo.pause(true/false);

ehsavoie: ok

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
ehsavoie: ok I wrote some compatibility tests

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
```diff
@@ -135,4 +135,31 @@ String sendMessage(@Parameter(name = "headers", desc = "The headers to add to th
...
+   void pause(@Parameter(name = "persist", desc = "if true, the pause state will be persisted.") boolean persist) throws Exception;
+
+   /**
+    * Resume all the queues bound of this address.Messages are delivered again to all its bounded queues.
+    * @throws java.lang.Exception
+    */
+   @Operation(desc = "Resumes the queues bound to this address", impact = MBeanOperationInfo.ACTION)
+   void resume() throws Exception;
+
+   @Attribute(desc = "indicates if the queues bound to this address are paused")
```
michaelandrepearce: nit: captial "I" at start of sentance.

clebertsuconic:
I improved the tests and disocvered a few bugs. I will have to work on it tomorrow.

clebertsuconic:
don't merge this yet

clebertsuconic:
This was replaced by #2761

PR #2733, opened 2019-06-26
ARTEMIS-2366 artemis-service start fails

jbertram:


artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service.xml
```diff
@@ -33,7 +33,7 @@
    <logmode>roll</logmode>
 
    <executable>%JAVA_HOME%\bin\java.exe</executable>
-   <startargument>-Xbootclasspath/a:%ARTEMIS_HOME%\lib\${logmanager}</startargument>
+   <startargument>-Xbootclasspath/a:%ARTEMIS_HOME%\lib\${logmanager};%ARTEMIS_HOME%\lib\${wildfly-common}</startargument>
```
michaelandrepearce: Have we an updated version yet of jar dep yet? Before we make further changes to support it. Incase it doesnt come and we have to revert the other pr that added it.

jbertram: I don't understand what you're asking.

michaelandrepearce: Have we updated wildfly common yet

clebertsuconic: @michaelandrepearce even thought there's no real issue in there, the PR is being merged and a new release is being made soon.

https://github.com/wildfly/wildfly-common/pull/47

jbertram: FWIW, I just pushed a commit to move Wildfly Common to 1.5.2.Final. That should address the license question.

michaelandrepearce: Thanks

PR #2734, opened 2019-06-26
[ARTEMIS-2403]Consumer command, clientID is not saved during JMS exception

shailendra14k:
Issue:- https://issues.apache.org/jira/browse/ARTEMIS-2403

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
@@ -61,6 +61,7 @@ public class ConnectionAbstract extends InputAbstract {
       }
       JmsConnectionFactory cf = new JmsConnectionFactory(user, password, brokerURL);
       if (clientID != null) {
+         System.out.println("Consumer:: clientID = " + clientID);
          cf.setClientID(clientID);
       }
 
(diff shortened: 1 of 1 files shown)


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
```diff
@@ -61,6 +61,7 @@ private ConnectionFactory createAMQPConnectionFactory() {
       }
       JmsConnectionFactory cf = new JmsConnectionFactory(user, password, brokerURL);
       if (clientID != null) {
+         System.out.println("Consumer:: clientID = " + clientID);
```
jbertram: This appears to have been left by mistake. Can you confirm?

jbertram:
Couple of things:

1. This needs a test.
2. Your commit message doesn't need any square brackets.
3. Get rid of lines like "Indentation correction" from your commit message.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
clebertsuconic: I am merging without this line

clebertsuconic:
@jbertram I am not sure how we could test this though, as this would need input with the user.

This is about setting the clientID after the user/password had to be input. not sure how we would test it?

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
clebertsuconic: Hmmm... actually the Core version had the same print on it.

jbertram:
@clebertsuconic, yes I see your point. Nothing is coming to mind except maybe some hackery with `System.setIn()`. Let me play around for a bit, and I'll let you know.

clebertsuconic:
@jbertram I actually figured out how to make a test...
I'm merging with my commit adding the test.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
shailendra14k: yes, I added the sysout to have this sync with the Core version, this gets logged on the console along with the filer
```
[shsingh@shsingh bin]$ ./artemis consumer --destination topic://Multicast --message-count 10 --clientID test_Demo --durable --protocol amqp
Consumer:: filter = null
Consumer:: clientID = test_Demo
```

shailendra14k:
Thank you @jbertram for the feedback, will take care in the future. when using clean install locally no ceckstyle violations was reported. Finally, I found the maven goals from Travis

PR #2738, opened 2019-07-02
ARTEMIS-2407 Large message file not deleted if broker crashes between page deleted and pending large message written

wy96f:
#2549 fixed the problem in journal mode. It ensures pending large message is inserted following by appending delete message record. 
But in page mode, the problem still exists. We should delete page file after pending large message record is written into disk and before deleting page complete record.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
@@ -343,6 +343,7 @@ public final class Page implements Comparable<Page> {
          logger.debug("Deleting pageNr=" + pageId + " on store " + storeName);
       }
 
+      List<Long> largeMessageIds = new ArrayList<>();
       if (messages != null) {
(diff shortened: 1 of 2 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
```diff
@@ -352,11 +353,15 @@ public boolean delete(final PagedMessage[] messages) throws Exception {
...
                // or it may still being delivered even though it has been acked already
                lmsg.decrementDelayDeletionCount();
+               largeMessageIds.add(lmsg.getMessageID());
             }
          }
       }
 
       try {
+         if (!storageManager.getContext().waitCompletion(5000)) {
+            logger.info("Time out waiting for large messages" + largeMessageIds + ", may be not deleted if crashed");
```
michaelpearce-gain: please create a paramatised logger with a log code, ActiveMQServerLogger.LOGGER

PR #2740, opened 2019-07-05
ARTEMIS-2414 Sync before closing file in case data loss

wy96f:
When we open a new page, the page sync timer might sync the current page, i.e. the new page and after it responses are sent back to client. And the old page is closed(not synced), we may lose messages if server crashes before dirty page cache written back into disk.

When we switch journal file to the new file, TimedBuffer::flushBatch will be called in the old file. Then the file would be synced except in the case user sets "journalDatasync" to false. For safety we'd better sync journal file before closing it.


michaelpearce-gain:
@franz1981 looks like your concern is addressed right?

PR #2741, opened 2019-07-05
ARTEMIS-2415 JDBCJournal miss pending tasks during shutdown

brusdev:
Wait deleting large message tasks during stop.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
@@ -65,6 +65,7 @@ import org.apache.activemq.artemis.core.server.LargeServerMessage;
 import org.apache.activemq.artemis.core.server.files.FileStoreMonitor;
 import org.apache.activemq.artemis.journal.ActiveMQJournalBundle;
 import org.apache.activemq.artemis.utils.ExecutorFactory;
+import org.apache.activemq.artemis.utils.ReusableLatch;
 import org.apache.activemq.artemis.utils.critical.CriticalAnalyzer;
 import org.jboss.logging.Logger;
 
@@ -85,6 +86,8 @@ public class JournalStorageManager extends AbstractJournalStorageManager {
    protected String largeMessagesDirectory;
 
    protected ReplicationManager replicator;
+
+   protected final ReusableLatch deletingLargeMessageTasks = new ReusableLatch();
 
    public JournalStorageManager(final Configuration config,
                                 final CriticalAnalyzer analyzer,
@@ -267,6 +270,8 @@ public class JournalStorageManager extends AbstractJournalStorageManager {
          if (journalLoaded && idGenerator != null)
             idGenerator.persistCurrentID();
       }
+
+      deletingLargeMessageTasks.await(30, TimeUnit.SECONDS);
 
       final CountDownLatch latch = new CountDownLatch(1);
       try {
@@ -513,6 +518,7 @@ public class JournalStorageManager extends AbstractJournalStorageManager {
 
       };
 
+      deletingLargeMessageTasks.countUp();
       getContext(true).executeOnCompletion(new IOCallback() {
          @Override
          public void done() {
@@ -521,11 +527,13 @@ public class JournalStorageManager extends AbstractJournalStorageManager {
             } else {
                executor.execute(deleteAction);
             }
+
+            deletingLargeMessageTasks.countDown();
          }
 
          @Override
          public void onError(int errorCode, String errorMessage) {
-
+            deletingLargeMessageTasks.countDown();
          }
       });
    }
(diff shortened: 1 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -86,6 +87,8 @@
 
    protected ReplicationManager replicator;
 
+   protected final AtomicInteger deletingLargeMessageTasks = new AtomicInteger(0);
```
franz1981: I would put this logic just into `JDBCJournalStorageManager` while exposing here these callbacks:
```java
protected void beforeSubmitDeleteLargeMessageRequest() {
}

protected void afterDeleteLargeMessageCompleted() {
}

protected void afterDeleteLargeMessageError(int code, String errorMessage) {
}
```
The names could be improved, but the idea would be to let just JDBC part to care about this (unless we know that the storage manager need to handle this too), wdyt?

brusdev: TBH, I can't reproduce the issue on JournalStorageManager/JournalImpl but in theory they should have the same issue of JDBCJournalStorageManager/JDBCJournalImpl. I think the issue on JournalStorageManager/JournalImpl is hidden by better performances.

franz1981: @brusdev You can try with mockito to inject slow large messages deletes and see what happen...good point!

brusdev:
@franz1981 wdyt to also add a timeout for deletingLargeMessageTasks.wait?

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
clebertsuconic: Take a look on ReusableLatch...

I created ReusableLatch for situations like this.

you make it a final... and you away(0) until it gets to 0.

it's the same overhead as an AtomicINteger
same usability as a CountDownLatch.

artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManagerTest.java
```diff
@@ -212,4 +212,36 @@ public void testAddBytesToLargeMessageNotLeakingByteBuffer() throws Exception {
...
+      manager.deleteLargeMessageFile(largeMessage);
+      final CountDownLatch latch = new CountDownLatch(1);
+      Executors.newSingleThreadExecutor().execute(() -> {
+         try {
+            manager.stop();
+            latch.countDown();
+         } catch (Exception ignore) {
+         }
+      });
+      Thread.sleep(500);
```
clebertsuconic: Why you need a sleep? I have been beaten pretty badly by Thread.sleep() on the testsuite many many times. Can't we avoid it with a latch, a Wait.. anything else?

Unless this is testing timeouts, I wouldn't use a wait.

brusdev: I replaced the sleep with a latch and a wait.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
brusdev: I replaced AtomicInteger with ReusableLatch.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -268,6 +271,8 @@ private boolean internalStop(boolean ioCriticalError, boolean sendFailover) thro
             idGenerator.persistCurrentID();
       }
 
+      deletingLargeMessageTasks.await(30, TimeUnit.SECONDS);
```
michaelpearce-gain: -1 This isnt a good idea. On case of split brain with network pinger and network is detected split we need the isolated to shutdown to occur immediately without any blocking.

artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManagerTest.java
michaelpearce-gain: -1 we should have no blocking see other comment

michaelpearce-gain:
-1 adding any form of wait blocking or locking on shutdown

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
brusdev: I see your point but it normally should not block and 10 lines forward I see the same code:

> latch.await(30, TimeUnit.SECONDS);

michaelpearce-gain: that doesnt mean its something we would want to proliferate. Just because a window is broken, doesnt mean its ok to just go inside... avoid the broken glass...

michaelandrepearce:
@clebert why did you merge this?

michaelandrepearce:
Im tempted just to revert on master

michaelandrepearce:
This is dangerous blocking stop. Also it doesnt solve the problem if the broker is killed anyhow. As mentioned really the broker should clear up lose ends on start up. This way it doesnt block stop and also fixes the case if its killed

clebertsuconic:
I did not realize your -1, and I confused this wait with a similar change on FDs.

I'm reverting it.


@brusdev  can you address the issue with a new PR? a restart of the broker should remove any previous messages that were not deleted.

PR #2742, opened 2019-07-08
ARTEMIS-2409 Convert HornetQ field names in consumer/queue selector s…

TomasHofman:
…trings

https://issues.apache.org/jira/browse/ARTEMIS-2409

When HornetQ client connects to Artemis broker and creates a consumer with a selector defined, the field names in the selector string are not translated into correct internal field names on the broker side.


tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/jms/client/SelectorTranslatorTest.java
```diff
@@ -204,6 +204,37 @@ public void testParseJMSType() {
       checkNoSubstitute("JMSType");
    }
 
+   @Test
```
clebertsuconic: The proper test would been on the compatibility-tests


I have added it as part of my merge. just letting you know for future reference.

PR #2743, opened 2019-07-08
ARTEMIS-2418 Race conditions between cursor movement and page writing

wy96f:
The current code of CursorIterator::internalGetNext is a little complicated and not easy to follow logically.

And there are two race conditions between cursor movement and page writing:

1. Suppose the cursor's initial position is (p1,0) and page p1 is live with 0 msg. When we call internalGetNext(), the cursor moves to next page, i.e. position is (p2, 0) now(after [code](https://github.com/apache/activemq-artemis/blob/b25afc9efa04e3a05c1dbc7b7f4abf7546caad99/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java#L398)). Meanwhile(before [code](https://github.com/apache/activemq-artemis/blob/b25afc9efa04e3a05c1dbc7b7f4abf7546caad99/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java#L400)) p1 is filled with message m1 and p2 is created with m2. Then we retrieve m2 from p2 rather than m1.

2. Suppose the cursor's initial position is (p1, 1) and the page p1 is non live with 1 msg. When we call  internalGetNext(), the cursor moves to position(p2, 0) and get null page cache since p2 is not yet created(after [code](https://github.com/apache/activemq-artemis/blob/b25afc9efa04e3a05c1dbc7b7f4abf7546caad99/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java#L400)). Then(before next [loop](https://github.com/apache/activemq-artemis/blob/b25afc9efa04e3a05c1dbc7b7f4abf7546caad99/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java#L396)) p2 is created with m1 and p3 is created with m2 which means current writing page no. is p3. After the while loop the cursor moves to position(p3, 0) and we retrieve m2 from p3 rather than m1.

In both cases we would miss message m1 and subsequent page files won't be deleted unless the broker restarts.

### DIFF
--- a/tests/extra-tests/src/test/java/org/apache/activemq/artemis/tests/extras/byteman/RaceOnCursorIteratorTest.java
+++ b/tests/extra-tests/src/test/java/org/apache/activemq/artemis/tests/extras/byteman/RaceOnCursorIteratorTest.java
@@ -1,0 +1,208 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p/>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p/>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.activemq.artemis.tests.extras.byteman;
+
+import java.util.HashMap;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+
+import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
+import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.api.core.client.ClientSession;
+import org.apache.activemq.artemis.api.core.client.ClientSessionFactory;
+import org.apache.activemq.artemis.api.core.client.ServerLocator;
+import org.apache.activemq.artemis.core.config.Configuration;
+import org.apache.activemq.artemis.core.message.impl.CoreMessage;
+import org.apache.activemq.artemis.core.paging.PagingStore;
+import org.apache.activemq.artemis.core.paging.cursor.PagedReference;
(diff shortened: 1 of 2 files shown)


tests/extra-tests/src/test/java/org/apache/activemq/artemis/tests/extras/byteman/RaceOnCursorIteratorTest.java
```diff
@@ -0,0 +1,208 @@
...
+      skipLivePageCache = true;
+      // Simulate scenario #1 depicted in https://issues.apache.org/jira/browse/ARTEMIS-2418
+      PagedReference ref = queue.getPageSubscription().iterator().next();
+      assertTrue("first msg should not be " + (ref == null ? "null" : ref.getPagedMessage().getMessage().getMessageID()),
+                 ref == null || ref.getPagedMessage().getMessage().getMessageID() == 1);
+   }
+
+   @Test
+   @BMRules(
+      rules = {@BMRule(
```
clebertsuconic: Did you really need the byteman test? These test tend to be fragile. I have been personally avoiding them.


Can you explain in english what the test is doing? I coudl read the byteman rule. but I got confused on what is the race.

wy96f: Test testSkipLivePageCache simulates scenario 1. 
> Meanwhile p1 is filled with message m1 and p2 is created with m2.
This is achieved by calling raceAddLivePageCache after entry of PageSubscriptionImpl::moveNextPage.

Test testSkipNullPageCache simulates scenario 2. 
> When we call internalGetNext(), the cursor moves to position(p2, 0) and get null page cache since p2 is not yet created. Then p2 is created with m1 and p3 is created with m2 which means current writing page no. is p3
This is achieved by calling raceAddTwoCaches after the first PageSubscriptionImpl::moveNextPage and after PageSubscriptionImpl::getPageCache.

I have added the code line in the previous comment. Hope this helped to understand the race.

I wrote the byteman test to justify the race conditions. As you said, the test is tricky and fragile, so it's ok for me to remove it :)

clebertsuconic: I'm ok on merging this, however this will clash with your other changes on paging.


Would you rebase after I merged this?

wy96f: Yes, I would rebase, feel free to merge it :)

PR #2747, opened 2019-07-09
ARTEMIS-2420 Add DLA prefix configuration property, to allow dynamic DLA/DLQ creation

PiotrKlimczak:
Please note this is my 1st PR to Apache Artemis and 2h ago it was 1st time I was seeing its source code.
I tried to modify as little as possible and to fit existing coding practices.

I am open for comments and happy to change anything as per comments.

We need this one and a few more features in order for us to migrate to Artemis.

PS. Recreated PR as I suspect there is build stability issue in Travis with non related tests.


jbertram:
Couple of things:

- The commits should be squashed and rebased on the head of `master`.
- You don't need to re-create PRs to trigger another Travis run. Just do something like `git commit --amend` and then `git push -f origin <branch>`.
- The commit message should be more specific than just "implementation". You'll find commit message recommendations in the [Hacking Guide](https://github.com/apache/activemq-artemis/blob/master/docs/hacking-guide/en/maintainers.md#commit-messages).

That said, I have some broader reservations about this PR. It takes care of the configuration side of things (i.e. it allows you to configure a DLA prefix). However, the prefixed dead-letter addresses will still need to be *manually* created which kind of defeats the purpose of this feature which is to automate this process for auto-created addresses and queues.

In order to behave more like 5.x the broker would need to auto-create the dead-letter address with the same name as the original address with the proper prefix. However, an address alone is not enough to store messages so a queue will need to be created as well. This is where, in my opinion, things start to break down due to fundamental differences in the message routing configuration and semantics between 5.x and Artemis. The 5.x broker deals with JMS-centric resources (i.e. queues & topics) whereas Artemis deals with more general and granular resources (i.e. addresses, queues, & routing-types). If you're only thinking about the JMS use-case then it would be fine for the broker to create the properly named dead-letter address with a single anycast queue of the same name. However, in order to be appropriately flexible it will need to support multicast as well and it's plausible that the queue (if one is created) should not, in fact, have the same name as the address. You can see how the configuration options and the test scenarios start to expand to something much larger than what you've included in this PR.

At this point I think the best solution is the one which is already available. That is, statically creating one or more dead-letter addresses and corresponding queues and using those to aggregate the undeliverable messages from auto-created queues. When the broker sends a message to a dead-letter address it adds properties for the original address, queue, and message ID. These properties can be used by consumers to filter the messages and find what they need.

PiotrKlimczak:
Thanks for your comment @jbertram.
 
As said, I just really started with Artemis and I was hoping with a few hours of coding, I can add missing features avoiding changes in our software/processes which is JMS centric indeed.

I don't understand why:
''the prefixed dead-letter addresses will still need to be manually created which kind of defeats the purpose of this feature which is to automate this process for auto-created addresses and queues.''
I was trying to find out all usages of dead letter address config and I haven't seen it being used for any address/queue creation anywhere unless I missed it. Or should we create explicitly each address for DLAs/DLQs manually- which I haven't seen in docs, but again I might have missed it?

Regarding the existing solution in Artemis, my problem is that we process lots of messages across many queues and there are occasions where we are getting lots of errors.
Therefore having everything in a huge bag DLA/DLQ is not the most convenient way to support failures, where customers are obviously in rush and they have different priorities for different processes.
Also we have integrations, where queue names are dynamic and differs env to env, in which case having fixed config is not possible, while due to our internal standards we must deploy same artifacts to each customer environment.
I like the ActiveMQ's dynamic nature and it supports us very well, in fact, 95% of our queues are dynamically created because they don't need any customisations to default config- which itself is customised quite a lot.
Indeed technically we could use filters, however again this is not the most convenient way and for me, this doesn't sound like a solution, but rather as a workaround, which I can imagine would be acceptable in short term but not as a permanent solution - I doubt our support department would agree for that.

Personally, I am happy to invest more of my time to get the solution that would satisfy everybody.
But it looks like it might require a bit more guidance from your end.

Is there any chance you could provide more details on what is needed (like which class to focus at, what would be your acceptance criteria etc.) so we could both get it done?
I've seen people asking on some forums/blogs for this feature, so I can't be the only one on this planet who would possibly benefit from this feature :).

WDYT?

Thanks again for your time,

Piotr

jbertram:
> I was trying to find out all usages of dead letter address config and I haven't seen it being used for any address/queue creation anywhere unless I missed it.

That's exactly my point. The broker currently doesn't automatically create a dead-letter address or queue (prefixed or not) for any reason. They have to be created manually. Your PR doesn't change this which is why it doesn't actually solve the problem. Notice that your tests are explicitly calling `createQueue` for the DLA/DLQ which is being tested.

> Therefore having everything in a huge bag DLA/DLQ is not the most convenient way to support failures, where customers are obviously in rush and they have different priorities for different processes.

Nothing is forcing you to use a "huge bag DLA/DLQ" assuming you have some kind of address/queue naming scheme which follows a particular pattern. This is a common practice in messaging solutions. For example, all queues dealing with function X might be named `functionX.queue1`, `functionX.queue2`, etc. Then you could have an address-setting whose match is `functionX.#` and dead-letter-address is `DLA.functionX`, etc. I'm not saying you have to do it this way. I'm just trying to outline what options you have at this point and to make clear that having a 1:1 mapping between an address and dead-letter address is not necessarily required for handling undeliverable messages in a dynamic environment.

> Is there any chance you could provide more details on what is needed...

The current PR is just one piece of the puzzle. Off the top of my head, I think you'll need to add parameters to cover:

- dead-letter address routing type
- dead-letter queue name
- a way to indicate whether or not they dead-letter address/queue should be auto-created (which should be false by default to preserve the existing behavior)

Also:

- you'll need to actually create the dead-letter address & queue with the proper routing type
- give some thought to security; we may want the ability to prevent users from somehow tricking the broker into creating a bunch of extra queues

I'm sure there's more that I'm not thinking about now. Your solution can evolve through the review process.

michaelandrepearce:
Some other considerations:

Could/should this be per address or per queue, e.g. in a multicast address with multiple queues, you may want a dedicated dlq per each queue.

What about expiry queues, i assume any automated logic for dlq, we would want corresponding capability to expiry.

PiotrKlimczak:
@michaelandrepearce @jbertram 
Thanks for your hints.

> Your solution can evolve through the review process.

This is exactly what I was hoping for 👍 

Cheers

PiotrKlimczak:
@jbertram @

I have crafted some code and tests and it works, but I am not fully satisfied with it. 
My problem is that I am using delivery of failed message to dla which then will apply a routing to deliver the message to some queue. 
I think this is not fully right as I would not expect my message to land in different dlq than corresponding original queue but this is possible if we have any cast with two queues- unless my understanding of addressing model is incorrect. 
Therefore I think the expected behavior should be that a failed message should always land in the expected queue without applying any routing OR by enforcing some sort of filters on routing. 

WDYT?

clebertsuconic:
I have tried to look into this, and this is a bit messy to read now with the merge commit. 

Please execute the following:


```
git pull --rebase upstream master
git push origin <your-branch-name> -f>
```


after that, you should also:

```
git rebase -i <your first commit>
```

squash them all into a single commit.



I have tried doing this myself and there are a lot of conflicts.

clebertsuconic:
This is probably because of recent changes done on DLQ by Justin on QueueIMpl.

Can you rebase against hist changes please?

PiotrKlimczak:
Sure. Will start tomorrow after work as I am in BST zone. Also this pr doesn’t have latest changes as I have hit some doubt as per my previous post. 

Cheers

clebertsuconic:
Perhaps a fresh new PR on top of master would be eaiser? although push -f would be the same.. whatever is eaiser fo you

clebertsuconic:
@PiotrKlimczak I could fix your PR with some gitfu

checked out your PR:
git reset <your merging point>

Now I have all your changes flat

I'm adding a single commit, with you as an author.

I will replace this PR

clebertsuconic:
nice feature BTW!

clebertsuconic:
This is being replaced by #2760 

Please I would like your review, to make sure everything is there before I merge it.


I will be running a full testsuite before merging it.

PiotrKlimczak:
@clebertsuconic 
Thanks for helping.
I have few more changes regarding dynamic DLAs addressing some of @michaelandrepearce and @jbertram highlighted areas. Will push them to new branch.
Sorry, I run out of time yesterday as I can work on it only late evenings.
Will do my best to push them tonight.

Cheers
Piotr

PR #2749, opened 2019-07-14
ARTEMIS-2408 Too many opened FDs after server stops

brusdev:
Wait netty event loop group shutdown to avoid too many opened FDs after
server stops, when netty configuration is used. Clear server
activateCallbacks to avoid reactivation of previous nodeManager and
consequent FD leaks on restart. Fix LargeServerMessageImpl.copy to avoid
FD leaks.

cherry-picking commit 9617058ba0649af4eea15ce8793f86de827c4b7f
NO-JIRA adding check for open FD on the testsuite

cherry-picking commit 0facb7ddf4d3baa14a3add4290684aff7fd46053
NO-JIRA addressing connections leaks on integration tests

### DIFF
--- a/artemis-commons/src/test/java/org/apache/activemq/artemis/utils/Wait.java
+++ b/artemis-commons/src/test/java/org/apache/activemq/artemis/utils/Wait.java
@@ -1,0 +1,144 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.utils;
+
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Assert;
+
+/**
+ * Utility adapted from: org.apache.activemq.util.Wait
+ */
+public class Wait {
+
+
+   public static final long MAX_WAIT_MILLIS = 30 * 1000;
+   public static final int SLEEP_MILLIS = 1000;
+
+   public interface Condition {
+
+      boolean isSatisfied() throws Exception;
+   }
+
+   public interface LongCondition {
+      long getCount() throws Exception;
+   }
+
+   public interface IntCondition {
+      int getCount() throws Exception;
+   }
+
+   public static boolean waitFor(Condition condition) throws Exception {
+      return waitFor(condition, MAX_WAIT_MILLIS);
+   }
+
+
+   public static void assertEquals(long size, LongCondition condition) throws Exception {
+      assertEquals(size, condition, MAX_WAIT_MILLIS);
+   }
+
+   public static void assertEquals(long size, LongCondition condition, long timeout) throws Exception {
+      assertEquals(size, condition, timeout, SLEEP_MILLIS);
+   }
+
+   public static void assertEquals(Long size, LongCondition condition, long timeout, long sleepMillis) throws Exception {
+      boolean result = waitFor(() -> condition.getCount() == size, timeout, sleepMillis);
+
+      if (!result) {
+         Assert.fail(size + " != " + condition.getCount());
+      }
+   }
+
+
+   public static void assertEquals(int size, IntCondition condition) throws Exception {
+      assertEquals(size, condition, MAX_WAIT_MILLIS);
+   }
+
+   public static void assertEquals(int size, IntCondition condition, long timeout) throws Exception {
+      assertEquals(size, condition, timeout, SLEEP_MILLIS);
+   }
+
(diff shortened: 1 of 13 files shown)


artemis-commons/src/test/java/org/apache/activemq/artemis/utils/Wait.java
```diff
@@ -0,0 +1,144 @@
...
+package org.apache.activemq.artemis.utils;
+
+import java.util.concurrent.TimeUnit;
+
+import org.junit.Assert;
+
+/**
+ * Utility adapted from: org.apache.activemq.util.Wait
+ */
+public class Wait {
```
michaelandrepearce: Can we avoid duplicating this, and simply make it more shareable between modules.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/imported/MQTTTestSupport.java
```diff
@@ -57,6 +57,11 @@
 import org.apache.activemq.artemis.spi.core.remoting.Acceptor;
 import org.apache.activemq.artemis.spi.core.security.ActiveMQJAASSecurityManager;
 import org.apache.activemq.artemis.tests.util.ActiveMQTestBase;
+import org.fusesource.hawtdispatch.DispatchPriority;
```
michaelandrepearce: Is this fusesource actually maintained, i see last update 2016, do we really want to be building further on it???

Can we implement using alternative lib or our own for better maintenance?

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/NoFDBehind.java
```diff
@@ -0,0 +1,202 @@
...
+import org.apache.activemq.artemis.nativo.jlibaio.LibaioContext;
+import org.apache.activemq.artemis.utils.Wait;
+import org.jboss.logging.Logger;
+import org.junit.rules.TestWatcher;
+import org.junit.runner.Description;
+
+/**
+ * This is useful to make sure you won't have leaking threads between tests
+ */
+public class NoFDBehind extends TestWatcher {
```
michaelandrepearce: Whats FD?

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -171,6 +171,9 @@
    @ClassRule
    public static ThreadLeakCheckRule leakCheckRule = new ThreadLeakCheckRule();
 
+   @Rule
+   public NoFDBehind noFDBehind = new NoFDBehind(-1, 1000);
```
michaelandrepearce: Whats FD, i can guess due to context of this PR, but the point is, i shouldnt have to guess and in future without PR context it will be harder to guess, please avoid using acronym or shorthand, for others readability.

michaelandrepearce:
@brusdev i would also appreciate if we could finish off the changes in https://github.com/apache/activemq-artemis/pull/2727/files, worried thats taken a back seat, and is on the back of an already merged PR change

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/imported/MQTTTestSupport.java
brusdev: FuseMQTTClientProvider (already used by some MQTT tests) has a transitive dependency for org.fusesource.hawtdispatch: HawtDispatcher is statically allocated by DispatcherConfig when org.fusesource.mqtt.client.MQTT.blockingConnection is called and It create a new GlobalDispatchQueue.
I used org.fusesource.hawtdispatch only to shuthdown the global queue at the end of the test to avoid file descriptor leaks.

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/NoFDBehind.java
brusdev: I renamed NoFDBehind to NoProcessFilesBehind.

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
brusdev: I renamed noFDBehind to noProcessFilesBehind.

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/NoFDBehind.java
clebertsuconic: FileDescriptors

artemis-commons/src/test/java/org/apache/activemq/artemis/utils/Wait.java
clebertsuconic: @brusdev you did this on top of my previous commit, I had the Wait there as a hack, and I intended to fix this before merging.

We need to fix the Wait here.

(I think you should have kept my commit as a reference of my ownership though and changed it on top.. but no big deal)

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
clebertsuconic: @michaelandrepearce I would have thought that FD is a common usage for File Descriptors?

artemis-commons/src/test/java/org/apache/activemq/artemis/utils/Wait.java
clebertsuconic: this is a separate issue. It's about the artemis-junit project duplicating some stuff.

I'm removing the dependency on artemis-junit, and I'm dealing with this on a separate PR. I am merging this and i will send a new PR.

PR #2750, opened 2019-07-15
ARTEMIS-2399 Improve performance when there are a lot of subscribers

wy96f:
Based on discussion from [http://activemq.2283324.n4.nabble.com/Improve-paging-performance-when-there-are-lots-of-subscribers-td4751333.html](url) I decided to take the approach suggested by franz which is different than the one before. Now file offset is put in the PagePostion for reading directly from file and the extra index cache layer is removed.

### DIFF
--- a/artemis-server/pom.xml
+++ b/artemis-server/pom.xml
@@ -90,6 +90,10 @@
          <scope>test</scope>
       </dependency>
       <dependency>
+         <groupId>org.jctools</groupId>
+         <artifactId>jctools-core</artifactId>
+      </dependency>
+      <dependency>
          <groupId>io.netty</groupId>
          <artifactId>netty-buffer</artifactId>
       </dependency>
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageReader.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageReader.java
@@ -1,0 +1,141 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.activemq.artemis.core.paging.cursor.impl;
+
+import org.apache.activemq.artemis.core.paging.PagedMessage;
+import org.apache.activemq.artemis.core.paging.cursor.NonExistentPage;
+import org.apache.activemq.artemis.core.paging.cursor.PageCache;
+import org.apache.activemq.artemis.core.paging.cursor.PagePosition;
+import org.apache.activemq.artemis.core.paging.impl.Page;
+import org.jboss.logging.Logger;
+
+public class PageReader implements PageCache {
+   private static final Logger logger = Logger.getLogger(PageReader.class);
+
+   private final Page page;
+   private final int numberOfMessages;
+   private PagedMessage[] pagedMessages = null;
+
+   public PageReader(Page page, int numberOfMessages) {
+      this.page = page;
+      this.numberOfMessages = numberOfMessages;
+   }
+
+   @Override
+   public long getPageId() {
+      return page.getPageId();
+   }
+
+   @Override
+   public int getNumberOfMessages() {
+      return numberOfMessages;
+   }
+
+   @Override
+   public void setMessages(PagedMessage[] messages) {
+      this.pagedMessages = messages;
+   }
+
+   @Override
+   public synchronized PagedMessage[] getMessages() {
+      if (pagedMessages != null) {
+         return pagedMessages;
+      } else {
+         try {
+            openPage();
+            return page.read().toArray(new PagedMessage[numberOfMessages]);
+         } catch (Exception e) {
+            throw new RuntimeException(e.getMessage(), e);
+         } finally {
+            close();
+         }
+      }
+   }
+
+   @Override
+   public boolean isLive() {
+      return false;
+   }
+
+   /**
+    * @param pagePosition   page position
+    * @param throwException if {@code true} exception will be thrown when message number is beyond the page
+    * @param keepOpen       if {@code true} page file would keep open after reading message
+    * @return the paged message
+    */
+   public synchronized PagedMessage getMessage(PagePosition pagePosition, boolean throwException, boolean keepOpen) {
+      if (pagePosition.getMessageNr() >= getNumberOfMessages()) {
+         if (throwException) {
+            throw new NonExistentPage("Invalid messageNumber passed = " + pagePosition + " on " + this);
+         }
+         return null;
+      }
+
+      boolean previouslyClosed = true;
+      try {
+         previouslyClosed = openPage();
+         PagedMessage msg;
+         if (pagePosition.getFileOffset() != -1) {
+            msg = page.readMessage(pagePosition.getFileOffset(), pagePosition.getMessageNr(), pagePosition.getMessageNr());
+         } else {
+            if (logger.isTraceEnabled()) {
+               logger.trace("get message from pos " + pagePosition, new Exception("trace get message without file offset"));
+            }
+            msg = page.readMessage(0, 0, pagePosition.getMessageNr());
+         }
+         return msg;
+      } catch (Exception e) {
+         throw new RuntimeException(e.getMessage(), e);
+      } finally {
+         if (!keepOpen && previouslyClosed) {
+            close();
+         }
+      }
+   }
+
+   @Override
+   public synchronized PagedMessage getMessage(PagePosition pagePosition) {
+      return getMessage(pagePosition, false, false);
+   }
+
+   /**
+    * @return true if file was previously closed
+    * @throws Exception
+    */
+   boolean openPage() throws Exception {
+      if (!page.getFile().isOpen()) {
+         page.getFile().open();
+         return true;
+      }
+      return false;
+   }
+
+   @Override
+   public synchronized void close() {
+      try {
+         page.close(false, false);
+      } catch (Exception e) {
+         logger.warn("Closing page " + page.getPageId() + " occurs exception:", e);
+      }
+   }
+
+   @Override
+   public String toString() {
+      return "PageReader::page=" + getPageId() + " numberOfMessages = " + numberOfMessages;
+   }
+}
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -29,7 +29,6 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Set;
-import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
@@ -108,6 +107,7 @@ import org.apache.activemq.artemis.utils.collections.TypedProperties;
 import org.apache.activemq.artemis.utils.critical.CriticalComponentImpl;
 import org.apache.activemq.artemis.utils.critical.EmptyCriticalAnalyzer;
 import org.jboss.logging.Logger;
+import org.jctools.queues.MpscUnboundedArrayQueue;
 
 /**
  * Implementation of a Queue
@@ -176,7 +176,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    // Messages will first enter intermediateMessageReferences
    // Before they are added to messageReferences
    // This is to avoid locking the queue on the producer
-   private final ConcurrentLinkedQueue<MessageReference> intermediateMessageReferences = new ConcurrentLinkedQueue<>();
+   private final MpscUnboundedArrayQueue<MessageReference> intermediateMessageReferences = new MpscUnboundedArrayQueue<>(8192);
 
    // This is where messages are stored
    private final PriorityLinkedList<MessageReference> messageReferences = new PriorityLinkedListImpl<>(QueueImpl.NUM_PRIORITIES);
@@ -209,7 +209,8 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
    private long pauseStatusRecord = -1;
 
-   private static final int MAX_SCHEDULED_RUNNERS = 2;
+   private static final int MAX_SCHEDULED_RUNNERS = 1;
+   private static final int MAX_DEPAGE_NUM = MAX_DELIVERIES_IN_LOOP * MAX_SCHEDULED_RUNNERS;
 
    // We don't ever need more than two DeliverRunner on the executor's list
    // that is getting the worse scenario possible when one runner is almost finishing before the second started
@@ -324,13 +325,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
          out.println("consumer: " + holder.consumer.debug());
       }
 
-      for (MessageReference reference : intermediateMessageReferences) {
-         out.print("Intermediate reference:" + reference);
-      }
-
-      if (intermediateMessageReferences.isEmpty()) {
-         out.println("No intermediate references");
-      }
+      out.println("Intermediate reference size is " + intermediateMessageReferences.size());
 
       boolean foundRef = false;
 
@@ -1042,14 +1037,13 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    public void deliverAsync() {
       if (scheduledRunners.get() < MAX_SCHEDULED_RUNNERS) {
          scheduledRunners.incrementAndGet();
+         checkDepage();
          try {
             getExecutor().execute(deliverRunner);
          } catch (RejectedExecutionException ignored) {
             // no-op
             scheduledRunners.decrementAndGet();
          }
-
-         checkDepage();
       }
 
    }
@@ -2513,6 +2507,8 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
          logger.debug(this + " doing deliver. messageReferences=" + messageReferences.size());
       }
 
+      scheduledRunners.decrementAndGet();
+
       doInternalPoll();
 
       // Either the iterator is empty or the consumer is busy
@@ -2699,7 +2695,14 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
     * @return
     */
    private boolean needsDepage() {
-      return queueMemorySize.get() < pageSubscription.getPagingStore().getMaxSize();
+      return queueMemorySize.get() < pageSubscription.getPagingStore().getMaxSize() &&
+         /**
+          * In most cases, one depage round following by at most MAX_SCHEDULED_RUNNERS deliver round,
+          * thus we just need to read MAX_DELIVERIES_IN_LOOP * MAX_SCHEDULED_RUNNERS messages. If we read too much, the message reference
+          * maybe discarded by gc collector in response to memory demand and we need to read it again at
+          * a great cost when delivering.
+          */
+         intermediateMessageReferences.size() + messageReferences.size() < MAX_DEPAGE_NUM;
    }
 
    private SimpleString extractGroupID(MessageReference ref) {
@@ -3634,8 +3637,6 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
          } catch (Exception e) {
             ActiveMQServerLogger.LOGGER.errorDelivering(e);
-         } finally {
-            scheduledRunners.decrementAndGet();
          }
       }
    }
--- a/tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/jmx/JmxConnectionTest.java
+++ b/tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/jmx/JmxConnectionTest.java
@@ -27,8 +27,8 @@ import java.rmi.server.RemoteObject;
 import java.rmi.server.RemoteRef;
 
 import io.netty.util.internal.PlatformDependent;
-import io.netty.util.internal.shaded.org.jctools.util.UnsafeAccess;
 import org.apache.activemq.artemis.tests.smoke.common.SmokeTestBase;
+import org.jctools.util.UnsafeAccess;
 import org.junit.Assert;
 import org.junit.Assume;
 import org.junit.Before;
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
@@ -251,6 +251,9 @@ public final class ActiveMQDefaultConfiguration {
    // The max number of concurrent reads allowed on paging
    private static int DEFAULT_MAX_CONCURRENT_PAGE_IO = 5;
 
+   // If true the whole page would be read, otherwise just seek and read while getting message
+   private static boolean DEFAULT_READ_WHOLE_PAGE = false;
+
    // the directory to store the journal files in
    private static String DEFAULT_JOURNAL_DIR = "data/journal";
 
@@ -841,6 +844,11 @@ public final class ActiveMQDefaultConfiguration {
     */
    public static int getDefaultMaxConcurrentPageIo() {
       return DEFAULT_MAX_CONCURRENT_PAGE_IO;
+   }
+
+
+   public static boolean isDefaultReadWholePage() {
+      return DEFAULT_READ_WHOLE_PAGE;
    }
 
    /**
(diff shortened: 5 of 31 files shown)


wy96f:
@michaelandrepearce I deployed the test without MAX_DEPAGE_NUM limit during depage. The performance result as follows:
1. Running in 51MB size page and 1 page cache in the case of 100 multicast
queues. The consumer tps dropped to ~11000 compared to ~16000 in pr.
image
image
2. Running in 5MB size page and 100 page cache in the case of 100 multicast
queues.  The consumer tps dropped to ~10000 compared to ~15000 in pr.
image
image
3. Running in 51MB size page and 1 page cache in the case of 1 queue. Performance results similar.
image
image

Seems like paged message removed and re reading file resulted in the performance drop particularly in the case of heavy usage of memory(test #1,#2). Even if there was enough free memory and messages were not gced(test #3), the performance didn't make a big difference - instead memory footprint might be increased. Considering we've totally relied on the kernel page cache, it seems not to need spare messages in memory. I personally assume this is an import feature.

michaelpearce-gain:
@wy96f any chance for higher resolution images? (and whilst at it english titles)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -324,13 +325,7 @@ public String debug() {
...
       }
 
-      for (MessageReference reference : intermediateMessageReferences) {
-         out.print("Intermediate reference:" + reference);
-      }
-
-      if (intermediateMessageReferences.isEmpty()) {
-         out.println("No intermediate references");
-      }
+      out.println("Intermediate reference size is " + intermediateMessageReferences.size());
```
michaelpearce-gain: Was this change needed, seem this is some existing debug thats been useful historically to debug issues, any harm leaving as is?

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/jmx/JmxConnectionTest.java
```diff
@@ -27,8 +27,8 @@
 import java.rmi.server.RemoteRef;
 
 import io.netty.util.internal.PlatformDependent;
-import io.netty.util.internal.shaded.org.jctools.util.UnsafeAccess;
 import org.apache.activemq.artemis.tests.smoke.common.SmokeTestBase;
+import org.jctools.util.UnsafeAccess;
```
michaelpearce-gain: @franz1981 what occurs post java 8 here? Same with any internal Unsafe access in the MPSC Queue used

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageReader.java
```diff
@@ -0,0 +1,88 @@
...
+   public int getNumberOfMessages() {
+      return page.getNumberOfMessages();
+   }
+
+   @Override
+   public void setMessages(PagedMessage[] messages) {
+   }
+
+   @Override
+   public PagedMessage[] getMessages() {
```
michaelpearce-gain: would this not be dangerous? either we have code relying on this, and need to support it, or we should remove it from the interface/api so people dont code against it.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageReader.java
```diff
@@ -0,0 +1,88 @@
...
+      return page.getPageId();
+   }
+
+   @Override
+   public int getNumberOfMessages() {
+      return page.getNumberOfMessages();
+   }
+
+   @Override
+   public void setMessages(PagedMessage[] messages) {
```
michaelpearce-gain: ditto on comment below for this one too

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageReader.java
```diff
@@ -0,0 +1,88 @@
...
+         throw new RuntimeException(e.getMessage(), e);
+      }
+   }
+
+   @Override
+   public synchronized void close() {
+      try {
+         page.close(false);
+      } catch (Exception e) {
+         logger.warn("Closing page " + page.getPageId() + " occurs exception:", e);
```
michaelpearce-gain: should be a parametised error with an error code.

michaelpearce-gain:
Forgot to say, great stuff as always!! Personally prefer this impl to the other.

michaelpearce-gain:
re test 3, it might be the graphs size, but on squiting it does seems a significant perf differece, from what i can make 15k to 30k, which one is this PR and which one is master?

clebertsuconic:
I just came back from vacations.. I will need another day to look into this. Please wait for me, let me review it.. don't merge it yet :)

wy96f:
> re test 3, it might be the graphs size, but on squinting it does seems a significant perf differece, from what i can make 15k to 30k, which one is this PR and which one is master? And same with latency, seems the bottom is 18 where top is 40, that also significant change.

I was not clear enough. The original graphs were only for tests without MAX_DEPAGE_NUM limit, from top to bottom: consumer tps, consumer latency, producer tps, producer latency.  I've done the larger screenshot, put the performance results with/without MAX_DEPAGE_NUM limit together for comparison, please see the latest comment.

> As i mentioned in the email lists, i think what ever change is made, as this is soo core to foundation of the broker, it will be a big and risky change perf wise, it is key (a must for me) this is something we let people toggle on, and we leave the default behaviour as was so user dont get suprises and also de-risks, at least for a few releases, this way if there is a perf issue or usecase not covered, there wont be any impact to other users.

Sorry I forgot this. I guess this is per address setting, right?

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/jmx/JmxConnectionTest.java
franz1981: @michaelandrepearce I see that JCtools qs are fine to use Unsafe until Java 11 (included) for both perf and compatibility (I know is weird) reasons :+1:

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
wy96f: MpscUnboundedArrayQueue doesn't support iterator, so I had to remove it.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/jmx/JmxConnectionTest.java
michaelpearce-gain: you say until 11...does that mean after its knackered?

franz1981: We are working on a VarHandle version and we already allows different AtomicXXX versions to be used (with a small perf drop): probably java 12 is fine too, but I haven't tested it personally...
The problem will be when some Unsafe specific methods will be scattered around VarHandle/Other classes and will be officially deleted from Unsafe...

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
franz1981: I will take a note  to implement one on JCTools :+1:

michaelpearce-gain:
@wy96f i was thinking simply at just a broker level would be enough for me, just as a simple enable/kill switch (toggle) for the feature. If you did that you could even make it default on, just leave the option for people to disable should they need.

michaelpearce-gain:
@wy96f that makes it alot easier to understand the graphs :) thanks. +1 on perf improvements

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelpearce-gain: Do we need MpscUnboundedArrayQueue then in this PR to make it work? Im conscious of removing useful debugging features. Perf shouldnt be at the cost of feature / operational need.

Can always be re-added back once it supports? In a later PR perf improvement.

michaelpearce-gain: @franz1981 is this relatively trivial? Could always copy the class in temporarily and add the feature in artemis copy as an interim. (also would avoid a karaf issue with extra deps)

artemis-server/pom.xml
```diff
@@ -89,6 +89,10 @@
          <version>${project.version}</version>
          <scope>test</scope>
       </dependency>
+      <dependency>
```
michaelpearce-gain: Need to ensure this is updated in the OSGI / Karaf bundles config to ensure its bundled and loaded correctly, else those running in Karaf / OSGI they will get issue.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
franz1981: Not so easy I think...for bounded versions of qs yet (already supported in the current/maybe next release), but for unbounded is not that easy...
We are sure we really use iterator? or is iterator + clear at the end?

michaelpearce-gain: So debug like this only make it into code if it was needed at some point to debug an operational issue, as such i would suggest it was needed. My question is how much perf improvement is actually gained from this one particular change. I would expect the real perf was in the change in paging logic not the change of queue.

wy96f: MpscUnboundedArrayQueue might perform well in journal mode where there are many producers and one consumer. The main reason I used MpscUnboundedArrayQueue is I need to calculate size of messages both in intermediateMessageReferences and messageReferences, i.e.`intermediateMessageReferences.size() + messageReferences.size() < MAX_DEPAGE_NUM` in QueueImpl::needsDepage. However time complexity of size() for ConcurrentLinkedQueue is O(n), so I replaced it with MpscUnboundedArrayQueue(I did't find other queues fit for our scenario). I didn't think about issues like kafraf, debug by introducing this.

artemis-server/pom.xml
clebertsuconic: I had to add this on features on my branch to be able to run tests.

clebertsuconic:
Can you rebase this and fix the merging issue? I have merged your other PR and there's a clash now.


In the process, please add the new component into features as a dependency.


on features.xml

```xml
	<feature name="artemis-core" version="${pom.version}" description="ActiveMQ Artemis broker libraries">
		<feature>artemis-common</feature>
		<feature>scr</feature>
		<configfile finalname="etc/org.apache.activemq.artemis.cfg">mvn:org.apache.activemq/artemis-features/${pom.version}/cfg</configfile>
		<configfile finalname="etc/artemis.xml">mvn:org.apache.activemq/artemis-features/${pom.version}/xml/artemis</configfile>

		<bundle dependency="true">mvn:org.apache.geronimo.specs/geronimo-jms_2.0_spec/${geronimo.jms.2.spec.version}</bundle>
		<bundle dependency="true">mvn:org.apache.commons/commons-configuration2/${commons.config.version}</bundle>
		<bundle dependency="true">mvn:org.apache.commons/commons-text/1.6</bundle>
		<bundle dependency="true">mvn:org.apache.commons/commons-lang3/${commons.lang.version}</bundle>

                <!--  ********   you need to add this next line  please: -->
		<bundle dependency="true">mvn:org.jctools/jctools-core/${jctools.version}</bundle>
      
		<!-- Micrometer can't be included until it supports OSGi. It is currently an "optional" Maven dependency. -->
		<!--bundle dependency="true">mvn:io.micrometer/micrometer-core/${version.micrometer}</bundle-->

		<bundle>mvn:org.apache.activemq/activemq-artemis-native/${activemq-artemis-native-version}</bundle>
		<bundle>mvn:org.apache.activemq/artemis-server-osgi/${pom.version}</bundle>
	</feature>

```

artemis-server/pom.xml
clebertsuconic: You have to add:

``xml
<bundle dependency="true">mvn:org.jctools/jctools-core/${jctools.version}</bundle>
```
 as a 
dependency on artemis-core. see my other comment

wy96f:
Rebased and fixed to add deps and configuration

clebertsuconic:
I will need some time to review this last week.


Can you rebase again after I merged your change?

wy96f:
> I will need some time to review this last week.
> 
> Can you rebase again after I merged your change?

I'm working on PageReader per queue these days. Unluckily some problem was encountered. After investigation, I found it was related to #2740 which would sync files while closing. Given we put PageReader in PageSubscription meaning 100 times of file open/close for each page in our case(100 queues), although page file is only opened for reading, the sync operation does cause great performance loss - ~5000 msg/s received and 13000 msg/s sent now compared with 16000 msg/s received and 11000 msg/s sent of pr. Should I file a new jira and pr to fix the problem introuduced by #2740 or just fix it in this one?

michaelpearce-gain:
I would use separate pr and as we haven't released yet. Even use that original jira to fix #2740

clebertsuconic:
If it's an easy fix, you should fix it first on a separate PR.

wy96f:
I've submitted a new pr #2769 to fix it. After reviewed and merged, I would rebase and push the new code with PageReader per queue.

wy96f:
Pushed to change PageReader per queue.

The perf result is below:
1. Running in 51MB size page and 1 page cache in the case of 100 multicast
queues. 19000 msg/s received and 9500 msg/s sent vs. 16000 msg/s received and 11000 msg/s sent with previous PageReader cache. 
consumer tps:
per_queue_multi_consumer
producer tps:
per_queue_multi_producer
2. Running in 5MB size page and 100 page cache in the case of 100 multicast
queues. 16000msg/s received and 8200 msg/s sent vs. 15000 msg/s received and 8500 msg/s sent with previous PageReader cache. 
consumer tps:
per_queue_multi_consumer_small
producer tps:
per_queue_multi_producer_small
3. Running in 51MB size page and 1 page cache in the case of 1 queue. 16000 msg/s received and 29000 msg/s sent vs. 13500 msg/s received and 30000 msg/s sent with previous PageReader cache. 
consumer tps:
per_queue_single_consumer
producer tps:
per_queue_single_producer
Generally speaking, the consumer tps increase a little and producer tps reduce less, with total tps increasing by about 1000. 

It's to be noted that there was frequent file opening/closing in second test scenario(5MB size page and 100 page cache). I added some logging and saw two subsequent deliver was scheduled a few seconds after depage. In the case that message of PagedReference was cleared, calling getPriority/checkExpired in deliver() would trigger reading message. Due to the small page file, reading messages were more likely to hit page before the current cursor. Given PageReader before current cursor was closed, this would cause constantly page file open/close(log showed approximately 6000 times per second in average). In first test scenario(51MB size page and 1 page cache), approximately 700 times per second in average of file open/closing occurred.

wy96f:
I've added a new commit to reduce file open/close.

The log before this commit showed for each queue depage was followed by two deliver after a few seconds, something like this:

`2019-07-31 11:24:12,915 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] Queue Memory Size after depage on queue=sub86 is 4470916 with maxSize = 5242880. Depaged 2000 messages, pendingDelivery=0, intermediateMessageReferences= 2000, queueDelivering=984
2019-07-31 11:24:15,583 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub86, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=5d09e339-b32d-11e9-b1ac-fa163e6757d7], temp=false]@2c21ad1e doing deliver. messageReferences=0
2019-07-31 11:24:22,596 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub86, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=5d09e339-b32d-11e9-b1ac-fa163e6757d7], temp=false]@2c21ad1e doing deliver. messageReferences=2
2019-07-31 11:24:22,620 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] Queue Memory Size after depage on queue=sub86 is 4490706 with maxSize = 5242880. Depaged 2000 messages, pendingDelivery=0, intermediateMessageReferences= 2000, queueDelivering=990
2019-07-31 11:24:25,620 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub86, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=5d09e339-b32d-11e9-b1ac-fa163e6757d7], temp=false]@2c21ad1e doing deliver. messageReferences=0
2019-07-31 11:24:33,095 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub86, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=5d09e339-b32d-11e9-b1ac-fa163e6757d7], temp=false]@2c21ad1e doing deliver. messageReferences=2`
`2019-07-31 11:20:34,082 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] Queue Memory Size after depage on queue=sub49 is 4473860 with maxSize = 5242880. Depaged 2000 messages, pendingDelivery=0, intermediateMessageReferences= 2000, queueDelivering=982
2019-07-31 11:20:34,161 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub49, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=5d09e339-b32d-11e9-b1ac-fa163e6757d7], temp=false]@41a2d788 doing deliver. messageReferences=0
2019-07-31 11:20:44,535 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub49, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=5d09e339-b32d-11e9-b1ac-fa163e6757d7], temp=false]@41a2d788 doing deliver. messageReferences=2
2019-07-31 11:20:44,802 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] Queue Memory Size after depage on queue=sub49 is 4473010 with maxSize = 5242880. Depaged 2000 messages, pendingDelivery=0, intermediateMessageReferences= 2000, queueDelivering=971
2019-07-31 11:20:44,999 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub49, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=5d09e339-b32d-11e9-b1ac-fa163e6757d7], temp=false]@41a2d788 doing deliver. messageReferences=0
2019-07-31 11:20:55,230 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub49, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=5d09e339-b32d-11e9-b1ac-fa163e6757d7], temp=false]@41a2d788 doing deliver. messageReferences=2`

It is likely paged message is cleared when delivering a few seconds after depage causing opening/closing file to read message. After this commit, depage is immediately followed by deliver. The log was like this:

`2019-08-02 16:35:11,317 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] Queue Memory Size after depage on queue=sub72 is 2227083 with maxSize = 5242880. Depaged 509 messages, pendingDelivery=491, intermediateMessageReferences= 509, queueDelivering=387
2019-08-02 16:35:11,317 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub72, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=e0d3e34f-b4df-11e9-956a-fa163e6757d7], temp=false]@36c11804 doing deliver. messageReferences=491
2019-08-02 16:35:16,467 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] Queue Memory Size after depage on queue=sub72 is 2224399 with maxSize = 5242880. Depaged 1000 messages, pendingDelivery=0, intermediateMessageReferences= 1000, queueDelivering=268
2019-08-02 16:35:16,467 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub72, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=e0d3e34f-b4df-11e9-956a-fa163e6757d7], temp=false]@36c11804 doing deliver. messageReferences=0`
`2019-08-02 16:36:03,048 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] Queue Memory Size after depage on queue=sub53 is 2225927 with maxSize = 5242880. Depaged 1000 messages, pendingDelivery=0, intermediateMessageReferences= 1000, queueDelivering=1501
2019-08-02 16:36:03,048 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub53, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=e0d3e34f-b4df-11e9-956a-fa163e6757d7], temp=false]@2965382d doing deliver. messageReferences=0
2019-08-02 16:36:06,825 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] Queue Memory Size after depage on queue=sub53 is 2228961 with maxSize = 5242880. Depaged 1000 messages, pendingDelivery=0, intermediateMessageReferences= 1000, queueDelivering=1590
2019-08-02 16:36:06,825 DEBUG [org.apache.activemq.artemis.core.server.impl.QueueImpl] QueueImpl[name=sub53, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::serverUUID=e0d3e34f-b4df-11e9-956a-fa163e6757d7], temp=false]@2965382d doing deliver. messageReferences=0`

This new added commit reduces file open/close by about 70% - approximately 220 times per second open/close in first test scenario and 1770 times per second in second test scenario.

franz1981:
@wy96f A quick question: in your tests have you ever experienced a certain degree of unfairness when consumers are catching up with publishers while consuming a big backlog?  
I have some test with durable subscribers that shows publishers getting considerably slowed down while consumer(s) are catching up...

michaelandrepearce:
@franz1981 which test? I know when we page full stop producers slow down this is because of the extra io as goes to paging not memory. Just want to make sure its not that

franz1981:
@michaelandrepearce 
> which test?

Just a specific scenario I've been fallen into recently, to summarize it:
- a topic with many publishers sending durable messages
- many durable subscribers
- 1 of the subscribers disconnect for some time
- publishers start paging 
- after 10 GB  of paged data, the subscriber come back online again
- while it catches up, the other publishers/subscribers rate are seriously affected (approaching to 1/2 msg/sec)
- when it catches up, things come back normal for everyone

michaelandrepearce:
@franz1981 what version was that in. I know alot of locking changes to improve that had been made already to alleviate that. Do you get that still in latest?

franz1981:
@michaelandrepearce I'm using the latest master: my concern is that during `PagingStoreImpl::page` we perform both `storageManagerLock.readLock` and `PagingStoreImpl.lock.writeLock::lock`...so maybe there are consuming operations that prevent these locks to be released

michaelandrepearce:
are you sure its not underlying disk/infra issue, im very surprised by scenario you explain. Be great to have a recreator. Also if in master already its unrelated to the work here. I would like to avoid conflating issues

franz1981:
> Be great to have a recreator.

Will work on provide one soon

> Also if in master already its unrelated to the work here (eg this work causes no regression in perf). I would like to avoid conflating issues

That's a good point: although I've hoped that would be kinda related to the work done here, but I see that is a different scenario...will move the discuttion into the dev list, thanks :+1:

wy96f:
@franz1981 In our test consumers were falling behind producers all the time. I saw ~80GB page files left after test lasting 1 hour.

I'm not sure whether it's related to this pr. Considering 10GB page data produced, it's possible the pages are constantly dropping in an out while consumers are cathing up, resulting in performance drop.  Could you try this pr to check it? :)

franz1981:
@wy96f 

> it's possible the pages are constantly dropping in an out while consumers are cathing up, resulting in performance drop

Yep, but that would affect both consumer(s) and producers, not just producers: anyway I'm struggling to reproduce it reliably, so probably will dig into it much further before checking with this PR :+1: 

> In our test consumers were falling behind producers all the time. I saw ~80GB page files left after test lasting 1 hour.

And while catching up, the producers are not seriously decreasing their rate?
This PR has helped in your tests?

wy96f:
@franz1981 
> And while catching up, the producers are not seriously decreasing their rate?
This PR has helped in your tests?

This PR helped to improve consumer throughput while entering into page mode with lots of subscribers. I didn't see the producers rate seriously decreased, although producers slowed down bcs consumers used extra io.

michaelandrepearce:
@franz1981 @clebertsuconic @wy96f is this ready to merge? Would be good to get this body of work into the next release cutting this week

michaelandrepearce:
All tests ive run so far dont have any regressions

michaelandrepearce:
@franz1981 @wy96f @clebertsuconic as no further comments/feedback will merge this tomorrow. Last call :)

michaelandrepearce:
@wy96f can you squash commits so we can merge?

wy96f:
@michaelandrepearce Can you wait a moment? When the browse cursor iterators(like the one created by QueueControl::browse, browse only consumer, etc) are closed, the page opened by them is not closed. I would push another commit to fix this.

wy96f:
pushed commit to close page reader when cursor iterator is closed

michaelandrepearce:
@wy96f can you squash commits pls

wy96f:
@michaelandrepearce squashed commits and pushed. BTW, I added some code to avoid file leak in race conditions in PageSubscriptionImpl::internalGetNext() :

`           serverMessage = ((PageReader) cache).getMessage(retPos, false, true);
            PageCache previousPageCache = pageReaders.putIfAbsent(retPos.getPageNr(), (PageReader) cache);
            if (previousPageCache != null && previousPageCache != cache) {
               // Maybe other cursor iterators have added page reader, we have to close this one to avoid file leak
               cache.close();
            }`

PR #2751, opened 2019-07-15
ARTEMIS-2322: Expose Queue.getRate() data as JMX metric

dmvolod:


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
```diff
@@ -249,6 +249,21 @@ public long getMessageCount() {
...
    }
 
+   @Override
+   public float getProducedRate() {
+      if (AuditLogger.isEnabled()) {
+         AuditLogger.getProducedRate(queue);
+      }
+      checkStarted();
+
+      clearIO();
```
michaelpearce-gain: Shouldnt be doing this on an attribute will cause perf issues when metric systems scrape blocking deliveries

clebertsuconic: @michaelandrepearce I think it's harmless though.

PR #2754, opened 2019-07-16
NO-JIRA: Fix RAT plugin warnings for built .NET examples

dmvolod:


jbertram:
This would be easier to track if you just added these `<exclude>` elements it to the apache-rat-plugin configuration in the main `pom.xml`. That controls the exclusions for everything in the project.

dmvolod:
Thanks for review, @jbertram .
Fixed.

PR #2756, opened 2019-07-17
ARTEMIS-2408 Effectively reverting FileDescriptors check as testsuite…

clebertsuconic:
… is not completing

At this commit point NetworkFailureFailoverTest would hung indefinetly

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
@@ -32,7 +32,6 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


clebertsuconic:
@brusdev I had to revert partially your commit here, as the testsuite is not completing.

In particular I'm talking about NetworkFailureFailoverTest

I'm going to merge this as soon as the PR check finishes as I need to run tests. But Please fix the test and reapply this change.


Thanks

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
```diff
@@ -408,27 +407,15 @@ public void stop(final boolean criticalError) throws Exception {
...
+            acceptor.stop();
          } catch (Throwable t) {
             ActiveMQServerLogger.LOGGER.errorStoppingAcceptor(acceptor.getName());
          }
       }
 
-      for (Map.Entry<Acceptor, Future<?>> acceptorFuture : acceptorFutures.entrySet()) {
-         if (acceptorFuture.getValue() != null) {
-            try {
-               acceptorFuture.getValue().get();
```
clebertsuconic: @brusdev there's probably some issue with some Failover tests on getting the return here without a timeout.

Probably a timeout of 1 sec here would suffice, as we just want to make sure we actually close the acceptors.

brusdev: I totally agree on the need for a timeout. On my machine it takes 1.5 seconds in the worst case, wdyt about a total timeout of 3 seconds?

An example just to clarify what I mean:

>       long acceptorTimeout = System.currentTimeMillis() + 3000;
>       for (Map.Entry<Acceptor, Future<?>> acceptorFuture : acceptorFutures.entrySet()) {
>          if (acceptorFuture.getValue() != null) {
>             try {
>                acceptorFuture.getValue().get(Math.max(0, acceptorTimeout - System.currentTimeMillis()), TimeUnit.MILLISECONDS);
>             } catch (Throwable t) {
>                ActiveMQServerLogger.LOGGER.errorStoppingAcceptor(acceptorFuture.getKey().getName());
>             }
>          }
>       }
>

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -171,9 +171,6 @@
    @ClassRule
    public static ThreadLeakCheckRule leakCheckRule = new ThreadLeakCheckRule();
 
-   @Rule
-   public NoProcessFilesBehind noProcessFilesBehind = new NoProcessFilesBehind(-1, 1000);
```
brusdev: Do you take NoProcessFilesBehind off because it's too heavy?

PR #2757, opened 2019-07-17
ARTEMIS-2412 Allow CF configuration through JNDI references

boekhold:
Most connection related properties, like the SSL ones, currently
have to be encoded in the brokerURL. When configuring connections
purely through JNDI bindings, this is not always desireable.
This commit allows one to configure all properties included
in TransportConstants.ALLOWABLE_CONNECTOR_KEYS to be listed separately
in the JNDI bindings. These properties are then zipped into any
provided brokerURL. For properties that appear in both places,
the one specified separately in the JNDI bindings takes priority.

This commit should not affect any configuration other than those
configure through JNDIReferenceFactory.


clebertsuconic:
Nice!!! this is a lot easier now! thanks

PR #2759, opened 2019-07-17
ARTEMIS-2408 Too many opened FDs after server stops

brusdev:
Limit acceptors stop time.

reapplying after commit 20ca827d79282ca41b4f60f28e6b09333c3d82d5

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
@@ -79,6 +79,8 @@ public class RemotingServiceImpl implements RemotingService, ServerConnectionLif
 
    private static final Logger logger = Logger.getLogger(RemotingServiceImpl.class);
 
+   private static final int ACCEPTOR_STOP_TIMEOUT = 3000;
+
    // Attributes ----------------------------------------------------
 
    private volatile boolean started = false;
@@ -407,13 +409,16 @@ public class RemotingServiceImpl implements RemotingService, ServerConnectionLif
          conn.disconnect(criticalError);
       }
 
+      CountDownLatch acceptorCountDownLatch = new CountDownLatch(acceptors.size());
       for (Acceptor acceptor : acceptors.values()) {
          try {
-            acceptor.stop();
+            acceptor.asyncStop(() -> acceptorCountDownLatch.countDown());
          } catch (Throwable t) {
             ActiveMQServerLogger.LOGGER.errorStoppingAcceptor(acceptor.getName());
          }
       }
+      //In some cases an acceptor stopping could be locked ie NettyAcceptor stopping could be locked by a network failure.
+      acceptorCountDownLatch.await(ACCEPTOR_STOP_TIMEOUT, TimeUnit.MILLISECONDS);
 
       acceptors.clear();
 
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
@@ -1269,6 +1269,8 @@ public class ActiveMQServerImpl implements ActiveMQServer {
          this.analyzer = null;
       }
 
+      activateCallbacks.clear();
+
       if (identity != null) {
          ActiveMQServerLogger.LOGGER.serverStopped("identity=" + identity + ",version=" + getVersion().getFullVersion(), tempNodeID, getUptime());
       } else {
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
@@ -171,6 +171,9 @@ public abstract class ActiveMQTestBase extends Assert {
    @ClassRule
    public static ThreadLeakCheckRule leakCheckRule = new ThreadLeakCheckRule();
 
+   @Rule
+   public NoProcessFilesBehind noProcessFilesBehind = new NoProcessFilesBehind(-1, 1000);
+
    /** We should not under any circunstance create data outside of ./target
     *  if you have a test failing because because of this rule for any reason,
     *  even if you use afterClass events, move the test to ./target and always cleanup after
@@ -276,6 +279,7 @@ public abstract class ActiveMQTestBase extends Assert {
 
    @After
    public void tearDown() throws Exception {
+      noProcessFilesBehind.tearDown();
       closeAllSessionFactories();
       closeAllServerLocatorsFactories();
 
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQComponent.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQComponent.java
@@ -16,26 +16,16 @@
  */
 package org.apache.activemq.artemis.core.server;
 
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.Future;
-
 public interface ActiveMQComponent {
 
    void start() throws Exception;
 
    void stop() throws Exception;
 
-   default Future<?> asyncStop() {
-      CompletableFuture<?> future = new CompletableFuture<>();
+   default void asyncStop(Runnable callback) throws Exception {
+      stop();
 
-      try {
-         stop();
-         future.complete(null);
-      } catch (Throwable t) {
-         future.completeExceptionally(t);
-      }
-
-      return future;
+      callback.run();
    }
 
    boolean isStarted();
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
@@ -36,8 +36,8 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
-import java.util.concurrent.Future;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -221,8 +221,6 @@ public class NettyAcceptor extends AbstractAcceptor {
    final AtomicBoolean warningPrinted = new AtomicBoolean(false);
 
    final Executor failureExecutor;
-
-   private Future<?> asyncStopFuture = null;
 
    public NettyAcceptor(final String name,
                         final ClusterConnection clusterConnection,
@@ -649,15 +647,18 @@ public class NettyAcceptor extends AbstractAcceptor {
    }
 
    @Override
-   public java.util.concurrent.Future<?> asyncStop() {
-      stop();
+   public void stop() throws Exception {
+      CountDownLatch latch = new CountDownLatch(1);
 
-      return asyncStopFuture;
+      asyncStop(() -> latch.countDown());
+
+      latch.await();
    }
 
    @Override
-   public synchronized void stop() {
+   public synchronized void asyncStop(Runnable callback) {
       if (channelClazz == null) {
+         callback.run();
          return;
       }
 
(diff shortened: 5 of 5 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
```diff
@@ -407,15 +408,28 @@ public void stop(final boolean criticalError) throws Exception {
...
          } catch (Throwable t) {
             ActiveMQServerLogger.LOGGER.errorStoppingAcceptor(acceptor.getName());
          }
       }
 
+      long acceptorTimeout = System.currentTimeMillis() + 3000;
+      for (Map.Entry<Acceptor, Future<?>> acceptorFuture : acceptorFutures.entrySet()) {
+         if (acceptorFuture.getValue() != null) {
+            try {
+               acceptorFuture.getValue().get(Math.max(0, acceptorTimeout - System.currentTimeMillis()), TimeUnit.MILLISECONDS);
```
clebertsuconic: Why not simply call get(3000) every time here?


Can you declare the constant as a private final field... 

Also, explain why there is a timeout here with a comment

brusdev: In a scenario with 2 acceptors if the first takes 2 seconds and the second takes 4 seconds the loop takes 4 seconds.

michaelpearce-gain: What happens to the acceptors that dont close in time? Are the port bindings just left hanging?

clebertsuconic: This is actually a fix for NetworkFailureFailoverTest. So no need for a test.


(@brusdev  please mention this is fixing networkFailureFailoverTest after reapplying the commit)


NetworkFailureFailoverTest was actually added to prevent a situation where a network card gone wouldn't prevent a server from stopping.

Instead of hunging during a shutdown the server will wait on a timeout, and the process will die and the OS will take care of cleaning up resources on that failure case.


On a regular scenario these will run really fast and never take very long.

clebertsuconic: @michaelandrepearce also I reverted a previous fix as networkFailurefailoverTest was broken, and this is fixing it.

michaelpearce-gain: I assume though that test case existed already right? And was previously passing.....

So why is this needed, what broke it, and why. Very cautious in areas like this. Sorry if im sucking eggs

clebertsuconic: @michaelandrepearce Domenico added a rule to validate if File Descriptors are temporarily leaking. The fix was to block until the acceptor was stopped.

That broke NetworkFailureFailoverTest, as the block was infinite

Now the fix is apply a timeout on the stop, which will fix both cases.

michaelpearce-gain: Has a physical test been done to prove the network one still works? If i remember we found it the hard way a year or two back and it took a physcial retest to prove it, i realise the test tried to simulate but the times a phsycial test has found issue where simulated didnt.

Just i am super cautious here, as these are failure scenarious already been bitten by

clebertsuconic: I'm not concerned as there is no semantic change if a timeout is applied.

I'm actually glad the test caught this situation right on the testsuite. It would been a regression if it wasn't for that test.


I will make a manual test before merging it. Although this test is automating how I used to test this manually (by adding and removing a network card through ifconfig)

clebertsuconic: This has to be merged anyways.. as I know it is broken now in master.

I'm just working thorugh a better way to do these timeouts.

michaelpearce-gain: im glad too :) , but you can understand my nervousness right :)

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
```diff
@@ -407,15 +410,28 @@ public void stop(final boolean criticalError) throws Exception {
          conn.disconnect(criticalError);
       }
 
+      Map<Acceptor, Future<?>> acceptorFutures = new HashMap<>();
       for (Acceptor acceptor : acceptors.values()) {
          try {
-            acceptor.stop();
+            //In some cases an acceptor stopping could be locked ie NettyAcceptor stopping could be locked by a network failure.
+            acceptorFutures.put(acceptor, acceptor.asyncStop());
```
clebertsuconic: it would be a lot simpler if instead of returning a future, you passed in a Runnable or Future

You could have the following code:

```java
// Instead of a Map<Acceptor, Future>
CountDownLatch latch = new CountDownLatch(acceptors.size());

for (Acceptor accetor: acceptors.values()) {
    acceptor.asyncStop(latch::countDown);
}
```

and the whole loop on each Future would be replaced by:
```java
   latch.await();
```

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -276,6 +279,7 @@ public ActiveMQTestBase() {
 
    @After
    public void tearDown() throws Exception {
+      noProcessFilesBehind.tearDown();
```
clebertsuconic: Why do you need a tearDown here? The Rule could clear itself.

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
```diff
@@ -407,13 +409,16 @@ public void stop(final boolean criticalError) throws Exception {
...
       for (Acceptor acceptor : acceptors.values()) {
          try {
-            acceptor.stop();
+            acceptor.asyncStop(() -> acceptorCountDownLatch.countDown());
          } catch (Throwable t) {
             ActiveMQServerLogger.LOGGER.errorStoppingAcceptor(acceptor.getName());
          }
       }
+      //In some cases an acceptor stopping could be locked ie NettyAcceptor stopping could be locked by a network failure.
+      acceptorCountDownLatch.await(ACCEPTOR_STOP_TIMEOUT, TimeUnit.MILLISECONDS);
```
clebertsuconic: a lot easier to read :)

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
brusdev: Some methods executed after tests ie shutdownDerby open new files (at the first execution) so I need a method called before to filter that files.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -1269,6 +1269,8 @@ void stop(boolean failoverOnServerShutdown, final boolean criticalIOError, boole
          this.analyzer = null;
       }
 
+      activateCallbacks.clear();
```
clebertsuconic: do you really need this?

There are a lot of things around activate callbacks. I don't see any testing around this.. so why is this needed?

clebertsuconic: For example, Wildfly will inject activations and a server may stopped and started.

unless you put really good though and testing on this.. you shouldn't really touch this area.

It does not seem related.

clebertsuconic: I will merge without this line.

If you need to clear activateCallbacks, please let me know why and we can review this!

clebertsuconic:
@brusdev I'm highly considering reverting this.

The testsuite is now taking 4 hours to finish. It was already high at 2 Hours and 30 minutes. 4 Hours is really too much.


There are other failures as well.

clebertsuconic:
It is taking too long because you are calling lsof over exec on every test executed.

That's too much!

I will revert that part only.

clebertsuconic:
Hmnmm... I don't know what's going on.. that part is disabled.

michaelpearce-gain:
Maybe the original PR should be revert also. I realise its trying to fix something. And the solution as a whole is re looked at.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
brusdev: Each time the server starts, it register a new activateCallback for FileLockNodeManager and for ManagementServiceImpl. After n stop and start without  `activateCallbacks.clear();` the server has n   activateCallback for FileLockNodeManager  and n  activateCallback for ManagementServiceImpl. After second start setUpServerLockFile is also called on previous activateCallbacks causing file leaks.

michaelpearce-gain: So the fix should be in the component holding the external resource then. Not here. If a piece of code opens an external resource the logic and lifecycle of that must be managed by it properly. Seems this is just hiding the problem that a component is not managing its rource correctly and simply this is hiding that issur

michaelpearce-gain: E.g. im worried we are treating symptoms, rather the the underlying cause

brusdev: @michaelandrepearce I see your point, we could replace `activateCallbacks.clear();` with a call to unregisterActivateCallback on LiveOnlyActivation.close for FileLockNodeManager and a call to unregisterActivateCallback on ManagementServiceImpl.stop for ManagementServiceImpl, wdyt?

michaelpearce-gain: still thats a change here, still not solving the underlying issue is dealing with the component that is holding the FD, it internally should be handling this, NOT ActiveMQServerImpl

brusdev: After start, stop and another start we have two activateCallbacks for FileLockNodeManager one from first start and one from second start. During second server start nodeManager.start/setUpServerLockFile is called on the new FileLockNodeManager but setLive/writeFileLockStatus is also called on previous FileLockNodeManager.
When setLive/writeFileLockStatus is called even on previous FileLockNodeManager the channel is closed (this is right, it is closed on the previous stop) so setUpServerLockFile is called an this is the file leak cause. So I think that who register an activateCallback for FileLockNodeManager on start has to unregister that activateCallback on stop, wdyt?

michaelpearce-gain: Yes if you add a callback, and its no longer required, the thing/component that does that has the responsibility to remove it, when its not required.

michaelpearce-gain: Whats the status on this i see this PR was merged is this discussion being addressed separately in a new PR?

clebertsuconic: @michaelandrepearce the callback.clear has been removed while I merged. @brusdev will address the other test that needs a cleanup.

clebertsuconic: I see now. .the issue is that a stop(shutdown=true) should clear the activateCallbacks. 

this is not really an issue when running on a real server, but if you call shutdown=true, it means the VM is going down and the process should cleanup anything left behind.

As the callback is leaking, if you reused the server after a stop(shutdown=true) on the testsuite, you will have the activation (on this case the one holding the lock) being called an extra time, what would then cause the file leak.


I will send another PR with checking the shutdown clause and add a test to avoid a future leak.

michaelpearce-gain: Query:

Doesnt network checker shutdown and restart in same vm? E.g. would that be liable to leaking?

PR #2762, opened 2019-07-19
ARTEMIS-2408 Changing FileDescriptors rule as classRule

clebertsuconic:


clebertsuconic:
@brusdev I'm running tests with these changes in. We can't really extend the testsuite for 1.5 on a fast box, and 3 hours more on a slow box (that's what I'm seeing when I run the testsuite now).

lets see how it goes with a classRule instead. perhaps it's what we need.

brusdev:
@clebertsuconic the lsof check is very useful to investigate on files leaks so it could be kept only if maxVariance > 0 or it could be moved in NoFilesBehind (when filesToCheck = {}) or it could be moved in a new rule to enable only if necessary.

clebertsuconic:
I'm a bit concerned about the lsof check, as being dead code.. the day you need it it hasn't been used forever and nobody knows what it does.

I changed it to show all the opened files in case there is a leak.. you as a developer would be able to figure out by looking at the list.

I also changed the timeout on shutdown. The issue is that once you added the block shutdown the tests were taking a lot of time to process those. with this at 0, I could get back to previous times.

If you could please review this?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/TransportConstants.java
```diff
@@ -281,6 +284,35 @@
...
+    *  When running on a test suite, we need it to be 0, You should see a property on the main pom.xml.
+    */
+   public static final int DEFAULT_QUIET_PERIOD = parseDefaultVariable("DEFAULT_QUIET_PERIOD", 100);
+
+   public static final String SHUTDOWN_TIMEOUT = "shutdownTimeout";
+
+   /** We let this to be defined as a System Variable, as we need a different timeout over our testsuite.
+    *  When running on a real server, this is the default we want.
+    *  When running on a test suite, we need it to be 0, You should see a property on the main pom.xml */
+   public static final int DEFAULT_SHUTDOWN_TIMEOUT = parseDefaultVariable("DEFAULT_SHUTDOWN_TIMEOUT", 30_000);
```
clebertsuconic: oops.. this is supposed to be 3 seconds

clebertsuconic: fixed

brusdev:
LGTM +1

PR #2767, opened 2019-07-24
ARTEMIS-2431: Test to reproduce the pipelined open auth issue with AMQP

jdanekrh:


clebertsuconic:
I'm not sure it's a good idea to add a failing test without a fix already. You could just open a JIRA, and add the test with a branch reference, a patch.. or anything like that.

jdanekrh:
I'll close the PR, then, and leave just the branch with currently failing test.

PR #2769, opened 2019-07-26
ARTEMIS-2414 Sync before closing file in case data loss

wy96f:
#2740 added sync before closing file in case data loss. Sync would also be called while closing read-only files. This would result in perf penalty in the case of files opened and closed constantly, such as large message deliver/delete, page file read by many queues, etc. Don't sync if file is only opened for reading.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
@@ -65,7 +65,7 @@ public final class LargeServerMessageImpl extends CoreMessage implements LargeSe
       ActiveMQBuffer buffer = coreMessage.getReadOnlyBodyBuffer();
       final int readableBytes = buffer.readableBytes();
       lsm.addBytes(buffer);
-      lsm.releaseResources();
+      lsm.releaseResources(true);
       lsm.putLongProperty(Message.HDR_LARGE_BODY_SIZE, readableBytes);
       return lsm;
    }
@@ -274,7 +274,7 @@ public final class LargeServerMessageImpl extends CoreMessage implements LargeSe
          throw new RuntimeException(e);
       } finally {
          try {
-            file.close();
+            file.close(false);
          } catch (Exception ignored) {
          }
       }
@@ -298,7 +298,7 @@ public final class LargeServerMessageImpl extends CoreMessage implements LargeSe
       } finally {
          if (closeFile) {
             try {
-               file.close();
+               file.close(false);
             } catch (Exception ignored) {
             }
          }
@@ -313,7 +313,7 @@ public final class LargeServerMessageImpl extends CoreMessage implements LargeSe
    @Override
    public synchronized void deleteFile() throws Exception {
       validateFile();
-      releaseResources();
+      releaseResources(false);
       storageManager.deleteLargeMessageFile(this);
    }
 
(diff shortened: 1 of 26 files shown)


artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/SequentialFile.java
```diff
@@ -53,7 +52,7 @@
 
    void fill(int size) throws Exception;
 
-   void delete() throws IOException, InterruptedException, ActiveMQException;
+   void delete() throws Exception;
```
michaelpearce-gain: Can you revert this and keep the explicit exceptions.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/mapped/TimedSequentialFile.java
```diff
@@ -94,7 +93,7 @@ public void fill(int size) throws Exception {
    }
 
    @Override
-   public void delete() throws IOException, InterruptedException, ActiveMQException {
+   public void delete() throws Exception {
```
michaelpearce-gain: Revert this and keep explicit

wy96f:
Pushed to keep explicit :)

clebertsuconic:
this is merged but i will open the PR to make sure we have some discussions about it, ok?

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
```diff
@@ -298,7 +298,7 @@ public int getBodyBufferSize() {
       } finally {
          if (closeFile) {
             try {
-               file.close();
+               file.close(false);
```
clebertsuconic: these will cause the channel to not flush, correct?

Why you making this change?


The thing is, I;ve run the testsuite after these changes and there are files leaking.

I'm running it one more time but this is the only commit after it started to fail. And I'm a bit confused about this after I merged and looking at the tests.

clebertsuconic: @wy96f  ^^^

wy96f: The getBodyBufferSize only retrieves the fileSize after opening file, and I assume this is only for reading. For this reason, I close it without flushing. I want to ask how you found files leaking? Are the leaking files large message ones?

clebertsuconic: when I run the whole testsuite, the NoProcessFilesBehind is reporting too many files open.

clebertsuconic:
Did you see .sync actually taking time on a profiling?

The information I had from a few kernel guys is that a sync call when there's no sync needed would be moot and not cause any issues. Although I can see some barrier between Java and Native layer here.

is that what you saw? a barrier between native and Java, or an actual sync ocurring when nothing was needed?

clebertsuconic:
this is being followed by #2774

wy96f:
> @wy96f I figured out: AIOSequentialFile::close(sync) was not closing the file if sync=false. I am fixing it. it's a previous issue that became unveiled when you changed the close.
> 
> BTW: Did you see .sync actually taking time on a profiling?
> 
> The information I had from a few kernel guys is that a sync call when there's no sync needed would be moot and not cause any issues. Although I can see some barrier between Java and Native layer here.
> 
> is that what you saw? a barrier between native and Java, or an actual sync ocurring when nothing was needed?

It might not be the actual sync. A few sync calls would not incur perf penalty, but a great many would.
The thing is, during development testing of #2750 , I found message of PagedReference was cleared by gc while delivering, which caused reading page file to retrieve message. There was no problem in PageReader cache version for the cached PageReader is always open. However in PageReader per queue version, when we moved to next page meaning the last PageReader was closed and removed, we had to reopen PageReader to retrieve message. This resulted in lots of sync when opening and closing, and perf penalty was emerging.

PR #2771, opened 2019-07-26
ARTEMIS-2434 Don't lock ServerConsumerImpl for long period of time

franz1981:


franz1981:
I have yet to verify that it works good on our CI: the only downside of this solution is that the lock won't be fairly attempted on `tryLock` if compared with an explicit timed `lock`.
I don't think this would be a problem, given that `setStarted` and `setTransferring` are slow path operations (ie rollback and connection transfert) and will be performend at least 300 times (every 100 ms for 30 seconds), the chance that they won't enter the lock is very small.

clebertsuconic:
If you can change the commit message to reflect what you are really doing here..

you are retrying the locks in a loop... I don't know what message though :) but can  you find something more in line to what you're doing?

PR #2772, opened 2019-07-26
 ARTEMIS-2423 Improving Consumer/Queue Delivery lock

clebertsuconic:


clebertsuconic:
@franz1981 this is a better fix, no locks at all.

The lock was added on hornetq when netty was blocked. We don't need that any longer.

PR #2773, opened 2019-07-27
 ARTEMIS-2434 Improving Consumer/Queue Delivery Lock

clebertsuconic:


clebertsuconic:
@franz1981 look at this one.. it's way less invasive than the previous one.

No more locks, and logic and current semantics still the same.

jbertram:
The JIRA referenced on the commit message is incorrect. It should be ARTEMIS-2434.

clebertsuconic:
@jbertram nice catch.. thanks

PR #2775, opened 2019-07-30
 ARTEMIS-2435 Configuration on device-block-size through CLI / broker.xml

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/utils/XMLConfigurationUtil.java
```diff
@@ -44,7 +44,6 @@ public static final Double getDouble(final Element e,
          validator.validate(name, val);
          return val;
       } else {
-         validator.validate(name, def);
```
michaelpearce-gain: ???

clebertsuconic: This is about validating the default value.

If the user didn't make a selection, there's no need to validate the default value of a selection.


I have been through specifically a case where device-block-size (a parameter that I will add) would be null if no selection, but in case there's a selection it must abide by the rules.

The default value == null would be allowed. 

There's no need to validate the default return value.

clebertsuconic: @michaelandrepearce any objection?

This should not be an issue, but I wanted to make sure.

clebertsuconic: @michaelandrepearce I need this fix in as I'm doing tests on a device that has block-size != 4096.

if you have any objections I can send further changes.

michaelpearce-gain: Surely though it would be valid for someone to set a default value, as such whilst a bit backwards, this would actually be (from unit testing) validate that the validators set allow for the default value. If they dont and they fail, then the validator isnt correct.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -266,10 +266,12 @@
    @Option(name = "--no-fsync", description = "Disable usage of fdatasync (channel.force(false) from java nio) on the journal")
    private boolean noJournalSync;
 
+   @Option(name = "--device-block-size", description = "The block size by the device, default at 4096.")
```
michaelpearce-gain: should this be journal-device-block-size.

e.g. what occurs in future where want to tune the others, and they maybe using other disks with other block sizes, might be best to namespace this now, so in future doesnt cause any issue

artemis-server/src/main/java/org/apache/activemq/artemis/utils/XMLConfigurationUtil.java
clebertsuconic: @michaelandrepearce I'm confused if you want me to just add a test or add the validator back?

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
clebertsuconic: ok.. I will send another PR with journal-*

artemis-server/src/main/java/org/apache/activemq/artemis/utils/XMLConfigurationUtil.java
michaelpearce-gain: Validator back

clebertsuconic: @michaelandrepearce well.. I don't see a point on that.. really.

It's moot, it's dead code.

michaelpearce-gain: Fair enuf. I see your point

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
clebertsuconic: everything else is already with journal on front.. I only missed it here.

PR #2777, opened 2019-08-01
NO-JIRA Fix NoProcessFilesBehind build on IBM JDK 8

brusdev:
Remove sun.management.VMManagement to build NoProcessFilesBehind with
IBM JDK 8.


artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/NoProcessFilesBehind.java
```diff
@@ -105,7 +105,7 @@ private static int getProcessId() throws ReflectiveOperationException {
       java.lang.management.RuntimeMXBean runtime = java.lang.management.ManagementFactory.getRuntimeMXBean();
       java.lang.reflect.Field jvmField = runtime.getClass().getDeclaredField("jvm");
       jvmField.setAccessible(true);
-      sun.management.VMManagement jvm = (sun.management.VMManagement) jvmField.get(runtime);
```
clebertsuconic: What about the other internal class we use for get files open, that one is ok?

brusdev: The class com.sun.management.UnixOperatingSystemMXBean LGTM, as matter of fact I executed NoProcessFilesBehind.getOpenFD with IBM JDK 8 and I got the right open file descriptor count.

PR #2779, opened 2019-08-01
ARTEMIS-2444 Artemis hawtio plugin overriding sublevel tabs

TonyJulin:
This fix addresses an issue in the HawtIO Artemis plugin where it was overriding the workspace sublevel tabs, thereby causing the HawtIO Camel plugin to only display a limited number of tabs in the UI.

### DIFF
--- a/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/artemisPlugin.js
+++ b/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/artemisPlugin.js
@@ -218,35 +218,7 @@ var ARTEMIS = (function(ARTEMIS) {
          }
       });
 
-      subLevelTabs = [];
-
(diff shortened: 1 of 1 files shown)


TonyJulin:
hawtio_plugin_before_camelview
hawtio_plugin_after_camelview
hawtio_plugin_after_artemisview

clebertsuconic:
Can you open a JIRA in Artemis?

Ideally you would have the jira as part of your commit message


ARTEMIS-XXX BLBLABLABLA


although that's not a big deal as I can ammend your commit before I merge it.. but it would be great if you could do it.


Nice on putting screenshots... I don't know if JIRA has a way to have it.. but at least we can keep a reference of them on the JIRA with a link here.

TonyJulin:
Opened a JIRA issue and ammended the commit message. Let me know if those work.

PR #2782, opened 2019-08-02
ARTEMIS-2441 Separate Lock Files

clebertsuconic:
Certain devices or file systems won't support record level locking.
For that reason I am changing FileLockNodeManager to use separate files (one for each position) instead of using tryLock(position);
A good example for this would be cephFS where channel.tryLock or channel.tryLock works but it fails at a record level.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
@@ -18,7 +18,9 @@ package org.apache.activemq.artemis.core.server.impl;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -82,6 +86,36 @@ public synchronized void start() throws Exception {
...
+      }
+   }
+
+   @Override
+   public synchronized void stop() throws Exception {
+      for (FileChannel channel : lockChannels) {
+         try {
+            channel.close();
+         } catch (Throwable e) {
+
```
clebertsuconic: hmmm... I should have logged it here. (reviewing my own PR)

PR #2785, opened 2019-08-05
ARTEMIS-2442 throw ActiveMQResourceLimitException if session/queue limit is reached

wy96f:
Now ActiveMQSessionCreationException is thrown when session/queue limit is reached. ActiveMQSessionCreationException  is used to indicate server is starting. It would cause session recreation retries all the time. Session creation should fail fast when limit is reached so client can try to connect other brokers in cluster.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQExceptionType.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQExceptionType.java
@@ -261,6 +261,12 @@ public enum ActiveMQExceptionType {
       public ActiveMQException createException(String msg) {
          return new ActiveMQReplicationTimeooutException(msg);
       }
+   },
+   RESOURCE_LIMIT_REACHED(221) {
+      @Override
+      public ActiveMQException createException(String msg) {
+         return new ActiveMQResourceLimitException(msg);
+      }
    };
    private static final Map<Integer, ActiveMQExceptionType> TYPE_MAP;
 
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQResourceLimitException.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQResourceLimitException.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.api.core;
+
+public final class ActiveMQResourceLimitException extends ActiveMQException {
+
+   private static final long serialVersionUID = 4630616634531784944L;
+
+   public ActiveMQResourceLimitException() {
+      super(ActiveMQExceptionType.RESOURCE_LIMIT_REACHED);
+   }
+
+   public ActiveMQResourceLimitException(String msg) {
+      super(ActiveMQExceptionType.RESOURCE_LIMIT_REACHED, msg);
+   }
+}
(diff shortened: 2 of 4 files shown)


artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQExceptionType.java
```diff
@@ -261,6 +261,12 @@ public ActiveMQException createException(String msg) {
       public ActiveMQException createException(String msg) {
          return new ActiveMQReplicationTimeooutException(msg);
       }
+   },
+   RESOURCE_LIMIT_REACHED(221) {
```
michaelandrepearce: What occurs with older clients that dont understand this new exception if this new exception is thrown and passed back to them

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQResourceLimitException.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.api.core;
+
+public final class ActiveMQResourceLimitException extends ActiveMQException {
+
```
michaelandrepearce: Probably should be an extension of the existing ActiveMQSessionCreationException so any old code with an explicit catch clause still gets caught

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQExceptionType.java
wy96f: For older client that don't understand, "ActiveMQException"(type code is GENERIC_EXCEPTION) would be decoded and thrown. There are two places where ActiveMQExceptionType.SESSION_CREATION_REJECTED is judged, one is ActiveMQSessionContext::recreateSession(), the other is BridgeImpl::connect(). For both of them, the logic is same in the case of GENERIC_EXCEPTION(old client) and RESOURCE_LIMIT_EXCEEDED(new client).

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQResourceLimitException.java
wy96f: Do you mean ActiveMQResourceLimitException extends ActiveMQSessionCreationException, and with RESOURCE_LIMIT_REACHED ActiveMQExceptionType? If so, older client still can't understand new exception by new ActiveMQExceptionType code when the code is passed back to client.

michaelandrepearce: Theyre independent comments. E.g. this is as well as the other

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQExceptionType.java
michaelandrepearce: People can use native core client. Not everyone uses jms.

wy96f: ActiveMQSessionContext::recreateSession() and BridgeImpl::connect() both are using native core client api.

michaelandrepearce: Point here just because theyre catching the base doesnt mean everyone is
 By changing the exception should try make it a extended of the current so people dont get any shock changes in their error handling logic.

michaelandrepearce: Like wise we may have remote clients using core, what occurs when server throws new exception but client is older, it wont know how to convert. As such need to handle this server side (most likely using something like the connection version to work out the client version and adjust for older)

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQResourceLimitException.java
wy96f: > Probably should be an extension of the existing ActiveMQSessionCreationException so any old code with an explicit catch clause still gets caught

I don't find code with an explicit catch ActiveMQSessionCreationException clause except in the tests. Most of them are catching ActiveMQException(or other extended exception) and judge type code. And when queue limit is reached(not session), throwing an extension of ActiveMQSessionCreationException sounds not proper, wdyt?

michaelandrepearce: The point here isnt code within the broker but code in end user clients. People who are using core api directly.

michaelandrepearce:
@wy96f what's occurring on this, are you actively wanting to develop this further, or dropping it and we close the PR? Just on a bit of a new year clean up of PR's

wy96f:
@michaelandrepearce closing it for now, it needs more discussion and refinement :)

PR #2788, opened 2019-08-06
ARTEMIS-2447 allow mapping admin to manage in LDAP plugin

jbertram:


clebertsuconic:
I ran this on the entire testsuite and it went fine.  Travis is acting up.

PR #2789, opened 2019-08-08
ARTEMIS-2448 group-name ignored in replicated colocated setup

franz1981:


franz1981:
I'm working on a full test, but I cannot find any colocated replicated test yet, @clebertsuconic do you know if where are some?

PR #2790, opened 2019-08-08
ARTEMIS-2449 limit size of producer details

jbertram:
The core server session tracks details about producers like what
addresses have had messages sent to them, the most recent message ID
sent to each address, and the number of messages sent to each address.
This information is made available to users via the
listProducersInfoAsJSON method on the various management interfaces
(JMX, web console, etc.). However, in situations where a server session
is long lived (e.g. in a pool) and is used to send to many different
addresses (e.g. randomly named temporary JMS queues) this info can
accumulate to a problematic degree. Therefore, we should limit the
amount of producer details saved by the session.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
@@ -95,6 +95,7 @@ import org.apache.activemq.artemis.spi.core.protocol.SessionCallback;
 import org.apache.activemq.artemis.utils.CompositeAddress;
 import org.apache.activemq.artemis.utils.JsonLoader;
 import org.apache.activemq.artemis.utils.PrefixUtil;
+import org.apache.activemq.artemis.utils.collections.MaxSizeMap;
 import org.apache.activemq.artemis.utils.collections.TypedProperties;
 import org.jboss.logging.Logger;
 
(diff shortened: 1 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -175,7 +176,7 @@
    private final OperationContext context;
 
    // Session's usage should be by definition single threaded, hence it's not needed to use a concurrentHashMap here
-   protected final Map<SimpleString, Pair<Object, AtomicLong>> targetAddressInfos = new HashMap<>();
+   protected final Map<SimpleString, Pair<Object, AtomicLong>> targetAddressInfos = new MaxSizeMap<>(100);
```
michaelandrepearce: Can this be configurable

jbertram: Yes, it could be configurable and you're right that 100 is arbitrary but so would any other not-ulimited value.

clebertsuconic: Will this represent any error in case the size > 100 or this is just for information?

If I send to a destination that was cleaned before, will I be able to send it again?


I believe this is just for informational purposes on the Console?

jbertram: There won't be any error. The map will just silently drop the oldest entry when the maxSize is hit. You'll be able to send to any destination again that was dropped previously due to maxSize. This information is only for administrative use.

jbertram:
BTW, I ran the PR build locally and it passed. I don't know what's going on with Travis lately.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
clebertsuconic: @michaelandrepearce / @jbertram I don't know.. we already have too many options to configure. I don't foresee anyone actually needing to tweak this value. Can't we just leave it be?

You are already removing temporary destinations from the session. And this is for the corner case where the Session sending the temporary destination is a different session.

PR #2791, opened 2019-08-09
ARTEMIS-2450 page-size-bytes should not be greater than Integer.MAX_VALUE

wy96f:


artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -505,6 +505,9 @@ public long getPageSizeBytes() {
    }
 
    public AddressSettings setPageSizeBytes(final long pageSize) {
+      if (pageSize > Integer.MAX_VALUE) {
+         throw new IllegalArgumentException("pageSize must be < " + Integer.MAX_VALUE);
```
michaelandrepearce: Why not change long to int, this would then be cleaner from anyone trying to use the method

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationParserTest.java
```diff
@@ -257,6 +257,26 @@ public void testParsingDefaultServerConfigWithENCMaskedPwd() throws Exception {
       assertEquals("helloworld", bconfig.getPassword());
    }
 
+   @Test
+   public void testParsingOverflowPageSize() throws Exception {
+      FileConfigurationParser parser = new FileConfigurationParser();
```
michaelandrepearce: This test, test the new change, but doesnt test for what we are protecting from. E.g whats the underlying issue being fixed

michaelandrepearce: E.g. other sizes are quite happily long currently also. Why is this special

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
wy96f: The AddressSettings maybe persisted in disk, and if we change long to int, we would incorrectly decode it. Does it make sense?

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationParserTest.java
wy96f: > E.g. other sizes are quite happily long currently also. Why is this special

I see in Page::write()/Page::read() all operations are limited to int range, such as PagingStoreImpl::currentPageSize, Page::size, and local variables like fileSize, processedBytes in Page::readFromSequentialFile(), etc.

> This test, test the new change, but doesnt test for what we are protecting from. E.g whats the underlying issue being fixed

I'm not sure what you mean. We're protecting page-size-bytes from being greater than Integer.MAX_VALUE(2147483647). So if we set page-size-bytes to 2147483648, the broker is expected to fail fast to throw exception , correct?

jbertram:
@wy96f, can you rebase and push -f to trigger a rebuild of the PR by Travis? An issue was recently fixed that was preventing the PR build from completing properly. Thanks!

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
michaelandrepearce: Not really. If you now decode something thats larger than int it will blow up with the exception youve added. No different. Except we are actually making the method take the correct primitive number

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationParserTest.java
michaelandrepearce: Point more so was i assume the same issue may lay in other persistence such as journal etc

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
michaelandrepearce: Also should update the schema to int so then when people do parse configs via xsd validators it would fail

wy96f:
@jbertram will re-work this based on comments. After that will rebase and push :)

wy96f:
pushed to handle "page-size-bytes" the same way as "journal-file-size"

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/Validators.java
```diff
@@ -122,6 +122,30 @@ public void validate(final String name, final Object value) {
       }
    };
 
+   public static final Validator POSITIVE_INT = new Validator() {
```
michaelpearce-gain: This isnt needed, simply change the xsd to int and it would fail

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/Validators.java
```diff
@@ -122,6 +122,30 @@ public void validate(final String name, final Object value) {
...
+         Number val = (Number) value;
+         if (val.longValue() > 0 && val.longValue() <= Integer.MAX_VALUE) {
+            // OK
+         } else {
+            throw ActiveMQMessageBundle.BUNDLE.inRangeOfPositiveInt(name, val);
+         }
+      }
+   };
+
+   public static final Validator MINUS_ONE_OR_POSITIVE_INT = new Validator() {
```
michaelpearce-gain: again not needed, if the xsd (schema is corrected)

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -1756,17 +1758,17 @@ private void parseClusterConnectionConfiguration(final Element e, final Configur
 
       double retryIntervalMultiplier = getDouble(e, "retry-interval-multiplier", ActiveMQDefaultConfiguration.getDefaultClusterRetryIntervalMultiplier(), Validators.GT_ZERO);
 
-      int minLargeMessageSize = getTextBytesAsIntBytes(e, "min-large-message-size", ActiveMQClient.DEFAULT_MIN_LARGE_MESSAGE_SIZE, Validators.GT_ZERO);
+      int minLargeMessageSize = getTextBytesAsIntBytes(e, "min-large-message-size", ActiveMQClient.DEFAULT_MIN_LARGE_MESSAGE_SIZE, Validators.POSITIVE_INT);
```
michaelpearce-gain: revert, not needed

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -1756,17 +1758,17 @@ private void parseClusterConnectionConfiguration(final Element e, final Configur
...
 
       long maxRetryInterval = getLong(e, "max-retry-interval", ActiveMQDefaultConfiguration.getDefaultClusterMaxRetryInterval(), Validators.GT_ZERO);
 
       int initialConnectAttempts = getInteger(e, "initial-connect-attempts", ActiveMQDefaultConfiguration.getDefaultClusterInitialConnectAttempts(), Validators.MINUS_ONE_OR_GE_ZERO);
 
       int reconnectAttempts = getInteger(e, "reconnect-attempts", ActiveMQDefaultConfiguration.getDefaultClusterReconnectAttempts(), Validators.MINUS_ONE_OR_GE_ZERO);
 
-      int confirmationWindowSize = getTextBytesAsIntBytes(e, "confirmation-window-size", ActiveMQDefaultConfiguration.getDefaultClusterConfirmationWindowSize(), Validators.GT_ZERO);
+      int confirmationWindowSize = getTextBytesAsIntBytes(e, "confirmation-window-size", ActiveMQDefaultConfiguration.getDefaultClusterConfirmationWindowSize(), Validators.POSITIVE_INT);
```
michaelpearce-gain: revert not needed

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -1756,17 +1758,17 @@ private void parseClusterConnectionConfiguration(final Element e, final Configur
...
 
-      int producerWindowSize = getTextBytesAsIntBytes(e, "producer-window-size", ActiveMQDefaultConfiguration.getDefaultBridgeProducerWindowSize(), Validators.MINUS_ONE_OR_GT_ZERO);
+      int producerWindowSize = getTextBytesAsIntBytes(e, "producer-window-size", ActiveMQDefaultConfiguration.getDefaultBridgeProducerWindowSize(), Validators.MINUS_ONE_OR_POSITIVE_INT);
```
michaelpearce-gain: revert not needed

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationTest.java
```diff
@@ -364,7 +364,7 @@ public void testDefaults() {
       assertEquals("a2.2", conf.getAddressesSettings().get("a2").getExpiryAddress().toString());
       assertEquals(5, conf.getAddressesSettings().get("a2").getRedeliveryDelay());
       assertEquals(932489234928324L, conf.getAddressesSettings().get("a2").getMaxSizeBytes());
-      assertEquals(7126716262626L, conf.getAddressesSettings().get("a2").getPageSizeBytes());
+      assertEquals(712671626L, conf.getAddressesSettings().get("a2").getPageSizeBytes());
```
michaelpearce-gain: should now be int, no need for L

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -597,11 +597,11 @@ public void parseMainConfig(final Element e, final Configuration config) throws
 
       config.setJournalSyncNonTransactional(getBoolean(e, "journal-sync-non-transactional", config.isJournalSyncNonTransactional()));
 
-      config.setJournalFileSize(getTextBytesAsIntBytes(e, "journal-file-size", config.getJournalFileSize(), Validators.GT_ZERO));
+      config.setJournalFileSize(getTextBytesAsIntBytes(e, "journal-file-size", config.getJournalFileSize(), Validators.POSITIVE_INT));
 
       int journalBufferTimeout = getInteger(e, "journal-buffer-timeout", config.getJournalType() == JournalType.ASYNCIO ? ArtemisConstants.DEFAULT_JOURNAL_BUFFER_TIMEOUT_AIO : ArtemisConstants.DEFAULT_JOURNAL_BUFFER_TIMEOUT_NIO, Validators.GE_ZERO);
 
-      int journalBufferSize = getTextBytesAsIntBytes(e, "journal-buffer-size", config.getJournalType() == JournalType.ASYNCIO ? ArtemisConstants.DEFAULT_JOURNAL_BUFFER_SIZE_AIO : ArtemisConstants.DEFAULT_JOURNAL_BUFFER_SIZE_NIO, Validators.GT_ZERO);
+      int journalBufferSize = getTextBytesAsIntBytes(e, "journal-buffer-size", config.getJournalType() == JournalType.ASYNCIO ? ArtemisConstants.DEFAULT_JOURNAL_BUFFER_SIZE_AIO : ArtemisConstants.DEFAULT_JOURNAL_BUFFER_SIZE_NIO, Validators.POSITIVE_INT);
```
michaelpearce-gain: change was not needed here.

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -597,11 +597,11 @@ public void parseMainConfig(final Element e, final Configuration config) throws
 
       config.setJournalSyncNonTransactional(getBoolean(e, "journal-sync-non-transactional", config.isJournalSyncNonTransactional()));
 
-      config.setJournalFileSize(getTextBytesAsIntBytes(e, "journal-file-size", config.getJournalFileSize(), Validators.GT_ZERO));
+      config.setJournalFileSize(getTextBytesAsIntBytes(e, "journal-file-size", config.getJournalFileSize(), Validators.POSITIVE_INT));
```
michaelpearce-gain: change was not needed here, int range should all be validated by xsd type.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/Validators.java
michaelpearce-gain: ignore i spotted, why finally, its because GB etc can be used in the string type, and then converted as such cannot limit this in xsd.

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
michaelpearce-gain: ignore i was wrong, i see the issue.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/Validators.java
michaelpearce-gain: ignore i spotted, why finally, its because GB etc can be used in the string type, and then converted as such cannot limit this in xsd.

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
michaelpearce-gain: ignore i spotted why you need.

PR #2792, opened 2019-08-12
NO-JIRA: Javadoc for ActiveMQServerMessagePlugin interface

jirkadanek:


jdanekrh:
I still need to try this with a broker cluster, to see what exactly happens there.

jbertram:
I think this would be better in the User Guide rather than the JavaDoc for this class.

jdanekrh:
Ok, I can move it there.

I saw some java classes with quite extensive Javadoc, and I felt that explaining how user provided methods are called belongs in a Javadoc.

jdanekrh:
(Does it look reasonable so far, what I wrote? Especially that using ThreadLocal is a good idea?)

jbertram:
I think what you wrote looks good. I wouldn't go as far as saying that using a `ThreadLocal` is a *good* idea, but it should work nonetheless. :)

jbertram:
I'm closing this as I don't expect it will receive any further updates. Please re-open if needed.

PR #2793, opened 2019-08-13
ARTEMIS-2452 group-name ignored in shared store colocated setup

franz1981:


franz1981:
This PR is adding `BackupRequestMessage`'s `nodeID` information for shared store colocated setup, to enable proper validation of backup group names: I don't know yet how to handle the case where old brokers are communicating with new ones, but throwing an exception + logging a warn message. I'm opened to other approaches ;)
I will add some tests on my PTO return, but I've left this here to make it available for comment/reviewes :)

franz1981:
@howardgao @jbertram wdyt?

michaelandrepearce:
Its important broker can be rolling upgraded, e.g. old broker instance can talk with new without issue so zero downtime upgrades can occur. I think the issue you called out must be addressed before this can merge, if it breaks that.

franz1981:
@michaelandrepearce Agree, but nonetheless my comment on the code hold:
```
// Older versions of artemis don't send the nodeID in BackupRequestMessage:
      // in this case we cannot trust the request, making the requesting server
      // to pair with this server in any case.
```
Old servers cannot be trusted, because they don't provide enough information to pair with new ones ie the colocated shared store with group name wasn't a feature correctly implemented.
If I let them behave as before it can break a new cluster: atm i can't see solutions to preserve a partial cluster upgrade.

michaelandrepearce:
Why not have a setting that makes it behave as old version, then once all servers uograded the setting can change to strict and then simply another round or rolling bounces.

michaelandrepearce:
Also why would having old logic break new? Another option is to version the message, and if you get v1 do old logic, if you get new v2 message do new. Then no need even for a flag.

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/BackupRequestMessage.java
```diff
@@ -34,12 +34,14 @@ public BackupRequestMessage() {
...
 
    public BackupRequestMessage(int backupSize,
+                               SimpleString nodeID,
                                String journalDirectory,
                                String bindingsDirectory,
                                String largeMessagesDirectory,
                                String pagingDirectory) {
       super(BACKUP_REQUEST);
       this.backupSize = backupSize;
+      this.nodeID = nodeID;
```
michaelandrepearce: Please make a v2 version, this would be a breaking change, for anyone with existing clusters

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/ha/ColocatedHAManager.java
```diff
@@ -163,6 +174,23 @@ private synchronized boolean activateSharedStoreBackup(String journalDirectory,
       return true;
    }
 
+   private TopologyMember validateBackupGroupName(SimpleString nodeID) {
+      // Older versions of artemis don't send the nodeID in BackupRequestMessage:
+      // in this case we cannot trust the request, making the requesting server
```
michaelandrepearce: Why can't we, we know we had old servers and setups working fine, as such why be so strict, simply fail back (or even tter have a v1 and v2 version of the message, that way if all v2 cluster you can be fully strict, else in a v1/v2 mix you can be less strict

franz1981:
The version proposal is probably the best solution indeed: the reason why the old logic breaks the new one is that the old one just ignored the group names and let any requesting server that arrive first to be able to form a pair on any group name regardless both the requesting server group name and the target server. It was just totally broken...

michaelandrepearce:
Well it can't be totally broken, theres users out there using it (mostly because they are forming single groups or ensuring clusters are rign fenced), yes theres a bug thats not ideal.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/ha/ColocatedHAManager.java
franz1981: It was working fine because you haven't had any other member of topology with a group name and been lucky, I believe :P

franz1981:
The feature of shared store colocated is not totally broken, but the shared store colocation using group names, yes.
But is normal, if the were not using at all group names in their logic I believe that's ok..

franz1981:
I understand your comment now, my comment was related to the code itself,not the feature as a whole: I see that there exists cases where it can work..

michaelandrepearce:
@franz1981 i assume you're going to have a think on this and re-work bits? e.g. no further comments needed for now right?

franz1981:
Yes, sure and will provide proper yes and coverage too as it deserves :)
Thanks for the suggestions Michael!!

clebertsuconic:
can you rebase this? it will trigger a new build after fixed.

michaelandrepearce:
@clebertsuconic its not ready for merge anhow

clebertsuconic:
@michaelandrepearce I'm marking as DO-NOT-MERGE-YET... it helps me to not mess up.

franz1981:
Closing this because inactive from long time

PR #2794, opened 2019-08-13
ARTEMIS-2451 limit size of destination cache

jbertram:


artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQConnection.java
```diff
@@ -97,7 +98,7 @@
 
    private final Set<SimpleString> tempQueues = new ConcurrentHashSet<>();
 
-   private final Set<SimpleString> knownDestinations = new ConcurrentHashSet<>();
+   private final ConcurrentMaxSizeCache<SimpleString> knownDestinations = new ConcurrentMaxSizeCache<>(100);
```
michaelandrepearce: Can we make this configurable (e.g. a setting in the CF), 100 is a bit of a magic number.

michaelandrepearce: this will actually cause issues if someone is using temp destinations, and has say more than 100 legitimately. and creates the consumers later, as the consumer will check it its known, which now it wont and will throw an error. 

This will at least mean people can increase that setting if they need.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/client/TemporaryDestinationTest.java
```diff
@@ -266,6 +266,24 @@ public void testForTempQueueTargetInfosSizeLimit() throws Exception {
...
+   public void testForKnownAddressSizeLimit() throws Exception {
+      try {
+         conn = createConnection();
+         Session s = conn.createSession(false, Session.AUTO_ACKNOWLEDGE);
+         for (int i = 0; i < 200; i++) {
+            TemporaryQueue temporaryQueue = s.createTemporaryQueue();
+            MessageProducer producer = s.createProducer(temporaryQueue);
+            producer.send(s.createMessage());
+         }
+         assertTrue(((ActiveMQConnection)conn).getKnownDestinations().size() <= 100);
```
michaelandrepearce: Need a test case where a temporary dest is created first and then later consumers are created on it, currently with this change the existing behaviour which allows  this and will mean on create consumer for the temp dest will throw error as will be not a known dest.

michaelandrepearce: e.g. where you make a temp dest for a reply from for other services, that is set in the replyto on the produced message, and then setup the consumer to listen to, currently if > 100 other services it would need to co-ordinate with, it will now break.

michaelandrepearce: An alternative solution to this problem is remove knownDestination entirely, and have a reference to the session & connection in the Destination itself, and then it will no longer be needed, as we can do the check directly on the destination, e.g. something like
   ` jbdest.getConnection == connection`

Removing the need for the set

michaelandrepearce: it seems we already have session inside destination as such this would be relatively easy.

expose the session on ActiveMQDestination e.g.
```
   public ActiveMQSession session() {
      return session;
   }

```
And then in this check:

```
      if (jbdest.isTemporary() && (jbdest.session() == null || connection != jbdest.session().getConnection())) {
```

lastly just clean up bits that we adding and removing temp address into the cache, as redundent.

michaelandrepearce: similar could be done for  method

void checkDestination(ActiveMQDestination destination) throws JMSException 

just check if the connection was the one that created the destination object, and thus remove need for cache

michaelandrepearce: ive very quickly tried to code up what im trying to explain here: https://github.com/apache/activemq-artemis/pull/2796

jbertram:
Closing in favor of #2796.

PR #2796, opened 2019-08-14
ARTEMIS-2451 remove need for knownDestination Cache

michaelandrepearce:
@jbertram this is partially what i was trying to explain as an alternative approach to simply remove the need for the cache https://github.com/apache/activemq-artemis/pull/2794.

I assume you've hit some client side memory leak issue, that you're trying to resolve

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/QueueAutoCreationTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/QueueAutoCreationTest.java
@@ -23,10 +23,7 @@ import org.apache.activemq.artemis.core.config.Configuration;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
 import org.apache.activemq.artemis.core.server.JournalType;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
-import org.apache.activemq.artemis.jms.client.ActiveMQConnection;
-import org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory;
-import org.apache.activemq.artemis.jms.client.ActiveMQQueue;
-import org.apache.activemq.artemis.jms.client.ActiveMQTopic;
+import org.apache.activemq.artemis.jms.client.*;
 import org.apache.activemq.artemis.utils.UUIDGenerator;
 import org.junit.After;
 import org.junit.Assert;
@@ -130,7 +127,7 @@ public class QueueAutoCreationTest extends JMSClientTestSupport {
          producer.send(session.createTextMessage("hello"));
       }
 
-      Assert.assertTrue(((ActiveMQConnection)connection).containsKnownDestination(addressName));
+      Assert.assertTrue(((ActiveMQDestination) topic).isCreated());
    }
 
    private void sendStringOfSize(int msgSize, boolean useCoreReceive) throws JMSException {
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/AutoCreateJmsDestinationTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/AutoCreateJmsDestinationTest.java
@@ -41,12 +41,8 @@ import org.apache.activemq.artemis.core.client.impl.ServerLocatorImpl;
 import org.apache.activemq.artemis.core.security.Role;
 import org.apache.activemq.artemis.core.server.Queue;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
-import org.apache.activemq.artemis.jms.client.ActiveMQConnection;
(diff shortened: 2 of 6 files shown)


jbertram:
At first glance this looks like a clever solution and better than the one I came up with. Let me review it in more detail.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/client/TemporaryDestinationTest.java
```diff
@@ -227,45 +228,6 @@ public void testForTempQueueCleanerUpperLeak() throws Exception {
       }
    }
 
-   @Test
```
jbertram: These tests are still valid and shouldn't be removed. They are testing the server-side accumulation of address info. The changes in this PR are for the client-side.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/client/TemporaryDestinationTest.java
```diff
@@ -81,7 +82,7 @@ public void testTemporaryQueueLeak() throws Exception {
 
          tempQueue.delete();
 
-         assertFalse(conn.containsKnownDestination(SimpleString.toSimpleString(tempQueue.getQueueName())));
+         assertFalse(((ActiveMQDestination) tempQueue).isCreated());
```
jbertram: This test fails because `ActiveMQDestination.delete()` doesn't call `setCreated(false)`.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/AutoCreateJmsDestinationTest.java
```diff
@@ -41,12 +41,8 @@
 import org.apache.activemq.artemis.core.security.Role;
 import org.apache.activemq.artemis.core.server.Queue;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
-import org.apache.activemq.artemis.jms.client.ActiveMQConnection;
-import org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory;
-import org.apache.activemq.artemis.jms.client.ActiveMQQueue;
-import org.apache.activemq.artemis.jms.client.ActiveMQTemporaryTopic;
-import org.apache.activemq.artemis.jms.client.ActiveMQTopic;
-import org.apache.activemq.artemis.tests.util.Wait;
+import org.apache.activemq.artemis.jms.client.*;
```
jbertram: Star imports are a check-style violation.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/QueueAutoCreationTest.java
```diff
@@ -23,10 +23,7 @@
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
 import org.apache.activemq.artemis.core.server.JournalType;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
-import org.apache.activemq.artemis.jms.client.ActiveMQConnection;
-import org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory;
-import org.apache.activemq.artemis.jms.client.ActiveMQQueue;
-import org.apache.activemq.artemis.jms.client.ActiveMQTopic;
+import org.apache.activemq.artemis.jms.client.*;
```
jbertram: Star imports are a check-style violation.

jbertram:
BTW, the issue I saw wasn't a leak, per se. It was simply unwanted accumulation based on the way it was designed.

michaelandrepearce:
@jbertram this was a quick scrap togeather of what i was trying to describe in review comment on your pr. Im happy to complete it fully for you, but it wont be till next week. Im more than happy if you wish to just take it and complete it (e.g. dont worry about keeping my authorship)

michaelandrepearce:
@jbertram 

re:

> BTW, the issue I saw wasn't a leak, per se. It was simply unwanted accumulation based on the way it was designed.

Won't we get similar issue on the tempoary queue set thats used to track temp queues?

jbertram:
It's *possible* that `tempQueues` could grow like `knownDestinations`. However, entries are only added to `tempQueues` for temporary destinations that the client actually creates and entries from `tempQueues` get removed when the corresponding temporary destination is deleted from the broker. An entry is added to `knownDestinations` for every unique destination to which a client sends a message and those entries *never* get removed unless they are temporary destinations which the client itself deletes. In my mind these facts make the pathological growth much less likely.

Also, in JMS terms the lifetime of a temporary destination is actually the lifetime of the *connection* that created it, not the session. The temporary destinations need to be tracked somehow so that the client implementation can delete them implicitly when the connection is closed in the case that the application doesn't delete them explicitly.

jbertram:
I just sent #2800 which is based on your work here, @michaelandrepearce.

PR #2799, opened 2019-08-15
ARTEMIS-2453 Fixing deadLock between destroyQueue and removeAddressInfo

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -649,9 +647,7 @@ public AddressInfo removeAddressInfo(SimpleString address, boolean force) throws
 
    @Override
    public AddressInfo getAddressInfo(SimpleString addressName) {
-      synchronized (addressLock) {
-         return addressManager.getAddressInfo(addressName);
-      }
+      return addressManager.getAddressInfo(addressName);
```
clebertsuconic: There's no need to lock here?

clebertsuconic: @jbertram What do you think about this?

jbertram: I don't see a need for the `synchronized (addressLock)` given that the underlying data structure in the address manager is a `ConcurrentHashMap` and no other operations are being done (unlike in the other places which use `addressLock`). I think this fix looks good.

clebertsuconic: @jbertram thanks.. that confirms what I thought.

I'm running the whole testsuite to confirm it.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -484,7 +482,7 @@ public QueueBinding updateQueue(SimpleString name,
                                    Long delayBeforeDispatch,
                                    SimpleString user,
                                    Boolean configurationManaged) throws Exception {
-      synchronized (addressLock) {
+      synchronized (this) {
```
michaelandrepearce: Why not make synchronized method for these methods if the whole block is syncd?

PR #2800, opened 2019-08-16
ARTEMIS-2451 eliminate knownDestinations cache

jbertram:
This is based on the work from #2796.  The connection tracking in that PR was removed as it's not necessary since we still track temporary queues.


michaelandrepearce:
+1

clebertsuconic:
@jbertram Can I ask you to add a test with this PR?

One that shows users doing this anti pattern:

```java
     int N = 100; // or any arbitrary number > 1
     for (int i = 0; i < N; i++) {
          Queue queue = session.createQueue("myQueue");
          MessageProducer prod = session.createProducer(queue);
          prod.send(....);
          prod.close();
     }
```

If an user did such anti-pattern, it should suffer the consequences of the extra roundtrip of course, and we should not optimize for it.


I just want to make sure it wouldn't fail for any reason.

clebertsuconic:
Other than that simple / dumb test to be added! +1

clebertsuconic:
@jbertram actually, leave the test with me

PR #2801, opened 2019-08-18
ARTEMIS-2454 Message Body damaged after re-encoding

clebertsuconic:


artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/converter/TestConversions.java
```diff
@@ -31,14 +33,17 @@
 import org.apache.activemq.artemis.protocol.amqp.converter.jms.ServerJMSTextMessage;
 import org.apache.activemq.artemis.protocol.amqp.util.NettyReadable;
 import org.apache.activemq.artemis.protocol.amqp.util.NettyWritable;
+import org.apache.activemq.artemis.protocol.amqp.util.TLSEncode;
```
michaelandrepearce: I dont think this needed to be added. I cant see it being used

clebertsuconic: it's failing on checkstyle right here actually.. fixing it.

PR #2802, opened 2019-08-20
ARTEMIS-2457 implement ring queue

jbertram:
Don't merge this yet. I've still got a few tests to add for basic configuration stuff, and I also need to finish the docs. I'm putting this up now to get reviews/opinions.


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueuePendingMessageMetrics.java
```diff
@@ -49,14 +52,18 @@
...
+   public QueuePendingMessageMetrics(final Queue queue, final String name) {
       Preconditions.checkNotNull(queue);
       this.queue = queue;
+      this.name = name;
    }
 
    public void incrementMetrics(final MessageReference reference) {
       long size = getPersistentSize(reference);
       COUNT_UPDATER.incrementAndGet(this);
+      logger.debugf("%s increment messageCount to %d: %s", this, messageCount, reference);
```
michaelandrepearce: can we wrap this in a isDebugEnabled

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueuePendingMessageMetrics.java
```diff
@@ -67,6 +74,7 @@ public void incrementMetrics(final MessageReference reference) {
    public void decrementMetrics(final MessageReference reference) {
       long size = -getPersistentSize(reference);
       COUNT_UPDATER.decrementAndGet(this);
+      logger.debugf("%s decrement messageCount to %d: %s", this, messageCount, reference);
```
michaelandrepearce: can we wrap this in a isDebugEnabled

michaelandrepearce:
In QPID both size in message count can be configured / or size in total bytes.  I cannot see how we can set the latter?

What's the reason to make this a brand new queue type entirely over just making it something configurable, i ask as it means cannot change an existing queue into a ring queue without deleting and recreating, where as having it a setting in QueueImpl like almost all others would give the dynamic ability. 

On top of the last comment, can we dynamically update the size, e.g. increase or decrease its limit, so that if incorrectly sized it can be tuned, without need of re-creating.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/QueueQueryImpl.java
```diff
@@ -95,70 +97,7 @@ public QueueQueryImpl(final boolean durable,
                          final SimpleString name,
                          final boolean exists,
                          final boolean autoCreateQueues) {
-      this(durable, temporary, consumerCount, messageCount, filterString, address, name, exists, autoCreateQueues, -1, false, false, RoutingType.MULTICAST);
```
michaelandrepearce: wont this be a breaking change removing the old api's if anyone is using CORE api directly? I appreciate its hidden for JMS users.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServer.java
```diff
@@ -449,7 +449,7 @@ Queue createQueue(AddressInfo addressInfo, SimpleString queueName, SimpleString
    Queue createQueue(AddressInfo addressInfo, SimpleString queueName, SimpleString filter,
                      SimpleString user, boolean durable, boolean temporary, boolean autoCreated, Integer maxConsumers,
                      Boolean purgeOnNoConsumers, Boolean exclusive, Boolean groupRebalance, Integer groupBuckets, SimpleString groupFirstKey, Boolean lastValue, SimpleString lastValueKey, Boolean nonDestructive,
-                     Integer consumersBeforeDispatch, Long delayBeforeDispatch, Boolean autoDelete, Long autoDeleteDelay, Long autoDeleteMessageCount, boolean autoCreateAddress) throws Exception;
+                     Integer consumersBeforeDispatch, Long delayBeforeDispatch, Boolean autoDelete, Long autoDeleteDelay, Long autoDeleteMessageCount, boolean autoCreateAddress, Long ringSize) throws Exception;
```
michaelandrepearce: Breaking api change, since last version, should have an override method.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -3272,7 +3273,8 @@ public Queue createQueue(final AddressInfo addrInfo,
                             final long autoDeleteDelay,
                             final long autoDeleteMessageCount,
                             final boolean autoCreateAddress,
-                            final boolean configurationManaged) throws Exception {
+                            final boolean configurationManaged,
+                            final long ringSize) throws Exception {
```
michaelandrepearce: breaking api change, new param needs a new method added (old method interface honoured)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueuePendingMessageMetrics.java
michaelandrepearce: can we put inside if debug enabled (as on hot path) so will be removed by jit if debug isnt on.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueuePendingMessageMetrics.java
michaelandrepearce: can we put inside if debug enabled (as on hot path) so will be removed by jit if debug isnt on.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RingQueue.java
```diff
@@ -0,0 +1,134 @@
...
+   private void enforceRing() {
+      enforceRing(null, false);
+   }
+
+   private void enforceRing(MessageReference refToAck, boolean scheduling) {
+      if (getMessageCountForRing() >= ringSize) {
+         refToAck = refToAck == null ? messageReferences.poll() : refToAck;
+
+         if (refToAck != null) {
+            logger.infof("Preserving ringSize %d by acking message ref %s", ringSize, refToAck);
```
michaelandrepearce: assume this is accidentally left in, would be very verbose logging at info

docs/user-manual/en/ring-queue.md
```diff
@@ -0,0 +1,99 @@
...
+
+However, before we get to those use cases let's look at the basic configuration
+of a ring queue.
+
+## Configuration
+
+There are 2 parameters related to ring queue configuration.
+
+The `ring-size` parameter can be set directly on the `queue` element. The
+default value is `-1` (i.e. no limit).
```
michaelandrepearce: wont the default value come from address setting?

docs/user-manual/en/ring-queue.md
```diff
@@ -0,0 +1,99 @@
...
+<address name="myRing">
+   <anycast>
+      <queue name="myRing" ring-size="3" />
+   </anycast>
+</address>
+```
+
+The `default-ring-size` is an `address-setting` which applies to queues on
+matching addresses which don't have an explicit `ring-size` set. This is
+especially useful for auto-created queues.
```
michaelandrepearce: if not set default is -1 at the address setting level

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/RingQueueTest.java
```diff
@@ -0,0 +1,186 @@
...
+import org.apache.activemq.artemis.core.server.impl.RingQueue;
+import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
+import org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory;
+import org.apache.activemq.artemis.tests.util.ActiveMQTestBase;
+import org.apache.activemq.artemis.tests.util.Wait;
+import org.apache.activemq.artemis.utils.RandomUtil;
+import org.junit.Before;
+import org.junit.Test;
+
+public class RingQueueTest extends ActiveMQTestBase {
```
michaelandrepearce: Can we add a test with persistence, also a test with non-destructive ring queue (ensure that when non-destructive it sill replaces/acks)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RingQueue.java
```diff
@@ -0,0 +1,134 @@
...
+   }
+
+   @Override
+   public long getRingSize() {
+      return ringSize;
+   }
+
+   public long getMessageCountForRing() {
+      return (long) pendingMetrics.getMessageCount();// + (pageSubscription == null ? 0 : pageSubscription.getMessageCount());
+   }
```
michaelandrepearce: hashcode and equals probably need implemented similar  like LVQ

cshannon:
This is kind of nit picky but I would prefer a different name like "FixedSizeQueue" instead of "RingQueue" or something just because the name RingQueue immediately made me think of Ring/Circular buffers where memory is re-used for performance so I thought this would be something similar to LMAX Disruptor, etc.  However, obviously the goal here is to just to maintain a max size or fix sized queue by acking the oldest message.  But I'm certainly not going to -1 over the name, if everyone wants to keep RingQueue that's fine.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
jbertram: Hmmm...Even though this is public it isn't part of the interface so I'm inclined to leave this.

michaelandrepearce:
I think in qpid it was ring queue, for those moving from qpid over its kind of nice to keep consistent but not strongly opinionated

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueuePendingMessageMetrics.java
jbertram: Will fix.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueuePendingMessageMetrics.java
jbertram: Will fix.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/QueueQueryImpl.java
jbertram: Will fix.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServer.java
jbertram: Will fix.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueuePendingMessageMetrics.java
jbertram: Will fix.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueuePendingMessageMetrics.java
jbertram: Will fix.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RingQueue.java
jbertram: Will fix.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RingQueue.java
```diff
@@ -0,0 +1,134 @@
...
+   private void enforceRing(MessageReference refToAck, boolean scheduling) {
+      if (getMessageCountForRing() >= ringSize) {
+         refToAck = refToAck == null ? messageReferences.poll() : refToAck;
+
+         if (refToAck != null) {
+            logger.infof("Preserving ringSize %d by acking message ref %s", ringSize, refToAck);
+            referenceHandled(refToAck);
+
+            try {
+               refToAck.acknowledge(null, AckReason.REPLACED, null);
```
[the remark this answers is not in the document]
jbertram: IMO it's different enough to warrant a new method.

docs/user-manual/en/ring-queue.md
jbertram: Will fix.

docs/user-manual/en/ring-queue.md
jbertram: Will fix.

jbertram:
@cshannon, @michaelandrepearce is right about the name coming from Qpid. I see your point, but for continuity with Qpid I think it makes sense to keep the current name.

cshannon:
@jbertram - Alright that's fine to keep the name then since there is some precedence on the name coming from Qpid.

jbertram:
> In QPID both size in message count can be configured / or size in total bytes. I cannot see how we can set the latter?

Only count is supported at this point.

> What's the reason to make this a brand new queue type entirely over just making it something configurable...

I was following the pattern established by the last-value queue implementation. I did consider just making it part of the queue implementation, but it seemed cleaner to separate out the ring code.

> ...can we dynamically update the size, e.g. increase or decrease its limit, so that if incorrectly sized it can be tuned, without need of re-creating.

There is no facility for changing the ring-size at this point. I'll look at adding that as it would be nice to have.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/RingQueueTest.java
jbertram: What do you mean a test with persistence? The existing tests use persistent messages. Were you talking about testing the persistence of the queue binding encoding?

jbertram:
I resolved the conflict, updated the code per the reviews, & added some additional tests. I'm still working on the docs, though so don't merge quite yet.

jbertram:
Docs are done. This is ready to merge from my perspective.

jbertram:
@wy96f, take a look at the code now. I believe I addressed your concerns.

michaelandrepearce:
Is there added ability to update the size? I really would like this sorted before merge as it stands with it not being implemented inside the main queueimpl i beleive it makes it very hard to implement. Thus my original feedback / query to why have it as independent type, instead of implementing within queueimpl.

michaelandrepearce:
Ill try make something this weekend to show what im meaning

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RingQueue.java
```diff
@@ -0,0 +1,166 @@
...
+   public boolean allowsReferenceCallback() {
+      return false;
+   }
+
+   private void enforceRing() {
+      enforceRing(null, false);
+   }
+
+   private void enforceRing(MessageReference refToAck, boolean scheduling) {
+      if (getMessageCountForRing() >= ringSize) {
```
wy96f: I think we should judge whether ref is already acked first.

E.g. for non destructive queue, suppose ring size is 1, m1 is sent and delivered to consumer c1. Then m2 is sent and delivered to consumer c2. Thus m1 is replace acked and m2 resides in "messageReferences"(message count is now 1). Now consumer c1 is closed without acking m1, m1 is rollbacked and replace acked again that would throw IllegalStateException.

jbertram:
@michaelandrepearce, I've moved the ring semantics into `QueueImpl`.

@wy96f, I'm still looking at the issue you identified.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -576,15 +612,17 @@ public QueueImpl(final long id,
...
 
    // Bindable implementation -------------------------------------------------------------------------------------
 
    @Override
    public boolean allowsReferenceCallback() {
-      // non descructive queues will reuse the same reference between multiple consumers
+      // non destructive queues will reuse the same reference between multiple consumers
       // so you cannot really use the callback from the MessageReference
-      return !nonDestructive;
+      return !isNonDestructive();
```
michaelandrepearce: ? Why the chang. Its internal field and allows access by field reference

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2733,7 +2799,7 @@ private boolean deliver() {
    }
 
    protected void removeMessageReference(ConsumerHolder<? extends Consumer> holder, MessageReference ref) {
-      if (!nonDestructive) {
+      if (!isNonDestructive()) {
```
michaelpearce-gain: why the change here? again its internal to class, and hot path, thus was field access

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3436,6 +3503,11 @@ public void postAcknowledge(final MessageReference ref, AckReason reason) {
 
       queue.decDelivering(ref);
 
+      // this is done to tell the difference between actual acks and just a closed consumer in the non-destructive use-case
+      if (isNonDestructive() && reason == AckReason.NORMAL) {
```
michaelpearce-gain: field access?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1556,7 +1614,7 @@ public void acknowledge(final MessageReference ref, final ServerConsumer consume
 
    @Override
    public void acknowledge(final MessageReference ref, final AckReason reason, final ServerConsumer consumer) throws Exception {
-      if (nonDestructive && reason == AckReason.NORMAL) {
+      if (isNonDestructive() && reason == AckReason.NORMAL) {
```
michaelpearce-gain: was field access why the change, its hotpath and was internal field.

michaelpearce-gain: i realise it maybe in-lined but no guarantee.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -165,7 +165,7 @@
 
    private volatile boolean queueDestroyed = false;
 
-   private final PageSubscription pageSubscription;
+   protected final PageSubscription pageSubscription;
```
michaelpearce-gain: Is this needed?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -179,17 +179,17 @@
    private final MpscUnboundedArrayQueue<MessageReference> intermediateMessageReferences = new MpscUnboundedArrayQueue<>(8192);
 
    // This is where messages are stored
-   private final PriorityLinkedList<MessageReference> messageReferences = new PriorityLinkedListImpl<>(QueueImpl.NUM_PRIORITIES, MessageReferenceImpl.getIDComparator());
+   protected final PriorityLinkedList<MessageReference> messageReferences = new PriorityLinkedListImpl<>(QueueImpl.NUM_PRIORITIES, MessageReferenceImpl.getIDComparator());
```
michaelpearce-gain: is this needed?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -179,17 +179,17 @@
...
 
    // The quantity of pagedReferences on messageReferences priority list
    private final AtomicInteger pagedReferences = new AtomicInteger(0);
 
    // The estimate of memory being consumed by this queue. Used to calculate instances of messages to depage
    private final AtomicInteger queueMemorySize = new AtomicInteger(0);
 
-   private final QueuePendingMessageMetrics pendingMetrics = new QueuePendingMessageMetrics(this);
+   protected final QueueMessageMetrics pendingMetrics = new QueueMessageMetrics(this, "pending");
```
michaelpearce-gain: again is this needed, making it protected

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3933,6 +4005,37 @@ void scheduleSlowConsumerReaper(AddressSettings settings) {
       }
    }
 
+   private void enforceRing() {
```
michaelpearce-gain: could ringsize != -1 check be done here, will allow better escaping and inlining where not being used., e.g. avoids need to call into method below which is many lines long, and unlikely to be inlined

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -881,13 +919,17 @@ public void addHead(final MessageReference ref, boolean scheduling) {
       enterCritical(CRITICAL_PATH_ADD_HEAD);
       synchronized (this) {
          try {
-            if (!scheduling && scheduledDeliveryHandler.checkAndSchedule(ref, false)) {
-               return;
-            }
+            enforceRing(ref, scheduling);
```
michaelpearce-gain: can a if check be wrapped around this, to avoid the method (which method is large and unlikly to be inlined) call when ringsize is disabled (-1)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/GroupFirstMessageReference.java
```diff
@@ -191,6 +191,16 @@ public void handled() {
       messageReference.handled();
    }
 
+   @Override
+   public void setInDelivery(boolean deliveredDirectly) {
+
```
michaelpearce-gain: why is it not invoking?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3460,7 +3532,7 @@ public void postAcknowledge(final MessageReference ref, AckReason reason) {
          message = null;
       }
 
-      if (message == null)
+      if (message == null || (isNonDestructive() && reason == AckReason.NORMAL))
```
michaelpearce-gain: by field access

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -576,15 +612,17 @@ public QueueImpl(final long id,
       if (this.addressInfo != null && this.addressInfo.isPaused()) {
          this.pause(false);
       }
+
+      this.ringSize = ringSize == null ? (addressSettingsRepository != null ? addressSettingsRepository.getMatch(address.toString()).getDefaultRingSize() : -1) : ringSize;
```
michaelpearce-gain: doesnt seem to match the logic of all others, which is that address setting resolvement if not set is meant to occur well before create queue object (e.g. ownership of that logic is else where) to keep things standardised where looking for where things occur, and if null reaches here (not expected, but a safety is that it defaults to the constant.

Currently if it was null and addresssettingrepo is null, you will get NPE. which others dont.


e.g.

expecting something more inline with others:

example:
      this.maxConsumers = maxConsumers == null ? ActiveMQDefaultConfiguration.getDefaultMaxQueueConsumers() : maxConsumers;

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -335,6 +335,66 @@ public void testCreateAndDestroyQueue_4() throws Exception {
...
+      Assert.assertEquals(durable, queueControl.isDurable());
+      Assert.assertEquals(maxConsumers, queueControl.getMaxConsumers());
+      Assert.assertEquals(purgeOnNoConsumers, queueControl.isPurgeOnNoConsumers());
+      Assert.assertEquals(false, queueControl.isTemporary());
+      Assert.assertEquals(exclusive, queueControl.isExclusive());
+//      Assert.assertEquals(groupRebalance, queueControl.getGroupRebalance());
+//      Assert.assertEquals(groupBuckets, queueControl.getGroupBuckets());
+//      Assert.assertEquals(groupFirstKey, queueControl.getGroupFirstKey());
+      Assert.assertEquals(lastValue, queueControl.isLastValue());
+//      Assert.assertEquals(lastValueKey, queueControl.getLastValueKey());
```
michaelpearce-gain: commented out code......

michaelpearce-gain: either comment it in, or remove.

michaelpearce-gain:
Looking better, left a number of comments, one critical bug i found was its possible to raise NPE in QueueImpl constructor, this is because its not following the same pattern as all other fields, lets keep things uniform / follow existing convention.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelpearce-gain: (nit)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/GroupFirstMessageReference.java
michaelpearce-gain: (nit)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelpearce-gain: (nit)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelpearce-gain: (nit)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelpearce-gain: (nit)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelpearce-gain: (nit)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelpearce-gain: (nit)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelpearce-gain: (nit)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
jbertram: I added these as place-holders to remind me what we don't have implemented yet. I'd like to keep them if possible and then un-comment them when the functionality is implemented on the QueueControl.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
jbertram: Fair enough.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/GroupFirstMessageReference.java
jbertram: Fixing...

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
jbertram: Will do.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
jbertram: Will do.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
jbertram: These access changes were left-over from when RingQueue was subclassed. I'll fix them.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
jbertram: I remember changing this, but I don't remember why. :)  I'll change them back.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/PagedReferenceImpl.java
```diff
@@ -252,6 +252,16 @@ public void handled() {
       getQueue().referenceHandled(this);
    }
 
+   @Override
+   public void setInDelivery(boolean inDelivery) {
+
```
wy96f: This should also be implemented

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -974,6 +1018,8 @@ public void addTail(final MessageReference ref) {
    public void addTail(final MessageReference ref, final boolean direct) {
       enterCritical(CRITICAL_PATH_ADD_TAIL);
       try {
+         enforceRing();
```
wy96f: Should put enforceRing below schedule and direct deliver code?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -881,13 +919,19 @@ public void addHead(final MessageReference ref, boolean scheduling) {
       enterCritical(CRITICAL_PATH_ADD_HEAD);
       synchronized (this) {
          try {
-            if (!scheduling && scheduledDeliveryHandler.checkAndSchedule(ref, false)) {
-               return;
+            if (ringSize != -1) {
+               enforceRing(ref, scheduling);
```
wy96f: Should put enforceRing below schedule code?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3933,6 +4007,39 @@ void scheduleSlowConsumerReaper(AddressSettings settings) {
...
 
+   private void enforceRing() {
+      if (ringSize != -1) { // better escaping & inlining when ring isn't being used
+         enforceRing(null, false);
+      }
+   }
+
+   private void enforceRing(MessageReference refToAck, boolean scheduling) {
+      if (getMessageCountForRing() >= ringSize) {
+         refToAck = refToAck == null ? messageReferences.poll() : refToAck;
```
wy96f: If a queue is both LVQ and RING queue, when we poll ref from messageReferences, we should also remove it from LastValueQueue::map.

wy96f:
In LastValueQueue::addHead,
https://github.com/apache/activemq-artemis/blob/9d96dda336514d2b49e26e73f2c4c7bd83f6ea31/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LastValueQueue.java#L152
https://github.com/apache/activemq-artemis/blob/9d96dda336514d2b49e26e73f2c4c7bd83f6ea31/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LastValueQueue.java#L154,

if ref is replace acked due to over the ring size, we shouldn't put ref in LastValueQueue::map.

PR #2803, opened 2019-08-20
ARTEMIS-2458 Fix AMQP Transaction Rollback Ordering

michaelandrepearce:
Ensure cancel of in-flight message, occurs before rollback, to ensure ordering guarantee. As like with Core.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
@@ -405,8 +405,15 @@ public class AMQPSessionCallback implements SessionCallback {
    public void cancel(Object brokerConsumer, Message message, boolean updateCounts) throws Exception {
       OperationContext oldContext = recoverContext();
       try {
-         ((ServerConsumer) brokerConsumer).individualCancel(message.getMessageID(), updateCounts);
-         ((ServerConsumer) brokerConsumer).getQueue().forceDelivery();
+         try {
+            ((ServerConsumer) brokerConsumer).individualCancel(message.getMessageID(), updateCounts);
+            ((ServerConsumer) brokerConsumer).getQueue().forceDelivery();
+         } catch (IllegalStateException e) {
+            //Ignore this silently, can occur if already cancelled.
+            if (logger.isDebugEnabled()) {
+               logger.debug("silently ignoring cancel", e);
+            }
+         }
       } finally {
          resetContext(oldContext);
       }
@@ -647,6 +654,10 @@ public class AMQPSessionCallback implements SessionCallback {
       } else {
          return false;
       }
+   }
+
+   public ServerSession getServerSession() {
+      return serverSession;
    }
 
    @Override
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/transaction/ProtonTransactionHandler.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/transaction/ProtonTransactionHandler.java
@@ -145,7 +145,7 @@ public class ProtonTransactionHandler implements ProtonDeliveryHandler {
             };
 
             if (discharge.getFail()) {
-               sessionSPI.withinContext(() -> tx.rollback());
+               sessionSPI.withinContext(() -> sessionSPI.getServerSession().doRollback(false, false, tx));
                sessionSPI.afterIO(ioAction);
             } else {
                sessionSPI.withinContext(() -> tx.commit());
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ServerSession.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ServerSession.java
@@ -286,6 +286,10 @@ public interface ServerSession extends SecurityAuth {
                       RoutingContext routingContext) throws Exception;
 
 
+   void doRollback(boolean clientFailed,
+                 boolean lastMessageAsDelived,
+                 Transaction theTx) throws Exception;
+
    RoutingStatus doSend(Transaction tx,
                         Message msg,
                         SimpleString originalAddress,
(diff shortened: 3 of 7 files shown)


clebertsuconic:
This might be valid on qpid-jms, but I'm not sure this is valid on the spec itself.

You may have other cases where a rollback should not issue a server's cancellation. I'm not sure this is valid.. as it may break other cases.


It's probably ok with qpid-jms, but you would need some soft of annotation to make sure this is valid.

michaelandrepearce:
The fact a session is rolled back, any inflight must be cancelled.

clebertsuconic:
@michaelandrepearce I know that's true for JMS. I'm just asking if that's true for other implementations. is there anything on the AMQP spec that guarantees that? 

I'm just trying to avoid breaking C++/.NET/ other implementations.

michaelandrepearce:
We found the issue in .NET , and recreated because its easier to show in QPID JMS as artemis is Java and we have the integration test capability. 
The test actually comes from QPID though they test against ActiveMQ5 where behaviour is correct, like wise correct behaviour when pointing .net at activemq5.

clebertsuconic:
@michaelandrepearce activemq5 is not removing the message from the queue. It's a different implementation.

I'm just trying to find if we should either add a method on the LinkedLilst such as addBackInPlace. However if this is not breaking other cases.. I'm 100% on it. I just want to make sure this is not breaking other cases.

michaelandrepearce:
Well i beleive our test suite has a number of non-JMS AMQP tests also right, so as long as all pass.....
As said issue was actually found outside of Java land ;) ... just re-creatable in it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/transaction/ProtonTransactionHandler.java
```diff
@@ -165,6 +170,16 @@ public void onError(int errorCode, String errorMessage) {
       }
    }
 
+   private void cancelRefs(ProtonTransactionImpl tx) throws Exception {
+      List<MessageReference> toCancel = new ArrayList<>();
```
michaelandrepearce: The only bit i wonder if we need to add here is the extra logic of stopping and restarting the consumer as like in ServerSessionImpl::cancelAndRollback

michaelandrepearce: actually will be cleaner just to invoke rollback in serversession impl, then will have same logic, and will gain from having any other fixes.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
```diff
@@ -405,7 +405,11 @@ public void ack(Transaction transaction, Object brokerConsumer, Message message)
    public void cancel(Object brokerConsumer, Message message, boolean updateCounts) throws Exception {
       OperationContext oldContext = recoverContext();
       try {
-         ((ServerConsumer) brokerConsumer).individualCancel(message.getMessageID(), updateCounts);
+         try {
+            ((ServerConsumer) brokerConsumer).individualCancel(message.getMessageID(), updateCounts);
+         } catch (IllegalStateException e) {
+            //Ignore this silently, can occur if already cancelled.
```
clebertsuconic: nit-pick: can you log.debug here instead?

Although I could ammend it myself during merging

clebertsuconic:
I think this is totally fine. This is just using the exact semantics for rollback than any other protocol in Artemis.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
michaelandrepearce: sure will add

michaelandrepearce: added - just pushed

clebertsuconic:
I have sent an alternative, in case this violates AMQP specification (in the case I was mentioning it could break other clients).

The PR is #2804

clebertsuconic:
Marking a label DO-NOT-MERGE-YET just to have a discussion on what approach is better.

gaohoward:
The spec doesn't specify what order should be kept after rollback. So the best thing to do is that the client should not reply on this behavior.

michaelandrepearce:
@howardgao i dont think this is a client issue, the fact the issue is seen in different produced clients (e.g. different language and teams creating) and the fact the issue is not seen when using these clients both in ActiveMQ5 and pointing at some propriety vendor brokers also, suggests it is indeed Artemis's issue.

michaelandrepearce:
@clebertsuconic i think your solution is more robust, as will provide the guarentee for anything else in future. Also with a little additional work, it can fix easily another ordering issue when the session is closed in AMQP. Ill fork your's add commit and PR that.

michaelandrepearce:
https://github.com/apache/activemq-artemis/pull/2805

gemmellr:
I know other solutions have been proposed now, I've not yet looked at them but will try to later, but for completeness of discussion here...

The issue definitely isn't with the client, which is doing what is expected and required from a JMS and AMQP perspective (and as noted works fine against a variety of other servers), but as suggested is rather with the behaviour Artemis introduces itself during the overall process.

I believe Clebert was quite right to be concerned about this change however. If this does what I think (and for it to make any different at all I'd guess must be, or something similar) then it wouldn't be a valid thing to do from an AMQP perspective, though I expect it would behave ok with Qpid JMS due to the clients overall [JMS] behaviour. AMQP transactions aren't session-wide the way e.g JMS ones are, they can mix with non-transacted delivery, there can be multiple transactions on a session at a time, or they can also (/do in some other clients) span multiple sessions, etc. As such rolling back an AMQP transaction does not itself affect delivery of messages that were never listed (by the client, in this and almost all cases) as being part of that transaction. There are also cases (which some other clients can/do also make use of) where even rolling back the transaction would still leave messages that were part of it live and allocated to the client.

clebertsuconic:
This is replaced by #2805 

Thanks @gemmellr for the review

PR #2804, opened 2019-08-21
ARTEMIS-2458 Fix AMQP Transaction Rollback Ordering by using a sorted…

clebertsuconic:
… add


michaelandrepearce:
I think your solution is actually better, there is actually another issue spotted when on session close, we also get ordering issue, a small additonal modification fixes that easily. have added a commit to yours and sent a new pr with both in https://github.com/apache/activemq-artemis/pull/2805

clebertsuconic:
This is replaced by #2805 

It will be safer

PR #2805, opened 2019-08-21
ARTEMIS-2458 Fix AMQP Transaction Rollback & Session Close Ordering 

michaelandrepearce:


clebertsuconic:
@michaelandrepearce it's easier to ask.. did you happen to change to semantics on core (to also use addSorted?) or that's preserved.

I'm not sure yet.. I'm reading in more detail to figure out.. but your answer would help me go faster :)

clebertsuconic:
@franz1981 I actually like the randomicity of the test in there. Just like our users, data is always random :)

clebertsuconic:
WHOA: This is a LinkedList of PRs that is changing a LinkedList... :) Seriously! :)

I have made an optimization to the addSorted on LinkedList, fixed a few check-style mistakes I made, and I sent a new PR (hopefully final now).

This is replaced by #2807

franz1981:
> @franz1981 I actually like the randomicity of the test in there. Just like our users, data is always random :)

Yep it is true, but it won'ìt help if the test start to fail, because we would like to make it fail 100% of the time and reproduce it locally too. Ok to make it random, but maybe it would be nice to report which random sequence has been used on test failure...

PR #2808, opened 2019-08-21
NO-JIRA fix log message format in ActiveMQRALogger

pazepaze:
MessageFormat uses single quotes to escape so string ${0} was logged.


clebertsuconic:
I have no way to test this, but I will trust you :)

PR #2809, opened 2019-08-26
ARTEMIS-2454 Fixing body re-encoding

clebertsuconic:


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -228,20 +227,23 @@ public MessageImpl getProtonMessage() {
...
     */
    public Header getHeader() {
-      ensureMessageDataScanned();
-      ensureDataIsValid();
+      ensureScanning();
       return scanForMessageSection(headerPosition, Header.class);
    }
 
+   private void ensureScanning() {
+      ensureDataIsValid();
```
clebertsuconic: Before someone asks.. the inversion of order here was on purpose...

We need to first ensure the data buffer is valid, and if needed to be re-encoded...

Then after the re-encoding, we need to re-scan the data.

PR #2813, opened 2019-08-27
ARTEMIS-2462 Allow store-forward queue to be deleted afte scaledown

gaohoward:
After a node is scaled down to a target node, the sf queue in the
target node is not deleted.

Normally this is fine because may be reused when the scaled down
node is back up.

However in cloud environment many drainer pods can be created and
then shutdown in order to drain the messages to a live node (pod).
Each drainer pod will have a different node-id. Over time the sf
queues in the target broker node grows and those sf queues are
no longer reused.

Although use can use management API/console to manually delete
them, it would be nice to have an option to automatically delete
those sf queue/address resources after scale down.

    
    In this PR it added a boolean configuration parameter called
    cleanup-sf-queue to scale down policy so that if the parameter
    is "true" the broker will send a message to the
    target broker signalling that the SF queue is no longer
    needed and should be deleted.
    
    If the parameter is not defined (default) or is "false"
    the scale down won't remove the sf queue.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
@@ -430,6 +430,9 @@ public final class ActiveMQDefaultConfiguration {
    // its possible that you only want a server to partake in scale down as a receiver, via a group. In this case set scale-down to false
    private static boolean DEFAULT_SCALE_DOWN_ENABLED = true;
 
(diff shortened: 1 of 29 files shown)


michaelandrepearce:
Can this be something we set in broker.xml. try avoid having system property configurations.

gaohoward:
ok I'm fine with it. I'll change it to broker.xml. thx guys.

gaohoward:
@michaelandrepearce @clebertsuconic  I've done refactoring and will watch the jenkins result.

clebertsuconic:
it looks alright... nice job!

let me run tests on it before merging.

PR #2814, opened 2019-08-28
ARTEMIS-2465 [JDBC-STORE] Adding index on txId

franz1981:


franz1981:
Please don't merge it yet: I need to test it with some different DBMS and see that's ok :+1:

PR #2817, opened 2019-08-29
Network check custom ping priority

twood2k:
this change allows a custom ping command to run during network failure and after network is recovered.
Our custom ping app does a lot of checking and is used to delay the start of artemis after a network failure is resolved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,4 @@ docs/hacking-guide/en/_book
 
 # overlay outpit
 **/overlays/**/*
+artemis-server/.sts4-cache/classpath-data.json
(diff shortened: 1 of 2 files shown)


clebertsuconic:
You made a mess on your PR here.. you don't need to keep every single commit.. you can just squash all your changes and keep one change only.  but I was able to squash them and keep your ownership.

I'm adding a test.. and I'm merging after I cleaned it up. but in the future (that's for any git based project.. not just artemis), you should cleanup your PR. you could:

```
# 5 here is number of commits you have
# use the list to squash them, google for git rebase -i, you will go through an editor such as vi
git rebase -i HEAD~5 
git push origin <your_branch_name> -f
```

twood2k:
so sorry. this is my first time using git. only have used subversion and
tfs on real projects.
thanks for the help and cleaning things up and ill try not to let it happen
again.

ugh, didnt even think about the test

thanks again and take care.

-tom

PR #2818, opened 2019-08-29
ARTEMIS-2467 Allowing create of static clusters through CLI

clebertsuconic:


clebertsuconic:
there is something broken .. I don't want to close it / reopen it... I will fix it tomorrow.

PR #2821, opened 2019-09-04
ARTEMIS-2471 JdbcNodeManager doesn't use jdbc-user and jdbc-password tags

franz1981:


jbertram:
Checkstyle violations:

```
[ERROR] src/test/java/org/apache/activemq/artemis/core/server/impl/jdbc/JdbcNodeManagerTest.java:[13,8] (imports) UnusedImports: Unused import - org.apache.activemq.artemis.jdbc.store.drivers.JDBCUtils.
[ERROR] src/test/java/org/apache/activemq/artemis/core/server/impl/jdbc/JdbcNodeManagerTest.java:[14,8] (imports) UnusedImports: Unused import - org.apache.activemq.artemis.jdbc.store.sql.SQLProvider.
```

PR #2828, opened 2019-09-09
ARTEMIS-2478 Expired message not removed in non destructive queue

wy96f:


clebertsuconic:
I ran the whole testsuite (2.5 hours run) with this change, no issues! merging!

clebertsuconic:
Actually, there's one issue (it seems the test added takes a long time to run.. 1 minute). I'm looking if we can speed it up before merging.. i will get back to you on that.

PR #2829, opened 2019-09-09
ARTEMIS-2475 log metrics plugin init details

jbertram:


clebertsuconic:
LGTM. I would merge now if I could.

PR #2830, opened 2019-09-09
ARTEMIS-2479 Initial connection will not work with infinite retry

clebertsuconic:


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
```diff
@@ -780,23 +798,25 @@ public ClientSessionFactory createSessionFactory() throws ActiveMQException {
...
                   addToConnecting(factory);
-                  factory.connect(initialConnectAttempts, failoverOnInitialConnection);
+                  // We always try to connect here with only one attempt,
+                  // as we will perform the initial retry here, looking for all possible connectors
+                  factory.connect(1, false);
                } finally {
                   removeFromConnecting(factory);
                }
             } catch (ActiveMQException e) {
+               if (factory.waitForRetry(retryInterval)) {
```
wy96f: Can we not wait and just throw exception if exception type is NOT_CONNECTED or attempts exceed?

clebertsuconic: I put it before as the factory close wouldn't let me wait

I modified the code slightly to what I think it's best


I'm still working on a test failure on this PR, I will remove the DO-NOT-MERGE-YET when ready.

clebertsuconic:
ready!

PR #2832, opened 2019-09-10
ARTEMIS-2482 Large messages could leak native ByteBuffers

franz1981:
Perform chunked read/write of large message files to save
NIO from leaking native ByBuffers:
see https://bugs.openjdk.java.net/browse/JDK-8147468

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
@@ -31,6 +31,8 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.PooledByteBufAllocator;
 import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
 import org.apache.activemq.artemis.api.core.ActiveMQIllegalStateException;
@@ -45,6 +47,7 @@ import org.apache.activemq.artemis.core.io.SequentialFile;
 import org.apache.activemq.artemis.core.io.SequentialFileFactory;
 import org.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory;
 import org.apache.activemq.artemis.core.io.mapped.MappedSequentialFileFactory;
+import org.apache.activemq.artemis.core.io.nio.NIOSequentialFile;
 import org.apache.activemq.artemis.core.io.nio.NIOSequentialFileFactory;
 import org.apache.activemq.artemis.core.journal.Journal;
 import org.apache.activemq.artemis.core.journal.impl.JournalFile;
@@ -832,6 +835,31 @@ public class JournalStorageManager extends AbstractJournalStorageManager {
       }
    }
 
+   static final int LARGE_MESSAGE_CHUNK_SIZE = 100 * 1024;
+
+   private void addBytesUsingTempNativeBuffer(final SequentialFile file, final ActiveMQBuffer bytes) throws Exception {
+      assert file instanceof NIOSequentialFile;
+      //we can't use the actual content of it as it is and need to perform a copy into a direct ByteBuffer
+      int readableBytes = bytes.readableBytes();
+      final int requiredCapacity = Math.min(LARGE_MESSAGE_CHUNK_SIZE, readableBytes);
+      final ByteBuf tempBuffer = PooledByteBufAllocator.DEFAULT.directBuffer(requiredCapacity, requiredCapacity);
+      try {
+         int readerIndex = bytes.readerIndex();
+         while (readableBytes > 0) {
+            final int size = Math.min(readableBytes, LARGE_MESSAGE_CHUNK_SIZE);
+            final ByteBuffer nioBytes = tempBuffer.internalNioBuffer(0, size);
+            final int position = nioBytes.position();
+            bytes.getBytes(readerIndex, nioBytes);
+            nioBytes.position(position);
+            file.blockingWriteDirect(nioBytes, false, false);
+            readerIndex += size;
+            readableBytes -= size;
+         }
+      } finally {
+         tempBuffer.release();
+      }
+   }
+
    public final void addBytesToLargeMessage(final SequentialFile file,
                                             final long messageId,
                                             final ActiveMQBuffer bytes) throws Exception {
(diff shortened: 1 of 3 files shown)


clebertsuconic:
@franz1981 do we have any other reading operation that would lead the same?

Can you also change our scripts to include the property to avoid caching if anything like that is ever used?

franz1981:
On paging we cache manually so is a different "issue" and probably for many user is fine.
Re the property I'm not sure: imo is a good idea but it will impact existing users where it is happening without some; they will start to perceive slower performance in a stealthy way..
Anyway I don't have a strong opinion on that so will do it if you prefer (just need to verify if is present in jdk 11 as well)

clebertsuconic:
We shouldn't be using native cache anyways... just in case it happens in our codebase, i would rather have a limit.

clebertsuconic:
@franz1981 can you add labels on github? use the DO-NOT-MERGE-YET?

franz1981:
@clebertsuconic seems that I cannot add the label...please check the code, because I've found another part on replication where it could happen...

franz1981:
@wy96f please take a look that you've recently have fun with `ByteBuf`s :)

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -832,46 +835,117 @@ public void stopReplication() {
...
-
-            if (isReplicated()) {
-               //copy defensively bytes
-               final byte[] bytesCopy = new byte[bytes.readableBytes()];
-               bytes.getBytes(bytes.readerIndex(), bytesCopy);
-               replicator.largeMessageWrite(messageId, bytesCopy);
+            //NIO -> need direct ByteBuffers, while JDBC the opposite
+            if (file instanceof NIOSequentialFile) {
+               if (bytes.byteBuf().isDirect()) {
+                  final ByteBuffer nioBytes = bytes.byteBuf().internalNioBuffer(bytes.readerIndex(), bytes.readableBytes());
```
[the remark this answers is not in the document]
franz1981: it is on https://github.com/apache/activemq-artemis/pull/2832/files/4fc608a382d9d1d7eb0401e808141c7f939b21f9#diff-6c51f3299aaf6c901fe05c48fcfa3bb7R869

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -832,46 +835,117 @@ public void stopReplication() {
...
+                  file.blockingWriteDirect(nioBytes, false, false);
+               } else {
+                  addBytesUsingTempNativeBuffer(file, bytes);
+               }
+            } else if (!bytes.byteBuf().isDirect()) {
+               final ByteBuffer nioBytes = bytes.byteBuf().internalNioBuffer(bytes.readerIndex(), bytes.readableBytes());
```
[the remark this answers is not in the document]
franz1981: it is on https://github.com/apache/activemq-artemis/pull/2832/files/4fc608a382d9d1d7eb0401e808141c7f939b21f9#diff-6c51f3299aaf6c901fe05c48fcfa3bb7R869

franz1981:
@wy96f 

> PooledByteBufAllocator::directBuffer uses java's ThreadLocal. Wouldn't this result in leak in the case if thread terminates due to idle timeout as in #2199 ?

Do you mean the thread local arena? Or the thread local NIO ByteBuffer?
Anyway, at the end of its usage we always release the `ByteBuf` and Netty will take care to release any referenced resources: it's a pool so it leaks "by definition" :+1:

wy96f:
> @wy96f
> 
> > PooledByteBufAllocator::directBuffer uses java's ThreadLocal. Wouldn't this result in leak in the case if thread terminates due to idle timeout as in #2199 ?
> 
> Do you mean the thread local arena? Or the thread local NIO ByteBuffer?
> Anyway, at the end of its usage we always release the `ByteBuf` and Netty will take care to release any referenced resources: it's a pool so it leaks "by definition"

I mean tinySubPageDirectCaches/smallSubPageDirectCaches/normalDirectCaches in PoolThreadCache. PoolChunk would be added into these caches first if released(not yet into shared arena). Given PoolThreadCache is thread local, would theses caches leak if thread terminates?

franz1981:
@wy96f Yes and no :)
The would leak (at first look, but let me take a better look), but they would contain just wrappers to direct memory that cannot be released (is part of the pool) and no cleaner (so they won't collect anything, when contains ByteBuffers). The impact should be way less then `IOUtil` that pool native `ByteBuffer`s holding exclusively native memory that won't be reused anymore...

franz1981:
@wy96f I've taken a second look to the way Netty handle thread local pools and I can confirm that `-Dio.netty.allocator.useCacheForAllThreads=false` should save non-netty threads from creating such thread local regions for tiny/small/medium sized buffers.
While related to 
> but they just contain wrappers to direct memory that cannot be released (is part of the pool and they have no cleaner). The impact should be way less then IOUtil that pool native ByteBuffers holding exclusively native memory that won't be reused anymore...

I have to add more detail and I was wrong: the tiny/small/medium thread local caches are actually leaking memory regions (with default `io.netty.allocator.useCacheForAllThread`) that need the holding thread to die in order to make the available to other threads, but their size is quite small by default and the size I've chosen as `LARGE_MESSAGE_CHUNK_SIZE` (ie 100 * 1024) is `medium` but doesn't fall in any of those caches, so no real leak will happen until we allocate something smaller (could happen but is rare).
IMO we should consider (for a separate issue) 2 separate improvements:
- allow threads to stay alive forever, but just idle (and maybe using the FJ thread pool executor instead of the AMQ thread pool executor in that case): in that case we can consider such caches to be ok to be used by *all* threads (including non-netty ones), because the leak is just to cope with future load
- for the AMQ thread pool executor as it is, make the presence of such caches configurable and off by default

wdyt?
Thanks for the comment, good catch!

wy96f:
@franz1981 

> I have to add more detail and I was wrong: the tiny/small/medium thread local caches are actually leaking memory regions (with default io.netty.allocator.useCacheForAllThread) that need the holding thread to die in order to make them available to other threads again

I assumed memory released into tiny/small/medium cache would be leaked if holding thread terminates until I looked at the code in PoolThreadCache:
```
/// TODO: In the future when we move to Java9+ we should use java.lang.ref.Cleaner.
    @Override
    protected void finalize() throws Throwable {
        try {
            super.finalize();
        } finally {
            free();
        }
    }

    /**
     *  Should be called if the Thread that uses this cache is about to exist to release resources out of the cache
     */
    void free() {
        // As free() may be called either by the finalizer or by FastThreadLocal.onRemoval(...) we need to ensure
        // we only call this one time.
        if (freed.compareAndSet(false, true)) {
            int numFreed = free(tinySubPageDirectCaches) +
                    free(smallSubPageDirectCaches) +
                    free(normalDirectCaches) +
                    free(tinySubPageHeapCaches) +
                    free(smallSubPageHeapCaches) +
                    free(normalHeapCaches);

            if (numFreed > 0 && logger.isDebugEnabled()) {
                logger.debug("Freed {} thread-local buffer(s) from thread: {}", numFreed,
                        Thread.currentThread().getName());
            }

            if (directArena != null) {
                directArena.numThreadCaches.getAndDecrement();
            }

            if (heapArena != null) {
                heapArena.numThreadCaches.getAndDecrement();
            }
        }
    }
```
So the cache will be freed and used by other threads when thread dies? If so, it seems no any leak problems with this pr, no need with improvements?

franz1981:
@wy96f 
> So the cache will be freed and used by other threads when thread dies?

Yes, it is, but my concern is that it needs a finalization run before getting there ie not that deterministic as I wish :)

wy96f:
As you said, size of tiny/small/medium caches is quite small. Even if not freed in time, the impact is very little compared to IOUtil that pool native ByteBuffers(default TEMP_BUF_POOL_SIZE=8) for each thread :)

franz1981:
@wy96f fair enough :)
Still waiting some results from the integration tests on the CI and will be ready to be merged, although I prefer to improve code quality of this PR, there are some changes due to JDBC that makes is uglier than I wished

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
```diff
@@ -511,6 +512,44 @@ protected void closeFile() throws Exception {
       }
    }
 
+   private static int read(final SequentialFile file, final ByteBuffer bufferRead) throws Exception {
```
wy96f: Given this is a hot path and in most cases(LargeMessageDeliverer::deliver, ClientProducerImpl::largeMessageSendServer, CoreMessage::getLargeMessageBuffer, etc) "bufferRead" is a heap buffer, can we construct "bufferRead" by using PooledByteBufAllocator before calling to save from coping buffer?

franz1981: @wy96f 
It was the second point on my list :) 
Yes, it can be done, but deserve lot of attention and probably some changes on the API on SessionCallback...

franz1981: @wy96f I still think that `Page` has the same issue and should use Netty pools instead of retaining direct memory as a state

wy96f: Do you mean readFileBuffer in Page? In most cases, we sequentially read messages from page and readFileBuffer might contains several messages(depends on message size) in buffer. This way we can just read following messages from buffer(no need bothering file system). If we use netty pools, we need to read the file every time we read a message, right? Will this somewhat affect the perf?

franz1981: > If we use netty pools, we need to read the file every time we read a message, right? Will this somewhat affect the perf?
We need to find a proper solution, but is a leak and we use a heap buffer too there so it means a leak on IOUtil too...
I see the same on decoding side for replicated files and on encoding side for non netty connections...basically everytime we performa FileChannel ops with heap ByteBuffers.

wy96f: Agreed :)

franz1981:
@clebertsuconic @michaelandrepearce Thanks to the work of @wy96f on replication, I've taken a further look into https://docs.oracle.com/javase/7/docs/api/java/io/RandomAccessFile.html#getChannel() and I think this issue can be solved much more easily by using `RandomAccessFile` when necessary (to write/read byte[] without leaks :))
I will provide soon a PR for it, but I can close this one, that is very complex and uneasy to be maintained

PR #2833, opened 2019-09-10
ARTEMIS-2483 Sync calculation should use the block size

clebertsuconic:


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/util/SyncCalculation.java
```diff
@@ -77,7 +77,10 @@ public static long syncTest(File datafolder,
                                int maxAIO,
                                JournalType journalType) throws Exception {
       SequentialFileFactory factory = newFactory(datafolder, fsync, journalType, blockSize * blocks, maxAIO);
-      final boolean asyncWrites = journalType == JournalType.ASYNCIO && !syncWrites;
+
+      if (factory instanceof AIOSequentialFileFactory) {
+         factory.setAlignment(blockSize);
```
clebertsuconic: Before anyone asks, I have no easy way to create a testcase for this. This is CLI, and hence this is like an UI tool.

besides these defaults are dependent on platform, so I am on purpose not writing a testcase for this change.

PR #2834, opened 2019-09-13
ARTEMIS-2488: Handle the case where source address is null

lulf:


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
```diff
@@ -319,7 +319,7 @@ public void initialise() throws Exception {
          if (CompositeAddress.isFullyQualified(source.getAddress())) {
             addressToUse = SimpleString.toSimpleString(CompositeAddress.extractAddressName(source.getAddress()));
             queueNameToUse = SimpleString.toSimpleString(CompositeAddress.extractQueueName(source.getAddress()));
-         } else {
+         } else if (source.getAddress() != null) {
```
clebertsuconic: Test?

michaelandrepearce: An alternate is to use the static method on SimpleString to convert String to SimpleString which has the null check.

lulf:
@clebertsuconic @michaelandrepearce Updated with your suggestions.

artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContextTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.activemq.artemis.core.server.AddressQueryResult;
+import org.apache.activemq.artemis.protocol.amqp.broker.AMQPSessionCallback;
+import org.apache.activemq.artemis.protocol.amqp.exceptions.ActiveMQAMQPNotFoundException;
+import org.apache.qpid.proton.amqp.messaging.Source;
+import org.apache.qpid.proton.engine.Sender;
+import org.junit.Test;
+
+import java.util.Collections;
+
+import static org.mockito.Mockito.*;
```
clebertsuconic: this will fail Checkstyle.. can you fix it?

lulf: Fixed, sorry about that!

PR #2836, opened 2019-09-13
[ARTEMIS-2487] Updated to org.jctools:jctools-core:2.1.2

MrEasy:
To resolve wrong import range due to wrong declaration in
jctools-core:2.1.1


wy96f:
Good job!

PR #2837, opened 2019-09-16
ARTEMIS-2476: New MQTT subscriptions receive older (not last published) retained message.

assens:


michaelandrepearce:
Whilst im not an mqtt person so cannot review code from a correctness pov, will need someone else to. As a standard requirement can a test case please be added that exposes the issues and proves the fix.

assens:
> Whilst im not an mqtt person so cannot review code from a correctness pov, will need someone else to. As a standard requirement can a test case please be added that exposes the issues and proves the fix.

I have created a project which demonstrates the problem. It's described in the related issue https://issues.apache.org/jira/browse/ARTEMIS-2476

I'll try to create a relevant unit test inside the artemis-mqtt-protocol maven module.

assens:
I've provided the MQTTRetainMessageManagerTest which asserts that retained messages received on a newly created subscriptions are matching the last published message on the given topic. 
@clebertsuconic

artemis-protocols/artemis-mqtt-protocol/src/test/java/org/apache/activemq/artemis/core/protocol/mqtt/MqttClientService.java
```diff
@@ -0,0 +1,126 @@
...
+import org.eclipse.paho.client.mqttv3.MqttClient;
+import org.eclipse.paho.client.mqttv3.MqttConnectOptions;
+import org.eclipse.paho.client.mqttv3.MqttException;
+import org.eclipse.paho.client.mqttv3.MqttMessage;
+import org.eclipse.paho.client.mqttv3.persist.MemoryPersistence;
+import org.jboss.logging.Logger;
+
+import io.netty.util.concurrent.DefaultThreadFactory;
+
+public class MqttClientService implements MqttCallback {
```
clebertsuconic: Can you explain me what is this? just to make it easier for me to understand?


Is this a server with only the MQTT portion?

I can certainly hack your code and try to udnerstand.. just asking some minimal english explanation to make my review easier.

assens: Hi. The MQTTRetainMessageManagerTest uses [Eclipse Paho](https://www.eclipse.org/paho/) MQTT client to connect to Artemis. MqttCallback is an interface defined by the Paho client, that provides a call back when a message is received. 

The MqttClientService used by the MQTTRetainMessageManagerTest uses the Paho MQTT client to publish messages and at the same time implements the MqttCallback to be able to subscribe and receive messages.

The MQTTRetainMessageManagerTest uses 4 instances of MqttClientService. 

- mqttConsumerCount is used only to count the messages received by it as consumer
- mqttConsumerBeforePublish is a consumer which subscribes to the fact topic, before any publishing starts. 
- arrivedCountAferPublish and arrivedCountAferPublish2 are consumers that subscribe after messages are published.

See how the actAndAssert method uses the 4 clients. 

The goal of the test is to assert that retained message received by all client subscriptions  (before and after the publish) is equal to the last retained message published.

In addition, I've created a [mqtt-test](https://github.com/assens/mqtt-test) project that demonstrates the problem we have with Artemis, and also shows that other brokers such as Ejabberd, Emqx, HiveMQ, Moquette, Mosquitto, RabbitMQ and VerneMQ don't have such problem.

clebertsuconic:
I will merge this after I release 2.10.1, ok?

assens:
> I will merge this after I release 2.10.1, ok?

Thanks. We already use a patched version in our project, but I'd like to update to a release version, as soon as it is out.

michaelandrepearce:
@clebertsuconic i think you promised @assens to merge this, are you still planning to? i can if not

csimhadri:
@clebertsuconic  @michaelandrepearce  do we have plans on merging this soon?

michaelandrepearce:
I'm merging this @clebertsuconic  i see no reason why not.

PR #2839, opened 2019-09-16
[ARTEMIS-2490] Prevent NumberFormatExc when reading large message

MrEasy:


artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -369,12 +369,11 @@ protected SequentialFile createFileForLargeMessage(final long messageID, final b
       LargeMessagePersister.getInstance().decode(buff, largeMessage);
 
       if (largeMessage.containsProperty(Message.HDR_ORIG_MESSAGE_ID)) {
-         // for compatibility: couple with old behaviour, copying the old file to avoid message loss
-         long originalMessageID = largeMessage.getLongProperty(Message.HDR_ORIG_MESSAGE_ID);
```
wy96f: Hi, I wonder in which case NumberFormatExc would be thrown because of the fact that message has already contained property HDR_ORIG_MESSAGE_ID.

ecki: i answered in jira, or do you prefer to discuss here (short form, reason for jms id in orig core property is unclear)

MrEasy: Problem is that you can set the Core-message-ID and feed it with a non-number. This then fails when reading it on journal initialization.
The format "ID:<UUID>" is used as JMS-message-id (in contrast to core-message-id), so this seems to end-up sometimes as core-message-id as well.

In any case, I would like to request the retrieval to be moved inside the compatibility block, since it is just unneeded outside.

clebertsuconic: test?

MrEasy: Any request to change the commit or other considerations?

jbertram:
This needs a test to validate the fix and prevent future regressions. There's also a conflict now which needs to be resolved. 

I'm closing this, but please feel free to re-open it with the requested changes if you need this fix. Thanks!

jbertram:
For what it's worth, I prefer [this fix](https://github.com/seeburger-ag/hornetq/commit/8ef99744f219b23d76803435201e4af33a53b3c5) to the one offered in this PR as it actually prevents any exception and logs a `WARN`. I would even consider merging that without a test.

ecki:
We can provide you with a new request for it, however what should happen with the id if it fails to find a long? We need help on how to actually construct such a broken message for adding a test.

PR #2840, opened 2019-09-17
ARTEMIS-2493 OpenWire session close doesn't cleanup consumer refs

gaohoward:
When an openwire client closes the session, the broker doesn't
clean up its server consumer references even though the core
consumers are closed. This results a leak when sessions within
a connection are created and closed when the connection keeps open.


clebertsuconic:
I'm running whole testsuite before we can merge this.

gaohoward:
thanks @clebertsuconic

PR #2844, opened 2019-09-20
ARTEMIS-1811 NIO Seq File should use RandomAccessFile with heap buffers

franz1981:
It use RandomAccessFile to allow using heap buffers without additional
copies and/or leaks of direct buffers, as performed by FileChannel JDK
implementation (see https://bugs.openjdk.java.net/browse/JDK-8147468)

### DIFF
--- a/artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/nio/NIOSequentialFile.java
+++ b/artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/nio/NIOSequentialFile.java
@@ -18,10 +18,10 @@ package org.apache.activemq.artemis.core.io.nio;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.RandomAccessFile;
 import java.nio.ByteBuffer;
 import java.nio.channels.ClosedChannelException;
 import java.nio.channels.FileChannel;
-import java.nio.file.StandardOpenOption;
 import java.util.List;
 import java.util.concurrent.Executor;
 
@@ -43,6 +43,8 @@ import org.apache.activemq.artemis.utils.Env;
 public class NIOSequentialFile extends AbstractSequentialFile {
 
    private FileChannel channel;
+
+   private RandomAccessFile rfile;
 
    private final int maxIO;
 
@@ -82,7 +84,9 @@ public class NIOSequentialFile extends AbstractSequentialFile {
    @Override
    public void open(final int maxIO, final boolean useExecutor) throws IOException {
       try {
-         channel = FileChannel.open(getFile().toPath(), StandardOpenOption.CREATE, StandardOpenOption.WRITE, StandardOpenOption.READ);
+         rfile = new RandomAccessFile(getFile(), "rw");
+
+         channel = rfile.getChannel();
 
          fileSize = channel.size();
       } catch (ClosedChannelException e) {
@@ -139,18 +143,27 @@ public class NIOSequentialFile extends AbstractSequentialFile {
       super.close();
 
       try {
-         if (channel != null) {
-            if (waitSync && factory.isDatasync())
-               channel.force(false);
-            channel.close();
+         try {
+            if (channel != null) {
+               if (waitSync && factory.isDatasync())
+                  channel.force(false);
+               channel.close();
+            }
+         } finally {
+            if (rfile != null) {
+               rfile.close();
+            }
          }
       } catch (ClosedChannelException e) {
          throw e;
       } catch (IOException e) {
          factory.onIOError(new ActiveMQIOErrorException(e.getMessage(), e), e.getMessage(), this);
          throw e;
+      } finally {
+         channel = null;
+         rfile = null;
       }
-      channel = null;
+
 
       notifyAll();
    }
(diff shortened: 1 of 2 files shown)


franz1981:
@wy96f @clebertsuconic This should solve the memory leak issue with the minimum amount of changes.
https://docs.oracle.com/javase/7/docs/api/java/io/RandomAccessFile.html#getChannel() doc shows that RandomAccessFile and FileChannel are required to always be in sync so NFS or other SAN storages cannot be unsynced due to using both to manipulate the underline file system

franz1981:
Just for completeness that's what would change (ie no additional copies with heap buffers anymore):

- https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/native/java/io/RandomAccessFile.c#L79
- https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/native/java/io/io_util.c#L138
- https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/solaris/native/java/io/io_util_md.h#L71
- https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/solaris/native/java/io/io_util_md.c#L164

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/nio/NIOSequentialFile.java
```diff
@@ -173,6 +173,40 @@ public int read(final ByteBuffer bytes) throws Exception {
       return read(bytes, null);
    }
 
+   //On *nix platform is the size by which RandomAccessFile write/read are performed using a
+   //stack-allocated copy of the java byte[]
```
michaelandrepearce: And what about windows

franz1981: This was just a POC commit to mitigate the issue , but we can discuss on the comments about it :)
Thanks to looking at it

franz1981: I've verified that is a the same behaviour for Windows: it is in the shared JNI code by any OSs

franz1981:
I'm going to "tune" the optimization while reading/writing RandonAccessFile in chunks of 8192, because on https://github.com/netty/netty/pull/9591#issuecomment-533904873 I've measured that the allocator isn't a bottleneck when chunk size > 8192 and assuming single threaded usage (malloc -> free on the same thread). 
ATM AMQP and OpenWire could cause big buffers allocation on native space (on JNI), so I would reduce the impact of this by tuning chunk size just to reduce the foot print over head of it.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/nio/NIOSequentialFile.java
```diff
@@ -160,14 +173,63 @@ public int read(final ByteBuffer bytes) throws Exception {
...
          if (channel == null) {
             throw new ActiveMQIllegalStateException("File " + this.getFileName() + " has a null channel");
          }
-         int bytesRead = channel.read(bytes);
+         final int bytesRead;
+         if (bytes.hasArray()) {
+            if (bytes.remaining() > BUF_SIZE) {
+               bytesRead = readRaf(rfile, bytes.array(), bytes.arrayOffset() + bytes.position(), bytes.remaining(), BUF_SIZE);
+            } else {
+               bytesRead = rfile.read(bytes.array(), bytes.arrayOffset() + bytes.position(), bytes.remaining());
```
wy96f: @franz1981 just curious: why you added a else branch? `readRaf` seemingly can cover both conditions.

wy96f:
> The overhead of file reading using `RandomAccessFile` and `byte[]` vs `FileChannel` and direct `ByteBuffer` is variable, but using the benchmark I've used on Netty [here](https://github.com/netty/netty/pull/9591#issuecomment-534002621) has given this result:
> 
> ```
> Benchmark                                  (chunkSize)    (fileName)  (fileSize)  (ioSize)  Mode  Cnt     Score     Error  Units
> ChunkedFileBenchmark.readFullyFile               32768  /tmp/raf.tmp    10485760     32768  avgt    4  2153.527 ± 715.641  us/op
> ChunkedFileBenchmark.readFullyFileWithNio        32768  /tmp/raf.tmp    10485760     32768  avgt    4  1775.999 ± 831.715  us/op
> ```
> 
> That means about a 25% perf difference while using super-fast disks and I assume to be the same on write side.

Do you mean 25% perf reduction using `RandomAccessFile` and `byte[]` compared to `FileChannel` and direct `ByteBuffer` ?

franz1981:
@wy96f yes, more then a good reason to make most of the code to avoid using byte[] + rafs, but...it saves from mem leaks :)

artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/nio/NIOSequentialFile.java
franz1981: to save inlining budget: for most of the cases the first check will safe the call on the happy path

wy96f: Understand :)

PR #2846, opened 2019-09-23
ARTEMIS-2500 CoreMessage doesn't make a ful copy of its props

gaohoward:
When CoreMessage is doing copyHeadersAndProperties() it doesn't
make a full copy of its properties (a TypedProperties object).
It will cause problem when multiple threads/parties are modifying the
properties of the copied messages from the same message.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
@@ -461,7 +461,7 @@ public class CoreMessage extends RefCountMessage implements ICoreMessage {
       priority = msg.getPriority();
 
       if (msg instanceof CoreMessage) {
-         properties = ((CoreMessage) msg).getProperties();
+         properties = new TypedProperties(((CoreMessage) msg).getProperties());
       }
    }
 
--- a/tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/message/impl/MessageImplTest.java
+++ b/tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/message/impl/MessageImplTest.java
@@ -37,6 +37,8 @@ import org.apache.activemq.artemis.core.protocol.core.impl.wireformat.SessionSen
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
 import org.apache.activemq.artemis.tests.util.ActiveMQTestBase;
 import org.apache.activemq.artemis.utils.RandomUtil;
+import org.apache.activemq.artemis.utils.UUID;
+import org.apache.activemq.artemis.utils.UUIDGenerator;
 import org.apache.activemq.artemis.utils.Wait;
 import org.junit.Assert;
 import org.junit.Test;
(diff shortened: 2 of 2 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
```diff
@@ -461,7 +461,7 @@ public void copyHeadersAndProperties(final Message msg) {
       priority = msg.getPriority();
 
       if (msg instanceof CoreMessage) {
-         properties = ((CoreMessage) msg).getProperties();
+         properties = new TypedProperties(((CoreMessage) msg).getProperties());
```
franz1981: SO the problem is that 'msg' and `this` don't have to share the same `TypedProperties` instance?

michaelandrepearce: What's the use case here, the body is shared without deepclone. So if issues with properties surely same issue with body

clebertsuconic:
@howardgao I believe you did this because this was fixed an issue in HornetQ.

Artemis already have these protected. Which is different from HornetQ. I don't think you need this change here.

clebertsuconic:
the user that reported an issue on HornetQ, reported it on a very old version of HornetQ, which is not an issue in Artemis.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
gaohoward: @franz1981 @michaelandrepearce guys,
Currently this method is used to make a copy of a large message. (LargeServerMessageImpl.copy(long). And it is used in Queue to send a message into DLA (or Expiry i think). So it's not used alone. It is used as part of copy and the buffer is not shared.
This creates problem when a large message for a topic (i.e. multicast) and multiple references of the same message is being sent to DLA concurrently. During the "send to DLA" action it will clean up certain internal properties of the copied message (i.e. _AMQ_ROUTE_TO etc) during reroute to DLA. However if the DLA is clustered, some of the copied messages may be routed to remote DLA, in which case an internal property (_AMQ_ROUTE_TOsf....) will be added to the message. You can see one is adding and one is removing. If the sequence happens at the right order and right moment, the messages routed to remote DLA will get their internal property removed and cause problem when the bridge forwards these messages to remote queue. See ClusterConnectionBridge.beforeForward() method. It would throw "no queueIDs defined" exception. If that happens the bridge will be removed from the queue's consumer list as bad consumers and the store and forward queue will have messages accumulated and never get forwarded.

clebertsuconic:
As a matter of fact.. that method is only used to convert an incoming Message as a Large Message.


perhaps we should just name it convertSomething instead of copy. and keep the use for LargeMessage

clebertsuconic:
AFAIK message.copy(newID) will not use this method you changed

if you look at the javadoc that method is only used to treat Large Messages.

A regular message when becomes a large messages, there's a transfer made for the properties. It is supposed to be that way.


I agree the name of the method is misleading here. which can be changed.

clebertsuconic:
copy will use the Copy Constructor:


Which is already doing such conversion

gaohoward:
I find there are 2 copy methods in LargeServerMessageImpl. One is copy() which uses copy constructor. The other is copy(newId) which does not. If we can use copy constructor in both copy methods then this PR is not necessary. In the case the copyHeadersAndProperties() shall be removed because it can cause subtle errors described above and hard to find.

clebertsuconic:
That method is not used as part of copy. It is used as part of sending a regular message on the client as Large Message.

as a matter of fact I just sent a PR renaming the method.

gaohoward:
I'm not sure about that... the properties shouldn't be shared imo.

clebertsuconic:
@gaohoward They are never shared on copy Howard!!!

First time a message became Large Message, their properties are moved from the CoreMessage to its brand new LargeMessage. That instance was never sent, it's just a message becoming a large message for the first time. So their properties are then moved to the LargeMessage, and the previous CoreMessage is discarded by GC.

gaohoward:
ok, I hope you are right. :) This issue has been reproduced in HQ but In HornetQ there is no CoreMessage so logic may be different.

clebertsuconic:
@gaohoward this was fixed long ago in Artemis Howard. That's what I'm trying to tell you.

There was also one specific issue fixed some time ago: 4fa2e75cdc5572e2e98d311fdcde0c2010726780

It's just that this user is persistent on not moving to any newer version (even a newer version of Hornetq) Where this fix is already applied.

clebertsuconic:
I will close this one, as it is superseded by #2849

gaohoward:
ah, ok thanks @clebertsuconic

clebertsuconic:
@gaohoward the only case where I could now see as not isolating copy is LargeMessage.copy()

on that case the new property is being shared. 


i will reopen your PR here

PR #2847, opened 2019-09-23
ARTEMIS-2494: [AMQP] Allow Modified disposition to be used signal address full to a sending peer

k-wall:


artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -264,6 +264,8 @@
 
    private static final String AMQP_USE_CORE_SUBSCRIPTION_NAMING = "amqp-use-core-subscription-naming";
 
+   private static final String AMQP_USE_MODIFIED_FOR_TRANSIENT_DELIVERY_ERRORS = "amqp-use-modified-for-transient-delivery-errors";
```
clebertsuconic: Same thing here, please use ProtonProtocolManager property for this. you can simply set a boolean property on the acceptors.

k-wall: @clebertsuconic you mean like amqpLowCredits?

`         <acceptor name="amqp">tcp://0.0.0.0:5672?tcpSendBufferSize=1048576;tcpReceiveBufferSize=1048576;protocols=AMQP;useEpoll=true;amqpCredits=1000;amqpLowCredits=300</acceptor>`

if so would.. be acceptable?
`amqpUseModifiedForTransientDeliveryErrors=<boolean>`

clebertsuconic: @k-wall yes.. but can't you find a shorter property name? :)

k-wall: @clebertsuconic configuration mechanism changed as suggested.  

on the name, I can't think how to shorten significantly without losing important information.  Best alternative I can up with was `amqpSendModifiedForTransientErrors` which save a few characters.  FWIW I don't think too many users will be configuring this option.  I'm happy to rename as you see fit.  Let me know.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
```diff
@@ -326,6 +325,37 @@ private void actualDelivery(Delivery delivery, Receiver receiver, ReadableBuffer
...
+            }
+            condition.setDescription(e.getMessage());
+
+            Rejected rejected = new Rejected();
+            rejected.setError(condition);
+            return rejected;
+         } else if (source.getDefaultOutcome() instanceof DeliveryState) {
+            return ((DeliveryState) source.getDefaultOutcome());
+         } else {
+            return Accepted.getInstance();
```
gemmellr: I think introducing this last bit as a new default behaviour will turn out to be very painful, and so think it should be gated to opt-in instead. It also feels like it somewhat goes against the fact this change introduces a toggle to mostly avoid changing the existing default questionable behaviour.

I've come to realise since seeing this PR that various other clients/servers are not setting the outcomes at all, though both using and expecting the full range of outcomes, and therefore it seems this change will have the effect of changing the default behaviour from the broker rejecting, to the broker accepting the message and probably dropping it on the floor.

I realise this is actually implementing the [strange] behaviour the AMQP 1.0 spec outlines for the outcomes / default-outcome field handling, but after all this time I think it would be a bad idea to introduce it by default knowing many existing peers will fall foul of it.

gemmellr: (and obviously those peers should also be fixed)

k-wall: @gemmellr good spot, I hadn't realised that.  I agree that we should make sure that issues are raised against the relevant projects to we move to a spec compliant population.  Which are they?

For this patch, for now, I need to change the else clause of line 354 to return a Rejected to maintain the existing behaviour for peers that specify neither outcomes or defaultOutcome.

gemmellr: At least some of the Proton bindings, perhaps AMQP.net lite (though its lower level and probably expected for you to configure it), probably Dispatch...and I'm sure there will be others. I'd say I'm currently more aware of things not setting these bits, or likely not to be, than I am those that actually are.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
```diff
@@ -326,6 +325,37 @@ private void actualDelivery(Delivery delivery, Receiver receiver, ReadableBuffer
       }
    }
 
+   private DeliveryState determineDeliveryState(final Source source, final boolean useModified, final Exception e) {
+      Outcome defaultOutcome = getEffectiveDefaultOutcome(source);
+      if (e instanceof ActiveMQAddressFullException && useModified &&
+          (outcomeSupported(source, Modified.DESCRIPTOR_SYMBOL) || defaultOutcome instanceof Modified)) {
+         Modified modified = new Modified();
+         modified.setDeliveryFailed(true);
+         modified.setUndeliverableHere(false);
```
gemmellr: I was too focused on the Accept and looking into its implications earlier that I missed this. I don't think undeliverable-here should be set, or else it should be configurable. There is nothing wrong with the message itself here, and the point of the related JIRA was you want the sender to be able to send the message again, but setting undeliverable-here specifically instructs them otherwise.

k-wall: `undeliverable-here` is set false.  only `delivery-failed` is set true.  I maybe missing your point.

gemmellr: You aren't missing anything, I've somehow completely misread that the second time round, thinking it was being set true...no idea how, read start of one line and end of other? :)

I probably wouldn't bother setting it, but don't see harm in doing so.

k-wall: That should be fixed.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
k-wall: I've changed the code to send Rejected in the case where neither outcomes not default-outcome is specified.

k-wall:
My initial patch missed the code path where the address full policy is BLOCK and the paging store signals itself full.    Previously reject would be sent unconditionally.  Now Modified is sent if enabled/supported on the link as per the other case.

There is a small behavioural change in this patch (a25a289).  Previously on the BLOCK code path the Reject for an address-full would be sent with a resource-limit-exceeded whereas the other code path would send a Reject for an address-full with a condition ""failed".  This patch uses resource-limit-exceeded consistently.

clebertsuconic:
Can you squash commits? I will then run tests and merge it.

k-wall:
@clebertsuconic squash done as requested, as rebased against master

PR #2848, opened 2019-09-23
ARTEMIS-2497: [AMQP] Allow handling of the reject disposition to be configured

k-wall:


artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -306,6 +306,8 @@
 
    private Long globalMaxSize;
 
+   private boolean amqpTreatRejectAsUnmodifiedDeliveryFailed = ActiveMQDefaultConfiguration.getDefaultTreatRejectAsUnmodifiedDeliveryFailed();
```
clebertsuconic: I would use the ProtocolManager configuration, being a property of the acceptor, instead of a global property for such thing.

That way you can configure this as part of the acceptor properties.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -54,6 +54,15 @@
             </xsd:annotation>
          </xsd:element>
 
+         <xsd:element name="amqp-treat-reject-as-unmodified-delivery-failed" type="xsd:boolean" maxOccurs="1" minOccurs="0" default="false">
```
clebertsuconic: If you use the protocol manager (AMQP Protocol Manager, such property would just be part of BeanUtils, and you wouldn't need to change any schema.

clebertsuconic:
Can you move this into the ProtonProtocolManager as a property, the acceptors can then configure this with a simple property.

clebertsuconic:
@k-wall I'm trying to minimize the number of options someone can configure on ConfigurationImpl. for protocol specific semantics, please use the ProtocolManager properties.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
clebertsuconic: Are you changing this similar to your other featrure? add it to ProtonProtocolManager instead?

k-wall:
@clebertsuconic review comments addressed and commits squashed.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
k-wall: Done

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
k-wall: Done

jbertram:
@k-wall can you rebase this? Conflicts have developed. Thanks!

PR #2850, opened 2019-09-24
ARTEMIS-2504 implement retroactive addresses

jbertram:
A new feature to preserve messages sent to an address for queues that will be
created on the address in the future. This is essentially equivalent to the
"retroactive consumer" feature from 5.x. However, it's implemented in a way
that fits with the address model of Artemis.


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1971,7 +1988,9 @@ private synchronized int iterQueue(final int flushLimit,
 
             if (filter1 == null || filter1.match(ref.getMessage())) {
                messageAction.actMessage(tx, ref);
-               iter.remove();
+               if (remove) {
```
brusdev: Do we need the same check for scheduledDelivery messages (https://github.com/apache/activemq-artemis/blob/master/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerImpl.java#L128) and paged messages (https://github.com/apache/activemq-artemis/blob/master/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java#L2002)?

michaelandrepearce:
A few bits as havent fully explored the feature just yet.

How does someone control if the retroactive address will behave when full, e.g. page, block, drop?

How or where can someone configure the naming of a retroactive address so they can control their names to avoid conflict or for security settings.

For a specific address it would be good to be able to configure the retroactive queue fully, eg at the address level define a queue (like we can others) and then specify the queue to use. E.g. for cases where retroactive you may want LVQ behaviour.

For queues created for retroactive, if there is a very high number of message groups and someone is using buckets so that the group map doesnt explode in size, how is this being avoided in the retroactive queue?

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -467,6 +478,83 @@ private boolean internalAddressInfo(AddressInfo addressInfo, boolean reload) thr
...
+                             .setAddress(retroactiveAddressName.toString())
+                             .setExclusive(false)
+                             .setForwardingAddress(internalAddressName.toString())
+                             .setRoutingType(ComponentConfigurationRoutingType.STRIP));
+   }
+
+   private void removeRetroactiveResources(SimpleString address) throws Exception {
+      String prefix = server.getConfiguration().getInternalNamingPrefix();
+
+      SimpleString internalDivertName = ResourceNames.getRetroactiveResourceName(prefix, address, ResourceNames.DIVERT);
```
michaelandrepearce: Why divert and separate address is needed. Why not have the retro queue under the original address?
With having separate address will mean message has to be copied rather than it simply being a message reference to existing message.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/RetroactiveAddressTest.java
```diff
@@ -0,0 +1,287 @@
...
+      Wait.assertTrue(() -> server.locateQueue(divertQueue).getRingSize() == COUNT, 1000, 100);
+      server.stop();
+      server.start();
+      server.getAddressSettingsRepository().addMatch(addressName.toString(), new AddressSettings().setRetroactiveMessageCount(COUNT * 2));
+      Wait.assertTrue(() -> server.locateQueue(divertQueue).getRingSize() == COUNT * 2, 1000, 100);
+      server.getAddressSettingsRepository().addMatch(addressName.toString(), new AddressSettings().setRetroactiveMessageCount(COUNT));
+      Wait.assertTrue(() -> server.locateQueue(divertQueue).getRingSize() == COUNT, 1000, 100);
+   }
+
+   @Test
```
michaelandrepearce: Should add a test that an anycast producer produces messages and ensures they dont get retroactively consumed by on multicast

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -467,6 +478,83 @@ private boolean internalAddressInfo(AddressInfo addressInfo, boolean reload) thr
...
+                            0L,
+                            false,
+                            retroactiveMessageCount);
+      }
+      server.deployDivert(new DivertConfiguration()
+                             .setName(internalDivertName.toString())
+                             .setAddress(retroactiveAddressName.toString())
+                             .setExclusive(false)
+                             .setForwardingAddress(internalAddressName.toString())
+                             .setRoutingType(ComponentConfigurationRoutingType.STRIP));
```
michaelandrepearce: I dont think we should be stripping. We need to ensure anycast messages only end up in anycast queues and like wise multicast.

michaelandrepearce: It may mean in a retro anycast and a retro multicast queue is required to preserve that

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -1170,6 +1172,10 @@ public SecuritySettingPlugin run() {
             addressSettings.setDefaultConsumerWindowSize(XMLUtil.parseInt(child));
          } else if (DEFAULT_RING_SIZE.equalsIgnoreCase(name)) {
             addressSettings.setDefaultRingSize(XMLUtil.parseLong(child));
+         } else if (RETROACTIVE_MESSAGE_COUNT.equalsIgnoreCase(name)) {
+            long retroactiveMessageCount = XMLUtil.parseLong(child);
+            Validators.GE_ZERO.validate(DEFAULT_ADDRESS_ROUTING_TYPE, retroactiveMessageCount);
```
michaelandrepearce: Default address routing type in the validator for msg count?

michaelandrepearce:
On the retroactive queue, would want a flag to tell us its a retro queue, thats then exposed to jmx so that metrics systems can get the tag, and so alerting rules on queue depths can ignore these queues automatically based on that flag.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -467,6 +478,83 @@ private boolean internalAddressInfo(AddressInfo addressInfo, boolean reload) thr
...
+      String prefix = server.getConfiguration().getInternalNamingPrefix();
+      final SimpleString internalAddressName = ResourceNames.getRetroactiveResourceName(prefix, retroactiveAddressName, ResourceNames.ADDRESS);
+      final SimpleString internalQueueName = ResourceNames.getRetroactiveResourceName(prefix, retroactiveAddressName, ResourceNames.QUEUE);
+      final SimpleString internalDivertName = ResourceNames.getRetroactiveResourceName(prefix, retroactiveAddressName, ResourceNames.DIVERT);
+
+      if (!reload) {
+         AddressInfo addressInfo = new AddressInfo(internalAddressName).addRoutingType(RoutingType.ANYCAST).setInternal(true);
+         addAddressInfo(addressInfo);
+         server.createQueue(internalAddressName,
+                            RoutingType.ANYCAST,
```
michaelandrepearce: Should message groups be disabled? So theyre not tracked as not needed in the retro queue and would avoid issues if people have very large numbers of groups.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -3404,6 +3404,14 @@
                   </xsd:documentation>
                </xsd:annotation>
             </xsd:element>
+
+            <xsd:element name="retroactive-message-count" type="xsd:long" default="0" maxOccurs="1" minOccurs="0">
```
michaelandrepearce: Can prefix default. In future or even in this pr it would be good to be able to explicitly set/control this at an address level. Thus this would just be a default thats then applied if none is set at that level

docs/user-manual/en/retroactive-addresses.md
```diff
@@ -0,0 +1,73 @@
...
+
+```xml
+<address-settings>
+   <address-setting match="orders">
+      <retroactive-message-count>100</retroactive-message-count>
+   </address-setting>
+</address-settings>
+```
+
+The value for `retroactive-message-count` can be updated at runtime either via
```
michaelandrepearce: This suggests it updates dynamically. It isnt entirely true as will only update the default for new addresses, existing addresses where the retro queue is created already it doesnt update. This really is acting as a default value, not the explicit. If you needed to change at runtime and existing you would need to find the internal queue and update it.

michaelandrepearce:
This feature would be more powerful if support for a timebased or bytes size retention, e.g. retro active for last hour etc. 

I realise that this may not be possible on this initial feature pr but this should impact the way we think about how we design the configuration so its most flexible for future expansion once the abilty for a queue to be ringqueue by byte size or time.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
michaelandrepearce: The more i think on this the more strongly i think we shouldnt have the divert and separate address, simply should be queue under the existing address.

jbertram:
> How does someone control if the retroactive address will behave when full, e.g. page, block, drop?

For the most part you can control the configuration of the internal retroactive address/queues using address settings just like any other address. The resources are named according to a specific pattern which can be used in the `match` for the address-setting.

> How or where can someone configure the naming of a retroactive address so they can control their names to avoid conflict or for security settings.

The naming pattern is outlined in the documentation. Conflicts can be avoided through the `internal-naming-prefix` setting.

> For queues created for retroactive, if there is a very high number of message groups and someone is using buckets so that the group map doesnt explode in size, how is this being avoided in the retroactive queue?

Configure the group buckets via an address-setting. I changed this in my latest update.

> Is routing type being preserved.

Yes, it is being preserved now. I changed this in my latest update.

> Why divert and separate address is needed. Why not have the retro queue under the original address?

My original plan was to have the ring queues for the retroactive address under the main address itself, but I didn't want those queues to be subject to the address-settings of that address. Separating the ring queues onto their own address means better configuration flexibility.

> Default address routing type in the validator for msg count?

Fixed in the my latest update.

> On the retroactive queue, would want a flag to tell us its a retro queue, thats then exposed to jmx so that metrics systems can get the tag, and so alerting rules on queue depths can ignore these queues automatically based on that flag.

Added on my latest update.

> In future or even in this pr it would be good to be able to explicitly set/control [the retroactive-message-count] at an address level. Thus this would just be a default thats then applied if none is set at that level

As far as I can tell that is already how `retroactive-message-count` is functioning. If not defined then the default of `0` is used. Addresses themselves have no direct configuration attributes other than `name`. The `address-settings` block is designed to configure the settings for an address. I see `retroactive-message-count` the same as settings like `max-size-bytes` and a bunch of others.

> This suggests it updates dynamically. It isnt entirely true as will only update the default for new addresses, existing addresses where the retro queue is created already it doesnt update.

That's not true. There is a listener which will update the `ring-size` on the ring queues dynamically.

> This feature would be more powerful if support for a timebased or bytes size retention, e.g. retro active for last hour etc.

Fair enough, but that's not the subject of this PR.

michaelpearce-gain:
@jbertram 

re: "My original plan was to have the ring queues for the retroactive address under the main address itself, but I didn't want those queues to be subject to the address-settings of that address. Separating the ring queues onto their own address means better configuration flexibility."

The issue here are 
1) you have to message copy, not just make new references - expensive if you have a large retro active to get through, also will cause paging faults to occur as copies occur, even if consumer isnt at that message.
2) its going to be very messy as an end user to make address settings for the internal's, especially as typically (as an end user, i probably just want the same settings for bit like paging etc)
3) address settings do not update the queue (in general they are defaults, overrides and actuals are always set at queue level), as such once queue is created its not going to be possible to adjust



Also having it at the address allows us to, and the way i was seing this working so then you can set this being better configurable and working nicely is to have these defined as sepecial queues under the address (eg. in broker.xml)

        <address name="MyAddress">
                <multicast>
                        <queue name="MyQueue" group-buckets="64" group-rebalance="true">
                                <filter string="ServiceId = 1" />
                                <durable>true</durable>
                        </queue>
                        <retro-active-queue group-buckets="0" >
                                <message-count>1024</message-count>
                        </queue>
                </multicast>
        </address>

This way then it can expose any queue settings, easily.

jbertram:
As noted, separating the ring queues onto their own address means better configuration flexibility. This is mainly for `max-size-bytes` and `address-full-policy` because of the way these specific settings can impact ring-queues (see the ring queue docs on paging for more details on that). In short, if your ring-queue for retroactive functionality is on the main address and that address starts paging then the retroactive functionality "breaks" in that newly created queues won't get any of the messages sent after paging started.

Address-settings can and do update queues. The `retroactive-message-count` that I implemented is one example of this. You can see other examples of this in `org.apache.activemq.artemis.core.server.impl.QueueImpl.AddressSettingsRepositoryListener`. If there are address-settings which currently don't update queues which should we can implement those a separate PR.

I'm not clear on what specifically will be "messy" about configuring address-settings for the separate address. For an address named "myRetroactiveAddress" with a `retroactive-message-count` > 0 users should be able to use a match like `*.myRetroactiveAddress.#.retro` to configure what they need.

michaelandrepearce:
It def shouldnt break if paging occurs. One of the bits myself and clebert have discussed once a time ago. Was the idea that you could have an address paging by default to achive this feature. And then could even have retro active for several days. Should you need replay etc.

jbertram:
> It def shouldnt break if paging occurs not entirely sure what you meant here.

The explanation is in my previous comment, "...newly created queues won't get any of the messages sent after paging started." I assumed you were familiar with the caveats for ring queues and paging given they are outlined in the documentation and you reviewed that PR.

> One of the bits myself and clebert have discussed once a time ago.

Clebert hasn't been involved in this implementation.

> For address setting match like that it will break if someone uses different wildcard configuration...

I addressed this issue in my last push. The names of the resources will reflect the delimiter configuration.

michaelandrepearce:
@jbertram tbh im struggling to understand still the issue, if an address is paging to disk, ring queue still gets new messages delivered.

on address-settings front thanks, that seems to make things work for custom wildcards.

When i have a large retro-active queue, and a new consumer joins, and that consumer may take 5 minutes to process those messages, im seeing some nasty GC spikes at the initial few seconds, from debugging that seems to be when the messages are copied back to route to original address and the new queue, as the broker goes as fast as it can regardless of the consumer speed. This kind of behaviour could destabilise the broker.

Ideas to solve or limit this issue:

1) is there a way to control the copy over so its not done as one big bang (e.g. as fast as broker can possible do), but feed the new queue at the rate the consumer is consuming there for avoiding the very big spike in object creation.

2) Is it possible to make it so that we dont have to copy, but it simply is a message ref? 

3) If 2 is not possible because a separate address, could we have a toggle that will deploy under the original address and when deployed like such a copy isnt needed and simply message ref is added to the queue, so then its a choice for end users. this then gives a kind of flexibility depending on end user case, with trade off for either, but at least then its up for the end user.

jbertram:
> if an address is paging to disk, ring queue still gets new messages delivered, theyre just paged references.

According to my tests the ring queue will *not* actually get those messages. From what I can tell messages are paged at the address level, and they aren't added to the actual queue until they are de-paged.

Regarding the performance issues, my initial recommendation would be to limit the size of the ring queue to avoid such problems. As for your ideas...

> is there a way to control the copy over so its not done as one big bang (e.g. as fast as broker can possible do), but feed the new queue at the rate the consumer is consuming there for avoiding the very big spike in object creation.

I understand your point about throttling the copy speed, but I'm not sure that will provide the semantics a consumer expects. Once the queue is created then it will receive any "new" message sent to the queue's address (at least in the multicast use-case). If the "old," retroactive messages are not all immediately added to the queue then the consumer might receive "new" messages before it receives all the "old" ones. 

Also, since messages are added to the queue when it is first created and before the client even consumes anything how can we calculate the consumption rate in order to throttle the copy rate?

> Is it possible to make it so that we dont have to copy, but it simply is a message ref?

I'll explore this possibility.

jbertram:
@michaelandrepearce, try it now. I eliminated the copy when populating the new queue with the old messages.

michaelandrepearce:
@jbertram im busy this week with some priority bits at work, thus also not had time tonrespin the amqp nms release. Im hoping to get time next week to look at this again

michaelandrepearce:
So im getting odd behaviour after restarts. Trying to work out what is going on. Have you done much destructive testing on this?

jbertram:
I'm not sure what you mean by "destructive testing." There are a number of tests on the PR one of which specifically tests restarting the broker, although there aren't any testing failover.

michaelandrepearce:
I mean running this a little more than just an integration test. E.g. actually deploying to a real environment and do bits like failover, outage, client disconnects

jbertram:
I added a simple failover test and push. Everything looks OK as far as I can tell.

As far as the "destructive testing" you mentioned goes, I've done a bit more than what is captured in the tests, but nothing I could categorize as industrial grade (e.g. no huge loads on big hardware).

michaelandrepearce:
No further comments from me atm other bits and bobs can always be added later in future prs /version and refined over time. Thanks for this @jbertram was prob the last big ticket for artemis to be in feature parity with activemq5.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -759,6 +761,15 @@ public AddressSettings setDefaultRingSize(final long defaultRingSize) {
...
 
+   public long getRetroactiveMessageCount() {
+      return retroactiveMessageCount != null ? retroactiveMessageCount : ActiveMQDefaultConfiguration.DEFAULT_RETROACTIVE_MESSAGE_COUNT;
+   }
+
+   public AddressSettings setRetroactiveMessageCount(final long defaultRetroactiveMessageCount) {
+      this.retroactiveMessageCount = defaultRetroactiveMessageCount;
+      return this;
+   }
+
```
clebertsuconic: I have tried to add a comment on a line you didn't touch on github and I couldn't  figure out how to do it...


It seems you missed implementing the merge method here,

you also missed implementing the persistence of the new attribute here.


So, I think you should add implementation on this new attribute, and you should probably also add tests on loading the new setting.

jbertram:
@clebertsuconic, I dealt with the address-setting persistence issue and found some other related issues along the way which I've also fixed in another commit.

PR #2851, opened 2019-09-24
ARTEMIS-2503 Improve wildcards for the roles access match

brusdev:
Improve wildcard support for the key attribute in the roles access
match element, allowing prefix match for the mBean properties ie
`<match domain="org.apache.activemq.artemis" key="queue=my*">`.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/JMXAccessControlList.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/JMXAccessControlList.java
@@ -18,79 +18,95 @@ package org.apache.activemq.artemis.core.server.management;
 
 import javax.management.ObjectName;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Pattern;
 
 public class JMXAccessControlList {
+   private static final String WILDCARD = "*";
 
-   private Access defaultAccess = new Access("*");
-   private Map<String, Access> domainAccess = new HashMap<>();
-   private ConcurrentHashMap<String, List<String>> whitelist = new ConcurrentHashMap<>();
-
-
-   public void addToWhiteList(String domain, String match) {
-      List<String> list = new ArrayList<>();
-      list = whitelist.putIfAbsent(domain, list);
-      if (list == null) {
-         list = whitelist.get(domain);
+   private Access defaultAccess = new Access(WILDCARD);
+   private ConcurrentHashMap<String, TreeMap<String, Access>> domainAccess = new ConcurrentHashMap<>();
+   private ConcurrentHashMap<String, TreeMap<String, Access>> whitelist = new ConcurrentHashMap<>();
+   private Comparator<String> keyComparator = (key1, key2) -> {
+      boolean key1ContainsWildCard = key1.contains(WILDCARD);
+      boolean key2ContainsWildcard = key2.contains(WILDCARD);
+      if (key1ContainsWildCard && !key2ContainsWildcard) {
+         return +1;
+      } else if (!key1ContainsWildCard && key2ContainsWildcard) {
+         return -1;
+      } else if (key1ContainsWildCard && key2ContainsWildcard) {
+         return key2.length() - key1.length();
       }
-      list.add(match != null ? match : "*");
+      return key1.length() - key2.length();
+   };
+
+   public void addToWhiteList(String domain, String key) {
+      TreeMap<String, Access> domainMap = new TreeMap<>(keyComparator);
+      domainMap = whitelist.putIfAbsent(domain, domainMap);
+      if (domainMap == null) {
+         domainMap = whitelist.get(domain);
+      }
+      Access access = new Access(domain, normalizeKey(key));
+      domainMap.putIfAbsent(access.getKey(), access);
    }
 
 
    public List<String> getRolesForObject(ObjectName objectName, String methodName) {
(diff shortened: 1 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/JMXAccessControlList.java
```diff
@@ -25,12 +25,22 @@
...
+
 public class JMXAccessControlList {
 
    private Access defaultAccess = new Access("*");
-   private Map<String, Access> domainAccess = new HashMap<>();
+   private HierarchicalRepository<Access> domainAccess;
    private ConcurrentHashMap<String, List<String>> whitelist = new ConcurrentHashMap<>();
 
+   public JMXAccessControlList() {
+      WildcardConfiguration domainAccessWildcardConfiguration = new WildcardConfiguration();
```
michaelandrepearce: Wildcard configuration should use the same as broker.xml

brusdev: The previous implementation of the wildcard for the key attribute (documented here https://github.com/apache/activemq-artemis/blob/master/docs/user-manual/en/management.md) isn't compatible with broker custom wildcard config defined in broker.xml. If wildcard configuration uses the same as broker.xml it could break the compatibility with the previous implementation.

michaelandrepearce: Point here is you could put a flag so that it uses old non flexible or use the broker.xml one. Lets not further diverge.

brusdev: I like your solution where would you put the flag, broker.xml or management.xml?

michaelandrepearce: As the deviation was in management.xml probably there

brusdev: I pushed a commit to use the same flexible wildcard support used for the addresses but I added a new node to customize the wildcard syntax for the authorisation key attributes because:
- the addresses and the keys wildcard support could have different requirements
- to avoid to introduce a dependence between management.xml and broker.xml

brusdev: @michaelandrepearce WDYT about the last changes?

clebertsuconic: @michaelandrepearce my personal opinion here: can't we keep this simple.. meaning. no configuration of wildcard for management?

I see this can be useful on queues.. etc.. but on management, if we could keep it simple??

I mean... management and broker should be kept independent.. I wouldn't load something from broker.xml to change semantics on management... 

And if you require the wildcard configuration, then you need testing..and more moving parts bound to fail and document...

I would prefer to keep it simpler?

clebertsuconic: I'm concerned about users abusing the configuration on management, things not working later.. and having to answer to user forum's / customer issues... etc...

michaelandrepearce: So my concern is trying to automate and unify the config the fact already slightly snowflaked gives me a massive headache as is. Really don't need further differing config. If we want to enchance the wildcards in management i feel strongly we need to align it

jbertram: These use-cases seem pretty different to me. The wildcard configuration in broker.xml is for address matching which is hierarchical in nature and therefore requires separating "words", matching individual and groups of words, etc. The "match" performed for keys defined in management.xml is basically just a **prefix**. It's not hierarchical and doesn't even use regular expressions. It's *very* simple and IMO it should stay that way. 

> If i need to secure a wildcard of address or queues on broker then no doubt i need to secure the same pattern from admin pov to.

I'm not sure is "no doubt" about that. Use cases for the security of actual messaging clients and management users can vary quite a bit. In any case, the same kind of matching isn't possible in management.xml as it is in broker.xml either before or after this PR. The use-case for this PR is just allowing a simple prefixing for the match "key" just like is available for the access "method" field. The syntax is the same. It's just allowing it in a new field. I don't see any issue with that.

clebertsuconic: @michaelandrepearce what do you suggest? merge management and broker.xml?

That's certainly desirable. but I don't think we can do it before a 3.0.

management is on a different domain from queues and addresses. it's about object  names.. not queue names.. hence something simpler would certainly apply here.

jbertram: It looks like there's a bit more going on here than just prefixing support. Looks like you implemented real regular expression matching and there's a bit in there with the TreeMaps that I don't quite understand. Certain matches are prioritized over others? Can you clarify?

brusdev: @jbertram with the last commit I tried to prioritize exact matches over wildcard matches and longer wildcard matches on shorter wildcard matches. I added the test testBasicRoleWithWildcardInKey that shows what I mean.

jbertram: @brusdev, that makes sense. I think something needs to be added to the docs explaining this otherwise the behavior might catch users by surprise.

brusdev: @jbertram I missed your last comment. I created the new PR #2952 to improve the doc.

PR #2852, opened 2019-09-25
ARTEMIS-2505: Fix wiring of the max-size-bytes-reject-threshold address-setting

k-wall:


k-wall:
@franz1981 unit test added as requested.

michaelandrepearce:
Can you squash the commits

k-wall:
@michaelandrepearce commits squashed, thanks for looking at this.

PR #2853, opened 2019-09-25
ARTEMIS-2506 MQTT doesn't cleanup underlying connection for bad clients

gaohoward:
When a bad MQTT clients drop its connection without proper closing
it the broker doesn't close the underlying physical connection.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/MQTTConnnectionCleanupTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/MQTTConnnectionCleanupTest.java
@@ -1,0 +1,85 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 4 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/MQTTConnnectionCleanupTest.java
```diff
@@ -0,0 +1,84 @@
...
+      assertEquals(1, acceptor.getConnections().size());
+
+      //now simulate a bad client by manually fail the server connection
+      RemotingConnection conn = server.getRemotingService().getConnections().iterator().next();
+
+      assertTrue(conn instanceof MQTTConnection);
+
+      conn.fail(new ActiveMQException("testBadClient"));
+
+      assertEquals("Server connection not cleaned up!", 0, acceptor.getConnections().size());
```
clebertsuconic: Can you use Wait.assertEquals here please? this is most likely an asynchronous operation that may eventual fail unless you use wait here.

gaohoward: ok will do

gaohoward: done.

PR #2855, opened 2019-09-30
ARTEMIS-2508 Crititical analyser trigger shutdown if removeAllMessages

brusdev:
The crititical analyser trigger the broker shutdown if try to
removeAllMessages with a huge queue. The iterQueue is split so as
not to keep the lock too time.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -1915,7 +1915,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    }
 
    @Override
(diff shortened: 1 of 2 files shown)


clebertsuconic:
@wy96f what about not use the page Executor for the delete?

The page executor was meant for cleanup and paging, not for deleting references on the queue.

clebertsuconic:
I don't understand why we are using an executor now for removing? why not just remove the locks and keep the change simple?

clebertsuconic:
hold a future waiting for the result has the same result as using the main thread anyway, right?

brusdev:
I used the page Executor because the iteration on paged queue seemed like quite similar to depage. Now I see your points of views so I'll just remove the locks and I'll keep it in the caller thread.

wy96f:
> @wy96f what about not use the page Executor for the delete?
> 
> The page executor was meant for cleanup and paging, not for deleting references on the queue.

Yes, that's what i thought :)

PR #2856, opened 2019-10-01
ARTEMIS-2509 Add some basic support for legacy openwire

michaelandrepearce:
This is needed to aid some lift and shift migration from activemq5 where non-java clients have support for older openwire protocols.

### DIFF
--- a/artemis-distribution/pom.xml
+++ b/artemis-distribution/pom.xml
@@ -164,6 +164,10 @@
            <artifactId>activemq-client</artifactId>
        </dependency>
        <dependency>
+           <groupId>org.apache.activemq</groupId>
+           <artifactId>activemq-openwire-legacy</artifactId>
+       </dependency>
+       <dependency>
           <groupId>org.eclipse.jetty.aggregate</groupId>
           <artifactId>jetty-all</artifactId>
           <type>jar</type>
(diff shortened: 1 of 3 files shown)


clebertsuconic:
This is just adding a jar on the /lib, how that helps? wouldn't that be simply a client library? 

just trying to understand.

michaelandrepearce:
Hi Clebert basically openwire tries to load dynamically the marshallers. Just by adding the legacy jar (as like in activemq5) now it can find the older version marshallers and handle.

clebertsuconic:
OIC.. thanks!

gemmellr:
Too late now its merged, but its not clear why the commit message has a cherry pick -x reference when it is the originating commit in this repo.

michaelandrepearce:
@gemmellr thats because to make PR it was cherry picked out onto the branch for pr from another internal one we have. An didnt spot the cherry added this to the comment normally i remove. I guess whom ever merge pr also didnt spot happy if someone wants to edit commit in master to remove

clebertsuconic:
it's not possible to push -f on master.. it's pointless to revert it / and send it again.


I didn't realize the -X in time.. if anyone is really strong opinionated about it we can revert it and send it again.

gemmellr:
No need to mess other folks up editing master, just pointing out it. Clebert suggested in a mail thread recently that theyve stopped that ability anyway anyway, which would make sense given the gitbox/github dual repo support would be screwed up by it.

gemmellr:
Gah, ninja'd by seconds :)

PR #2859, opened 2019-10-09
ARTEMIS-2513 Large message's copy may be interfered by other threads

gaohoward:
In LargeMessageImpl.copy(long) it need to open the underlying
file in order to read and copy bytes into the new copied message.
However there is a chance that another thread can come in and close
the file in the middle, making the copy failed
with "channel is null" error.

This is happening in cases where a large message is sent to a jms
topic (multicast address). During delivery it to multiple
subscribers, some consumer is doing delivery and closed the
underlying file after. Some other consumer is rolling back
the messages and eventually move it to DLQ (which will call
the above copy method). So there is a chance this bug being hit on.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
@@ -368,52 +368,51 @@ public final class LargeServerMessageImpl extends CoreMessage implements LargeSe
       try {
          LargeServerMessage newMessage = storageManager.createLargeMessage(newID, this);
 
-         boolean originallyOpen = file != null && file.isOpen();
+         //clone a SequentialFile to avoid concurrent access
+         ensureFileExists(false);
+         SequentialFile cloneFile = file.cloneFile();
 
-         validateFile();
+         try {
+            byte[] bufferBytes = new byte[100 * 1024];
 
-         byte[] bufferBytes = new byte[100 * 1024];
+            ByteBuffer buffer = ByteBuffer.wrap(bufferBytes);
 
-         ByteBuffer buffer = ByteBuffer.wrap(bufferBytes);
-
-         long oldPosition = file.position();
-
-         if (!file.isOpen()) {
-            file.open();
-         }
-         file.position(0);
-
-         for (;;) {
-            // The buffer is reused...
-            // We need to make sure we clear the limits and the buffer before reusing it
-            buffer.clear();
-            int bytesRead = file.read(buffer);
-
-            byte[] bufferToWrite;
-            if (bytesRead <= 0) {
-               break;
-            } else if (bytesRead == bufferBytes.length && !this.storageManager.isReplicated()) {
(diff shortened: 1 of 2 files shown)


gaohoward:
pls hold this, it's getting test failures in jenkins.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
```diff
@@ -368,52 +368,48 @@ public Message copy(final long newID) {
...
-         }
-
-         file.position(oldPosition);
-
-         if (!originallyOpen) {
-            file.close(false);
-            newMessage.getFile().close();
+         } finally {
+            cloneFile.close();
          }
```
wy96f: Do we need to close underlying file of new large message after copy?

clebertsuconic:
There are files leaking in your test.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
gaohoward: yes it should be closed as original code. That'll cause file leak.

gaohoward:
@clebertsuconic Yes I'm fixing it. Also same issue with hornetq. I'll fix that too.

gaohoward:
I'll run jenkins again.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
wy96f: @clebertsuconic @gaohoward I don't understand why we only close file when old file is not originally opened? In your case where other consumers open file and deliver, the file of new large message might not be closed. This will result in file leak and maybe data corrupt if broker crashes, wdyt?

gaohoward: tbh I don't know either. I just keep the old behavior.

gaohoward:
@clebertsuconic Jenkins has 5 failures, not likely relevant and those failed tests are passing on my local. So it's safe now.

clebertsuconic:
@gaohoward can I keep this open for 1 or 2 days?

I'm doing some work on large messages and I want to make sure why we have that code to close or keep files open.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
clebertsuconic: @wy96f I'm doing some work with large messages, and I want to double check that as well.

clebertsuconic: @wy96f  / @gaohoward actually.. I will merge this right away.. as whatever I do would clash here.


and I will make sure I address how we open and close files.

PR #2861, opened 2019-10-10
ARTEMIS-2515 pageIterator.hasNext spends too much time in the case of no messages matched

wy96f:
Discussion thread here http://activemq.2283324.n4.nabble.com/DISCUSS-Artemis-pageIterator-hasNext-spends-too-much-time-in-the-case-of-no-messages-matched-td4752397.html

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
@@ -47,6 +47,7 @@ import org.apache.activemq.artemis.core.paging.cursor.PagePosition;
 import org.apache.activemq.artemis.core.paging.cursor.PageSubscription;
 import org.apache.activemq.artemis.core.paging.cursor.PageSubscriptionCounter;
 import org.apache.activemq.artemis.core.paging.cursor.PagedReference;
+import org.apache.activemq.artemis.core.paging.cursor.PagedReferenceImpl;
 import org.apache.activemq.artemis.core.paging.impl.Page;
 import org.apache.activemq.artemis.core.persistence.StorageManager;
 import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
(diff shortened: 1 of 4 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -1323,7 +1328,13 @@ private PagedReference moveNext() {
             PagePositionAndFileOffset lastPosition = position;
             PagePositionAndFileOffset tmpPosition = position;
 
+            long timeout = System.currentTimeMillis() + DELIVERY_TIMEOUT;
```
franz1981: Use nanoTime here that is not susceptible to NTP updates :)
Now I can return to my honeymoon :P

wy96f: Resolved it, thanks :)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2680,24 +2684,24 @@ private boolean deliver() {
...
             // long
 
-            deliverAsync();
+            deliverAsync(true);
 
             return false;
          }
 
-         if (System.currentTimeMillis() > timeout) {
+         if (System.nanoTime() > timeout) {
```
clebertsuconic: I remember not long ago, we had to change a few nanoTime operations to avoid overflow.

I think, based on that the correct would be System.nanoTime() - timeout > 0

There was a previous big commit where we changed a few like this... the commit I could find where Franz changed something similar was 35c3475092576e7b008220f7c39b78e15850d5fd


Can you make that change here please?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2942,7 +2946,14 @@ private void depage(final boolean scheduleExpiry) {
...
       int depaged = 0;
-      while (timeout > System.currentTimeMillis() && needsDepage() && pageIterator.hasNext()) {
+      while (timeout > System.nanoTime() && needsDepage()) {
+         int status = pageIterator.tryNext();
+         if (status == 2) {
+            continue;
+         } else if (status == 0) {
+            break;
+         }
+
```
clebertsuconic: Same here

wy96f:
@clebertsuconic Changed it, thanks :)

PR #2864, opened 2019-10-11
ARTEMIS-2519: Use proper enum type inside ActiveMQUnexpectedRoutingTypeForAddress

cshannon:


artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQUnexpectedRoutingTypeForAddress.java
```diff
@@ -22,10 +22,10 @@
 public final class ActiveMQUnexpectedRoutingTypeForAddress extends ActiveMQException {
 
    public ActiveMQUnexpectedRoutingTypeForAddress() {
-      super(ActiveMQExceptionType.MAX_CONSUMER_LIMIT_EXCEEDED);
```
clebertsuconic: oops

PR #2865, opened 2019-10-16
ARTEMIS-2521 add documentation for role-mapping

sdirbach:
Add documentation for role-mapping implemented in ARTEMIS-1116


jbertram:
Thanks for the PR!

PR #2866, opened 2019-10-19
ARTEMIS-2523 Deprecate the parameter failoverOnInitialConnection

brusdev:
The parameter failoverOnInitialConnection wouldn't seem to be used and
makes no sense any more, because the connectors are retried in a loop.
So someone can just add the backup in the initial connection.


clebertsuconic:
LGTM however there's a real test failure shown on the PR check. We can merge this after fixed.

brusdev:
@clebertsuconic thanks, I pushed a commit to fix the test failure.

PR #2867, opened 2019-10-21
ARTEMIS-2524 Remove message from map in LVQ if it's deleted/moved/expired/changed

wy96f:
Message is acked but not removed from lvq map while calling QueueImpl::deleteReference/expireReference/moveReference.

Moreover calling QueueImpl::changeReferencePriority will make `HolderReference` refers to itself which finally leads to stack overflow.


clebertsuconic:
LGTM. I can't merge it right now.. will do it asap.

PR #2868, opened 2019-10-23
ARTEMIS-2466 PageSyncTimer::timeSync isn't configurable using ASYNCIO

brusdev:
Add the config parameter `page-sync-timeout` to set a customized value,
because if the broker is configured to use ASYNCIO journal, the timeout
has the same value of NIO default journal buffer timeout ie 3333333.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
@@ -109,6 +109,7 @@ public class Create extends InputAbstract {
 
    public static final String ETC_GLOBAL_MAX_SPECIFIED_TXT = "etc/global-max-specified.txt";
    public static final String ETC_GLOBAL_MAX_DEFAULT_TXT = "etc/global-max-default.txt";
+   public static final String ETC_PAGE_SYNC_SETTINGS = "etc/page-sync-settings.txt";
    public static final String ETC_JOLOKIA_ACCESS_XML = "jolokia-access.xml";
 
    @Arguments(description = "The instance directory to hold the broker's configuration and data.  Path must be writable.", required = true)
@@ -1002,6 +1003,7 @@ public class Create extends InputAbstract {
    private void performAutoTune(HashMap<String, String> filters, JournalType journalType, File dataFolder) {
       if (noAutoTune) {
          filters.put("${journal-buffer.settings}", "");
+         filters.put("${page-sync.settings}", "");
       } else {
          try {
             int writes = 250;
@@ -1018,6 +1020,7 @@ public class Create extends InputAbstract {
 
                filters.put("${journal-buffer.settings}", readTextFile(ETC_JOURNAL_BUFFER_SETTINGS, syncFilter));
 
+               filters.put("${page-sync.settings}", readTextFile(ETC_PAGE_SYNC_SETTINGS, syncFilter));
             } else {
                long time = SyncCalculation.syncTest(dataFolder, 4096, writes, 5, verbose, !noJournalSync, false, "journal-test.tmp", ActiveMQDefaultConfiguration.getDefaultJournalMaxIoAio(), journalType);
                long nanoseconds = SyncCalculation.toNanos(time, writes, verbose);
@@ -1034,11 +1037,23 @@ public class Create extends InputAbstract {
                                      " writes per millisecond, your journal-buffer-timeout will be " + nanoseconds);
 
                filters.put("${journal-buffer.settings}", readTextFile(ETC_JOURNAL_BUFFER_SETTINGS, syncFilter));
+
+               if (noJournalSync) {
+                  syncFilter.put("${nanoseconds}", "0");
+               } else if (journalType != JournalType.NIO) {
+                  long nioTime = SyncCalculation.syncTest(dataFolder, 4096, writes, 5, verbose, !noJournalSync, false, "journal-test.tmp", ActiveMQDefaultConfiguration.getDefaultJournalMaxIoAio(), JournalType.NIO);
+                  long nioNanoseconds = SyncCalculation.toNanos(nioTime, writes, verbose);
+                  syncFilter.put("${nanoseconds}", Long.toString(nioNanoseconds));
+               }
+
+               filters.put("${page-sync.settings}", readTextFile(ETC_PAGE_SYNC_SETTINGS, syncFilter));
             }
+
 
 
          } catch (Exception e) {
             filters.put("${journal-buffer.settings}", "");
+            filters.put("${page-sync.settings}", "");
             e.printStackTrace();
             System.err.println("Couldn't perform sync calculation, using default values");
          }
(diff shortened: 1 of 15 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -675,6 +675,8 @@ public void parseMainConfig(final Element e, final Configuration config) throws
 
       config.setCriticalAnalyzerPolicy(CriticalAnalyzerPolicy.valueOf(getString(e, "critical-analyzer-policy", config.getCriticalAnalyzerPolicy().name(), Validators.NOT_NULL_OR_EMPTY)));
 
+      config.setPageSyncTimeout(getInteger(e, "page-sync-timeout", config.getJournalBufferTimeout_NIO(), Validators.GE_ZERO));
```
clebertsuconic: I did not see any xsd changes.. or the FileConfiguration parser tests...

brusdev: I'm sorry, I missed to commit the xsd file. I pushed a new commit to add the xsd file and some changes for the configuration tests.

docs/user-manual/en/paging.md
```diff
@@ -171,6 +171,13 @@ Once that limit is reached any message will be blocked. (unless the protocol
 doesn't support flow control on which case there will be an exception thrown
 and the connection for those clients dropped).
 
+## Page Sync Timeout
+
+The pages are synced periodically and the sync period is configured through
```
clebertsuconic: can you change the CLI create to perform an NIO test.. and set this value with the same amount as NIO.

if NIO, this value should be the same .. if AIO a separate test should be done, and then applied to page-sync timeout.

brusdev: I pushed a new commit that changes the CLI create to autotune page-sync-timeout too.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -1002,43 +1003,51 @@ private void applyAddressesAndQueues(HashMap<String, String> filters) {
...
-               String writesPerMillisecondStr = new DecimalFormat("###.##").format(writesPerMillisecond);
-
                HashMap<String, String> syncFilter = new HashMap<>();
                syncFilter.put("${nanoseconds}", Long.toString(nanoseconds));
                syncFilter.put("${writesPerMillisecond}", writesPerMillisecondStr);
                syncFilter.put("${maxaio}", journalType == JournalType.ASYNCIO ? "" + ActiveMQDefaultConfiguration.getDefaultJournalMaxIoAio() : "1");
 
-               System.out.println("done! Your system can make " + writesPerMillisecondStr +
-                                     " writes per millisecond, your journal-buffer-timeout will be " + nanoseconds);
+               System.out.println("Your journal-buffer-timeout will be " + nanoseconds);
```
clebertsuconic: I would keep this writers per millisecond information ... it's been quite useful for me when I'm dealing with certain slow devices and network.... that's me as an user of my own code there.

brusdev: I moved that println at line 1018, because the test is always executed.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -1002,43 +1003,51 @@ private void applyAddressesAndQueues(HashMap<String, String> filters) {
...
                filters.put("${journal-buffer.settings}", readTextFile(ETC_JOURNAL_BUFFER_SETTINGS, syncFilter));
             }
 
+            HashMap<String, String> syncFilter = new HashMap<>();
+            syncFilter.put("${nanoseconds}", Long.toString(nanoseconds));
+
+            System.out.println("Your page-sync-timeout will be " + nanoseconds);
+
+            filters.put("${page-sync.settings}", readTextFile(ETC_PAGE_SYNC_SETTINGS, syncFilter));
+
```
wy96f: I think when noJournalSync is true ie disable usage of fdatasync, we should set page-sync-timeout to 0 for we won't call fsync and enabling `PageSyncTimer` does nothing but increases producer send latency.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -1002,43 +1003,51 @@ private void applyAddressesAndQueues(HashMap<String, String> filters) {
...
       } else {
          try {
             int writes = 250;
             System.out.println("");
             System.out.println("Auto tuning journal ...");
 
+            long time = SyncCalculation.syncTest(dataFolder, 4096, writes, 5, verbose, !noJournalSync, false, "journal-test.tmp", ActiveMQDefaultConfiguration.getDefaultJournalMaxIoAio(), journalType);
+            long nanoseconds = SyncCalculation.toNanos(time, writes, verbose);
+            double writesPerMillisecond = (double) writes / (double) time;
+            String writesPerMillisecondStr = new DecimalFormat("###.##").format(writesPerMillisecond);
```
wy96f: If AIO or MAPPED, we should run another test using `NIOSequentialFile` to apply to page-sync-timeout.

wy96f: If AIO or MAPPED, we should run another test using `NIOSequentialFile` to apply to page-sync-timeout.

clebertsuconic: I thought this was done as part of this change?

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
brusdev: I pushed a commit to set page-sync-timeout to 0 when noJournalSync is true.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
brusdev: > If AIO or MAPPED, we should run another test using `NIOSequentialFile` to apply to page-sync-timeout.

Do you mean `SyncCalculation.syncTest(..., JournalType.NIO)`?

wy96f: Yes, we need to call it if `journalType` is not nio :)

brusdev: Done.

clebertsuconic:
This all looks good to me.


Is this squashed in a way you want to merge, or you should squash it?

Is this ready to be merged?

clebertsuconic:
I would squash them on a single commit. don't know if you intend this as is (in case you want to have this cherry-pickable into a different branch or something).

brusdev:
IMO It is ready to be merged. I squashed them on a single commit.

PR #2871, opened 2019-10-25
NO-JIRA improve transformer docs

jbertram:


jbertram:
Awhile back transformers were updated to support configurable properties, but none of the related docs were updated. This update bridges that gap.

PR #2872, opened 2019-10-25
ARTEMIS-2531: Fix filter in FederatedQueue to prevent infinite consumer creation in a circular or bidrectional setup

cshannon:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/federation/FederatedQueueTest.java
```diff
@@ -177,6 +177,11 @@ public void testFederatedQueueBiDirectional() throws Exception {
          producer1.send(session1.createTextMessage("hello"));
          assertNotNull(consumer0.receive(1000));
 
+         //Sleep to see if extra consumers are created - this tests to make sure there is no loop and tests the FederatedQueue metaDataFilterString
+         //is working properly - should only be 1 consumer on each (1 for the local consumer on broker0 and 1 for the federated consumer on broker1)
+         Thread.sleep(1000);
+         assertEquals(1, ((QueueBinding) getServer(0).getPostOffice().getBinding(SimpleString.toSimpleString(queueName))).consumerCount());
+         assertEquals(1, ((QueueBinding) getServer(1).getPostOffice().getBinding(SimpleString.toSimpleString(queueName))).consumerCount());
```
jbertram: One thought here...Instead of using `Thread.sleep(1000)` you could do something like this:
```
assertFalse(Wait.waitFor(() -> getServer(0).locateQueue(SimpleString.toSimpleString(queueName)).getConsumerCount() > 1, 500, 100));
assertFalse(Wait.waitFor(() -> getServer(1).locateQueue(SimpleString.toSimpleString(queueName)).getConsumerCount() > 1, 500, 100));
```
I used 500 on both calls because you were originally sleeping for 1000ms and 1000/2=500. The advantage here (albeit small) is that the test would fail faster. Also, using `locateQueue()` is a bit more elegant than `getPostOffice().getBinding()`.

I don't feel strongly about the change. Just thought I'd throw it out there.

cshannon: Yeah that's fine, i can go ahead and change that.

cshannon:
@jbertram - I updated the test

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/queue/FederatedQueue.java
```diff
@@ -63,7 +63,7 @@ public FederatedQueue(Federation federation, FederationQueuePolicyConfiguration
       Objects.requireNonNull(config.getName());
       this.config = config;
       this.priorityAdjustment = federationUpstream.getPriorityAdjustment() + (config.getPriorityAdjustment() == null ? -1 : config.getPriorityAdjustment());
-      String metaDataFilterString = config.isIncludeFederated() ? null : FederatedQueueConsumer.FEDERATION_NAME +  " IS NOT NULL";
+      String metaDataFilterString = config.isIncludeFederated() ? null : "hyphenated_props:" + FederatedQueueConsumer.FEDERATION_NAME +  " IS NOT NULL";
```
michaelandrepearce: It is meant to be circular in a directional setup, if theres a local consumer messages wont go back but should a message move to a broker and then consumer go then message should return back to first broker if there is one there

michaelandrepearce: If you want to restrict that then you should use hops to limit it as per docs

michaelandrepearce: -1 as this would break the original intent.

cshannon:
@michaelandrepearce - I'm not really sure what you are talking about but maybe you think I'm referring to addresses.  I'm talking about queues...and there is no max hops flag.  Also the docs specifically say there's a "includeFederated" flag to enable allowing the propagation of federated consumers.  In this case I'm simply using what is already there...but the filter you created is wrong, I'm just fixing the bug with the filter and not changing the intended behavior.

cshannon:
@michaelandrepearce - As a follow up to maybe make it more clear, the fix is fixing the flag that already exists called `include-federated` as described here: https://activemq.apache.org/components/artemis/documentation/latest/federation-queue.html and not changing behavior at all. 

If anything this is making sure the intended behavior actually happens correctly as documented because as you will see if you run the FederatedQueueTest I updated unchanged without the fix it breaks because a ton of consumers get created when they should have been excluded as the filter that was created (by the flag `include-federated` being false) doesn't work as intended because of the hyphen.

michaelpearce-gain:
@cshannon so thats why the metadata flag exists, FederatedQueueConsumer.FEDERATION_NAME is put on the session meta data see FederatedQueueConsumer,, which is added .  nowhere in code base can i find anything appending a "hyphenated_props"

cshannon:
@michaelandrepearce - So maybe I didn't explain the original intent of this as well.  The ONLY problem here is that while the existing filter is correct it does not actually work properly because the SQL92 parsing logic breaks because the key for FederatedQueueConsumer.FEDERATION_NAME has a hyphen in it as the key name is `federation-name`  Because of the hyphen you have to prefix the key with hyphenated_props in order for the filter to actually work.  That's it...nothing else is wrong with the filter.  The other solution is to simply change the name of the key to not have a hyphen and problem fixed.  This is just a one line fix.

Take a look at this documentation for filter expressions at the bottom: https://activemq.apache.org/components/artemis/documentation/latest/filter-expressions.html

> The JMS spec also states that property identifiers (and therefore the identifiers which are valid for use in a filter expression) are an, "unlimited-length sequence of letters and digits, the first of which must be a letter. A letter is any character for which the method Character.isJavaLetter returns true. This includes _ and $. A letter or digit is any character for which the method Character.isJavaLetterOrDigit returns true." This constraint means that hyphens (i.e. -) cannot be used. **However, this constraint can be overcome by using the hyphenated_props: prefix.For example, if a message had the foo-bar property set to 0 then the filter expression hyphenated_props:foo-bar = 0 would match it.**

If you are still confused or don't quite get what I am talking about I encourage you to run my updated test with and without the fix and you can see what I mean.  Without the fix the brokers will create infinite numbers of consumers because the filter is ignored because of the hyphen.  With the fix everything works as originally intended.  If you use break points in the debugger you will see the expression get evaluated to false even though it should be true and not work right.

So again, I'm NOT trying to change the original intent..Im simply trying to actually fix what's supposed to already be there.  The original filter is correct and will work as intended if the parser can actually handled the hyphen.

michaelandrepearce:
Ok gotcha now. +1

PR #2879, opened 2019-10-30
ARTEMIS-2535 Add ignorePartialResultException option to LDAPLoginModule

SmithJosh:
Active Directory servers are unable to handle referrals automatically. This causes a PartialResultException to be thrown if a referral is encountered beneath the base search DN, even if `Context.REFERRAL` is set to `ignore`. This PR adds an option to ignore these exceptions, allowing login to proceed with the query results received before the exception was encountered.

This is the same solution used by Spring. See https://github.com/spring-projects/spring-ldap/blob/master/core/src/main/java/org/springframework/ldap/core/LdapTemplate.java#L379-L387

Issue: https://issues.apache.org/jira/browse/ARTEMIS-2535

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/LDAPLoginModule.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/LDAPLoginModule.java
@@ -23,6 +23,7 @@ import javax.naming.Name;
 import javax.naming.NameParser;
 import javax.naming.NamingEnumeration;
 import javax.naming.NamingException;
+import javax.naming.PartialResultException;
 import javax.naming.directory.Attribute;
 import javax.naming.directory.Attributes;
 import javax.naming.directory.DirContext;
@@ -84,6 +85,7 @@ public class LDAPLoginModule implements LoginModule {
    private static final String SASL_LOGIN_CONFIG_SCOPE = "saslLoginConfigScope";
    private static final String AUTHENTICATE_USER = "authenticateUser";
    private static final String REFERRAL = "referral";
+   private static final String IGNORE_PARTIAL_RESULT_EXCEPTION = "ignorePartialResultException";
    private static final String PASSWORD_CODEC = "passwordCodec";
    private static final String CONNECTION_POOL = "connectionPool";
    private static final String CONNECTION_TIMEOUT = "connectionTimeout";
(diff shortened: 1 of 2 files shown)


SmithJosh:
This is Active Directory-specific behavior. I tried, but couldn't reproduce the issue with the ApacheDS test server

jbertram:
I haven't verified this, but based on the aforementioned [commit in the Karaf code-base](https://github.com/apache/karaf/commit/2a1b963d2f958f70f6968f8f6896fe6c95faf656) it does appear that the behavior can be replicated in ApacheDS.

If there's no way to test this in any way I'm not sure the PR should be merged. On what basis can we say that the code actually works?

SmithJosh:
Thanks for the link. I actually tried to write a unit test based on that Karaf example, but the Karaf test doesn't actually reproduce the issue (again it's AD-specific behavior). Their test passes regardless the value of `ignorePartialResultException`

jbertram:
Is this behavior acknowledge by Microsoft as a defect? If so, are there any plans to fix it. If not, why not?

SmithJosh:
That I don't know. When I was researching this I didn't find any information from Microsoft. But it looks like a pretty old issue that many projects have had to workaround with their LDAP implementations:

Spring's had this workaround since v1.0: https://github.com/spring-projects/spring-ldap/blob/6a9bde34811b87b5425c05068a31ff61d7e59170/changelog.txt#L554-L556
Here's a redhat bug from 2010: https://bugzilla.redhat.com/show_bug.cgi?id=582471

I can see if I can find any official info from Microsoft

jbertram:
Given that this issue looks to be well known as well as the relative simplicity of the code changes I'm inclined to give this my blessing.

That said, this can't be merged as is. The PR test suite failed due to the failure of `org.apache.activemq.artemis.core.security.jaas.LDAPLoginModuleTest#testLogin`. I pulled your PR branch and ran the test locally and I can confirm the failure isn't spurious. To my surprise, the problem appears to be the removal of a seemingly innocuous bit of code that isn't even really related to the goal of this PR. In the method `org.apache.activemq.artemis.spi.core.security.jaas.LDAPLoginModule#resolveDN` this `if` statement was removed:
```
if (results.hasMore()) {
   // ignore for now
}
```
I'm not sure what but *something* is happening in that call which is important. If removed the aforementioned test fails. If added back the test succeeds.  The test also succeeds if the `if` statement is replaced with this:
```
results.close();
```
My recommendation would be to just leave the `if` statement alone since it's not directly related to your PR.

Also, I recommend adding something to your commit statement about the lack of testing so it will be clear in the future to whatever poor soul is digging through this code.

Once the PR build passes successfully I'll merge it.

SmithJosh:
Whoops, I'll check it out. That if statement was causing a PartialResultException, which is why I removed it. Should be able to wrap in a try/catch instead.

So I did a little more digging. Still didn't find anything from Microsoft, but the root issue is that if `Context.REFERRAL` is set to `ignore`, JNDI expects target directory services to support the `ManageDsaIT` control, which Active Directory does not. See https://docs.oracle.com/javase/jndi/tutorial/ldap/referral/jndi.html

SmithJosh:
@jbertram tests are passing and the commit message is updated as you requested

jbertram:
Merged it. Nice work, and thanks for the contribution!

PR #2881, opened 2019-11-04
ARTEMIS-2538 Removing all messages from a huge queue causes OOM

brusdev:
The iterQueue transaction commits are locked by the synchronization
context. So removing all messages from a huge queue causes the creation
of too locked transactions for the paged messages and so OOM.
The iteration on paged message is executed out the iterQueue
synchronization context to avoid to lock the transaction commits.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -227,6 +227,8 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    //This lock is used to prevent deadlocks between direct and async deliveries
    private final ReentrantLock deliverLock = new ReentrantLock();
 
+   private final ReentrantLock depageLock = new ReentrantLock();
+
    private volatile boolean depagePending = false;
 
    private final StorageManager storageManager;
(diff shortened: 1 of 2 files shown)


clebertsuconic:
No test possible?

I understand it's not possible to write a test that would run for 10 minutes... but is there anything that can at least exercise more than one batch on the tx.commit() while removing?

We wouldn't be testing for the OME as that''s not feasible here.. but at least making sure the semantic works as expected?

clebertsuconic:
there was no other semantic changes here?

brusdev:
I'm moving the iteration on paged message out the iterQueue
synchronization context as I tried with the commit https://github.com/apache/activemq-artemis/commit/28d1a53630bea2ce9ef137cda4724fbf106dd96e but that commit contains a mistake and it is partially reverted by the commit https://github.com/apache/activemq-artemis/commit/84067d8fef311d641ef7d35b9cdda3bc48759a37.

brusdev:
I added a test to check how many iterQueue transactions are locked on afterCommit.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3880,9 +3889,12 @@ private DepageRunner(boolean scheduleExpiry) {
       @Override
       public void run() {
          try {
+            depageLock.lock();
```
franz1981: move `depageLock.lock()` right before the `try` statement

PR #2882, opened 2019-11-05
ARTEMIS-2504 fix divert test

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/divert/DivertTest.java
```diff
@@ -169,7 +169,7 @@ public void testCrossProtocol() throws Exception {
 
          session.createQueue(new SimpleString(testForConvert), RoutingType.ANYCAST, SimpleString.toSimpleString(testForConvert), null, true);
 
-         session.createQueue(new SimpleString(forwardAddress), RoutingType.MULTICAST, queueName2, null, true);
```
clebertsuconic: what made the test fail after the Ring Queue changes? 

was there any semantic change for regular cases? That's my concern here.

jbertram: @gemmellr, not a nitpick, IMO. Well spotted.

PR #2883, opened 2019-11-05
ARTEMIS-2540 Display LargeMessage column in message browser of admin UI

sebthom:


clebertsuconic:
nice one! thanks! tested and it works

PR #2884, opened 2019-11-06
NO-JIRA Fix indention in browse.js

sebthom:
This only changes the formatting, no functional changes. This will make it easier to review future functional changes of the file.


sebthom:
The github diff mechanism apparently isn't very intelligent, you can use this link https://github.com/apache/activemq-artemis/pull/2884/files?w=1 to see that no functional changes were applied.

PR #2885, opened 2019-11-07
ARTEMIS-2544 Remove rolledback PageTransactionInfo to free up memory

wy96f:
If page transaction is rolled back, it will reside in transactions map unless broker restarts. Remove it when messages in the transaction are all acked.


michaelandrepearce:
@wy96f i think you can merge this, seems no one has given opposition, and tests pass

clebertsuconic:
@wy96f did you run a complete testsuite on this change?

@michaelandrepearce for changes like this we should run the whole testsuite... we should at least indicate if that was executed or not.

wy96f:
@clebertsuconic I didn't run a complete testsuite, but I added a test specifically for this change.

clebertsuconic:
@wy96f Ignore my last comment.  I was dealing with my Ci broken.  It was out of space. 

Funny thing is that I brought to another box that was also out of space.  Which threw me off.  And it was all failing on blocking paging due to out of space. 


Sorry about that.

PR #2886, opened 2019-11-07
ARTEMIS-2541 Improve message browser of Admin UI

sebthom:


clebertsuconic:
any chance you could attach before and after images to the PR discussion?

sebthom:
@clebertsuconic I added another screenshot before the modifications.

artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/browse.js
```diff
@@ -26,44 +26,55 @@ var ARTEMIS = (function(ARTEMIS) {
...
+         width: '50px'
+      }, {
+         field: 'persistentSize',
+         displayName: 'Size',
+         cellTemplate: '<div class="ngCellText" title="{{row.entity.persistentSize}}">{{formatPersistentSize(row.entity.persistentSize)}}</div>',
+         width: '100px'
+      }, {
+         field: 'userID',
+         displayName: 'User ID',
+         cellTemplate: '<div class="ngCellText">{{row.entity.userID ? row.entity.userID : row.entity.StringProperties._AMQ_VALIDATED_USER}}</div>',
```
sebthom: @clebertsuconic I actually would prefer to show the _AMQ_VALIDATED_USER value by default and falling back to userID if it is not present. WDYT?

clebertsuconic: thanks.. I didnt realize that.. thakns

clebertsuconic: > @clebertsuconic I added another screenshot before the modifications to the JIRA issue.
thanks a lot, I didn't realize that before.

clebertsuconic: > @clebertsuconic I actually would prefer to show the _AMQ_VALIDATED_USER value by default and falling back to userID if it is not present. WDYT?

would that be intuitive to users? shouldn't we keep it separated columns?
as long as users can understand the result without having to read any doc about it.

sebthom: > > @clebertsuconic I actually would prefer to show the _AMQ_VALIDATED_USER value by default and falling back to userID if it is not present. WDYT?
> 
> would that be intuitive to users? shouldn't we keep it separated columns?
> as long as users can understand the result without having to read any doc about it.

Sure, we could do that too.

sebthom: @clebertsuconic I added a separated validated user column and updated the screenshot in the JIRA issue.

clebertsuconic:
I checked AMQP, and this property is only filled with Core Protocol. I checked and it worked fine.

Perhaps we could add with extra properties.

clebertsuconic:
but that will be an improvement. I'm merging this.

PR #2887, opened 2019-11-08
NO-JIRA make OpenWire test more robust

jbertram:
This test fails intermittently due to a race condition. The added code resolves the race.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/cluster/TemporaryQueueClusterTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/cluster/TemporaryQueueClusterTest.java
@@ -17,6 +17,8 @@
 package org.apache.activemq.artemis.tests.integration.openwire.cluster;
 
 import org.apache.activemq.artemis.api.core.RoutingType;
+import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.utils.Wait;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/cluster/TemporaryQueueClusterTest.java
```diff
@@ -64,20 +66,23 @@ public void testClusteredQueue() throws Exception {
...
          this.waitForBindings(servers[0], QUEUE_NAME, true, 1, 0, 2000);
          this.waitForBindings(servers[1], QUEUE_NAME, true, 1, 0, 2000);
          this.waitForBindings(servers[1], QUEUE_NAME, false, 1, 0, 2000);
          this.waitForBindings(servers[0], QUEUE_NAME, false, 1, 0, 2000);
 
-
          MessageProducer prod1 = session1.createProducer(targetQueue1);
          MessageConsumer cons2 = session2.createConsumer(targetQueue2);
 
+         this.waitForBindings(servers[0], QUEUE_NAME, false, 1, 1, 2000);
```
michaelandrepearce: Should be able to send immediately.  A real use case wouldnt wait. If its failing without would suggest annissue that needs to be fixed.

clebertsuconic: This test is about clustering. a lot of tests will make sure the other servers will have their remote bindings before validating a cluster scenario.

This should work fine on a local send. But since the test is validating a cluster scenario, this must be done to guarantee the cluster formation.

I think this is fine. There are a lot of other clustering tests that will wait for the expected binding.

jbertram: I agree with @clebertsuconic here. This pattern is followed by most, if not all, of our clustering tests.

michaelandrepearce: Still dont fully get why being clustered or not is any difference as an end user would just create a producer an immediately send no differently e.g. they would be agnostic. But as its seems to done by other tests ok.

clebertsuconic: It's not about clustering.. it's about the test.

The test to be valid, clustering must be formed properly. otherwise it's just like sending locally.

clebertsuconic: That wait is just waiting for the conditions on what the test was meant to be working.

jbertram: To be clear, the issue isn't a race between creating a producer and sending the message. It's between creating the consumer on node 2 of the cluster and sending the message to node 1. When the consumer is created on node 2 the broker tells node 1 about it so messages can be routed from node 2 to node 1 if necessary. This happens asynchronously which introduces the possibility of a race (i.e. if the message is sent to node 1 before it knows about the consumer on node 2).

PR #2888, opened 2019-11-08
Enable Continuation WebSocket Frame support

3venthorizon:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/protocol/websocket/WebSocketServerHandler.java
```diff
@@ -117,7 +118,8 @@ private boolean handleWebSocketFrame(ChannelHandlerContext ctx, WebSocketFrame f
       } else if (frame instanceof PingWebSocketFrame) {
          ctx.writeAndFlush(new PongWebSocketFrame(frame.content().retain()));
          return false;
-      } else if (!(frame instanceof TextWebSocketFrame) && !(frame instanceof BinaryWebSocketFrame)) {
+      } else if (!(frame instanceof TextWebSocketFrame) && !(frame instanceof BinaryWebSocketFrame) &&
```
clebertsuconic: small change! zero test on a core hot path functionality!!! Really dangerous to merge this man!

Can you provide tests for this? or some background on why a test is not needed? (would this be covered by existing tests? which I don't believe it's the case).

clebertsuconic: I will close this for now: 

you would need to provide:

- A JIRA
- a test as part of your PR
- change your commit name to include the JIRA name.


I will have to close it, as this is not even close to something I would be able to merge.

clebertsuconic:
lets reopen it when you provide those things.

Sorry about that.. let me know when I can help you to make progress here.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/protocol/websocket/WebSocketServerHandler.java
3venthorizon: Thank you I will create the Jira ticket etc.

PR #2889, opened 2019-11-08
ARTEMIS-1925 combine STRICT and OFF with redistribution

jbertram:
This commit adds two new message-load-balancing types:
STRICT_WITH_REDISTRIBUTION and OFF_WITH_REDISTRIBUTION. These allow
redistribution with the semantics of the existing STRICT and OFF
message-load-balancing types respectively. Previously you could only get
redistribution if you were using ON_DEMAND. The semantics for STRICT and
OFF were not changed so as not to impact existing users.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
@@ -172,7 +172,7 @@ public class Create extends InputAbstract {
    @Option(name = "--max-hops", description = "Number of hops on the cluster configuration")
    private int maxHops = 0;
 
-   @Option(name = "--message-load-balancing", description = "Load balancing policy on cluster. [ON_DEMAND (default) | STRICT | OFF]")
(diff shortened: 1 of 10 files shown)


clebertsuconic:
only thing I don't like the term OFF_WITH_REDISTRIBUTION

What about REDISTRIBUTION_ONLY ?

jbertram:
That's a fair point. I chose to use the `_WITH_REDISTRIBUTION` suffix to maintain continuity with the existing setting names so it's as simple as possible for users to switch if they want. Ultimately my goal is to remove these two new settings in the next major release and properly isolate the behavior of redistribution to the `redistribution-delay` setting alone. I didn't do that here so as not to impact semantics for existing users.

AntonRoskvist:
Hi,

I had a test with the load balancing type "OFF_WITH_REDISTRIBUTION" and fail to see a difference in the semantics of straight up "ON_DEMAND" load balancing. All incoming load was distributed evenly to all nodes that had registered consumers for my testing queue. The redistribution-delay made no difference to this behavior, even when I increased it to 60000 milliseconds (just for verification).

Am I missing something?

jbertram:
@AntonRoskvist, no, I think there's something wrong with the PR. I'll take a closer look.

clebertsuconic:
@jbertram when you reopen, please consider renaming OFF_WITH_REDISTRIBUTION to REDISTRIBUTION ? OFF_WITH_SOMETHING is not really OFF.. makes it difficult to understand.

PR #2892, opened 2019-11-12
ARTEMIS-2549 - Add Downstream support to Federation

cshannon:
This commit introduces the ability to configure a downstream connection
for federation.  This works by sending information to the remote broker
and that broker will parse the message and create a new upstream back
to the original broker.

In the future we could look at enhancements with this including only needing 1 connection instead of 2 when doing duplex but for the first version I thought it was best to just re-use the existing upstream base to make it simpler and just create a new connection. Plus having separate connections for the downstream/upstream has advantages too in that you don't interfere with each other when things like flow control kick on (I had this problem with duplex bridging in 5.x)

I have a couple more PRs to submit after this including plugin support, metrics support, and also divert binding demand.


michaelandrepearce:
@cshannon i saw this, we just hit a nasty issue, fix clebert merged yesterday. But the washup and sorting that is taking my focus. Will look over this asap after that.

cshannon:
Alright, sounds good

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/config/federation/FederationStreamConfiguration.java
```diff
@@ -33,7 +35,7 @@ public String getName() {
       return name;
    }
 
-   public FederationUpstreamConfiguration setName(String name) {
+   public FederationStreamConfiguration setName(String name) {
```
michaelandrepearce: Can you make the class <T extends FederationStreamConfiguration> and then here, and other returns, return T, this then mans when its returned you get the correct type, which would have any extra setters or getters.

cshannon: @michaelandrepearce - Good point, I made the change and pushed an update to the branch

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/address/FederatedAddress.java
```diff
@@ -113,7 +116,31 @@ public void start() {
     */
    @Override
    public synchronized void afterCreateQueue(Queue queue) {
-      createRemoteConsumer(queue);
+      conditionalCreateRemoteConsumer(queue);
+   }
+
+   private void conditionalCreateRemoteConsumer(Queue queue) {
+      //Always create demand if not a federated queue
+      if (!queue.getName().toString().startsWith(FEDERATED_QUEUE_PREFIX)) {
```
michaelandrepearce: @cshannon im still struggling to understand why this change is needed. It seems to be breaking some setups i have.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/address/FederatedAddress.java
```diff
@@ -113,7 +116,31 @@ public void start() {
...
+         createRemoteConsumer(queue);
+      } else {
+         //For federated queues we want to make sure that we aren't creating demand
+         //if the federated queue is part of the same federation stream as we would get into a situation
+         //where a local consumer on broker A created a federated queue on broker B and in turn broker
+         //A would respond to the federated queue on B and created a federated queue on A even though there is
+         //no local demand on B
+
+         //For queues part of a different federation stream we want to create demand
+         String[] federationToken = queue.getName().toString().split("\\.");
```
michaelandrepearce: Where is the "." coming from? Is it wildcard settings separator? If so will need to honor that.

michaelandrepearce:
@cshannon im pretty happy to merge this, i have one case that's not working as was originally expecting, i think its around the address federation createconsumer change. It could well be also we simply had a bug before thats now resolved. Maybe we could organise a call to discuss?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/address/FederatedAddress.java
cshannon: @michaelandrepearce - This change is primarily there for the use case of 2 brokers each configured with a downstream with each other.  (so bi-directional).  If a consumer is created on broker0 then a federated consumer should be created on broker 1 and there should just be 1 consumer on each broker.  However, without this change you end up with a federated consumer on broker 0 as well (so 2 total consumers) eventhough there is no demand from a local consumer on broker 1.  I will push an update in a minute with a new test to show this.

cshannon:
@michaelandrepearce - I just pushed an update an added a new test to FederatedAddress called testDownstreamFederatedAddressReplicationConsumerCounts.   https://github.com/apache/activemq-artemis/commit/31aa464bafa9a8c1f6d6c2861ffc70981aa24199#diff-6585fa255ca2a39e4fa2a48bf28e6947R123

Take a look at that and see if that makes sense, you will see the test break if you revert that conditional method check.  Unfortunately this right now fixes the two downsteram case but with two upstreams the extra consumer is still created which I want to fix in another update but is more tricky however I wouldn't expect this change to break existing use cases.

If you still have questions we can chat more, we can setup a call or using Slack might be easier which you can message me on

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/address/FederatedAddress.java
michaelpearce-gain: On federated address through it wasn't meant to be demand based on consumer, it was purely meant to be demand based on if there is a queue, e.g. multicast dispatch over the joined brokers. e.g. producers producers to "foo" on brokerA if brokerB had a queue under the address then a remote queue would be created on brokerA and the message would dispatch to brokerB, regardless if that queue had subscriber or not, e.g. handles cases of durable subscriptions, what was nice, is JMS non durable subs, as soon as consumer goes anyhow queue would go, and thus demand goes, as such demand just needed to be down to if queue or not.

michaelpearce-gain: e.g. address level federation was desgined to be queue demand based, and queue level federation is designed to be consumer demand based

cshannon:
@michaelandrepearce - I reverted that small change as I think it needs to be revisited anyways (it doesn't quite work the way I intended after digging into it more) plus with plugin support coming it may be unnecessary as it could go into a custom plugin.

Take a look and see if you are ok merging it now.

michaelpearce-gain:
Looks good to me +1 i can merge tonight when i get home from work, or feel free to merge yourself.

cshannon:
Cool thanks, I will go ahead and just do the merge myself then.

PR #482, opened 2016-04-22
ARTEMIS-214 exception when sending 10k message

gaohoward:

PR #2893, opened 2019-11-15
ARTEMIS-1925 combine STRICT and OFF with redistribution

jbertram:
This commit adds two new message-load-balancing types:
STRICT_WITH_REDISTRIBUTION and REDISTRIBUTION_ONLY. These allow
redistribution with the semantics of the existing STRICT and OFF
message-load-balancing types respectively. Previously you could only get
redistribution if you were using ON_DEMAND. The semantics for STRICT and
OFF were not changed so as not to impact existing users.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
@@ -175,7 +175,7 @@ public class Create extends InputAbstract {
    @Option(name = "--max-hops", description = "Number of hops on the cluster configuration")
    private int maxHops = 0;
 
-   @Option(name = "--message-load-balancing", description = "Load balancing policy on cluster. [ON_DEMAND (default) | STRICT | OFF]")
+   @Option(name = "--message-load-balancing", description = "Load balancing policy on cluster. [ON_DEMAND (default) | STRICT | OFF | STRICT_WITH_REDISTRIBUTION | REDISTRIBUTION_ONLY]")
    private MessageLoadBalancingType messageLoadBalancing = MessageLoadBalancingType.ON_DEMAND;
 
    @Option(name = "--replicated", description = "Enable broker replication")
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/Validators.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/Validators.java
@@ -219,7 +219,9 @@ public final class Validators {
          String val = (String) value;
          if (val == null || !val.equals(MessageLoadBalancingType.OFF.toString()) &&
             !val.equals(MessageLoadBalancingType.STRICT.toString()) &&
-            !val.equals(MessageLoadBalancingType.ON_DEMAND.toString())) {
+            !val.equals(MessageLoadBalancingType.ON_DEMAND.toString()) &&
+            !val.equals(MessageLoadBalancingType.REDISTRIBUTION_ONLY.toString()) &&
+            !val.equals(MessageLoadBalancingType.STRICT_WITH_REDISTRIBUTION.toString())) {
             throw ActiveMQMessageBundle.BUNDLE.invalidMessageLoadBalancingType(val);
          }
       }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/Binding.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/Binding.java
@@ -46,6 +46,8 @@ public interface Binding extends UnproposalListener {
 
    boolean isHighAcceptPriority(Message message);
 
+   boolean isHighAcceptPriority(Message message, boolean redistributing);
+
    boolean isExclusive();
 
    Long getID();
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/BindingsImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/BindingsImpl.java
@@ -200,7 +200,7 @@ public final class BindingsImpl implements Bindings {
 
    @Override
    public boolean allowRedistribute() {
-      return messageLoadBalancingType.equals(MessageLoadBalancingType.ON_DEMAND);
+      return messageLoadBalancingType.equals(MessageLoadBalancingType.ON_DEMAND) || messageLoadBalancingType.equals(MessageLoadBalancingType.STRICT_WITH_REDISTRIBUTION) || messageLoadBalancingType.equals(MessageLoadBalancingType.REDISTRIBUTION_ONLY);
    }
 
    @Override
@@ -257,7 +257,7 @@ public final class BindingsImpl implements Bindings {
 
          Filter filter = binding.getFilter();
 
-         boolean highPrior = binding.isHighAcceptPriority(message);
+         boolean highPrior = binding.isHighAcceptPriority(message, true);
 
          if (highPrior && binding.getBindable() != originatingQueue && (filter == null || filter.match(message))) {
             theBinding = binding;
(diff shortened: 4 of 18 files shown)


jbertram:
@AntonRoskvist, can you give this branch a test?

AntonRoskvist:
Sure can, @jbertram .

Initial tests are looking great! I will keep this running in a "busy" test environment over the weekend to weed out any potential issues, but for the past few hours it's looking good. The system is _mostly_ behaving as expected, with the load balancing type "REDISTRIBUTION_ONLY" configured. Forwards are WAY down and the system is running much smoother.

The only odd behavior I noticed  was in regards to queues that require redistribution. As an example: If a consumer is connected to broker "1" on a new queue and then messages get produced to broker "2", the messages will not get forwarded but rather backs up on broker "2". If I disconnect and reconnect the consumer on broker "1", then redistribution starts and keeps on going as expected.

If messages are produced on broker "2" first and a consumer later connects to broker "1", then all works as expected. Otherwise it's looking fantastic!

Br,
Anton

jbertram:
@AntonRoskvist so the "odd" behavior your seeing isn't specifically related to this PR. It is simply the way the broker deals with auto-created queues in a cluster at the moment. I'll try to explain...

Each broker has its own set of addresses and queues - even in a cluster. When a queue is created on one broker in a cluster the broker on which the queue is created creates what we call a "local" binding. It then informs all the other nodes in the cluster about this binding at which point the other nodes create a "remote" binding. Therefore, at this point all the nodes in the cluster know about the queue. Likewise, when a consumer is created on a queue on one broker all the other nodes are informed about it. *If* another broker in the queue has a matching _local_ binding with the same name (i.e. the queue has actually been created on that broker) then a "redistributor" will be created on that queue so that messages sent to it can be redistributed if necessary. In your use-case there is no local binding so the redistributor isn't created.

As I mentioned previously, this issue wasn't caused by this PR. IMO a new JIRA should be opened and a new PR should be sent to address it. In the mean-time you can work-around the issue by pre-creating all the queues for which you need redistribution.

AntonRoskvist:
Oh, okay. In that case I think this PR is looking good. I have been running it through heavy load tests continuously since Friday and I have seen no issues except the one I mentioned above.

I'd be happy to publish a new Jira for the other minor problem, but I am not quite sure how to formulate it... This is the expected behavior right now and we might want to alter how or when redistributors are formed, is that right?

Unfortunately I can not configure static queues as I am working with several hundreds of them, in several sites that are created ad-hoc by external teams/components that I have little to no control over. But I'll work out some temporary solution for that, like forcing disconnects for consumers on the queues that are not currently getting redistributions.

Br,
Anton

clebertsuconic:
My vote is to just respect redistribution-delay with STRICT.

When we wrote STRICT, we thought we didn't need redistribution, as with forward-when-no-consumers=off, it would mean the queue would only receive messages if there was a consumer.


so, if in fact redistribution-delay and STRICT makes sense, just change STRICT. There's no sense on adding a third option.

if you have a strong opinion about it.. call it something else.. I suggest... ROUND_ROBIN, make it the default, and deprecate STRICT. I see no value on STRICT after the change though.

clebertsuconic:
ROUND_ROBIN is actually not a bad name if you want to keep a third option.

clebertsuconic:
ha! I just googled.. and most other-brokers I just found actually call that kind of distribution... guess what? ... "ROUND ROBIN"

clebertsuconic:
well.. it's OFF_WITH_REDISTRIBUTION. I got it wrong.

But still.

jbertram:
All things being equal I would agree with you, @clebertsuconic. However, as noted in the commit message, I decided not to change the semantics of `STRICT` and `OFF` as I didn't want anyone using those settings along with a non-zero `redistribution-delay` to suddenly start having redistribution when up until this commit the `redistribution-delay` would just be ignored in those cases. Certainly an argument could be made that such a configuration is rare and might even be considered "wrong." However, the fact remains that using `STRICT` or `OFF` with a non-zero `redistribution-delay` has always been a technically valid configuration and the documentation explicitly states that one must use `ON_DEMAND` to enable redistribution. I just think changing semantics in a minor version like you're suggesting is an objectively bad thing.

Perhaps this should just wait until the next major release where we can properly break stuff. What do you think?

michaelandrepearce:
@jbertram just reading discussion on this, i agree with you until next major version, we should avoid breaking changes. If you're leaving this parked, any objections if we mark the PR with the "Do not merge yet label" just so its clear and gets skipped when reviewing active PR's?

jbertram:
@michaelandrepearce, I don't have any objections to that. I added the label.

AntonRoskvist:
Just out of curiosity... when can I expect to see this implemented in a released version then? Is there a schedule for major releases?

jbertram:
@AntonRoskvist, it depends on the outcome of this discussion. I'm still waiting to hear back from @clebertsuconic. 

As far as major releases go, there is no schedule.

AntonRoskvist:
Hi,

I don't really know if this is expected behavior or not, but I have run this PR in one environment since it was created and just noticed something. When running with REDISTRIBUTION_ONLY, topics (multicast) are not forwarded across the cluster, so given a cluster of broker A and B, messages produced to broker A only get delivered to subscribers directly connected to broker A. Subscribers connected to broker B gets none. If no consumers are connected to broker A then the message is lost (using non-durable topics). Using ON_DEMAND the redistribution seem to work as expected, but then all anycast queues get distributed as well, which is what I was hoping to avoid in the first place.

I guess this might relate to the other issue I had where remote bindings don't get created unless there is also a local one present?

Besides this I have had no issue running this PR, even ported into the latest version of the broker

Br,
Anton

AntonRoskvist:
@jbertram I have made a fix to the issue I mentioned above and also fixed the conflicts in the tests. How would I go about sending in that fix? Can I just append it here somehow?

jbertram:
@AntonRoskvist, push your branch to GitHub somewhere and let me know where it is. Then I can cherry-pick your commit onto mine and update the PR.

AntonRoskvist:
Okay cool, you can find it here:
https://github.com/AntonRoskvist/activemq-artemis/tree/ARTEMIS-1925-CHANGES
Actual fix is on line #482 in BindingsImpl.java
I built this against tag 2.16.0 as master branch is currently not building

Br,
Anton

jbertram:
@AntonRoskvist, I couldn't find any change on line 482 of the `BindingsImpl` on [commit d4c502d](https://github.com/AntonRoskvist/activemq-artemis/commit/d4c502d245ce4131dba1f5416b55781efe47d401) of your branch. Also, it looks like you squashed your changes and my changes together which makes it hard to identify what you actually did change. In any event, I've resolved the conflicts and pushed again. Let me know how I can get your fix onto my branch.

AntonRoskvist:
@jbertram 
Hmm, I don't really know how to share that in some simple way then...
The reason you are not seeing any change is probably because compared to the master branch there is no change, it's a revert of that code line from what is in your original pr back to what master branch is using. So now it looks like it's on line :489, I changed:

`if ((messageLoadBalancingType.equals(MessageLoadBalancingType.OFF) || messageLoadBalancingType.equals(MessageLoadBalancingType.REDISTRIBUTION_ONLY)) && theBinding instanceof RemoteQueueBinding) {`

back to:

`if (messageLoadBalancingType.equals(MessageLoadBalancingType.OFF) && theBinding instanceof RemoteQueueBinding) {`

It's in the fork I shared, but maybe the best would be if I worked from your new pr and made just this change to that?
Apologies for possibly making this more complicated than it needs to be, I'm not really used to this workflow.

Br,
Anton

jbertram:
@AntonRoskvist all you need to do is take my branch, add your changes on top of it, and then push it to your repo. Then I can take your changes and cherry-pick them onto my branch here.

AntonRoskvist:
@jbertram
Okay, then you can find it here:
https://github.com/jbertram/activemq-artemis/pull/12

PR #2895, opened 2019-11-18
ARTEMIS-2550 Websocket Continuation Frame Support

3venthorizon:
**Websocket Continuation Frame Support:**
- Receiving of Continuation Frames
- Sending Continuation Frames when payload length exceeds the configured maximum payload length by splitting up the remainder of the payload into new Continuation Frames.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/protocol/websocket/WebSocketFrameEncoder.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/protocol/websocket/WebSocketFrameEncoder.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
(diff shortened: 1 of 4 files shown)


jbertram:
This looks good. The only change I'd recommend is renaming `WebSocketContinuationFrameEncoder` to something more generic since it handles both normal binary frames as well as continuation frames. Perhaps just `WebSocketFrameEncoder` would be best.

jbertram:
Also, the commit message should reference the JIRA. See the [Hacking Guide](https://github.com/apache/activemq-artemis/blob/master/docs/hacking-guide/en/maintainers.md#commit-messages) for more details about the commit message.

clebertsuconic:
Thanks for doing this.. it's in very good shape now.

I still need to do some review.. but it's a lot better than what I looked a few weeks ago.

jbertram:
LGTM. Nice work, @3venthorizon!

jbertram:
@clebertsuconic, have you had a chance to review this? Do you think it's ready to merge?

PR #2897, opened 2019-11-18
ARTEMIS-2558 Add the commented out args to dump the java heap on OOME

brusdev:
Add the commented out args to dump the first java heap on OOME for
process and backup the previous java heap on OOME at startup.

### DIFF
--- a/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis
+++ b/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis
@@ -112,6 +112,11 @@ if [ -z "$WILDFLY_COMMON" ] ; then
   WILDFLY_COMMON="$ARTEMIS_HOME/lib/${wildfly-common}"
 fi
 
+if [ -f "$ARTEMIS_OOME_DUMP" ] ; then
+  # backup the last OOME heap dump
+  mv $ARTEMIS_OOME_DUMP $ARTEMIS_OOME_DUMP.bkp
+fi
+
 exec "$JAVACMD" \
     $JAVA_ARGS \
     -Xbootclasspath/a:"$LOG_MANAGER:$WILDFLY_COMMON" \
--- a/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/artemis.profile
+++ b/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/artemis.profile
@@ -19,6 +19,7 @@ ARTEMIS_HOME='${artemis.home}'
 ARTEMIS_INSTANCE='${artemis.instance}'
 ARTEMIS_DATA_DIR='${artemis.instance.data}'
 ARTEMIS_ETC_DIR='${artemis.instance.etc}'
+ARTEMIS_OOME_DUMP='${artemis.instance.oome.dump}'
 
 # The logging config will need an URI
 # this will be encoded in case you use spaces or special characters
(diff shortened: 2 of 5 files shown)


artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/artemis.profile
```diff
@@ -41,6 +41,11 @@ JAVA_ARGS="${java-opts} -XX:+PrintClassHistogram -XX:+UseG1GC -Xms512M -Xmx2G -D
 # Replace "all_pauses.log" with the file name you want to log to.
 # JAVA_ARGS="$JAVA_ARGS -XX:+PrintSafepointStatistics -XX:PrintSafepointStatisticsCount=1 -XX:+PrintGCApplicationStoppedTime -XX:+PrintGCApplicationConcurrentTime -XX:+LogVMOutput -XX:LogFile=all_pauses.log"
 
+#
+# Enables the dumping of the java heap to the file oom_dump.hprof by using the heap profiler (HPROF)
+# when a java.lang.OutOfMemoryError exception is thrown.
+# JAVA_ARGS="$JAVA_ARGS -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=${ARTEMIS_INSTANCE}/oom_dump.hprof"
```
clebertsuconic: this warrants a JIRA.

clebertsuconic: @brusdev  ^^

brusdev:
I created the jira ARTEMIS-2558 and I added some lines to backup the last OOM heap dump at startup.

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis
```diff
@@ -112,6 +112,11 @@ if [ -z "$WILDFLY_COMMON" ] ; then
   WILDFLY_COMMON="$ARTEMIS_HOME/lib/${wildfly-common}"
 fi
 
+if [ -f "$ARTEMIS_INSTANCE/log/oomheapdump.hprof" ] ; then
```
michaelandrepearce: Rem this, also best to use a var for the filename as everyone will have their own naming conventions
 Lastly you may want to have a max history else you can easily fill disk

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/artemis.profile
```diff
@@ -41,6 +41,11 @@ JAVA_ARGS="${java-opts} -XX:+PrintClassHistogram -XX:+UseG1GC -Xms512M -Xmx2G -D
 # Replace "all_pauses.log" with the file name you want to log to.
 # JAVA_ARGS="$JAVA_ARGS -XX:+PrintSafepointStatistics -XX:PrintSafepointStatisticsCount=1 -XX:+PrintGCApplicationStoppedTime -XX:+PrintGCApplicationConcurrentTime -XX:+LogVMOutput -XX:LogFile=all_pauses.log"
 
+#
+# Enables the dumping of the java heap to the file oom_dump.hprof by using the heap profiler (HPROF)
+# when a java.lang.OutOfMemoryError exception is thrown.
+# JAVA_ARGS="$JAVA_ARGS -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=${ARTEMIS_INSTANCE}/log/oom_dump.hprof"
```
michaelandrepearce: Would be good to include timestamp to the heapdump file name

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis
michaelandrepearce: E.g. we where i am atm all heap dumps include the process id, service name (as multiple instances on a host) and timestamp.

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/artemis.profile
brusdev: I set a fixed filename to prevent the creation of multiple heap dumps that can fill the disk.

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis
brusdev: I'll add var for the filename but I set a fixed filename to prevent the creation of multiple heap dumps that can fill the disk.
I added this check because if the file exists, no other dumps are created. The mv command replaces the previous backup so there will be at most 2 dumps.

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/artemis.profile
clebertsuconic: timestamp would be generated based on when the broker started on this case.. not when the dump happened... 

I would keep it simple here.

If you want a variable name.. fine.. but other than that... you start the broker now.. it dumps 5 months from now.. the timestamp is useless.

brusdev:
I added a var for the filename in artemis profile.

PR #2901, opened 2019-11-22
ARTEMIS-2421 periodic journal lock evaluation

jbertram:
If a broker loses its file lock on the journal and doesn't notice (e.g.
network connection failure to an NFS mount) then it can continue to run
after its backup activates resulting in split-brain.

This commit implements periodic journal lock evaluation so that if a live
server loses its lock it will automatically restart itself.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
@@ -22,6 +22,12 @@ import java.io.RandomAccessFile;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 
 import org.apache.activemq.artemis.api.core.ActiveMQIllegalStateException;
 import org.apache.activemq.artemis.api.core.SimpleString;
@@ -54,7 +60,9 @@ public class FileLockNodeManager extends NodeManager {
 
    private static final long LOCK_ACCESS_FAILURE_WAIT_TIME = 2000;
 
-   private FileLock liveLock;
+   private static final int LOCK_MONITOR_TIMEOUT_MILLIES = 2000;
+
+   private volatile FileLock liveLock;
 
    private FileLock backupLock;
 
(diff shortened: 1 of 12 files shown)


jbertram:
Since the original author apparently abandoned #2547 I took their branch, resolved the conflicts, eliminated `Thread.sleep()` from the tests, and added a proper commit message.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -388,4 +404,109 @@ protected FileLock lock(final int lockPosition) throws Exception {
       return null;
    }
 
+   private synchronized void startLockMonitoring() {
```
clebertsuconic: Did you consider extending ActiveMQScheduledComponent instead?

clebertsuconic:
If you lost connectifity into NFS, you would get an IOException and the broker would stop?

or are you concerned on the case where the broker is idle?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
jbertram: No, I didn't consider extending `ActiveMQScheduledComponent`. I just took what was on the original PR and cleaned it up so it could actually be merged.

jbertram:
> or are you concerned on the case where the broker is idle?

@clebertsuconic, yes. I believe that's the use-case here. You can read more about it on [ARTEMIS-2421](https://issues.apache.org/jira/browse/ARTEMIS-2421). It's a pretty straight-forward scenario to reproduce the problem (although I haven't tried it).

clebertsuconic:
@jbertram sorry, I didn't realize where this was coming from.. I see it was a user PR originally.

let me see if I can adapt the ActiveMQScheduledComponent here.although it's not that easy given the location of the component (CLI).

clebertsuconic:
there is a chance of thread leaks on the testsuite. I will run the whole testsuite before merging this.


I already have a commit to use the ScheduledComponent.

clebertsuconic:
Testsuite hung on RealNodeManagerTest. I'm running again, but please don't merge this yet.

clebertsuconic:
The issue is exactly what I mentioned. the use of the ScheduledExecutor left non daemon threads running on a few tests that were using Spawn. Causing the VM to never finish on those cases.

The testsuite didn't go far enough to leak on the Junit Rules that we have written.

But using the ScheduledComponent will provide a fix as it already has the proper Thread factory in place among a few other benefits (avoiding for instance calls happening too fast which could happen when the run takes longer).

PR #2902, opened 2019-11-28
NO-JIRA close page after recovering page counter

wy96f:


franz1981:
Please explain the effects of this change in the comments of the PR, was a leak of some type?

wy96f:
> Please explain the effects of this change in the comments of the PR, was a leak of some type?

Not really. Just close page file in time and not bother finalization.

PR #2903, opened 2019-12-02
ARTEMIS-2565 - Add plugin support for Federated Queues/Addresses

cshannon:
Add a new interface called ActiveMQServerFederationPlugin to allow
customization of the federated queue/address feature of the broker

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedAbstract.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedAbstract.java
@@ -19,14 +19,20 @@ package org.apache.activemq.artemis.core.server.federation;
 
 import java.util.HashMap;
 import java.util.Map;
+
+import org.apache.activemq.artemis.api.core.ActiveMQException;
 import org.apache.activemq.artemis.core.config.WildcardConfiguration;
 import org.apache.activemq.artemis.core.config.federation.FederationTransformerConfiguration;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
-import org.apache.activemq.artemis.core.server.federation.FederatedQueueConsumer.ClientSessionCallback;
+import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
+import org.apache.activemq.artemis.core.server.federation.FederatedQueueConsumerImpl.ClientSessionCallback;
 import org.apache.activemq.artemis.core.server.plugin.ActiveMQServerBasePlugin;
 import org.apache.activemq.artemis.core.server.transformer.Transformer;
+import org.jboss.logging.Logger;
 
 public abstract class FederatedAbstract implements ActiveMQServerBasePlugin {
+
+   private static final Logger logger = Logger.getLogger(FederatedAbstract.class);
 
    private static final WildcardConfiguration DEFAULT_WILDCARD_CONFIGURATION = new WildcardConfiguration();
    protected final Federation federation;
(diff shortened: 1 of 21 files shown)


cshannon:
@clebertsuconic or @michaelandrepearce  - Can one of you take a look and see if it's ok to merge this?

michaelandrepearce:
@cshannon was out last week, will aim to look this week for you

cshannon:
@michaelandrepearce - Sounds good, thanks

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumer.java
```diff
@@ -163,11 +181,29 @@ private void disconnect() throws ActiveMQException {
    @Override
    public void onMessage(ClientMessage clientMessage) {
       try {
+         if (server.hasBrokerFederationPlugins()) {
+            try {
+               server.callBrokerFederationPlugins(plugin -> plugin.beforeFederatedQueueConsumerMessageHandled(this, clientMessage));
+            } catch (ActiveMQException t) {
+               logger.warn("Error executing beforeFederatedQueueConsumerMessageHandled plugin method: {}", t.getMessage(), t);
```
michaelandrepearce: should be a logger with code, e.g. explicit method logger with code as at warn (same for all logger.warns in this PR)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederationStream.java
```diff
@@ -106,4 +107,25 @@ public int getPriorityAdjustment() {
       return config.getConnectionConfiguration().getPriorityAdjustment();
    }
 
+   protected void callFederationStreamStartedPlugins() {
+      if (server.hasBrokerFederationPlugins()) {
+         try {
+            server.callBrokerFederationPlugins(plugin -> plugin.federationStreamStarted(this));
```
michaelandrepearce: If we are going to pass this to the plugin, we should extract interface, and method of federationStreamStarted should take the interface, not the implementation. As plugin should just have api's, else will cause issues in any future refactor (like you had done in last pr)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedAbstract.java
```diff
@@ -104,9 +109,27 @@ public synchronized void createRemoteConsumer(FederatedConsumerKey key, Transfor
...
+               }
+            }
             remoteQueueConsumer = new FederatedQueueConsumer(federation, server, transformer, key, upstream, callback);
             remoteQueueConsumer.start();
             remoteQueueConsumers.put(key, remoteQueueConsumer);
+
+            if (server.hasBrokerFederationPlugins()) {
+               try {
+                  final FederatedQueueConsumer finalConsumer = remoteQueueConsumer;
+                  server.callBrokerFederationPlugins(plugin -> plugin.afterCreateFederatedQueueConsumer(finalConsumer));
```
michaelandrepearce: we should extract interface of FederationQueueConsumer, and make plugin use the interface, not concrete implementation, as this plugins will cause it to be a public api for end users, making any future refactoring impossible.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedAbstract.java
```diff
@@ -104,9 +109,27 @@ public synchronized void createRemoteConsumer(FederatedConsumerKey key, Transfor
       if (started) {
          FederatedQueueConsumer remoteQueueConsumer = remoteQueueConsumers.get(key);
          if (remoteQueueConsumer == null) {
+            if (server.hasBrokerFederationPlugins()) {
+               try {
+                  server.callBrokerFederationPlugins(plugin -> plugin.beforeCreateFederatedQueueConsumer(key));
```
michaelandrepearce: extract interface, so that plugin does not expose concrete implementation, but simply api.

michaelandrepearce:
@cshannon as you would have seen, two main bits to address here.

1) logging anything that is info, warn, error should always have a logging code, and use a specific log method.

2) As plugins make things public, then interfaces should be extracted, so plugin takes interface, not the concrete implementation. Also should only expose what's needed e.g. fields not methods maybe. Else will make any future refactoring on concrete classes, like just done in previous PR impossible.

cshannon:
@michaelandrepearce - Those changes sound good to me and I can get to it either today or Monday probably.  

As a preview, 2 more things I'm working on:

1) I also have a prototype fix I'm still testing and polishing for that other bug I had found that caused extra consumers to get created.  It's a bit complicated in terms of the problem so I will write up something in the Jira to try and describe the issue and solution well when I think it's ready for a PR.'

2) I have another prototype patch to support divert bindings instead of just queue bindings over federated addresses. I still need to write tests for it and polish and then can do the PR.

cshannon:
@michaelandrepearce - I'm hoping to get back to this tomorrow/friday and finish it up before I am out of the office for the holidays.  I had to take care of some other stuff this week that came up.

cshannon:
@michaelandrepearce - not sure if you saw or not but I updated the PR with the changes

michaelandrepearce:
Sorry away on PTO. Lgtm. Feel free to merge

cshannon:
@michaelandrepearce - sounds good, I will rebase and merge today

PR #2904, opened 2019-12-02
ARTEMIS-2564 retryMessages incorrectly removes msgs

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1996,8 +1980,7 @@ private int iterQueue(final int flushLimit,
                   }
 
                   if (filter1 == null || filter1.match(ref.getMessage())) {
-                     messageAction.actMessage(tx, ref);
-                     if (remove) {
+                     if (messageAction.actMessage(tx, ref)) {
```
brusdev: Why pageIterator loop [1] hasn't the same check?

[1] https://github.com/jbertram/activemq-artemis/blob/ARTEMIS-2564/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java#L2015

clebertsuconic: There's no need to do anything different.

We could have the same test performed over paging though.

clebertsuconic: let me handle this in this afternoon. I already changed the test to use paging... but I can't make that part of the code to be used. 

leave it with me for a few hours.

clebertsuconic: I give up... and there's no need for that..


I will merge it.. if anyone has a strong opinion of adding the test.. he's welcome to do it :)

brusdev: I created the PR #2914 to fix the removing of paged messages.

PR #2906, opened 2019-12-03
ARTEMIS-2560 Duplicate amqp messages over cluster

gaohoward:
When AMQPMessages are redistributed from one node to
another, the internal property of message is not
cleaned up and this causes a message to be routed
to a same queue more than once, causing duplicated
messages.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/AMQPMessageRedistributionTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/AMQPMessageRedistributionTest.java
@@ -1,0 +1,158 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


clebertsuconic:
AMQP is accessing a CoreMessage property... I don't like it on a first look

clebertsuconic:
That was a mistake on the import, I will amend while merging it.

gaohoward:
thx @clebertsuconic @franz1981

clebertsuconic:
@gaohoward you actually don't need that import static at all. Those classes are implemeting / extending Message. so no need to static import (apparently)

gaohoward:
@clebertsuconic I'll remove that. It was added before I moved the predict to Message. It passed the style check so it's been ignored by me.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/AMQPMessageRedistributionTest.java
```diff
@@ -0,0 +1,158 @@
...
+      Connection connection = factory.createConnection();
+      Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+      MessageConsumer consumer = session.createConsumer(session.createQueue(queue));
+      connection.start();
+
+      for (int i = 0; i < num; i++) {
+         TextMessage msg = (TextMessage) consumer.receive(5000);
+         assertNotNull(msg);
+      }
+      Message msg = consumer.receive(2000);
```
clebertsuconic: Please... do not receive(2 seconds) and assert for null

if you expect it to be null, please receiveNoWait

clebertsuconic: but I already changed it here, no need to ammend.

clebertsuconic:
@gaohoward I already merged locallt before you fixed the import.. and I am merging after I run some tests. leave it with me.

PR #2908, opened 2019-12-04
ARTEMIS-2566 race condition in concurrent dynamic queue lookup

jbertram:


artemis-jms-client/src/test/java/org/apache/activemq/artemis/jndi/JndiTest.java
```diff
@@ -0,0 +1,78 @@
...
+import javax.naming.InitialContext;
+import java.util.Hashtable;
+import java.util.UUID;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import static org.apache.activemq.artemis.jndi.ActiveMQInitialContextFactory.DYNAMIC_QUEUE_CONTEXT;
+
+public class JndiTest {
```
clebertsuconic: Would be nice if this was on integration-tests instead, as this would benefit from ActiveMQTestBase and rules on Threading.

That's a nit-pick though.. I will merge it.. just leaving here as a suggestion.

PR #2909, opened 2019-12-05
ARTEMIS-2567 restore locking semantics on server.lock when reading st…

jbertram:
…atus

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
@@ -108,7 +108,16 @@ public class FileLockNodeManager extends NodeManager {
    protected synchronized void setUpServerLockFile() throws IOException {
       super.setUpServerLockFile();
 
-      for (int i = 0; i < 3; i++) {
+      if (lockChannels[0] == null || !lockChannels[0].isOpen()) {
+         File fileLock = newFile(NodeManager.SERVER_LOCK_NAME);
+         if (!fileLock.exists()) {
+            fileLock.createNewFile();
+         }
+         RandomAccessFile randomFileLock = new RandomAccessFile(fileLock, "rw");
+         lockChannels[0] = randomFileLock.getChannel();
+      }
+
+      for (int i = 1; i < 3; i++) {
          if (lockChannels[i] != null && lockChannels[i].isOpen()) {
             continue;
          }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/NodeManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/NodeManager.java
@@ -30,7 +30,7 @@ import org.apache.activemq.artemis.utils.UUIDGenerator;
 public abstract class NodeManager implements ActiveMQComponent {
 
    protected static final byte FIRST_TIME_START = '0';
-   private static final String SERVER_LOCK_NAME = "server.lock";
+   public static final String SERVER_LOCK_NAME = "server.lock";
    private static final String ACCESS_MODE = "rw";
 
    protected final boolean replicatedBackup;


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -108,7 +108,16 @@ public synchronized void start() throws Exception {
    protected synchronized void setUpServerLockFile() throws IOException {
       super.setUpServerLockFile();
 
-      for (int i = 0; i < 3; i++) {
+      if (lockChannels[0] == null || !lockChannels[0].isOpen()) {
+         File fileLock = newFile(NodeManager.SERVER_LOCK_NAME);
+         if (!fileLock.exists()) {
+            fileLock.createNewFile();
```
franz1981: what if return false, because already exists?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -108,7 +108,16 @@ public synchronized void start() throws Exception {
...
+         }
+         RandomAccessFile randomFileLock = new RandomAccessFile(fileLock, "rw");
```
franz1981: Not important if we're in hurry, but you can use `FileChannel.open(file.toPath(), StandardOpenOption.CREATE,StandardOpenOption.WRITE, StandardOpenOption.READ)`

clebertsuconic:
is there any chance we could add a test?

it regressed previously for the lack of test.. and we are fixing it again without a test.

jbertram:
I'll try to work up some kind of synthetic test since we can't guarantee access to NFS in the test-suite.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
jbertram: If the `if` returns false then that means `lockChannels[0]` is not equal to `null` and it's open which means we don't need to create the file or get the channel. It's the same basic logic that was already in place. I didn't change the semantics on that part.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
jbertram: It's not clear to me what specific benefit that would provide.

franz1981: You save to create one instance that will reference the same file descriptor, but as I've said is a minor thing :)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
franz1981: I mean createNewFile: it return a boolean

clebertsuconic: if file already existed, we just use the previous.. that's the semantic in place

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
clebertsuconic: we would have to change it everywhere. if you feel compelled to do it, make it a separate task?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
franz1981: :+1:

jbertram:
Upon further review I don't see any straight-forward way to add a test here.

PR #2912, opened 2019-12-09
NO-JIRA Minor code improvement in QueueImpl

sebthom:
Fixed some compiler warnings (missing generic type, private class can be final) and performance related sevntu checkstyle warnings, like "All method calls are advised to move to end of logic expression." or "Variable 'xyz' can be moved inside the block at line '2,864' to restrict runtime creation.". Also removed unused private constant FLUSH_TIMEOUT.

@franz1981 Maybe you want to review this?

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -129,9 +129,9 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    protected static final int CRITICAL_CHECK_DEPAGE = 4;
 
    private static final Logger logger = Logger.getLogger(QueueImpl.class);
-   private static final AtomicIntegerFieldUpdater dispatchingUpdater = AtomicIntegerFieldUpdater.newUpdater(QueueImpl.class, "dispatching");
-   private static final AtomicLongFieldUpdater dispatchStartTimeUpdater = AtomicLongFieldUpdater.newUpdater(QueueImpl.class, "dispatchStartTime");
-   private static final AtomicLongFieldUpdater consumerRemovedTimestampUpdater = AtomicLongFieldUpdater.newUpdater(QueueImpl.class, "consumerRemovedTimestamp");
+   private static final AtomicIntegerFieldUpdater<QueueImpl> dispatchingUpdater = AtomicIntegerFieldUpdater.newUpdater(QueueImpl.class, "dispatching");
+   private static final AtomicLongFieldUpdater<QueueImpl> dispatchStartTimeUpdater = AtomicLongFieldUpdater.newUpdater(QueueImpl.class, "dispatchStartTime");
+   private static final AtomicLongFieldUpdater<QueueImpl> consumerRemovedTimestampUpdater = AtomicLongFieldUpdater.newUpdater(QueueImpl.class, "consumerRemovedTimestamp");
    private static final AtomicReferenceFieldUpdater<QueueImpl, Filter> filterUpdater = AtomicReferenceFieldUpdater.newUpdater(QueueImpl.class, Filter.class, "filter");
 
    public static final int REDISTRIBUTOR_BATCH_SIZE = 100;
@@ -147,8 +147,6 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
     * or depage executor should be hanging on
     */
    public static final int DELIVERY_TIMEOUT = 1000;
-
-   private static final int FLUSH_TIMEOUT = 10000;
 
    public static final int DEFAULT_FLUSH_LIMIT = 500;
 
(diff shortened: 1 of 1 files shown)


jbertram:
Checkstyle failed:

> [ERROR] src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java:[1605,10] (whitespace) WhitespaceAround: 'if' is not followed by whitespace.
> [ERROR] src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java:[3027] (indentation) Indentation: 'if' has incorrect indentation level 7, expected level should be 6.
> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-checkstyle-plugin:3.0.0:check (default) on project artemis-server: You have 2 Checkstyle violations. -> [Help 1]

michaelandrepearce:
Is there any perf or other benefits from this change. It could be argued this is just personal preference. If its just code pref changes id be a -0.5 for it as we could endless be making changes back and forth just on someones pref.

If there are perf improvements there should be some stats to back it up.

gemmellr:
Clearly not suited for NO-JIRA. Changes to actual code in one of the most critical parts of the broker deserve a JIRA, no matter how small they may seem.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -129,9 +129,9 @@
    protected static final int CRITICAL_CHECK_DEPAGE = 4;
 
    private static final Logger logger = Logger.getLogger(QueueImpl.class);
-   private static final AtomicIntegerFieldUpdater dispatchingUpdater = AtomicIntegerFieldUpdater.newUpdater(QueueImpl.class, "dispatching");
-   private static final AtomicLongFieldUpdater dispatchStartTimeUpdater = AtomicLongFieldUpdater.newUpdater(QueueImpl.class, "dispatchStartTime");
-   private static final AtomicLongFieldUpdater consumerRemovedTimestampUpdater = AtomicLongFieldUpdater.newUpdater(QueueImpl.class, "consumerRemovedTimestamp");
+   private static final AtomicIntegerFieldUpdater<QueueImpl> dispatchingUpdater = AtomicIntegerFieldUpdater.newUpdater(QueueImpl.class, "dispatching");
```
clebertsuconic: What's the difference here? what's the advantage?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1604,9 +1602,11 @@ public long getDurableScheduledSize() {
 
       for (String lingerSessionId : lingerSessionIds) {
          ServerSession serverSession = server.getSessionByID(lingerSessionId);
-         List<MessageReference> refs = serverSession == null ? null : serverSession.getInTxLingerMessages();
-         if (refs != null && !refs.isEmpty()) {
-            mapReturn.put(serverSession.toManagementString(), refs);
+         if (serverSession != null) {
+            List<MessageReference> refs = serverSession.getInTxLingerMessages();
```
clebertsuconic: I believe there's one semantic change on your change here.

I think it's possible that getInTxLingerMessages() may return null,
so, you changed serverSession != null, but you stopped verifying the actual result. this may cause NPE and tests breaks.

clebertsuconic: your code change may get more complex here than the intended cleanup. I would leave this one, unless you have a creative choise.

clebertsuconic:
Most of the changes here are harmless... (verifying the flag before a method result, that'a s minimal change).

However the actual code change performed here scares me, especially on a component like this... on a rare usecase (lingering TXs).. this may break 1% of the usecases when using XA and that kind of thing, and that kind of thing it's not easy to debug remotely.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
sebthom: This just solves a lot of (16) compiler warnings. When declared: "AtomicIntegerFieldUpdater is a raw type. References to generic type AtomicIntegerFieldUpdater<T> should be parameterized"
When referenced: "Type safety: The method get(Object) belongs to the raw type AtomicIntegerFieldUpdater. References to generic type AtomicIntegerFieldUpdater<T> should be parameterized"

michaelandrepearce:
> Most of the changes here are harmless... (verifying the flag before a method result, that'a s minimal change).
> 
> However the actual code change performed here scares me, especially on a component like this... on a rare usecase (lingering TXs).. this may break 1% of the usecases when using XA and that kind of thing it's not easy to debug remotely.

I really want to see some perf tests and other jvm metrics proving the benefits of this change,

sebthom:
I removed the edits where I re-orded the if statements to reduce the amount of changed code.

clebertsuconic:
@michaelandrepearce the way this is setting now, there are 0 semantic changes. just some reorganization of the code, as it's what @sebthom intended.

michaelpearce-gain:
@clebertsuconic agreed this looks alot safer change now.

PR #2913, opened 2019-12-10
ARTEMIS-2571 Remove synchronization of ActiveMQServerImpl#getSessions()

sebthom:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -294,7 +294,7 @@
 
    private FileStoreMonitor fileStoreMonitor;
 
-   private final Map<String, ServerSession> sessions = new ConcurrentHashMap<>();
+   private final ConcurrentMap<String, ServerSession> sessions = new ConcurrentHashMap<>();
```
michaelandrepearce: Can a perf test or some other benefit for this change be provided. This is core parts of the broker where historically concurrent issues introduced by such small changes have had tragic effects

sebthom: @michaelandrepearce I tried to explain the reasoning for the change as part of the Jira issue, have you read that already?

michaelandrepearce: So our concerns are this is in core. Its an un-proven optimization, that without any proven hot path benefit brings alot of risk, as small changes like this have previously introduced unexpected issues at first not obvious

sebthom: ATM I am not really sure how I can provide a perf test for the admin UI. The issue is that getSessions() can be very 

 The getSessions() methods are only used by the ActiveMQServerControlImpl and the ConnectionView class, so I would think that this particular change cannot negatively affect the hot path.

ATM I am not really sure how I can provide a perf test for the effect in the webconsole.

The getSessions() methods are only used by the ActiveMQServerControlImpl and the ConnectionView class to list session attributes or to close a session manually, so I can't see how removing the synchronized keyword from these methods can affect the performance of the hot path negatively.

Also if this synchronization really is required than something else must already be broken here since ActiveMQServerImpl.removeSession(), ActiveMQServerImpl.createSession() and ActiveMQServerImpl.getSessionByID() access/modify the sessions map without synchronizing on `this`.

So even without a perf test I would argue something should be changed, either createSession/getSessionByID/removeSession also need to synchronize on `this` or the `synchronized` keyword on `getSessions()` methods is useless.

sebthom: @michaelandrepearce I attached a testcase to the JIRA issue where I try to highlight the impact of this PR. I hope this satisfies your request.

michaelpearce-gain: So idea is to include test cases etc in PRs so they live in the repo

clebertsuconic:
I'm not sure where we would add performance tests as part of the testsuite. I have ran those outside in the past using external tools.

perhaps an example?

franz1981:
@clebertsuconic I strongly suggest to use proper tools to get reliable results ie JMH
I think https://github.com/apache/qpid-proton-j/tree/master/tests/performance-jmh could be a good way to start having some of them. On Netty there is so etching similar and JMH is not just for micro-benchmarks, but is ok for macro-benchmarks as well. Next week I can help to provide some useful examples on that, if you agree :)

clebertsuconic:
@franz1981 my point was more on a place holder for perf tests as part of the testsuite. it's a general subject I think

jbertram:
I think this change looks good. Even if there was no performance testing (which there is on the JIRA) static analysis is enough for me given that (as @sebthom already pointed out) there is a lot of unsynchronized access to the `sessions` map and the map is *already* a `ConcurrentHashMap`. I don't see any good reason for those `getSessions` methods to use `synchronized`.

PR #2915, opened 2019-12-11
ARTEMIS-2573 Removing tcpnative dependency from distribution

clebertsuconic:


clebertsuconic:
this dependency is actually moot, as @gemmellr pointed me, this is not used by the server, so the binary distribution never included it.

gemmellr:
but its still worth removing the config since its unused :)

clebertsuconic:
@gemmellr yeah.. I was just being proactive and explaining why I'm removing, before someone complained.

PR #2917, opened 2019-12-11
ARTEMIS-2574 allow security manager config via XML

jbertram:
The test-suite has long used the broker's ability to configure the
security manager. This commit implements this functionality via XML
configuration.


clebertsuconic:
nit-pick: there's a script to run all the tests as part of the distribution:

./artemis-distribution/src/test/scripts/run-standard-examples.sh
./artemis-distribution/src/test/scripts/run-examples.sh


These can be used on a CI to validate the examples, or before release

Other than that I have nothing to add. if you could ammend these two on  your PR please?

clebertsuconic:
The reason these scripts exist is because it's currently not possible to run all tests with mvn -Prelease example on the root... As there is some sort of leak that makes impossible.

jbertram:
@clebertsuconic, I always forget those. :)

I just pushed the change.

PR #2918, opened 2019-12-16
NO-JIRA minor changes on test

wy96f:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionImpl.java
```diff
@@ -801,7 +801,6 @@ private void createNewRecord(final long eventUID,
       targetLocator.setInitialConnectAttempts(0);
       targetLocator.setClientFailureCheckPeriod(clientFailureCheckPeriod);
       targetLocator.setConnectionTTL(connectionTTL);
-      targetLocator.setInitialConnectAttempts(0);
```
clebertsuconic: what is the outcome of this? faster ? failure? or just to be correct?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/AddressControlTest.java
```diff
@@ -282,7 +281,7 @@ public void testGetNumberOfBytesPerPage() throws Exception {
       session.createQueue(address, address, true);
 
       AddressControl addressControl = createManagementControl(address);
-      Assert.assertEquals(ActiveMQDefaultConfiguration.getDefaultJournalFileSize(), addressControl.getNumberOfBytesPerPage());
```
clebertsuconic: my previous question also applies here.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionImpl.java
wy96f: `setInitialConnectAttempts` has been set to 0 two lines before, no need to set it again.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/AddressControlTest.java
wy96f: It's not semantically right to compare `pageSizeBytes` with default journal file size. The default page size should be AddressSettings.DEFAULT_PAGE_SIZE, see https://github.com/apache/activemq-artemis/blob/eb3c4931c2871a58fea9208eebaf90bc8ffbe462/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java#L514

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionImpl.java
clebertsuconic: OIC

PR #2920, opened 2019-12-16
ARTEMIS-2574 fix example

jbertram:


clebertsuconic:
I think you should just remove bootstrap.xml from the example, as the CLI will create it again during the CLI.


Do you have a --security-manager option to the CLI create, on which case you would be able to pass in the config through CLI instead?

clebertsuconic:
never mind actually.. it works this way

PR #2921, opened 2019-12-17
ARTEMIS-2578 clarify storage capacity messages

jbertram:
This is a surprisingly large change just to fix some log messages, but
the changes were necessary in order to get the relevant data to where it
was being logged. The fact that the data wasn't readily available is
probably why it wasn't logged in the first place.


clebertsuconic:
nice one...


I didn't include it because it was a simple version.

nice enhancement

clebertsuconic:
I can't merge it myself now.. but I aprove the changes. if no one merges it before me, I will do it later today.

PR #2927, opened 2020-01-06
ARTEMIS-2588 redundant lock on server.lock file

jbertram:


gemmellr:
Nitpicking: the JIRA indicates the issue was caused by earlier changes for https://issues.apache.org/jira/browse/ARTEMIS-2567 which haven't yet been released and seem to be largely undone by this PR. Given that, I think the commit should be referencing the original JIRA as well (or perhaps even instead, since no release will actually have the problem from 2588).

I'll leave actual review for folks familiar with the code :)

PR #2930, opened 2020-01-08
NO-JIRA regenerate expired SSL test resources

jbertram:


clebertsuconic:
how did you do it? perhaps a little note somewhere on how to generate / regenerate them?

clebertsuconic:
@jbertram  ^^

jbertram:
Instructions for generating the resources are in [the test that requires them](https://github.com/apache/activemq-artemis/blob/master/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/imported/MQTTSecurityCRLTest.java#L41) and in [the example documentation](https://github.com/apache/activemq-artemis/blob/master/examples/features/standard/ssl-enabled-crl-mqtt/readme.md) as well.

PR #2936, opened 2020-01-13
NO-JIRA Adding test name into temporary folder name directory

clebertsuconic:
This is to make it possible to identify what test is leaking files whenever that is happening.
That is because future tests will report the leaks, and it's difficult to identify where it happened.

Also i'm changing NoProcessFilesBehind to show the getOpenFD propertly


clebertsuconic:
These files will be removed by another rule in ActiveMQTestBase:

https://github.com/apache/activemq-artemis/blob/3fff744de4d0ba3b209c821cc1404d920455157c/artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java#L233-L236


I had a case where I had leaks on a change I made, and I needed this to find the issue.

I wanted to keep this a permanent change as it will be helpful in the future.

jbertram:
Nice one!

PR #2938, opened 2020-01-14
ARTEMIS-2596 (kill -9) AMQ causes tmp web dir space usage to increase

gaohoward:
If you kill the server without invoking a normal shutdown, tmp
web files are not cleaned out.  This leaves old webapp folders
lingering until a normal shutdown.

In a failover test environment that repeatedly kills the server,
this causes disk space usage issues.

The fix is to add a cleanup method before the web server starts.
It searches the tmp web dir if there is any leftover files/dirs
and delete them if any.

### DIFF
--- a/artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
+++ b/artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
@@ -46,6 +46,7 @@ import org.eclipse.jetty.server.handler.RequestLogHandler;
 import org.eclipse.jetty.server.handler.ResourceHandler;
 import org.eclipse.jetty.util.ssl.SslContextFactory;
 import org.eclipse.jetty.webapp.WebAppContext;
+import org.eclipse.jetty.webapp.WebInfConfiguration;
 import org.jboss.logging.Logger;
 
 public class WebServerComponent implements ExternalComponent {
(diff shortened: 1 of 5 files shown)


artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
```diff
@@ -226,6 +227,7 @@ public void start() throws Exception {
       if (isStarted()) {
          return;
       }
+      cleanupTemporaryFilesIfAnyBeforeStart();
```
clebertsuconic: Can you just rename this to cleanupTmp(); ?

clebertsuconic:
There's a binary as part of this PR? is that a mistake?

if not? you shoudln't commit binaries without their source code.

artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
clebertsuconic: I don't understand one thing.. can't you reuse the WebTmpCleaner here?

gaohoward: It eventually calls WebTmpCleaner but before that we need to figure out where the tmp dir is. (because before the web server start there is no tmpdir in each web context. The tmpdir only
exists in the web context after the webapp (war) is deployed).

gaohoward:
I'll remove the binary war (copied from artemis-hawtio) and add a new war for the test.

jbertram:
@clebertsuconic, can you take a look at @gaohoward's update?

PR #2939, opened 2020-01-15
ARTEMIS-2597 Memory Leak when closing AMQP Consumers in the context

brusdev:
Remove server senders on remote link close.


clebertsuconic:
Did you run a full testsuite on this? at least on AMQP?

As long as you ran the tests.. LGTM

PR #2941, opened 2020-01-15
NO-JIRA: fix .asf.yaml invalid GitHub labels

jdanekrh:
```
An error occurred while running github feature in .asf.yaml!:
.asf.yaml: Invalid GitHub label 'Apache' - must be lowercase alphanumerical and <= 35 characters!
```

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -18,18 +18,19 @@ github:
   description: "Mirror of Apache ActiveMQ Artemis"
   homepage: https://activemq.apache.org/components/artemis
   labels:
-    - Apache
-    - ActiveMQ
+    - apache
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -18,18 +18,19 @@ github:
   description: "Mirror of Apache ActiveMQ Artemis"
   homepage: https://activemq.apache.org/components/artemis
   labels:
-    - Apache
-    - ActiveMQ
+    - apache
+    - activemq
+    - activemq-artemis
```
jdanekrh: I checked that there are repos with a minus in the label, https://github.com/topics/activemq-artemis

PR #2942, opened 2020-01-16
ARTEMIS-2599 DescribeJournal isn't correctly counting surviving msg

franz1981:
This PR:

- enhance performance of the original version by saving some boxed `Integer` allocation
- enhance the info provides: it shows large messages count
- fix the non large message count


clebertsuconic:
the integer boxing.. I don't think it was a big deal as this is a reporting tool only. not really intended for high performance. although it was written that way just because I first did it to solve / debug an issue, commited on the repo and it's there until today.

Now, if there are issues on the counter, it warrants a JIRA.

clebertsuconic:
@franz1981 can you open a JIRA for this one and amend your commit?. this one would be nice on the release notes.

franz1981:
@clebertsuconic done!

PR #2944, opened 2020-01-17
ARTEMIS-2602 Improve Journal loading heap usage

franz1981:


franz1981:
I don't know why I cannot require the review from @wy96f, but I would like to ask for it :)

franz1981:
@clebertsuconic @michaelandrepearce `LongHashSet` is a porting of https://github.com/real-logic/agrona/blob/master/agrona/src/main/java/org/agrona/collections/IntHashSet.java (and the tests too). Please recommend me the better way to deal this this :)

michaelpearce-gain:
@franz1981 i would add the lib as a library as its avaiable as a fair small jar in maven with not much dependency, the main reasons we didnt do that with the bookeeper one was it wasn;t nicely seperated code, also we had to make some fixes

franz1981:
@michaelpearce-gain the fact is that `LongHashMap` doesn't exists on Agrona: I've made it by myself...

michaelandrepearce:
ah ok, gotcha i missed its not a copy, but a modifcation to use long. tbh lgtm then. as long as theres maybe a small comment in the code giving reference to original source of the idea, e.g. give credit due

michaelandrepearce:
btw do you have any nice pretty graphs to go with this PR showing the before and after on the heap......

franz1981:
@michaelandrepearce 
I'm producing a journal to show the improvement :)
I like the "nice pretty part" :)

clebertsuconic:
LGTM

clebertsuconic:
It wasn't clear it wasn't ready to me.. .please send another commit with further improvments

PR #2946, opened 2020-01-17
ARTEMIS-2603 Deadlock between pageStore.getCurrentID and page.cleanup

clebertsuconic:
This happened when JournalStorageManager.startReplication and PageStore.cleanup were called between multiple threads.


clebertsuconic:
full test suite pass on this commit, I'm merging this myself.

PR #2947, opened 2020-01-17
ARTEMIS-2602 Reduce number of copies for non JDBC Journal

franz1981:


clebertsuconic:
Testsuite is completely crazy here. some stack overflows (load method calling load).

it can't be merged.

franz1981:
Will take a look at this on Monday! That's strange, I have just saved a copy of the sparse array list while loading the journal...maybe there is something going on with the previous implementation of  the sparse array list's remove; that now has always the chance to be used

franz1981:
@clebertsuconic I've made an invifnite recursive call, now should be fixed, I'm kickin' the CI :+1:

franz1981:
@clebertsuconic I've fixed the NPE too, now running some tests :)

clebertsuconic:
I don't know what did you do, but you just cut 10 min from the testsuite run.

Is this ready to merge?

franz1981:
@clebertsuconic eheh eheh is not limited by the speed journal now we are faster to load :)
And can be made faster too

franz1981:
@wy96f this should improve anything related to loading the journal, especially with many small messages and a proper heap size and GC configuration

clebertsuconic:
testsuite is 100% pass, from 2 hours and 25 minutes, it ran 10 min faster on both runs.

if I could review this tomorrow, just to make sure. I want to review it.

franz1981:
@clebertsuconic Thanks bud!

> from 2 hours and 25 minutes, it ran 10 min faster on both runs

Yep, the previous changes on loading already improved by ~10 min on test suite and this new ones will reduce it by another couple of minutes too :)

franz1981:
Probably will create some follow-up PRs to fix another issue: anothe big load of garbage is due to `Long` boxing for all the `Map<Long, ...>` we use while loading the journal...Probably worths to save those as well

PR #2948, opened 2020-01-20
Fixing Postgresql JDBC store when the driver is not directly accessible or when using a connection pool.

ehsavoie:
The postgresql dedicated code in Artemis suppose that the connection is a 'direct' connection from the driver. This may not always be the case and thus the connection needs to be unwrapped and if a different classloader is used then use reflection to use it.


clebertsuconic:
I - JIRA
II - test
III - explain?

I know it's not possible to simulate Wildfly within our testsuite. (well.. perhaps it is, I don't know if it's feasible)


But at least it would be possible to add a test on the new class you added, so we wouldn't break the integration you're working on now.

ehsavoie:
Sorry this is just a WIP

clebertsuconic:
ah.. ok.. thanks

ehsavoie:
Superseded by https://github.com/apache/activemq-artemis/pull/2987

PR #2950, opened 2020-01-21
ARTEMIS-2604 Optimizations on journal loading

franz1981:
I've opened this for discussion, so PLEASE DON'T MERGE IT YET.

I've yet to implement the same optimization for other properties (ie `getDuplicateProperty` and `getLastValueProperty`) and for the `AMQPMessage`, while providing adeguate test coverage.

I will need some help from @gemmellr and @tabish121 , because I see that implementing an `AMQPMessage::searchProperty(SimpleString key)` that won't force encoding (if not already happened) of the whole message isn't trivial due to my ignorance of the protocol and probably the wrong approach to the problem.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/TypedProperties.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/TypedProperties.java
@@ -372,6 +372,79 @@ public class TypedProperties {
       }
    }
 
+   /**
+    * Performs a search among the valid key properties contained in {@code buffer}, starting from {@code from}
+    * assuming it to be a valid encoded {@link TypedProperties} content.
+    *
+    * @throws IllegalStateException if any not-valid property is found while searching the {@code key} property
+    */
+   public static boolean searchProperty(SimpleString key, ByteBuf buffer, int startIndex) {
+      // It won't implement a straight linear search for key
+      // because it would risk to find a SimpleString encoded property value
+      // equals to the key we're searching for!
+      int index = startIndex;
+      byte b = buffer.getByte(index);
+      index++;
+      if (b == DataConstants.NULL) {
+         return false;
+      }
+      final int numHeaders = buffer.getInt(index);
+      index += Integer.BYTES;
+      for (int i = 0; i < numHeaders; i++) {
+         final int keyLength = buffer.getInt(index);
+         index += Integer.BYTES;
+         if (key.equals(buffer, index, keyLength)) {
+            return true;
+         }
+         if (i == numHeaders - 1) {
+            return false;
+         }
+         index += keyLength;
+         byte type = buffer.getByte(index);
+         index++;
+         switch (type) {
+            case NULL: {
+               break;
+            }
+            case CHAR:
+            case SHORT: {
+               index += Short.BYTES;
+               break;
+            }
+            case BOOLEAN:
+            case BYTE: {
+               index += Byte.BYTES;
+               break;
+            }
+            case BYTES:
+            case STRING: {
+               index += (Integer.BYTES + buffer.getInt(index));
+               break;
+            }
+            case INT: {
+               index += Integer.BYTES;
+               break;
+            }
+            case LONG: {
+               index += Long.BYTES;
+               break;
+            }
+            case FLOAT: {
+               index += Float.BYTES;
+               break;
+            }
+            case DOUBLE: {
+               index += Double.BYTES;
+               break;
+            }
+            default: {
+               throw ActiveMQUtilBundle.BUNDLE.invalidType(type);
+            }
+         }
+      }
+      return false;
+   }
+
    public synchronized void decode(final ByteBuf buffer,
                                    final TypedPropertiesDecoderPools keyValuePools) {
       byte b = buffer.readByte();
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/CompositeAddress.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/CompositeAddress.java
@@ -35,7 +35,15 @@ public class CompositeAddress {
    }
 
    public static SimpleString extractQueueName(SimpleString name) {
-      return name == null ? null : new SimpleString(extractQueueName(name.toString()));
+      if (name == null) {
+         return null;
+      }
+      final String nameString = name.toString();
+      final String queueName = extractQueueName(nameString);
+      if (queueName.equals(nameString)) {
+         return name;
+      }
+      return new SimpleString(queueName);
    }
 
    public static String extractQueueName(String queue) {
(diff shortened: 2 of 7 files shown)


franz1981:
@clebertsuconic Sadly seems that while exchanging messages, we trigger the full decode anyway, due to other properties accessed eg `CoreMessage:.getRoutingType`
image

But, good news about the change already done:
that's from *ARTEMIS 2.11*
image

And the same heap configuration ie `-XX:+UseG1GC -Xms4G -Xmx4G -XX:+AlwaysPreTouch` for *this PR*:
image

The difference in pause time is just too much to compare them :)
There are other metrics, but the point is that we won't produce tons of garbage to decode messages if the `ScheduledDeliveryTime` property isn't set and should cover the most of the cases.

The previous results has been performed by loading a journal obtained by running 30 pairs of Core JMS producers/consumers sending persistent byte messages of 10 bytes size until the journal start to page, with a slow consumer (that perform Thread::sleep(200) while consuming messages): this has filled the message journal with a mix of unacked/acked small messages that would stress the heap/GC part of the loading activity instead of the disk. 
Indeed the whole `data`  folder is just `2.5 GB`.
I would try with different sizes too, to put under stress the new search property method, that is being benchmarked separately to make it as fast as possible.

I strongly encourage any user that make heavy use of the journal and has stress tests while loading journal to try this, being aware that ATM it just improve this for Core protocol, but with some help, will get the same for AMQP as well.

franz1981:
@clebertsuconic The downside of this optimization is that it still need to read the whole part of buffer with the properties in the best case: the ideal thing would be to always place `ScheduledDeliveryTime` in a fixed position (as first property?)

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/TypedProperties.java
```diff
@@ -372,6 +372,79 @@ private synchronized void forEachInternal(BiConsumer<SimpleString, PropertyValue
       }
    }
 
+   /**
+    * Performs a search among the valid key properties contained in {@code buffer}, starting from {@code from}
+    * assuming it to be a valid encoded {@link TypedProperties} content.
+    *
+    * @throws IllegalStateException if any not-valid property is found while searching the {@code key} property
+    */
+   public static boolean searchProperty(SimpleString key, ByteBuf buffer, int startIndex) {
```
clebertsuconic: wouldn't be less costly to just lookup for the sequence of bytes?

You're still parsing the key at its exact position.

if just looking for the sequence , say _HDR_SQUEDULED_TIME anywhere in the buffer would be enough.

now if the search is more costly than the partial parsing, then it's fine.

At least you're not reading the string to verify the equal object.


if this was C, you would control to reuse the same integer every time, but in Java every time you do buffer.getInt(index) you will create a few bytes in memory.

clebertsuconic:
"the ideal thing would be to always place ScheduledDeliveryTime"

the ideal thing would be to use a fixed value within the protocol, like what AMQP guys are doing. it would pretty much be redefining the protocol, it would incur in versioning anyways.

Using a fixed property would in fact require you to do versioning on the protocol, to make sure you still parse the whole buffer for older clients.. so it's not feasible without breaking compatibility, and if you break it, it would be best to do the right thing.


I would say it's a no. .since AMQP takes care of it. we could instead invest to make AMQP faster than core.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/TypedProperties.java
franz1981: I have added a comment about this and the risk is to find a value that matches the property we are searching for. The other reason is that we know how many bytes to skip by using the decoded info and would help to reach quicker what we need without comparing every single byte (including values)

clebertsuconic: if you could name this as hasProperty(SimpleString key).

since you use the parsed property if the buffer is not parsed already.

franz1981:
To make amqp to be faster then core we should start dropping boxed values as viable decoded types and just use primitives: that would help a lot

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/TypedProperties.java
franz1981: I will change it for sure :+1

clebertsuconic:
>To make amqp to be faster then core we should start dropping boxed values as viable decoded types and just use primitives: that would help a lot

That requires a new codec, I think it's on the way.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/TypedProperties.java
clebertsuconic: >> "the risk is to find a value that matches the property we are searching for"

So what? if someone uses the key as the value, all it will incur is for us to do the older semantic, which is to parse the whole buffer. and we would incur in the old semantic.


Add a test with the value as the key, just to make sure.. although I doubt someone will use the property _HDR_SCHEDULED_TIME anywhere in the values :) and if they did it would still be fine.

clebertsuconic: what I mean is.. if parsing the whole buffer is faster.. just go for it.. don't worry about values as keys.

franz1981: But is not faster according my tests...in one case we perform more comparisons...

franz1981: Re your comments

> You're still parsing the key at its exact position.

Nope: to compare `SimpleString` and `ByteBuf` is using another optimized method I've implemented some time ago that allow to compare them very quickly directly ie https://github.com/apache/activemq-artemis/blob/0293d8057441e6bfeb9e6533774862ffc86d74eb/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java#L345 

> in Java every time you do buffer.getInt(index) you will create a few bytes in memory.

That's the signature of the Netty method: https://netty.io/4.0/api/io/netty/buffer/ByteBuf.html#getInt-int-

it would return a primitive `int`: it's not allocating anything. I've taken care to design that method to be garbage-free :+1:

michaelandrepearce:
just adding the DNMY label, as you said this isn't to be merged in first comment, just so its clearer when looking over all PR's. (I'm playing catchup from a nice long holiday :) )

franz1981:
@michaelandrepearce I've yet to add some tests, change some name and cleanup some APIs by addressing some of the advices from @clebertsuconic , but feel free to take a look/review so I will address altogether:  from the pov of journal loading (and ha shared store) I see that when the disk is fast enough this could be really a game changer together with the other optimizations already merged on master, if compared to 2.11 :)

franz1981:
@michaelandrepearce @clebertsuconic I've added some tests and this is ready to get a full review. 
After addressing all the review comments I'll squash the changes into one. The CI is already in a good shape on this :+1:

franz1981:
@michaelandrepearce @wy96f I finally got some numbers about the the loading of journal:

this commit:
start: 19:28:08,193
end: 19:28:28,391

~20 seconds

2.11:
start: 12:12:22,651
end: 12:13:15,625

~ 52 seconds

Is about twice as faster assuming a very fast disk :)

clebertsuconic:
I'm merging this

PR #2951, opened 2020-01-23
Artemis-2563 Add option to create queue on all active clustered nodes

AntonRoskvist:
This is to help with redistribution as mentioned in this Jira:
https://issues.apache.org/jira/browse/ARTEMIS-2563

Address option added to enable queue creation on all active nodes in a cluster, which happens on BINDING_ADDED. Since this creates a local binding for remote queues, a redistributor will also get created once a consumer registers

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -278,12 +278,12 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
 
                boolean clusteredQueues = addressSettings.getClusteredQueues();
 
-               if (clusteredQueues && binding == null) {
+               if (clusteredQueues && binding == null && distance != 0) {
                   try {
                      if (routingName.equals(address)) {
                         server.createQueue(address, RoutingType.ANYCAST, address, filterString, true, false);
                      } else {
-                        server.createQueue(address, RoutingType.MULTICAST, address, filterString, true, false);
+                        server.createQueue(address, RoutingType.MULTICAST, routingName, filterString, true, false);
                      }
                   } catch (Exception e) {
                      logger.error("Could not create queue", e);
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -272,6 +272,19 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
 
                queueInfos.put(clusterName, info);
 
+               //Global queue creation fix
+               Binding binding = getBinding(routingName);
+
+               if (true && binding == null) {
+                  if(routingName.equals(address)) {
+                     try {
+                        server.createQueue(address, RoutingType.ANYCAST, address, filterString, true, false);
+                     } catch (Exception e) {
+                        logger.error("Could not create queue", e);
+                     }
+                  }
+               }
+
                break;
             }
             case BINDING_REMOVED: {
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -94,8 +94,9 @@ import org.apache.activemq.artemis.utils.collections.TypedProperties;
 import org.jboss.logging.Logger;
 
 /**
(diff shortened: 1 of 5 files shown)


clebertsuconic:
- I have not seen any tests added on your PR
- we should have a discussion about the feature itself.. as it may be extremely dangerous. I would prefer having a single subscription on a single node, and clients connecting to that node.

- The PR as it stands it can't be merged.. lack of tests.. you actually are breaking compatibility on the compatility testsuite... (the PR is probably failing because of that).

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -1192,6 +1212,8 @@ public void encode(ActiveMQBuffer buffer) {
       BufferHelper.writeNullableBoolean(buffer, defaultLastValueQueue);
 
       BufferHelper.writeNullableLong(buffer, redistributionDelay);
+      
+      BufferHelper.writeNullableBoolean(buffer, clusteredQueues);
```
michaelandrepearce: this cannot be added here, new bits in the journal must ALWAYS be at the end for compatibility reasons

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -984,6 +1001,8 @@ public void decode(ActiveMQBuffer buffer, boolean tryCompatible) {
       defaultLastValueQueue = BufferHelper.readNullableBoolean(buffer);
 
       redistributionDelay = BufferHelper.readNullableLong(buffer);
+      
+      clusteredQueues = BufferHelper.readNullableBoolean(buffer);
```
michaelandrepearce: must be at the end of the serialisation for compatibility reasons

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -3276,6 +3276,15 @@
                   </xsd:documentation>
                </xsd:annotation>
             </xsd:element>
+            
+            <xsd:element name="clustered-queues" type="xsd:boolean" default="false" maxOccurs="1" minOccurs="0">
```
michaelandrepearce: please add also the the schema in tests, adding tests also around the config.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -272,6 +272,23 @@ public void onNotification(final Notification notification) {
 
                queueInfos.put(clusterName, info);
 
+               //Global queue creation fix
+               Binding binding = getBinding(routingName);
+               
+               AddressSettings addressSettings = addressSettingsRepository.getMatch(address.toString());
+
+               boolean clusteredQueues = addressSettings.getClusteredQueues();
```
michaelandrepearce: should this not be a value that's configurable at the queue level, e.g. under an address you may want one queue not clustered and another clustered. it shouldn't be an all or nothing for all queues, e.g. address setting should be a default, but the actual value should be at the queue level

docs/user-manual/en/address-model.md
```diff
@@ -720,6 +721,12 @@ before it will begin to dispatch messages. Default is `-1` (wait forever).
 closed on a queue before redistributing any messages. Read more about
 [clusters](clusters.md#message-redistribution).
 
+`clustered-queues` enables cluster-wide creation of `ANYCAST` queues.
```
michaelandrepearce: why only anycast, surely queue type should not matter, and use cases for mutlicast would exist, e.g. share durable subscriptions

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -272,6 +272,23 @@ public void onNotification(final Notification notification) {
...
+
+               if (clusteredQueues && binding == null) {
+                  if(routingName.equals(address)) {
+                     try {
+                        server.createQueue(address, RoutingType.ANYCAST, address, filterString, true, false);
```
michaelandrepearce: need to support multicast also.

AntonRoskvist:
Thanks for all of your feedback! I will start working with fixing what you have pointed out. Should I close this PR or can I apply my fixes here somehow?

@clebertsuconic You said 
"we should have a discussion about the feature itself.. as it may be extremely dangerous. I would prefer having a single subscription on a single node, and clients connecting to that node."

I do not really understand that last part, how would such an implementation work?

Br,
Anton

AntonRoskvist:
I will close this for now, got some issues rebasing and formatting. I will open a new one later

PR #2952, opened 2020-01-24
NO-JIRA improve role-access match priority doc

brusdev:


gaohoward:
Hi @brusdev Do we have a test to verify this logic? I think a test should be put in there.

brusdev:
Hi @gaohoward I added the test `testBasicRoleWithWildcardInKey` with the PR #2851, but I could add more tests if needed.

gaohoward:
That's great @brusdev. I think that's ok as long as we have a test. I'll go merge it.
thx.

PR #2954, opened 2020-01-25
ARTEMIS-2608 Fix ClassCastException on binary properties conversion

fvaleri:


jbertram:
Can you add a test to this to confirm the fix?

fvaleri:
Hi @jbertram , for some reason I forgot to commit the test, now you have it.

You can run it with:
`mvn -Ptests -DfailIfNoTests=false -Dtest=OpenWireMessageConverterTest test -pl artemis-protocols/artemis-openwire-protocol`

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -610,7 +613,7 @@ private static ActiveMQMessage toAMQMessage(MessageReference reference,
          amqMsg.setCorrelationId(corrId.toString());
       }
 
-      final byte[] dsBytes = (byte[]) coreMessage.getObjectProperty(AMQ_MSG_DATASTRUCTURE);
+      final byte[] dsBytes = getBytesPropertyOrNull(coreMessage, AMQ_MSG_DATASTRUCTURE);
```
clebertsuconic: This would been either a mistake from the user, or someone maliciously trying to break things. 

shouldn't you log such condition instead of silently failing?

or perhaps I would prefer rejecting the message and sending an error to the client, meaning the previous semantic was probably ok?

fvaleri: I thought about this, but this actually came out from a legitimate use case as described in the Jira:
1. Send a message to the broker using OpenWire client
2. Consume that message using AMQP client
3. Send that message back to the broker using AMQP client

Considering the previous use case, at which level you would log?

clebertsuconic: Definitely log.warn

clebertsuconic: >>Send a message to the broker using OpenWire client

Perhaps we could fix whatever conversion made that happen?

fvaleri: > > > Send a message to the broker using OpenWire client
> 
> Perhaps we could fix whatever conversion made that happen?

Agree. I think the problem is located in the AMQP module but it needs more investigation. Can we create another Jira for that and release this fix?

fvaleri:
Hi @jbertram @clebertsuconic , what about merging this?

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -963,4 +966,13 @@ private static void setAMQMsgObjectProperties(final ActiveMQMessage amqMsg,
          }
       }
    }
+
+   private static byte[] getBytesPropertyOrNull(ICoreMessage coreMessage, SimpleString property) {
+      try {
+         return coreMessage.getBytesProperty(property);
+      } catch (ActiveMQPropertyConversionException e) {
+         LOG.warnf("%s", e.getMessage());
```
michaelandrepearce: should be a explicit and coded log entry.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -963,4 +966,13 @@ private static void setAMQMsgObjectProperties(final ActiveMQMessage amqMsg,
          }
       }
    }
+
+   private static byte[] getBytesPropertyOrNull(ICoreMessage coreMessage, SimpleString property) {
+      try {
+         return coreMessage.getBytesProperty(property);
```
michaelandrepearce: this shouldn't be catching property conversion exception, instead it should check if coremessage has property and return null, should still be allowing ActiveMQPropertyConversionException to propagate up if there's an actual conversion issue.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
michaelandrepearce: > Hi @jbertram @clebertsuconic , what about merging this?

reading others reviews, isn't this waiting of finding the underlying cause why the properties go missing in AMQP? we should be fixing underlying issue, not plastering over.

michaelandrepearce:
> Hi @jbertram @clebertsuconic , what about merging this?

reading others reviews, isn't this waiting of finding the underlying cause why the properties go missing in AMQP? we should be fixing underlying issue, not plastering over.

fvaleri:
@michaelandrepearce the problem is within the AMQP client. If you create a simple AMQP message resender, then you will see some invalid properties added to the original message (see ARTEMIS-2608 for the reproducer). 

```
3 = {java.util.HashMap$Node@2093} "__HDR_MESSAGE_ID" -> "\x00\x00\x00Ln\x00\x01{\x01\x00%ID:mint.local-50285-1574947728615-1:1\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00"
 key = "__HDR_MESSAGE_ID"
 value = {org.apache.qpid.proton.amqp.Binary@2101} "\x00\x00\x00Ln\x00\x01{\x01\x00%ID:mint.local-50285-1574947728615-1:1\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x00"
4 = {java.util.HashMap$Node@2094} "__HDR_PRODUCER_ID" -> "\x00\x00\x009{\x01\x00%ID:mint.local-50285-1574947728615-1:1\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01"
 key = "__HDR_PRODUCER_ID"
 value = {org.apache.qpid.proton.amqp.Binary@2103} "\x00\x00\x009{\x01\x00%ID:mint.local-50285-1574947728615-1:1\x00\x00\x00\x00\x00\x00\x00\x01\x00\x00\x00\x00\x00\x00\x00\x01"
 ```

The OpenWire consumer expects binary data for those keys and the aim of this fix is to make it more robust, while also solving a real use case problem.

**Comment by gemmellr** [comment]
I dont think the issue is within the AMQP client. The broker sends the AMQP consumer binary values in the message, so unless its mutated before resending (i.e clearing the properties and setting new values) those same binary values are going to go back to the broker. If it doesn't handle them that seems like its on it, given it created them to start with.

**Comment by fvaleri** [comment]
@gemmellr is right. The problem is in the `AmqpCoreConverter.setProperty` method which convert proton-j's Binary to String instead of byte[]:

```java
...
    } else if (value instanceof Binary) {
         msg.setStringProperty(key, value.toString());
```

Please review my last commit which fixes the issue, adds a converter test and bit of refactoring to have separate classes for CoreOpenWireConverter and OpenWireCoreConverter (much like in the AMQP protocol module).

Thanks.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -308,9 +308,7 @@ private static ServerJMSMessage processApplicationProperties(ServerJMSMessage jm
    private static ServerJMSMessage processExtraProperties(ServerJMSMessage jms, TypedProperties properties) {
       if (properties != null) {
          properties.forEach((k, v) -> {
-            if (!k.equals(AMQPMessage.ADDRESS_PROPERTY)) {
```
clebertsuconic: why?

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenwireMessage.java
```diff
@@ -30,8 +30,7 @@
 
 import io.netty.buffer.ByteBuf;
 
-// TODO: Implement this
-public class OpenwireMessage implements Message {
+public class OpenWireMessage implements Message {
```
clebertsuconic: you didn't need to change this.. I would keep your changes to a minimal.. only what you really need.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
fvaleri: my fault, let me fix this

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenwireMessage.java
fvaleri: ok

fvaleri:
Let me commit the one-line change to fix this. Then, I will create a distinct PR for proposed refactoring (same semantics).

brusdev:
How does the test check the fix?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -477,7 +477,7 @@ private static void setProperty(javax.jms.Message msg, String key, Object value)
       } else if (value instanceof Decimal32) {
          msg.setFloatProperty(key, ((Decimal32) value).floatValue());
       } else if (value instanceof Binary) {
-         msg.setStringProperty(key, value.toString());
+         msg.setObjectProperty(key, ((Binary) value).getArray());
```
tabish121: This isn't a complete fix as the Binary value may contain a subsequence of some other array value possibly a slice from another buffer owned by someone else so you need to copy this if the Binary has an array offset or relative length value that doesn't match the underlying array otherwise you end up with a value on the other side that doesn't match the original scoped array contents.

clebertsuconic:
@fvaleri the test does not fail without the fix.. I don't see how this test is validating anything.


It would be nice to understand what is the issue so we can fix the converter.

fvaleri:
@clebertsuconic @brusdev added the relevant integration test

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
fvaleri: Hi @tabish121 , thank you for this note. I fixed this, please let me know if it's fine now.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/AMQPResenderToOpenwireTest.java
```diff
@@ -0,0 +1,158 @@
...
+   @After
+   public void tearDown() throws Exception {
+      if (server != null) {
+         server.stop();
+         server = null;
+      }
+   }
+
+   @Test(timeout = 60_000)
+   public void openwireInternalBinaryPropertiesShouldBeResentAsByteArrays() throws Exception {
```
franz1981: it wouldn't been more deterministic and simpler to make it single threaded? I know that's not reproducing a client behaviour (that should span different machines anyway), but it would make the timing lot more deterministic

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/AMQPResenderToOpenwireTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+               Connection connection = factory.createConnection();
+               Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+
+               Queue queueZero = session.createQueue(queueZeroName);
+               Queue queueOne = session.createQueue(queueOneName);
+
+               MessageProducer producer = session.createProducer(queueZero);
+               Message testMessage = session.createTextMessage("test");
+               resenderReady.await(20, TimeUnit.SECONDS);
```
brusdev: Why do you need to wait for resenderReady?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/AMQPResenderToOpenwireTest.java
```diff
@@ -0,0 +1,158 @@
...
+               connection.close();
+
+            } catch (Exception e) {
+               errors.add(e);
+            }
+         }
+      });
+
+      executor.shutdown();
+      executor.awaitTermination(20, TimeUnit.SECONDS);
```
brusdev: Can you reduce this timeout?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/AMQPResenderToOpenwireTest.java
fvaleri: yes, I think it can be avoided, let me do some tests before removing it

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/AMQPResenderToOpenwireTest.java
fvaleri: I guess it can be reduced, will do after some testing

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/crossprotocol/AMQPResenderToOpenwireTest.java
fvaleri: good point, let me explore this suggestion

tabish121: @franz1981 is correct here, there is simply no need to make this a multi threaded test that requires more overhead and time vs just creating two methods that take a connection factory one the sends and one that receives and then test that messages sent from one protocol are received by another they contain the expected message body data types.  There are plenty of existing test cases in the AMQP integrations tests that do this sort of thing already

fvaleri:
@tabish121 @brusdev @clebertsuconic @franz1981 something like this would be good?

brusdev:
LGTM +1

PR #2959, opened 2020-01-29
NO-JIRA Fix Javadoc warning in ActiveMQTestBase

sebthom:
Fixes Javadoc warnings "Javadoc: Exception Exception is not declared" and "Javadoc: Exception InterruptedException is not declared"


artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -654,7 +654,6 @@ public static void forceGC() {
     * Verifies whether weak references are released after a few GCs.
     *
     * @param references
-    * @throws InterruptedException
```
clebertsuconic: thanks.. this was probably generated by some IDE.

PR #2969, opened 2020-02-03
NO-JIRA Fix compiler warning in FederationStreamConnectMessage v2

sebthom:
Second attempt to fix the following compiler warning that is reported in Travis builds, this time using the correct cast type `Class<?>[]` which prevents temporary object allocation because of var-args handling:

```java
/home/travis/build/apache/activemq-artemis/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/FederationStreamConnectMessage.java:151: warning: non-varargs call of varargs method with inexact argument type for last parameter;

         return (FederationPolicy) Class.forName(clazz).getConstructor(null).newInstance();

  cast to Class<?> for a varargs call
  cast to Class<?>[] for a non-varargs call and to suppress this warning
```


clebertsuconic:
tests are good.. merging

PR #2970, opened 2020-02-04
ARTEMIS-2615 Update netty version to 4.1.45.Final

brusdev:
Update netty version to 4.1.45.Final and netty-tcnative version to 2.0.28.Final.


clebertsuconic:
@brusdev I had to remove caches on travis

merging it.

PR #2971, opened 2020-02-04
NO-JIRA Let SparseArrayLinkedList to reuse the last empty array

franz1981:
That's an optimization on `SparseArrayLinkedList` that allows it to reuse the last array chunk when is empty (too): in the previous version the last remaining array wasn't pooled and adding a single element would need a new array allocation.

This change allows to replace some `ArrayList` usage with this list, when the API and usage allows it.


clebertsuconic:
this is a real optimization, shouldn't be a JIRA?

franz1981:
@clebertsuconic 
Given that there wasn't an original JIRA just for this data-structure I wasn't sure to do it; but if it helps tracking it, I will do it :+1:

PR #2972, opened 2020-02-05
Restart Sequence

shchavan-source:
This document contains the restart sequence for the brokers under various situations.
These sequences can help to restart the brokers in a proper sequence so as to avoid the client applications to be restarted.


franz1981:
I suggest to sign the apache ICLA (see https://www.apache.org/dev/new-committers-guide.html#submitting-your-individual-contributor-license-agreement-icla) and create an appropriate JIRA for this document change eg https://issues.apache.org/jira/browse/ARTEMIS-1730
The name of PR and the commit message should reflect the JIRA name to help tracking the motivation of the change :+1:

clebertsuconic:
nice doc... it needs a JIRA

clebertsuconic:
nice doc... it needs a JIRA

jbertram:
IMO this belongs in the "High Availability" chapter rather than its own chapter.

michaelpearce-gain:
@jbertram if no objections in next few hours, i will merge this, we can then move it later if wanted. I think the doc is useful

michaelpearce-gain:
@clebertsuconic it seems this went quiet, but i think its a good contribution, to avoid issues with no Jira i have created one for it: https://issues.apache.org/jira/browse/ARTEMIS-2822

PR #2973, opened 2020-02-06
NO-JIRA clarify & verify web socket support for MQTT

jbertram:


docs/user-manual/en/amqp.md
```diff
@@ -155,4 +155,20 @@ This contains a real example for configuring amqpIdleTimeout:
...
\ No newline at end of file
+```
+
+## Web Sockets
+
+Apache ActiveMQ Artemis also supports AMQP over [Web
+Sockets](https://html.spec.whatwg.org/multipage/web-sockets.html).  Modern web
+browsers which support Web Sockets can send and receive AMQP messages.
+
+MQTT over Web Sockets is supported via a normal AMQP acceptor:
```
tabish121: Think you mean AMQP here, MQTT over the AMQP accepter seems wrong

jbertram: Fixed. Thanks!

PR #2974, opened 2020-02-06
ARTEMIS-2607 interceptor returns false but processing continues

jbertram:


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
```diff
@@ -507,8 +507,9 @@ private void serverSend(final ProtonServerReceiverContext context,
                            final Receiver receiver,
                            final RoutingContext routingContext) throws Exception {
       message.setConnectionID(receiver.getSession().getConnection().getRemoteContainer());
-      invokeIncoming((AMQPMessage) message, (ActiveMQProtonRemotingConnection) transportConnection.getProtocolConnection());
-      serverSession.send(transaction, message, directDeliver, false, routingContext);
+      if (invokeIncoming((AMQPMessage) message, (ActiveMQProtonRemotingConnection) transportConnection.getProtocolConnection()) == null) {
+         serverSession.send(transaction, message, directDeliver, false, routingContext);
+      }
 
       afterIO(new IOCallback() {
```
clebertsuconic: Did you think about the implications of the interceptor on the afterIO?

jbertram: In the case where the interceptor returns `false` I assumed the `IOCallback` wouldn't be invoked since `send` wasn't invoked. Is that not how it works?

clebertsuconic: not really... afterIO will wait any pending IO to be executed, or to be executed right away if there is no pending IO. in that sense if context is empty, it will execute in thread.

That means that the confirmation (ack for the sends) will happen right away.


I'm not sure what is the expected semantic from the user on this case.



I would expect the disposition to be rejected on that case... and not accepted. from this POV this is being accepted as if the message was actually stored and delivered.

clebertsuconic: this is one of the reasons it was never captured at this point. but if you decide to take the outcome of the interceptor into consideration, then you need proper cancellations at the send.

clebertsuconic: this was well done, but the semantic here is wrong.. I think this should be -1

unless we find the best definition here.

jbertram: I reworked this a bit essentially reproducing what we do on core.

PR #2975, opened 2020-02-09
ARTEMIS-2617 Improve AMQP Journal loading

franz1981:
I've opened this PR for discussion.

Currently it brings 4 improvements:

1. enable G1GC string deduplication
2. uses `CoreMessageObjectPools` on journal loading to save allocations
3. allows AMQP lazy message data scanning on journal loading
4. half-brute force search `x-opt-delivery-time` to avoid message data scanning

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
@@ -18,6 +18,7 @@ package org.apache.activemq.artemis.protocol.amqp.broker;
 
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
@@ -43,6 +44,7 @@ import org.apache.activemq.artemis.protocol.amqp.util.TLSEncode;
 import org.apache.activemq.artemis.reader.MessageUtil;
 import org.apache.activemq.artemis.utils.ByteUtil;
 import org.apache.activemq.artemis.utils.DataConstants;
+import org.apache.activemq.artemis.utils.algo.KMPNeedle;
 import org.apache.activemq.artemis.utils.collections.TypedProperties;
 import org.apache.qpid.proton.amqp.Binary;
 import org.apache.qpid.proton.amqp.Symbol;
@@ -74,12 +76,49 @@ import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.buffer.Unpooled;
 import org.jboss.logging.Logger;
 
-// see https://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#section-message-format
+/**
+ * See <a href="https://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#section-message-format">AMQP v1.0 message format</a>
+ * <pre>
+ *
+ *                                                      Bare Message
+ *                                                            |
+ *                                      .---------------------+--------------------.
+ *                                      |                                          |
+ * +--------+-------------+-------------+------------+--------------+--------------+--------+
+ * | header | delivery-   | message-    | properties | application- | application- | footer |
+ * |        | annotations | annotations |            | properties   | data         |        |
+ * +--------+-------------+-------------+------------+--------------+--------------+--------+
+ * |                                                                                        |
+ * '-------------------------------------------+--------------------------------------------'
+ *                                             |
+ *                                      Annotated Message
+ * </pre>
+ * <ul>
+ *    <li>Zero or one header sections.
+ *    <li>Zero or one delivery-annotation sections.
+ *    <li>Zero or one message-annotation sections.
+ *    <li>Zero or one properties sections.
+ *    <li>Zero or one application-properties sections.
+ *    <li>The body consists of one of the following three choices:
+ *    <ul>
+ *       <li>one or more data sections
+ *       <li>one or more amqp-sequence sections
+ *       <li>or a single amqp-value section.
+ *    </ul>
+ *    <li>Zero or one footer sections.
+ * </ul>
+ */
 public class AMQPMessage extends RefCountMessage {
 
    private static final Logger logger = Logger.getLogger(AMQPMessage.class);
 
    public static final SimpleString ADDRESS_PROPERTY = SimpleString.toSimpleString("_AMQ_AD");
+   // used to perform quick search
+   private static final Symbol[] SCHEDULED_DELIVERY_SYMBOLS = new Symbol[]{
+      AMQPMessageSupport.SCHEDULED_DELIVERY_TIME, AMQPMessageSupport.SCHEDULED_DELIVERY_DELAY};
+   private static final KMPNeedle[] SCHEDULED_DELIVERY_NEEDLES = new KMPNeedle[]{
+      AMQPMessageSymbolSearch.kmpNeedleOf(AMQPMessageSupport.SCHEDULED_DELIVERY_TIME),
+      AMQPMessageSymbolSearch.kmpNeedleOf(AMQPMessageSupport.SCHEDULED_DELIVERY_DELAY)};
 
    public static final int DEFAULT_MESSAGE_FORMAT = 0;
    public static final int DEFAULT_MESSAGE_PRIORITY = 4;
@@ -89,7 +128,10 @@ public class AMQPMessage extends RefCountMessage {
 
    // Buffer and state for the data backing this message.
    private ReadableBuffer data;
-   private boolean messageDataScanned;
+   private static final byte NOT_SCANNED = 0;
+   private static final byte RELOAD_PERSISTENCE = 1;
+   private static final byte SCANNED = 2;
+   private byte messageDataScanned;
 
    // Marks the message as needed to be re-encoded to update the backing buffer
    private boolean modified;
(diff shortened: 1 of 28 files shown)


franz1981:
@tabish121 @gemmellr @clebertsuconic @michaelandrepearce 

This PR contains many changes and I know there are parts that could be simplified a lot.
The most important bits that can really be a game-changer for AMQP on journal loading are on https://github.com/apache/activemq-artemis/pull/2975/commits/fc77a546ca1ee047218e938f0413c55ed8c838c6.

So please review me especially that part.

An important note: 
I've found that `AMQPMessagePersisterV2::decode` does something weird re `AMQPMessage::setAddress`: `AMQPMessagePersister::decode` can `AMQPMessage::setAddress`, allocating `AMQPMessage::extraProperties`, but 
`AMQPMessagePersisterV2::decode`  can decode a new `extraProperties` totally overwriting the existing one. Is it a bug or it's suppose to work like that?
I see that this behaviour has been introduced on `ARTEMIS-1858`  in 1ae2784dc6075875b18780fa8ba40f86cb895f7b with this comment:
```java
   /**
    * This will set the address on CoreMessage.
    *
    * Note for AMQPMessages:
    * in AMQPMessages this will not really change the address on the message. Instead it will add a property
    * on extraProperties which only transverse internally at the broker.
    * Whatever you change here it won't affect anything towards the received message.
    *
    * If you wish to change AMQPMessages address you will have to do it directly at the AMQP Message, however beware
    * that AMQPMessages are not supposed to be changed at the broker, so only do it if you know what you are doing.
    * @param address
    * @return
    */
   Message setAddress(SimpleString address);
```
I've tried to mitigate this by using `CoreMessageObjectPools` to save `SimpleString` allocations [here](https://github.com/franz1981/activemq-artemis/blob/918800b309009fd440bf83b69b54a0d4f49db940/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessagePersisterV2.java#L78) and by allowing `TypedProperties::decode` to append/replace any existing `extraProperty`, saving an `HashMap` allocation, but IMO it could be addressed by properly fixing it, wdyt? any idea how?

franz1981:
@michaelandrepearce I suppose you are interested in numbers: in my syntethic tests a 2 GB journal with very small messages has wasn't able to be loaded by a broker with `4GB` heap size, always leading to a OOM: now it can load without relevant GC pauses eg
image

We're not quite yet at the same level of Core messages, but very near: I think that the duplicate `address` is responsible of the difference here.

The commit that allows to half-brute force search `x-opt-delivery-time` is what I'm not sure about:

- that method should be optimized and performed optimally by proton itself IMO (nudge nudge @tabish121 @gemmellr), but I would be super happy to provide a PR for it
- currently it can give rare false positive (causing message scanning when not necessary)
- I've yet to verify how it behave with many message annotations and longer messages

If @michaelandrepearce has some real journal to test with and compare with `master` it would be great, given that this change can affect positively your usage of the broker ;)

franz1981:
In addition, I hope I haven't missed to force message data scanning on getter of other properties eg routingType.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1183,8 +1219,126 @@ public Object getCorrelationID() {
...
+      }
+      // messageDataScanned != SCANNED;
+      final ReadableBuffer data = this.data;
+      DecoderImpl decoder = TLSEncode.getDecoder();
+      final int position = data.position();
+      decoder.setBuffer(data.rewind());
+      try {
+         while (data.hasRemaining()) {
+            TypeConstructor<?> constructor = decoder.readConstructor();
+            if (MessageAnnotations.class.equals(constructor.getTypeClass())) {
```
tabish121: You should account for the message not having any MessageAnnotations otherwise you will scan in every type constructor for the follow on sections if present for no reason.  There's no requirement that a message have a Message Annotations section.

artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessageTest.java
```diff
@@ -156,6 +157,72 @@ public void testCreateMessageForPersistenceDataReload() throws ActiveMQException
       assertEquals(TEST_TO_ADDRESS, message.getAddress());
    }
 
+   @Test
```
tabish121: The tests seem a bit simplistic for such a change, what about annotations sections with many different annotations or annotations that don't match what you are looking for or are very very close but not quite what you are looking for etc.

franz1981: In this case it would lead to a false positive, that's ok in our case, given that is very unlikely that both values or keys are named similarly to a specific existing symbol.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
franz1981: How to quick search for this case? So you think it would beneficial to have such method directly on proton?

artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessageTest.java
franz1981: Probably I have understood what you mean and yes agree, I can improve the coverage with other cases :)

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
franz1981: I'm looking at 
image

If that's accurate I supposeI can use this structure to stop searching for `MessageAnnotations` earlier: any advices here is wellcome :)

tabish121: Once you pass the Header and Delivery Annotations (should they actually be there) than either you will get a Message Annotations section, or you will get one of the four possible sections that follows it or nothing if the message contains no sections following the delivery annotations.

franz1981: Great, tomorrow will add a commit to improve that + some tests.
Now my other concern is related to the lazely message scan: I've introduced a whole new state in the message and I hope it won't mess with some earlier assumptions now broken (routingType now force a message scan, while before it was assuming to always being correctly populated)

tabish121: As far as I recall the broker decodes the message annotations on each message specifically to read the routing type and delivery time on receive.  I'm unsure for routing type if that is needed on reload from journal but wouldn't be surprised if it doesn't get called at some point.

franz1981: Luckily enough on journal loading routing type seems not needed (we read it from other bits on journal record probably), while the only required message info seems the scheduled delivery time indeed

tabish121: Wouldn't it be nice if the journal entry stored all needed metadata in an easily parsed canonical format so that all protocols could benefit from the same optimizations.  :)

franz1981:
@tabish121 I've tried to improve the fail-fast behaviour for both the search algorithm and the `MessageAnnotations` section search. The latter by using an `IdentityHasMap` to save checking the N possible message sections that would terminate the algorithm: let me know if that makes sense according to the AMQP spec :+1:

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1183,8 +1263,129 @@ public Object getCorrelationID() {
...
+         return false;
+      }
+      // messageDataScanned != SCANNED;
+      final ReadableBuffer data = this.data;
+      DecoderImpl decoder = TLSEncode.getDecoder();
+      final int position = data.position();
+      decoder.setBuffer(data.rewind());
+      try {
+         while (data.hasRemaining()) {
+            TypeConstructor<?> constructor = decoder.readConstructor();
```
tabish121: As an alternative to the IdentityHashMap you might try casting the TypeConstructor to an AbstractDescribedType which the valid Sections in the message would be and then get the Descriptor code which is on UnsignedLong.  The values for each section would be 

- Header (70)
- DeliveryAnnotations (71)
- MessageAnnotations (72)
- Properties (73)
- ApplicationProperties (74)
- Data (75)
- AmqpSequence (76)
- AmqpValue (77) 
- Footer (78)

You'd need to handle the error cases but that could make for simpler logic, thing you want it 72 so keep going is less, stop if greater.

franz1981: uuuh that's much better indeed, thanks!
Error cases? I will take a look to AMQPMessage to see if there are any or you can point me some?

tabish121: Basically need to deal with case where TypeConstructor you get back isn't an AbstractDescribedType for some unknown reason, should have blown up on ingress but you never know.

franz1981: In this case probably throwing IllegalStateEx is the safer choice here

clebertsuconic:
I need this merged, as this will impact the changes I'm making for large messages.

I see that you still need other work done, such as adding more tests.. on which case can you please add any further changes into a new PR?

thanks a lot.

PR #2976, opened 2020-02-09
ARTEMIS-2325 + a test fix

jbertram:


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerImpl.java
```diff
@@ -138,24 +135,19 @@ public void send(SimpleString address1,
...
-      if (confirmationWindowEnabled) {
-         doSend(address1, message, handler);
-      } else {
-         doSend(address1, message, null);
-         if (handler != null) {
-            if (logger.isDebugEnabled()) {
-               logger.debug("Handler was used on producing messages towards address " + (address1 == null ? null : address1.toString()) + " however there is no confirmationWindowEnabled");
-            }
+      if (handler != null) {
+         handler = new SendAcknowledgementHandlerWrapper(handler);
```
clebertsuconic: this will create a new object for every message send called when no handlers in place?

that's a heavy change, no?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/SendAcknowledgementHandlerWrapper.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+package org.apache.activemq.artemis.core.client.impl;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import org.apache.activemq.artemis.api.core.Message;
+import org.apache.activemq.artemis.api.core.client.SendAcknowledgementHandler;
+
+public class SendAcknowledgementHandlerWrapper implements SendAcknowledgementHandler {
```
clebertsuconic: is this for every message sent?


if you need such thing I would rather add something to the ClientMessage itself instead of creating a new object.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerImpl.java
jbertram: No, it won't create a new object for every message send. It will only create a wrapper for the handler if a handler is set when a message is sent.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/SendAcknowledgementHandlerWrapper.java
jbertram: No, it isn't for every message sent. It's only used when a handler is set (either on the session or passed-in when sent).

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java
```diff
@@ -534,62 +524,56 @@ private CompletionListenerWrapper(CompletionListener listener,
...
-               try {
-                  ((BytesMessage) jmsMessage).reset();
-               } catch (JMSException e) {
-                  // HORNETQ-1209 XXX ignore?
-               }
+         if (jmsMessage instanceof StreamMessage) {
+            try {
+               ((StreamMessage) jmsMessage).reset();
+            } catch (JMSException e) {
+               // HORNETQ-1209 XXX ignore?
```
clebertsuconic: not related to your change but can you please remove this XXX?

This is mark I used to add on code.

I don't recall if I did this, but this is a thing I used to do to mark the code with stuff like During a big refactoring I would mark XXX, and remove it before merging.. so it was intended to be removed from 10 years ago :) 

you could replace this thing by just logger.debug

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/SendAcknowledgementHandlerWrapper.java
```diff
@@ -0,0 +1,62 @@
...
+
+   private SendAcknowledgementHandler wrapped;
+
+   /**
+    * It's possible that a SendAcknowledgementHandler might be called twice due to subsequent
+    * packet confirmations on the same connection. Using this boolean avoids that possibility.
+    * A new SendAcknowledgementHandlerWrapper is created for each message sent so once it's
+    * called once it will never be called again.
+    */
+   private AtomicBoolean active = new AtomicBoolean(true);
```
clebertsuconic: fair enough.. can you at least use a regular volatile boolean here?

the sendAck is called from a single thread. I think it's safe to use a regular boolean here.

if you really want Atomics used, can you use a AtomicIntegerFieldUpdater instead? have a counter and only call this if the counter == 0. AtomicBoolean will be heavier then a regular field.

That's the only change I will ask, on top of the favour I asked you. (removing the XXX)

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java
jbertram: done

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/SendAcknowledgementHandlerWrapper.java
jbertram: I've reworked this a bit to avoid the wrapper altogether.

jbertram: Scratch that. My non-wrapper solution had compatibility issues. :(

PR #2978, opened 2020-02-10
ARTEMIS-2563 Added option to create queues on all clustered nodes

AntonRoskvist:
This is to help with redistribution as mentioned in this Jira:
https://issues.apache.org/jira/browse/ARTEMIS-2563

Address option added to enable queue creation on all active nodes in a cluster, which happens on BINDING_ADDED. Since this creates a local binding for remote queues, a redistributor will also get created once a consumer registers

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
@@ -275,6 +275,8 @@ public final class FileConfigurationParser extends XMLConfigurationUtil {
 
    private static final String RETROACTIVE_MESSAGE_COUNT = "retroactive-message-count";
 
+   private static final String CLUSTERED_QUEUES_NODE_NAME = "clustered-queues";
+
 
    // Attributes ----------------------------------------------------
 
@@ -1185,6 +1187,8 @@ public final class FileConfigurationParser extends XMLConfigurationUtil {
             long retroactiveMessageCount = XMLUtil.parseLong(child);
             Validators.GE_ZERO.validate(RETROACTIVE_MESSAGE_COUNT, retroactiveMessageCount);
             addressSettings.setRetroactiveMessageCount(retroactiveMessageCount);
+         } else if (CLUSTERED_QUEUES_NODE_NAME.equalsIgnoreCase(name)) {
+            addressSettings.setClusteredQueues(XMLUtil.parseBoolean(child));
          }
       }
       return setting;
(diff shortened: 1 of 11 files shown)


jbertram:
I think there may be a simpler solution to this that would avoid adding a new configuration option. Currently {{org.apache.activemq.artemis.core.server.impl.QueueImpl#addRedistributor}} is only invoked when receiving a notification about a consumer being created or closed on another cluster member. I think it should be possible to check if there are consumers on other nodes when creating a queue and if so add a redistributor.

AntonRoskvist:
Yeah, that sounds good... but wouldn't it be preferable to have it configurable, to not change the way the broker works without other users being aware of it? Or do you think it would that be safe enough to implement it like that anyway? I'll certainly look in to it in that case, any idea where such an addition should be made?

jbertram:
In my opinion the current behavior is a bug since it's a gap between the way pre-configured and auto-created queues work. Therefore, I don't think it needs to be configurable.

Look at the callers of `org.apache.activemq.artemis.core.server.impl.QueueImpl#addRedistributor`. Some meta-data about the remote queues are stored there that you should be able to access when a queue gets created on the broker.

AntonRoskvist:
Okay, sounds good. I will have a look at it.

One other thing though, Tests are failing for my current pr, but they all pass when I run them locally, I even reran the specific tests that was failing here with no issues. Have I set something up incorrectly in my env? I get some log4j errors in the log.

jbertram:
The test failures look legitimate to me since you've modified `org.apache.activemq.artemis.core.settings.impl.AddressSettings` and `org.apache.activemq.artemis.tests.compatibility.JournalCompatibilityTest` is failing at the point where it's trying to decode address settings. It looks to me like the problem is in `org.apache.activemq.artemis.core.settings.impl.AddressSettings#decode(org.apache.activemq.artemis.api.core.ActiveMQBuffer, boolean)` because you're not checking `buffer.readableBytes() > 0` when trying to decode your new parameter which fails on an old journal which doesn't have that set.

jbertram:
Just to be clear...

Are you working on the alternative solution that I mentioned previously? I wouldn't merge this PR until it was clear that other potential solution wasn't viable.

AntonRoskvist:
Sorry, I am looking in to the alternative solution but have gotten nowhere. I am pretty new to coding and have gotten completely swamped with other stuff so I am not making the progress I was hoping for.

I have yet to find a good way to poll the cluster members for consumer information on individual queues. Once I find that I'll be good to go, but it might take a while considering my current other work load.

AntonRoskvist:
Would it be safe to just add a redistributor on receiving core notification ADD_BINDING if distance on that notification message is 0? That is, add a redistributor (if redistributiondelay is greater than -1) on creation of all local bindings? I tried it and all tests (as far as i can tell) are passing...

PR #2980, opened 2020-02-10
ARTEMIS-2619 allow disabling STOMP 'server' header

jbertram:


artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/v11/StompFrameHandlerV11.java
```diff
@@ -85,7 +85,10 @@ public StompFrame onConnect(StompFrame frame) {
          response.addHeader(Stomp.Headers.Connected.SESSION, connection.getID().toString());
 
          // server
-         response.addHeader(Stomp.Headers.Connected.SERVER, connection.getActiveMQServerName());
+         Object disableServerHeader = connection.getAcceptorUsed().getConfiguration().get(TransportConstants.DISABLE_STOMP_SERVER_HEADER);
+         if (disableServerHeader == null || (disableServerHeader != null && !Boolean.parseBoolean(disableServerHeader.toString()))) {
```
gaohoward: Is the disableServerHeader != null check not necessary?

jbertram: Yes, I think you're right. If `disableServerHeader == null` then it will short-circuit the other check and enter the block.

PR #2981, opened 2020-02-11
NO-JIRA Fix federated-address-downstream pom

brusdev:


cshannon:
The bigger issue here seems to be that the release profile in the broker-federation parent pom is missing all the modules so that caused the version to not be updated

cshannon:
I will go ahead and fix the commit with that change and merge it

PR #2982, opened 2020-02-12
ARTEMIS-2620 Avoid Exception on client when large message file corrupted

gaohoward:
The broker should be more resilient so that it can avoid throwing an
exception on the client side when it encounters an invalid or empty
large message file(corrupted for any reason). It would be better to
deal with it within the broker and not allow the issue to manifest
itself on the consuming client.

Also fixed an ActiveMQServerMessagePlugin issue where beforeSend(),
afterSend(), sendOnException() are not get called with large
messages.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/LargeBodyReader.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/LargeBodyReader.java
@@ -62,4 +62,13 @@ public interface LargeBodyReader {
     * This method must not be called directly by ActiveMQ Artemis clients.
     */
    long getSize() throws ActiveMQException;
+
+
+   /**
+    * Check if the large message file matches its size
+    * @return true if matches, false otherwise
+    */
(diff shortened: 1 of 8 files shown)


clebertsuconic:
rebase?

clebertsuconic:
You have changed the ServerSessionHandler... if the large message is damaged at that point, any exceptions should definitely be sent to the client.

I don't think we should change this.

perhaps we should look closely to what was the original issue before this was raised?

clebertsuconic:
I agree with adding a plugin event.. 

but I don't think we should change semantics on the sending exceptions to the client.

gaohoward:
That was user's idea that the error should be detect early and not to send to client.
I'd agree that we fix the original issue (so far we lack information as to details how the large message was corrupted and lack of a reproducer).

jbertram:
@gaohoward, should this just be closed at this point?

gaohoward:
@clebertsuconic @jbertram Sorry guys it's been a long time and this hasn't been mentioned by any ever since. And the PR has very much outdated. I agree with Justin that we close this at the moment.

PR #2984, opened 2020-02-14
ARTEMIS-2617 Fixing KMPNeedle::searchInto API and specializing it

franz1981:
KMPNeedle::searchInto has been specialized and copied
to handle ReadableBuffer in order to save polymorphic
calls on it that would make it slower on hot paths.


franz1981:
I still need to write some tests on it: I've benchmarked the version with a polymorphic lamda getting mixed feeling on it and I've chosen to safely specialize it to get the best possible performance for it.

PR #2985, opened 2020-02-17
ARTEMIS-2587 & ARTEMIS-2624 auto-create dead-letter & expiry resources

jbertram:
This is a reimplementation of the IndividualDeadLetterQueueStrategy
from 5.x in a way that makes sense with the Artemis addressing model.


jbertram:
Most of the changes here are for the new address-settings, tests, docs, etc. The actual substantive changes to support this feature is a 10-line function - `org.apache.activemq.artemis.core.server.impl.QueueImpl#createDeadLetterResources`

michaelandrepearce:
Nice. Is it possible to do same for expiry?

jbertram:
@michaelandrepearce, yes the same can be done for expiry. However, that would be on different Jira & PR.

clebertsuconic:
@jbertram perhaps instead of creating a specialized option for DLQ and a specialized option for Expiry, we could make it generic, so the same feature would apply to both?

instead of auto-create-DLQ.. we could call it auto-create-output-resources ? something on that line.. dunno?

jbertram:
FYI - I just pushed a commit to deal with the expiry resources.

clebertsuconic:
@jbertram looks good.. there's a test failure on the PR though. seems legit.

clebertsuconic:
it's probably due to the config changes performed. the test probably needs to be adapted.

PR #2986, opened 2020-02-19
ARTEMIS-1975 Real Large Message support into AMQP

clebertsuconic:
This is Large commit, where I am refactoring largeMessage Body out of CoreMessage
which is now reused with AMQP.

I had also to fix Reference Counting to fix how Large Messages are Acked

And I also had to make sure Large Messages are transversing correctly when in cluster.


artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
```diff
@@ -19,66 +19,155 @@
...
+         debugRefs();
+         fired = true;
+         releaseComplete();
+      }
+   }
+   /**
+    *
+    * This method will be useful if you remove commented out code around #ifdef AND #endif COMMENTS
+    * */
+   public final void debugRefs() {
```
clebertsuconic: I'm keeping these if I ever need to debug these again.

Too bad there's no IFDEF in java.

clebertsuconic:
I have done my own review and I realized I missed some flow control logic. this last commit is adding it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
```diff
@@ -278,39 +283,66 @@ public void onMessage(Delivery delivery) throws ActiveMQAMQPException {
...
+            receiver.advance();
+            delivery.settle();
+
+            // Replenish the credit if not doing a drain
+            if (!receiver.getDrain()) {
+               receiver.flow(1);
+            }
+
+            return;
+         } else if (delivery.isPartial()) {
```
tabish121: I think you need to add some more metrics here on what constitutes a "large" message as there isn't anything that requires a client to strictly adhere to max frame size limits other that it shouldn't go over, it can certainly send under that amount.  A client could split a 4k message into 4 1k chunks etc so assuming that a message is large just based on partial will lead to likely considering messages that don't meet the normal Artemis definition of a large message as being large.  Also even if not partial you ought to be checking if what you have in a complete message that meets your definition of a large message as the AMQP max frame size is configurable on the protocol adapter so you might need a second metric there to say it's not partial but it exceeds what we consider "not large" and would prefer to store it in the large message way.

clebertsuconic: @tabish121 makes sense.. thanks a lot.. will add another commit to the REVIEW

clebertsuconic: I was actually looking for that information.. if it made sense to use frame size or not.

clebertsuconic: @tabish121 commit added

These commits will be squahsed after we are done with the review.

clebertsuconic:
Doing my own review here: since I now added some configuration option for minlargeMessageSize, I will need to touch some documentation around Large Messages.

I'm adding some paragraphs to the large message size chapter. I will add a new commit shortly.

clebertsuconic:
@tabish121 
@franz1981 
@jbertram 

I'm going to squash this tomorrow. I had added separate commits this time instead of squashing to make the review eaiser.


Please let me know what you think... I think this is ready to be merged to I will squash tomorrow.

clebertsuconic:
I'm marking do not merge yet.. but please keep reviewing.

clebertsuconic:
I rebased.. and I think this is now ready to be merged.

franz1981:
Going to merge this: well done bud!

PR #2987, opened 2020-02-20
[ARTEMIS-2626]: Postgresql Journal implementation requires the jdbc driver to be in the same classloader

ehsavoie:
Issue: 
The BLOB manipulation is done using PostgreSQL internal classes starting from PGConnection. This leads to ClasCastExceptions if the connection is wrapped in a pool or if the driver is in a different classloader (WildFly).

Fix: 
unwrap the connection and if the PostgreSQL classes are not directly available uses reflection to manipulate the BLOBs.
Jira: https://issues.apache.org/jira/browse/ARTEMIS-2626

Supersedes https://github.com/apache/activemq-artemis/pull/2948

### DIFF
--- a/artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/file/PostgresLargeObjectManager.java
+++ b/artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/file/PostgresLargeObjectManager.java
@@ -1,0 +1,209 @@
+/*
+ * Copyright 2019 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


clebertsuconic:
@ehsavoie sort of a nit-pic but the only thing wrong here is the title for the commit

the bug is Postgres JDBC requires JDBC driver..


The fix is... allowing Postgres to use the driver in a certain way... etc..


Can you be a bit more descriptive of the fix itself instead of using the bug description?

clebertsuconic:
It's just that when I look for a list of commits, I like to understand what was changed.. to follow cherry-picks.. etc..

ehsavoie:
Is this better ?

clebertsuconic:
Nice !!!   I thought about something simple. You went beyond complete now.  Thanks a lot.

PR #2990, opened 2020-02-21
ARTEMIS-2627 simpleSecureServer failing on IBM JDK 8

brusdev:
Remove excluded cipher suites matching the prefix `SSL` because the names of the
IBM JDK 8 cipher suites have the prefix `SSL` while the
`DEFAULT_EXCLUDED_CIPHER_SUITES` of org.eclipse.jetty.util.ssl.SslContextFactory
includes "^SSL_.*$". So all IBM JDK cipher suites are excluded by
SslContextFactory using the `DEFAULT_EXCLUDED_CIPHER_SUITES`.

### DIFF
--- a/artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
+++ b/artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
@@ -72,6 +72,20 @@ public class WebServerComponent implements ExternalComponent {
          SslContextFactory.Server sslFactory = new SslContextFactory.Server();
          sslFactory.setKeyStorePath(webServerConfig.keyStorePath == null ? artemisInstance + "/etc/keystore.jks" : webServerConfig.keyStorePath);
          sslFactory.setKeyStorePassword(webServerConfig.getKeyStorePassword() == null ? "password" : webServerConfig.getKeyStorePassword());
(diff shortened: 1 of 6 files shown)


artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
```diff
@@ -72,6 +74,11 @@ public void configure(ComponentDTO config, String artemisInstance, String artemi
          SslContextFactory.Server sslFactory = new SslContextFactory.Server();
          sslFactory.setKeyStorePath(webServerConfig.keyStorePath == null ? artemisInstance + "/etc/keystore.jks" : webServerConfig.keyStorePath);
          sslFactory.setKeyStorePassword(webServerConfig.getKeyStorePassword() == null ? "password" : webServerConfig.getKeyStorePassword());
+         sslFactory.setIncludeProtocols("TLSv1.2");
```
jbertram: I'm concerned about changing the default here for all cases. This should probably be made configurable and then the test can configure it as necessary.

brusdev: I added some parameters to the WebServerDTO so the test can configure them.

PR #2991, opened 2020-02-21
ARTEMIS-2629 ensure queue auto-delete after expiration

jbertram:


jbertram:
I just realized that this won't work if, for example, a message is added which expires after a broker restart.

PR #2992, opened 2020-02-22
ARTEMIS-2629 ensure queue auto-delete after expiration

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueManagerImpl.java
```diff
@@ -84,8 +84,7 @@ public static boolean messageCountCheck(Queue queue) {
    }
 
    public static boolean delayCheck(Queue queue) {
-      long consumerRemovedTimestamp =  queue.getConsumerRemovedTimestamp();
-      return consumerRemovedTimestamp != -1 && System.currentTimeMillis() - consumerRemovedTimestamp >= queue.getAutoDeleteDelay();
+      return System.currentTimeMillis() - queue.getConsumerRemovedTimestamp() >= queue.getAutoDeleteDelay();
```
brusdev: Removing the condition `consumerRemovedTimestamp != -1` could change the semantic of the **public** method `delayCheck`. With this change `delayCheck` will return **true** before removing a consumer.

jbertram: It *will* change the semantic, and that was intended. This particular check is no longer desirable as it will preclude auto-deleting a queue where all the messages have expired but no consumer has ever connected. I implemented a new check in `org.apache.activemq.artemis.core.postoffice.impl.PostOfficeImpl.AddressQueueReaper#queueWasUsed` which resolves this issue. The `delayCheck` method is only used internally as it's not part of the user-facing API so I don't have any qualms with changing its semantics.

PR #2994, opened 2020-02-28
ARTEMIS-2637 Making UDP client discovery resilient

clebertsuconic:
In case there is a hardware, firewal or any other thing making the UDP connection to go deaf
we will now reopen the connection in an attempt to go over possible issues.

This is also improving locking around DiscoveryGroup initial connection.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/topology/NettyHAClientTopologyWithDiscoveryTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/topology/NettyHAClientTopologyWithDiscoveryTest.java
@@ -16,11 +16,146 @@
  */
 package org.apache.activemq.artemis.tests.integration.cluster.topology;
 
+import org.apache.activemq.artemis.api.core.DiscoveryGroupConfiguration;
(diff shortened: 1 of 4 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/topology/NettyHAClientTopologyWithDiscoveryTest.java
```diff
@@ -16,11 +16,146 @@
...
 
    @Override
    protected boolean isNetty() {
       return true;
    }
 
+
+
+   @Test
+   public void testRecoveryBadUDPWithRetry() throws Exception {
```
jbertram: There are no assertions in this test?

clebertsuconic: no. the retry should fix it and it should succeed.

The assertion is no failures.

clebertsuconic:
there's a test failure on the main testsuite (the one that takes 2 hours).

and I'm adding some checks around serialization.

clebertsuconic:
fixed the test failure

PR #2995, opened 2020-02-29
Suggest not using blowfish for security concern

YYTVicky:


michaelandrepearce:
This would be a breaking change for those using it. Best way here is to make a new option that people can use, but doesnt break existing setups. Also corresponding doc updates needed

jbertram:
I agree with @michaelandrepearce.

jbertram:
I'm closing this for now. Once you've addressed @michaelandrepearce's comments please re-open.

Also, it's worth noting that anybody who doesn't like `org.apache.activemq.artemis.utils.DefaultSensitiveStringCodec` for whatever reason can implement and use their own version of `org.apache.activemq.artemis.utils.SensitiveDataCodec`.

YYTVicky:
@michaelandrepearce Thanks for your feedback, since blowfish is a broken algorithm, there are some document list the reasons, such as this one https://threatpost.com/new-collision-attacks-against-3des-blowfish-allow-for-cookie-decryption/120087/.  our tool report that as a vulnerability issue, would you mind trying **AES** to see if it could pass the test? 
Thanks!

michaelandrepearce:
The point is that aes should be additive not in replacement. It cannot be a breaking change

YYTVicky:
@michaelandrepearce @jbertram  Thanks a lot for this information, we would try to improve our project to handle this case and try to generate more flexible fix suggestions!

PR #2998, opened 2020-03-04
ARTEMIS-2641 Openwire client runs out of credits after reconnection

brusdev:
Clear the messagePullHandler on setting prefetchSize.

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
@@ -27,6 +27,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.activemq.advisory.AdvisorySupport;
 import org.apache.activemq.artemis.api.core.ActiveMQQueueExistsException;
@@ -68,7 +69,7 @@ public class AMQConsumer {
    private int prefetchSize;
    private final AtomicInteger currentWindow;
    private long messagePullSequence = 0;
-   private MessagePullHandler messagePullHandler;
+   private final AtomicReference<MessagePullHandler> messagePullHandler = new AtomicReference<>(null);
    //internal means we don't expose
    //it's address/queue to management service
    private boolean internalAddress = false;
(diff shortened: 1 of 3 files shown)


artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
```diff
@@ -380,6 +383,11 @@ public void setPrefetchSize(int prefetchSize) {
       this.prefetchSize = prefetchSize;
       this.currentWindow.set(prefetchSize);
       this.info.setPrefetchSize(prefetchSize);
+      if (this.prefetchSize == 0) {
+         messagePullHandler.compareAndSet(null, new MessagePullHandler());
```
franz1981: what happen if it won't succeed?

franz1981: @michaelandrepearce any suggestions?

michaelpearce-gain: what do you mean if it wont succeed? if its not null it means it already had a pullhandler set, no need to assign a new one.

michaelpearce-gain:
LGTM (obviously ;) ), nice test, that was the only bit i was struggling with how to think to implement.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
franz1981: that can happen if: 

1. there is a racing thread doing it 
2. the value isn't null

in both cases probably logging some trace/debug is ok if we expect it to happen?

franz1981: Anyway It's fine to me; if there are no objections I'm going to merge this ;)

michaelpearce-gain: if its not null, we dont want to change it. e.g. the case there is client was already at prefetch 0 and simply sent a second update which set it to the same size, as such you want to leave the original pullhandler.

michaelpearce-gain: if you really want a log i would add it as a trace, i dont think its needed tbh i think it just clutter the code base.

brusdev: no need to set messagePullHandler value if it isn't null

franz1981: :+1:  thanks for the explanation bud :)

michaelpearce-gain: > Anyway It's fine to me; if there are no objections I'm going to merge this ;)

no objections here, please go ahead....

PR #3000, opened 2020-03-04
ARTEMIS-2716 Moving ServerLocatorConfig together with ServerLocator

clebertsuconic:


artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/client/ServerLocator.java
```diff
@@ -19,7 +19,7 @@
 import java.util.concurrent.Executor;
 import java.util.concurrent.ScheduledExecutorService;
 
-import org.apache.activemq.artemis.api.config.ServerLocatorConfig;
+import org.apache.activemq.artemis.api.core.client.loadbalance.ServerLocatorConfig;
```
clebertsuconic: oops.. that's not where I meant it to be..

Glad I'm doing my own review

michaelandrepearce:
what's the reason we are moving a config file out of the config package? isn't the idea all config classes live in config packaged?

clebertsuconic:
@michaelandrepearce this is a new class, recently introduced. Howard introduced it to externalize the configuration as part of ARTEMIS-2716

this is tightly coupled with ServerLocator, and I consider the API to be cleaner if they live together in the same package.

Right now there's a single class in config for Default configuration. I didn't think it would fit in there.

PR #3002, opened 2020-03-05
ARTEMIS-2644 Include client id into non durable subscriber queue name

brusdev:


michaelandrepearce:
Guys this is a breaking change as no longer with use core naming enabled the naming patterns match with amqp. 

Also this merged very quickly for a non critical fix can we please ask for giving others time to review in future.

Can either this revert this or ensure amqp is alinged when amqp core naming is toggled

jbertram:
@michaelandrepearce, feel free to revert.

michaelandrepearce:
Will revert in the morning then.

brusdev:
@michaelandrepearce @jbertram I'm going to create a new PR to avoid breaking change.

michaelpearce-gain:
@brusdev will you have the PR today? If so ill hold off the revert, else will revert still, just so it doesnt leave master with the issue.

brusdev:
@michaelandrepearce I'm working on it.

michaelpearce-gain:
@brusdev looking at the jira, this implementation requires the client to set client id, which isnt always advisable. To meet the needs of the jira, 
other identifying information also be present, e.g. the user + hostname

brusdev:
@michaelandrepearce I created the PR #3006 to fix the breaking change, ATM I'm using only the clientId if present. Can you review the PR, please?

michaelpearce-gain:
i think we need to address both the issues, to not revert this,

PR #3003, opened 2020-03-05
ARTEMIS-2645 refactor CLI FQQN support

jbertram:
FQQN support for the CLI was implemented via ARTEMIS-1840 before general
FQQN support was added for producers via ARTEMIS-1867. The CLI's FQQN
functionality is slightly different from what is now generally available
and it can be confusing for users. By refactoring the CLI to use the
general FQQN support the code can be much simpler and consistent with
the expected behavior. Refactoring includes:

 - Deprecating the use of "fqqn://". The CLI commands use JMS so using
   "fqqn://" (instead of "queue://" or "topic://") makes the destination
   type ambiguous which can yield unexpected message routing behavior.
   Now "queue://" and "topic://" can be used with the normal FQQN syntax
   (e.g. address::queue).
 - Eliminating the use of the _AMQ_ROUTE_TO header when sending messags
   to an FQQN. The _AMQ_ROUTE_TO header is an internal header used when
   routing messages over a cluster bridge. Using it in the CLI for FQQN
   support was a clever hack, but using the general FQQN support
   eliminates complexity and makes behavior consistent between
   standalone JMS clients using FQQN and the CLI.
 - De-duplicating MessageSerializer initialization boilerplate.
 - Removing limitation where using an FQQN with an anycast address
   required the same name for the address and queue.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/Browse.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/Browse.java
@@ -49,10 +49,15 @@ public class Browse extends DestAbstract {
             } else {
                session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
             }
-            Destination dest = lookupDestination(session);
+            Destination dest = getDestination(session);
             threadsArray[i] = new ConsumerThread(session, dest, i);
 
-            threadsArray[i].setVerbose(verbose).setSleep(sleep).setMessageCount(messageCount).setFilter(filter).setBrowse(true);
+            threadsArray[i]
+               .setVerbose(verbose)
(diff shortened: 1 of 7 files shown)


michaelandrepearce:
@jbertram I'm getting test failures

Failed tests: 
  MessageSerializerTest.testTextMessageImportExport:177->checkSentMessages:104->CliTestBase.consumeMessages:147 null
Tests in error: 
  MessageSerializerTest.testMapMessageImportExport:227 Â» InvalidDestination AMQ2...

jbertram:
@michaelandrepearce, there must be a race condition somewhere. When I run those tests I don't get any failures. I'll continue investigating.

jbertram:
@michaelandrepearce I was able to reproduce the failures and I've sent #3009 to address them.

PR #3005, opened 2020-03-06
ARTEMIS-2636: Introduce Disk Usage Metrics

atris:


clebertsuconic:
the PR test is failing.. can you check why it failed? running tests...

clebertsuconic:
sorry.. I closed it by accident.. reopened it.

clebertsuconic:
thjere's a test failure here:

 MessageSerializerTest.testObjectMessageImportExport:202->CliTestBase.consumeMessages:147 null


Are you sure it's not related? I will have to take a look.

clebertsuconic:
can y ou run the test locally and let us know what you see?

jbertram:
I did some work in `org.apache.activemq.cli.test.MessageSerializerTest` for PR #3003, but every test in that class passes for me locally (both with and without the commit from this PR). Perhaps it's a spurious failure?

atris:
The test failed again -- doesnt seem associated

michaelandrepearce:
@jbertram I'm getting test failures in CLI tests locally. 

Failed tests: 
  MessageSerializerTest.testTextMessageImportExport:177->checkSentMessages:104->CliTestBase.consumeMessages:147 null
Tests in error: 
  MessageSerializerTest.testMapMessageImportExport:227 Â» InvalidDestination AMQ2...

michaelandrepearce:
checking build histories of master on travis, seems as soon as PR #3003  was merged, the build starting failing with the issue, as such does look like it is related to PR #3003 

here is the build on master straight after 3003 was merged
https://travis-ci.org/apache/activemq-artemis/builds/658977926?utm_source=github_status&utm_medium=notification

michaelandrepearce:
Actually seems since 2999 was merged https://travis-ci.org/apache/activemq-artemis/builds/658977370?utm_source=github_status&utm_medium=notification

there seems to been a number of merges very close together as such could be any that were merged before 2999 as unfortunately as merges so close together previous builds cancelled for next.

merges between last success build and first build failed, as such these are the smoking guns
2973, 2992, 2993, 2997, 2999

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.java
```diff
@@ -28,6 +28,7 @@
    String CONNECTION_COUNT_DESCRIPTION = "Number of clients connected to this server";
    String TOTAL_CONNECTION_COUNT_DESCRIPTION = "Number of clients which have connected to this server since it was started";
    String ADDRESS_MEMORY_USAGE_DESCRIPTION = "Memory used by all the addresses on broker for in-memory messages";
+   String DISK_STORE_USAGE_DESCRIPTION = "Memory used by the disk store";
```
criew: Space / bytes, not memory? Also JavaDoc comments below.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -2996,6 +2996,7 @@ private void registerMeters() {
             builder.register(BrokerMetricNames.CONNECTION_COUNT, this, metrics -> Double.valueOf(getConnectionCount()), ActiveMQServerControl.CONNECTION_COUNT_DESCRIPTION);
             builder.register(BrokerMetricNames.TOTAL_CONNECTION_COUNT, this, metrics -> Double.valueOf(getTotalConnectionCount()), ActiveMQServerControl.TOTAL_CONNECTION_COUNT_DESCRIPTION);
             builder.register(BrokerMetricNames.ADDRESS_MEMORY_USAGE, this, metrics -> Double.valueOf(getPagingManager().getGlobalSize()), ActiveMQServerControl.ADDRESS_MEMORY_USAGE_DESCRIPTION);
+            builder.register(BrokerMetricNames.DISK_STORE_USAGE, this, metrics -> Double.valueOf(getPagingManager().getDiskTotalSpace() - getPagingManager().getDiskUsableSpace()), ActiveMQServerControl.DISK_STORE_USAGE_DESCRIPTION);
```
criew: Add two more metrics - ADDRESS_MEMORY_USAGE_PERCENTAGE and DISK_STORE_USAGE_PERCENTAGE here

atris: Since address memory usage is not exposed as a metric here, I think it warrants a separate PR (to add both percentage types to exposed metrics). I will follow up with a PR

atris:
> Thanks for adding the new features suggested in the ticket.
> I was not sure where to add my comments, so I created a review, hope that's ok.
> Would it be possible to add *_PERCENTAGE metrics as well?

Thanks, fixed your comments.

atris:
I dont understand the test failure -- the test passes locally for me

jbertram:
The failure is caused by a race condition from #3003. I believe that's now been resolved via #3009. Please rebase and squash your commits.

atris:
Done, please see.

clebertsuconic:
@atris you should rebase, and squash  your commit

atris:
@clebertsuconic I believe I did that? (The PR is rebased on top of your latest commit to master, and is squashed to a single commit. What did I miss?)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
clebertsuconic: You are duplicating the attributes here?

Also DISK_STORE_USAGE.. this seems to give you the amount of free space available, instead of DISK Usage. Isn't that a mistake?

atris: Isnt disk usable space the free space?

michaelandrepearce: Typically the words used for disk reporting are used and free to avoid any confusion.

Youre using the word usage which to me suggest its the used space not the freem

I would recommend we use clearer words like used and free

atris:
Rebased

michaelandrepearce:
Have you sorted better naming so its clearer? See inline comments by clebert and myself

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -707,6 +707,24 @@ public long getAddressMemoryUsage() {
       }
    }
 
+   @Override
+   public long getDiskStoreUsed() {
```
clebertsuconic: I'm a bit confused between the metric and the math here... 

disk space - disk-usable? what that means!?

atris: If I understand correctly, disk usable represents the available space. Subtracting that from the total disk should give the disk used?

There is no documentation for disk usable method so I might be wrong.

jbertram: Ultimately the calls here are using [`java.nio.file.FileStore#getUsableSpace`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileStore.html#getUsableSpace--) and [`java.nio.file.FileStore#getTotalSpace`](https://docs.oracle.com/javase/8/docs/api/java/nio/file/FileStore.html#getTotalSpace--) so the math looks right.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -726,6 +744,25 @@ public int getAddressMemoryUsagePercentage() {
...
+      // no max size set implies 0% used
+      if (globalMaxSize <= 0) {
+         return 0;
+      }
+
+      long diskUsed = getDiskStoreUsed();
+      if (diskUsed <= 0) {
+         return 0;
+      }
+      double result = (100D * diskUsed) / globalMaxSize;
```
jbertram: You should use `org.apache.activemq.artemis.core.server.files.FileStoreMonitor#calculateUsage` here.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/BrokerMetricNames.java
```diff
@@ -22,5 +22,5 @@
    public static final String CONNECTION_COUNT = "connection.count";
    public static final String TOTAL_CONNECTION_COUNT = "total.connection.count";
    public static final String ADDRESS_MEMORY_USAGE = "address.memory.usage";
-
+   public static final String DISK_STORE_USED = "disk.store.usage";
```
jbertram: Your names & descriptions alternate between "used" and "usage". You should use the same verbiage everywhere to avoid confusion. For example, this line could be `DISK_STORE_USED = "disk.store.used"` or `DISK_STORE_USAGE = "disk.store.usage"`. I would go with "usage" since the address memory metric already follows this pattern.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
atris: Thanks for validating!

atris:
@jbertram @clebertsuconic @criew Fixed, please see and let me know your comments

clebertsuconic:
How did you come up with these formulas? did they come from the JIRA? or you made up them?
as an user I would prefer seeing how much space I have left.. and what's the percentage of the disk that it's used

atris:
@clebertsuconic Updated, please see and let me know if it looks fine now

PR #3006, opened 2020-03-06
ARTEMIS-2644 Fix AMQP naming convention in compatibility mode with CORE

brusdev:


michaelpearce-gain:
@brusdev i think we need to address the issue of no client id before this can merge, dont really want to have two lots of changes in changing the naming strategy;s

brusdev:
@michaelandrepearce I used user and hostname if no clientid for AMQP queue. Can you check if that's what you meant?

michaelandrepearce:
I kind of meant use all of it

michaelandrepearce:
E.g.  ID:<clientid>-IP:<ip>-USER:<user>-<uuid>

michaelandrepearce:
As well should align openwire here also just noticed it has same logic in its naming see AMQConsumer

michaelandrepearce:
To be honest the more you look at this changing this logic everywhere is lookinh dangerous as breaking some bits possible.

Is the issue that you just need to identify the creator of a queue? Instead of changing the naming strategy why not have meta data on the queue itself, e.g  add client id, user and remote address onto the queue meta data?

This also then would work even if clients not upgraded which with core youre relying on.

michaelandrepearce:
Im going to go ahead and revert whats in master till this is solved, the more i look at this having the extra details at queue level rather than mucking around with changing the existing naming strategy will be less impactful to all

michaelandrepearce:
If a user really really wants the queue name to have a clientid in, they can make a shared non durable subscription with clientid set on connection and then set subscription name to a uuid via jms api

brusdev:
I could revert all changes and add a prefix to the UUID without a separator, ie
<clientid>-<user>-<ip>-<uuid>. WDYT?

michaelandrepearce:
I think we dont touch the naming. If people need to know who created a queue put it on the queue as meta data.

If people really want to play with queue names its actually possible already today as i had pointed out.

PR #3008, opened 2020-03-09
ARTEMIS-2647: Fix append-to-file for mysql

graben:


brusdev:
LGTM. I checked it using mysql-connector-java 5.1.48 with mysql-server 5.5, 5.6 and 5.7.

PR #3011, opened 2020-03-10
ARTEMIS-2650 The delivering count is wrong after reconnecting an openwire client

brusdev:
Fix the conversion of the message id from the CORE messages.
Fix the credits acquired for acknowledges related to undelivered messages.


clebertsuconic:
checkstyle

PR #3012, opened 2020-03-10
ARTEMIS-2642 Fixing Drain Timeout Issue on AMQP

clebertsuconic:
The drain control has to immediately flush
otherwise a next flow control event may remove the previous status from Proton.
So, this really cannot wait the next executor, and it has to be done immediately.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/DrainTimeoutTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/DrainTimeoutTest.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


clebertsuconic:
@michaelandrepearce can you review this one please?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/DrainTimeoutTest.java
```diff
@@ -0,0 +1,112 @@
...
+      final AtomicInteger errors = new AtomicInteger(0);
+      final String queueName = getQueueName();
+      JmsConnectionFactory connectionFactory =
+         new JmsConnectionFactory(
+            "tqadmin",
+            "admin",
+            "amqp://localhost:5672?jms.prefetchPolicy.all=1&jms.connectTimeout=60000&amqp.drainTimeout=1000");
+      connectionFactory.setExceptionListener(
+         e -> {
+            System.out.println("Got a JMSException.  Terminating the VM.");
```
clebertsuconic: removing this line.. part of my own review.

michaelpearce-gain:
@clebertsuconic looks promising to fix one of the issues we saw under high load, not sure it 100% explains flow stopping and producer not getting its acks. Will need to patch our existing version and test in a server environment, not sure you're aware we are unable to reproduce the load and the issue locally

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/DrainTimeoutTest.java
```diff
@@ -0,0 +1,93 @@
...
+      try (JMSContext context = connectionFactory.createContext(Session.AUTO_ACKNOWLEDGE)) {
+         final Message message = context.createMessage();
+         message.setStringProperty("selector", "dude");
+         JMSProducer producer = context.createProducer();
+         Queue queue = context.createQueue(queueName);
+         while (sendCount.sum() < NUMBER_OF_MESSAGES && !Thread.interrupted()) {
+            producer.send(queue, message);
+            sendCount.increment();
+            long sent = sendCount.sum();
+            if (sent % 100 == 0) {
```
clebertsuconic: hmmm... removing this
I used to have a System.out here

michaelpearce-gain:
Have taken this, built and deployed, and we have hit same issue with the flow ceasing, i dont think this has fixed the underlying issue we have.

PR #3014, opened 2020-03-11
ARTEMIS-2653: Update to Proton-J 0.33.3 and Qpid JMS 0.50.0

gemmellr:


michaelandrepearce:
Isnt qpid jms 0.50.0 about to release? Should we just wait for that?

gemmellr:
It will be available next week if it passes its vote, however I don't see a reason to wait. It's mainly a test dep, theres more useful changes in this than 0.50.0, and the proton-j dep is the same for both anyway.

michaelpearce-gain:
@gemmellr make sense +1 then from me

PR #3015, opened 2020-03-11
ARTEMIS-2654 Improving Page cleanup and Operation Context

clebertsuconic:


clebertsuconic:
@michaelandrepearce I think this is probably more likely to be your issue.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
```diff
@@ -517,6 +528,23 @@ protected void cleanupComplete(ArrayList<PageSubscription> cursorList) throws Ex
...
+         @Override
+         public void done() {
+
+            for (PageSubscription cursor : cursorList) {
+               cursor.enableAutoCleanup();
+            }
+         }
+
+         @Override
+         public void onError(int errorCode, String errorMessage) {
```
michaelpearce-gain: what about error handling?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
```diff
@@ -612,26 +640,15 @@ protected void onDeletePage(Page deletedPage) throws Exception {
...
-         // we just need to make sure the storage is done..
-         // if the thread pool is full, we will just log it once instead of looping
-         if (!storageManager.waitOnOperations(5000)) {
-            ActiveMQServerLogger.LOGGER.problemCompletingOperations(storageManager.getContext());
-         }
-      } finally {
-         for (PageSubscription cursor : cursorList) {
-            cursor.enableAutoCleanup();
-         }
+      // Move every cursor to the next bookmarked page
```
michaelpearce-gain: how now are we making sure storage is done?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
```diff
@@ -530,11 +529,6 @@ void slowConsumerDetected(String sessionID,
    @Message(id = 222023, value = "problem cleaning page address {0}", format = Message.Format.MESSAGE_FORMAT)
    void problemCleaningPageAddress(@Cause Exception e, SimpleString address);
 
-   @LogMessage(level = Logger.Level.WARN)
-   @Message(id = 222024, value = "Could not complete operations on IO context {0}",
```
michaelpearce-gain: this just removes the log line we are seeing, isnt that now just masking the problem?

clebertsuconic: No... we are using asynchronously now.. before it was blocking.. no need to the handler.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
clebertsuconic: some place else would be already listening for it and the broker would shutdown on critical error.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
clebertsuconic: On another fix there was a loop removed here.. the thread is exausting the pool and causing the log.warn.


notice I'm also fixing the context used on cleanup.

michaelpearce-gain: just deployed to real server in testing env, seeing if issue of flow stopping is resolved. will let you know

michaelpearce-gain:
Just watching logs like a hawk atm whilst testing this, we're not using large messages, and likewise there is no ids, but with master + this pr, i see in the logs the below, which is spurious (i dont think its this PR, but something to check - and maybe fix, not sure what introduced it)

2020-03-12 13:54:30,271 INFO  [org.apache.activemq.artemis.core.server] AMQ224100: Timed out waiting for large messages deletion with IDs [], might not be deleted if broker crashes atm
2020-03-12 13:54:35,272 INFO  [org.apache.activemq.artemis.core.server] AMQ224100: Timed out waiting for large messages deletion with IDs [], might not be deleted if broker crashes atm

michaelpearce-gain:
fyi, this latest doesnt seem to fix the issue with the flow stopping, it seems from what i can tell is the producers get blocked

image

michaelpearce-gain:
happy to have a session for you to see first hand....

clebertsuconic:
@michaelandrepearce did you apply the other fix for AMQP Flow control? did you get master + this fix?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
clebertsuconic: I'm setting on a callback on the storage now.

michaelpearce-gain:
so the flowcontrol one is merged to master i saw, so i locally merged this fix with master and then built

clebertsuconic:
@michaelandrepearce 

```
2020-03-12 13:54:30,271 INFO [org.apache.activemq.artemis.core.server] AMQ224100: Timed out waiting for large messages deletion with IDs [], might not be deleted if broker crashes atm
```

This is another place doing the same thing I did before.. 

One thing to notice is you are starting to page.. and then depaging really fast.. why is that happening?

I can fix this blocking issue, but you should also look at that cause.

michaelpearce-gain:
so the simulation test, was a way we re-created what occured in prod, basically we had a massive volatilty in the market meaning alot of messages burst into the session, it took a we while for consumers to scale up causing to page, but then we had some blocked flow, the recreator basically is a way we've found to re-create the problem

clebertsuconic:
How can we talk offline?

michaelpearce-gain:
Do you have teams? hangouts? or bluejeans can do any

michaelpearce-gain:
sent a teams invite to your apache.org email

PR #3018, opened 2020-03-13
ARTEMIS-2655 support auto-creation w/FQQN & STOMP

jbertram:


michaelandrepearce:
How are we differentiating if it should be anycast or multicast queue?

jbertram:
I added a blurb to [the documentation](https://github.com/apache/activemq-artemis/blob/master/docs/user-manual/en/stomp.md#routing-semantics) not too long ago describing how control routing semantics with STOMP.

As far as the code goes, a `RoutingType` is passed to the `autoCreateDestinationIfPossible` method which is based on the frame's header or destination prefix. If that is `null` then it uses the default value specified in the address-settings.

michaelandrepearce:
Thanks.

michaelpearce-gain:
i think you can merge @jbertram its been 2 weeks an no other comments

PR #3019, opened 2020-03-16
ARTEMIS-2658 AMQP message read from page has wrong encode size

franz1981:


franz1981:
@clebertsuconic I'm not very proud of https://github.com/apache/activemq-artemis/pull/3019/commits/2105479304a7efead14ecb343686bf31042f1731#diff-d3a97ffe6cb7d4ab3fd0c4d24dea022bR100, but it seems covered by some existing tests eg `AmqpBridgeClusterRedistributionTest`. It should retain the nice performance of https://issues.apache.org/jira/browse/ARTEMIS-2617: I'm now running a test to verify it.
@tabish121 is aware of this issue ie duplicate decoding of  `address` : I'm thinking to create a `AMQPMessagePersisterV3` that save the encoding/decoding of `address` to happen twice, while "fixing" `AMQPMessagePersisterV2` trying to preserve the nice optimization on journal loading. wdyt? advices?

franz1981:
@clebertsuconic Seems that regardless my efforts of trying to preserve the perf optimization...this PR still have some perf issue see
image

franz1981:
@clebertsuconic The issue is more evident on master, so this PR should be fine...I'll open a JIRA to investigate the performance issue while loading the journal separately...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessagePersisterV2.java
```diff
@@ -72,25 +75,36 @@ public void encode(ActiveMQBuffer buffer, Message record) {
...
-         if (existingExtraProperties == null) {
-            extraProperties = new TypedProperties(Message.INTERNAL_PROPERTY_NAMES_PREDICATE);
-         }
-         extraProperties.decode(buffer.byteBuf(), pool != null ? pool.getPropertiesDecoderPools() : null, existingExtraProperties == null);
-         if (extraProperties != existingExtraProperties) {
-            message.setExtraProperties(extraProperties);
-         }
+         final TypedProperties extraProperties = record.createExtraProperties();
+         extraProperties.decode(buffer.byteBuf(), pool != null ? pool.getPropertiesDecoderPools() : null);
+         assert Objects.equals(address, extraProperties.getSimpleStringProperty(AMQPMessage.ADDRESS_PROPERTY)) :
```
clebertsuconic: Can't we just log.debug if something is wrong here?
log.warn...

i would rather have a big warning on the system, with the system somewhat working, than having AssertException and having the system freezing.

PR #3021, opened 2020-03-17
ARTEMIS-2662 Using previously stored encodeSize on page record offset

clebertsuconic:
There is no guarantee that the encodeSize size is the same in AMQP right after read.
    As the protocol may add additional bytes right after decoded such as header, extra properties.. etc.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
@@ -1357,7 +1357,7 @@ public final class PageSubscriptionImpl implements PageSubscription {
                   break;
                }
 
-               int nextFileOffset = message.getPosition().getFileOffset() == -1 ? -1 : message.getPosition().getFileOffset() + message.getPagedMessage().getEncodeSize() + Page.SIZE_RECORD;
+               int nextFileOffset = message.getPosition().getFileOffset() == -1 ? -1 : message.getPosition().getFileOffset() + message.getPagedMessage().getStoredSize() + Page.SIZE_RECORD;
                tmpPosition = new PagePositionAndFileOffset(nextFileOffset, message.getPosition());
 
                boolean valid = true;
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagedMessageImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagedMessageImpl.java
@@ -47,6 +47,8 @@ public class PagedMessageImpl implements PagedMessage {
 
    private long transactionID = 0;
 
+   private final int storedSize;
+
    private volatile StorageManager storageManager;
 
    public PagedMessageImpl(final Message message, final long[] queueIDs, final long transactionID) {
@@ -57,10 +59,22 @@ public class PagedMessageImpl implements PagedMessage {
    public PagedMessageImpl(final Message message, final long[] queueIDs) {
       this.queueIDs = queueIDs;
       this.message = message;
+      this.storedSize = 0;
    }
 
-   public PagedMessageImpl(StorageManager storageManager) {
+   public PagedMessageImpl(int storedSize, StorageManager storageManager) {
       this.storageManager = storageManager;
+      this.storedSize = storedSize;
+   }
+
+
+   @Override
+   public int getStoredSize() {
+      if (storedSize <= 0) {
+         return getEncodeSize();
+      } else {
+         return storedSize;
+      }
    }
 
    @Override
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageReaderTest.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageReaderTest.java
@@ -58,7 +58,7 @@ public class PageReaderTest extends ActiveMQTestBase {
             PagePosition pagePosition = new PagePositionImpl(10, i);
             pagedMessage = pageReader.getMessage(pagePosition);
          } else {
-            int nextFileOffset = pagedMessage == null ? -1 : offsets[i - 1] + pagedMessage.getEncodeSize() + Page.SIZE_RECORD;
+            int nextFileOffset = pagedMessage == null ? -1 : offsets[i - 1] + pagedMessage.getStoredSize() + Page.SIZE_RECORD;
             PagePositionAndFileOffset startPosition = new PagePositionAndFileOffset(nextFileOffset, new PagePositionImpl(10, i - 1));
             PagePosition pagePosition = startPosition.nextPagePostion();
             assertEquals("Message " + i + " has wrong offset", offsets[i], pagePosition.getFileOffset());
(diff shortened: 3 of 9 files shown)


clebertsuconic:
@franz1981 I was working with you on this. this is the issue we were running after.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/paging/AmqpPageReaderTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+      Assert.assertEquals(persister.getEncodeSize(message), buffer.writerIndex());
+
+      // the very first byte is the persisterID, we skip that since we are calling the Persister directly
+      buffer.readerIndex(1);
+      Message messageRead = persister.decode(buffer, null, null);
+
+      // The current persister does not guarantee the same encode size after loading
+      // and it does not need to
+      // Assert.assertEquals(persister.getEncodeSize(message), persister.getEncodeSize(messageRead));
```
franz1981: This test class should be cleaned up a bit

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -1357,7 +1357,7 @@ private PagedReference moveNext() {
                   break;
                }
 
-               int nextFileOffset = message.getPosition().getFileOffset() == -1 ? -1 : message.getPosition().getFileOffset() + message.getPagedMessage().getEncodeSize() + Page.SIZE_RECORD;
+               int nextFileOffset = message.getPosition().getFileOffset() == -1 ? -1 : message.getPosition().getFileOffset() + message.getPagedMessage().getStoredSize() + Page.SIZE_RECORD;
```
franz1981: I would encapsulate this computation as part of `PagedMessage` or `PagePosition` , possibly adding a test

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagedMessageImpl.java
```diff
@@ -59,8 +61,19 @@ public PagedMessageImpl(final Message message, final long[] queueIDs) {
...
-   public PagedMessageImpl(StorageManager storageManager) {
+   public PagedMessageImpl(int storedSize, StorageManager storageManager) {
       this.storageManager = storageManager;
+      this.storedSize = storedSize;
+   }
+
+
+   @Override
+   public int getStoredSize() {
+      if (storedSize <= 0) {
```
franz1981: Instead of using this logic, woudn't simpler to correctly initialize `storedSize` for both protocols: for core `== encodeSize`, for AMQP, using a different value?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagedMessageImpl.java
```diff
@@ -47,6 +47,8 @@
 
    private long transactionID = 0;
 
+   private int storedSize = 0;
```
franz1981: If it cannot change anymore, I would make it `final` instead

artemis-server/src/test/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageReaderTest.java
```diff
@@ -58,7 +58,7 @@ public void testPageReadMessage() throws Exception {
             PagePosition pagePosition = new PagePositionImpl(10, i);
             pagedMessage = pageReader.getMessage(pagePosition);
          } else {
-            int nextFileOffset = pagedMessage == null ? -1 : offsets[i - 1] + pagedMessage.getEncodeSize() + Page.SIZE_RECORD;
+            int nextFileOffset = pagedMessage == null ? -1 : offsets[i - 1] + pagedMessage.getStoredSize() + Page.SIZE_RECORD;
```
franz1981: I don't know why, but my IDEA isn't happy with this test, see
image
Although I can run it and it compiles correctly...:(

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/paging/AmqpPageReaderTest.java
clebertsuconic: I intended to keep it commented out, as it would fail if commented out.. nothing to be cleaned.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagedMessageImpl.java
clebertsuconic: I didn't think it was final.. I thought I was going to have a setter when I wrote this.. changing it.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
clebertsuconic: I agree, and I raised that with you. but this is not part of my change here though. this could be a different PR.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagedMessageImpl.java
clebertsuconic: I am not sure it's safe to do that in case the encodeSize changed between the contruction and when getStoredSize is used.

We can change this If you're sure it's safe. I wasn't bold enough.

artemis-server/src/test/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageReaderTest.java
clebertsuconic: PageReaderTest is from a different testsuite. IDEA may not like it.

clebertsuconic: It's not that simple to move this test. it uses a lot of package private methods.. moving would need to make them public or a bigger refactoring.

I had a version where I did this at some point. I would go for the simpler version now in case we ever need to cherry-pick to a point release.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagedMessageImpl.java
franz1981: > We can change this If you're sure it's safe. I wasn't bold enough.

Agree, I'm not sure of it as well (let me check around on the code)

artemis-server/src/test/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageReaderTest.java
franz1981: > PageReaderTest is from a different testsuite. IDEA may not like it.

What's strange is that I rememeber while doing it the same (basically same code) IDEA wasn't complaining that much, but is really a minor eh, can be ignored

PR #3022, opened 2020-03-17
ARTEMIS-2663 Add customizer support for the embedded web server

brusdev:


jbertram:
I thought this could use a test so I sent #3032.

PR #3023, opened 2020-03-17
ARTEMIS-2661 AMQP Journal loading is triggering reencode

franz1981:


franz1981:
I'm now running some performance check to see if the fix is sound: in case I would create a follow-up PR to do something similar for the Core protocol.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPStandardMessage.java
```diff
@@ -162,14 +162,7 @@ public void reloadPersistence(ActiveMQBuffer record, CoreMessageObjectPools pool
       // Message state is now that the underlying buffer is loaded, but the contents not yet scanned
       resetMessageData();
       modified = false;
-      messageDataScanned = RELOAD_PERSISTENCE;
```
clebertsuconic: really? that's the offending line? 👍

franz1981: from now on: there was already a method on AMQPMessage that will call reencode, if needed

PR #3024, opened 2020-03-17
ARTEMIS-2664 The prefetch size is exceeded after delivered acks

brusdev:
Decrement delivered acks before acquiring credits.

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
@@ -68,6 +68,7 @@ public class AMQConsumer {
 
    private int prefetchSize;
    private final AtomicInteger currentWindow;
+   private final AtomicInteger deliveredAcks;
    private long messagePullSequence = 0;
    private final AtomicReference<MessagePullHandler> messagePullHandler = new AtomicReference<>(null);
    //internal means we don't expose
@@ -87,6 +88,7 @@ public class AMQConsumer {
       this.scheduledPool = scheduledPool;
       this.prefetchSize = info.getPrefetchSize();
       this.currentWindow = new AtomicInteger(prefetchSize);
+      this.deliveredAcks = new AtomicInteger(0);
       if (prefetchSize == 0) {
          messagePullHandler.set(new MessagePullHandler());
       }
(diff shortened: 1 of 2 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/amq/JmsClientAckTest.java
```diff
@@ -132,6 +137,72 @@ public void testUnAckedMessageAreNotConsumedOnSessionClose() throws JMSException
...
+      Wait.assertEquals((long)messageCount, () -> queueControl.getMessagesAcknowledged(), 3000, 100);
+      Wait.assertEquals(0, () -> queueControl.getDeliveringCount(), 3000, 100);
+
+      // Reset the session.
+      session.close();
+      session = connection.createSession(false, Session.CLIENT_ACKNOWLEDGE);
+
+      // Attempt to Consume the message...
+      consumer = session.createConsumer(queue);
+      msg = consumer.receive(1000);
```
clebertsuconic: Please, don't use receive(timeout) if you expect to receive null... use receiveNoWait.

clebertsuconic: I am ammending myself.. you don't need to do anything.. just pointing you for future reference.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
```diff
@@ -305,8 +306,20 @@ public void acknowledge(MessageAck ack) throws Exception {
       List<MessageReference> ackList = serverConsumer.getDeliveringReferencesBasedOnProtocol(removeReferences, first, last);
 
       if (removeReferences && (ack.isIndividualAck() || ack.isStandardAck() || ack.isPoisonAck())) {
-         acquireCredit(ackList.size());
+         this.deliveredAcks.getAndUpdate(deliveredAcks -> {
+            if (deliveredAcks >= ackList.size()) {
+               return deliveredAcks - ackList.size();
+            }
+
+            acquireCredit(ackList.size() - deliveredAcks);
```
brusdev: It is dangerous to call acquireCredit inside the deliveredAcks update because it could be executed multiple times. I created the PR #3029 to fix this.

PR #3026, opened 2020-03-17
ARTEMIS-2662 Make Page to be resilient to invalid position

clebertsuconic:
in case of an invalid position on getMessage, it should instead of runtimeException
be more resilient and fix itself.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
@@ -1664,6 +1664,11 @@ public interface ActiveMQServerLogger extends BasicLogger {
       format = Message.Format.MESSAGE_FORMAT)
    void federationBindingsLookupError(@Cause Throwable e, SimpleString address);
 
+   @LogMessage(level = Logger.Level.WARN)
+   @Message(id = 222288, value = "Page {0}, message {1} could not be found on offset {2}, with starting message {3}. This represents a logic error or inconsistency on the data, and the system will try once again from the beggining of the page file.",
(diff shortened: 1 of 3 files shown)


franz1981:
The idea is great, but I would use a warn (maybe): it shouldn't really happen, but when it happens it means that something isn't behaving correctly...I have not a strong opinion on this eh

clebertsuconic:
I thought about using asserts... but I have a test that's testing for that condition, the assert would fail.

I am not sure about using log.warn here though.

clebertsuconic:
so, I thought about log.warn. .but I wasn't 100% sure I would want it


I will do it log.warn.

clebertsuconic:
@franz1981 since I don't have a strong opinion either.. I ammended to add log.warn

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
```diff
@@ -1664,6 +1664,11 @@ void slowConsumerDetected(String sessionID,
       format = Message.Format.MESSAGE_FORMAT)
    void federationBindingsLookupError(@Cause Throwable e, SimpleString address);
 
+   @LogMessage(level = Logger.Level.WARN)
+   @Message(id = 222288, value = "Page {0}, message {1} could not be found on offset starring {2}, with starting message {3}. This represents a logic error or inconsistency on the data, and the system will try once again from the beggining of the page file.",
```
wy96f: typo "starring"?

clebertsuconic: typo fixed. thanks

PR #3029, opened 2020-03-18
ARTEMIS-2664 Fix the credits acquiring.

brusdev:
Move acquireCredit outside the deliveredAcks update function, since it may be
re-applied when attempted updates fail.

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
@@ -68,7 +68,7 @@ public class AMQConsumer {
 
    private int prefetchSize;
    private final AtomicInteger currentWindow;
-   private final AtomicInteger deliveredAcks;
+   private int deliveredAcks;
    private long messagePullSequence = 0;
    private final AtomicReference<MessagePullHandler> messagePullHandler = new AtomicReference<>(null);
    //internal means we don't expose
@@ -88,7 +88,7 @@ public class AMQConsumer {
       this.scheduledPool = scheduledPool;
       this.prefetchSize = info.getPrefetchSize();
       this.currentWindow = new AtomicInteger(prefetchSize);
-      this.deliveredAcks = new AtomicInteger(0);
+      this.deliveredAcks = 0;
       if (prefetchSize == 0) {
          messagePullHandler.set(new MessagePullHandler());
       }
@@ -306,18 +306,15 @@ public class AMQConsumer {
       List<MessageReference> ackList = serverConsumer.getDeliveringReferencesBasedOnProtocol(removeReferences, first, last);
 
       if (removeReferences && (ack.isIndividualAck() || ack.isStandardAck() || ack.isPoisonAck())) {
-         this.deliveredAcks.getAndUpdate(deliveredAcks -> {
-            if (deliveredAcks >= ackList.size()) {
-               return deliveredAcks - ackList.size();
-            }
-
+         if (deliveredAcks < ackList.size()) {
             acquireCredit(ackList.size() - deliveredAcks);
-
-            return 0;
-         });
+            deliveredAcks = 0;
+         } else {
+            deliveredAcks -= ackList.size();
+         }
       } else {
          if (ack.isDeliveredAck()) {
-            this.deliveredAcks.addAndGet(ack.getMessageCount());
+            this.deliveredAcks += ack.getMessageCount();
          }
 
          acquireCredit(ack.getMessageCount());


artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
```diff
@@ -68,7 +68,7 @@
 
    private int prefetchSize;
    private final AtomicInteger currentWindow;
-   private final AtomicInteger deliveredAcks;
+   private int deliveredAcks;
```
michaelandrepearce: please leave this atomic, to avoid concurrency visibility issues with different threads.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
```diff
@@ -306,18 +306,15 @@ public void acknowledge(MessageAck ack) throws Exception {
...
-            return 0;
-         });
+            deliveredAcks = 0;
+         } else {
+            deliveredAcks -= ackList.size();
+         }
       } else {
          if (ack.isDeliveredAck()) {
-            this.deliveredAcks.addAndGet(ack.getMessageCount());
+            this.deliveredAcks += ack.getMessageCount();
```
michaelandrepearce: this isn't concurrent safe, if different threads modify. thus why it was atomic

brusdev: I reverted the AtomicInteger. I did not think the `acknowledge` method requires to be thread-safe.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
```diff
@@ -306,15 +306,12 @@ public void acknowledge(MessageAck ack) throws Exception {
...
 
       if (removeReferences && (ack.isIndividualAck() || ack.isStandardAck() || ack.isPoisonAck())) {
-         this.deliveredAcks.getAndUpdate(deliveredAcks -> {
-            if (deliveredAcks >= ackList.size()) {
-               return deliveredAcks - ackList.size();
-            }
-
-            acquireCredit(ackList.size() - deliveredAcks);
+         int previousDeliveredAcks = this.deliveredAcks.getAndUpdate(
+            deliveredAcks -> deliveredAcks > ackList.size() ? deliveredAcks - ackList.size() : 0);
```
michaelandrepearce: why not >= as was before?

michaelandrepearce: i guess it still result in same result as 0, just be interesting if there was any reason.

clebertsuconic: I guess the commit message is not clear enough.. I think the problem is the nubmer of times acquiredCredits is called?

I would be a bit more descriptive on the commit message if this is the issue.

brusdev: @michaelandrepearce before was `>=` to not call acquireCredit with 0.

brusdev: @michaelandrepearce maybe I could simplify this replacing the AtomicInteger with an `int`. Indeed `deliveredAcks` is used only by the `acknowledge` method and it is only executed by the `EpollEventLoop` thread bounded with the relative connection channel. WDYT?

brusdev: I created https://github.com/brusdev/activemq-artemis/tree/fix_delivered_acks_decrement_simplified to check the simplified version and I executed all openwire tests in integration-tests against it and LGTM.

michaelandrepearce: I would go with the least change. More changes increase risk of introducing  new or other bugs.

Using int isn't thread safe.

franz1981: The original code wasn't having any AtomicInteger declaration at all

michaelandrepearce: Id def be a -1 moving from atomic

franz1981: I am not sure is a good idea to developing it assuming a concurrent access that is proved to not be present: it means making the core more complex and the testing should account and stress that too. 
@brusdev has already verified that the code can be single-threaded so I vote for the simplest solution here.

franz1981: @michaelandrepearce FYI is @brusdev himself that has used AtomicInteger the first time there, that's why I insist that is ok to simplify it with the right threading model assumption...

franz1981: We can add some check too in form of assert to get the CI to fail I'd the assumption won't hold in the future.

michaelpearce-gain: My concern is not concurrent updates, but that updates can occur on different thread (possibly even different core) and thus the update isnt visible. 

Reading the comment it seems there has been made an assumption based on EpollEventLoop. What about KQueue or even if on windows, where neither is present? Are the guarantees the events come on same thread.

franz1981: The event loop executions re the same Netty channel always happen on the same Thread (is a strong assumption in Netty or FastThreadLocal wouldn't work) so as-single-threaded, visibility shouldn't be a concern due to it.

michaelandrepearce: As long as were 100%. Im just worried after past issues

franz1981: That's perfectly reasonable indeed, from any pov I agree :)
In general I often see ownership and threading model not always clear in our code base so I would like to simplify things when we have the chance to do it to make easier the life of future contributors (including ourselves in the future). Probably thinking on some check/assert is a nice addition to make this more robust over time...

brusdev: I replaced the AtomicInteger with an int. Indeed deliveredAcks is used only by
the acknowledge method and it is only executed by the EpollEventLoop thread
bounded with the relative connection channel.

michaelpearce-gain:
@franz1981 are you merging then?

PR #3031, opened 2020-03-18
ARTEMIS-2665 AMQP use createSharedQueue like Core

michaelandrepearce:


michaelandrepearce:
Ensure's the queue is marked temporary for AMQP as it is for CORE.

PR #3036, opened 2020-03-20
ARTEMIS-2669 not durable AMQP messages cannot became durable on depaging

franz1981:
Currently depaging AMQP not durable messages turn them into durable ones and reencoded again: this shouldn't really happen, hence the mentioned code should be deleted.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
@@ -836,15 +836,6 @@ public abstract class AMQPMessage extends RefCountMessage implements org.apache.
       scanMessageData();
       messageDataScanned = MessageDataScanningStatus.SCANNED.code;
       modified = false;
-
-      // Message state should reflect that is came from persistent storage which
-      // can happen when moved to a durable location.  We must re-encode here to
-      // avoid a subsequent redelivery from suddenly appearing with a durable header
-      // tag when the initial delivery did not.
-      if (!isDurable()) {
-         setDurable(true);
-         reencode();
-      }
    }
 
    @Override
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/paging/AmqpPagingTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/paging/AmqpPagingTest.java
@@ -16,9 +16,12 @@
  */
 package org.apache.activemq.artemis.tests.integration.amqp.paging;
 
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
 import org.apache.activemq.artemis.tests.integration.amqp.AmqpClientTestSupport;
@@ -30,8 +33,22 @@ import org.apache.activemq.transport.amqp.client.AmqpSender;
 import org.apache.activemq.transport.amqp.client.AmqpSession;
 import org.junit.Assert;
 import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
 
+@RunWith(Parameterized.class)
 public class AmqpPagingTest extends AmqpClientTestSupport {
+
+   @Parameterized.Parameters(name = "durability={0}")
+   public static Collection getParams() {
+      return Arrays.asList(new Object[][]{{Boolean.TRUE}, {Boolean.FALSE}, {null}});
+   }
+
+   private final Boolean durable;
+
+   public AmqpPagingTest(Boolean durable) {
+      this.durable = durable;
+   }
 
    @Override
    protected void addConfiguration(ActiveMQServer server) {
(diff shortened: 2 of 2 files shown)


franz1981:
I've opened this for duscussion: @clebertsuconic and @tabish121 mates you're invited at the party :)

franz1981:
The test is failing on purpose: it is showing something that I suppose is a weak point of the current behaviour ;)

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -837,16 +837,26 @@ protected synchronized void lazyScanAfterReloadPersistence() {
...
+      makeDurable();
+   }
+
+   private void makeDurable() {
+      assert messageDataScanned == MessageDataScanningStatus.SCANNED.code;
       // Message state should reflect that is came from persistent storage which
       // can happen when moved to a durable location.  We must re-encode here to
       // avoid a subsequent redelivery from suddenly appearing with a durable header
       // tag when the initial delivery did not.
       if (!isDurable()) {
```
clebertsuconic: Can't we just remove this?

franz1981: It would be the best possible scenario: just don't know if we have test coverage to be sure that it won't break anything...still need to investigate why we have this..idea?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPStandardMessage.java
```diff
@@ -190,6 +191,89 @@ public void reencode() {
       scanMessageData();
    }
 
+   private void reencodeOnDurabilityChangedHeaderAlreadyPresent() {
+      // only the durability property has changed: header size hasn't changed
```
tabish121: This method makes an assumption that adjusting the durable value and re-encoding the Header will not result in a size change but that in fact is not at all certain.  The encoding done by the sender cannot be guaranteed to be exactly how the version of proton-j the broker is using will result in the same exact size and structure.  The Header described type can be marked with either a Symbol or an UnsignedInteger which itself can either be encoded as a small uint (2 bytes) or a normal uint (5 bytes).  Same goes for the boolean values which can be one byte or two depending on the optimizations done in the source encoder.  Same again for the TTL and DeliveryCount fields which have compact and normal sized encodings.  If that happens you will likely underwrite and leave corrupt data in the Delivery payload as proton-j will be optimizing each value for size.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
tabish121: Removing would be a good option as the broker is currently altering the Delivery in a way that changes what the sender originally intended for it.  Any forwarding or resending from other intermediaries will see behavioural changes down the line because the Delivery shifted from non-durable to durable without (as far as I can tell) a valid reason for it doing so.

clebertsuconic: @franz1981 I will do the following: mark this as DO-NOT-MERGE-YET... if you could remove these lines... run the testsuite and see if anything breaks...



I believe this was written as in.. I feel like it should have been durable.. why not set it to durable.. as opposed to it's really breaking anything.

clebertsuconic: just completed a testsuite.. .and I see no need for this re-encoding.

I say lets get rid of this.

@tabish121 do you remember why you added this on a851a8f93f30972d252f2bff0bb3d5847cfd7b5f ?


if you're ok with removing it... it should be gone.

franz1981: Thanks guys for the help on this, highly appreciated!!! I am super happy ro follow the path that both have found and I can close this!! 
A code removed is a code we doesn't have to maintain anymore, yeeeeee

tabish121: The change I made was to stabilize what was already happening where on reload from persistence the broker tagged the durable flag in AMQPMessage as true and then depending on other factors in the message content and redelivery state it would sometimes pop out marked with a durable header and other times not.  I chose to preserve the behaviour that happened the most which was that the message reappeared with a re-encode that resulted in a Header being applied.

clebertsuconic: @franz1981 it's safe to remove it then. amend this PR to remove it.

franz1981: The label can be removed: now the PR is removing the legacy code and improving the test coverage to spot any regression on this behaviour :+1:

clebertsuconic:
@franz1981 this is ready to merge now, right?

franz1981:
@clebertsuconic Yep :) 
Now AMQP depage should be more correct and delivering a very nice performance & stability boost

PR #3037, opened 2020-03-20
removed flag ' -l qpidtypes' from proton-cpp example

1984shekhar:
With '-l qpidtypes', error "/usr/bin/ld: cannot find -lqpidtypes" is received in console while executing compile.sh script.

### DIFF
--- a/examples/protocols/amqp/proton-cpp/compile.sh
+++ b/examples/protocols/amqp/proton-cpp/compile.sh
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 # This requires g++ and qpid-cpp-client-devel
-g++ src/main/cpp/hello.cpp  -o hello -l qpid-proton-cpp -l qpidtypes
+g++ src/main/cpp/hello.cpp  -o hello -l qpid-proton-cpp


examples/protocols/amqp/proton-cpp/compile.sh
```diff
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 # This requires g++ and qpid-cpp-client-devel
-g++ src/main/cpp/hello.cpp  -o hello -l qpid-proton-cpp -l qpidtypes
```
clebertsuconic: can I understand why?

1984shekhar: Hi clebert,
I don't find any qpidtypes shared library after installing qpid-proton-cpp client.  Also I had discussion with Justin Ross on Red Hat messaging@redhat.com regarding this with subject "AMQ C++ Client". He suggested that qpidtypes should not be there.
List of available shared libraries:

[chandrashekhar@localhost lib64]$ pwd
/usr/lib64
[chandrashekhar@localhost lib64]$ ls -al|grep qpid
lrwxrwxrwx.   1 root root        25 Mar  3 00:58 libqpid-proton-core.so -> libqpid-proton-core.so.10
lrwxrwxrwx.   1 root root        29 Mar  3 00:58 libqpid-proton-core.so.10 -> libqpid-proton-core.so.10.8.1
-rwxr-xr-x.   1 root root    255408 Mar  3 00:58 libqpid-proton-core.so.10.8.1
lrwxrwxrwx.   1 root root        24 Mar  3 00:58 libqpid-proton-cpp.so -> libqpid-proton-cpp.so.12
lrwxrwxrwx.   1 root root        28 Mar  3 00:58 libqpid-proton-cpp.so.12 -> libqpid-proton-cpp.so.12.6.1
-rwxr-xr-x.   1 root root    421424 Mar  3 00:58 libqpid-proton-cpp.so.12.6.1
lrwxrwxrwx.   1 root root        28 Mar  3 00:58 libqpid-proton-proactor.so -> libqpid-proton-proactor.so.1
lrwxrwxrwx.   1 root root        32 Mar  3 00:58 libqpid-proton-proactor.so.1 -> libqpid-proton-proactor.so.1.5.1
-rwxr-xr-x.   1 root root     49288 Mar  3 00:58 libqpid-proton-proactor.so.1.5.1
lrwxrwxrwx.   1 root root        20 Mar  3 00:58 libqpid-proton.so -> libqpid-proton.so.11
lrwxrwxrwx.   1 root root        24 Mar  3 00:58 libqpid-proton.so.11 -> libqpid-proton.so.11.9.1
-rwxr-xr-x.   1 root root    368760 Mar  3 00:58 libqpid-proton.so.11.9.1
[chandrashekhar@localhost lib64]$

clebertsuconic: This is years old... it probably changed. thanks.. I will merge it.

1984shekhar: Thanks clebert

gemmellr: While C++ is not my thing, I believe the only time I have seen qpidtypes referred to was with the older 'Qpid Messsaging' C++ client, i.e. its never been used for Proton and so should indeed be removed.

It looks like the example used the other client previously, since its even mentioned in the script on the line right above the change here, it says it requires qpid-cpp-client-devel rather than qpid-proton-cpp-devel.

Actually, the example readme also still discusses the other client extensively in its 'steps'. Seems like most of that should just be removed as it doesn't reflect the example at all.

PR #3041, opened 2020-03-22
Two JIRAs around Paging and AMQP

clebertsuconic:
These two JIRAs are related but I wanted to keep them separate as commits.

Both changes were part of the same testing.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/ProcessorBase.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/ProcessorBase.java
@@ -22,7 +22,6 @@ import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
-import java.util.concurrent.locks.LockSupport;
 import java.util.function.Consumer;
 
 import org.jboss.logging.Logger;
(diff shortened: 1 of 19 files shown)


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/ProcessorBase.java
```diff
@@ -99,45 +98,23 @@ public void shutdown(long timeout, TimeUnit unit) {
...
-               //give other threads a chance to make progress
-               LockSupport.parkNanos(100_000_000L);
-            }
-         }
-         while (!stateUpdater.compareAndSet(this, STATE_NOT_RUNNING, STATE_FORCED_SHUTDOWN));
-         //this could happen just one time: the forced shutdown state is the last one and
-         //can be set by just one caller.
-         //As noted on the execute method there is a small chance that some tasks would be enqueued
+      if (!inHandler()) {
+         flush(1, TimeUnit.SECONDS);
```
franz1981: I would use some named const somewhere to help understand why is 1 second or I won't wait at all

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/ProcessorBase.java
```diff
@@ -99,45 +98,23 @@ public void shutdown(long timeout, TimeUnit unit) {
...
-      //there is a small chance that execute() could race with this cleanup: the lock allow an all-or-nothing behaviour between them
-      synchronized (tasks) {
-         T item;
-         while ((item = tasks.poll()) != null) {
-            onPendingItem.accept(item);
-            pendingItems++;
-         }
+
+      T item;
+      while ((item = tasks.poll()) != null) {
```
franz1981: the synchronize was there for 2 reasons: 

1. if we would have started using some multi-producer single-consume queue (the ones on JCTools for example...) 
2. onAddedTaskIfNotRunning is using a synchronize to allow all-or-nothing behaviour

artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContextTest.java
```diff
@@ -154,6 +160,8 @@ private void doOnMessageWithDeliveryException(List<Symbol> sourceSymbols,
 
       rc.onMessage(mockDelivery);
 
+      Thread.sleep(1000);
```
franz1981: that's safe in any env?

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
```diff
@@ -43,6 +44,8 @@
  */
 public class OperationContextImpl implements OperationContext {
 
+   private static final Logger logger = Logger.getLogger(OperationContextImpl.class);
```
brusdev: is this logger used?

artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContextTest.java
clebertsuconic: I forgot to remove this sleep

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/ProcessorBase.java
clebertsuconic: I was afraid of failures on the testsuite. In a production system the component would be going down anyways.


I will try to remove the condition for good.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/ProcessorBase.java
clebertsuconic: I am removing the clear on onAddedTaskifNotRunning

franz1981:
If the few comments are being addressed (when makes sense) I'm ok with this, can you merge it your self?

PR #3042, opened 2020-03-23
ARTEMIS-975: Add transactional records to deletedRecords to check for committed transactions that also hold references to

graben:
Especially needed for paging/depaging records that are hold in several transactions not getting cleaned up completely.

### DIFF
--- a/artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/journal/JDBCJournalImpl.java
+++ b/artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/journal/JDBCJournalImpl.java
@@ -232,6 +232,7 @@ public class JDBCJournalImpl extends AbstractJDBCDriver implements Journal {
                   holder = transactions.get(record.getTxId());
                   for (RecordInfo info : holder.recordsToDelete) {
                      deletedRecords.add(record.getId());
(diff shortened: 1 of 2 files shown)


jbertram:
This really needs a test in order to validate the fix and also to mitigate against regressions in the future.

artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/journal/JDBCJournalImpl.java
```diff
@@ -232,6 +232,7 @@ public synchronized int sync() {
                   holder = transactions.get(record.getTxId());
                   for (RecordInfo info : holder.recordsToDelete) {
                      deletedRecords.add(record.getId());
+                     deletedRecords.add(info.id);
```
clebertsuconic: without a test I can't really say what is broken.

graben:
I'll provide a test case.

graben:
@jbertram , @clebertsuconic 
I add a test case out of my trace log that simulates actual behavior of jdbc backend while writing messages to page and consuming from page. It shows that some records (the commit record!) are lost in db storage if used without patch and everything gets cleared completely with patch applied.

jbertram:
Nice work!

PR #3043, opened 2020-03-23
ARTEMIS-2682: OpenWireMessageConverter refactoring

fvaleri:
This is just a refactoring of the OpenWireMessageConverter with no semantics change and is coming out of #2954. Proposed change is similar to what we already have in the AMQP protocol module with a singleton class as the protocol converter entry point and two dedicated classes for conversion algorithms.


michaelandrepearce:
This requires a jira

clebertsuconic:
@michaelandrepearce I talked to @fvaleri and I had told him that if he's sure this is a no-semantic change we could have a NO-JIRA as being just a reorg on the code.

but if you see a need we can certainly issue a JIRA.

michaelpearce-gain:
Its not a small change tho, lets have a JIRA. If it was just a move of files or something, then it show as a git move, but it seems that its wholesale class changes...with that we should be careful as its very hard to cross validate before and after.... in git diff, having to manually review this currently.

michaelpearce-gain:
fyi, because of the amounts of changed lines, annoying it hasnt been done so it shows it as a move, and just the edited lines, this is taking some time to go through.

fvaleri:
@michaelpearce-gain there was a small error in a class rename, now fixed.

I completely understand that stability matter the most and it may take some time to be verified by another developer. I'm also open to discussion here. 

The whole point of refactoring is to alter the internal structure without changing its external behavior, so there shouldn't be new features or fixes. The reward in this case should be better maintainability, at least this is the intention.

Specifically, this is copying the AMQP protocol module layout, which splits the inbound and outbound converters in dedicated classes (they were already distinct but hosted on the same big class). In my opinion this is much more clean.

michaelandrepearce:
The main point is this needs a jira. Its not a small change

fvaleri:
> The main point is this needs a jira. Its not a small change

Agree and now you have it.

michaelpearce-gain:
its still not compiling.....even after your latest commit, i really am thinking is this PR worth it, if its only a refactor, but not benefit. 

The biggest thing for me in this space if someone was to refactor is to actually introduce Native OpenWireMessage support, so for openwire to openwire we dont convert. Then it be worth the risk/effort.

fvaleri:
Same reason as before, I hope the fix is pushed now and I also added an initial set of unit tests.

I think that a feature like native OpenWire message support would inevitably include some refactoring. If you are not sensible to better maintainability, then you could also take this as a first step in that direction.

michaelandrepearce:
This still isnt compiling.

michaelandrepearce:
I really now have reservations on this PR from a stability standpoint

clebertsuconic:
@michaelpearce-gain I actually rebased this locally and it's building fine. it's probably some intermittent issue on the CI.

Can you rebase it and push -f?


can you describe what you have moved here?


i'm trying to understand the gain of these changes.


I remember when i wrote this I couldn't put the OpenWire in the same classModel I did for AMQP. Is that what you accomplished here?


if so.. what exactly did you move? it's kind of hard to see if there was no changes in here. and the benefit of such change.

clebertsuconic:
These tests are good BTW. we should include them some way or the other.

michaelandrepearce:
@clebertsuconic i keep getting a compile error with this pr due to a classname not matching its file.

I will try  full rebase tomorrow if given time. That said i havent even started to review this as need to have it compile first.

Also with all the basic compile issues, and the amount of change which makes it hard to work out what actually had changed, this is a high risk pr that clearly can introduce an accidental issue (as example it hasnt been compiling)

clebertsuconic:
@michaelandrepearce try `git clean -df` after you checkout this branch.

I created a branch called forMichael on my fork, if that helps.

fvaleri:
Hi the build failure seems to be related to a Travis environment issue because I can run the same build command without issues.
```log
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

@clebertsuconic this is what I've done:

- removed `OpenwireMessage` bean because it was just an empty implementation and no other class was using it
- given that the inbound and outbound conversion logic was already distinct inside the  `OpenWireMessageConverter` class, I made it as Singleton that is now calling `OpenWireMessageConverter.createInboundMessage` and `CoreOpenWireConverter.createMessageDispatch`
- the conversion logic inside previous methods is exactly the same as it was, I just removed some minor code duplication (*)
- created `OpenWireMessageSupport` class to host some simple utility method and to group all property constants that are in common between `OpenWireMessageConverter` and `CoreOpenWireConverter`. This class also useful in unit tests.

(*) you can easily see my changes by comparing the following:
- `OpenWireMessageConverter.inbound` --> `OpenWireCoreConverter.createInboundMessage`
- `OpenWireMessageConverter.createMessageDispatch` --> `CoreOpenWireConverter.createMessageDispatch`

I know it's just a starter, but I'm really looking forward to your comments. I think the converter is much more manageable than before and, along with unit tests, it should allow us to be more confident in case of future changes.

clebertsuconic:
can you rebase this?

michaelandrepearce:
So i managed to get this to compile finally. 

I have to say the tests are a nice addition. 

That said having a single class (as was) for the conversion to and from core or two seperate is really negligible in difference and is more personal coding preference/ style of a coder.. it makes no difference in ability to maintain.

Im personally much more in favour of a single class containing the conversion logic as it keeps it clearly contained.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageSupport.java
```diff
@@ -0,0 +1,86 @@
...
+      }
+      Object object = coreMessage.getObjectProperty(propertyKey);
+      if (object == null || !(object instanceof byte[])) {
+         return null;
+      }
+      ByteSequence seq = new ByteSequence((byte[]) object);
+      return format.unmarshal(seq);
+   }
+
+   public static void setBinObjectProperty(WireFormat format,
```
michaelandrepearce, line 63: Bin? Please lets call it binary

fvaleri: Sure

fvaleri:
OK, changed Bin to Binary and rebased.

PR #3047, opened 2020-03-24
ARTEMIS-2680 use q name for addr if none specified via config

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/ConfigurationTest.java
```diff
@@ -47,6 +49,22 @@ public void testStartWithDuplicateQueues() throws Exception {
       }
    }
 
+   @Test
+   public void testQueueWithoutAddressName() throws Exception {
+      final SimpleString QUEUE_NAME = RandomUtil.randomSimpleString();
+      ActiveMQServer server = createServer(false, createDefaultInVMConfig());
+      try {
+         server.getConfiguration().addQueueConfiguration(new CoreQueueConfiguration().setName(QUEUE_NAME.toString()));
```
clebertsuconic: how would that look like on the xml?

michaelandrepearce:
Surely if a queue has no address then its invalid configuration, shouldnt the loading of the config error...

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/ConfigurationTest.java
jbertram: I don't think this would be possible via the XML. It's only possible by using the `CoreQueueConfiguration` object directly.

jbertram:
@michaelandrepearce, currently if you add a `CoreQueueConfiguration` object with no address to `Configuration` and start the broker the queue will fail to load with a WARN message. However, if you invoke one of the overloaded `createQueue` methods with effectively no address the queue will be created using the queue's name as the address name. The behavior from `createQueue` seems OK to me so I think using `CoreQueueConfiguration` should work the same.

To be clear, the use-case here is an embedded broker without XML configuration.

michaelpearce-gain:
isnt the overloaded createQueue a legacy of pre 2.x model, where queues and topics were modelled differently and thus to keep some compatibilty and for JMS server bits that were still about, the createQueue was more for JMS create a JMS queue where queue and address did match.

Now the question i would ask, is should you be able to make a queue without an address, to me it seems we shouldnt so if anything we should probably clean out that createQueue method that creates and address based on queueName , createQueue really must/should provide an address IMO.

jbertram:
JMS queues and JMS topics have always been modeled essentially the same way. Back in Artemis 1.x and in HornetQ we had a specific JMS management API and a specific core management API and even then in the core management API we would default the address name to the queue name if the address name was `null`. It wasn't a JMS thing. It was the chosen default behavior for core, and was even advertised [via JavaDoc](http://activemq.apache.org/components/artemis/documentation/javadocs/javadoc-1.5.3/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.html#deployQueue-java.lang.String-java.lang.String-java.lang.String-). I still think it's a reasonable behavior, and I also think changing it could potentially break existing users.

It's also worth noting that the overloaded `createQueue` method which does this is the one that is always ultimately called. In other words, it doesn't matter which `createQueue` method you invoke, if you pass an effectively `null` address name then the queue name will be used.

michaelpearce-gain:
fair enough, im not heavily against so go ahead and merge if others also happy, just really doesnt make much sense to have a queue but not to define an address.

clebertsuconic:
this needs to be rebased.

jbertram:
This is moot now that the `QueueConfiguration` work has been merged.

jbertram:
Actually, the test is still worth merging...

jbertram:
@michaelandrepearce, for what it's worth as I was going through the tests to update them to use the new `createQueue` API I noticed that the same address and queue name are used in hundreds of places so defaulting the address name to the queue name is a nice little usability tweak.

PR #3048, opened 2020-03-25
ARTEMIS-1975 Fixing LargeMessage encoding for replication

clebertsuconic:


clebertsuconic:
today I have sent two commits directly into master, as I thought they were obvious change that didn't require peer review.

in the process I broke org.apache.activemq.artemis.tests.integration.amqp.largemessages.AmqpReplicatedLargeMessageTest


this is fixing AmqpReplicatedLargeMessageTest


Instead of using a Thread local for the cache of the saved buffer, I'm storing the record on the Message.

However for replication, this may be broken due to multi-threading.

This is fixing that by adding reference counting.


I did some manual testing to make sure there are no leaks. I couldn't find a way to add a permanent test without being invasive in performance.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
```diff
@@ -126,6 +127,20 @@ public void closeLargeMessage() throws Exception {
       parsingData = null;
    }
 
+   public synchronized void releaseEncodedBuffer() {
```
franz1981: Why synchronized? We perform these operations concurrently? 
If the semantic of accesses is as-single-threaded, but is happening on different Threads one-by-one, is ok to remove the synchronization (and better, because better document that this class shouldn't be used for concurrent accesses)

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
```diff
@@ -126,6 +127,20 @@ public void closeLargeMessage() throws Exception {
...
+         temporaryBuffer = null;
+      }
+   }
+
+   public synchronized ByteBuf getSavedEncodeBuffer() {
+      if (temporaryBuffer == null) {
+         temporaryBuffer = PooledByteBufAllocator.DEFAULT.buffer(getEstimateSavedEncode());
+         saveEncoding(temporaryBuffer);
+      }
+      return temporaryBuffer.retain();
```
franz1981: retain is a costly operation, is it necessary? Or we just document that the saved encode buffer cannot be conserved but just immediately used?
I would document it anyway to save a user to forget to release it and cause memory leaks

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessagePersister.java
```diff
@@ -96,10 +91,11 @@ public void encode(ActiveMQBuffer buffer, Message record) {
          properties.encode(buffer.byteBuf());
       }
 
-      ByteBuf savedEncodeBuffer = getSavedEncodeBuffer(record);
+      ByteBuf savedEncodeBuffer = msgEncode.getSavedEncodeBuffer();
       buffer.writeBytes(savedEncodeBuffer, 0, savedEncodeBuffer.writerIndex());
-      savedEncodeBuffer.release();
-      msgEncode.temporaryBuffer = null;
+      msgEncode.releaseEncodedBuffer();
+      msgEncode.releaseEncodedBuffer(); // We need two releases here, one for the retain when the savedEncodeBuffer was called
```
franz1981: Having 2 separate releases is prone to future errors: I believe this can be simplified to make esplicit in the method names what's happening here

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
michaelandrepearce: Largemessage can be concurrent accessed and used. In case of topic with multiple queues. 

Remember how we had to fix loads of concurrent access issue on core message the other year

michaelandrepearce: Id rather this is sync. Dont need to be chasing such difficult issues again

franz1981: Yep, agree I just wanted be sure was a correctness requirement :)

clebertsuconic: This is about persistence.  Which will not happen concurrently unless it’s replication.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
clebertsuconic: Retain is just a ref counter increment.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessagePersister.java
clebertsuconic: I will make a new method to do a release(2)

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
clebertsuconic: So, I will keep the synchronize here.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
franz1981: I know but looking at the naming used on Netty I think is a good idea to communicate when something is to be relased after usage and when not, just to be less error prone...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessagePersister.java
franz1981: Just to write here an example: https://netty.io/4.1/api/io/netty/buffer/ByteBuf.html#readRetainedSlice-int-

clebertsuconic:
I'm just doing some testing.. please don't merge it yet.

clebertsuconic:
the issue I found is totally not related.. I'm sending a separate PR

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
clebertsuconic: I renamed the method. This is not much my style of coding. I have seen other code you've written and you would create a new method.  but I will submit and created the new method now.

pushed it ! :)

clebertsuconic:
do not merge again.. as I'm dealing with the method rename.

clebertsuconic:
done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessagePersister.java
franz1981: I'm still fighting to find some water :( 
I would like you to be happy about the naming anyway, impl-wire you can use https://netty.io/4.1/api/io/netty/util/ReferenceCounted.html#release-int- that would release the buffer the number of times you need (and can return a boolean if you need to do something right after too)

PR #3049, opened 2020-03-25
ARTEMIS-2681 timestamp not set on notif msgs

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/SecurityNotificationTest.java
```diff
@@ -155,6 +163,8 @@ public void testCONSUMER_CREATED() throws Exception {
       Assert.assertEquals("guest", notifications[0].getObjectProperty(ManagementHelper.HDR_VALIDATED_USER).toString());
       Assert.assertEquals(address.toString(), notifications[0].getObjectProperty(ManagementHelper.HDR_ADDRESS).toString());
       Assert.assertEquals(SimpleString.toSimpleString("unavailable"), notifications[0].getSimpleStringProperty(ManagementHelper.HDR_CERT_SUBJECT_DN));
+      Assert.assertTrue(notifications[0].getTimestamp() >= start);
+      Assert.assertTrue((Long) notifications[0].getObjectProperty(ManagementHelper.HDR_NOTIFICATION_TIMESTAMP) >= start);
```
clebertsuconic: shouldn't we assert for == .getTimestamp()? given the way it's implemented?

jbertram: Are you saying we should assert `getTimestamp() == start` or something else? Both timestamps should be the same and both should be *after* `start`. I'm just asserting the latter.

michaelpearce-gain: i think the point here, is shouldnt the test be check timestamp >= start, and check timestamp aloing with hdr notif timestamp field are the same.

current logic in test doesnt check both are the same.

jbertram: Fair enough. I updated the tests to assert both timestamps are equal.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/NotificationTest.java
```diff
@@ -79,13 +79,17 @@ public void testBINDING_ADDED() throws Exception {
...
+      long start = System.currentTimeMillis();
       session.createQueue(address, queue, durable);
 
       //the first message received will be for the address creation
       ClientMessage[] notifications = NotificationTest.consumeMessages(2, notifConsumer);
       Assert.assertEquals(BINDING_ADDED.toString(), notifications[1].getObjectProperty(ManagementHelper.HDR_NOTIFICATION_TYPE).toString());
       Assert.assertEquals(queue.toString(), notifications[1].getObjectProperty(ManagementHelper.HDR_ROUTING_NAME).toString());
       Assert.assertEquals(address.toString(), notifications[1].getObjectProperty(ManagementHelper.HDR_ADDRESS).toString());
+      Assert.assertTrue(notifications[1].getTimestamp() > start);
+      Assert.assertTrue((long) notifications[1].getObjectProperty(ManagementHelper.HDR_NOTIFICATION_TIMESTAMP) >= start);
```
franz1981: I would use Assert.assertThat and Matchers.greaterXXX so when it fails we have a more communicative error (that is just a lil' thing but I have introduced the core matchers of hamcrest for this reason some time ago)

clebertsuconic: I'm trying to understand what this means?? :) can you clarify?

What for anyways? it's just an assertion? I guess I'm not understanding you.

jbertram: It just means that the notification was sent *after* the binding was added. It's a simple validation of the timestamp.

PR #3052, opened 2020-04-01
ARTEMIS-2693 Improve log of starting acceptor errors

brusdev:
Add the log of starting acceptor errors to simplify the detection of the
failing acceptor in case of multiple acceptors.


michaelandrepearce:
please raise a jira

brusdev:
jira created https://issues.apache.org/jira/browse/ARTEMIS-2693

PR #3055, opened 2020-04-01
ARTEMIS-1975 Fix AMQP paged large message usage counter

franz1981:


franz1981:
I've noticed a thing while working on this PR on `PagedMessageImpl::decode`:
```java
      if (isLargeMessage) {
         int largeMessageHeaderSize = buffer.readInt();

         if (storageManager == null) {
            largeMessageLazyData = new byte[largeMessageHeaderSize];
            buffer.readBytes(largeMessageLazyData);
         } else {
```
the case where `storageManager == null` and is a core message; why `usageUp` isn't called?

PR #3057, opened 2020-04-01
ARTEMIS-2685 Not Block Netty Thread in any way for OpenWire

clebertsuconic:


franz1981:
I remember OpenWire was using some Netty mechanism to disable AutoRead on a channel: this change still left this mechanism to work fine?

clebertsuconic:
Nothing was changed on flowncontrol.

PR #3058, opened 2020-04-02
ARTEMIS-2686 Fix MQTT connect message rejection

brusdev:
Initialize the session state with a default value to fix a NPE, when an incoming
MQTT interceptor rejects a MqttConnectMessage.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/imported/MQTTRejectingInterceptorTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/imported/MQTTRejectingInterceptorTest.java
@@ -16,6 +16,9 @@
  */
 package org.apache.activemq.artemis.tests.integration.mqtt.imported;
 
+import java.util.concurrent.CountDownLatch;
+
+import io.netty.handler.codec.mqtt.MqttConnectMessage;
 import io.netty.handler.codec.mqtt.MqttMessage;
 import io.netty.handler.codec.mqtt.MqttPublishMessage;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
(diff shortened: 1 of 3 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/imported/MQTTRejectingInterceptorTest.java
```diff
@@ -61,4 +62,36 @@ public boolean intercept(MqttMessage packet, RemotingConnection connection) thro
...
+
+      Thread publishThread = new Thread(() -> {
+         MQTTClientProvider publishProvider = getMQTTClientProvider();
+
+         try{
+            initializeConnection(publishProvider);
+            publishProvider.disconnect();
+            fail("The connection should be rejected!");
+         } catch (Exception e) {
+            e.printStackTrace();
```
franz1981, line 85: If this is the expected behaviour you can just ignore the exception or we risk to read the test logs and think that there is some error!

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/imported/MQTTRejectingInterceptorTest.java
```diff
@@ -61,4 +62,36 @@ public boolean intercept(MqttMessage packet, RemotingConnection connection) thro
...
+
+      server.getRemotingService().addIncomingInterceptor((MQTTInterceptor) (packet, connection) -> {
+         if (packet.getClass() == MqttConnectMessage.class) {
+            return false;
+         } else {
+            return true;
+         }
+      });
+
+      Thread publishThread = new Thread(() -> {
```
franz1981, line 77: We really need to use a separate Thread for this?

brusdev: The MQTTClientProvider hasn't a connection timeout. I could remove the thread and use the test timeout but I would lose the fail message "The connection is stuck!".

brusdev:
@franz1981 I pushed some changes according to your suggestions.

PR #3060, opened 2020-04-03
ARTEMIS-2695 Return exception message in AMQP conversion error response

djc:


michaelandrepearce:
please create a jira for this change

clebertsuconic:
the commit should be part of your commit message, no the github title

on a command line:
```sh
git commit -a --amend
on the vim, change the title, and add the JIRA
git push origin -f <your-branch-name>
```

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
```diff
@@ -514,7 +514,7 @@ private static Section convertBody(ServerJMSMessage message, Map<Symbol, Object>
             }
          } catch (Throwable ignored) {
```
clebertsuconic, line 515: if you're really using the JIRa, rename the argument here as ``e`` or whatever other name you like, 
since it's not ignored any longer.

djc: Fair point! Fixed.

PR #3063, opened 2020-04-05
ARTEMIS-2692 refactor queue creation

jbertram:
- Deprecates existing overloaded createQueue, createSharedQueue,
  createTemporaryQueue, & updateQueue methods for ClientSession,
  ServerSession, ActiveMQServer, & ActiveMQServerControl where
  applicable.
- Deprecates QueueAttributes, QueueConfig, & CoreQueueConfiguration.
- Deprecates existing overloaded constructors for QueueImpl.
- Implements QueueConfiguration with JavaDoc to be the single,
  centralized configuration object for both client-side and broker-side
  queue creation including methods to convert to & from JSON for use in
  the management API.
- Implements new createQueue, createSharedQueue, createTemporaryQueue &
  updateQueue methods with JavaDoc for ClientSession, ServerSession,
  ActiveMQServer, & ActiveMQServerControl as well as a new constructor
  for QueueImpl all using the new QueueConfiguration object.
- Changes all tests and internal code to use the new methods.


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
```diff
@@ -298,6 +299,28 @@ public void createAddress(final SimpleString address, RoutingType routingType, b
...
+      startCall();
+      try {
+         sessionContext.createSharedQueue(queueConfiguration);
+      } finally {
+         endCall();
+      }
+   }
+
+   @Override
+   public void createTemporaryQueue(QueueConfiguration queueConfiguration) throws ActiveMQException {
```
Havret, line 320: I'm wondering, do we still need all these overloads like `createSharedQueue` and `createTemporaryQueue`. Shouldn't the intent be expressed by the state of `QueueConfiguration` object itself? As the client of the api, I wouldn't expect that the configuration I am passing will be mutated.

michaelandrepearce:
Could i suggest this PR is split in two. One with the refactor where tests remain untouched. And the once that is merged and a full build with all tests run not just the pr ones. Then a subsequent pr is raised to update tests. 

Reason for this is that this is a sizeable refactor and even human eye looking over may miss something. Where as we do expect if refactor all historic tests should pass without modification. By having both refactor and tests change its hard to get confidence.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
michaelandrepearce: I would agree, essentially a shared queue or temporary queue is just all attributes of a queue. As such having a more simplified single createQueue method.

jbertram: I'm in favor of that. I'll take those out to simplify it even further.

jbertram:
@michaelandrepearce, I think that's a great idea. I'll split it up and push ASAP.

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
```diff
@@ -357,9 +363,29 @@ private void slowPacketHandler(final Packet packet) {
                case CREATE_QUEUE_V2: {
                   CreateQueueMessage_V2 request = (CreateQueueMessage_V2) packet;
                   requiresResponse = request.isRequiresResponse();
-                  session.createQueue(request.getAddress(), request.getQueueName(), request.getRoutingType(), request.getFilterString(), request.isTemporary(), request.isDurable(), request.getMaxConsumers(), request.isPurgeOnNoConsumers(),
-                                      request.isExclusive(), request.isGroupRebalance(), request.getGroupBuckets(), request.getGroupFirstKey(), request.isLastValue(), request.getLastValueKey(), request.isNonDestructive(), request.getConsumersBeforeDispatch(), request.getDelayBeforeDispatch(),
-                                      request.isAutoDelete(), request.getAutoDeleteDelay(), request.getAutoDeleteMessageCount(), request.isAutoCreated(), request.getRingSize());
+                  session.createQueue(new QueueConfiguration(request.getQueueName())
```
franz1981, line 366: I would encapsulate this logic in a separate method on a separate class to make it more readeable an d shorter ie translation from `request` to `QueueConfiguration`

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
```diff
@@ -382,9 +411,25 @@ private void slowPacketHandler(final Packet packet) {
                   requiresResponse = request.isRequiresResponse();
                   QueueQueryResult result = session.executeQueueQuery(request.getQueueName());
                   if (!(result.isExists() && Objects.equals(result.getAddress(), request.getAddress()) && Objects.equals(result.getFilterString(), request.getFilterString()))) {
-                     session.createSharedQueue(request.getAddress(), request.getQueueName(), request.getRoutingType(), request.getFilterString(), request.isDurable(), request.getMaxConsumers(), request.isPurgeOnNoConsumers(),
-                                               request.isExclusive(), request.isGroupRebalance(), request.getGroupBuckets(), request.getGroupFirstKey(), request.isLastValue(), request.getLastValueKey(), request.isNonDestructive(), request.getConsumersBeforeDispatch(), request.getDelayBeforeDispatch(),
-                                               request.isAutoDelete(), request.getAutoDeleteDelay(), request.getAutoDeleteMessageCount());
+                     session.createSharedQueue(new QueueConfiguration(request.getQueueName())
```
franz1981, line 414: I would encapsulate this logic in a separate method on a separate class to make it more readeable an d shorter ie translation from request to QueueConfiguration

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
jbertram: Will do.

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
jbertram: Will do.

clebertsuconic:
@jbertram let me know when this is ready to be merged / reviewed?

I know you have been running tests on this.

jbertram:
I just pushed these changes:

- Broke the single commit into 2 commits - one for the broker code-base and one for the tests. Both commits pass the test-suite indicating no regressions (suggested by @michaelandrepearce).
- Removed `createTemporaryQueue` from the API completely. I did not remove `createSharedQueue` because that is going to require some wire-format changes and I didn't want to introduce those into these commits (suggested by @Havret).
- Encapsulated the logic to translate a few packets to a `QueueConfiguration` (suggested by @franz1981).

IMO this is ready to be merged.

michaelandrepearce:
Can we make the tests a seperate pr in itself. E.g. as suggested in dev thread almost we do one release without the tests changed, and then as soon as this release goes out the tests merge and then go in the next release. 

This way should there be any issues in release we know tests are not changed and avoid any goose chases.

jbertram:
@michaelandrepearce, I'm still fuzzy on the logic for putting the commits in different releases. The tests aren't part of the release anyway so I don't see why it really matters. Could you elaborate a bit further (perhaps with a contrived but plausible example)?

clebertsuconic:
-1 on the tests.  We should go all in.  Or wait the next release.

clebertsuconic:
I don’t understand why the commits are separated anyways ?

jbertram:
@clebertsuconic, I separated the commits and ran the full test-suite on both to (1) ensure there were no inadvertent semantic changes for existing users and (2) to make sure the new API was working as expected.

clebertsuconic:
@jbertram can you please add a test using the deprecated API?

I think it's fair that we move the testsuite to the new test, but we should still have a test validating the old API works.

clebertsuconic:
- can you treat the defaults on the QueueConfiguration?

it would be nice if these defaults to be treated directly there.

with that, you won't need to check if queue.isDurable() == null, you can just return true (or defaultValue from another class) if null, it would save you a lot of duplicated code?

jbertram:
I just pushed a few updates:

- Setting defaults in `QueueConfiguration` directly (suggested by @clebertsuconic).
- Modified `ActiveMQServerControlTest`, `SessionTest`, and `CreateQueueTest` to use both old a new API calls for creating queues to validate that the old API still works (suggested by @clebertsuconic).
- Added a `serialVersionUID` to `QueueConfiguration`.

michaelandrepearce:
Point to split changing the tests to be in a next release is that honestly i got to about 100ish classes reviewed and ill be honest it became more of a skim review, once i got off the core classes. So id say im 60 % confident in that there isnt an issue.

That said and reason for split pr / release is that we all know its not until its released and peoples qa teams run much larger test suites or deploy into pre prod environments that we find the edge cases. By keeping the tests the same makes it easier to check against previous version with a test case and make/edit existing test that shows a regression as same test would work on last release also.

That said I've run a full test suite with the first commit and all was good.

cshannon:
@jbertram - Nice work, this is a great chance going forward to simplify queue creation

michaelandrepearce:
@jbertram / @clebertsuconic There seems to be an issue, ive been looking at the lastest master, and testing it will some real world scenarios and it seems where previously methods may have been called without values previously, the address settings default would have been used, where as now it falls through as being null in queueconfig, and then just gets set to the static default in queue impl.

michaelandrepearce:
e.g. where say a piece of code used to call

   public Queue createQueue(final SimpleString address,
                            final RoutingType routingType,
                            final SimpleString queueName,
                            final SimpleString filterString,
                            final boolean durable,
                            final boolean temporary) throws Exception {

Basically all the other values would have been by delegation of each new variable, picked up the default from the address settings, but now it seems code is just calling createQueue(QueueConfiguration) with the values of other fields not being set, but no lookup on the address settings for non set values, as such later when it hits QueueImpl its just being set to the static default not the address setting default.

michaelandrepearce:
Easiest solution and safest would probably be to as the first line in the new method, get address settings for address, and then where value is null, set the value to the default at the address settings level.

   public Queue createQueue(final QueueConfiguration queueConfiguration, boolean ignoreIfExists) throws Exception {

michaelandrepearce:
@jbertram / @clebertsuconic ignore, this seems its being handled in QueueConfigurationUtils 
Ill have to dig a bit further then as why using latest master with an existing config set, the address settings maybe not being applied when previously there were.

PR #3064, opened 2020-04-05
NO-JIRA Document NO-JIRA use cases

brusdev:


clebertsuconic:
funny enough.. this one wouldn't need a JIRA. 


I'm not sure I would document things that are not part of the user's release note.

PR #3066, opened 2020-04-06
ARTEMIS-2699 Warn if queue stats are limited by default maxRows

brusdev:
Print a warning if the queues are greater than the max rows value.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
@@ -54,6 +54,8 @@ public class StatQueue extends AbstractAction {
       CONTAINS, EQUALS, GREATER_THAN, LESS_THAN
    }
 
+   public static final int DEFAULT_MAX_ROWS = 50;
+
    @Option(name = "--queueName", description = "display queue stats for queue(s) with names containing this string.")
    private String queueName;
 
@@ -67,7 +69,7 @@ public class StatQueue extends AbstractAction {
    private String value;
 
    @Option(name = "--maxRows", description = "max number of queues displayed. Default is 50.")
-   private int maxRows = 50;
+   private int maxRows = DEFAULT_MAX_ROWS;
 
    //easier for testing
    public StatQueue setQueueName(String queueName) {
@@ -114,11 +116,11 @@ public class StatQueue extends AbstractAction {
          context.out.println("filter is '" + filter + "'");
          context.out.println("maxRows='" + maxRows + "'");
       }
-      printStats(context, filter, maxRows);
+      printStats(context, filter);
       return null;
    }
 
-   private void printStats(final ActionContext context, final String filter, int maxRows) throws Exception {
+   private void printStats(final ActionContext context, final String filter) throws Exception {
       performCoreManagement(new ManagementCallback<ClientMessage>() {
          @Override
          public void setUpInvocation(ClientMessage message) throws Exception {
@@ -151,10 +153,15 @@ public class StatQueue extends AbstractAction {
       }
 
       JsonObject queuesAsJsonObject = JsonUtil.readJsonObject(result);
-      JsonArray array = (JsonArray) queuesAsJsonObject.get("data");
+      int count = queuesAsJsonObject.getInt("count");
+      JsonArray array = queuesAsJsonObject.getJsonArray("data");
 
       for (int i = 0; i < array.size(); i++) {
          printQueueStats(array.getJsonObject(i));
+      }
+
+      if (count > maxRows) {
+         context.out.println(String.format("WARNING: the displayed queues are %d/%d, set maxRows to display more queues.", maxRows, count));
       }
    }
 
--- a/artemis-cli/src/test/java/org/apache/activemq/cli/test/ArtemisTest.java
+++ b/artemis-cli/src/test/java/org/apache/activemq/cli/test/ArtemisTest.java
@@ -1243,8 +1243,8 @@ public class ArtemisTest extends CliTestBase {
          statQueue.setMaxRows(1);
          statQueue.execute(context);
          lines = getOutputLines(context, false);
-         // Header line + 1 queue only
-         Assert.assertEquals("rows returned by maxRows=1", 2, lines.size());
+         // Header line + 1 queue only + warning line
+         Assert.assertEquals("rows returned by maxRows=1", 3, lines.size());
 
       } finally {
          stopServer();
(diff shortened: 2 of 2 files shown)


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
```diff
@@ -67,7 +69,7 @@ String getJsonId() {
    private String value;
 
    @Option(name = "--maxRows", description = "max number of queues displayed. Default is 50.")
-   private int maxRows = 50;
+   private Integer maxRows;
```
michaelpearce-gain, line 72: why change this?

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
```diff
@@ -156,6 +160,10 @@ private void printStats(String result) {
       for (int i = 0; i < array.size(); i++) {
          printQueueStats(array.getJsonObject(i));
       }
+
+      if (maxRows == null && array.size() == DEFAULT_MAX_ROWS) {
```
michaelpearce-gain, line 164: first of all array.size() == DEFAULT_MAX_ROWS seems duff logic, surely if size > DEFAULT_MAX_ROWS is the logix expected based on the output below.

surely better to have this as if maxRows >= array.size() this way even if someone sets it to some other value, thats still smaller than array.size then you want to ouytput that the output has been limited.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
michaelpearce-gain: leave this as maxRows = DEFAULT_MAX_ROWS or 50./

brusdev: I would like to print the warning only if maxRows is not set by user.

michaelpearce-gain: but surely the user anyhow would like to know if more rows than they gave anyhow right?

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
brusdev: The maxRows should always be >=  array.size() but I would like to print the warning only if maxRows is not set by user and array.size() == DEFAULT_MAX_ROWS.
If you call `queue stat` without the parameter `maxRows` then you get at most 50 queues but you could not know that results are limited by the default maxRows value.
The purpose of the warning is only to help users that don't know that default maxRows value is 50.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
brusdev: I see your point but ATM the `listQueues(String options, int page, int pageSize)` return only maxRows queues. How can I help the users to know if there are more rows?

brusdev: I found a way using the attribute `count`. Let me try.

michaelpearce-gain: sounds positive :)

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
```diff
@@ -110,25 +112,27 @@ public Object execute(ActionContext context) throws Exception {
          return null;
       }
 
+      int printingMaxRows = this.maxRows != null ? this.maxRows : DEFAULT_MAX_ROWS;
+
       if (verbose) {
          context.out.println("filter is '" + filter + "'");
-         context.out.println("maxRows='" + maxRows + "'");
+         context.out.println("maxRows='" + printingMaxRows + "'");
```
michaelpearce-gain, line 119: why the need for printingMaxRows....

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
```diff
@@ -110,25 +112,27 @@ public Object execute(ActionContext context) throws Exception {
          return null;
       }
 
+      int printingMaxRows = this.maxRows != null ? this.maxRows : DEFAULT_MAX_ROWS;
```
michaelpearce-gain, line 115: this can revert now as you just use maxRows

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
```diff
@@ -151,11 +155,16 @@ private void printStats(String result) {
...
       JsonObject queuesAsJsonObject = JsonUtil.readJsonObject(result);
-      JsonArray array = (JsonArray) queuesAsJsonObject.get("data");
+      int count = queuesAsJsonObject.getInt("count");
+      JsonArray array = queuesAsJsonObject.getJsonArray("data");
 
       for (int i = 0; i < array.size(); i++) {
          printQueueStats(array.getJsonObject(i));
       }
+
+      if (maxRows == null && count > printingMaxRows) {
```
michaelpearce-gain, line 165: just count > maxRows

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
michaelpearce-gain: in your latest this can now just be reverted

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
brusdev: I would not to bore users with a warning if they set the maxRows parameter. WDYT?

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
brusdev: I only change the variable name. The maxRows was printed using verbose before my changes.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
brusdev: The maxRows is null if the user doesn't set the parameter. I need a way to undestand if the user set the parameter maxRows. I would print a warning only if the user doesn't set the parameter maxRows.

michaelpearce-gain: but why discriminate? surely just simplify the logic, because if i set 100 but count was 1001 it still be useful

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
michaelpearce-gain: i think its a) simpler to not discriminate b) even if i set 100 manually, it still useful to get an output to tell me whilst i set 100 it wasnt all the messages (and include the count in the output)

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
michaelpearce-gain: i was meaning no need for the printingMaxRows vairable, just keep it all back on single maxRows.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
brusdev: If you limit a sql query you get a limited set of record and you don't get a warning but if you don't limit a sql query you should get all matching records.
I think that the warning could break a script that is processing the queue stat output, so I would print the total available queues only in verbose mode if the user set the maxRows. WDYT?

michaelpearce-gain: this isnt a sql query though, its an admin tool, and like you said in sql if you dont limit you get all, thats not behaviour here, so its not the same at all.

michaelpearce-gain:
+1 looks good now.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
```diff
@@ -139,7 +141,7 @@ public void requestFailed(ClientMessage reply) throws Exception {
       });
    }
 
-   private void printStats(String result) {
+   private void printStats(String result, int maxRows) {
```
michaelpearce-gain, line 144: why do you need to pass this in? Is it not addressable at the object (class level var)

brusdev: To handle the case printStats will be called with a value different from the class level var, in the future.

brusdev: I agree atm its not needed but to be sure to not leave any broken path I'm going to remove the parameter `maxRows` from the signature `printStats(final ActionContext context, final String filter, int maxRows)` too.

PR #3067, opened 2020-04-06
ARTEMIS-2701 Improving DLQ/check over previously removed records

clebertsuconic:


clebertsuconic:
I added two operations to the journal...
tryUpdate
tryDelete


and used them on DLQ / Rollback


if something previously removed the message during the DLQ check.. we will just ignore the operation normally.

clebertsuconic:
@michaelandrepearce can you review this one?

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -1048,6 +1048,12 @@ public AMQSession getSession(SessionId sessionId) {
 
    public void removeDestination(ActiveMQDestination dest) throws Exception {
       if (dest.isQueue()) {
+
+         if (!dest.isTemporary()) {
+            // this should not really happen,
+            // so I'm not creating a Logger for this
+            logger.warn("OpenWire client sending a queue remove towards " + dest.getPhysicalName());
```
michaelandrepearce, line 1055: I dont get this change,  why shouldnt this happen?

PR #3068, opened 2020-04-06
ARTEMIS-2702 QuorumVoteServerConnect with requestToStayLive is voting order sensitive

franz1981:
QuorumVoteServerConnect with requestoToStayLive should perform the same correct decision regardless the order of a negative votes: this could affect the vote-on-replication-failure functionality if an unlucky timing of vote would happen.


franz1981:
I still need to validate the behavior at an higher level of abstraction ie

- ServerConnectVoteHandler unit test
- with an end 2 end test

And run the whole test suite to see if this isn't causing regressions (especially on `QuorumVoteServerConnect::await` that doesn't have any test coverage and I'm not sure of the original semantic. Probably it can create issues on the live when `vote-on-replication-failure` is set and we're awaiting the votes to complete.

franz1981:
I've sent a second commit that (assuming that I've correctly understood the original logic) simplify the current logic. @clebertsuconic @jbertram @andytaylor please take a look at the 2 commits separately, if I've misunderstood how `QuorumVoteServerConnect` is supposed to work...

In addition: I see that we should check for `targetNodeId` vs `ServerConnectVote::getNodeId` on `QuorumVoteServerConnect::vote` too (probably).

franz1981:
I've decided to merge the 2 commits into one, hardening the expected behaviour after solving some test failures.

franz1981:
I need to take care of a couple of log messages on the test and It seems ok (looking at the CI test too)

PR #3069, opened 2020-04-07
ARTEMIS-2703 Update commons-configuration2 version to 2.7

brusdev:


artemis-features/src/main/resources/features.xml
```diff
@@ -63,6 +63,7 @@
 		<configfile finalname="etc/org.apache.activemq.artemis.cfg">mvn:org.apache.activemq/artemis-features/${pom.version}/cfg</configfile>
 		<configfile finalname="etc/artemis.xml">mvn:org.apache.activemq/artemis-features/${pom.version}/xml/artemis</configfile>
 
+		<bundle dependency="true">mvn:javax.mail/mail/1.4.7</bundle>
```
clebertsuconic, line 66: javax.mail??/ what's tat?

brusdev: commons-configuration2 2.7 has a new Import-Package in the manifest: javax.mail.internet

clebertsuconic: but we certainly don't expose java mail... is there any way we don't include this? It's weird.


I can foresee people asking how to email messages with artemis.

michaelandrepearce: Lmao agreed , im a pmc and committer and i was even thinking if i missed that our broker had email feature by this.

jbonofre: The problem is on commons-configuration: javax.mail should be an optional import. With updating to commons-configuration 2.7, we have to add a bundle exporting javax.mail package. However, as commons-configuration doesn't define any version range, I would use javax.mail 1.6.2 (recent one) instead of 1.4.7.

brusdev: @jbonofre thanks for your clarification.
If I replace the bundle `mvn:javax.mail/mail/1.4.7` with `mvn:javax.mail/javax.mail-api/1.6.2` but I get the missing requirement error for the import package: `com.sun.mail.util`.
If I replace the bundle `mvn:javax.mail/mail/1.4.7` with `mvn:com.sun.mail/javax.mail/1.6.2` I don't get any missing requirement error. WDYT?

clebertsuconic: if you can't get around.. .add a big comment on why it had to be added... at least we will know.
say we don't use it but for some misconfiguration on the component, we were forced to import it in OSGI

PR #3073, opened 2020-04-08
[ARTEMIS-2704]: Provide a SPI to manage and cache SSLContext.

ehsavoie:
* Adding a new SPI to allow for SSLContext reuse accross the broker.
 * Providing a default gehaviour similar to the existing one.

Jira: https://issues.apache.org/jira/browse/ARTEMIS-2704

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/DefaultSSLContextFactory.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/DefaultSSLContextFactory.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2020 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.core.remoting.impl.ssl;
+
+import java.util.Map;
+import javax.net.ssl.SSLContext;
+import org.apache.activemq.artemis.core.remoting.impl.netty.TransportConstants;
+import org.apache.activemq.artemis.spi.core.remoting.ssl.SSLContextFactory;
+import org.apache.activemq.artemis.utils.ConfigurationHelper;
+
+/**
+ * Simple SSLContextFactory for use in NettyConnector and NettyAcceptor.
+ */
+public class DefaultSSLContextFactory implements SSLContextFactory {
+
+
+   @Override
+   public SSLContext getSSLContext(Map<String, Object> configuration,
+           String keystoreProvider, String keystorePath, String keystorePassword,
+           String truststoreProvider, String truststorePath, String truststorePassword,
+           String crlPath, String trustManagerFactoryPlugin, boolean trustAll) throws Exception {
+      return createSSLContext(configuration,
+              keystoreProvider, keystorePath, keystorePassword,
+              truststoreProvider, truststorePath, truststorePassword,
+              crlPath, trustManagerFactoryPlugin, trustAll);
+   }
+
+   protected SSLContext createSSLContext(Map<String, Object> configuration,
+           String keystoreProvider, String keystorePath, String keystorePassword,
+           String truststoreProvider, String truststorePath, String truststorePassword,
+           String crlPath, String trustManagerFactoryPlugin, boolean trustAll) throws Exception {
+      if (log.isDebugEnabled()) {
+         final StringBuilder builder = new StringBuilder();
+         configuration.forEach((k, v) -> builder.append("\r\n").append(k).append("=").append(v));
+         log.debugf("Creating SSL context with configuration %s", builder.toString());
+      }
+      boolean useDefaultSslContext = ConfigurationHelper.getBooleanProperty(TransportConstants.USE_DEFAULT_SSL_CONTEXT_PROP_NAME, TransportConstants.DEFAULT_USE_DEFAULT_SSL_CONTEXT, configuration);
+      if (useDefaultSslContext) {
+         return SSLContext.getDefault();
+      }
+      return new SSLSupport()
+              .setKeystoreProvider(keystoreProvider)
+              .setKeystorePath(keystorePath)
+              .setKeystorePassword(keystorePassword)
+              .setTruststoreProvider(truststoreProvider)
+              .setTruststorePath(truststorePath)
+              .setTruststorePassword(truststorePassword)
+              .setTrustAll(trustAll)
+              .setCrlPath(crlPath)
+              .setTrustManagerFactoryPlugin(trustManagerFactoryPlugin)
+              .createContext();
+   }
+   @Override
+   public int getPriority() {
+      return 5;
+   }
+}
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/remoting/ssl/SSLContextFactory.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/remoting/ssl/SSLContextFactory.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2020 The Apache Software Foundation.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.spi.core.remoting.ssl;
+
+import java.util.Map;
+import javax.net.ssl.SSLContext;
+import org.jboss.logging.Logger;
+
+/**
+ * Service interface to create a SSLContext for a configuration.
+ * This is NOT used by OpenSSL.
+ * To create and use your own implementation you need to create a file
+ * <code>META-INF/services/org.apache.activemq.artemis.spi.core.remoting.ssl.SSLContextFactory</code>
+ * in your jar and fill it with the full qualified name of your implementation.
+ */
+public interface SSLContextFactory extends Comparable<SSLContextFactory> {
+   Logger log = Logger.getLogger(SSLContextFactory.class);
+
+   /**
+    * Obtain a SSLContext from the configuration.
+    * @param configuration
+    * @param keystoreProvider
+    * @param keystorePath
+    * @param keystorePassword
+    * @param truststoreProvider
+    * @param truststorePath
+    * @param truststorePassword
+    * @param crlPath
+    * @param trustManagerFactoryPlugin
+    * @param trustAll
+    * @return a SSLContext instance.
+    * @throws Exception
+    */
+   SSLContext getSSLContext(Map<String, Object> configuration,
+           String keystoreProvider, String keystorePath, String keystorePassword,
+           String truststoreProvider, String truststorePath, String truststorePassword,
+           String crlPath, String trustManagerFactoryPlugin, boolean trustAll) throws Exception;
+
+   default void clearSSLContexts() {
+   }
+
+   /**
+    * The priority for the SSLContextFactory when resolving the service to get the implementation.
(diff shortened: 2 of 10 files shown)


clebertsuconic:
This PR is breaking the testsuite pretty badly.

e.g .CoreClientOverOneWaySSLTest

clebertsuconic:
still failing:

e.g: org.apache.activemq.artemis.tests.integration.ssl.CoreClientOverTwoWayOpenSSLServerTest.testTwoWaySSLVerifyClientTrustAllTrue[storeType=JCEKS]

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/DefaultSSLContextFactory.java
```diff
@@ -0,0 +1,70 @@
...
+              keystoreProvider, keystorePath, keystorePassword,
+              truststoreProvider, truststorePath, truststorePassword,
+              crlPath, trustManagerFactoryPlugin, trustAll);
+   }
+
+   protected SSLContext createSSLContext(Map<String, Object> configuration,
+           String keystoreProvider, String keystorePath, String keystorePassword,
+           String truststoreProvider, String truststorePath, String truststorePassword,
+           String crlPath, String trustManagerFactoryPlugin, boolean trustAll) throws Exception {
+      if (LOGGER.isDebugEnabled()) {
```
clebertsuconic: You should probably squash both commits.

like this commit here wouldn't compile as you didn't declare LOGGER, and the next one you remove this LOGGER.


As a matter of fact, I would call it log just like everywhere else, but that would be too much OCD on my side.

clebertsuconic:
I think this is in good shape now.. But can we squash these commits into one?


Something like:


```
    [ARTEMIS-2704]: Provide a SPI to manage and cache SSLContext.
    [ARTEMIS-2718]: Take advantage of ARTEMIS-2704 to cache SSLContexts.
    
     * Adding a new SPI to allow for SSLContext reuse accross the broker.
     * Providing a default behaviour similar to the existing one.
     * Adding a cache for SSLContexts and reusing them accross acceptors and
      connectors.
```

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/DefaultSSLContextFactory.java
```diff
@@ -0,0 +1,68 @@
...
+      if (!SSL_CONTEXTS.containsKey(sslContextName)) {
+         SSL_CONTEXTS.put(sslContextName, createSSLContext(configuration,
+              keystoreProvider, keystorePath, keystorePassword,
+              truststoreProvider, truststorePath, truststorePassword,
+              crlPath, trustManagerFactoryPlugin, trustAll));
+      }
+      return SSL_CONTEXTS.get(sslContextName);
+   }
+
+   protected String getSSLContextName(Map<String, Object> configuration) {
```
jbertram, line 52: Instead of passing the `configuration` map and then potentially extracting the keystore and truststore path from it why not simply pass through the `keystorePath` and `truststorePath` which were passed in to `getSSLContext` (i.e. the caller)?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/SslContextFactory.java
```diff
@@ -0,0 +1,52 @@
...
+import io.netty.handler.ssl.SslContext;
+import java.util.Map;
+import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
+import org.apache.activemq.artemis.core.remoting.impl.netty.TransportConstants;
+import org.apache.activemq.artemis.utils.ConfigurationHelper;
+
+/**
+ * Simple class to build a SslContext from a configuration.
+ */
+public class SslContextFactory {
```
jbertram, line 27: This class doesn't appear to be used anywhere? If it's not actually used anywhere it should be removed. Otherwise it should probably have a different name since there is already a class named `SSLContextFactory` with different capitalization and in a different package.

artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/remoting/ssl/SSLContextFactory.java
```diff
@@ -0,0 +1,57 @@
...
+    */
+   SSLContext getSSLContext(Map<String, Object> configuration,
+           String keystoreProvider, String keystorePath, String keystorePassword,
+           String truststoreProvider, String truststorePath, String truststorePassword,
+           String crlPath, String trustManagerFactoryPlugin, boolean trustAll) throws Exception;
+
+   default void clearSSLContexts() {
+   }
+
+   int getPriority();
```
jbertram, line 51: What is priority used for? Some JavaDocs might help clarify this.

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
```diff
@@ -594,21 +596,13 @@ public synchronized SslHandler getSslHandler(ByteBufAllocator alloc, String peer
    private SSLEngine loadJdkSslEngine(String peerHost, int peerPort) throws Exception {
```
jbertram, line 596: How come there aren't corresponding changes in `loadOpenSslEngine`?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/BasicSSLContextFactory.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.activemq.artemis.core.remoting.impl.netty.TransportConstants;
+import org.apache.activemq.artemis.spi.core.remoting.ssl.SSLContextFactory;
+import org.apache.activemq.artemis.utils.ConfigurationHelper;
+
+/**
+ * Simple SSLContextFactory for use in NettyConnector and NettyAcceptor.
+ */
+public class BasicSSLContextFactory implements SSLContextFactory {
+
+   private static final Map<String, SSLContext> SSL_CONTEXTS = Collections.synchronizedMap(new HashMap<>());
```
jbertram, line 31: `SSL_CONTEXTS` isn't actually used anywhere in this class. It should be removed.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/DefaultSSLContextFactory.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import javax.net.ssl.SSLContext;
+import org.apache.activemq.artemis.core.remoting.impl.netty.TransportConstants;
+import org.apache.activemq.artemis.utils.ConfigurationHelper;
+
+/**
+ *  Default SSLContextFactory providing a cache using the SSL_CONTEXT_PROP_NAME value as key.
+ */
+public class DefaultSSLContextFactory extends BasicSSLContextFactory {
```
jbertram, line 28: I think should be renamed to something like `CachingSSLContextFactory` to differentiate it from other `SSLContextFactory` implementations and to clarify its semantics.

jbertram:
The commit message says it's, "Providing a default behaviour similar to the existing one." However, the default `SSLContextFactory` (i.e. `org.apache.activemq.artemis.core.remoting.impl.ssl.DefaultSSLContextFactory`) implements caching which is not similar to the existing implementation (which doesn't use any caching).

jbertram:
Can you provide a brief statement about why this change is necessary? If it's for performance reasons do you have any benchmark data you can share?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/DefaultSSLContextFactory.java
ehsavoie: yes sorry this is a leftover

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/SslContextFactory.java
ehsavoie: the goal was to provide netty SslContext but this is not required

ehsavoie:
@jbertram the initial idea was to be able to inject SSLContext already created. Also I noticed that we were using the same code in acceptor and connector and that we were creating the same context  several times. I don't have any numbers to back this.

jbertram:
This looks better, but I still have a few questions...

1. The function of `priority` of an `SSLContextFactory` is still unclear to me. Will an `SSLContextFactory` with a mathematically larger `priority` be used in lieu of an `SSLContextFactory` with a mathematically lower `priority`? What's the purpose of defining multiple `SSLContextFactory` in the first place? 
2. Why does the `org.apache.activemq.artemis.spi.core.remoting.ssl.SSLContextFactory` file loaded by `SSLContextFactoryProvider` have 2 implementations now?
3. Why are you only using `SSLContextFactoryProvider` for the JDK provider and not the OpenSSL provider?

ehsavoie:
1. Yes, I'll try to make the Javadoc clearer: the goal is to be able to add your implementation with a jar on the classpath to replace the existing implementation.
2. to list the 2 available implementations
3. Because for OpneSSL you are returning a SslContext and not a SSLContext and I'm not sure about sharing a SslContext with several threads

ehsavoie:
@jbertram I've updated the javadoc and only defined a single service implementation in the org.apache.activemq.artemis.spi.core.remoting.ssl.SSLContextFactory file

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
ehsavoie: Because they don't produce a SSLContext but a SslContext

artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/remoting/ssl/SSLContextFactory.java
```diff
@@ -0,0 +1,62 @@
...
+           String keystoreProvider, String keystorePath, String keystorePassword,
+           String truststoreProvider, String truststorePath, String truststorePassword,
+           String crlPath, String trustManagerFactoryPlugin, boolean trustAll) throws Exception;
+
+   default void clearSSLContexts() {
+   }
+
+   /**
+    * The priority for the SSLContextFactory when resolving the service to get the implementation.
+    * This is used when selecting the implementation when several implemetnations are loadeds.
```
jbertram, line 53: Spelling/typo issues for "implemetnations" and "loadeds"

jbertram: The JavaDoc still doesn't explain the semantics of priority. For example, if I wanted to select impl `A` instead of impl `B` and `B` had a priority of `5` how should I set the priority of `A`? Would it need to be greater than `5` or less than `5`?

artemis-core-client/src/main/resources/META-INF/services/org.apache.activemq.artemis.spi.core.remoting.ssl.SSLContextFactory
```diff
@@ -0,0 +1 @@
+org.apache.activemq.artemis.core.remoting.impl.ssl.CachingSSLContextFactory
```
jbertram, line 1: This is changing the default behavior of the broker. I'm not sure we want to do that in a minor release.

jbertram:
Aside from the line-items I just noted I think the final thing necessary here is some documentation that outlines what this is for, how to implement your own factory, how to choose between the versions we ship (i.e. caching vs. the default non-caching), that it doesn't work from the OpenSSL provider, etc.

artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/remoting/ssl/SSLContextFactory.java
```diff
@@ -0,0 +1,67 @@
...
+           String keystoreProvider, String keystorePath, String keystorePassword,
+           String truststoreProvider, String truststorePath, String truststorePassword,
+           String crlPath, String trustManagerFactoryPlugin, boolean trustAll) throws Exception;
+
+   default void clearSSLContexts() {
+   }
+
+   /**
+    * The priority for the SSLContextFactory when resolving the service to get the implementation.
+    * This is used when selecting the implementation when several implementations are loadeds.
```
jbertram, line 57: Spelling issue with *loadeds*.

ehsavoie:
@jbertram done :)

jbertram:
There are still 2 checkstyle violations:
```
[ERROR] src/main/java/org/apache/activemq/artemis/spi/core/remoting/ssl/SSLContextFactory.java:[25] (regexp) RegexpSingleline: Line has trailing spaces.

[ERROR] src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/CachingSSLContextFactory.java:[29] (regexp) RegexpSingleline: Line has trailing spaces.
```
Also, there's still no documentation.

ehsavoie:
@jbertram should be ok with some doc

PR #3074, opened 2020-04-08
ARTEMIS-2653 Update proton-j to version 0.33.4

tabish121:
Update proton to latest point release


clebertsuconic:
looked at the commit list, and there's nothing that I think it would affect us..

merging it as soon as the build is finished to make sure we won't hit any maven caching issues.

PR #3075, opened 2020-04-10
Fixing failures on RamDisk due to Java API Files.getFileStore() 

madhukar-git:
Added a dummy implementation of FileStore object. Its a workaround to handle the exception thrown from org.apache.activemq.artemis.core.server.files.FileStoreMonitor.addStore(File)  when running on a RAMDISK machine or similar virtual file system


clebertsuconic:
You need a JIRA,
you need a test.

I don't understand even what's being fixed. without a test it makes a bit more difficult.


I would also squash your committs into one, with the JIRA mentioned. look at the hacking guide please:

https://github.com/apache/activemq-artemis/tree/master/docs/hacking-guide/en

or the formatted version through the webiste:

https://activemq.apache.org/components/artemis/documentation/

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/files/FileStoreMonitor.java
```diff
@@ -73,8 +75,8 @@ public FileStoreMonitor addStore(File file) throws IOException {
             try {
                addStore(Files.getFileStore(file.toPath()));
             } catch (IOException e) {
-               logger.error("Error getting file store for " + file.getAbsolutePath(), e);
-               throw e;
+               logger.warn("Error getting file store for " + file.getAbsolutePath() + ". Trying dummy FileStore", e);
```
michaelandrepearce, line 78: This will mean now real filesystems where we need to trust the store has been successful will silently fail. E.g. if permission or actual io issue with disk

michaelandrepearce: You will need to differentiate file systems if youre doing this as majority will be using normal disk filestores and expecting error if unsuccessful

franz1981:
I don't understand the use case: with RAM disk you mean Linux `tmpfs`? 
With it, NIO and Mapped should work fine AFAIK.

madhukar-git:
> I don't understand the use case: with RAM disk you mean Linux `tmpfs`?
> With it, NIO and Mapped should work fine AFAIK.

RAM disk file system stores its files in system RAM. ImDisk Toolkit is an example of RAM disk virtual file system.

madhukar-git:
> Added a dummy implementation of FileStore object. Its a workaround to handle the exception thrown from org.apache.activemq.artemis.core.server.files.FileStoreMonitor.addStore(File) when running on a RAMDISK machine or similar virtual file system


Regardless of your java version, we get below exception while executing the Java statement java.nio.file.Files.getFileStore(path) //where “path” points to RAM disk drive named R:\ .

java.nio.file.FileSystemException: R:\: Incorrect function.
       at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
       at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
       at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
       at java.base/sun.nio.fs.WindowsFileStore.create(WindowsFileStore.java:94)
       at java.base/sun.nio.fs.WindowsFileSystemProvider.getFileStore(WindowsFileSystemProvider.java:487)
       at java.base/java.nio.file.Files.getFileStore(Files.java:1490)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/files/FileStoreMonitor.java
madhukar-git: Thanks for the valuable feedback. We recently upgraded from Wildfly 10 to 17. This patch was intended to fix  our product regressions which run on RAM disk.

PR #3076, opened 2020-04-10
Updating text casing on connectionTtl

gamedev8:
Updating the text from `connectionTtl` to `connectionTTL` 

I actually copied and pasted `connectionTtl` in my configuration and spent time trying to figure out what was wrong. I finally realized it was the casing, so I'm proposing this change just in case someone faces the same issue.

### DIFF
--- a/docs/user-manual/en/connection-ttl.md
+++ b/docs/user-manual/en/connection-ttl.md
@@ -95,7 +95,7 @@ closing it down. If the server doesn't receive any packets on a connection
 for the connection TTL time, then it will automatically close all the
 sessions on the server that relate to that connection.
 
-The connection TTL is configured on the URI using the `connectionTtl`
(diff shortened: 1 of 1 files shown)


docs/user-manual/en/connection-ttl.md
```diff
@@ -186,4 +186,4 @@ thread adds a little more latency. These packets are:
 
 To disable asynchronous connection execution, set the parameter
 `async-connection-execution-enabled` in `broker.xml` to
-`false` (default value is `true`).
```
clebertsuconic, line 189: what was changed on this line? I can't see it.

clebertsuconic: NM: I see .. an enter was removed.. thanks

gamedev8: That is weird. I'll remove it.

PR #3078, opened 2020-04-13
ARTEMIS-2708 JDK bug causes missed props reload

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/ReloadableProperties.java
```diff
@@ -137,7 +137,14 @@ private void load(final File source, Properties props) throws IOException {
...
 
    private boolean hasModificationAfter(long reloadTime) {
-      return key.file.lastModified() > reloadTime;
+      /**
+       * A bug in JDK 8/9 (i.e. https://bugs.openjdk.java.net/browse/JDK-8177809) causes java.io.File.lastModified() to
+       * lose resolution past 1 second. Because of this, the value returned by java.io.File.lastModified() can appear to
+       * be smaller than it actually is which can cause the broker to miss reloading the properties if the modification
+       * happens close to another "reload" event (e.g. initial loading). In order to *not* miss file modifications that
+       * need to be reloaded we artificially inflate the value returned by java.io.File.lastModified() by 1 second.
+       */
```
michaelandrepearce, line 146: Just a note here wont this break once the jdk bug is fixed? Is this fixed in 11?

Is it worth adding some check to jdk version? Before doing the workaround so only do workaround on broken jdk versions

jbertram: No, it won't break. The worst that will happen is that the file might be reloaded a few more times than necessary which will take more time relatively speaking but really shouldn't be a big deal.

PR #3082, opened 2020-04-14
ARTEMIS-2706 Discounting Transfer payload size

clebertsuconic:


tabish121:
Tested here and frames are now written in a reasonable size distribution

gemmellr:
yep, much improved. A variant of the issue will still present for clients that have a lower max frame size than the brokers outgoing limit (the smaller it is, the less noticable it would actually be).

PR #3083, opened 2020-04-15
ARTEMIS-2714 handle 'Address already in use' better

jbertram:
Instead of blowing up with a stack-trace when hitting the common
'Address already in use' the broker will now log an ERROR with the
salient details of the acceptor and keep going.


michaelandrepearce:
This is dangerous no? as user if broker starts up you expect acceptors to have started fine. By swallowing do we run risk of broker starting up now with the acceptor not binding to port.

michaelandrepearce:
E.g. if the broker cannot bind i would want it to fail

jbertram:
I suppose whether or not is preferable to fail broker initialization completely or to keep going with an `ERROR` message depends on your use-case, but it is a semantic change so I'll revert that bit. The main goal here is to log more details since a broker can have multiple acceptors and users are currently left to just guess which one failed.

clebertsuconic:
nice one!

The semantic change was actually a reason I hadn't merged this one before...
I was going to talk to you but then got side tracked!!!


Merging now.

PR #3085, opened 2020-04-15
ARTEMIS-2692 update dependency for JMS client

jbertram:


jbertram:
@clebertsuconic, this should probably be merged before the release.

artemis-jms-client/pom.xml
```diff
@@ -50,6 +50,11 @@
          <artifactId>artemis-core-client</artifactId>
          <version>${project.version}</version>
       </dependency>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-commons</artifactId>
```
michaelandrepearce, line 55: Why is this needed it comes as a transitive anyhow

jbertram: Yes, it comes in as transitive, but `org.apache.activemq.artemis.jms.client.ActiveMQDestination` uses `org.apache.activemq.artemis.api.core.QueueConfiguration` (from `artemins-commons`) explicitly so IMO the dependency should be explicit as well instead of leaving it as transitive.

michaelandrepearce: Fair enough +1

PR #3088, opened 2020-04-17
ARTEMIS-2720 Optimize HierarchicalObjectRepository

jbertram:


jbertram:
I just had an idea about an alternate approach that might be more flexible. Closing this while I investigate...

PR #3089, opened 2020-04-17
ARTEMIS-2720 Optimize HierarchicalObjectRepository

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/HierarchicalObjectRepository.java
```diff
@@ -172,7 +172,12 @@ private void addMatch(final String match, final T value, final boolean immutable
       lock.writeLock().lock();
       try {
          String modifiedMatch = matchModifier.modify(match);
-         clearCache();
+         // an exact match (i.e. one without wildcards) won't impact any other matches so no need to clear the cache
+         if (usesWildcards(modifiedMatch)) {
```
clebertsuconic, line 183: This is tempting, but I don't think this is correct.

you may be removing an entry that was part of another wildcard. The cache has to be cleared on the other side as well.


I will add a test that would be broken by your change here.

clebertsuconic: I see.. I thought the other way around. we actually cache the result, not the wildcards.

so, this is fine.


I will merge a test I wrote as part of this.

PR #3090, opened 2020-04-17
ARTEMIS-2721 Activation keeps retrying even after server.stop()

clebertsuconic:
This won't be an issue on a real server (Production system)

however, on the testsuite or while embedded this could cause issues,
if a same instance is stopped then started.

This is the reason why BackupAuthenticationTest was intermittently failing.

I also adapted the test since I would need to stop the server and reactivate it in order to change the configuration.
The previous test wasn't acting like a real server.


clebertsuconic:
ran the whole testsuite. .and it's passing.

PR #3092, opened 2020-04-18
ARTEMIS-2723 Read the default CLI connector from the related broker

brusdev:
Read the CLI connector from the related broker instance if it isn't set by user.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
@@ -21,16 +21,29 @@ import javax.jms.Connection;
 import javax.jms.ConnectionFactory;
 import javax.jms.JMSException;
 import javax.jms.JMSSecurityException;
+import java.io.File;
+import java.net.InetAddress;
+import java.net.URI;
+import java.util.Map;
 
 import io.airlift.airline.Option;
+import org.apache.activemq.artemis.api.core.TransportConfiguration;
+import org.apache.activemq.artemis.cli.commands.ActionContext;
 import org.apache.activemq.artemis.cli.commands.InputAbstract;
+import org.apache.activemq.artemis.core.config.FileDeploymentManager;
+import org.apache.activemq.artemis.core.config.impl.FileConfiguration;
+import org.apache.activemq.artemis.core.remoting.impl.netty.TransportConstants;
 import org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory;
+import org.apache.activemq.artemis.utils.ConfigurationHelper;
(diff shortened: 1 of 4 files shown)


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
```diff
@@ -79,6 +95,67 @@ public void setProtocol(String protocol) {
...
+            for (TransportConfiguration acceptorConfiguration: fileConfiguration.getAcceptorConfigurations()) {
+               Map<String, Object> acceptorParams = acceptorConfiguration.getParams();
+               String protocols = ConfigurationHelper.getStringProperty(TransportConstants.PROTOCOLS_PROP_NAME, null, acceptorParams).toUpperCase();
+
+               if (protocols == null || protocols.contains(protocol.toUpperCase())) {
+                  String scheme = ConfigurationHelper.getStringProperty(TransportConstants.SCHEME_PROP_NAME, SchemaConstants.TCP, acceptorParams);
+                  String host = ConfigurationHelper.getStringProperty(TransportConstants.HOST_PROP_NAME, "localhost", acceptorParams);
+                  int port = ConfigurationHelper.getIntProperty(TransportConstants.PORT_PROP_NAME, 61616, acceptorParams);
+
+                  if (scheme.equals(SchemaConstants.VM)) {
```
clebertsuconic, line 139: why don't you do something simple?

Look for broker.xml on the etc.

look for the acceptor named "artemis"

```<acceptor name="artemis">tcp://0.0.0.0:61616?tcpSendBufferSize=1048576;tcpReceiveBufferSize=1048576;amqpMinLargeMessageSize=102400;protocols=CORE,AMQP,STOMP,HORNETQ,MQTT,OPENWIRE;useEpoll=true;amqpCredits=1000;amqpLowCredits=300;amqpDuplicateDetection=true</acceptor```


if you can't find it, just use the DEFAULT and the user will have to be more ellaborative on the cli options.


no reason to complicate things here IMO.

clebertsuconic: For instance.. there' s nothing the client CLI can do with a inVM acceptor. that's only valid for embedded cases, which is definitely not the case for the CLI.

clebertsuconic: if anything failed.. like you can't find the broker.xml on the ETC.. just use the default option.

That will be true for the cases where you use the CLI from the main distribution. as a matter of fact you should test the options outside of the broker instance.

brusdev: I pushed a new commit to simply.

PR #3094, opened 2020-04-20
ARTEMIS-2717 Isolate the current Quorum Vote implementation

franz1981:
This PR is part of the effort of providing a new pluggable QuorumVote implementation and aims to first isolate and refactor any "leak"  of the quorum vote logic outside `QuorumManager`.
This is not yet ready to be merged and is open to suggestions.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/ClusterControl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/ClusterControl.java
@@ -16,6 +16,8 @@
  */
 package org.apache.activemq.artemis.core.server.cluster;
 
+import java.util.Optional;
+
 import org.apache.activemq.artemis.api.core.ActiveMQException;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.api.core.TransportConfiguration;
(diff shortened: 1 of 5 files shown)


franz1981:
@michaelandrepearce @brusdev 
In the current state, the only services exposed by `QuorumManager` are:

- `registerQuorum`/`unRegisterQuorum`: to register/unregister a listener for `nodeDown`/`nodeUP` events issued by `ClusterController::defaultLocator`. The high-level function, looking at `SharedNothingBackupQuorumm`, isto detect a change in the live role (ie live is going down)
- `isStillLive`: used by the live with `vote-on-replication-failure` to validate that is still live
- `hasLive`: used by the backup (on `SharedNothingBackupQuorumm`) to check, after received a disconnection failure or a `nodeDown` event (see `registerQuorum`/`unRegisterQuorum`), if there is any live in charge. 
- `handleQuorumVote`: to manage `QuorumVoteMessage` packets, perform a vote and send back a response. It's tied to `ClusterController` (again)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/ClusterControl.java
```diff
@@ -64,6 +62,10 @@ public ClusterControl(ClientSessionFactoryInternal sessionFactory, ActiveMQServe
       this.clusterPassword = server.getConfiguration().getClusterPassword();
    }
 
+   public Optional<Channel> getClusterChannel() {
```
franz1981, line 65: @michaelandrepearce @brusdev Not proud to have "exposed" the lower level `Channel` here, but it seems the simplest way to move away he many  `Quorum*` deps away from here

brusdev:
Nice work. I have only one concern about `vote` used by `ColocatedActivation`.

clebertsuconic:
I will merge this..

But please keep an eye on the testsuite.

PR #3102, opened 2020-04-27
ARTEMIS-2738 implement per-acceptor security domains

jbertram:


michaelpearce-gain:
How does this map into existing simple, roles.properties and user.properties is there a new security_domains.properties or something?

michaelpearce-gain:
I think the documents need a bit of work, its not very clear to me how to use this, how it works etc and how it maps to JAAS?

jbertram:
> How does this map into existing simple, roles.properties and user.properties is there a new security_domains.properties or something?

I'm not entirely sure what you're asking. Everything maps like it always has. There's no new properties files or anything like that.

> I think the documents need a bit of work, its not very clear to me how to use this, how it works etc and how it maps to JAAS?

The document is pretty simple because the configuration and function is simple. Just specify the domain you want to use on your your acceptor using the `securityDomain` URL parameter. Then any client using that acceptor will be secured with that domain. It maps to JAAS just like the standard stuff. The specified `securityDomain` overrides the broker's configuration which includes the certificate domain.

clebertsuconic:
What tests did you add for AMQP? I wasn't sure I understood correctly,  but it feels like the CoreSession one you added could been Parameterized to use AMQP / Core.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/SecurityPerAcceptorTest.java
```diff
@@ -0,0 +1,244 @@
...
+   @Override
+   @Before
+   public void setUp() throws Exception {
+      super.setUp();
+
+      locator = createNettyNonHALocator();
+   }
+
+   @Test
+   public void testJAASSecurityManagerAuthentication() throws Exception {
```
clebertsuconic, line 67: Yes.. on a second look, you are missing tests for AMQP, Core, OpenWire, MQTT, Stomp.

You should add at least for AMQP, Core, OpenWire... where you pass the parameter of the connection Factory. Using CFUtil.createCF, and passing the protocol as a @parameter on the test.

Given the changes you made are actually affecting these classses I think it would be important to add tests for at least  AMQP, Core and openWire.

jbertram: Agreed.

clebertsuconic:
@jbertram there is one real failures on your PR:

org.apache.activemq.artemis.tests.integration.server.SecureConfigurationTest.testSecureDurableSubscriber[2: protocol=OPENWIRE]

clebertsuconic:
@jbertram as a matter of fact you can probably extend this test for your use?

artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/protocol/AbstractProtocolManager.java
```diff
@@ -50,13 +52,15 @@ protected String invokeInterceptors(final List<I> interceptors, final P message,
 
    @Override
    public void setAnycastPrefix(String anycastPrefix) {
+      ActiveMQServerLogger.LOGGER.info("Setting a prefix", new Exception());
```
clebertsuconic, line 55: Remove this please.. it looks like Debug you added for your own benefit and it was forgotten here.

I actually got a big exception on the broker I jsut started with your PR checked out, and I thought it had an error.

jbertram:
@clebertsuconic, I added some more tests & removed the debug stuff. I couldn't get `org.apache.activemq.artemis.tests.integration.server.SecureConfigurationTest#testSecureDurableSubscriber` to fail even though I ran it on repeat for awhile.

clebertsuconic:
Ok, I am about to start a testsuite on my box anyways.. I will bring your commit with mine and if everything is ok I will merge it.

clebertsuconic:
LGTM.. merging

PR #3104, opened 2020-04-29
ARTEMIS-2742 Warn complete XA transaction on the wrong session

brusdev:
Add warnings for XA transaction completed on a session different from the one
on which they started, making these situations easier to investigate.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/xa/BasicXaTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/xa/BasicXaTest.java
@@ -44,6 +44,7 @@ import org.apache.activemq.artemis.api.core.RoutingType;
 import org.apache.activemq.artemis.core.server.Queue;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
 import org.apache.activemq.artemis.core.transaction.impl.XidImpl;
+import org.apache.activemq.artemis.logs.AssertionLoggerHandler;
 import org.apache.activemq.artemis.ra.ActiveMQRAXAResource;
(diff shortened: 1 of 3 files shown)


clebertsuconic:
I am not sure this is not a WARN at all... it happens all the time...


-1000


this is supposed to happen that way... the TransactionManager will commit the Transaction from its own session in a lot of instances.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/xa/BasicXaTest.java
```diff
@@ -472,6 +473,96 @@ public void testRollbackError() throws Exception {
...
+      ClientSession clientSession2 = sessionFactory.createSession(true, false, false);
+
+      Xid xid = newXID();
+      startXATransaction(xid);
+
+      AssertionLoggerHandler.startCapture();
+      try {
+         clientSession2.prepare(xid);
+
+         assertTrue("Expected to find AMQ224105", AssertionLoggerHandler.findText("AMQ224105"));
```
clebertsuconic, line 505: You cannot make this as a rule...

clebertsuconic:
If you have a case where an user must have such thing. The best you can do, is to add a System.property to enable such behavior as a DEBUG tool.

Something like -Dorg.apache.activemq.artemis.core.server.impl.ServerSession.XA_EXPECT_SINGLE_SESSION = true,

And only then have such behavior working.


if this is something you initiated without an user asking... just forget about it

clebertsuconic:
@brusdev are you addressing this? you should probably close this until you can figure out how to properly get the TX ID out of the Transaction.

brusdev:
@clebertsuconic I'll close it

PR #3107, opened 2020-04-30
ARTEMIS-2648 - audit logging improvements

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-2648

A description of what I have done can be found on the Jira. Id like however some feedback on something, where I have added new logging I have removed the log message that was previously in the same place so where  there was a log message for creating a queue I have added a success and failure log. If anyone thinks this is a big deal i can add back the old messages but this will mean there are multiple logs potentially occurring.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/EventHandler.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/EventHandler.java
@@ -91,4 +91,7 @@ public interface EventHandler {
       return true;
    }
 
+   default String getRemoteAddress() {
+      return "";
+   }
 }
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/ProtonHandler.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/ProtonHandler.java
@@ -28,6 +28,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.channel.EventLoop;
 import org.apache.activemq.artemis.api.core.ActiveMQSecurityException;
+import org.apache.activemq.artemis.logs.AuditLogger;
 import org.apache.activemq.artemis.protocol.amqp.proton.AMQPConnectionContext;
 import org.apache.activemq.artemis.protocol.amqp.proton.ProtonInitializable;
 import org.apache.activemq.artemis.protocol.amqp.sasl.ClientSASL;
@@ -492,6 +493,11 @@ public class ProtonHandler extends ProtonInitializable implements SaslListener {
       }
       try {
          inDispatch = true;
+         if (AuditLogger.isAnyLoggingEnabled()) {
+            for (EventHandler h : handlers) {
+               AuditLogger.setRemoteAddress(h.getRemoteAddress());
+            }
+         }
          while ((ev = collector.peek()) != null) {
             for (EventHandler h : handlers) {
                if (log.isTraceEnabled()) {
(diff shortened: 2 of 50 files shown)


clebertsuconic:
ArtemisFeatureTest fails as we talked about.

just using the tag to avoid someone to merge it before it's fixed.


let me know when it's updated and I will remove the tag.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/EventHandler.java
```diff
@@ -91,4 +91,7 @@ default boolean flowControl(ReadyListener readyListener) {
       return true;
    }
 
+   default String getRemoteAddress() {
```
clebertsuconic, line 94: You don't need this. you're using AMQPConnection directly, no need to change to change EventHandler.

Can you remove it?

clebertsuconic: actually, you do use it...

But you are calling this on every proton event. isn't there a better place to call that?

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -287,6 +288,9 @@ public void bufferReceived(Object connectionID, ActiveMQBuffer buffer) {
          if (logger.isTraceEnabled()) {
             traceBufferReceived(connectionID, command);
          }
+         if (AuditLogger.isAnyLoggingEnabled()) {
+            AuditLogger.setRemoteAddress(getRemoteAddress());
```
clebertsuconic, line 292: this won't work...

openWireActor.act() will transfer the execution to a different thread.

you will lose the ThreadLocal here.

The netty loop is reused for many connections.

clebertsuconic: this could eventually fail in other protocols as well. you would endup informing a wrong remote address for Audit Operations.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/EventHandler.java
andytaylor: It is on every event so it is available to the Audit logger

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
andytaylor: ok, I will address

clebertsuconic: Just an idea. We have OperationContext, which will hold operations to the Session, and it has a thread local already.

Why don't you somehow inject the session on the operation context and use the thread local from there. We recover it back whenever we do a new operation on the session. It is a reliable place.

clebertsuconic: I mean.. why don't you inject the data you need on the operation context.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/ProtonHandler.java
```diff
@@ -497,6 +498,9 @@ private void dispatch() {
                if (log.isTraceEnabled()) {
                   log.trace("Handling " + ev + " towards " + h);
                }
+               if (AuditLogger.isAnyLoggingEnabled()) {
```
clebertsuconic, line 501: Do you need to do this for every event? It seems you only need to this only once.. so I think you could do this at the beggining of the method, before the loop starts.

clebertsuconic: there are no test failures on your branch now.. we just need to resolve the thread local versus thread changes due to asynchronous operations on the broker.

clebertsuconic:
this branch has no test failures now. We just need to address asynchronous operations and the thread local that was needed.

PR #3108, opened 2020-04-30
SecrectKeySpec contant key update

YYTVicky:


michaelandrepearce:
-1 as pee previous please add another implementation of sensitivsstringcodec.

jbertram:
The reason a constant key is used is so that the encode & decode will actually work. Keep in mind that this is just used for password *masking* so the broker has to be able to encode *and* decode the the values which, as far as I understand, require using the same key. If you generate a new key every time then you won't be able to decode anything. 

Also, If you want something more secure then Blowfish then please implement your own codec as [described in the documentation](http://activemq.apache.org/components/artemis/documentation/latest/masking-passwords.html#choosing-a-codec-for-password-masking).

jbertram:
FWIW, the [aforementioned documentation](http://activemq.apache.org/components/artemis/documentation/latest/masking-passwords.html#choosing-a-codec-for-password-masking) says that the default codec uses a constant key:

> Whenever no codec is specified in the configuration, the default codec is used. The class name for the default codec is `org.apache.activemq.artemis.utils.DefaultSensitiveStringCodec`. It has hashing, encoding, and decoding capabilities. It uses `java.crypto.Cipher` utilities to hash or encode a plaintext password and also to decode a masked string **using the same algorithm and key**. [emphasis mine]

YYTVicky:
Hi, 
  Thanks for your response! @jbertram @michaelandrepearce 
  our tool generates another fix template as: 
   KeyGenerator keyGen = KeyGenerator.getInstance("AES");
   keyGen.init(128);
   SecretKey secretKey = keyGen.generateKey();
   byte[] keyBytes= secretKey.getEncoded();
  for your encode and decode file. the keygenerator are able to generate the Symmetric key, the encode and decode part can use the same key now.

jbertram:
@YYTVicky, I don't really understand what you mean. Where is this tool you're referring to when you say "our tool"?

PR #3110, opened 2020-04-30
ARTEMIS-2746 Reduce memory footprint of journal loading

franz1981:


franz1981:
I've found a  more important fix around journal loading then this: although in a good state I will close this to decide if re-push it in the future.

PR #3113, opened 2020-05-04
ARTEMIS-2445 : fix issue of large message lost on divert

bhaveshkharwa:


artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
```diff
@@ -81,7 +81,7 @@
    // We use properties to establish routing context on clustering.
    // However if the client resends the message after receiving, it needs to be removed, so we mark these internal
    Predicate<SimpleString> INTERNAL_PROPERTY_NAMES_PREDICATE =
-      name -> (name.startsWith(Message.HDR_ROUTE_TO_IDS) && !name.equals(Message.HDR_ROUTE_TO_IDS)) ||
```
clebertsuconic, line 84: HDR_ROUTE_TO_IDS is used to transfer messages to a different cluster.

Perhaps you had messages moving to a different cluster and thought they were lost.

Without a proper test, this PR cannot be merged. If you have a real case you need to prove it at least.

jbertram:
I'm closing this for now. Please re-open when you are able to provide a valid test in the test-suite to validate the fix and prevent future regressions.

PR #3120, opened 2020-05-07
ARTEMIS-2753 Fixing OpenWire Temporary queue names over wildcard

clebertsuconic:


michaelpearce-gain:
Im just looking at this, on windows it seems it doesnt pick up. Another note re wildcards looks like something might be up with Advisories, we dont have many test in that space though.... so i havent got much to steal and test with,....unless you know of somewhere and can point me to?

clebertsuconic:
@michaelpearce-gain  I don't think this is a real issue in most environments... however since I have tests failing now... I wanted to be diligent and make sure it's fixed.

PR #3121, opened 2020-05-07
ARTEMIS-2749 - Broadcast Group Control Improvements

andytaylor:
added some default behaviour for when overridden

https://issues.apache.org/jira/browse/ARTEMIS-2749


artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessage.java
```diff
@@ -91,7 +91,7 @@
             entry.getKey().equals("priority")) {
             // Ignore
          } else if (entry.getKey().equals("userID")) {
-            jmsMessage.put("JMSMessageID", entry.getValue().toString());
+            jmsMessage.put("  JMSMessageID", entry.getValue().toString());
```
brusdev, line 94: It looks like a typo.

andytaylor: fixing now

PR #3123, opened 2020-05-08
ARTEMIS-2755 Improve SoakPagingTest stability

brusdev:
Await producers and consumer creation before to await the test running time.


clebertsuconic:
there are no code changes here, why do we need a JIRA?

brusdev:
I created the PR #3125 to add timeouts avoiding to await for infinite time

PR #3124, opened 2020-05-09
ARTEMIS-2614 Create queues based on FQQN for AMQP protocol

Havret:
This is my attempt to address the problem of the auto-creation of queues based on FQQN. The simple scenario seems to work, but some decisions need to be made before I can move forward with this. 

1) There is a problem with the `RoutingType` mismatch. Previously it didn't matter what `RoutingType` receiver tries to attach to, as the implementation assumed pre-existence of the queue. Now the routing type matters, as we need to create a properly configured queue if it doesn't exist. That's why `testQueueConsumerReceiveTopicUsingFQQN
` test fails currently. To make it pass I would need to subscribe to a topic instead of a queue. I'm not sure but is this a breaking change or not? 
2) I'm not sure what is the expected behavior when durability configuration mismatches. The problem is described in a test `testAttachToPreConfiguredNonDurableQueueUsingDurableFQQN
`. Should we reject the attempt to attach when the pre-configured queue is non-durable and attach request has the source with `TerminusDurability
` of `UNSETTLED_STATE` or `CONFIGURATION`? And what about the opposite, pre-configured queue is durable and we received source with `NONE` value for `TerminusDurability`?

I would greatly appreciate your help. :)

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
@@ -301,6 +301,16 @@ public class AMQPSessionCallback implements SessionCallback {
       }
    }
 
+   public void createQueue(QueueConfiguration queueConfiguration) throws Exception {
+      try {
+         serverSession.createQueue(queueConfiguration);
+      } catch (ActiveMQSecurityException se) {
+         throw ActiveMQAMQPProtocolMessageBundle.BUNDLE.securityErrorCreatingConsumer(se.getMessage());
+      } catch (ActiveMQQueueExistsException e) {
+         // ignore as may be caused by multiple, concurrent clients
+      }
+   }
+
    public QueueQueryResult queueQuery(SimpleString queueName, RoutingType routingType, boolean autoCreate) throws Exception {
       QueueQueryResult queueQueryResult = serverSession.executeQueueQuery(queueName);
 
@@ -323,8 +333,6 @@ public class AMQPSessionCallback implements SessionCallback {
 
       return queueQueryResult;
    }
-
-
 
    public boolean checkAddressAndAutocreateIfPossible(SimpleString address, RoutingType routingType) throws Exception {
       boolean result = false;
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
@@ -29,6 +29,7 @@ import org.apache.activemq.artemis.api.core.ActiveMQSecurityException;
 import org.apache.activemq.artemis.api.core.Message;
 import org.apache.activemq.artemis.api.core.RoutingType;
 import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.api.core.QueueConfiguration;
 import org.apache.activemq.artemis.core.io.IOCallback;
 import org.apache.activemq.artemis.core.message.LargeBodyReader;
 import org.apache.activemq.artemis.core.persistence.OperationContext;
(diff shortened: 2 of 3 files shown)


jbertram:
> Previously it didn't matter what RoutingType receiver tries to attach to, as the implementation assumed pre-existence of the queue.

That's not technically true. The broker supports auto-creation of addresses and queues. When auto-creating resources the routing type is based on the following (in descending order of precedence):

1. The terminus capability set on the link as described in section 5.2 of the [AMQP-to-JMS mapping document](https://www.oasis-open.org/committees/download.php/60574/amqp-bindmap-jms-v1.0.pdf).
2. The [prefix](http://activemq.apache.org/components/artemis/documentation/latest/address-model.html#using-prefixes-to-determine-routing-type) set on the acceptor.
3. The dynamic default routing-type set in the `address-settings`.
4. The [static default routing-type](https://github.com/apache/activemq-artemis/blob/master/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java#L537) set the the code-base.

Havret:
> That's not technically true. The broker supports auto-creation of addresses and queues. When auto-creating resources the routing type is based on the following (in descending order of precedence):

That's true, but not when you specify address+queue using FQQN. If you do, it will attach to any address+queue combination that matches irrespectively of the defined routing type.

gemmellr:
TerminusDurability is not the same thing as queue durability and so shouldn't be used for indicating such.

Havret:
> TerminusDurability is not the same thing as queue durability and so shouldn't be used for indicating such.

But currently seems to be used that way, isn't it?

jbertram:
> That's true, but not when you specify address+queue using FQQN.

Of course, but that wasn't my point. My point was simply that the resources didn't have to pre-exist in the first place (i.e. without FQQN) as you seemed to assert.

Havret:
@jbertram Yes, but why should we consider not FQQN based scenarios?

gemmellr:
> > TerminusDurability is not the same thing as queue durability and so shouldn't be used for indicating such.
> 
> But currently seems to be used that way, isn't it?


No. It is only used in one place, appropriately, to indicate the subscription to a topic node is durable or not. That is, the link/terminus is durable or not. Not the queue the broker happens to create (which is an implementation detail, there need not be any queue).

Havret:
@gemmellr I am not sure I follow. Isn't this just JMS limitation? What I'm aiming for here is to support programmatically what is already supported via broker configuration and via GUI.

jbertram:
@Havret, I suppose I misunderstood what you were asserting in your original comment when you were talking about how "the implementation assumed pre-existence of the queue."

jbertram:
> What I'm aiming for here is to support programmatically what is already supported via broker configuration and via GUI.

By "programmatically" you mean specifically from an AMQP client, right? You can already create the resources any client would need programmatically through the management API.

Havret:
@jbertram No worries. I supposed I could've been more specific. :)

gemmellr:
@Havret No its not related to JMS. TerminusDurability is simply not equal to queue durability.

For example, were you to conflate terminus and queue config like that, I'd note that the terminus expiry policy would then dictate the broker should be deleting your queues when you go away in almost all situations. Which is typically not going to be preferred behaviour for most folks, and fortunately isnt the case since terminus != queue. The core AMQP spec doesn't deal with creation of nodes (queues/topics/others), which again reinforces terminus durability isnt queue durability. Its expected other mechanisms are used for fine grained creation of nodes (queues/topics/others) such as layered management like Artemis own existing management abilities Justin notes, which can be used using AMQP messages also, or alternatively the AMQP Management layered [draft] spec.

Havret:
@gemmellr 
But it seems it's already conflated, as it is being used to create shared durable queues like that? And they aren't deleted?

> Its expected other mechanisms are used for fine grained creation of nodes (queues/topics/others) such as layered management like Artemis own existing management abilities Justin notes, which can be used using AMQP messages also

This sounds really promising! It is almost like it is done in RabbitMQ. Could you please point me where I can find more regarding this API? 

The whole effort I am making here is just a result of a discussion on mailing list where Martyn Taylor said:

> My understanding was that auto-creation of FQQN queues was always an
> intention but was never implemented.  Enabling this in the AMQP (and other
> protocols) *should* allow any JMS over AMQP clients to utilize this
> feature.  There may be some details here in terms of order of
> precedence when using topic/queue capabilities (as defined in the JMS AMQP
> mapping) and what is specified in the FQQN, but could be implemented
> relatively easily.

jbertram:
> Could you please point me where I can find more regarding this API?

The [management API](http://activemq.apache.org/components/artemis/documentation/latest/management.html) generally and [management messages](http://activemq.apache.org/components/artemis/documentation/latest/management.html#using-management-message-api) in particular are both discussed in the documentation. The documentation discusses this mainly in terms of core and JMS. The broker ships the [management example](https://github.com/apache/activemq-artemis/tree/master/examples/features/standard/management) where you can see management messages in action. You can also see an example of an AMQP client using management messages in [the test-suite](https://github.com/apache/activemq-artemis/blob/master/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpManagementTest.java).

gemmellr:
> @gemmellr
> But it seems it's already conflated, as it is being used to create shared durable queues like that? And they aren't deleted?
> 

This isnt the same. The client is creating a subscription to a 'topic' node's address, it says nothing about queues or their durability. It asks for the link terminus to be retained or not. For the durable-subscription case it specifically sets a terminus expiration of never, that is the terminus (/subscription) should be retained when it goes away. In the 'shared' cases there is a further layered mechanism (which has been negotiated as supported) further governing the behaviour of 'shared', 'global', link names etc. In no case does the client ask for a specific queue to be made, durable or otherwise, its simply a broker implementation detail that the broker chooses to model its 'topic' subscriptions as a queue encoding details such as a particular name.

If nothing else you will break things for most existing usage if you were to now conflate terminus and queue durability at this point when they never have been before. Similarly if you conflate Terminus config with queue config, youd really also need to conflate terminus expiry with the queue lifecycle. As earlier, in most cases (i.e mostly everything except 'never', as above) thats going to result in a queue being deleted when you dont want.

The core spec does not deal with regular queue creation, so it also doesnt cater to their durability, terminus durability isnt queue durability. 

> 
> Could you please point me where I can find more regarding this API?

Its not something I have used, but Justin looks to have already linked to the details I was aware of in his last post above. 

One place I know it is actually used as part of doing Artemis management using AMQP messages is:
https://github.com/EnMasseProject/enmasse/blob/0.31.1/agent/lib/artemis.js#L58

HavretGC:
@gemmellr
I'm not sure I'd break anything as most jms based clients out there doesn't use FQQNs. 
Could you please refer to the points from documentation and Martyn's words? I believe he is responsible for the original implementation. 

I believe that your option is a little bit biased, as you're looking at AMQP as the implementation detail for JMS. I don't believe this is the case. Artemis as it is doesn't care about, `shared`, `global`, `topics`, and `subscriptions`. These are JMS bits. It is modeled on addresses and queues and their properties.

Havret:
@gemmellr
I'm not sure I'd break anything as most jms based clients out there doesn't use FQQNs. 
Could you please refer to the points from documentation and Martyn's words? I believe he is responsible for the original implementation. 

I believe that your option is a little bit biased, as you're looking at AMQP as the implementation detail for JMS. I don't believe this is the case. Artemis as it is doesn't care about, `shared`, `global`, `topics`, and `subscriptions`. These are JMS bits. It is modeled on addresses and queues and their properties.

gemmellr:
I'm really not looking at it that way at all. I'd say I more typically advocate for not catering to JMS specifically, since AMQP is entirely separate from JMS, regardless whether a certain JMS client I work on uses it.

That I can recall, JMS hasn't really even entered my replies here except largely dismissing it as a factor when you keep bringing it up (one last time though: those bits you refer to are not only for JMS clients, other AMQP clients can and do use them). This discussion has zero dependence on JMS in my mind.

I'm looking at this in terms of the AMQP spec itself, and having discussed this area previously with folks who actually wrote it. Terminus durability isnt queue durability. Again, the core spec does not cover queues or creating them, it simply attaches to nodes [configured out of band] with a given address.

You would similarly break behaviour for any AMQP client out there using FQQNs already by all of a sudden considering terminus durability to be queue durability.

Havret:
@gemmellr I'm bringing up JMS because JMS-AMQP mapping was the first thing you referred me to. Speaking in terms of AMQP spec, words like `shared`, `global`, `topic`, and `subscription` doesn't appear there even once, and you keep bringing them up. I'm not an expert in this area, I've never spoken with the authors of the spec. The only resources I have are spec itself, Artemis Source Code, and its documentation. Forgive me my ignorance but if I read in docs for amqp connector that

> If the Terminus Durability is either UNSETTLED_STATE or CONFIGURATION then the queue will be made durable (similar to a JMS durable subscription) (...).

And I see this being used to implement JMS features I interpret it this way.

Shared Durable Subscription that you're recalling as the only feasible case to use TerminusDurability is not an AMQP feature. It's a way that JMS based clients can attach to multicast queues where topic maps to address and subscription name maps to durable queue name. The fact that the durable queue is created there is not an implementation detail from my perspective. When I configure broker I'm not configuring it in terms of topics and subscriptions. When I use FQQN I don't even need to know that such things exist. I have a broker with configured addresses and queues and I want to attach to them. In the ideal world, it would be great if it was possible to do it without the need to pre-configure (for instance to make development easier).

gemmellr:
> 
> 
> @gemmellr I'm bringing up JMS because JMS-AMQP mapping was the first thing you referred me to. 

I didn't refer you to it. Someone else may have briefly linked to it.

> Speaking in terms of AMQP spec, words like `shared`, `global`, `topic`, and `subscription` doesn't appear there even once, and you keep bringing them up. 

Topic and subscriptions aren't AMQP specific terms (or JMS specific), though are generally understood messaging concepts, but since the spec doesn't deal specifically with queues or topics, merely nodes, plus links, terminus etc, their lack of use isnt surprising to me.

You are correct that the 'shared' and 'global' terms arent in the draft JMS mapping document either, because I have yet to publish the update with them in it. I did only mention those terms after you brought up shared durable subscriptions however, and again I didn't actually direct you to that document.

JMS has no special bearing in this discussion from my view, its purely AMQP from my perspective. 

> I'm not an expert in this area, I've never spoken with the authors of the spec. The only resources I have are spec itself, Artemis Source Code, and its documentation. 

Thats fair enough. You could possibly include in that resource list, discussion from those who people might say are quite familiar with the area, and are indicating having discussed this very specific item with authors on more than one occasion (including yesterday).

> Forgive me my ignorance but if I read in docs for amqp connector that
> 
> > If the Terminus Durability is either UNSETTLED_STATE or CONFIGURATION then the queue will be made durable (similar to a JMS durable subscription) (...).
> 
> And I see this being used to implement JMS features I interpret it this way.
> 

Specifically the doc section it appears you may be quoting from is "AMQP and Multicast Addresses (Topics)", i.e its about topics rather than queues, aligning with what I have said.

The broker models its 'topic' subscriptions (i.e consumer link attaching using address of, err, an address ) using a queue that it happens to expose. There is no need for an actual exposed queue here at all, its a topic subscription, the queue is essentially a synthetic creation. It is the link terminus attaching to the topic-like node (so, address), i.e effectively the subscription representation, that is being indicated as durable (and separately marked as never expiring, to ensure its contuing existance) in this case to retain it. This is quite a different situation than if the link terminus were intended for attaching directly to an actual queue.

> Shared Durable Subscription that you're recalling as the only feasible case to use TerminusDurability is not an AMQP feature.

I am not saying (and dont recall ever having done so here) that Shared Durable Subscriptions are the only feasible case for TerminusDurability. I have said topic-like subscriptions of any kind are the main use case for it (along with expiry in some cases). I have said that TerminusDurability isnt the same direct thing as queue durability. The spec doesnt deal with creating them and so doesn't deal with their durability.

Separately though, they are however an AMQP feature - a layered extension feature yes, but a fully negotiated one that any AMQP client or server can use doing regular AMQP interactions, and again already do.

Not really important, but for historic clarity, as it happens shared AMQP subscriptions existed in actual use in non-JMS client and servers long before any work was done in that area for a JMS 2 mapping for use in AMQP JMS clients. That existing behaviour was used as a starting point of reference (and one of many many approaches looked at) and expanded slightly to cover other requirements.

> It's a way that JMS based clients can attach to multicast queues where topic maps to address and subscription name maps to durable queue name.

All AMQP clients or servers attach to 'topic-like' nodes (in this case, the brokers 'addresses') in the same way, not just JMS ones. Again, the queue existing in this scenario is completely a broker implementation detail (an approach that others also choose, admittedly), and it is entirely possible to model that topic subscription without any exposed queue.


> The fact that the durable queue is created there, is not an implementation detail from my perspective. When I configure broker I'm not configuring it in terms of topics and subscriptions. When I use FQQN I don't even need to know that such things exist. I have a broker with configured addresses and queues and I want to attach to them. In the ideal world, it would be great if it was possible to do it without the need to pre-configure (for instance to make development easier).

I understand that it might be a nice side effect that you desire in certain cases, but that doesnt mean it isnt an impl detail, or that terminus durability is queue durability.

gemmellr:
The way I know some other servers have facilitated things in this space is to enable pre-defining configuration that will apply to a queue/topic/other when things are being auto created, e.g allowing you to pattern match what configuration that applies when a link attaches to a given address. Possibly much like the the existing address config stuff can pre-configure address behaviour without actual addresses.

Havret:
@jbertram, @gemmellr 

So after all it wasn't too productive discussion. I debugged the code, and apparently methods `createSharedDurableQueue` and `createUnsharedDurableQueue` don't explicitly create durable queues. :) The queues are durable by default if it's not specified otherwise. So if drop "setDurable" bits, FQQN queues will be created as durable which makes sense, and doesn't break anything.

However, there are still some things that need to be cleared up regarding my first question. 

> There is a problem with the RoutingType mismatch. Previously it didn't matter what RoutingType receiver tries to attach to, as the implementation assumed pre-existence of the queue. Now the routing type matters, as we need to create a properly configured queue if it doesn't exist. That's why testQueueConsumerReceiveTopicUsingFQQN test fails currently. To make it pass I would need to subscribe to a topic instead of a queue. I'm not sure but is this a breaking change or not?

michaelandrepearce:
First off you shouldn't affect the existing logic, e.g. the original use case which was to allow a fqqn to bind to a queue matching the name ignoring the routing type. E.g. if a fqqn matches an existing queue it should bind to it regardless routingtype.

For stomp how its done is"

A RoutingType is passed to the autoCreateDestinationIfPossible method which is based on the frame's header or destination prefix. If that is null then it uses the default value specified in the address-settings.

michaelandrepearce:
If you look in  ProtonServerSenderContext, there already is a field derived and caclulated "routingTypeToUse"

Also i would use the same logic  already in there thats working out if its a durable or non-durable queue (aka if its volatile)

Havret:
So the routing type is calculated there multiple times, using slightly different variations of the same logic. It's pretty messy tbh, maybe it would be a good idea to unify this, and extract one method to determine routing type.

michaelandrepearce:
To be honest i would expect for FQQN it to apply all the exact logic thats in there, except the where we call createQueueName you use queueNameToUse if it present instead, that was possibly extracted earlier in the composite address extraction

e.g. line 413

               queue = createQueueName(connection.isUseCoreSubscriptionNaming(), clientId, pubId, shared, global, false);

to 

               queue = queueNameToUse == null ? createQueueName(connection.isUseCoreSubscriptionNaming(), clientId, pubId, shared, global, false) : queueNameToUse;


And like wise line 438 extactly the same

I wouldnt expect changes anywhere else to achieve this tbh.

Havret:
I wish it was that simple. Unfortunately for anycast routing type we don't even get to this line.

michaelandrepearce:
But thats just original logic added for FQQN, so just change line 515 e.g. this

         if (!result.isExists()) {
            throw new ActiveMQAMQPNotFoundException("Queue: '" + queueName + "' does not exist");

to return null, instead of throw exception (as like for non fqqn)

do keep the rest of it.

Havret:
So if I make these changes, it will work partially for multicast routing type (but all queues will be created using `createUnsharedDurableQueue` method. Not sure if this is desired. And it doesn't work at all for anycast, because of this if:

image

Any suggestions?

michaelandrepearce:
For AMQP clients they should really be using correct structures of AMQP not coding to FQQN. Its intent was solely for JMS legacy workaround where client has not yet upgraded to JMS 2 which supports shared subs etc. And in JMS 2.0 Amqp client like qpid they correctly set the amqp structures.

E.g. in your case such where NMS doesnt yet support 2.0 (wip) then youre restricted to fqqn for shared subs. Obviously the whole issue is removed when you get NMS 2.0 upto 2.p spec.

michaelandrepearce:
Just to note and for the record, i am cautious of this change, as the FQQN was added for legacy stuff, making this change could be impacting to those expected behaviours. 

Unlike stomp AMQP has a very rich structure that supports this, as such its not like comparing stomp or openwire which are limited.

Personally i think for any higher level AMQP clients (e.g JMS/NMS) they would be better to expend time into adding support for the equivalent of jms shared sub in their abstractions, , so in case of NMS that would be to get that updated to  similar spec level with JMS 2.0

Havret:
So the behavior shouldn't change unless somebody relies on the fact, that the broker returns an error if one tries to attach to the queue that doesn't exist, instead of simply creating it.

> If i was to guess at a default queue type, i would go with a durable shared queue. You could look to make use of ParameterisedAddress like in Core to get extra queue config for a queue off the name, to allow for more configuration to be able to be present in the fqqn address

So I think this is the current behavior as the only difference between `createUnsharedDurableQueue` and is `createSharedDurableQueue` is that Unshared set consumer limit to 1.

michaelandrepearce:
I was editing my comment, as hadn;'t fully finished.

Mappings i would expect from a black box testing:

JmsQueueConsumer (createConsumer(Queue)-> shared durable
JmsTopicConsumer (createConsumer(Topic)) -> non durable shared
JmsTopicConsumer (createDurableconsumer(Topic)) -> non shared durable

For nonshared, non durable a user would just not use FQQN.

michaelandrepearce:
re:

> So the behavior shouldn't change unless somebody relies on the fact, that the broker returns an error if one tries to attach to the queue that doesn't exist, instead of simply creating it.


They do, if security settings etc, youll need tests on this.

Havret:
I'm not sure if mixing JMS concepts into this is the best idea. Determining whether a queue is durable/non-durable or shared/non-shared based on routing type seems to me pretty non-intuitive.

Moreover, the way that JMS differentiate `createDurableconsumer(Topic)` and `createConsumer(Topic)` is based on `TerminusDurability` and `TerminusExpiryPolicy`. As @gemmellr  tried to explain to me throughout the big chunk of this thread, this property cannot be used to determine whether a queue is durable or not. 

In the ideal world, I would just use `TerminusDurability` and `shared` capability. Having these two properties with FQQN, would make it possible to fully leverage Artemis Address Model in a dead-simple way:

| Address   | Queue  | Routing Type  | Durable | Shared|amqp address  | capabilities  | durable |
|---|---|---|---|---|---|---|---|
| a1  | q1  | Multicast  | true  | true  | a1:q1 | shared, topic | 2|
| a1  | q1  | Anycast| true  | true  | a1:q1 | shared, queue | 2|
| a1  | q1  | Multicast  | true  | false| a1:q1 | topic | 2|
| a1  | q1  | Anycast| true  | false  | a1:q1 | queue | 2|
| a1  | q1  | Multicast  | false | true  | a1:q1 | shared, topic | 0 |
| a1  | q1  | Anycast| true  | false  | a1:q1 | shared, queue | 0 |
| a1  | q1  | Multicast  | false  | false| a1:q1 | topic | 0 |
| a1  | q1  | Anycast| false  | false  | a1:q1 | queue | 0 |

If we create a queue based on FQQN, only if the queue doesn't exist, we wouldn't break any existing users, as they are using FQQN to attach to pre-configured queues anyway.

michaelandrepearce:
Why should it not be the whole feature was to sort a jms issue in 1.1

As i said as a non jms amqp user, simply shouldnt be looking to use fqqn, but full amqp native and correct states.

I would be -1 this if the idea is to use it outside jms/nms scope for other uses.

michaelandrepearce:
Like wise to note as using fqqn with jms, i would expect the broker to behave the same if openwire, amqp or core. So anything done here i would expect alignment in core and openwire with agnostic jms test validating this (just switching the jms provider) we do this for many of the cross protocol jms tests

michaelpearce-gain:
If you look at FQQNOpenWireTest youll notice some existing cross protocol tests, you will want to add AMQP to that, and enhance for any extra cases, to ensure behaviour across JMS is aligned.

clebertsuconic:
I would like to -1000 on this.

it's already complicated having auto-creating and routing messages.

Doing this kind of thing would require doing a lot of computation on queue adding. I don't really think we should use FQQN on auto-create.

quite the opposite you should not be using auto-creatin at all on a production system IMHO.

If you do, use it for simple cases.. this is definitely not a simple case and it would induce a lot of calculations during simple message routes on regular cases.

Havret:
Yes, I totally agree. I followed @jbertram advice and implemented topology management system similar to what is available for RabbitMQ. It is more robust and powerful solution. Sorry for the hassle. I will close this PR by tomorrow.

clebertsuconic:
@Havret thanks for contributing.. I have seen other nice contributions from you...

it's just that I hit another recent around mirroring and broker connection that I was working on, and I have seen another issue around this.. when I saw this I came with the -1... that's why I didn't bring this up before.

clebertsuconic:
I will close it based on the discussion!

PR #3125, opened 2020-05-10
ARTEMIS-2755 Add SoakPagingTest timeouts

brusdev:
Limit time awaiting producers and consumers.

### DIFF
--- a/tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/replicationflow/SoakPagingTest.java
+++ b/tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/replicationflow/SoakPagingTest.java
@@ -28,6 +28,7 @@ import javax.jms.Topic;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.activemq.artemis.tests.smoke.common.SmokeTestBase;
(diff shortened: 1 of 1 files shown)


tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/replicationflow/SoakPagingTest.java
```diff
@@ -157,10 +158,14 @@ public void run() {
          }
 
          System.out.println("Awaiting producers...");
-         producersLatch.await();
+         if (!producersLatch.await(30000, TimeUnit.MILLISECONDS)) {
+            System.out.println("Awaiting producers timeout");
```
clebertsuconic, line 162: -1000

Producers never finishing here would be an issue that should fail the test, not silently fail it.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/replicationflow/SoakPagingTest.java
```diff
@@ -157,10 +158,14 @@ public void run() {
...
+         }
 
          System.out.println("Awaiting consumers...");
-         consumersLatch.await();
+         if (!consumersLatch.await(30000, TimeUnit.MILLISECONDS)) {
+            System.out.println("Awaiting consumers timeout");
```
clebertsuconic, line 167: same here.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/replicationflow/SoakPagingTest.java
brusdev: @clebertsuconic I see your point. I could return 0 if the Producers never finishing here. WDYT?

clebertsuconic: System.exit(0) with a System.err.println notification would be ok.

PR #3126, opened 2020-05-11
ARTEMIS-2752 - Properly cleanup Federated Address Divert consumers

cshannon:
DivertBindings are now properly cleaned up when a queue binding is
removed that matches the divert. The correct key is now used to remove
the queue address from the set and the correct address is now used to
remove the remote consumer.


clebertsuconic:
LGTM: I would merge it right now if I could.. will do it at the first chance.

PR #3127, opened 2020-05-11
ARTEMIS-2709: Fix LiveToLiveFailoverTest::scaleDownDelay

atris:
Test fails with the primary server being killed by the crash and the backup server is killed
by the tearDown before ScaleDownHandler can kick in. This commit adds a wait method to allow
ScaleDownHandler to process before the test completes.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/failover/LiveToLiveFailoverTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/failover/LiveToLiveFailoverTest.java
@@ -32,8 +32,10 @@ import org.apache.activemq.artemis.core.config.ha.ColocatedPolicyConfiguration;
 import org.apache.activemq.artemis.core.config.ha.SharedStoreMasterPolicyConfiguration;
 import org.apache.activemq.artemis.core.config.ha.SharedStoreSlavePolicyConfiguration;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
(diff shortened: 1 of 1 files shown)


atris:
@clebertsuconic Please see and let me know your comments

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/failover/LiveToLiveFailoverTest.java
```diff
@@ -207,6 +207,9 @@ public void scaleDownDelay() throws Exception {
 
       crash(session);
 
+      // Wait for failover to happen
+      Thread.sleep(100);
```
clebertsuconic, line 211: it would be nice if we could instead use a class called Wait, that would keep trying until a certain condition was reached in the server.

This type of sleep will certainly eventually fail.

i'm looking for a case where the test always pass.. with this we will still have eventually a failure.

atris: Can you suggest a condition that can be used to check if ScaleDownHandler has kicked in?

atris: Fixed, thanks

atris:
@clebertsuconic Fixed the same, thanks.

clebertsuconic:
wow.. you really did it. nice one!

PR #3128, opened 2020-05-12
Remove deprecated sudo setting.

dengliming:
[Travis are now recommending removing the sudo tag.](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration)

_"If you currently specify sudo: false in your .travis.yml, we recommend removing that configuration"_


clebertsuconic:
thank you

PR #3130, opened 2020-05-12
ARTEMIS-2757 improving flow control in AMQP

clebertsuconic:


tabish121:
I've spent the afternoon testing this and from what I can tell while it does seem to mitigate the issue to some degree the underlying problem still presents if the broker it pressed.  

In this trace the broker is still pretty far behind in sending dispositions while it is granting more credit to the sender all the while.  
> -> AMQP: Transfer{handle=0, deliveryId=599999, deliveryTag={504876}, messageFormat=null, settled=false, more=false, rcvSettleMode=null, state=null, resume=null, aborted=null, batchable=null}
<- AMQP: Disposition{role=RECEIVER, first=519174, last=519174, settled=true, state=Accepted{}, batchable=false}
<- AMQP: Flow{nextIncomingId=599984, incomingWindow=2147483647, nextOutgoingId=1, outgoingWindow=2147483647, handle=0, deliveryCount=599983, linkCredit=1000, available=0, drain=false, echo=false, properties=null}

The issue doesn't seem to occur to the same degree as before as my Quiver benchmark runs tend to succeed without the massive memory overhead that ensues when the broker gets horribly far behind but it seems like it is still possible to get there over the long haul.

clebertsuconic:
I only changed flow control at the consimers size.  Could be this caused by the producers as well ?  Perhaps I need a similar change on the other side ?

tabish121:
I would be surprised if the consumer had anything to do with this particular issue.  It seems from external observation that the broker grants credit when the link credit hits the low water mark not taking into consideration if the actual messages that were dispatched to the broker for enqueue and dispatch have actually been accepted and settled.  

My guess would be that the credit runnable is triggered to send a flow before it really should be but I've not been in the broker code in quite awhile and I'm having a hard time working out the current mechanics without spending more time with it.  

Just a wild guess at this point but the code currently doesn't seem to track the difference between the current credit and the amount of messages dispatched outwith to the broker executors for handling and so the first couple actually handled messages that come back through for settlement might trigger the flow much to early because the client has already used up link credit to the point that the link.getCredit call will be below the low water mark because that many message did enter the broker.

clebertsuconic:
The code definitely don't take pending settlements into account when issuing credits..

all that's being done, as you said is:

```java
       if (receiver.getCredit() <= threshold) {
            int topUp = refill - receiver.getCredit();
            if (topUp > 0) {
               // System.out.println("Sending " + topUp + " towards client");
               receiver.flow(topUp);
               connection.instantFlush();
            }
         }
```


To be honest I thought Proton was only updating the credit after settlements. 

I can update the code tomorrow to take settlements into account.

tabish121:
Yes, that was the bit I was looking at.  Proton can't do anything in relation to settlements with regards to link credit because they are two different things.  The credit is consumed as each delivery arrives, settlements merely signal to the sending peer that you've accepted them (or rejected etc) and that it can forget about them as you've taken ownership.  We had a similar sort of issue in Qpid JMS a couple years ago in regard to the prefetch buffer and issuing credit without accounting for messages sitting idle vs actually getting dispatched which is what got me thinking along these lines.

gemmellr:
Credit and settlement (+ delivery outcomes) are separate things, they have separate protocol interactions so they specifically aren't tied together. If you say 'here is more credit' then you have given credit.

gemmellr:
Yeah what he said, lol.

clebertsuconic:
Ok.  I will update that tomorrow. 

this has been this way for ages.  No regression in here. (Not saying I won’t do it. Just saying that no recent change modified that )


Leave it with me and I will change it tomorrow.

clebertsuconic:
I know what to change. non durable messages will be sent without the executor, and I will add a pending metrics for the in flight executions for durable.


will be done tomorrow.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
```diff
@@ -496,15 +499,21 @@ public void serverSend(ProtonServerReceiverContext context,
             }
          } else {
             message.setConnectionID(receiver.getSession().getConnection().getRemoteContainer());
-            // We need to transfer IO execution to a different thread otherwise we may deadlock netty loop
-            sessionExecutor.execute(() -> inSessionSend(context, transaction, message, delivery, receiver, routingContext));
+            // If there are pending messages (durable) on the executors, we should still use the executors
+            // otherwise we risk running out of order
+            if (pendingSettles == 0 && !message.isDurable()) {
```
clebertsuconic, line 504: This should fix your test... as this will guarantee no execution for non durable.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
```diff
@@ -543,7 +552,7 @@ private void inSessionSend(final ProtonServerReceiverContext context,
             afterIO(new IOCallback() {
                @Override
                public void done() {
-                  connection.runLater(() -> {
+                  connection.runNow(() -> {
```
clebertsuconic, line 555: @tabish121 this will also fix your test, as it will avoid using an executor for non durable

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
```diff
@@ -128,10 +147,11 @@ public void run() {
          }
 
          connection.requireInHandler();
-         if (receiver.getCredit() <= threshold) {
-            int topUp = refill - receiver.getCredit();
+         int pending = context != null ? context.pendingSettles : 0;
```
clebertsuconic, line 150: @tabish121 also, in case you mix durable and non durable and use the executors, this will adjust pending settles on the credits.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
clebertsuconic: I tested this.. and the number of pending settlements is always 0 after this change for non durables... with durables it can go to a few dozen at a times.. .but credits will now take pending settlements into consideration.

clebertsuconic:
I added a tag do-not-merge-yet just because the second commit will be squashed into the first once review is done.

tabish121:
I've been able to produce some signs of this at the start of quiver runs where the client will backlog by as much as 16k unsettled messages indicating that the broker is still over offering credit but it is not yet been repeatable with any sort of tracing enabled so it is hard to capture the chain of events.  The depth of unsettled never seems to top 20k so that's progress anyway.

clebertsuconic:
@tabish121 I think I'm stoopid...

```java
         int pending = context != null ? context.pendingSettles : 0;
         int adjustedCredit = receiver.getCredit() + pending;
         if (adjustedCredit <= threshold) {
```

adjustedCredit should discount pending, not add it!

clebertsuconic:
although you shouldn't really have pending settles with non persistent.

clebertsuconic:
the calculation is off.. let me fix that...

clebertsuconic:
and @tabish121 sorry about that.. I thought I had it right

clebertsuconic:
The calculations were actually ok from what I could later test.  I will need to get your tests.

tabish121:
Tested with some fixes to local test suite and this appears to sort the issues that I was seeing with over eager credit granting.

PR #3132, opened 2020-05-13
ARTEMIS-2760 Filter AddressQueueReaper misleading error messages

brusdev:


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1744,6 +1746,15 @@ public void run() {
 
                   server.removeAddressInfo(address, null);
                }
+            } catch (ActiveMQAddressDoesNotExistException e) {
+               // the address and queue reaper is asynchronous so it may happen
+               // that the address is removed before the reaper removes it
+               ActiveMQServerLogger.LOGGER.debug(e.getMessage(), e);
```
clebertsuconic, line 1752: for anything Debug, please declare a log on PostOfficeimpl, and then use log.debug(e.getMessage, e);


Also, are you sure there's no leak on the address / versus Reaper?

just make sure you're not hiding a bug under rug by eatin an exception as debug.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1744,6 +1746,15 @@ public void run() {
...
+            } catch (ActiveMQShutdownException e) {
+               // the address and queue reaper is asynchronous so it may happen
+               // that the broker is shutting down while the reaper iterates
+               // through the addresses, next restart this operation will be retried
+               ActiveMQServerLogger.LOGGER.debug(e.getMessage(), e);
```
clebertsuconic, line 1757: log.debug please. same note as before

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
brusdev: I added a check to see if the address still exists

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
brusdev: I pushed a commit to replace the wrong logger.

PR #3133, opened 2020-05-13
ARTEMIS-2759 Warn ignored connection factory URI parameters

brusdev:


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/uri/URISchema.java
```diff
@@ -31,7 +31,7 @@ public T newObject(URI uri, P param) throws Exception {
    }
 
    public void populateObject(URI uri, T bean) throws Exception {
-      BeanSupport.setData(uri, bean, parseQuery(uri.getQuery(), null));
```
clebertsuconic, line 34: we have ignored this per design in the past.

- the Protocol manager will try to set properties that will not be available on every protocol, and that will be using the TCP Schema as far as I know.
- Also there's an issue between versions. We store the JNDI with the properties for the connection factory. and it may not have a certain property between versions.


But I am more concerned about the protocol manager. Have you tested this?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/SimpleJNDIClientTest.java
```diff
@@ -115,6 +115,18 @@ public void testEmptyConnectionFactoryString() throws NamingException, JMSExcept
 
    }
 
+   @Test
+   public void testConnectionFactoryStringWithInvalidParameter() throws NamingException, JMSException {
+      Hashtable<String, String> props = new Hashtable<>();
```
clebertsuconic, line 120: Are you asserting anything here? you could use the AssertionLogger (whatever the name that is)

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/uri/URISchema.java
brusdev: where does the protocol manager set the properties using the TCP schema?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/SimpleJNDIClientTest.java
brusdev: I added the assertion using the AssertionLogger.

PR #3134, opened 2020-05-13
ARTEMIS-2762 JournalRecord is using too much memory to track JournalFile updates

franz1981:


franz1981:
I've used a journal that exibith many cross updates file references and that's the difference I'm getting:
image
and OOM at 6 GB of heap. I need 16 GB of heap to be able to complete the test.
while with this PR:
image

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalRecord.java
```diff
@@ -33,7 +31,15 @@
 
    private final int size;
 
-   private List<Pair<JournalFile, Integer>> updateFiles;
+   // use this singleton to save using a separated boolean field to mark the "deleted" state
```
clebertsuconic, line 34: TBH I don't get the logic here..  

it seems that you are using some sort of auto-expandable int arrays instead of the previous List?


Can't we encapsulate the collections somehow it's a bit more understandable?


I won't be able to maintain this class any longer after this change. I want to get a clear udnerstanding and simplify the logic here.. even if we have to add an additional layer just to manage the auto-expanding collection here.

right now the data-structure is mixed with the logic here, and I don't understand what's going on.. perhaps it's just that I didn' get your logic and a video conference would help me understand it.

franz1981: > Can't we encapsulate the collections somehow it's a bit more understandable?

Good point, tomorrow I'll pack this in a more understandable version; in short, it contains 2 optimizations:
1) it accumulates any incoming addToUpdateFile while preserving the neg/pos count and size, if the last added file is unchanged
2) it amortize the footprint of a single entry ie 
Before: 24 bytes Pair + 16 bytes boxed Integer + 4 bytes array list entry
Now: 4 bytes JournalFile array slot + 8 bytes pos/neg count and size packed together

So, 44 bytes vs 12 bytes for each entry

franz1981: I've added a new commit to improve the readability of the proposed solution

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalRecord.java
```diff
@@ -29,38 +28,37 @@
 public class JournalRecord {
 
    /**
-    * Ordered collection of file updates ie the tuple (file, size, posCount):
-    * while appending new updates it tries to merge them into the last added, if possible,
-    * to reduce the amount of tuples.
+    * Ordered collection of (T, int, int) tuples with positive integers and not null T.
     */
-   private static final class FileUpdateList {
+   private static final class ObjIntIntArrayList<T> {
```
clebertsuconic, line 33: different file please?

clebertsuconic:
Nice job here!

just to let you know.. I made a small ammend into JournalRecord to change the order of the methods. as it made clearer the diff on git diff or IDE diff.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalRecord.java
```diff
@@ -45,26 +44,51 @@ public JournalRecord(final JournalFile addFile, final int size) {
       addFile.addSize(size);
    }
 
-   void addUpdateFile(final JournalFile updateFile, final int size) {
-      if (updateFiles == null) {
-         updateFiles = new ArrayList<>();
+   private void checkNotDeleted() {
+      if (fileUpdates == DELETED) {
```
clebertsuconic, line 48: I made a small ammend to this while I merged. I moved

I moved checkNotDeleted to the end of the file so the diffs are clearer.. method by method.


you dont' have to do anything.. just letting you know.

clebertsuconic: @franz1981  ^

PR #3136, opened 2020-05-14
ARTEMIS-2765 Fix parsing error in XMLUtil

Ceiu:
Fix a StringIndexOutOfBoundsException in XMLUtil while parsing system
properties in XML that contains curly braces or unterminated property
references


jbertram:
Nice work. Thanks for the commit!

PR #3138, opened 2020-05-15
ARTEMIS-2766 Not parsing application properties for duplicate property

clebertsuconic:


franz1981:
Bud sorry that I have been able to read this just now from my phone: the IdentityHashMap (on MSG_BODY_TYPES) I have used was meant to be used to search just MessageAnnotations: you should remove any (potentially) terminal section before the one you're searching to build the right one for your use case. 
In the current state of this PR if you hit a MessageAnnotations while searching for Data you will return with a false from the research of the Symbol, that's not correct.

I suggest, for simplicity, to build your own static final Identity hashmap with the correct terminal symbols for the data section (including the data section) and just expose an "anyData" search method instead of a more generic one. Another solution can be to remove the map and use some of the loops used by Robbie or Tim to do similar things, in order to exit when some terminal symbol beyond what you're expecting to find is being found.

clebertsuconic:
ApplicationProperties.class is on the list.. I don't see a problem?

clebertsuconic:
@franz1981 MSG_BODY_TYPES are the types that will possibly hold data. We need that list regardless.

there's an exposed method to look for the one I need. I think this is correct. and everything seems to be working fine.


I didn't release yet BTW as I was dealing with an issue.

clebertsuconic:
I see the issue.

you used that list to stop the loop... you didn't want to keep looking.

clebertsuconic:
It wouldn't been an issue because I'm starting the search right at the position of the applicaitonProperties. But I'm fixing it for correctness.

clebertsuconic:
https://github.com/apache/activemq-artemis/pull/3140 will contain a fix.

Still a WIP and I will finish tomorrow.

PR #3143, opened 2020-05-21
Fix libaio1 docker dependency and rename Dockerfile-ubuntu to Dockerfile-debian

zekeoptimo:
The current Dockerfile based on openjdk:8 does not build due to an unavailable version of libaio1 specified. In addition, the Dockerfile states it is Ubuntu, when in fact openjdk:8 is now Debian.

I removed the hardcoded 0.3.110-3 libaio1 version, renamed the Dockerfile and updated the scripts and readme.md to indicate what changed.


clebertsuconic:
awesome work...

only thing. it would be nice if you could create a JIRA.. and squash all these commits into one explaining all the changes... associated with the JIRA

```
ARTEMIS-XXX Docker Improvements

explain your changes here...
```


However, let me know if you can't do it.. and I will do it myself while merging your changes.

zekeoptimo:
Thank you! I have created a Jira ticket, squashes the commits and created a new PR.

clebertsuconic:
@zekeoptimo you didn't need to create a new PR.. push -f would have done it..

thanks!

zekeoptimo:
@clebertsuconic thanks! Yeah, sorry about that, I wasn't familiar with squashing the commits and such - thanks for the tip, that makes sense.

Question: for the Jira ticket I created, do I close that myself ?

Lastly, I discovered another bug in the Dockerfiles. The ANONYMOUS_LOGIN environment var is never used and the result is that anonymous access is always enabled. I'm going to file another ticket for that and get another PR going once I've tested everything - I also have another simple change that will give more control over how the cluster is created that I'd like to add.

PR #3144, opened 2020-05-21
ARTEMIS-2773 Dockerfile Improvements

zekeoptimo:
The current Dockerfile based on openjdk:8 does not build due to an unavailable version of libaio1 specified. It explicitly references version 0.3.110-3 of libaio1 however, it appears that 0.3.112-3 is the current stable version and when attempting to build the Dockerfile the following error occurs:

```
Step 9/20 : RUN groupadd -g 1000 -r artemis && useradd -r -u 1000 -g artemis artemis && apt-get -qq -o=Dpkg::Use-Pty=0 update && apt-get -qq -o=Dpkg::Use-Pty=0 install -y --no-install-recommends libaio1=0.3.110-3 && rm -rf /var/lib/apt/lists/*
 ---> Running in f56ff568c477
E: Version '0.3.110-3' for 'libaio1' was not found
``` 

In addition, I received this error attempting to run prepare-docker.sh:

`cp: cannot stat './{Dockerfile-centos,Dockerfile-ubuntu,docker-run.sh}': No such file or directory`
 
Finally, the Dockerfile states it is Ubuntu, when in fact openjdk:8 is now Debian, so the names of the scripts and the readme should reflect this.

### DIFF
--- a/artemis-docker/Dockerfile-debian
+++ b/artemis-docker/Dockerfile-debian
@@ -31,8 +31,7 @@ ENV CREATE_ARGUMENTS --user ${ARTEMIS_USER} --password ${ARTEMIS_PASSWORD} --sil
 # add user and group for artemis
 RUN groupadd -g 1000 -r artemis && useradd -r -u 1000 -g artemis artemis \
  && apt-get -qq -o=Dpkg::Use-Pty=0 update && \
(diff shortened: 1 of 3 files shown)


clebertsuconic:
I made one little ammend to the commit name before pushing. Take a look on master just for future PRs.

Thanks a lot.

PR #3146, opened 2020-05-21
ARTEMIS-2776 Improve Dockerfiles

zekeoptimo:
Fixes the Dockerfiles handling of ANONYMOUS_LOGIN and adds an
EXTRA_ARGS environment variable to allow specifying extra arguments
separate from the username, password and anonymous login options.


zekeoptimo:
This solves a couple of problems: 

1) The ANONYMOUS_LOGIN environment variable was not being used, and in testing I found that it would default to allow anonymous logging no matter what (due to the --silent command being forced to make a choice)

2) I believe this will make configuration easier, particular in an environment where usernames and/or passwords are kept as secrets. Allowing the ability to specify other parameters via EXTRA_ARGS allows specifying usernames and passwords separately from other configuration.

PR #3147, opened 2020-05-22
ARTEMIS-2778 AcceptorControl returns only transport parameters

brusdev:
Fix AcceptorControl to return all acceptor parameters.


jbertram:
The fix looks good, but it needs a test to verify and prevent future regressions.

clebertsuconic:
+1 on the test...

brusdev:
I added a unit test.

PR #3149, opened 2020-05-26
ARTEMIS-2781 return only local queue names for AddressControl.getQueu…

jbertram:
…eNames()


clebertsuconic:
@jbertram this will need to be reverted.

Not only the semantic of getQueues is changing, there are methods inside the broker needing this.

For instance destroyTopic tests are failing because only local queues are now destroyed


I suggest a new method getLocalQueues is created and the current returns the combined version.

clebertsuconic:
followed up with https://github.com/apache/activemq-artemis/pull/3151

clebertsuconic:
and then followed up with #3152

PR #3151, opened 2020-05-28
ARTEMIS-2781 Restoring original semantic on AddressControl.getQueue

clebertsuconic:


jbertram:
Take a look at #3152.

jbertram:
To be clear, I don't think the original semantic should be restored due to the discrepancy with `ActiveMQServerControl.getQueueNames()`. I updated the Jira description to be more clear.

PR #3153, opened 2020-05-29
ARTEMIS-2649 always over-write ORIG message props

jbertram:
ORIG message propertes like _AMQ_ORIG_ADDRESS are added to messages
during various broker operations (e.g. diverting a message, expiring a
message, etc.). However, if multiple operations try to set these
properties on the same message (e.g. administratively moving a message
which eventually gets sent to a dead-letter address) then important
details can be lost. This is particularly problematic when using
auto-created dead-letter or expiry resources which use filters based on
_AMQ_ORIG_ADDRESS and can lead to message loss.

This commit simply over-writes the existing ORIG properties rather than
preserving them so that the most recent information is available.

### DIFF
--- a/docs/user-manual/en/copied-message-properties.md
+++ b/docs/user-manual/en/copied-message-properties.md
@@ -1,0 +1,37 @@
+# Properties for Copied Messages
+
+There are several operations within the broker that result in copying a
+message. These include:
+
+- Diverting a message from one address to another.
+- Moving an expired message from a queue to the configured `expiry-address`
+- Moving a message which has exceeded its `max-delivery-attempts` from a queue
+  to the configured `dead-letter-address`
+- Using the management API to administratively move messages from one queue to
+  another
+
+When this happens the body and properties of the original message are copied to
+a new message. However, the copying process removes some potentially important
+pieces of data so those are preserved in the following special message
+properties:
+
+- `_AMQ_ORIG_ADDRESS`
+
+  a String property containing the *original address* of the message
+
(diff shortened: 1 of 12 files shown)


docs/user-manual/en/copied-message-properties.md
```diff
@@ -0,0 +1,37 @@
...
+  to the configured `dead-letter-address`
+- Using the management API to administratively move messages from one queue to
+  another
+
+When this happens the body and properties of the original message are copied to
+a new message. However, the copying process removes some potentially important
+pieces of data so those are preserved in the following special message
+properties:
+
+- `_AMQ_ORIG_ADDRESS`
```
clebertsuconic, line 18: These are also documented in AMQP. 

Perhaps you could merge the content of that section here, and add a note about message annotations for AMQP messages?

jbertram: They're referenced in the AMQP documentation but only insofar as how they are translated. I think the AMQP doc is where that should be.

clebertsuconic: Can you just a link then, stating there are some small differences with AMQP messages perhaps?

clebertsuconic:
There are two test failures on the whole testsuite ATM:


- org.apache.activemq.artemis.tests.integration.divert.DivertTest.testSingleDivertWithExpiry
- org.apache.activemq.artemis.tests.integration.amqp.DLQAfterExpiredMessageTest.testDoubleTransfer

PR #3154, opened 2020-05-29
ARTEMIS-2785 io.netty.util.internal.OutOfDirectMemoryError during uncompress

franz1981:
Netty's `UnreleasableByteBuf::readBytes(int length)` allocates a direct `ByteBuf` without `Cleaner` that can leak and cause OOM on `ChannelBufferWrapper::readBytes(final int length)`

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQBuffer.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQBuffer.java
@@ -800,19 +800,6 @@ public interface ActiveMQBuffer extends DataInput {
    String readUTF();
 
    /**
-    * Transfers this buffer's data to a newly created buffer starting at
-    * the current {@code readerIndex} and increases the {@code readerIndex}
-    * by the number of the transferred bytes (= {@code length}).
-    * The returned buffer's {@code readerIndex} and {@code writerIndex} are
-    * {@code 0} and {@code length} respectively.
-    *
-    * @param length the number of bytes to transfer
-    * @return the newly created buffer which contains the transferred bytes
-    * @throws IndexOutOfBoundsException if {@code length} is greater than {@code this.readableBytes}
-    */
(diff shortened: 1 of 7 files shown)


franz1981:
Although it seems like a Netty issue, it's not: the problem is on `ChannelBufferWrapper` because it's using `UnreleasableByteBuf::readBytes(int length)` and wrapping its returned `ByteBuf` (which nature depends on the wrapped allocator) into an `UnreleaeasableByteBuf` that would prevent it to be released and correctly deallocated. 

This PR is both:

- avoiding using an intermediate `readBytes` + `toByteBuffer` to read into a `byte[]` the data
- removing the wrong method on `ChanneBufferWrapper` given that isn't used anymore

clebertsuconic:
perhaps we should consider release 2.13.1 soon because of this?

sebthom:
@franz1981 @clebertsuconic is this an issue that was introduced in 2.12.0? We are currently running 2.11.0 in production and thinking about upgrading to 2.13.0

clebertsuconic:
@sebthom  this has been an issue since at least 2.6.0, possibly longer as I believe I copied the code from somewhere else when I committed e86acd4824d3a94d124693fbd7887a1f1b9b072c

sebthom:
@clebertsuconic Thanks for the update. I think then we are safe in our scenario to upgrade. We did not experience a memory leak so far. Are you still considering a 2.13.1 release?

clebertsuconic:
@sebthom I actually am...

although I'm not branching 2.13.. I will release from master.. and if there are features it promotes the number to 2.14.0...

Since there's nothing breaking any compatibility, I'm considering just releasing from master.

michaelandrepearce:
@clebertsuconic we have a feature needed just finishing it up. As had a prod issue. Can you hold any release for a day or two so i can get it in please

michaelandrepearce:
@clebertsuconic as mentioned before next release id like https://github.com/apache/activemq-artemis/pull/3169 included (now pr'd)

PR #3164, opened 2020-06-04
ARTEMIS-1322 Add a ServerLocator option to ignore topology for load-balancing

gaohoward:
(2nd commit) Disable wait for topolocy when ignore topology for
load-balancing option is true.


jbertram:
You might still want the topology for failover even if you don't want to use it for load balancing.

gaohoward:
yes that's a problem.

clebertsuconic:
@howardgao you would have to come up with a different implementation. I don't think we should go with this implementation.

gaohoward:
I'm thinking of adding a similar option 'useClientFailover' to allow user to disable client side failover. If the user disable both loadbalancing and failover (as mostly useful in cloud envirnoment) then don't care about the topology waiting.

jbertram:
@howardgao, don't we already have that option via the `ha` URL parameter? If `ha=false` then the client won't failover.

gaohoward:
I found ha=false not quite disable the failover thing. The connection retry still looking for a backup in the topology regardless whether ha is on or off. Anyway I saw the Jira is closed. Maybe we can drop this PR?

PR #3165, opened 2020-06-04
ARTEMIS-2790 Fix no examples documentation in the bin archive

brusdev:
Move the markdown generator for the examples to artemis-distribution.

### DIFF
--- a/artemis-distribution/pom.xml
+++ b/artemis-distribution/pom.xml
@@ -287,4 +287,34 @@
       </plugins>
    </build>
 
+   <profiles>
+      <profile>
+         <id>release</id>
+         <build>
+            <plugins>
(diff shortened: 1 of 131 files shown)


artemis-distribution/pom.xml
```diff
@@ -287,4 +287,34 @@
       </plugins>
    </build>
 
+   <profiles>
+      <profile>
+         <id>release</id>
```
clebertsuconic, line 292: why just on the release?

if I generate a binary, it would be nice to have the HTMLs as well, not just as part of the release.

brusdev: I left in the release profile as it was before, but I haven't any concerns to change it.

PR #3167, opened 2020-06-04
ARTEMIS-2793 - Only log quorum voting occurred when it actually does …

andytaylor:
…happen

https://issues.apache.org/jira/browse/ARTEMIS-2793

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/qourum/QuorumVoteServerConnect.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/qourum/QuorumVoteServerConnect.java
@@ -141,7 +141,11 @@ public class QuorumVoteServerConnect extends QuorumVote<ServerConnectVote, Boole
       return votesNeeded == 0;
    }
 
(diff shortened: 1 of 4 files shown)


clebertsuconic:
LGTM

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/qourum/QuorumVoteServerConnect.java
```diff
@@ -141,7 +141,11 @@ public synchronized Boolean getDecision() {
       return votesNeeded == 0;
    }
 
-   public void await(int latchTimeout, TimeUnit unit) throws InterruptedException {
+   public void await(int latchTimeout, TimeUnit unit, int maxClusterSize) throws InterruptedException {
```
franz1981, line 144: IMO this check should be moved into `QuorumManager`, given that:
- `maxClusterSize` is already owned and managed by it
- QuorumVoteServerConnect isn't aware of how many cluster members could perform the vote, just the expected number to unblock the quorum

PR #3168, opened 2020-06-04
ARTEMIS-2793 quorum logging implies it happens when single pair

franz1981:
This should replace https://github.com/apache/activemq-artemis/pull/3167


franz1981:
Given that #3167 was "LGTM" by @clebertsuconic and this one is doing exactly the same + adding an additional log info I'm going to merge it

PR #3169, opened 2020-06-04
ARTEMIS-2787 - Add ability to disable and enable a queue

michaelandrepearce:
Add feature
Add tests
Add docs
Add missing bits noticed in ring-size


clebertsuconic:
I am trying to understand what disabling a queue would mean. There's already pause on a queue.

clebertsuconic:
The only outcome that I saw from setting enabled or disabled is the boolean changed itself.. being a metadata that serves no purpose to the broker itself.

wouldn't be best to implement a generic meta-data for this kind of purpose?

clebertsuconic:
unless you tell me there's a function for disabling the queue on the broker... I would prefer to serve your feature with a generic metadata, that you set at your own desire. as I couldn't find a function for the broker itself.

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/QueueAttributes.java
```diff
@@ -103,6 +105,8 @@ public void set(String key, String value) {
             setAutoDeleteMessageCount(Long.valueOf(value));
          } else if (key.equals(RING_SIZE)) {
             setRingSize(Long.valueOf(value));
+         } else if (key.equals(ENABLED)) {
+            setRingSize(Long.valueOf(value));
          }
```
jbertram, line 110: Shouldn't this be `setEnabled(Boolean.valueOf(value));`?

jbertram:
@clebertsuconic, look at `QueueImpl.route()`. That's where the functionality is actually implemented.

jbertram:
I think it would be worth making an explicit statement in the documentation that disabling all the queues on an address means that any message sent to that address will be silently dropped. I'm sure this point will be clear to savvy users, but others may not connect the dots as readily.

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/QueueAttributes.java
michaelpearce-gain: yes good catch

michaelpearce-gain:
@clebertsuconic re
`I am trying to understand what disabling a queue would mean. There's already pause on a queue.`

pause only works when no consumers, in this case we need to stop messages entering the queue, even if consumers are attached.

michaelpearce-gain:
@jbertram will add extra line to the docs to mention that.

michaelpearce-gain:
@clebertsuconic its used by the queue to disable it thats what the feature adds, as such its not meta-data.

docs/user-manual/en/address-model.md
```diff
@@ -527,6 +527,34 @@ Open the file `<broker-instance>/etc/broker.xml` for editing.
...
+<addresses>
+   <address name="foo.bar">
+      <multicast>
+         <queue name="orders1" enabled="false"/>
+      </multicast>
+   </address>
+</addresses>
+```
+
+Warning: Disabling all the queues on an address means that any message sent to that address will be silently dropped.
```
michaelpearce-gain, line 555: @jbertram  added doc in line to be explicit if someone disables all queues.

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/QueueAttributes.java
michaelpearce-gain: fixed

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/addressing/AddressingTest.java
```diff
@@ -268,6 +268,62 @@ public void testPurgeOnNoConsumersFalse() throws Exception {
       Wait.assertEquals(1, server.locateQueue(queueName)::getMessageCount);
    }
 
+
+   @Test
+   public void testQueueEnabledDisabled() throws Exception {
```
michaelpearce-gain, line 273: @clebertsuconic see here this is the actual functionality test, where enabling / disabling the queue.

(other tests added are just around checking config works and updates etc should reload etc occur)

clebertsuconic:
I missed the QueueImpl::route, I only saw the change the PostOfficeImpl... 

it makes sense as a feature.

michaelpearce-gain:
@clebertsuconic cool :) , anything else needed before merge? I fixed the bit justin spotted, and added the warning to docs regarding if someone disabled all.

michaelpearce-gain:
@clebertsuconic i dont follow sorry. its stored in the queue config, as such if you update the queue config, as like any other queue config change it will be persisted.

clebertsuconic:
@michaelandrepearce queue.pause / queue.resume.. there's an option on the JMX Console method that will store the value on the journal (not on the config)


I was just asking if wouldn't make sense to do the same.

michaelpearce-gain:
so thats taken care of if you update config via jmx console

https://github.com/apache/activemq-artemis/pull/3169/files#diff-1fae8ea746270e6f0b1f0c99a6148ab2

michaelpearce-gain:
i see what you mean, i think holding state somewhere else would be dangerous, we could though do the updae via post office, as such it will update the binding and then queue that way, as such it will go via same route as config change


e.,g. 
         server.getPostOffice().updateQueue(queue.getQueueConfiguration().setEnabled(false));

michaelpearce-gain:
Ill send that in two ticks.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
```diff
@@ -618,6 +618,52 @@ public boolean isPurgeOnNoConsumers() {
...
+   @Override
+   public void enable() throws Exception {
+      if (AuditLogger.isEnabled()) {
+         AuditLogger.enable(queue);
+      }
+      checkStarted();
+
+      clearIO();
+      try {
+         server.getPostOffice().updateQueue(queue.getQueueConfiguration().setEnabled(true));
```
michaelpearce-gain, line 645: now it updates the config, as such persists in the queue binding.

clebertsuconic: nice one... I will run tests and merge it

clebertsuconic:
@michaelandrepearce / @michaelpearce-gain a test you wrote is failing 100% of the time. Can you take a look?

org.apache.activemq.artemis.tests.integration.client.SessionTest::testCreateQueue

michaelpearce-gain:
@clebertsuconic  fixed , the new case is only valid for new createQueue (not legacy , e.g the new unified createQueue(QueueConfig))

clebertsuconic:
I'm just confirming one non related failure on my last run. if everything ok I will merge this today.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/SessionQueueQueryResponseMessage_V3.java
```diff
@@ -475,6 +501,11 @@ public boolean equals(Object obj) {
             return false;
       } else if (!ringSize.equals(other.ringSize))
          return false;
+      if (enabled == enabled) {
```
ehsavoie, line 504: This is wrong I think

michaelandrepearce: good spot, https://github.com/apache/activemq-artemis/pull/3174 fix here.

PR #3170, opened 2020-06-05
ARTEMIS-2788 clear openwire producer exchange, created on send, on pr…

gtully:
…oducer close event


gtully:
follow up, additional state created on first message send, private to connection, that also needs to be cleared.

PR #3172, opened 2020-06-09
ARTEMIS-2797 - Reset queue properties by unsetting them in broker.xml

jsmucr:
Now it is possible to remove queue filter by setting it to an empty string or null.
Before this commit the server would just ignore such request.

EDIT: This pull request turned out to cover the ability to reset all queue properties, not just the filter string.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -16,22 +16,6 @@
  */
 package org.apache.activemq.artemis.core.postoffice.impl;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.Executor;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
-
 import org.apache.activemq.artemis.api.core.ActiveMQAddressDoesNotExistException;
 import org.apache.activemq.artemis.api.core.ActiveMQAddressFullException;
 import org.apache.activemq.artemis.api.core.ActiveMQDuplicateIdException;
@@ -79,6 +63,7 @@ import org.apache.activemq.artemis.core.server.RoutingContext;
 import org.apache.activemq.artemis.core.server.cluster.impl.MessageLoadBalancingType;
 import org.apache.activemq.artemis.core.server.group.GroupingHandler;
 import org.apache.activemq.artemis.core.server.impl.AddressInfo;
+import org.apache.activemq.artemis.core.server.impl.QueueConfigurationUtils;
 import org.apache.activemq.artemis.core.server.impl.QueueManagerImpl;
 import org.apache.activemq.artemis.core.server.impl.RoutingContextImpl;
 import org.apache.activemq.artemis.core.server.management.ManagementService;
@@ -96,6 +81,23 @@ import org.apache.activemq.artemis.utils.CompositeAddress;
 import org.apache.activemq.artemis.utils.UUIDGenerator;
 import org.apache.activemq.artemis.utils.collections.TypedProperties;
 import org.jboss.logging.Logger;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * This is the class that will make the routing to Queues and decide which consumer will get the messages
@@ -632,7 +634,7 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
             boolean changed = false;
 
             //validate update
-            if (queueConfiguration.getMaxConsumers() != null && queueConfiguration.getMaxConsumers().intValue() != Queue.MAX_CONSUMERS_UNLIMITED) {
+            if (queueConfiguration.getMaxConsumers() != null && queueConfiguration.getMaxConsumers() != Queue.MAX_CONSUMERS_UNLIMITED) {
                final int consumerCount = queue.getConsumerCount();
                if (consumerCount > queueConfiguration.getMaxConsumers()) {
                   throw ActiveMQMessageBundle.BUNDLE.invalidMaxConsumersUpdate(queueConfiguration.getName().toString(), queueConfiguration.getMaxConsumers(), consumerCount);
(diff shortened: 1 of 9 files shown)


michaelandrepearce:
Can a xml config update test be made to cover reload from xml.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -689,7 +689,10 @@ public QueueBinding updateQueue(QueueConfiguration queueConfiguration) throws Ex
                queue.setDelayBeforeDispatch(queueConfiguration.getDelayBeforeDispatch().longValue());
             }
             Filter filter = FilterImpl.createFilter(queueConfiguration.getFilterString());
-            if (filter != null && !filter.equals(queue.getFilter())) {
+            if ((filter == null) && (queue.getFilter() != null)) {
```
michaelandrepearce, line 692: Should think about applying the same to all, e.g. if now null set back to default for address setting (or static default)

jsmucr: I've added the config reload test.

I'm not sure about what you mean by the default behavior. Shall I add some sort of constant (like `DEFAULT_FILTER`) or just document the fact that there's no filter by default and that setting it to `null` (by removing it from the configuration) or to `""` reverts it back to the default value?

michaelandrepearce: I was meaning other queue configurations that you can update e.g like the one immediately above delaybeforedispatch. E.g. if change of behaviour from currently if null do not update, now youre proposing to make it if null set back to default which for filter is null. We would want that behaviour consistent. Else we end up with some ignoring and some reverting. 

Its very important behaviour is consistent for all, so either all do nothing when new value is null (ignore) or all update back to their defaults (being either the default at address setting or where a default constant its used)

tests/integration-tests/src/test/resources/reload-queue-filter-updated-empty.xml
```diff
@@ -0,0 +1,43 @@
...
+
+      <acceptors>
+         <acceptor name="artemis">tcp://0.0.0.0:61616</acceptor>
+      </acceptors>
+
+      <addresses>
+         <address name="myFilterQueue">
+            <anycast>
+               <queue name="myFilterQueue">
+                  <filter string=""/>
```
michaelandrepearce, line 37: This isnt filtering. I would expect the test to be a real filter string not an empty one. Empty filter is same as no filter so this isnt validating much

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/RedeployTest.java
```diff
@@ -287,6 +287,99 @@ public void run() {
...
+      } finally {
+         embeddedActiveMQ.stop();
+      }
+   }
+
+   @Test
+   public void testRedeployRemoveFilter() throws Exception {
+      doTestRemoveFilter(RedeployTest.class.getClassLoader().getResource("reload-queue-filter-updated-empty.xml"));
+      doTestRemoveFilter(RedeployTest.class.getClassLoader().getResource("reload-queue-filter-removed.xml"));
+   }
```
michaelandrepearce, line 381: I would expect the test to first validate a filter works and queueimpl has filter matching with a real filter, then reload config and validate both the queueimpl is null and now no filtering is applied on flow. To vaildate update does indeed remove a filter if previously set.

tests/integration-tests/src/test/resources/reload-queue-filter-updated-empty.xml
jsmucr: The `doTestRemoveFilter` method first applies the `reload-queue-filter.xml` configuration that has the filter set up correctly, then tests if it works and after that it applies the intended modification (either `reload-queue-filter-updated-empty.xml` or `reload-queue-filter-removed.xml`).

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
jsmucr: I see. I'll take look at that.

jsmucr: Hm, I must have overlooked the fact that the other properties behave the same too.
Is the behavior I propose a correct one? Or is there a reason for ignoring null values?

Now that it's obvious - even to me :-) - that the issue is a lot broader, shall I continue with implementing the rest here or would that be a whole new pull request?

My use case in an automatic cluster configuration deployment and the current behavior sort of breaks the idea.

jsmucr: Clearing the value using `""` does not work, at least not for a filter.
```
//...
public static Filter createFilter(final SimpleString filterStr) throws ActiveMQException {
      if (filterStr == null || filterStr.length() == 0) {
         return null;
      }
// ...
```
So there would have to be a breaking change anyway, just to achieve this.

I understand the issue with resetting the whole configuration. Maybe there could be a toggle to enable this.

I'm working for a cloud-based EDI provider and our ultimate goal is to have all of our infrastructure configurable via Ansible (or similar tools) so that in case of a major failure there'd be no need to perform too many manual steps in order to rebuild the service from scratch. That makes the Ansible repository the only place to store our configuration, and therefore it would be highly beneficial if I could redeploy the Artemis configuration from there. :-) It would also open the way to create a bit more user friendly tool to design the MQ related infrastructure.
I'd be happy to hear about the way you do it in your company. :-)

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -647,66 +649,121 @@ public QueueBinding updateQueue(QueueConfiguration queueConfiguration) throws Ex
...
+                    ? ActiveMQDefaultConfiguration.getDefaultMaxQueueConsumers()
+                    : queueConfiguration.getMaxConsumers();
+            if (queue.getMaxConsumers() != newMaxConsumers) {
                changed = true;
-               queue.setMaxConsumer(queueConfiguration.getMaxConsumers());
+               queue.setMaxConsumer(newMaxConsumers);
             }
-            if (queueConfiguration.getRoutingType() != null && queue.getRoutingType() != queueConfiguration.getRoutingType()) {
+            // routingType
+            final RoutingType newRoutingType = queueConfiguration.getRoutingType() == null
```
michaelpearce-gain, line 662: So to get defaults, should be using the address settings.

Theres actually a common method that does this thats used else where, as its quite common

     QueueConfigurationUtils.applyDynamicQueueDefaults(queueConfiguration, addressSettingsRepository.getMatch(queueConfiguration.getAddress().toString()));

This then will correctly set defaults on everything in the queueConfiguration

michaelpearce-gain: e.g. you could just do this once before the section of atomic updates.

michaelandrepearce:
Looking alot better. I note you still have this as draft pr. Is there more outstanding?

jsmucr:
@michaelandrepearce There are no tests (apart from the one testing filters) validating the new behavior. There's still a lot of _fun_ waiting for me. :-)

jsmucr:
Waiting for someone to check this before I continue with docs.

michaelpearce-gain:
@jsmucr LGTM

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3954,7 +3954,7 @@ public QueueConfiguration getQueueConfiguration() {
          .setAddress(address)
          .setId(id)
          .setRoutingType(routingType)
-         .setFilterString(filter.getFilterString())
+         .setFilterString(filter == null ? null : filter.getFilterString())
```
michaelpearce-gain, line 3957: can you rebase, this is already fixed in master

michaelpearce-gain:
@jsmucr can you rebase and squash commits

michaelpearce-gain:
@jsmucr thanks for the contribution.

PR #3175, opened 2020-06-09
fixed typo in code example

Faldrian:
Fixed a typo which could lead to some frustration if people just copy the string and try it.


clebertsuconic:
thanks

PR #3176, opened 2020-06-09
ARTEMIS-2798 expiration for AMQP msgs not reloaded

jbertram:
@franz1981 or @clebertsuconic could either of you review this to see if I'm doing this right. Seems like a simple change, but want to make sure.


franz1981:
Goog catch! And we were not covered on this..so good to have a test too...please check if the message isn't getting fully decoded before rerouted to the queue after journal loading (as I suspect)

PR #3177, opened 2020-06-11
ARTEMIS-2801 Fix ByteUtil.getHumanReadableByteCount() giving inconsistent results

jsmucr:
HumanReadableByteCountTest test is no longer failing under environments with locales defining different number format.
The function now returns values according to the Locale.ROOT locale specification.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java
@@ -21,6 +21,7 @@ import java.io.StringWriter;
 import java.nio.ByteBuffer;
 import java.nio.ReadOnlyBufferException;
 import java.util.Arrays;
+import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java
```diff
@@ -449,6 +450,7 @@ public static String getHumanReadableByteCount(long bytes) {
          i++;
       }
 
-      return String.format("%.1f%sB", bytes / BYTE_MAGNITUDES[i], BYTE_SUFFIXES[i]);
+      // Set locale to language/country neutral to avoid different behavior for non-US users
```
clebertsuconic, line 453: Nice!!! @ehsavoie  you posted about this on the dev list the other day, right?

jbertram:
Doesn't this mean that the "human readable format" won't be formatted according to the JVM's default locale? It seems like this is the opposite of what we want. Wouldn't we want to change the *test* so that it formats the assertion value according to the default locale?

jbertram:
I sent #3178 to demonstrate what I mean.

ehsavoie:
@jbertram that was the point I tried to raise in my email. What do we want ? ;)

jbertram:
@ehsavoie, I just responded to your email on the dev list.

jsmucr:
@jbertram I too thought about that but for me it seemed a bit tough and not very transparent to cover various locales in a test. What I mean is that you'd either have to enumerate the possible values, or compare the method return value to something you cannot really see at the moment of testing.
Maybe I'm not fully awake yet but wouldn't that mean to pretty much copy-paste the method functionality to the test?

clebertsuconic:
@jsmucr I think should be closed, as the test on #3178 would fail.

If you want to still discuss this, please reopen the PR.

clebertsuconic:
beware that I merged this by accident, then I reverted it.

PR #3178, opened 2020-06-11
ARTEMIS-2801 make HumanReadableByteCountTest locale sensitive

jbertram:


clebertsuconic:
the broker.xml is written in English. I would keep the locale always in english on the xml.

mixing it up would make it weird. eventually moving the xml from one place to another would require you to "translate" the config.

jbertram:
This really isn't about configuration. It's just about how a particular log message is formatted using `String.format`.

clebertsuconic:
Ok, I thought this was about interpreting human formats on the XML. I misunderstood it.

PR #3161, opened 2020-06-03
Java 11

ehsavoie:
For testing purpose only.
DO NOT MERGE or REVIEW

PR #3179, opened 2020-06-11
ARTEMIS-2109: enable building Artemis on JDK 11+

gemmellr:
This allows building on JDK11+, and passes the current Travis CI job tasks on JDK 8, 11, and 14 (Travis config extended to add those).

Notes:
This incorporates Emmanuel's changes from #3161 as well as various additions from me.

The build targets 8 when using 8, which is still required to be used for releases, however it targets Java 11 when using Java 11+. This is because the '--release 8' compiler config isnt usable due to use of things like Unsafe in the codebase. Equally, just using --target 1.8 equivalent config while building with JDK11+ still wouldnt produce bytecode that works on 8 due to using the 11 signatures. Not unless we modify the source with type casting etc to force it to that is.

I think this is in shape to be merged and can then be iterated over time to improve rather than it continuing along in a branch any longer, hence the PR.


gemmellr:
Added missing ARTEMIS-2109 ref to the title.

PR #3180, opened 2020-06-12
ARTEMIS-2705 Test Artemis on aarch64 CPU architecture

martin-g:
1) Update Netty to 4.1.50 that includes aarch64 binaries

2) Use Maven wrapper

This way it will use the preferred version of Maven.
Also it seems Maven is not coming pre-installed on ARM64

3) Update netty-tcnative to 2.0.31.Final

4) Update Maven to 3.6.2 because artemis-console fails with 3.3.9

5) travis_wait the build of 'examples'

because it takes more time and Travis kills it: https://travis-ci.org/github/martin-g/activemq-artemis/jobs/697514020

"The job exceeded the maximum time limit for jobs, and has been terminated."

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ matrix:
       env:
         - EXAMPLES_PROFILE="release"
     - os: linux
+      jdk: openjdk8
+      arch: arm64
+      env:
+        - EXAMPLES_PROFILE="release"
+    - os: linux
       jdk: openjdk11
       env:
         - EXAMPLES_PROFILE="noRun"
@@ -24,9 +29,9 @@ before_install:
 # By setting anything to org.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED we are disabling libaio loading on the testsuite
 script: 
 - set -e
-- mvn -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Djdk8-errorprone -Pfast-tests -Pextra-tests -Ptests-CI -B install -q
+- ./mvnw -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Dnetty-transport-native-epoll-classifier=linux-`uname -m` -Djdk8-errorprone -Pfast-tests -Pextra-tests -Ptests-CI -B install -q
 - cd examples
-- mvn install -P${EXAMPLES_PROFILE} -B -q
+- ../mvnw install -P${EXAMPLES_PROFILE} -B -q
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


clebertsuconic:
can you rebase it?


What is the point of testing the new platform? there are no natives in the testsuite for travis (we disable libaio on travis).

.travis.yml
```diff
@@ -12,9 +15,9 @@ before_install:
 # By setting anything to org.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED we are disabling libaio loading on the testsuite
 script: 
 - set -e
-- mvn -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Pfast-tests -Pextra-tests -Ptests-CI -B install -q
+- ./mvnw -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Dnetty-transport-native-epoll-classifier=linux-`uname -m` -Pfast-tests -Pextra-tests -Ptests-CI -B install -q
```
michaelandrepearce, line 18: Can we keep this as plain mvn no need for mvnw

emiliofernandes:
> can you rebase it?

Rebased!

> What is the point of testing the new platform? there are no natives in the testsuite for travis (we disable libaio on travis).

In that case maybe there is no point in testing on arm64.
Should I close this PR or enable libaio just on arm64 ?

.travis.yml
emiliofernandes: There is no `mvn` pre-installed on the ARM64 VMs at TravisCI.

martin-g: There is no `mvn` pre-installed on ARM64 VMs at TravisCI

martin-g:
> can you rebase it?

Rebased!
 
> What is the point of testing the new platform? there are no natives in the testsuite for travis (we disable libaio on travis).

Should I close this PR ?
May end goal is to make
`./modules/system/layers/base/org/apache/activemq/artemis/journal/main/lib/linux-x86_64/libartemis-native-64.so` available for `aarch64` too.

clebertsuconic:
@martin-g you should probably close this then, as we don't test the .so on travis. (environment issues on travis were killing us, so we disabled it).


The issue really is how to build it?


Someone willing to use artemis-native on aarch64 should do the cmake .; make and place the natives in the right place.

Now if you have a way we can release with aarch64, you can probably look at the compilations scripts on artemis-native if you can come up with a way to do it in the future.

clebertsuconic:
like, I don't have access an aarch64 box so i could build.

Again, someone willing to use it, can just compile it themselves in 10 seconds.

martin-g:
> The issue really is how to build it?

No. Custom build is easy.
The idea is to provide the aarch64 binary within the Wildfly distribution.
I will see what I can do!

martin-g:
https://github.com/apache/activemq-artemis-native/pull/8

PR #3181, opened 2020-06-12
ARTEMIS-2804 Human readable timestamp is incorrectly parsed in web console

avi5kdonrh:
The web console displays the day of the week instead of the day of the month in the timestamp column while browsing a queue from the web console.


jbertram:
The commits should be squashed and the comment should be removed and added to the commit message as it only makes sense in the context of the change.

jbertram:
@avi5kdonrh, can you update the commit message to follow [the pattern outlined in the hacking guide](https://github.com/apache/activemq-artemis/blob/master/docs/hacking-guide/en/code.md#commitMessageDetails)?

jbertram:
@avi5kdonrh, don't worry about the commit message. I fixed it during merge. Thanks for the fix!

avi5kdonrh:
@jbertram, @franz1981  Thank you very much. Next time I'll keep the guidelines in mind.

PR #3183, opened 2020-06-15
ARTEMIS-2109: require JDK 8 for release

gemmellr:
Requires JDK 8 for the 'release' profile. This is the same requirement as before, but the existing profile check only set an '8 minimum' requirement. Now that it can build on 11+ (but, targetting 11 while doing so currently) unlike before, the check needs to be specifically 8-only. 

Enforces use of JDK8 or 11+ in regular non-release builds, giving 9 or 10 users a clearer error.

Renames profiles for consistency and clarity.

Updates CI to not use the release profile for examples on 11 or 14, only 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,19 @@
 language: java
 install: true
-jdk:
-  - openjdk8
-  - openjdk11
-  - openjdk14
+matrix:
+  include:
+    - os: linux
+      jdk: openjdk8
+      env:
+        - EXAMPLES_PROFILE="release"
+    - os: linux
+      jdk: openjdk11
+      env:
+        - EXAMPLES_PROFILE="noRun"
+    - os: linux
+      jdk: openjdk14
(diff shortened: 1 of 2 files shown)


gemmellr:
More changes following #3179

gemmellr:
I reverted the plugin version changes on master in https://github.com/apache/activemq-artemis/commit/61e920d010f94894b04a1b43a3880a0cdbdd629d.

I had noticed a NPE failure building the assembly during "mvn clean package" (but not "mvn clean install"), which some search suggeted relates to use of the enforcer plugin 1.4.1 in modules with pom packaging. That was the effective change from removing the 1.4 version definitions as the pluginManagement already has a 1.4.1 entry.

Must be some sort of weird interaction with other bits in use as well, since we use 1.4.1+assembly in lots of other places, but anyway...I've restored that bit to how it was before to put things as they were.

PR #3185, opened 2020-06-16
ARTEMIS-2806 deployQueue missing address argument

qihongxu:
In ActiveMQServerControlImpl, deployQueue method is missing address argument which results in creating non-matched addresses and queues. (For example, using deployQueue to create a subscriber A_0 under existing address A, it finally returns a new topic A_0 with a subscriber A_0 )

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
@@ -952,7 +952,7 @@ public class ActiveMQServerControlImpl extends AbstractControl implements Active
       clearIO();
       try {
          server.createQueue(new QueueConfiguration(name)
-                            .setName(name)
+                            .setAddress(address)
                             .setFilterString(filterStr)
                             .setDurable(durable));
       } finally {


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -952,6 +952,7 @@ public void deployQueue(final String address,
       clearIO();
       try {
          server.createQueue(new QueueConfiguration(name)
+                            .setAddress(address)
                             .setName(name)
```
jbertram, line 956: You can actually remove `setName(name)` as this is already taken care of by passing `name` to the constructor. The bug here is that `setName(name)` is actually called instead of `setAddress(address)`.

jbertram:
We certainly want to fix this bug, but it's worth noting that all versions of `deployQueue` are deprecated so I recommend you migrate to `createQueue` as soon as possible. Furthermore, there are no actual tests for `deployQueue` which is almost certainly why this bug was not caught.

qihongxu:
> We certainly want to fix this bug, but it's worth noting that all versions of `deployQueue` are deprecated so I recommend you migrate to `createQueue` as soon as possible. Furthermore, there are no actual tests for `deployQueue` which is almost certainly why this bug was not caught.

Thanks for your advice! This bug is caught coincidently when we tried to upgrade some artemis servers to newer version. Although deployQueue is deprecated, I checked the commit log and found it was modified 2 months ago. We will migrate to createQueue in our projects. This pr is just for someone that might still use legacy methods as we do. :)

PR #3186, opened 2020-06-16
 ARTEMIS-2109: fix ErrorProne for JDK8, move to profile

gemmellr:
Fixes the updated configuration for ErrorProne so it can actually run on JDK8.

Also moves it to a specific profile for optional use on JDK8, as the 2.4.0+ javac plugin approach requires forking the compiler on JDK8, which is slow and adds a couple of minutes to the build from all the modules. On JDK8 Error Prone is activated on request. On JDK11+ it is still enabled by default. 

Updates the CI job config as well to still enable ErrorProne for JDK8  (i.e. on all JDKs being used).

Follows the changes in #3179 (and #3183).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ before_install:
 # By setting anything to org.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED we are disabling libaio loading on the testsuite
 script: 
 - set -e
-- mvn -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Pfast-tests -Pextra-tests -Ptests-CI -B install -q
+- mvn -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Djdk8-errorprone -Pfast-tests -Pextra-tests -Ptests-CI -B install -q
 - cd examples
 - mvn install -P${EXAMPLES_PROFILE} -B -q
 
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -24,7 +24,7 @@ before_install:
 # By setting anything to org.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED we are disabling libaio loading on the testsuite
 script: 
 - set -e
-- mvn -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Pfast-tests -Pextra-tests -Ptests-CI -B install -q
+- mvn -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Djdk8-errorprone -Pfast-tests -Pextra-tests -Ptests-CI -B install -q
```
clebertsuconic, line 27: didn't you mean -Pjdk8-errorprone?


-P

gemmellr:
No, as that would enable the JDK8 arguments on 11+ and break them. The profile can be activated explicitly with -P or by JDK8+property. Using the property as here stops it from actually adding the JDK-8 specific bits on JDK11+ when run there (since the job script is shared across all 3) which would break.

.travis.yml
gemmellr: No, as that would enable the JDK8 arguments on 11+ and break them. The profile can be activated explicitly with -P or by JDK8+property. Using the property as here stops it from actually adding the JDK-8 specific bits on JDK11+ when run there (since the job script is shared across all 3) which would break.

clebertsuconic: @gemmellr thanks for the clarification.

PR #3188, opened 2020-06-16
ARTEMIS-2807 avoid notifications on critical IO error

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -1071,6 +1071,11 @@ void stop(boolean failoverOnServerShutdown, final boolean criticalIOError, boole
          }
          state = SERVER_STATE.STOPPING;
 
+         if (criticalIOError) {
+            // notifications trigger disk IO so we don't want to send any on a critical IO error
```
michaelandrepearce, line 1075: Why would i not want to know about critical errors on the management notification channel. This doesnt make sense to me

jbertram: As the comment and the Jira state, sending notification messages triggers disk IO. Depending on the nature of the underlying issue which caused the original critical IO error additional disk IO may cause the broker shutdown process to be delayed or hang indefinitely. This has implications for failover in that failover may not actually occur as expected. Giving up a few notifications to help ensure the broker shuts down in timely manner is a suitable compromise in my opinion.

clebertsuconic: I agree with this change.

Although I think we could follow up with an option to HALT on critical IO like we do on the critical analyzer. (not just shutdown).

PR #3195, opened 2020-06-20
ARTEMIS-2815 Fix null pointer exception on attempt to update queue without filter

Havret:
I'm not sure if there is more elegant way to do it in Java. Any hints would be greatly appreciated.


michaelandrepearce:
You can inline it by filter == null ? null : filter.getFilterString():

Havret:
Thanks Mike, fixed.

Havret:
BTW, what's about with this --> https://github.com/apache/activemq-artemis/blob/4e77a9fb79503af017695c0955438517450c0d50/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java#L3965-L3966

brusdev:
@Havret I guess the second one should be .setLastValueKey(null). @jbertram WDYT?

> BTW, what's about with this -->
> 
> https://github.com/apache/activemq-artemis/blob/4e77a9fb79503af017695c0955438517450c0d50/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java#L3965-L3966

michaelpearce-gain:
> BTW, what's about with this -->
> 
> https://github.com/apache/activemq-artemis/blob/4e77a9fb79503af017695c0955438517450c0d50/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java#L3965-L3966

This is because the QueueImpl doesnt support LVQ, theres a special extended version, as such its not possible atm to update (you have to remove and re add a queue currently) This is something though that could be rolled into a single queue, and make it updateable, i had a poc a while back, just never got time to finish it. And no one is shouting for it so like all things prioritised down.

clebertsuconic:
LGTM...

@michaelandrepearce I'm not merging this as I'm not sure if you settled your discussion here. if you could merge when done please.

michaelpearce-gain:
@clebertsuconic setLastValueKey needs to be fixed, I believe @Havret  is looking to push a fix on that tonight or tomorrow.

Havret:
@michaelandrepearce @clebertsuconic I've pushed the changes.

clebertsuconic:
@michaelandrepearce cool.. I will start the release tomorrow.  I will wait a final review from you tomorrow, or if you don't merge it by then I will do it.

michaelpearce-gain:
@Havret merged thanks

PR #3196, opened 2020-06-22
Stop the timeout future when channel is closed

emagiz:
Solves a small issue we are having with the AWS load balancer doing a health check. The logging explodes with warnings I think.
The cause is that there is no event handling the graceful closure of a connection. This fix should solve it.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/ProtocolHandler.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/ProtocolHandler.java
@@ -101,7 +101,7 @@ public class ProtocolHandler {
 
       private final boolean httpEnabled;
 
-      private ScheduledFuture timeoutFuture;
+      private ScheduledFuture<?> timeoutFuture;
 
       private int handshakeTimeout;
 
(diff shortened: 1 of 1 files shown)


clebertsuconic:
what's the reason behind this? What failures did you see for this?

is it possible to include a test?


Also: the commit-id needs to be ammended with a JIRA:

``ARTEMIS-XXX Stop the timeout future when channel is closed``

clebertsuconic:
@emagiz The full testsuite pass... I just wanted to confirm what failing scenario did you see with this as this requires a JIRA.

emagiz:
Sure, I'll create a JIRA because I don't have one right now. I can amend the commit and include the JIRA issue when I create it. Can I keep the JIRA issue short as I probably don't have the logging anymore. Basically the scheduled future is not canceled on gracefull close and we have a log statement per port check.

jbertram:
This is a nice little fix. It eliminates the false positives we get for clients (e.g. load balancer health-checkers ) that connect to an acceptor, don't complete a protocol handshake, but still close their connection. Previously we recommended a configuration change to work-around it.

PR #3198, opened 2020-06-22
Closing the directbytebuffer locally to prevent leaks

emagiz:
Related to the jira issue: 
https://issues.apache.org/jira/browse/ARTEMIS-2696
Assuming this can be done locally, this could be a fix for the above issue.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/ProtonHandler.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/ProtonHandler.java
@@ -276,12 +276,17 @@ public class ProtonHandler extends ProtonInitializable implements SaslListener {
                break;
             }
 
-            // We allocated a Pooled Direct Buffer, that will be sent down the stream
             ByteBuf buffer = PooledByteBufAllocator.DEFAULT.directBuffer(pending);
-            buffer.writeBytes(head);
+            try {
+               // We allocated a Pooled Direct Buffer, that will be sent down the stream
+               buffer.writeBytes(head);
 
-            for (EventHandler handler : handlers) {
-               handler.pushBytes(buffer);
+               for (EventHandler handler : handlers) {
+                  handler.pushBytes(buffer);
+               }
+            } finally {
+               // We need to release the buffer when it has been sent downstream
+               buffer.release();
             }
 
             transport.pop(pending);


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/ProtonHandler.java
```diff
@@ -276,12 +276,17 @@ private void actualFlush() {
...
+               buffer.writeBytes(head);
 
-            for (EventHandler handler : handlers) {
-               handler.pushBytes(buffer);
+               for (EventHandler handler : handlers) {
+                  handler.pushBytes(buffer);
+               }
+            } finally {
+               // We need to release the buffer when it has been sent downstream
+               buffer.release();
```
franz1981, line 289: I'm not sure about this: if this release would happen before the `ByteBuf` is being handled by some of the handler they will get a failure because the ByteBuf is released...

clebertsuconic: actually, the tests are failing pretty badly with this change. look at the PR tests.

emagiz: Indeed test results do not look good, is the ci/cd harder on the tests for the ci/cd? I did a local build on the combined PRs we did today and they succeeded. Is the ci/cd heavier then when you "mvn clean install" it?

clebertsuconic: @emagiz I'm not sure about the CI/CD.. if you're talking about redhat's CI.. they are harder then the PR checks here actually.

emagiz: Ok, I tried with the mvn command from the automated build with the additional testing profiles and it fails on my laptop as well now. But I might just need to do a restart because it complains on CLI testcases where it complains about a process already using some file.

But is the fix correct in the sense there is a release missing on this buffer which needs to be added somewhere (maybe not here, but somewhere?

We have been running this fix on an environment for a week and we have had no netty LEAK reports anymore (they started from version 2.11) and have noticed no issues related to data quality. But maybe we have a narrow case with very little coverage. If we remove this patch we get netty LEAK reports and the server crashed in around 6 hours because of no pooled direct buffer memory left to be used.

So if the assumption of the release call is correct, we can maybe change it. Something like: give all the handlers their own view and let them release it themselves. Would something like that be an option?

ldebello: Hi @emagiz I think the PR https://github.com/apache/activemq-artemis/pull/3252 fix the issue you are seeing. it shoud be release as part of 2.16.0

jbertram:
[ARTEMIS-2696](https://issues.apache.org/jira/browse/ARTEMIS-2696) was resolved awhile back. If this is still an issue please open a new Jira or update this PR. Thanks!

PR #3199, opened 2020-06-23
ARTEMIS-2819: Exact match address setting removal ineffective

k-wall:
Removal of address settings with an exact match address is currently ineffective.  PR adds a test that demonstrates the problem and a patch for the HierarchicalObjectRepository to clear the cache on the extact match path.


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -1309,6 +1309,265 @@ public void testAddressSettings() throws Exception {
...
+              addrMinExpiryDelay,
+              addrMaxExpiryDelay,
+              enableMetrics);
+      AddressSettingsInfo addrInfo = AddressSettingsInfo.from(serverControl.getAddressSettingsAsJSON(addr));
+
+      assertEquals("settings for addr should carry update", addrMinExpiryDelay, addrInfo.getMinExpiryDelay());
+      assertEquals("settings for addr should carry update", addrMaxExpiryDelay, addrInfo.getMaxExpiryDelay());
+
+// I noticed if I overwrite the settings with the root settings, then delete, a reread produces the expected value.
+//      serverControl.addAddressSettings(addr,
```
clebertsuconic, line 1504: do you need this commented out code? it seems the next line is all you need?

k-wall: that actually the workaround I've been using in EnMasse (add the root settings, then delete).  I've removed the code now.

clebertsuconic: @k-wall Ohh.. I see.. that commented out code had actually meaning.. I thought you had just forgotten in there. you can keep it if it has a meaning to be there.

k-wall: no, it was just my experiment, it has no place if this PR becomes the fix.

PR #3201, opened 2020-06-24
ARTEMIS-2820 Undeploy diverts by removing them from broker.xml

jsmucr:
This PR adds the ability to undeploy diverts by removing them from the `broker.xml` configuration file.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/RedeployTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/RedeployTest.java
@@ -43,6 +43,7 @@ import org.apache.activemq.artemis.api.core.RoutingType;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.postoffice.Binding;
 import org.apache.activemq.artemis.core.postoffice.QueueBinding;
+import org.apache.activemq.artemis.core.postoffice.impl.DivertBinding;
 import org.apache.activemq.artemis.core.postoffice.impl.LocalQueueBinding;
 import org.apache.activemq.artemis.core.security.Role;
 import org.apache.activemq.artemis.core.server.cluster.impl.MessageLoadBalancingType;
@@ -52,6 +53,7 @@ import org.apache.activemq.artemis.core.server.impl.AddressInfo;
 import org.apache.activemq.artemis.core.settings.impl.AddressFullMessagePolicy;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
 import org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory;
+import org.apache.activemq.artemis.jms.client.ActiveMQDestination;
 import org.apache.activemq.artemis.tests.unit.core.postoffice.impl.FakeQueue;
 import org.apache.activemq.artemis.tests.util.Wait;
 import org.apache.activemq.artemis.tests.util.ActiveMQTestBase;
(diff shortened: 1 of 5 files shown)


michaelpearce-gain:
@jsmucr thanks again for the contributions.

clebertsuconic:
@michaelpearce-gain  / @jsmucr there was a test broken after this PR, and I'm fixing it on #3209

however the test itself clash with the semantic you're presenting here. so I need you to review the change. and if the semantics are clashing we will need to address it.

thanks

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/RedeployTest.java
```diff
@@ -448,6 +450,63 @@ public void testRedeployQueueDefaults() throws Exception {
       }
    }
 
+   @Test
+   public void testUndeployDivert() throws Exception {
```
clebertsuconic, line 454: @michaelpearce-gain  / @jsmucr any chance you can look at why this test is failing?

if you run it as part of the test suite (mvn -Ptests test) it breaks most of the time.

And this is blocking me from releasing.


Another point is the clash of the user redeployment as it's clashing with a previous definition.. see my PR.

michaelandrepearce: @clebertsuconic re the user in other pr, i think the issue is the test that failed was enforcing you cannot clear a user, that test probably should have been changed. @jsmucr will have to send a pr tomorrow to revert your change and fix the test so that its legal to clear a user. That said if you need to release it can be sorted in separate ticket.

Re this one on diverts if @jsmucr doesnt get back to you, revert the PR

jsmucr: @clebertsuconic @michaelandrepearce Thank you for the review, I'll get on with that.

jsmucr: \*sad face\* 
My PR is wrong. What I originally wanted to implement was to reset the values when reading `broker.xml`. What I've done instead is that it resets the queue configuration every time an incomplete configuration object is passed to the `updateQueue` method. Which in fact breaks the `testUpdateQueueWithNullUser` test.

clebertsuconic: @jsmucr if you could check and fix RedeployTest::testUndeployDivert and testRedeployQueueDefaults

with my latest PR merged 
org.apache.activemq.artemis.tests.integration.client.UpdateQueueTest.testUpdateQueueWithNullUser is passing.


But as you and Michael said something else seems to be going on.


I won't release 2.14 for another week, so we have time to fix this.

clebertsuconic: I have no idea why testUndeployDivert fails... when I run it on my laptop it always pass. when I run it as part of mvn -Ptests test, it fails most of the time. (9 in 10). so I have not been successful on debugging it.

if you have no idea I will probably create a branch with lots of traces.

(perhaps the collector from your change fails with ConcurrentModificationException and redeployment does not work?) that's the only theory I have but I tried to prove it and I couldn't.

jsmucr: @clebertsuconic I'll keep this in mind and try to resolve it later.

Your PR resolves this by reverting part of the functionality that I've implemented. It does not fix the problem as a whole.

For now I think it would be for the best to revert #3172 so neither the build or the configuration mechanism is broken. I shall come with a different approach later.

PR #3204, opened 2020-06-29
ARTEMIS-2823 Use datasource with JDBC store db connections

uomik:
Replaces direct jdbc connections with dbcp2 datasource. Adda
configuration options to use alternative datasources and to alter the
parameters. While adding slight overhead, this vastly improves the
management and pooling capabilities with db connections.


sebthom:
Before on settling down on dbcp. Did you evaluate other connection pool implementations. I remember that dbcp had performance and memory issues in the past (maybe nowadays too?). HikariCP might be a better alternative.

https://github.com/brettwooldridge/HikariCP
https://programminghints.com/2017/07/hikaricp/

uomik:
> Before on settling down on dbcp. Did you evaluate other connection pool implementations. I remember that dbcp had performance and memory issues in the past (maybe nowadays too?). HikariCP might be a better alternative.

Yes I did evaluate HikariCP and c3p0 (https://www.mchange.com/projects/c3p0/) along with dbcp2. Each of them worked just fine and I saw no noticeable performance difference beween them. Yes there used to be some major performance issues with dbcp2, but to my knowledge the latest version is much better. I did not run any real performance tests with each of the connection pools, but have done it with other applications. Dbcp2 has been slightly slower, but proven to be more reliable in an environment where database drops connections very aggressively, when compared to HikariCP. c3p0 was simply dropped because I'm not that familiar with it.

At the end, I made the choice based on the fact that with default settings dbcp2 was the only one that I got to work flawlessly in our test cluster. Both HikariCP and c3p0 required some extra configuration.

artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/file/Db2SequentialFileDriver.java
```diff
@@ -30,33 +31,30 @@ public Db2SequentialFileDriver() {
...
+      this.selectFileNamesByExtension = sqlProvider.getSelectFileNamesByExtensionSQL();
    }
 
    @Override
    public int writeToFile(JDBCSequentialFile file, byte[] data) throws SQLException {
       if (data == null || data.length == 0) {
          return 0;
       }
-      synchronized (connection) {
-         try {
```
gtully, line 59: One thought on this change, to have the same semantic, the connectionPool *must* have a max pool size of 1. That needs to be enforced.
With the original code, the entire jdbc operation is serialised on the connection, and that will only still hold if the connection pool blocks when it single entry is in use. ie: if it has a max pool size of 1.

franz1981 (changes requested):

artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/file/Db2SequentialFileDriver.java
uomik: I've done this through the broker.xml config and completely forgot to enforce the max size in code if not configured. Will try to find a proper solution for this

franz1981:
@uomik 
I'm getting some test failures eg `JDBCJournalTest::testCleanupTxRecords`:
```
[main] 22:01:31,134 INFO  [org.apache.activemq.artemis.tests.util.ActiveMQTestBase] **** start #test testCleanupTxRecords[authentication = false]() ***
[main] 22:01:31,136 INFO  [org.apache.activemq.artemis.jdbc.store.drivers.JDBCDataSourceUtils] Initialising JDBC data source: org.apache.commons.dbcp2.BasicDataSource {driverClassName=org.apache.derby.jdbc.EmbeddedDriver, url=jdbc:derby:/home/forked_franz/IdeaProjects/activemq-artemis-upstream/tests/integration-tests/./target/tmp/JDBCJournalTest/junit12974489528425881245/derby;create=true}
[main] 22:01:31,878 INFO  [org.apache.activemq.artemis.tests.util.ActiveMQTestBase] **** end #test testCleanupTxRecords[authentication = false]() ***

java.lang.AssertionError: 
Expected :0
Actual   :1
<Click to see difference>


	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at org.apache.activemq.artemis.tests.integration.jdbc.store.journal.JDBCJournalTest.testCleanupTxRecords(JDBCJournalTest.java:175)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:47)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:44)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:48)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:48)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:48)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
	at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:271)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:70)
	at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:50)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:53)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:229)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:309)
	at org.junit.runners.Suite.runChild(Suite.java:127)
	at org.junit.runners.Suite.runChild(Suite.java:26)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:238)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:63)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:236)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:53)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:229)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:55)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:309)
	at org.junit.runner.JUnitCore.run(JUnitCore.java:160)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:220)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:53)
```

franz1981:
@uomik I'm sending a PR to your branch to fix the error

franz1981:
Same for the tests using JDBC authentication, but probably is just a derby issue

artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/file/Db2SequentialFileDriver.java
franz1981: @gtully 
I'm searching trough the `master` code base and although it seems that it's using a single database connection I'm 100% sure of it: I will verify it tomorrow with a running broker tomorrow.
Probably we've many JDBC connections opened (see `AbstractJDBCDriver` childrens -> journal(s), node manager and sequential file fatories)
Probably there is no need to hard code such limit

franz1981:
Please rebase over master and try the latest commits on https://github.com/franz1981/activemq-artemis/tree/2.13.0-jdbc-ds
Seems that I cannot send a PR to your branch

artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/file/Db2SequentialFileDriver.java
uomik: I did a quick check by just starting "stock" Artemis and it at least opens several connections to the database on startup. After a while it seems to settle down to two active connections. And I have to correct myself, I forgot that I actually decided to drop the setting form the configs as well as I came to conclusion that removing it has no negative effect. But correct me if I'm wrong.

franz1981:
@uomik any news on the changes I've proposed on https://github.com/apache/activemq-artemis/pull/3204#issuecomment-690766208?

uomik:
> @uomik any news on the changes I've proposed on [#3204 (comment)](https://github.com/apache/activemq-artemis/pull/3204#issuecomment-690766208)?

Sorry. I'm burden under work currently. Had a quick look on changes (that seem to ok and be on point) and did the rebase, but was little confused about the outcome (I'm no means a git expert), and have had absolutely no time to dig into it. The unit tests did go through though.

franz1981:
@uomik I will try to send a PR so you can accept/merge it and push force the new version, do you need me to help more?

franz1981:
@uomik 
That's the PR: https://github.com/uomik/activemq-artemis/pull/1

franz1981:
@uomik Just a note: I see that you have modified both deps and features hence the commons-dbcp2 artifact is not deployed with the broker libs: this is done on purpose? How you do in production, if you're using this?

uomik:
> @uomik Just a note: I see that you have modified both deps and features hence the commons-dbcp2 artifact is not deployed with the broker libs: this is done on purpose? How you do in production, if you're using this?

This was not intended, if this is the case it's definitely a mistake, or something I have overlooked.

We are not yet on production with this feature. In test environment we patch the official Artemis release with modified jars and add commons-dbcp2, commons-logging and commons-pool2 jars to lib (along with jdbc driver). This is done in CI/CD pipeline. We chose this approach simply because it was easier in our environment when developing this datasource feature and is in no means the way we would like to do things in production.

franz1981:
thanks @uomik 

> In test environment we patch the official Artemis release with modified jars and add commons-dbcp2, commons-logging and commons-pool2 jars to lib (along with jdbc driver)

Probably is a better way to handle this for users too: forcing a specific data source isn't good although we assume dbcp2 to be used (looking at the unwrapping logic for PosgresSQL) or whatever pooled datasource. 
Users with existing JDBC connection configuration need to change their config (and libs) if they want "decent" performance

franz1981:
I've managed to run some tests with MySql and found something odd: if you add this configuration to `broker.xml`:
```xml
	<ha-policy>
	    <shared-store>
	        <master/>
	    </shared-store>
	</ha-policy>
```
And stop the database, the broker will stop after:
```
2020-09-15 19:59:05,080 WARN  [org.apache.activemq.artemis.core.server] AMQ222010: Critical IO Error, shutting down the server. file=NULL, message=Critical error while on live renew: java.lang.IllegalStateException: com.mysql.cj.jdbc.exceptions.CommunicationsException: Communications link failure

The last packet successfully received from the server was 2 milliseconds ago. The last packet sent successfully to the server was 2 milliseconds ago.
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcLeaseLock.renew(JdbcLeaseLock.java:169) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.ActiveMQScheduledLeaseLock.run(ActiveMQScheduledLeaseLock.java:91) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.ActiveMQScheduledComponent$2.run(ActiveMQScheduledComponent.java:306) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.OrderedExecutor.doTask(OrderedExecutor.java:42) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.OrderedExecutor.doTask(OrderedExecutor.java:31) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.executePendingTasks(ProcessorBase.java:65) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [rt.jar:1.8.0_265]
        at org.apache.activemq.artemis.utils.ActiveMQThreadFactory$1.run(ActiveMQThreadFactory.java:118) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
Caused by: com.mysql.cj.jdbc.exceptions.CommunicationsException: Communications link failure
```
I suppose that the data source pool need to be configured somehow or the lease lock renew will make it stop. 
That's the expected behaviour?

uomik:
> I suppose that the data source pool need to be configured somehow or the lease lock renew will make it stop.
> That's the expected behaviour?

The drawback with shared store, being database or something else, is that it essentially becomes a single point of failure for the whole cluster. If we lose the database, we lose the ability to sync data between master and backup, the ability to ensure data integrity and broker high availability. In this case, on my opinion, the only thing we can do is to shut down the broker node as precautionary measure. This is the way it was before I fiddled with anything, and I saw no reasons to change it.

The way you can mitigate this single point of failure is to have multiple master-backup pairs distributed accross multiple databases (servers). Or build some redundancy in the infrastrusture (clustering the database etc.).

franz1981:
@uomik 

> The drawback with shared store, being database or something else, is that it essentially becomes a single point of failure for the whole cluster. If we lose the database, we lose the ability to sync data between master and backup, the ability to ensure data integrity and broker high availability. In this case, on my opinion, the only thing we can do is to shut down the broker node as precautionary measure. This is the way it was before I fiddled with anything, and I saw no reasons to change it.

Agree, but consider the simple case of a database upgrade ( ie a database restart): a shared store implies a live and backup connected to it; the expectation during a database upgrade would be that the live will go down (that's what is happening now) while the slave should keep trying to became live (ie while now it just die).

uomik:
@franz1981 

> Agree, but consider the simple case of a database upgrade ( ie a database restart): a shared store implies a live and backup connected to it; the expectation during a database upgrade would be that the live will go down (that's what is happening now) while the slave should keep trying to became live (ie while now it just die).
> Probably the backup behaviour of shared store ha should be different. It seems a separate problem to me then this PR, just thinking loud.

My time to completely agree :) I thought you were just referring to the behaviour of the master. Backup should definitely behave differently on these occasions.

franz1981:
I've found another issue with shared store HA on live stop:
```
2020-09-16 09:51:36,116 WARN  [org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider] Unable to set a network timeout on the JDBC connection: java.util.concurrent.RejectedExecutionException: Task org.apache.activemq.artemis.utils.actors.ProcessorBase$$Lambda$61/1932332324@40491594 rejected from org.apache.activemq.artemis.utils.ActiveMQThreadPoolExecutor@6cfdff08[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 92]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379) [rt.jar:1.8.0_265]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.onAddedTaskIfNotRunning(ProcessorBase.java:186) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.task(ProcessorBase.java:174) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.OrderedExecutor.execute(OrderedExecutor.java:54) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at com.mysql.cj.jdbc.ConnectionImpl.setNetworkTimeout(ConnectionImpl.java:2488) [mysql-connector-java-8.0.21.jar:8.0.21]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider.getConnection(JDBCConnectionProvider.java:62) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcLeaseLock.release(JdbcLeaseLock.java:267) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.pauseLiveServer(JdbcNodeManager.java:413) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.SharedStoreLiveActivation.close(SharedStoreLiveActivation.java:161) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl.stop(ActiveMQServerImpl.java:1247) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl.stop(ActiveMQServerImpl.java:1033) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl.stop(ActiveMQServerImpl.java:876) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.integration.FileBroker.stop(FileBroker.java:98) [artemis-cli-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.cli.commands.Run.stop(Run.java:166) [artemis-cli-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.cli.commands.Run$2.run(Run.java:157) [artemis-cli-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
```
`activation.close(failoverOnServerShutdown, restarting);` on server stop is causing `JdbcLeaseLock.release` to set a network timeout on the JDBC connection while using an already closed executor (previously on stop). 
Need to check how to fix this

uomik:
> `activation.close(failoverOnServerShutdown, restarting);` on server stop is causing `JdbcLeaseLock.release` to set a network timeout on the JDBC connection while using an already closed executor (previously on stop).
> Need to check how to fix this

I've come across this as well. After investigating I came to conclusion that this might be an issue with MySQL or Connector/J. But have not yet had a proper look into this, so might be something else.

franz1981:
Nope it seems related to how we configure the network connection timeout (still need to understand if it makes sense...).

I see a similar problem on backup side while live is restarting, see:
```
2020-09-16 15:07:09,990 INFO  [org.apache.activemq.artemis.core.server] AMQ221008: live server wants to restart, restarting server in backup
2020-09-16 15:07:10,181 INFO  [org.apache.activemq.artemis.core.server] AMQ221002: Apache ActiveMQ Artemis Message Broker version 2.14.0-SNAPSHOT [b25dd37c-f77c-11ea-9566-1aabf22f14df] stopped, uptime 1 minute
2020-09-16 15:07:10,183 WARN  [org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider] Unable to set a network timeout on the JDBC connection: java.util.concurrent.RejectedExecutionException: Task org.apache.activemq.artemis.utils.actors.ProcessorBase$$Lambda$61/1394940518@15f4294c rejected from org.apache.activemq.artemis.utils.ActiveMQThreadPoolExecutor@69d38d0a[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 265]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379) [rt.jar:1.8.0_265]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.onAddedTaskIfNotRunning(ProcessorBase.java:186) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.task(ProcessorBase.java:174) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.OrderedExecutor.execute(OrderedExecutor.java:54) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at com.mysql.cj.jdbc.ConnectionImpl.setNetworkTimeout(ConnectionImpl.java:2488) [mysql-connector-java-8.0.21.jar:8.0.21]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider.getConnection(JDBCConnectionProvider.java:62) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.jdbc.store.drivers.AbstractJDBCDriver.createTableIfNotExists(AbstractJDBCDriver.java:90) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.jdbc.store.drivers.AbstractJDBCDriver.createTable(AbstractJDBCDriver.java:63) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcSharedStateManager.createSchema(JdbcSharedStateManager.java:68) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.jdbc.store.drivers.AbstractJDBCDriver.start(AbstractJDBCDriver.java:50) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcSharedStateManager.usingConnectionProvider(JdbcSharedStateManager.java:58) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.lambda$usingConnectionProvider$0(JdbcNodeManager.java:91) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.start(JdbcNodeManager.java:162) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.SharedStoreBackupActivation$FailbackChecker$1.run(SharedStoreBackupActivation.java:217) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at java.lang.Thread.run(Thread.java:748) [rt.jar:1.8.0_265]

2020-09-16 15:07:10,191 WARN  [org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider] Unable to set a network timeout on the JDBC connection: java.util.concurrent.RejectedExecutionException: Task org.apache.activemq.artemis.utils.actors.ProcessorBase$$Lambda$61/1394940518@15f4294c rejected from org.apache.activemq.artemis.utils.ActiveMQThreadPoolExecutor@69d38d0a[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 265]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379) [rt.jar:1.8.0_265]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.onAddedTaskIfNotRunning(ProcessorBase.java:186) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.task(ProcessorBase.java:174) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.OrderedExecutor.execute(OrderedExecutor.java:54) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at com.mysql.cj.jdbc.ConnectionImpl.setNetworkTimeout(ConnectionImpl.java:2488) [mysql-connector-java-8.0.21.jar:8.0.21]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider.getConnection(JDBCConnectionProvider.java:62) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcSharedStateManager.setup(JdbcSharedStateManager.java:193) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.start(JdbcNodeManager.java:164) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.SharedStoreBackupActivation$FailbackChecker$1.run(SharedStoreBackupActivation.java:217) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at java.lang.Thread.run(Thread.java:748) [rt.jar:1.8.0_265]

2020-09-16 15:07:10,195 WARN  [org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider] Unable to set a network timeout on the JDBC connection: java.util.concurrent.RejectedExecutionException: Task org.apache.activemq.artemis.utils.actors.ProcessorBase$$Lambda$61/1394940518@15f4294c rejected from org.apache.activemq.artemis.utils.ActiveMQThreadPoolExecutor@69d38d0a[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 265]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379) [rt.jar:1.8.0_265]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.onAddedTaskIfNotRunning(ProcessorBase.java:186) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.task(ProcessorBase.java:174) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.OrderedExecutor.execute(OrderedExecutor.java:54) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at com.mysql.cj.jdbc.ConnectionImpl.setNetworkTimeout(ConnectionImpl.java:2488) [mysql-connector-java-8.0.21.jar:8.0.21]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider.getConnection(JDBCConnectionProvider.java:62) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcSharedStateManager.readState(JdbcSharedStateManager.java:285) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.readSharedState(JdbcNodeManager.java:477) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.awaitLiveStatus(JdbcNodeManager.java:451) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.SharedStoreBackupActivation$FailbackChecker$1.run(SharedStoreBackupActivation.java:219) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at java.lang.Thread.run(Thread.java:748) [rt.jar:1.8.0_265]

2020-09-16 15:07:12,198 WARN  [org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider] Unable to set a network timeout on the JDBC connection: java.util.concurrent.RejectedExecutionException: Task org.apache.activemq.artemis.utils.actors.ProcessorBase$$Lambda$61/1394940518@15f4294c rejected from org.apache.activemq.artemis.utils.ActiveMQThreadPoolExecutor@69d38d0a[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 265]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379) [rt.jar:1.8.0_265]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.onAddedTaskIfNotRunning(ProcessorBase.java:186) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.task(ProcessorBase.java:174) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.OrderedExecutor.execute(OrderedExecutor.java:54) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at com.mysql.cj.jdbc.ConnectionImpl.setNetworkTimeout(ConnectionImpl.java:2488) [mysql-connector-java-8.0.21.jar:8.0.21]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider.getConnection(JDBCConnectionProvider.java:62) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcSharedStateManager.readState(JdbcSharedStateManager.java:285) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.readSharedState(JdbcNodeManager.java:477) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.awaitLiveStatus(JdbcNodeManager.java:451) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.SharedStoreBackupActivation$FailbackChecker$1.run(SharedStoreBackupActivation.java:219) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at java.lang.Thread.run(Thread.java:748) [rt.jar:1.8.0_265]

2020-09-16 15:07:12,201 WARN  [org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider] Unable to set a network timeout on the JDBC connection: java.util.concurrent.RejectedExecutionException: Task org.apache.activemq.artemis.utils.actors.ProcessorBase$$Lambda$61/1394940518@15f4294c rejected from org.apache.activemq.artemis.utils.ActiveMQThreadPoolExecutor@69d38d0a[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 265]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379) [rt.jar:1.8.0_265]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.onAddedTaskIfNotRunning(ProcessorBase.java:186) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.task(ProcessorBase.java:174) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.OrderedExecutor.execute(OrderedExecutor.java:54) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at com.mysql.cj.jdbc.ConnectionImpl.setNetworkTimeout(ConnectionImpl.java:2488) [mysql-connector-java-8.0.21.jar:8.0.21]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider.getConnection(JDBCConnectionProvider.java:62) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.jdbc.store.sql.PropertySQLProvider$Factory.investigateDialect(PropertySQLProvider.java:405) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.jdbc.store.sql.PropertySQLProvider$Factory.<init>(PropertySQLProvider.java:366) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.with(JdbcNodeManager.java:67) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl.createNodeManager(ActiveMQServerImpl.java:501) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl.internalStart(ActiveMQServerImpl.java:578) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl.start(ActiveMQServerImpl.java:526) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.SharedStoreBackupActivation$FailbackChecker$1.run(SharedStoreBackupActivation.java:230) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at java.lang.Thread.run(Thread.java:748) [rt.jar:1.8.0_265]

2020-09-16 15:07:12,202 WARN  [org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider] Unable to set a network timeout on the JDBC connection: java.util.concurrent.RejectedExecutionException: Task org.apache.activemq.artemis.utils.actors.ProcessorBase$$Lambda$61/1394940518@15f4294c rejected from org.apache.activemq.artemis.utils.ActiveMQThreadPoolExecutor@69d38d0a[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 265]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379) [rt.jar:1.8.0_265]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.onAddedTaskIfNotRunning(ProcessorBase.java:186) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.task(ProcessorBase.java:174) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.OrderedExecutor.execute(OrderedExecutor.java:54) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at com.mysql.cj.jdbc.ConnectionImpl.setNetworkTimeout(ConnectionImpl.java:2488) [mysql-connector-java-8.0.21.jar:8.0.21]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider.getConnection(JDBCConnectionProvider.java:62) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.jdbc.store.drivers.AbstractJDBCDriver.createTableIfNotExists(AbstractJDBCDriver.java:90) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.jdbc.store.drivers.AbstractJDBCDriver.createTable(AbstractJDBCDriver.java:63) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcSharedStateManager.createSchema(JdbcSharedStateManager.java:68) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.jdbc.store.drivers.AbstractJDBCDriver.start(AbstractJDBCDriver.java:50) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcSharedStateManager.usingConnectionProvider(JdbcSharedStateManager.java:58) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.lambda$usingConnectionProvider$0(JdbcNodeManager.java:91) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.start(JdbcNodeManager.java:162) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl.internalStart(ActiveMQServerImpl.java:580) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl.start(ActiveMQServerImpl.java:526) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.SharedStoreBackupActivation$FailbackChecker$1.run(SharedStoreBackupActivation.java:230) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at java.lang.Thread.run(Thread.java:748) [rt.jar:1.8.0_265]

2020-09-16 15:07:12,205 WARN  [org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider] Unable to set a network timeout on the JDBC connection: java.util.concurrent.RejectedExecutionException: Task org.apache.activemq.artemis.utils.actors.ProcessorBase$$Lambda$61/1394940518@15f4294c rejected from org.apache.activemq.artemis.utils.ActiveMQThreadPoolExecutor@69d38d0a[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 265]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379) [rt.jar:1.8.0_265]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.onAddedTaskIfNotRunning(ProcessorBase.java:186) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.task(ProcessorBase.java:174) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.OrderedExecutor.execute(OrderedExecutor.java:54) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at com.mysql.cj.jdbc.ConnectionImpl.setNetworkTimeout(ConnectionImpl.java:2488) [mysql-connector-java-8.0.21.jar:8.0.21]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider.getConnection(JDBCConnectionProvider.java:62) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcSharedStateManager.setup(JdbcSharedStateManager.java:193) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.start(JdbcNodeManager.java:164) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl.internalStart(ActiveMQServerImpl.java:580) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl.start(ActiveMQServerImpl.java:526) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.SharedStoreBackupActivation$FailbackChecker$1.run(SharedStoreBackupActivation.java:230) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at java.lang.Thread.run(Thread.java:748) [rt.jar:1.8.0_265]

2020-09-16 15:07:12,207 INFO  [org.apache.activemq.artemis.core.server] AMQ221000: backup Message Broker is starting with configuration Broker Configuration (clustered=true,jdbcDriverClassName=com.mysql.cj.jdbc.Driver,jdbcConnectionUrl=jdbc:mysql://localhost:3306/mysql,messageTableName=MESSAGE_TABLE,bindingsTableName=BINDINGS_TABLE,largeMessageTableName=LARGE_MESSAGES_TABLE,pageStoreTableName=PAGE_TABLE,)
2020-09-16 15:07:14,586 INFO  [org.apache.activemq.artemis.core.server] AMQ221032: Waiting to become backup node
2020-09-16 15:07:14,587 WARN  [org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider] Unable to set a network timeout on the JDBC connection: java.util.concurrent.RejectedExecutionException: Task org.apache.activemq.artemis.utils.actors.ProcessorBase$$Lambda$61/1394940518@15f4294c rejected from org.apache.activemq.artemis.utils.ActiveMQThreadPoolExecutor@69d38d0a[Terminated, pool size = 0, active threads = 0, queued tasks = 0, completed tasks = 265]
        at java.util.concurrent.ThreadPoolExecutor$AbortPolicy.rejectedExecution(ThreadPoolExecutor.java:2063) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.reject(ThreadPoolExecutor.java:830) [rt.jar:1.8.0_265]
        at java.util.concurrent.ThreadPoolExecutor.execute(ThreadPoolExecutor.java:1379) [rt.jar:1.8.0_265]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.onAddedTaskIfNotRunning(ProcessorBase.java:186) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.ProcessorBase.task(ProcessorBase.java:174) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.utils.actors.OrderedExecutor.execute(OrderedExecutor.java:54) [artemis-commons-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at com.mysql.cj.jdbc.ConnectionImpl.setNetworkTimeout(ConnectionImpl.java:2488) [mysql-connector-java-8.0.21.jar:8.0.21]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.Jdbc41Bridge.setNetworkTimeout(Jdbc41Bridge.java:386) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.commons.dbcp2.DelegatingConnection.setNetworkTimeout(DelegatingConnection.java:985) [commons-dbcp2-2.7.0.jar:2.7.0]
        at org.apache.activemq.artemis.jdbc.store.drivers.JDBCConnectionProvider.getConnection(JDBCConnectionProvider.java:62) [artemis-jdbc-store-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcLeaseLock.tryAcquire(JdbcLeaseLock.java:175) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.LeaseLock.tryAcquire(LeaseLock.java:91) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.LeaseLock.tryAcquire(LeaseLock.java:109) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.lock(JdbcNodeManager.java:267) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.jdbc.JdbcNodeManager.startBackup(JdbcNodeManager.java:360) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.SharedStoreBackupActivation.run(SharedStoreBackupActivation.java:62) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
        at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl$ActivationThread.run(ActiveMQServerImpl.java:3981) [artemis-server-2.14.0-SNAPSHOT.jar:2.14.0-SNAPSHOT]
```

franz1981:
The problem on  seems related to: https://github.com/apache/activemq-artemis/blob/d55ec37195dff04377a43957cc0afa9d0de14b89/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/SharedStoreBackupActivation.java#L217

where the `nodeManager` is being used while the server (and the thread pools created on it) are stopped.

franz1981:
@uomik 

One question: the comment here is valid?
```java
/**
    * The DataSource to use to store Artemis data in the data store (can be {@code null} if {@code jdbcConnectionUrl} and {@code jdbcDriverClassName} are used instead).
    *
    * @return the DataSource used to store Artemis data in the JDBC data store.
    */
   private DataSource getDataSource() {
      if (dataSource == null) {
         if (dataSourceProperties.isEmpty()) {
            addDataSourceProperty("driverClassName", jdbcDriverClassName);
            addDataSourceProperty("url", jdbcConnectionUrl);
            if (jdbcUser != null) {
               addDataSourceProperty("username", jdbcUser);
            }
            if (jdbcPassword != null) {
               addDataSourceProperty("password", jdbcPassword);
            }
         }
         dataSource = JDBCDataSourceUtils.getDataSource(dataSourceClassName, dataSourceProperties);
      }
      return dataSource;
   }
```

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/storage/DatabaseStorageConfiguration.java
```diff
@@ -151,6 +175,33 @@ public void setDataSource(DataSource dataSource) {
       this.dataSource = dataSource;
    }
 
+   public JDBCConnectionProvider getConnectionProvider() {
+      if (connectionProvider == null) {
+         connectionProvider = new JDBCConnectionProvider(getDataSource());
+      }
+      return connectionProvider;
+   }
+   public void addDataSourceProperty(String key, String value) {
```
franz1981, line 184: why not directly calling `dataSourceProperties.put(key, value)`?
It will be up to the actual driver to validate the passed parameters, wdyt?

franz1981:
With these last changes I believe it's nearly ok, I'm now working to check how it behaves with different databases.
I will further work to make this better for the shared store use case too, similarly to what we do for NFS with file shared store (maybe).

I've just one concern looking at the default pool configuration: the number of available JDBC connections in the pool and the behaviour when there are none available.

I see that paging + large messsages can use an unbounded number of connections (it really depends by the concurrent thread that perform page/large msg streaming operations), the 2 journals requires 1 connections each (binding + message journal) and the node manager requires 2 (the lease lock + the node manager shared state) connections.
That means that without paging or large messages a minimum of 2 + 2 = 4 database connections.
The problem I see is paging and large messages: if the are too many some of the other operations will hang, awaiting an available connection, that means that probably we need to add the critical analyzer there or just let users knows they can configure pool appropriately.

franz1981:
Please use the rebase to remove the last merge commit and try rebasing against `master` manually instead so the merge commit won't appear when the PR will be merged

uomik:
> Please use the rebase to remove the last merge commit and try rebasing against `master` manually instead so the merge commit won't appear when the PR will be merged

I'll try, but it will take me a while to understand how I do that.

franz1981:
@uomik don't worry try with

```git reset --hard 76caeaccbc61f154cb9e26ed74e172534a1f4da0```

**Comment by uomik** [comment]
@franz1981 I may have succeeded :) Hopefully I didn't mess anything up

**Comment by franz1981** [comment]
Sorry @uomik we've had several fixes around JDBC, please try to fix the conflicts if you can so I would merge the PR: I'm still testing this with different DBMS :)

**Comment by franz1981** [comment]
@uomik that's https://gist.github.com/franz1981/ffb4cff1c1fc025f9d0fed1aa961f116 that fixed version of the conflicted file

**Comment by uomik** [comment]
@franz1981 
> I've just one concern looking at the default pool configuration: the number of available JDBC connections in the pool and the behaviour when there are none available.
> 
> I see that paging + large messsages can use an unbounded number of connections (it really depends by the amount of concurrent threads that perform page/large msg streaming operations), the 2 journals requires 1 connections each (binding + message journal) and the node manager requires 2 (the lease lock + the node manager shared state) connections.
> That means that without paging or large messages a minimum of 2 + 2 = 4 database connections.
> The problem I see is paging and large messages: if the are too many some of the other operations will hang, awaiting an available connection, that means that probably we need to add the critical analyzer there or just let users knows they can configure pool appropriately.

Getting back to this. We've lately opened our test cluster for "external" clients, this has increased load and amount of transactions quite a lot. As a result we've quite frequently seen an event where master restarts due to issues with live locks. We are still using default settings for the pool and I'm suspecting that this may actually be a result of DB connection starvation. I will experiment with pool settings next week and will hopefully find something to overcome this issue.

`2020-09-25 11:53:10,958 WARN  [org.apache.activemq.artemis.core.server] AMQ222010: Critical IO Error, shutting down the server. file=NULL, message=Critical error while on live renew: java.lang.IllegalStateException: live lock can't be renewed
	at org.apache.activemq.artemis.core.server.impl.jdbc.ActiveMQScheduledLeaseLock.run(ActiveMQScheduledLeaseLock.java:92) [artemis-server-2.13.0.jar:2.13.0]
	at org.apache.activemq.artemis.core.server.ActiveMQScheduledComponent$2.run(ActiveMQScheduledComponent.java:306) [artemis-commons-2.13.0.jar:2.13.0]
	at org.apache.activemq.artemis.utils.actors.OrderedExecutor.doTask(OrderedExecutor.java:42) [artemis-commons-2.13.0.jar:2.13.0]
	at org.apache.activemq.artemis.utils.actors.OrderedExecutor.doTask(OrderedExecutor.java:31) [artemis-commons-2.13.0.jar:2.13.0]
	at org.apache.activemq.artemis.utils.actors.ProcessorBase.executePendingTasks(ProcessorBase.java:65) [artemis-commons-2.13.0.jar:2.13.0]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [rt.jar:1.8.0_221]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [rt.jar:1.8.0_221]
	at org.apache.activemq.artemis.utils.ActiveMQThreadFactory$1.run(ActiveMQThreadFactory.java:118) [artemis-commons-2.13.0.jar:2.13.0]

2020-09-25 11:53:10,981 ERROR [org.apache.activemq.artemis.core.server.impl.jdbc.ActiveMQScheduledLeaseLock] live lock renew tooks 39255 ms, while is supposed to take <4000 ms`

**Comment by franz1981** [comment]
@uomik 
Yep, that was expected, sadly: I see 2 possible solutions to this.

1. set the max number of connections of the pool to be infinite (and document it)
2. use different data sources (and pools): one for the parts of the broker that requires a well defined amount (ie journals, node manager & lease locks), while another one for paging/large messages (both use sequential file factory IIRC), documenting the requirements for both

**Comment by uomik** [comment]
One thing thats a little strange is that the pool had only three active db connections at the time the error occurred (DBCP2 default max is 8, that should be enough in most cases). 

This may actually be the same issue we've had with other applications with this particular database server: the server randomly closes some of the connections after one hour, being active or not. Based on earlier tests I suspected that JDBC4 driver with isValid checks would be enough to cope with this. I will first check whether the same pool configuration we use elsewhere resolves the issue. If the problem persists will experiment with the other options

**Comment by franz1981** [comment]
> Based on earlier tests I suspected that JDBC4 driver with isValid checks would be enough to cope with this

And it is, the point is that only lease lock and node manager are alive with an idle broker ie 2 connections
I don't know what can be the third one...

**Comment by uomik** [comment]
> I don't know what can be the third one...

This might be that invalid connection. At least this happened with the other application I once fought with this db server (gradually filling the pool with invalid connections). That was using jdbc3 driver, though.

**Comment by franz1981** [comment]
And that makes sense indeed, it won't get changed unless none will request a new one...I know dbcp can do it on a separate thread too that save latency outliers on start up but I prefer this default behaviour.

I am more concern about the starvation of lease lock connection due to others; I am sure this need some attention before merging this pr...

File: artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/sql/PropertySQLProvider.java [hunk 3]
```diff
@@ -362,8 +363,12 @@ public Factory(SQLDialect dialect) {
          }
       }
 
-      public Factory(DataSource dataSource) {
```
ehsavoie, line 365: Can't we keep this method and call this(new JDBCConnectionProvider(ds));

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreFactoryDatabase.java
```diff
@@ -106,8 +105,8 @@ public PagingStoreFactoryDatabase(final DatabaseStorageConfiguration dbConf,
                                      final ScheduledExecutorService scheduledExecutor,
```
franz1981, line 105: I see that `JDBCSequentialFile directoryList` and `JDBCSequentialFileFactoryDriver dbDriver`

franz1981 (changes requested):

franz1981:
@uomik any news on this? 
We can just set an unbounded number of connections on the pool to begin with in order to proceed on this :+1:

artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/sql/PropertySQLProvider.java
ehsavoie: @uomik this is still missing in your latest changes

uomik: @ehsavoie Fixed now. For some reason the commit did not make it to my last push.

uomik:
@franz1981 

Getting back to the comment https://github.com/apache/activemq-artemis/pull/3204#issuecomment-698902202

I've now run our Artemis cluster with updated datasource / connection pool configuration (forced idle connection eviction after 30 minutes) for a couple of days. The random reboots have disappeared, so I'm concluding this really was an issue where our database server is dropping connections. The same fix (eviction runs) has been in use with other applications using the same database with similar results, so I'm quite confident with my conclusion.

Additionally did a commit for fixing the datasource pool size as unbounded as default.

franz1981:
@uomik happy to hear it!
I'll run this again on our CI as well and I would move to the next task based on this: https://issues.apache.org/jira/browse/ARTEMIS-2918

I'll call you out to test it if you like, because the scenario is a bit tricky, but I think it could help many users :)

BTW, great contribution, many thanks!

artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/file/PostgresLargeObjectManager.java
```diff
@@ -56,31 +56,31 @@ public PostgresLargeObjectManager(Connection connection) throws SQLException {
...
-   public final Long createLO() throws SQLException {
+   public final Long createLO(Connection connection) throws SQLException {
       if (shouldUseReflection) {
-         Object largeObjectManager = getLargeObjectManager();
+         Object largeObjectManager = getLargeObjectManager(connection);
          try {
-            Method method = largeObjectManager.getClass().getMethod("createLO");
-            return (Long) method.invoke(largeObjectManager);
+            Method method = largeObjectManager.getClass().getMethod("createLO", Connection.class);
+            return (Long) method.invoke(largeObjectManager, connection);
```
ehsavoie, line 64: connection is not a parameter of this method

artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/file/PostgresLargeObjectManager.java
```diff
@@ -56,31 +56,31 @@ public PostgresLargeObjectManager(Connection connection) throws SQLException {
...
-   public Object open(long oid, int mode) throws SQLException {
+   public Object open(Connection connection, long oid, int mode) throws SQLException {
       if (shouldUseReflection) {
-         Object largeObjectManager = getLargeObjectManager();
+         Object largeObjectManager = getLargeObjectManager(connection);
          try {
-            Method method = largeObjectManager.getClass().getMethod("open", long.class, int.class);
-            return method.invoke(largeObjectManager, oid, mode);
+            Method method = largeObjectManager.getClass().getMethod("open", Connection.class, long.class, int.class);
+            return method.invoke(largeObjectManager, connection, oid, mode);
```
ehsavoie, line 78: connection is not a parameter of this method

ehsavoie (changes requested):
Those 2 changes are required for postgresql to work properly. Continuing testing tomorrow

ehsavoie (changes requested):

franz1981 (approved):
LGTM I'm going to merge this one: many thanks for this great contribution!!

franz1981:
Please @uomik try to squash all the commits into one if possible and I'm going to merge it :+1:

uomik:
> Please @uomik try to squash all the commits into one if possible and I'm going to merge it 👍

Done.

This is great! Thank you @franz1981, @ehsavoie, @sebthom and @gtully for you invaluable input, help and effort with this one 👍

franz1981:
@uomik Just merged it, but...argh I see something strange...
I cannot see some of the changes I was expecting to see...Asre you sure you have squashed (using fixup) the commits or just deleted them?

uomik:
> @uomik Just merged it, but...argh I see something strange...
> I cannot see some of the changes I was expecting to see...Are you sure you have squashed (using fixup) the commits or just deleted them?

Thats strange. I used squash instead of fixup (and altered the commit message a bit). And did a diff with remote before pushing just to be sure. It did show no difference in code after the squash other than one extra commit from the upstream/master.

franz1981:
@uomik I can give you an example: https://github.com/uomik/activemq-artemis/pull/1/commits/8e2c7bab20b78556677275cb398542956df89cdb

uomik:
> @uomik I can give you an example: [uomik@8e2c7ba](https://github.com/uomik/activemq-artemis/commit/8e2c7bab20b78556677275cb398542956df89cdb)
> 
> This change seems disappeared from the commits of this PR

Yes, I can verify that this was already missing before I squashed. All the latest changes for Postgre large message handling are included, so it seems that some commits from the middle have disappeared. Seems I messed things up at some point,  already before the squash

franz1981:
@uomik I believe that both the 2 PRs I've sent you on your repo are missing: some of them are key to make both test and this PR to work so I'm going to send a follow-up PR that add the missing bits :+1:

PR #3206, opened 2020-06-29
ARTEMIS-2826 SSL configuration for 'data imp' tool

vkamble60:
Jira https://issues.apache.org/jira/browse/ARTEMIS-2826


jbertram:
Instead of adding a handful of new input parameters to deal with SSL/TLS specifics I recommend deprecating the `host` and `port` parameters and adding a `url` parameter. That way *any* parameter can be set if desired (include SSL/TLS stuff).

vkamble60:
@jbertram Thank you for recommendation. I was thinking same approach.. let me try with  URL parameter.

jbertram:
Closing this for now. Re-open when you've got a solution using the URL.

PR #3209, opened 2020-06-30
ARTEMIS-2797 Fixing user update Null

clebertsuconic:
UpdateQueueTest::testUpdateQueueWithFilter was broken.


clebertsuconic:
@jbertram can you please review this?

there was a test broken after #3201 

The definition of your test seems to clash with #3201 ... so I need your review here.

michaelandrepearce:
@clebertsuconic merge this. Can be fixed separately if there is need to remove a user in update.

clebertsuconic:
@michaelandrepearce I will run some more tests . I added another commit here trying to fix another issue with redeployment on the divert.

I want to run more tests before merging this... (I lost my window for the release anyways.. it's 4th July weekend and I have some personal time off as well).

PR #3211, opened 2020-07-02
ARTEMIS-2835 Porting HORNETQ-1575 and HORNETQ-1578

gaohoward:
ARTEMIS-2835 Porting HORNETQ-1575 and HORNETQ-1578
    
    2 of 2) - Porting of HORNETQ-1578
    
    Exceptions are swallowed, making it hard to diagnose issues
    
    1 of 2) - Porting of HORNETMQ-1575
    
    In a live-backup scenario, when live is down and backup becomes live, clients
    using HA Connection Factories can failover automatically. However if a
    client decides to create a new connection by itself (as in camel jms case)
    there is a chance that the new connection is pointing to the dead live
    and the connection won't be successful. The reason is that if the old
    connection is gone the backup will not get a chance to announce itself
    back to client so it fails on initial connection.
    
    The fix is to let CF remember the old topology and use it on any
    initial connection attempts.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
@@ -165,8 +165,6 @@ public final class ServerLocatorImpl implements ServerLocatorInternal, Discovery
 
    private TransportConfiguration clusterTransportConfiguration;
 
-   private boolean useTopologyForLoadBalancing;
-
    /** For tests only */
    public DiscoveryGroup getDiscoveryGroup() {
       return discoveryGroup;
@@ -422,7 +420,7 @@ public final class ServerLocatorImpl implements ServerLocatorInternal, Discovery
       clusterTransportConfiguration = locator.clusterTransportConfiguration;
    }
 
-   private TransportConfiguration selectConnector() {
+   private synchronized Pair<TransportConfiguration, TransportConfiguration> selectConnector(boolean useInitConnector) {
       Pair<TransportConfiguration, TransportConfiguration>[] usedTopology;
 
       flushTopology();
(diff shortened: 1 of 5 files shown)


clebertsuconic:
There are about 21 tests broke with this PR.

FederatedAddressTest mainly.


Can you address these failures before this can be merged please?

gaohoward:
ok will do.

clebertsuconic:
on a positive note, this made the test suite 10 min faster.

gaohoward:
that's maybe because those failed tests are failing earlier. :)

gaohoward:
@clebertsuconic I've updated the branch and checked all fedoration tests passed. Also investigated other tests and seems the RedeployTest has some real failures but not related to this PR.

clebertsuconic:
@gaohoward RedeployTest is now fixed on master.

Thank you


What was the issue?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
```diff
@@ -682,12 +689,32 @@ public ClientSessionFactory createSessionFactory() throws ActiveMQException {
...
+                           if (staticTried) {
+                              throw ActiveMQClientMessageBundle.BUNDLE.cannotConnectToServers();
+                           } else {
+                              topologyArrayTried = true;
+                              shouldTryStatic = true;
+                              attempts = 0;
+                           }
+                        }
                      }
                      if (factory.waitForRetry(config.retryInterval)) {
```
clebertsuconic, line 719: I see, I had an old copy of your PR, you removed these 3 extra lines.

I guess that would explain it.


thank you

gaohoward: actually I somehow removed the line
if (factory.waitForRetry(config.retryInterval)) {
...blah blah
During porting. It makes no sense. I guess I mistakenly removed it during porting.

clebertsuconic: thanks @gaohoward

PR #3212, opened 2020-07-03
NO-JIRA Fix #3172 mechanism

jsmucr:
Queue settings are reset to their default values upon broker.xml reload ONLY.
Regular calls to PostOfficeImpl#updateQueue are no longer affected.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
@@ -16,40 +16,6 @@
  */
 package org.apache.activemq.artemis.core.server.impl;
 
-import javax.management.MBeanServer;
-import java.io.File;
-import java.io.IOException;
-import java.io.PrintWriter;
-import java.io.StringWriter;
-import java.lang.management.ManagementFactory;
-import java.net.URL;
-import java.security.AccessController;
-import java.security.PrivilegedAction;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Date;
-import java.util.EnumSet;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Map.Entry;
-import java.util.Set;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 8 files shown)


jsmucr:
@clebertsuconic @michaelandrepearce This covers the #3209 issue too. That bug was caused by #3172 unintentionally affecting all `updateQueue` calls while originally I was targeting `broker.xml` updates only. I am sorry about such severe mistake.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -16,40 +16,6 @@
  */
 package org.apache.activemq.artemis.core.server.impl;
 
-import javax.management.MBeanServer;
```
clebertsuconic, line 19: can you avoid these lines on the diff?

it gets difficult to track changes over time.

I don't think you really needed these.

clebertsuconic:
I don't think this is a NO-JIRA.

Although you can just apply the previous JIRA given that we didn't release yet. Just use the previous JIRA-id on the commit here.

clebertsuconic:
@jsmucr although leave those two changes with me.. I will do some git-fu, checkout your branch.. make these changes.. and merge with you as the author.

clebertsuconic:
this is superseded by #3213 

please check

clebertsuconic:
there are tests failures as part of your changes:


org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlTest.testRemoveQueueFilter[legacyCreateQueue=true]
org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlTest.testRemoveQueueFilter[legacyCreateQueue=false]
org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlUsingCoreTest.testRemoveQueueFilter[legacyCreateQueue=true]
org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlUsingCoreTest.testRemoveQueueFilter[legacyCreateQueue=false]

jsmucr:
That's disappointing. I spent literally hours running tests to make sure everything's fine. 

I'll take look at that.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
jsmucr: Will do.

PR #3213, opened 2020-07-06
ARTEMIS-2797 Fixing redeploy mechanism

clebertsuconic:
Queue settings are reset to their default values upon broker.xml reload ONLY.
Regular calls to PostOfficeImpl#updateQueue are no longer affected.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -701,7 +701,7 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
             final SimpleString empty = new SimpleString("");
             Filter oldFilter = FilterImpl.createFilter(queue.getFilter() == null ? empty : queue.getFilter().getFilterString());
             Filter newFilter = FilterImpl.createFilter(queueConfiguration.getFilterString() == null ? empty : queueConfiguration.getFilterString());
-            if ((forceUpdate || newFilter != null) && !Objects.equals(oldFilter, newFilter)) {
+            if ((forceUpdate || newFilter != oldFilter) && !Objects.equals(oldFilter, newFilter)) {
                changed = true;
                queue.setFilter(newFilter);
             }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -63,7 +63,6 @@ import org.apache.activemq.artemis.core.server.RoutingContext;
 import org.apache.activemq.artemis.core.server.cluster.impl.MessageLoadBalancingType;
 import org.apache.activemq.artemis.core.server.group.GroupingHandler;
 import org.apache.activemq.artemis.core.server.impl.AddressInfo;
-import org.apache.activemq.artemis.core.server.impl.QueueConfigurationUtils;
 import org.apache.activemq.artemis.core.server.impl.QueueManagerImpl;
 import org.apache.activemq.artemis.core.server.impl.RoutingContextImpl;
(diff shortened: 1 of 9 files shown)


clebertsuconic:
this supersedes #3212

clebertsuconic:
I'm running a full set of tests.. it will take me 2.5 hours to complete it. if all good this will be merged. (unless someone complains about it in the next 2 and 1/2 hours).

clebertsuconic:
4 tests failures on this PR:

org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlTest.testRemoveQueueFilter[legacyCreateQueue=true]
org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlTest.testRemoveQueueFilter[legacyCreateQueue=false]
org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlUsingCoreTest.testRemoveQueueFilter[legacyCreateQueue=true]
org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlUsingCoreTest.testRemoveQueueFilter[legacyCreateQueue=false]

clebertsuconic:
@jsmucr Why such semantic change regarding updateQueue? that's a contract break on the API.. and a big semantic change.

As a matter of fact those testRemoteQueueFilter stopped working because of this.

clebertsuconic:
@jsmucr: never min, when I read this:

"Regular calls to PostOfficeImpl#updateQueue are no longer affected."

I thought it meant that updateQueue wouldn't no longer have any effect... I read it again.. and I see the real meaning.


I am adding a new commit to fix the offended tests.. if you could please review it?

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -701,7 +701,7 @@ public QueueBinding updateQueue(QueueConfiguration queueConfiguration, boolean f
             final SimpleString empty = new SimpleString("");
             Filter oldFilter = FilterImpl.createFilter(queue.getFilter() == null ? empty : queue.getFilter().getFilterString());
             Filter newFilter = FilterImpl.createFilter(queueConfiguration.getFilterString() == null ? empty : queueConfiguration.getFilterString());
-            if ((forceUpdate || newFilter != null) && !Objects.equals(oldFilter, newFilter)) {
+            if ((forceUpdate || newFilter != oldFilter) && !Objects.equals(oldFilter, newFilter)) {
```
clebertsuconic, line 704: @jsmucr here is the change that should fix the test.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -514,8 +514,8 @@ public void testRemoveQueueFilter() throws Exception {
 
       producer.send(m);
 
-      assertNotNull(consumer1.receiveImmediate());
-      assertNotNull(consumer2.receiveImmediate());
+      assertNotNull(consumer1.receive(1000));
```
clebertsuconic, line 517: @jsmucr: just a note here. if you intended to assertNotNull, do a receive with a timeout.

if you assertNull, do a receiveImmediate (or receiveNoWait in JMS terms).. or a very short timeout.

assertNotNull(receiveImmediate) is a bit fragile and it may break a test intermittently.

jsmucr: I'll keep that in mind. Thank you.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
jsmucr: Yes, this is what I originally intended. Thank you.

clebertsuconic: @jsmucr merged, thank you

PR #3215, opened 2020-07-09
A few fixes for audit logging

jbertram:


clebertsuconic:
Just FYI I already cut 2.14.0, this won't be part of the cut..

is that ok?

jbertram:
@clebertsuconic, yeah, I don't think it's a big deal.

PR #3219, opened 2020-07-15
ARTEMIS-2846 Cannot define hawtio.role with whitespace

brusdev:
Move the `hawtio.role` property definition to avoid the word splitting.


clebertsuconic:
@brusdev it would probably be a good idea to change artemis-distribution/src/test/scripts/validate-installation.sh to do some testing with spaces on the role? is there a way we could increase coverage through such test?

I do use this test on our internal CI.

brusdev:
@clebertsuconic I added a check for Artemis Jolokia REST API in artemis-distribution/src/test/scripts/validate-installation.sh and a test for the roles with spaces in org.apache.activemq.cli.test.ArtemisTest.

PR #3221, opened 2020-07-16
ARTEMIS-2847 socks5h support

swerner0:
Added 'socksRemoteDNS' transport parameter.
If set to true, remote destination socket is created unresolved and DNS resolution is disabled.


clebertsuconic:
tests?

swerner0:
Moved previous SOCKS tests into separate class and added test for my use case

jbertram:
@swerner0, can you squash the commits? Everything else looks good to me.

clebertsuconic:
I can actually squash it while I merge. don't worry about it.


I will merge this.... tests are good.

swerner0:
ok, thanks. lemme know if you need anything else

PR #3223, opened 2020-07-17
ARTEMIS-2849 Eliminate zeroing of buffers while writing the ASYNCIO journal

franz1981:
See https://issues.apache.org/jira/browse/ARTEMIS-2849

### DIFF
--- a/artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/AbstractSequentialFile.java
+++ b/artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/AbstractSequentialFile.java
@@ -256,11 +256,11 @@ public abstract class AbstractSequentialFile implements SequentialFile {
       return file;
    }
 
-   protected ByteBuffer newBuffer(int size, int limit) {
+   protected ByteBuffer newBuffer(int size, int limit, boolean zeroed) {
       size = factory.calculateBlockSize(size);
       limit = factory.calculateBlockSize(limit);
 
-      ByteBuffer buffer = factory.newBuffer(size);
+      ByteBuffer buffer = factory.newBuffer(size, zeroed);
       buffer.limit(limit);
       return buffer;
    }
@@ -271,7 +271,7 @@ public abstract class AbstractSequentialFile implements SequentialFile {
       public void flushBuffer(final ByteBuf byteBuf, final boolean requestedSync, final List<IOCallback> callbacks) {
          final int bytes = byteBuf.readableBytes();
          if (bytes > 0) {
-            final ByteBuffer buffer = newBuffer(byteBuf.capacity(), bytes);
+            final ByteBuffer buffer = newBuffer(byteBuf.capacity(), bytes, false);
             buffer.limit(bytes);
             byteBuf.getBytes(byteBuf.readerIndex(), buffer);
             buffer.flip();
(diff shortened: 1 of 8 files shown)


artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/AbstractSequentialFile.java
```diff
@@ -271,7 +271,7 @@ protected ByteBuffer newBuffer(int size, int limit) {
       public void flushBuffer(final ByteBuf byteBuf, final boolean requestedSync, final List<IOCallback> callbacks) {
          final int bytes = byteBuf.readableBytes();
          if (bytes > 0) {
-            final ByteBuffer buffer = newBuffer(byteBuf.capacity(), bytes);
+            final ByteBuffer buffer = newBuffer(byteBuf.capacity(), bytes, false);
```
clebertsuconic, line 274: In case we pool a buffer, and say you wrote 300 bytes, on a place that previously had 4K filled

you will fill it with 300 bytes... the buffer will be flipped, but when the libaio write takes place.. I'm afraid this could not cleanup the rest of the 4K.

I did not run any tests on this.. just from what I remember on the pool mechanism.

just make sure the previously pooled buffer wouldn't make to the journal.

clebertsuconic: Also: did you run the whole test suite on such change?

franz1981: Every borrowed pooled buffer always start at position 0, so the observer, while flushing always overwrite exactly the bytes it has requested despite being zeroed or not

> Also: did you run the whole test suite on such change?

Yet to do it, will do it now :)
NIO and MAPPED already use this optimization from long time (the jira link explain it): ASYNCIO was the only one missing yet

franz1981: About the tests: I see that only this one has failed

```
Error Message
expected:<10> but was:<11>
Stacktrace
java.lang.AssertionError: expected:<10> but was:<11>
	at org.apache.activemq.artemis.tests.unit.core.postoffice.impl.DuplicateDetectionUnitTest.testReloadDuplication(DuplicateDetectionUnitTest.java:127)
```
And probably it isn't related

clebertsuconic: I actually had access to the same build as you, and as I can see the build did not complete.. lots of failures due to open files. I will restart that build

clebertsuconic: @franz1981 the failure is real... this PR is broken.

franz1981:
@clebertsuconic The fix seems to work fine, but I'm not very proud of the solution: if we want to write 128 bytes and the block size is 4096, `bufferControl` allocates (or can reuse) a 4096 sized ByteBuffer without zeroing it (that's nice), but the journal record still need to zero `4096 - 128` bytes to save having garbage data on the journal. If we write enough data this optimization can save lot of zeroing but cannot eliminate it.

clebertsuconic:
that's a result of reusing the buffers from previous writes. Which is what I mentioned to you.

on Mapped and NIO there's no alignment, meaning you don't need to set zeroes to a previously reused buffer.

With libaio we have our own pool... I think this is expected.

clebertsuconic:
squash the commits and I will merge this

PR #3226, opened 2020-07-28
ARTEMIS-2856 consumer can be created on closed session

jbertram:
Due to a lack of concurrency protections it's possible to create a
consumer on a closed session. I've not been able to reproduce this with
a test, but I've seen it in the wild. Static code analysis points to a
need for better concurrency controls around closing the session and
creating consumers.


clebertsuconic:
I'm running a full set of tests before I merge this. I'm leaving the tests running overnight and if they are good I will merge it tomorrow morning.

PR #3228, opened 2020-08-03
ARTEMIS-2854 Non-durable subscribers stop receiving after failover

gaohoward:
In a cluster scenario where non durable subscribers fail over to
backup while another live node forwarding messages to it,
there is a chance that the the live node keeps the old remote
binding for the subs and messages go to those
old remote bindings will result in "finding not found".


clebertsuconic:
@gaohoward did you run the extensive test suite on this change? Clustering is a dangerous area, especially that you are messing up with remote records.

I need to know what tests you ran as part of this.

Meanwhile I started a full test suite on my internal CI machine.

gaohoward:
@clebertsuconic thanks. I didn't run the testsuite on ci machine.
I think it shouldn't be causing any troubles. :)

clebertsuconic:
@gaohoward please rebase this against master...

run a full test suite and let us know the results?

gaohoward:
I'm starting the test suite. (the travis-ci seems to be hanging)

gaohoward:
testing still in progress

gaohoward:
@clebertsuconic test suite finished with 6 failures but I don't think they are related.
So the PR is safe to merge.

PR #3235, opened 2020-08-05
NO-JIRA: fix RandomUtilDistributionTest assertion and improve failure message

gemmellr:
I saw RandomUtilDistributionTest fail in a CI run. Checking, the error message indicated that all tries were duplicated and a value of 40 didnt satisfy the minimum expected value of 40.

> [ERROR] Failures: 
[ERROR]   RandomUtilDistributionTest.testDistribution:73 The Random distribution is pretty bad. All tries have returned duplicated randoms. value=40, minimum expected = 40

This tweaks the assertion to use >= so as to actually allow the expected minimum count to pass, since 40 was at least 40, and by using > the restriction was over 10% more stringent than it suggests it is trying to be. Also tweaks the message to indicate many rather than all were duplications, since only '1' would indicate the latter.

Given the limited selection choices available im surprised this hasnt failed more often. I rather wonder if its a test worth actually throwing >5 seconds and ~2.5GB of mem at to begin with, but this just fixes the actual issues spotted for now, other adjustments can be made again...

### DIFF
--- a/tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/util/RandomUtilDistributionTest.java
+++ b/tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/util/RandomUtilDistributionTest.java
@@ -69,9 +69,8 @@ public class RandomUtilDistributionTest {
       // Be careful removing it (make sure you know what you're doing in case you do so)
       int minimumExpected = (int) ((iterations * numberOfStarts) * 0.80);
 
-      log.debug("value=" + value + ", minimum expected = " + minimumExpected);
-      Assert.assertTrue("The Random distribution is pretty bad. All tries have returned duplicated randoms. value=" + value + ", minimum expected = " + minimumExpected, value > minimumExpected);
-
+      log.debug("values = " + value + ", minimum expected = " + minimumExpected);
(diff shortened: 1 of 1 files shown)


clebertsuconic:
Back in 2016 I had the case of needing to test Randoms.

The case was.. you started multiple clients.. they were not actually random, since the seed was using a similar value... 

I don't remember some of the details as this has been 5 years now (actually, pretty impressive, when you realized how fast time is going).

I think by now, this test is probably not needed. I don't foresee anyone tweaking with randomUtil.

We can remove it if you are strongly opinionated. 

I don't want to invest much time on it as it would not have much benefit over the real codebase.

gemmellr:
Im not strongly opinionated on removing it, I dont think its particularly valuable but its fine with leaving it if anyone does, hence fixing the obvious issue and making it slightly less likely to fail, rather than a PR to delete it. If it fails anything more than a trivial amount however I would say it should either be fully reworked (dont think its worth that), made signifcantly less stringent (e.g 60% rather than 80% allowance) or just removed. I cant say I've ever seen it fail before it did yesterday, but humorously it did actually fail a second time on my fork while I prepared this PR, getting 39 out of required 40.

gemmellr:
I raised https://issues.apache.org/jira/browse/ARTEMIS-2874 for tracking the other test failure this PR hit, since it seems to occur a lot.

PR #3236, opened 2020-08-05
[ARTEMIS-2873] Ensure configuation managed queues are not auto delete…

michaelandrepearce:
…d these should only be removed if removed in configuration.

Auto Delete Queues and Auto Delete Created Queues should only apply to NON configuration managed queues.


michaelandrepearce:
replaced by https://github.com/apache/activemq-artemis/pull/3237

PR #3237, opened 2020-08-05
[ARTEMIS-2873] Ensure configuation managed queues are not auto delete…

michaelandrepearce:
…d these should only be removed if removed in configuration.

Auto Delete Queues and Auto Delete Created Queues should only apply to NON configuration managed queues.


jbertram:
I was going to merge this, but the test succeeds even without the fix.

michaelandrepearce:
@jbertram good catch, i missed a local change, i forgot to commit, queueConfigurationManaged was under wrong address, so didnt pick up the address settings. Should be failing without the change and passing with.

PR #3238, opened 2020-08-06
ARTEMIS-2859 Wildcard topic test cases resulting in negative address size warnings from broker

swerner0:


swerner0:
We were initially seeing an issue with clustered topics, but then saw negative address size warnings on a single broker when we had a consumer on the full topic and another on a wildcard, thought it looked strange. Also relates to ARTEMIS-2768

gtully:
this test is included with the fix in #3265
nice test :-), much appreciated. Thanks!

PR #3241, opened 2020-08-13
ARTEMIS-2878 Add numberOfPages as metric

bgutjahr:
Added metric 'pages.count' to provide numberOfPages for an address.

Please let me know if 'pages.count' is OK or if I should rename the metric to something different.
I felt that a metric name like 'number.of.pages' or 'pages.numberof' looked strange.


jbertram:
The name `number.of.pages` may look a bit odd, but I think consistency in naming between metrics and the management API is important. Can you rename it? Thanks!

bgutjahr:
I have renamed the metric to `number .of.pages`

PR #3242, opened 2020-08-13
ARTEMIS-2877 Fix journal replication scalability

franz1981:
This will include:

- reverting ARTEMIS-2837 Avoiding bursts on writes and pending callbacks
- reimplementing it by limiting the amount of opened files (@clebertsuconic need you here: probably limiting the capacity of `openedFiles` is fine too?)
- using netty batching to allow bursty journal appends during replication (this one should both help with https://issues.apache.org/jira/browse/ARTEMIS-2852 and @michaelandrepearce concern of perf regression on replication, so please take a look)

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/RemotingConnectionImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/RemotingConnectionImpl.java
@@ -392,6 +392,15 @@ public class RemotingConnectionImpl extends AbstractRemotingConnection implement
    }
 
    @Override
+   public void endOfBatch(Object connectionID) {
+      super.endOfBatch(connectionID);
+      // TODO we really need a lock here?
+      synchronized (transferLock) {
+         channels.forEach((channelID, channel) -> channel.endOfBatch());
+      }
+   }
+
+   @Override
    public String getTransportLocalAddress() {
       return getTransportConnection().getLocalAddress();
    }
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
@@ -1227,6 +1227,27 @@ public class ClientSessionFactoryImpl implements ClientSessionFactoryInternal, C
             logger.debug("TheConn == null on ClientSessionFactoryImpl::DelegatingBufferHandler, ignoring packet");
          }
       }
+
+      @Override
+      public void endOfBatch(final Object connectionID) {
+         RemotingConnection theConn = connection;
+
+         if (theConn != null && connectionID.equals(theConn.getID())) {
+            try {
+               theConn.endOfBatch(connectionID);
+            } catch (final RuntimeException e) {
+               ActiveMQClientLogger.LOGGER.disconnectOnErrorDecoding(e);
+               threadPool.execute(new Runnable() {
+                  @Override
+                  public void run() {
+                     theConn.fail(new ActiveMQException(e.getMessage()));
+                  }
+               });
+            }
+         } else {
+            logger.debug("TheConn == null on ClientSessionFactoryImpl::DelegatingBufferHandler, ignoring packet");
+         }
+      }
    }
 
    private final class DelegatingFailureListener implements FailureListener {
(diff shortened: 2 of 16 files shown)


franz1981:
The CI has gone well, but here I see something has gone wrong, no idea why: please take a look if you can or I will on my back

franz1981:
Just few numbers to show the amount of performance improvement re this:
2 brokers with replication using MAPPED journal configured as https://issues.apache.org/jira/browse/ARTEMIS-2852

with 1 producer/consumer 1000 bytes messages using core protocol

pre PR:
```
**************
EndToEnd Throughput: 4410 ops/sec
**************
mean                231.26
min                 168.96
50.00%              226.30
90.00%              259.07
99.00%              333.82
99.90%              475.14
99.99%             2670.59
max               10354.69
count               100000
```
this PR:
```
**************
EndToEnd Throughput: 5106 ops/sec
**************
mean                201.19
min                 141.31
50.00%              196.61
90.00%              231.42
99.00%              294.91
99.90%              364.54
99.99%              667.65
max                2605.06
count               100000
```
2.2:
```
**************
EndToEnd Throughput: 4718 ops/sec
**************
mean                217.82
min                 152.58
50.00%              209.92
90.00%              256.00
99.00%              333.82
99.90%              724.99
99.99%             2392.06
max                4587.52
count               100000
```

The end 2 end latencies are being reported just for information although are not comparable because the throughputs are very different.

Let's what append with 32 producer-consumers using 1000 bytes messages, instead

franz1981:
Regardless the case, this PR is improving over 2.2 (that has been mentioned on https://issues.apache.org/jira/browse/ARTEMIS-2852 as the best performed) and over pre-PR (without the commits on this PR) too.
And the is more to come (after this PR) :)

michaelandrepearce:
Its pretty bad that we had such a perf drop. Have you tested with the actual setup and scenario theyre using. Also i would suggest a  release is performed thats a public blog and test case it be best to have it public against a tag and released version

michaelandrepearce:
Also are we sure where the perf drop was introduced? E.g. has their perf test setup been run on some of the previous releases to see where/which version it dropped in?

franz1981:
Yep, I have tested exactly with their test scenario. I am preparing some regression perf tests to check against regressions of such type (when I am back from PTO).

TBH ASYNCIO isn't getting such bad hit on performance as MAPPED, unless a very fast disk is being used and the rate of appends is very high.
I agree that we should have something on upstream to check against this perf regression and already spoken with @clebertsuconic about which approach to use ;)
Nonetheless this PR is not just fixing the scalability issue but is actually improving beyond it

michaelandrepearce:
Yes but i think once this is merged we should call for a release spin quickly. That blog js very public and well known. It be pretty bad to have them report this issue or to see artemis slip in its position

franz1981:
Agree although consider that I have found their topology to be completely wrong from the beginning and the test itself wasn't testing what it was thinking to test. Anyway it has been useful for me to spot a real perf regression :)

michaelandrepearce:
Well yes but it still puts artemis in good light. And to be honest once we get new leadership election stuff in it wouldnt be actually that bad

michaelandrepearce:
Do we have any insight in when they plan to publish. Not sure how fast clebert can spin a release but ill be waiting to give my +1

franz1981:
@michaelandrepearce no idea, i believe they need to fix the benchmark first, so feel free to step in and propose th correct clustering config on their issue on https://issues.apache.org/jira/browse/ARTEMIS-2852

If I work a bit more during the PTO my wife will kill me :)

michaelandrepearce:
i dont think they need to fix anything, they have a performance metric and in that particular setup, we've massively regressed. we just need to ship a new version asap with that regression reverted.

franz1981:
@michaelandrepearce agree about the new version, but they said that they need to fix the benchmark setup/configuration; see https://issues.apache.org/jira/browse/ARTEMIS-2852 comments

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/RemotingConnectionImpl.java
```diff
@@ -391,6 +391,15 @@ public void bufferReceived(final Object connectionID, final ActiveMQBuffer buffe
       }
    }
 
+   @Override
+   public void endOfBatch(Object connectionID) {
+      super.endOfBatch(connectionID);
+      // TODO we really need a lock here?
```
clebertsuconic, line 397: Can you clear this TODO before we merge it?

clebertsuconic: That's my only point on this PR.

franz1981: I don't have my laptop around, but I can try do it tomorrow somehow...
The answer in the to-do is still valid: do we need a lock there?

PR #3252, opened 2020-09-01
ARTEMIS-2696 Releasing ByteBuf after reading content

ldebello:
We try to migrate from Artemis 2.9.0 to 2.13.0 and we got a OutOfDirectMemoryError. After some investigation we found the issue https://issues.apache.org/jira/projects/ARTEMIS/issues/ARTEMIS-2696 which points out to WebSocketFrameEncoder. So we include a log line in netty to track PlatformDependent#incrementMemoryCounter and PlatformDependent#decrementMemoryCounter but we only see calls to incrementMemoryCounter without any "decrementMemoryCounter". 

The WebSocketFrameEncoder was added in https://github.com/apache/activemq-artemis/commit/9fac4b866cf9cefbb6f7c13b820e07455b6649f5#diff-924a36a0e320185d284c29e46ddfdb75 changing the logic of wrapping the buffer into BinaryWebSocketFrame now the buffer is copied and the copy is wrapped in BinaryWebSocketFrame and if the size is bigger than the max using a continuation frame but the original buffer is not release.

I did this small change a run "mvn clean package" now I am creating the PR to check if this change breaks something because when trying to execute the build locally using "mvn -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Djdk8-errorprone -Pfast-tests -Pextra-tests -Ptests-CI -B install -q" it was not working for me even without doing any change.

I will also testing this change in our current env

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/protocol/websocket/WebSocketFrameEncoder.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/protocol/websocket/WebSocketFrameEncoder.java
@@ -63,5 +63,7 @@ public class WebSocketFrameEncoder extends ChannelOutboundHandlerAdapter {
          byteBuf.readBytes(fragment, length);
          ctx.writeAndFlush(new ContinuationWebSocketFrame(finalFragment, 0, fragment), promise);
       }
+
+      byteBuf.release();
    }
 }


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/protocol/websocket/WebSocketFrameEncoder.java
```diff
@@ -63,5 +63,7 @@ private void writeContinuationFrame(ChannelHandlerContext ctx, ByteBuf byteBuf,
          byteBuf.readBytes(fragment, length);
          ctx.writeAndFlush(new ContinuationWebSocketFrame(finalFragment, 0, fragment), promise);
       }
+
+      byteBuf.release();
```
clebertsuconic, line 67: would be possible to add a test?

ldebello: Yes I will add a test, unit test should be easy anyway I am trying to understand how to reproduce it in a consistent way because connecting and desconnecting does not trigger the issue at least inmac, assuming due to kqueue vs the linux impl. I will try to dig more if you have an idea about a good test I can try to create it

clebertsuconic: @ldebello a mock test perhaps?

ldebello: I have added a test using a direct buffer to verify the release is called. the test is using a direct buffer and I checked the increment and decrement is called I was not able to add that condition in the test because netty does not expose that information

ldebello: is the test good enough?

ldebello:
@clebertsuconic could you help me with the build? it is failing but not sure why, it mentions a maven error. Another question do we need to point this PR to 2.6.x, isnt it?

ldebello:
> @clebertsuconic could you help me with the build? it is failing but not sure why, it mentions a maven error. Another question do I need to point this PR to 2.6.x, isnt it?

forget my comment I double check the output from the build and the issue is the checkstyle  ,fixing that

artemis-server/src/test/java/org/apache/activemq/artemis/core/server/protocol/websocket/WebSocketFrameEncoderTest.java
```diff
@@ -78,7 +78,7 @@ public void testWriteNonByteBuf() throws Exception {
 
    @Test
    public void testWriteReleaseBuffer() throws Exception {
-      String content = "Using direct buffer";
```
clebertsuconic: git commit --amend
git push -f

(without changing anything would do to force a build) :)

clebertsuconic:
@ldebello You asked about 2.6.x... I haven't been maintaining 2.6.x.. it's too old now...

do you need a patch release with this?

artemis-server/src/test/java/org/apache/activemq/artemis/core/server/protocol/websocket/WebSocketFrameEncoderTest.java
ldebello: thanks for the tip :)

ldebello:
> @ldebello You asked about 2.6.x... I haven't been maintaining 2.6.x.. it's too old now...
> 
> do you need a patch release with this?

no, we can wait unti 2.16. just asking because I saw the branch there and maybe you use that for development. So if there PR against master is ok that is for me.

Do you want me to test something else? or are we good to merge it? I am still running the artemis instance in our lower envs checking the fix. I will keep the instance running for at least a week more. also I have artemis 2.15 running in another env just deployed today.

PR #3254, opened 2020-09-04
ARTEMIS-2886 put address/FQQN into new security manager interface

jbertram:
The default JAAS security manager doesn't need the address/FQQN for
authorization, but I'm putting it back into the interface because there
are other use cases which *do* need it.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
@@ -270,7 +270,7 @@ public class SecurityStoreImpl implements SecurityStore, HierarchicalRepositoryC
          final Boolean validated;
          if (securityManager instanceof ActiveMQSecurityManager5) {
             Subject subject = getSubjectForAuthorization(session, ((ActiveMQSecurityManager5) securityManager));
-            validated = ((ActiveMQSecurityManager5) securityManager).authorize(subject, roles, checkType);
+            validated = ((ActiveMQSecurityManager5) securityManager).authorize(subject, roles, checkType, isFullyQualified ? fqqn.toString() : bareAddress.toString());
          } else if (securityManager instanceof ActiveMQSecurityManager4) {
             validated = ((ActiveMQSecurityManager4) securityManager).validateUserAndRole(user, session.getPassword(), roles, checkType, bareAddress.toString(), session.getRemotingConnection(), session.getSecurityDomain()) != null;
          } else if (securityManager instanceof ActiveMQSecurityManager3) {
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/ActiveMQJAASSecurityManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/ActiveMQJAASSecurityManager.java
@@ -113,7 +113,8 @@ public class ActiveMQJAASSecurityManager implements ActiveMQSecurityManager5 {
    @Override
    public boolean authorize(final Subject subject,
                             final Set<Role> roles,
-                            final CheckType checkType) {
+                            final CheckType checkType,
+                            final String address) {
       boolean authorized = false;
 
       if (subject != null) {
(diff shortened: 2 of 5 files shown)


clebertsuconic:
LGTM... feel free to merge it

luisalves00:
`if (bareQueue != null) {
            fqqn = CompositeAddress.toFullyQualified(bareAddress, bareQueue);
            if (securityRepository.containsExactMatch(fqqn.toString())) {
               roles = securityRepository.getMatch(fqqn.toString());
               isFullyQualified = true;
            }
         }`

What is the securityRepository used for? Currently it is not matching the **fqqn** (against what?) and therefore I don't receive the **fqqn** in the ActiveMQSecurityManager5 implementation.
Is it possible to says if address is prefixed by "keyword" than always match?

luisalves00:
Didn't get and feedback. Should I open a new feature request?

jbertram:
I recommend sending an email to the ActiveMQ user mailing list. This PR was merged 22 days ago. The time to comment on the way it behaves was *before* it was merged. It's worth starting a new conversation at this point. Either that or send your own PR and we can discuss that.

luisalves00:
@jbertram sorry, next time I'll try to check it out earlier. I'll try to come up with an PR, but I kind of worried that I did not fully understood why did you use the exact match instead of a wildcard match.

jbertram:
Detecting the use of an exact match is important so I wouldn't change that. I recommend something really simple like this:
```java
validated = ((ActiveMQSecurityManager5) securityManager).authorize(subject, roles, checkType, fqqn != null ? fqqn.toString() : bareAddress.toString());
```
So instead of using the `isFullyQualified` `boolean` it will just look at the `fqqn` `SimpleString`. I think that would get you what you want.

luisalves00:
That seems quite simple, but ActiveMQSecurityManager5 implementations need to be aware they'll receive a fqqn even if they don't have an exact match in the security-settings. Is ActiveMQSecurityManager5 already released?

jbertram:
`ActiveMQSecurityManager5` hasn't been released yet so it doesn't matter. We're still free to change it at this point without backwards-compatibility considerations.

luisalves00:
Great. I like your proposal as it solve all the problems I have now, but what about the roles passed to the authorize method? (I don't used them but others might need them) I think the whole point of your IF is to push the correct roles into the method, no? I'll have a more detailed look tomorrow.

jbertram:
I'm not following you. Please elaborate on what you mean.

luisalves00:
So...let me explain with the code (comments inline):

```java

//best match for roles using the wildcards
Set<Role> roles = securityRepository.getMatch(bareAddress.toString());
...
if (bareQueue != null) {
   fqqn = CompositeAddress.toFullyQualified(bareAddress, bareQueue);
   if (securityRepository.containsExactMatch(fqqn.toString())) {
      //if I don't have an exact match I don't set the roles here
      //and this seems to be the main purpose of the IF, i.e. to find specific roles to the fqqn, no?
      roles = securityRepository.getMatch(fqqn.toString());
      isFullyQualified = true;
   }
}
```
so now calling the authorize with the code you've propose:

```java
validated = ((ActiveMQSecurityManager5) securityManager).authorize(subject, roles, checkType, fqqn != null ? fqqn.toString() : bareAddress.toString());
```

the roles passed to the method will not be the ones of the exact match, right?
As I said, I don't need the roles, I'll ignore them as I'm going to get them from an access token, but might be important to other users that implement the ActiveMQSecurityManager5.

jbertram:
Are you saying the change I proposed changes how `roles` is treated? As far as I can tell the change I proposed has nothing to do with the `roles` passed to `authorize`. It's only impacting the `address` parameter of `authorize`. Am I wrong about that?

Aside from that, the basic idea here is that if there is an exact match for the FQQN then I pass those roles to `authorize` otherwise I pass whatever roles came from the bare address match. It should be pretty straight-forward.

My change *will* allow a new semantic. Namely, it will allow the FQQN to be passed to `authorize` even if there isn't an exact match. I believe this is the semantic change you asked for, no?

I am not seeing a problem here.

luisalves00:
I think you are totally correct :) thanks.

luisalves00:
Did a pull request with your recommendation and some changes in the caching as well.

PR #3255, opened 2020-09-07
ARTEMIS-2875 - Failed to handle re-establish connection failover

skrutzler:
While handling a re-establish connection on failover it can happen that the client can connect to the broker and re-establish a connection, but while reattaching all sessions to the new connection another connection loss can happen and there the failover fails without any possibility to recover on its own.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
@@ -632,7 +632,21 @@ public class ClientSessionFactoryImpl implements ClientSessionFactoryInternal, C
 
                connector = null;
 
-               reconnectSessions(oldConnection, reconnectAttempts, me);
+               boolean allSessionReconnected;
+               int failedReconnectSessionsCounter = 0;
+               do {
+                  allSessionReconnected = reconnectSessions(oldConnection, reconnectAttempts, me);
+                  if (oldConnection != null) {
+                     oldConnection.destroy();
+                  }
+
+                  if (!allSessionReconnected) {
+                     failedReconnectSessionsCounter++;
+                     oldConnection = connection;
+                     connection = null;
+                  }
+               }
+               while ((reconnectAttempts == -1 || failedReconnectSessionsCounter < reconnectAttempts) && !allSessionReconnected);
 
                if (oldConnection != null) {
                   oldConnection.destroy();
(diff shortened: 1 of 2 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
```diff
@@ -632,11 +632,21 @@ private void failoverOrReconnect(final Object connectionID,
...
+                  allSessionReconnected = reconnectSessions(oldConnection, reconnectAttempts, me);
+                  if (oldConnection != null) {
+                     oldConnection.destroy();
+                  }
+                  oldConnection = connection;
 
-               if (oldConnection != null) {
-                  oldConnection.destroy();
+                  if (!allSessionReconnected) {
+                     failedReconnectSessionsCounter++;
```
clebertsuconic, line 645: shouldn't you close the connection before setting it to null?


also... do you have a test showing the issue?


And 3rd... can you amend the commit with the JIRA?

```sh
git commit --amend
# on Vim or your configured editor, add the JIRA to the commit message (not the GitHub title)
git push origin -f
```

(If you happen to have a test meanwhile you're doing that?)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/remoting/ReconnectTest.java
```diff
@@ -361,7 +361,7 @@ public boolean intercept(Packet packet, RemotingConnection connection) throws Ac
 
       final long retryInterval = 50;
       final double retryMultiplier = 1d;
-      final int reconnectAttempts = 10;
+      final int reconnectAttempts = 1;
```
clebertsuconic, line 364: this is the only test you changed... would this be enough to replicate the issue?


shouldn't we parameterize the test to add this condition, since this seems to be changing other tests...


I need some context on why you changed this.. if this is your test or what is this?


thanks!

skrutzler: this test never used the reconnectAttempts because on a failed connection while processing a failover we immediatelly set the FailoverEventType.FAILOVER_FAILED. With my changes we now retry to reconnect for as often as reconnectAttempts is set, therefore on second try we would succeed to reconnect and never send the FailoverEventType.FAILOVER_FAILED.

This test was not done by me but shows that we do not try to reconnect for the specified amount of times.

Adding a test for the issue I wanted to solve seems rather tricky because the issue is only present when the client can connect to the broker and while trying to handle the failover on the session looses the new connection.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
skrutzler: 1. the connection is not closed because the oldConnection is still using it. but the oldConnection might not get closed, I will fix that

2. quite tricky to come up with a test as described in the other comment

3. will do

clebertsuconic:
@skrutzler nice one!!! thanks a lot for this PR!

skrutzler:
@clebertsuconic glad I could help. Do you know when a new release with this fix will be made?

PR #3258, opened 2020-09-11
ARTEMIS-589 flow control for individual STOMP subscribers

jbertram:


rtista (changes requested):
The changes made seem to be perfect for a successful flow control implementation in the STOMP protocol. The "0" value defaulting to 1 message is indeed a very nice touch on the implementation as well as the parameter normalization facing other protocols and deprecation of the _stompConsumerCredits_ parameter.

I have, however, one concern, by supporting the _activemq.prefecthSize_ for backwards compatibility with very different units (message count _VS_ message size in bytes) this may cause some client applications to be unable receive messages. If I have an _activemq.prefectSize_ of 5, this would be parsed as 5 bytes and that would render all consumers with that configuration unable to process any messages.

Should we attempt to maintain a more ActiveMQ 5.x like behavior by multiplying the _activemq.prefectSize_ by a default message size? Or maybe even not support it at all? I would rather have a migration to Artemis culminate on consumers only being able to consume a reduced amount of messages than not consuming at all.

Regardless of the decision, thank you very much and great job @jbertram!

jbertram:
I am concerned that the change in semantics for `activemq.prefetchSize` will be confusing for users despite the details in the documentation. Therefore, it may be better *not* to support `activemq.prefetchSize` as you suggest.

However, there is no setting for either `consumerWindowSize` or `activemq.prefetchSize` that will prevent consumers from receiving messages so your specific concern about having an `activemq.prefetchSize` of `5` is not valid. Even if the setting is `0` consumers will still get messages. They will just get them one at a time, as the documentation notes.

I actually did consider multiplying `activemq.prefetchSize` by a default size (as you mentioned), but I decided against it because use-cases can vary so widely and the more translation happens the harder it is to understand and the greater the chance of bugs and confusion for users. The simpler the implementation the better, in my opinion.

In any case, consumers will still get their messages although likely more slowly than before. Take the branch, build it, try it out and let me know if you find it behaves otherwise.

PR #3260, opened 2020-09-14
ARTEMIS-2901 support namespace for temporary queues

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/TempQueueNamespaceTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+      session.createQueue(new QueueConfiguration(queue).setAddress(address).setDurable(false).setTemporary(true));
+
+      assertEquals(10, (long) server.locateQueue(queue).getQueueConfiguration().getRingSize());
+
+      session.close();
+   }
+
+   @Test
+   public void testTempQueueNamespaceNegative() throws Exception {
```
brusdev, line 44: What is the purpose of this test?

clebertsuconic: @brusdev he's verifying the name was not applied when there was no match.

PR #3263, opened 2020-09-15
ARTEMIS-2903 support admin objects in JCA RA

jbertram:


clebertsuconic:
I have no way to review this.. if you are sure it's ok, please go ahead and merge it

PR #3265, opened 2020-09-16
ARTEMIS-2888 - ensure the pageStore uses the publish address rather t…

gtully:
…han the subscriptions value which may have a wildcard

There is still something odd, the test works however I see an unexpected WARN:
[Thread-2 (activemq-netty-threads)] 17:54:02,694 WARN  [org.apache.activemq.artemis.core.server] AMQ222038: Starting paging on address 'A.*'; size is currently: 1,025 bytes; max-size-bytes: 50; global-size-bytes: 1,025

note the address is A.*! 
I think that is wrong. more investigation needed. maybe folks more familiar will see what is going on.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/QueueConfiguration.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/QueueConfiguration.java
@@ -109,6 +109,7 @@ public class QueueConfiguration implements Serializable {
    private Boolean internal;
    private Boolean _transient;
    private Boolean autoCreated;
+   private transient SimpleString pageStoreName;
 
    /**
     * Instantiate this object and invoke {@link #setName(SimpleString)}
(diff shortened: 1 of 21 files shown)


clebertsuconic:
lets talk about this when you can?

gtully:
sure, there is also a test failure that I will investigate: org.apache.activemq.artemis.core.protocol.mqtt.MQTTRetainMessageManagerTest

jbertram:
I'm pretty sure this is related to both [ARTEMIS-2859](https://issues.apache.org/jira/browse/ARTEMIS-2768) and [ARTEMIS-2768](https://issues.apache.org/jira/browse/ARTEMIS-2768) as the size of the page-store underneath the address is being manipulated incorrectly when messages are sent & consumed.

gtully:
@jbertram  can you have a quick peek over this to see if i missed anything on the xml attribute addition and doc

clebertsuconic:
this LGTM...

@jbertram since you're reviewing this per Gary's request if you could merge it when done please?

jbertram:
@gtully, can you grab the tests from #3238 and add them here? Other than that it looks good.

gtully:
@jbertram  thanks, that is a nice test. pulled it in and proves the point that this new page-store-name setting is needed even if paging does not kick in because the dispatch is always through cursors. updated the doc to reflect that.

PR #3269, opened 2020-09-21
ARTEMIS-2910: consider routing type annotations during auto-creation

gemmellr:
Inspects messages sent via anonymous-producers for routing type details to use in auto-creation of the appropriate type of address/queues, instead of just the address name (similar to the additional terminus detail that can be considered when fixed-destination senders are in used). Adds or enhances various tests around auto-creation and anonymous producers.

More detail at https://issues.apache.org/jira/browse/ARTEMIS-2910


tabish121:
+1 lgtm

PR #3273, opened 2020-09-23
ARTEMIS-2912 Server start exception before activation can cause a zombie broker

franz1981:
@clebertsuconic I'm awaiting this to complete running on our CI first
And I would provide some better test too


franz1981:
@jbertram IF you want to take a look bud :)
I would like to request reviewers but seems I cannot do it, there is some missing authorizations for me?

franz1981:
I need to think better about this: the solution isn't working for the 2 failure cases (stop issued)

PR #3274, opened 2020-09-24
ARTEMIS-2912 Server start exception before activation can cause a zombie broker

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-2912

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
@@ -1089,17 +1089,24 @@ public class ActiveMQServerImpl implements ActiveMQServer {
          state = SERVER_STATE.STOPPING;
 
          if (criticalIOError) {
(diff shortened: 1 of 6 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -1075,17 +1089,24 @@ void stop(boolean failoverOnServerShutdown, final boolean criticalIOError, boole
          state = SERVER_STATE.STOPPING;
 
          if (criticalIOError) {
-            // notifications trigger disk IO so we don't want to send any on a critical IO error
-            managementService.enableNotifications(false);
+            final ManagementService managementService = this.managementService;
```
brusdev, line 1092: why do we need this variable? who is setting to null this variable on stopping?

franz1981: It's thanks to Java Language Specification: the local `managementService` cannot change once read

PR #3277, opened 2020-09-24
ARTEMIS-2768 - add warning if wildcard subscription does not have a m…

gtully:
…atching page-store-name address setting

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
@@ -103,6 +103,7 @@ import org.apache.activemq.artemis.core.postoffice.Binding;
 import org.apache.activemq.artemis.core.postoffice.BindingType;
 import org.apache.activemq.artemis.core.postoffice.PostOffice;
 import org.apache.activemq.artemis.core.postoffice.QueueBinding;
+import org.apache.activemq.artemis.core.postoffice.impl.AddressImpl;
 import org.apache.activemq.artemis.core.postoffice.impl.DivertBinding;
 import org.apache.activemq.artemis.core.postoffice.impl.LocalQueueBinding;
 import org.apache.activemq.artemis.core.postoffice.impl.PostOfficeImpl;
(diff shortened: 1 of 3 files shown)


gtully:
@clebertsuconic is there a better place for this check? It seems as if addressInfo should know about wildcards but that logic is currently in addressImpl and the address manager.

hmm, maybe it also needs to be conditional on wildcard routing being enabled, maybe folks run with it disabled for a tiny perf lift

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -3512,7 +3513,14 @@ public Queue createQueue(final QueueConfiguration queueConfiguration, boolean ig
          }
       }
 
-      QueueConfigurationUtils.applyDynamicQueueDefaults(queueConfiguration, addressSettingsRepository.getMatch(getRuntimeTempQueueNamespace(queueConfiguration.isTemporary()) + queueConfiguration.getAddress().toString()));
+      final AddressSettings addressSettings = addressSettingsRepository.getMatch(getRuntimeTempQueueNamespace(queueConfiguration.isTemporary()) + queueConfiguration.getAddress().toString());
+      QueueConfigurationUtils.applyDynamicQueueDefaults(queueConfiguration, addressSettings);
+
+      if (ActiveMQServerLogger.LOGGER.isEnabled(Logger.Level.WARN)) {
```
clebertsuconic, line 3519: I don't think you need to check for warn enabled... it shouldn't be a big burden anyways.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -3512,7 +3513,14 @@ public Queue createQueue(final QueueConfiguration queueConfiguration, boolean ig
...
+         if (new AddressImpl(queueConfiguration.getAddress(), configuration.getWildcardConfiguration()).containsWildCard() && addressSettings.getPageStoreName() == null) {
```
clebertsuconic, line 3520: isn't there a better way instead of new AddressImpl() ?


that's the only concern I have.

gtully: I share that concern, i did not find a way to get the AddressImpl back from the binding created by the wildcard address manager. It needs some surgery to access.

The addressImpl has a reference to the wildcard config, I guess I could duplicate that logic in place of the new ..

clebertsuconic: @gtully I've sent a Pull Request towards your personal fork:

https://github.com/gtully/activemq-artemis/pull/1


if you can squash it with your commit.  I was looking through the code it was easier to just write the change instead of explaining it.

clebertsuconic: actually, I will merge with my change squashed...

you can follow up with additional commits if you don't agree with anything.


thanks a lot for this PR.. very nice!

gtully: Thank you, that is a nice improvement 😷

PR #3278, opened 2020-09-25
ARTEMIS-2908 - Persist Divert Configuration in Bindings journal

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-2908

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/BufferHelper.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/BufferHelper.java
@@ -18,6 +18,7 @@ package org.apache.activemq.artemis.utils;
 
 import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
 import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.logs.ActiveMQUtilBundle;
 
 /**
  * Helper methods to read and write from ActiveMQBuffer.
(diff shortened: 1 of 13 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/BufferHelper.java
```diff
@@ -16,8 +16,10 @@
  */
 package org.apache.activemq.artemis.utils;
 
+import io.netty.buffer.ByteBufUtil;
 import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
```
brusdev, line 20: This import doesn't look used.

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/BufferHelper.java
```diff
@@ -155,5 +157,38 @@ public static Double readNullableDouble(ActiveMQBuffer buffer) {
...
    }
 
+   public static int sizeOfNullableString(String s) {
+      if (s == null) {
+         return DataConstants.SIZE_BOOLEAN;
+      }
+      return DataConstants.SIZE_BOOLEAN + sizeOfString(s);
+   }
+
+   public static int sizeOfeString(String s) {
```
franz1981, line 167: This one doesn't look correct

franz1981 (changes requested):

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/BufferHelper.java
andytaylor: fixed

PR #3279, opened 2020-09-27
ARTEMIS-2919 support timestamping incoming messages

jbertram:


artemis-tools/src/test/resources/artemis-configuration.xsd
```diff
@@ -3456,6 +3456,14 @@
                </xsd:annotation>
             </xsd:element>
 
+            <xsd:element name="enable-ingress-timestamp" type="xsd:boolean" default="true" maxOccurs="1" minOccurs="0">
```
gemmellr, line 3459: Similar to default value comment for earlier xsd file (this one looks test-only so may not matter though)

docs/user-manual/en/address-model.md
```diff
@@ -996,3 +997,14 @@ queues created on the matching address. Defaults to 0. Read more about
...
+
+`enable-ingress-timestamp` determines whether or not the broker will add its
+timestamp to messages sent to the matching address. When `true` the broker will
+treat AMQP and non-AMQP messages differently. For AMQP messages the broker will
+add a *message annotation* named `x-opt-ingress-timestamp`. For core messages
+(used by the core and OpenWire protocols) the broker will add a property named
+`_AMQ_INGRESS_TIMESTAMP`. For STOMP messages the broker will add a frame header
+named `ingress-timestamp`. The value will be a 64-bit
+[epoch](https://en.wikipedia.org/wiki/Unix_time) timestamp with
+[*nano*second](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--)
```
gemmellr, line 1009: Per earlier comment, this doesnt match what the value sent does, as System.nanoTime() isnt epoch based.

docs/user-manual/en/address-model.md
```diff
@@ -996,3 +997,14 @@ queues created on the matching address. Defaults to 0. Read more about
 `enable-metrics` determines whether or not metrics will be published to any
 configured metrics plugin for the matching address. Default is `true`. Read more
 about [metrics](metrics.md).
+
+`enable-ingress-timestamp` determines whether or not the broker will add its
+timestamp to messages sent to the matching address. When `true` the broker will
+treat AMQP and non-AMQP messages differently. For AMQP messages the broker will
+add a *message annotation* named `x-opt-ingress-timestamp`. For core messages
```
gemmellr, line 1004: Should perhapssay it will be a "long" typed annotation.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpSendReceiveTest.java
```diff
@@ -1260,5 +1263,33 @@ public void testReceiveRejecting() throws Exception {
...
+      Queue queueView = getProxyToQueue(getQueueName());
+      assertEquals(1, queueView.getMessageCount());
+
+      receiver.flow(1);
+      AmqpMessage receive = receiver.receive(5, TimeUnit.SECONDS);
+      assertNotNull(receive);
+      instanceLog.info(receive);
+      Object ingressTimestamp = receive.getMessageAnnotation(AMQPMessageSupport.INGRESS_TIMESTAMP.toString());
+      assertNotNull(ingressTimestamp);
+      assertTrue(ingressTimestamp instanceof Date);
```
gemmellr, line 1286: This test fails, as it is checking for the wrong type. It should check for a long.

It should also inspect the value and verify it is set within some delta (e.g <5sec since just before sending) so as to ensure the value is set properly.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/IngressTimestampTest.java
```diff
@@ -0,0 +1,150 @@
...
+
+   @Test
+   public void testSendOpenWireReceiveOpenWire() throws Throwable {
+      internalSend(Protocol.OPENWIRE, Protocol.OPENWIRE);
+   }
+
+   /**
+    *  we don't test AMQP here because there's no way to access the message annotations from the Qpid JMS client
+    *
+   @Test
```
gemmellr, line 88: Rather than the commented out tests, why not add to the test class in the amqp package.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
```diff
@@ -636,6 +641,15 @@ default Object getBrokerProperty(SimpleString key) {
       return getObjectProperty(key);
    }
 
+   default Message setIngressTimestamp() {
+      setBrokerProperty(HDR_INGRESS_TIMESTAMP, System.nanoTime());
+      return this;
+   }
+
+   default Object getIngressTimestamp() {
```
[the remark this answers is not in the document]
jbertram: `Object` is not needed. That was leftover from when I was using a `java.util.Date` for AMQP and a `long` everywhere else.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1317,6 +1317,17 @@ public Object getBrokerProperty(SimpleString key) {
       return extra.getProperty(key);
    }
 
+   @Override
+   public final org.apache.activemq.artemis.api.core.Message setIngressTimestamp() {
+      setMessageAnnotation(AMQPMessageSupport.INGRESS_TIMESTAMP, System.nanoTime());
```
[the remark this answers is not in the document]
jbertram: Thanks for the feedback. I misunderstood the semantics of `nanoTime()`.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -3695,6 +3695,14 @@
                   </xsd:documentation>
                </xsd:annotation>
             </xsd:element>
+
+            <xsd:element name="enable-ingress-timestamp" type="xsd:boolean" default="true" maxOccurs="1" minOccurs="0">
```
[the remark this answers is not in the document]
jbertram: Typo. Will fix.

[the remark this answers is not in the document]
clebertsuconic: I hope this is false by default, as something that's always re-encoding messages should be off by default.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/IngressTimestampTest.java
jbertram: Will do.

jbertram:
@gemmellr, I think I've resolved all the issues you enumerated. Please review and let me know if there's anything left. Thanks for the feedback. It's always appreciated.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/IngressTimestampTest.java
gemmellr: Rather than leave the commented out code  + comment, I'd instead suggest adding a reference to the $PROTOCOL-->AMQP tests added in the AMQP package. Or just remove the comment entirely.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1317,6 +1317,21 @@ public Object getBrokerProperty(SimpleString key) {
       return extra.getProperty(key);
    }
 
+   @Override
+   public final org.apache.activemq.artemis.api.core.Message setIngressTimestamp() {
+      setMessageAnnotation(AMQPMessageSupport.INGRESS_TIMESTAMP, System.currentTimeMillis());
```
gemmellr, line 1322: I see the impl changed to supplying epoch based millisecond value.

If this is going to be the case then I'd say it shouldnt use the same x-opt-ingress-timestamp annotation key as the qpidd broker the feature inspiration came from, since the two would then be using the same key for two differnet styles of value. That would rather break interop between applications and/or the two brokers (at least without significant hoop jumping), rather defeating the point of using the same protocol and annotations.

jbertram:
@gemmellr, I updated the PR to use `x-opt-ingress-timestamp-millis` for AMQP. Let me know if you have additional feedback.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
```diff
@@ -161,6 +161,11 @@
     */
    SimpleString HDR_ROUTING_TYPE = new SimpleString("_AMQ_ROUTING_TYPE");
 
+   /**
+    * The Routing Type for this message.  Ensures that this message is only routed to queues with matching routing type.
```
brusdev, line 165: The comment is relative to HDR_ROUTING_TYPE: `The Routing Type ...`

jbertram: Thanks, @brusdev. I just fixed it.

docs/user-manual/en/address-model.md
```diff
@@ -964,3 +965,13 @@ queues created on the matching address. Defaults to 0. Read more about
 `enable-metrics` determines whether or not metrics will be published to any
 configured metrics plugin for the matching address. Default is `true`. Read more
 about [metrics](metrics.md).
+
+`enable-ingress-timestamp` determines whether or not the broker will add its
+timestamp to messages sent to the matching address. When `true` the exact behavior
+will depend on the specific protocol in use. For AMQP messages the broker will
+add a `long` *message annotation* named `x-opt-ingress-timestamp-millis`. For core
+messages (used by the core and OpenWire protocols) the broker will add a property
```
gemmellr, line 973: Perhaps "`long` property" instead of just "property" ?

docs/user-manual/en/address-model.md
```diff
@@ -964,3 +965,13 @@ queues created on the matching address. Defaults to 0. Read more about
...
+named `_AMQ_INGRESS_TIMESTAMP`. For STOMP messages the broker will add a frame
+header named `ingress-timestamp`. The value will be a 64-bit
+[epoch](https://en.wikipedia.org/wiki/Unix_time) timestamp with millisecond
+precision. Default is `false`.
```
gemmellr, line 977: I'd suggest "The value will be the number of milliseconds since the epoch" in order to be clearer on the units (vs precision, which is different).

(Unix time is in units of seconds, and though currentTimeMillis is in units of milliseconds, it is not guaranteed as being millisecond precision)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpClientTestSupport.java
```diff
@@ -218,7 +230,7 @@ protected TransportConfiguration addAcceptorConfiguration(ActiveMQServer server,
    }
 
    protected String getConfiguredProtocols() {
-      return "AMQP,OPENWIRE";
+      return "AMQP,OPENWIRE,CORE";
```
gemmellr, line 233: This seems like the wrong place for this change. If a specific test wants to enable CORE it can override the method, thats what it exists for. (Same applies for the OPENWIRE one already there, but another time..)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpSendReceiveTest.java
```diff
@@ -1260,5 +1262,60 @@ public void testReceiveRejecting() throws Exception {
...
+
+      assertEquals(1, queueView.getMessageCount());
+
+      connection.close();
+   }
+
+   private enum Protocol {
+      CORE, AMQP, OPENWIRE
+   }
+
```
gemmellr, line 1319: I'd probably give these their own test class, its testing a specific broker feature rather than 'general AMQP sending and receiving' that this class is suited to, and its got its own class in the other module. That would also help with its need for enabling the CORE protocol being specific to these tests without having to do so for all the others here.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpSendReceiveTest.java
```diff
@@ -1260,5 +1262,60 @@ public void testReceiveRejecting() throws Exception {
...
+      receiver.flow(1);
+      AmqpMessage receive = receiver.receive(5, TimeUnit.SECONDS);
+      assertNotNull(receive);
+      instanceLog.info(receive);
+      Object ingressTimestamp = receive.getMessageAnnotation(AMQPMessageSupport.INGRESS_TIMESTAMP.toString());
+      assertNotNull(ingressTimestamp);
+      assertTrue(ingressTimestamp instanceof Long);
+      long ingressTs = (Long) ingressTimestamp;
+      assertTrue(beforeSend <= ingressTs && afterSend >= ingressTs);
```
gemmellr, line 1308: Getting a more useful error should it ever fail would be probably be good. E.g a Hamcrest based assertion might be something like:

```
Matcher<Long> inRange = both(greaterThanOrEqualTo(beforeSend)).and(lessThanOrEqualTo(afterSend));
MatcherAssert.assertThat("Ingress time should be set in expected range", ingressTs, inRange);
```

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/StompTest.java
```diff
@@ -709,6 +709,24 @@ public void testSubscribeWithAutoAck() throws Exception {
...
+      conn.connect(defUser, defPass);
+
+      subscribe(conn, null, Stomp.Headers.Subscribe.AckModeValues.AUTO);
+
+      sendJmsMessage(getName());
+
+      ClientStompFrame frame = conn.receiveFrame(10000);
+      Assert.assertEquals(Stomp.Responses.MESSAGE, frame.getCommand());
+      Assert.assertEquals(getQueuePrefix() + getQueueName(), frame.getHeader(Stomp.Headers.Send.DESTINATION));
+      Assert.assertNotNull(frame.getHeader(Stomp.Headers.Message.INGRESS_TIMESTAMP));
```
gemmellr, line 724: Same as other protocols, validating the value is as actually set as expected would be good.

gemmellr (commented):
Is the value expected to be persisted with the message through broker restart? If so a test for that would seem useful.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
```diff
@@ -636,6 +641,19 @@ default Object getBrokerProperty(SimpleString key) {
...
+   default Message setIngressTimestamp() {
+      setBrokerProperty(HDR_INGRESS_TIMESTAMP, System.currentTimeMillis());
+      return this;
+   }
+
+   default Long getIngressTimestamp() {
+      Object result = getBrokerProperty(HDR_INGRESS_TIMESTAMP);
+      if (result == null) {
+         return null;
+      }
```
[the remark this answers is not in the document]
jbertram: For some reason I had in my head that trying to cast a `null` would throw a `NullPointerException`. Will fix.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1339,6 +1339,21 @@ public Object getBrokerProperty(SimpleString key) {
...
+      return this;
+   }
+
+   @Override
+   public Long getIngressTimestamp() {
+      Object result = getMessageAnnotation(AMQPMessageSupport.INGRESS_TIMESTAMP);
+      if (result == null) {
+         return null;
+      }
```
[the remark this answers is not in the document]
jbertram: For some reason I had in my head that trying to cast a `null` would throw a `NullPointerException`. Will fix.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
```diff
@@ -99,6 +99,11 @@
     */
    public static final Symbol ROUTING_TYPE = Symbol.getSymbol("x-opt-routing-type");
 
+   /**
+    * Attribute used to mark the Broker-defined ingress time assigned to the message
+    */
+   public static final Symbol INGRESS_TIMESTAMP = Symbol.getSymbol("x-opt-ingress-timestamp-millis");
```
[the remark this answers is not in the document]
jbertram: I like `x-opt-ingress-time`. Kudos.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -283,6 +284,8 @@ protected static ServerJMSMessage processMessageAnnotations(ServerJMSMessage jms
                if (delay > 0) {
                   jms.setLongProperty(HDR_SCHEDULED_DELIVERY_TIME.toString(), System.currentTimeMillis() + delay);
                }
+            } else if (AMQPMessageSupport.INGRESS_TIMESTAMP.toString().equals(key) && entry.getValue() != null) {
+               jms.setLongProperty(HDR_INGRESS_TIMESTAMP.toString(), ((Number) entry.getValue()).longValue());
```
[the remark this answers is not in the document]
jbertram: Fair enough.

docs/user-manual/en/address-model.md
jbertram: Will do.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpClientTestSupport.java
clebertsuconic: I agree with Robbie here.

jbertram: Will fix.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpSendReceiveTest.java
jbertram: Good point.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/IngressTimestampTest.java
```diff
@@ -0,0 +1,150 @@
...
+      try (Connection connection = factoryConsume.createConnection()) {
+         connection.start();
+         try (Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE)) {
+            javax.jms.Queue queue = session.createQueue(QUEUE.toString());
+            try (MessageConsumer consumer = session.createConsumer(queue)) {
+               TextMessage message = (TextMessage) consumer.receive(1000);
+               Assert.assertNotNull(message);
+               Assert.assertEquals(0, message.getIntProperty("mycount"));
+               Assert.assertEquals("hello", message.getText());
+               assertNotNull(message.getObjectProperty(Message.HDR_INGRESS_TIMESTAMP.toString()));
```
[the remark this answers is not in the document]
jbertram: Will do.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpSendReceiveTest.java
jbertram: Will do.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/StompTest.java
jbertram: Will do.

docs/user-manual/en/address-model.md
jbertram: Will do.

jbertram:
I added a broker restart to a couple of the tests. The STOMP test didn't need it as that should be covered by the other test using core messages.

jbertram:
@gemmellr, any additional feedback here?

franz1981:
@jbertram i am quite interested into this bud, please hold on a bit, I just need some days to try it first, because I have recently some personal/family issues

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
```diff
@@ -198,6 +200,8 @@
    public static final Symbol JMS_DEST_TYPE_MSG_ANNOTATION = getSymbol("x-opt-jms-dest");
    public static final Symbol JMS_REPLY_TO_TYPE_MSG_ANNOTATION = getSymbol("x-opt-jms-reply-to");
 
+   public static final Symbol INGRESS_TIME = getSymbol(HDR_INGRESS_TIME);
```
gemmellr, line 203: Might be good to group and align the name with the other _MSG_ANNOTATION constants above?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -283,6 +284,8 @@ protected static ServerJMSMessage processMessageAnnotations(ServerJMSMessage jms
                if (delay > 0) {
                   jms.setLongProperty(HDR_SCHEDULED_DELIVERY_TIME.toString(), System.currentTimeMillis() + delay);
                }
+            } else if (AMQPMessageSupport.HDR_INGRESS_TIME.equals(key) && entry.getValue() != null) {
+               jms.setLongProperty(HDR_INGRESS_TIME.toString(), ((Number) entry.getValue()).longValue());
```
gemmellr, line 288: Might be good to rename AMQPMessageSupport.HDR_INGRESS_TIME to avoid the qualification and confusing it can cause, and be more consistent with the symbol name created from it?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/SimpleTest.java
```diff
@@ -83,6 +83,7 @@ public void setUp() throws Exception {
 
    @Test
    public void simpleTest() throws Exception {
+      System.out.println((String) null);
```
gemmellr, line 86: Leftover debug?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpIngressTimeTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+      receiver.flow(1);
+      AmqpMessage receive = receiver.receive(5, TimeUnit.SECONDS);
+      assertNotNull(receive);
+      instanceLog.info(receive);
+      Object ingressTimestamp = receive.getMessageAnnotation(AMQPMessageSupport.HDR_INGRESS_TIME);
+      assertNotNull(ingressTimestamp);
+      assertTrue(ingressTimestamp instanceof Long);
+      long ingressTs = (Long) ingressTimestamp;
+      assertTrue("Ingress time should be set in expected range",beforeSend <= ingressTs && afterSend >= ingressTs);
```
gemmellr, line 82: Same point as last time about the unhelpful failure this will produce

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/IngressTimeTest.java
```diff
@@ -0,0 +1,145 @@
...
+         try (Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE)) {
+            javax.jms.Queue queue = session.createQueue(QUEUE.toString());
+            try (MessageConsumer consumer = session.createConsumer(queue)) {
+               TextMessage message = (TextMessage) consumer.receive(1000);
+               Assert.assertNotNull(message);
+               Object ingressTime = message.getObjectProperty(Message.HDR_INGRESS_TIME.toString());
+               assertNotNull(ingressTime);
+               assertTrue(ingressTime instanceof Long);
+               long ingress = (Long) ingressTime;
+               assertTrue("Ingress time should be set in expected range",beforeSend <= ingress && afterSend >= ingress);
```
gemmellr, line 124: Same point as last time about the unhelpful failure this will produce

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/StompTest.java
```diff
@@ -709,6 +709,27 @@ public void testSubscribeWithAutoAck() throws Exception {
...
+      sendJmsMessage(getName());
+      long after = System.currentTimeMillis();
+
+      ClientStompFrame frame = conn.receiveFrame(10000);
+      Assert.assertEquals(Stomp.Responses.MESSAGE, frame.getCommand());
+      Assert.assertEquals(getQueuePrefix() + getQueueName(), frame.getHeader(Stomp.Headers.Send.DESTINATION));
+      String ingressTimeHeaderValue = frame.getHeader(Stomp.Headers.Message.INGRESS_TIME);
+      Assert.assertNotNull(ingressTimeHeaderValue);
+      long ingressTime = Long.parseLong(ingressTimeHeaderValue);
+      assertTrue("Ingress time should be set in expected range", before <= ingressTime && after >= ingressTime);
```
gemmellr, line 728: Same point as last time about the unhelpful failure this will produce

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpIngressTimeTest.java
```diff
@@ -0,0 +1,98 @@
...
+      } else if (protocol == Protocol.OPENWIRE) {
+         sendMessagesOpenWire(getQueueName(), 1, true);
+      } else {
+         sendMessages(getQueueName(), 1, true);
+      }
+      long afterSend = System.currentTimeMillis();
+
+      server.stop();
+      server.start();
+      assertTrue(server.waitForActivation(3, TimeUnit.SECONDS));
```
gemmellr, line 63: Its good there is now a test that verifies it survives restart...but now it seems we dont have any that verifies it works without the restart except STOMP? With my earlier comment I was more thinking of at least a single added test that checked it survived restart, rather than swapping the existing ones.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
jbertram: I had it that way at first, but I was getting an "Illegal forward reference" error due to the location of `HDR_INGRESS_TIME`. Moving `HDR_INGRESS_TIME` would also require moving `X_OPT_PREFIX` so it seemed simpler to just put `INGRESS_TIME` where it is.

gemmellr: I wasnt suggesting anything that should cause illegal forward references. I simply mean delete the empty line above it, and add _MSG_ANNOTATION to the name like the other message annotation constants that exist.

gemmellr: Ok I see now what you thought I was saying - I didt mean the internal property names ~25 lines up used to carry detail during protocol conversions, just the other specific message annotation symbols a line up.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpIngressTimeTest.java
jbertram: Are you talking about the assertions that have no error message or the specific message on the assertion which does?

clebertsuconic:
@jbertram I did not have time to look into this.. but if you could please make sure you test with large messages? (Make sure you reboot the server between you send and receive)?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpIngressTimeTest.java
gemmellr: Im not sure why GitHub is showing 4 lines as I only selected one at the time I commented, but its the last assertTrue with the message I mean, i.e "assertTrue("Ingress time should be set in expected range",beforeSend <= ingressTs && afterSend >= ingressTs);"

The above will basically give 'it failed here, good luck on your quest', basically requiring you run it again to try and figure out what happened (which could be racey and hard to reproduce, might cause us to modify the test to help figure the reason, etc), whereas a more useful failure would show what the value was and what it should have been [between]. That might even allow figuring out the issue without needing to run the test again (or if we cant get it to fail again) as the failure message was far more useful and potential causes potentially could be reasoned from it...e.g 'oh it was 0, weve stopped setting it, oops \<checks recent changes\>' or 'it was 1ms outside expectation, possibly we introduced some unexpected async behaviour'

E.g one option from one of the earlier comments: https://github.com/apache/activemq-artemis/pull/3279#discussion_r533325701
```
Matcher<Long> inRange = both(greaterThanOrEqualTo(beforeSend)).and(lessThanOrEqualTo(afterSend));
MatcherAssert.assertThat("Ingress time should be set in expected range", ingressTs, inRange);
```

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
jbertram: Done.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/SimpleTest.java
jbertram: Fixed.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
jbertram: Fixed

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpIngressTimeTest.java
jbertram: Fixed.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/StompTest.java
jbertram: Fixed

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpIngressTimeTest.java
jbertram: Fixed.

jbertram:
@clebertsuconic, I finally got back around to this and there are indeed problems with large messages. Specifically, when sending an AMQP large message and then consuming it with core or OpenWire (see `org.apache.activemq.artemis.tests.integration.client.IngressTimestampTest`). It looks to me like the problem is that `org.apache.activemq.artemis.protocol.amqp.broker.AMQPLargeMessage#reencode` doesn't really do anything, but I'm not sure how to fix it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/IngressTimeTest.java
jbertram: Done.

jbertram:
@clebertsuconic, I've resolved the issues with large messages, and I ran the whole test-suite. There was one failure in `org.apache.activemq.artemis.tests.integration.cluster.distribution.ClusteredGroupingTest`, but that appears completely unrelated to my change, and I couldn't reproduce it locally. Please review and merge before the next release. Thanks!

PR #3280, opened 2020-09-28
ARTEMIS-2920 Fix ActiveMQ Artemis Features build on IBM JDK 1.8

brusdev:


artemis-features/pom.xml
```diff
@@ -30,6 +30,20 @@
     </properties>
 
     <dependencies>
+        <dependency>
+            <groupId>org.apache.karaf.features</groupId>
+            <artifactId>framework</artifactId>
+            <version>${karaf.version}</version>
+            <type>xml</type>
+            <classifier>features</classifier>
```
gemmellr, line 38: Should these new deps be 'provided' scope like the other existing dep was?

Also, since the JIRA says the build is failing while downloading these deps, that suggests it already knows about the deps being required, and the build passing on other JDKs suggests that process succeeds there. That makes me wonder if perhaps the fix needed actually isnt that they need added as deps? From the description it sounds like maybe Karaf failed to download them, so I guess this will just make Maven download them initially instead so Karaf then doesnt have to?

brusdev: @gemmellr the exception is only due to the verify goal failure of karaf-maven-plugin. It can't download the required features XML to verify and validate artemis features. I'm not an expert of karaf but reading [karaf maven dependencies](https://svn.apache.org/repos/asf/karaf/site/production/manual/latest/custom-distribution.html#_maven_dependencies) I think the compile scope should be fine.

gemmellr: So what I am saying is that since this works on other JDKs already, and even on the IBM JDK it clearly knows about the items for it to get as far as trying and failing to download them, that would suggest to me that it isnt a pom dependency issue at the core of the problem. I wouldnt expect that to really shouldnt change between e.g different JDK8 instances. Other stuff like TLS handling etc is known to vary though, and might be more likely to cause issues; is there more detail around the download failure errors?

It just seems like by doing this you would perhaps (I'm no karaf expert either) instead be avoiding the actual issue by adding the bits as dependencies of the pom and having Maven download them into the local repo in advance during the build so Karaf doesnt need to, rather than fixing whatever is preventing Karaf from downloading them when it needs them (which I am presuming is how it is succeeding in doing the build on other JDKs for this to work there currently?).

If it is done it seems odd to have one dep at provided scope and the others at compile. That would mean there was a fixed-version dep on particular karaf bits in the pom, which may not be the karaf verison something using this uses, which using provided scope would presumably avoid (I'd guess thats why the existing dep uses provided).

brusdev: I have just added the full error stack trace to the related JIRA: https://issues.apache.org/jira/browse/ARTEMIS-2920
I have deepen the karaf-maven-plugin configuration. The maven dependencies scope for features and kar archives should affect only the `assembly` goal.

Si it should be indifferent from the karaf-maven-plugin POV setting the scope to `compile` or `provided` for all 3 dependencies  because we are not using the `assembly` goal.

The new dependencies are required just as workaround for the `verify` goal of karaf-maven-plugin on IBM JDK.

gemmellr: If making this change of adding deps is the simplest albeit hacky workaround then thats fair enough, but I still think they should be at provided scope like the existing dep, so that anything referencing the pom in turn wont pick the unecessary deps up. To be clear, its the maven sense of things rather than the karaf sense of things I am considering here. Having the deps at compile scope would would mean anything referencing this pom will pick them up and end up downloading those dependencies even if they arent strictly needed (otherwise it wouldnt work anywhere currently and this change wouldnt be effective). If there isnt a reason using provided wont work, then that seems like what should be used.

--

Seperately though I would also say the expanded error just confirms what I was thinking. The download failure occurring while the karaf plugin attempts the download is due to a TLS protocol version issue, likely it doesnt have the required TLS version enabled/supported and so it fails. This change will avoid that causing maven itself to download the artifacts earlier as dependencies of the pom (and make them available when karaf goes looking locally), and presumably due to using a different stack for the process it somehow does have the required TLS version supported.

I'd guess this is because some code involved is using a TLS protocol lookup of "TLS", which in non-IBM JDKs tends to allow all protocol versions but in the IBM JDK doing so defaults to allowing only 1.0, support for which has been dropped from the maven repositories. Searching suggests various sys props used to avoid issues such as this on the IBM JDK (or JDK7 in cases), such as:
https.protocols=TLSv1.2

com.ibm.jsse2.overrideDefaultProtocol=TLSv12
com.ibm.jsse2.overrideDefaultTLS=true

The first is something I actually used recently in an unrelated ASF project CI job failing for similar reasons on JDK7, using "mvn <foo> -Dhttps.protocols=TLSv1.2" to enable TLS 1.2 and resolve the issue as JDK 7 supported but didnt enable it by default for compatibility.

The last one is something that looks to be used during the Artemis tests automatically due to Surefire configuration in the parent pom:
https://github.com/apache/activemq-artemis/blob/2.15.0/pom.xml#L896

In this case, I dont know that the karaf plugin has a way to configure such things specifically, so I expect it would be needed to pass one or more of them when executing maven for it to be effective.

(I also just noticed there are non-https repositories in your custom maven configuration, and so it is presumably being redirected to the https, you probably want to resolve that; who knows it may even be related)

brusdev: @gemmellr thanks now I see your POV, I have just updated the dependencies scope.

PR #3281, opened 2020-09-29
 ARTEMIS-1884 add plugin API for message level authorization policies

ryeats:


gtully:
this looks nice but it seem too many hoops to go through to have a plugin implement a filter and a very specific new plugin type. A generic filter may be more generally applicable. canAccept behaving like filter.match and gating delivery.

maybe the existing broker plugin can have a new: 
 - boolean canAccept(ServerConsumer, MessageReference)

and with the new getSubject accessor that can implement authorization.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -2444,6 +2446,25 @@ public void callBrokerMessagePlugins(final ActiveMQPluginRunnable<ActiveMQServer
       callBrokerPlugins(getBrokerMessagePlugins(), pluginRun);
    }
 
+   @Override
+   public boolean callBrokerMessagePluginsCanAccept(ServerConsumer serverConsumer, MessageReference messageReference) throws ActiveMQException {
+      for (ActiveMQServerMessagePlugin plugin : getBrokerMessagePlugins()) {
```
michaelandrepearce, line 2451: Default return should surely be false.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -403,6 +404,13 @@ public HandleStatus handle(final MessageReference ref) throws Exception {
          return HandleStatus.BUSY;
       }
 
+      if (!server.callBrokerMessagePluginsCanAccept(this, ref)) {
```
michaelandrepearce, line 407: This is hot path. Needs to be a simple bool check if it should go into this code area e.g. hasBrokerMessagePlugins which shoild be a boolen flag behind it not a list

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -1547,4 +1555,9 @@ public String getConnectionLocalAddress() {
    public String getConnectionRemoteAddress() {
       return this.session.getRemotingConnection().getTransportConnection().getRemoteAddress();
    }
+
+   @Override
+   public Subject getSubject() {
```
michaelandrepearce, line 1560: Dont expose tbis

michaelandrepearce: if something needs it like a plugun the server can be passed jn at injt stage or just pass server

artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
```diff
@@ -339,6 +339,22 @@ public static String getUserFromSubject(Subject subject) {
...
+    * Get the cached Subject. If the Subject is not in the cache then authenticate again to retrieve
+    * it.
+    *
+    * @param session contains the authentication data
+    * @return the authenticated Subject with all associated role principals or null if not
+    * authenticated or JAAS is not supported by the SecurityManager.
+    */
+   @Override
+   public Subject getSessionSubject(SecurityAuth session) {
+      if (securityManager instanceof ActiveMQSecurityManager5) {
```
michaelandrepearce, line 352: This is a little specific, why only 5

clebertsuconic:
@ryeats for big features like this, it would be nice to also have a DISCUSS thread on the dev list explaining how things works, some designs.. etc...


I'm not trying to pedantic or anything.. just saying it would be nice.. (like if you look at the dev-list now, you will see a thread I started myself for a PR I am going to make next week).

ryeats:
Yeah this turned into a bigger change than i was thinking i am requesting permission to post to the dev forums now.

jbertram:
@ryeats, I think this looks good for the most part, but it's a little hard to review with the commits separated out. Can you squash all the commits together?

ryeats:
Hacked at the Perf test to try out a few cases. Turns out Roles are stored in the javax.security.auth.Subject as a "Set" thats really backed by linked list so they do effect performance which is disappointing. I tried to run the test in a way that made the worst case take place every time so canAccept returns true on the 100th role on the 100th consumer.

**Performance Test on a Queue**:
Baseline performance
4743.47 msg/s (250000 messages in 52.70s
Role\Consumers | 1|11|101
------------ | ------------- | ------------- | ------------- 
2  |56.10s|56.64s|66.37s
11 |57.53s|58.96s|68.48s
101|56.23s|60.53s|84.66s

I also tested against a topic worst case 50 consumers have the role and it is their 100th role. Had some trouble hitting memory or other system limits on this one so adjusted the consumers down to avoid that noise but it still wasn't super consistent.
**Performance Test on a Topic with durability off**:
Baseline Performance
1 consumers 83084.08 msg/s (250000 messages in 3.01s)
10 consumers  21347.45 msg/s (250000 messages in 11.71s)
50 consumers 21396.78 msg/s (250000 messages in 11.68s)

**Performance Test on a Topic**:
Role\Consumers | 1|11|51
------------ | ------------- | ------------- | ------------- 
2|3.84s|8.16s|13.15s
11|2.49s|12.20s|18.93s
101|3.08s|11.58s|25.01s

The numbers were not super consistent but do seem to show a trend towards a tradeoff with more consumers and roles. I think there has to be some expectation of a trade off in performance when doing message level authentication though. It is disappointing that java's role principles are not backed by a hash lookup but the number of roles is normally small.

jbertram (commented):
You a need "negative" queue test in `tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/MessageAuthorizationTest.java` (i.e. a test where the consumers *can't* receive the messages due to the incorrect role).

examples/features/standard/broker-msg-auth-plugin/readme.md
```diff
@@ -0,0 +1,5 @@
+# Broker Plugin Example
+
+To run the example, simply type **mvn verify** from this directory, or **mvn -PnoServer verify** if you want to start and create the broker manually.
+
+This example shows how a message plugin can be used to filter message sent to a consumer depending on that consumers roles. Credentials for a user are by default invalidated every 10 minutes so this plugin may cause excessive authentication if used without configuring the security-invalidation-interval limit appropriately.
```
jbertram, line 5: The default `security-invalidation-interval` is actually 10000 milliseconds (i.e. 10 seconds), not 10 minutes.

ryeats:
@jbertram don't know a better way to touch base but wanted to checkin and see if there was anything else holding this PR up and if so what are the next steps are to push it along?

KalCramer (approved):

jbertram:
@ryeats, sorry it took me so long to get back to this. Everything looks good. I'll merge it ASAP.

jbertram:
@ryeats, nice work, BTW!

PR #3282, opened 2020-09-30
ARTEMIS-2859 - track owning page store as a transient message attribute to ensure correct usage tracking

gtully:
… and forward queue in error

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -1201,7 +1201,7 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
    @Override
    public MessageReference reload(final Message message, final Queue queue, final Transaction tx) throws Exception {
 
-      MessageReference reference = MessageReference.Factory.createReference(message, queue);
+      MessageReference reference = MessageReference.Factory.createReference(message, queue, pagingManager.getPageStore(message.getAddressSimpleString()));
 
       Long scheduledDeliveryTime;
       if (message.hasScheduledDeliveryTime()) {
@@ -1211,6 +1211,7 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
          }
       }
 
+      queue.refUp(reference);
       queue.durableUp(message);
 
       if (tx == null) {
@@ -1455,8 +1456,14 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
          deliveryTime = message.getScheduledDeliveryTime();
       }
 
+      PagingStore owningStore = pagingManager.getPageStore(message.getAddressSimpleString());
       for (Map.Entry<SimpleString, RouteContextList> entry : context.getContexListing().entrySet()) {
-         PagingStore store = pagingManager.getPageStore(entry.getKey());
+         PagingStore store;
+         if (entry.getKey() == message.getAddressSimpleString() || entry.getKey().equals(message.getAddressSimpleString())) {
+            store = owningStore;
+         } else {
+            store = pagingManager.getPageStore(entry.getKey());
+         }
 
          if (store != null && storageManager.addToPage(store, message, context.getTransaction(), entry.getValue())) {
             if (message.isLargeMessage()) {
@@ -1469,14 +1476,14 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
          }
 
          for (Queue queue : entry.getValue().getNonDurableQueues()) {
-            MessageReference reference = MessageReference.Factory.createReference(message, queue);
+            MessageReference reference = MessageReference.Factory.createReference(message, queue, owningStore);
 
             if (deliveryTime != null) {
                reference.setScheduledDeliveryTime(deliveryTime);
             }
             refs.add(reference);
 
-            queue.refUp(message);
+            queue.refUp(reference);
          }
 
          Iterator<Queue> iter = entry.getValue().getDurableQueues().iterator();
@@ -1484,7 +1491,7 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
          while (iter.hasNext()) {
             Queue queue = iter.next();
 
-            MessageReference reference = MessageReference.Factory.createReference(message, queue);
+            MessageReference reference = MessageReference.Factory.createReference(message, queue, owningStore);
 
             if (context.isAlreadyAcked(context.getAddress(message), queue)) {
                reference.setAlreadyAcked();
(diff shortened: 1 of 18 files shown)


gtully:
This seems quite a big change but I think it is the root cause. I see that routing to the remote binding can attempt to page the message to the cluster connection queue store, which competes with the local queue binding paging to the target address, the topic address. This seems wrong.
Having the remote binding just add the store and forward queue to the topic binding, rather than an entry in the routing context bindings map  for the store and forward address, avoids the contention. It seems wrong that it would page to the store and forward queue address!

Lets hope some existing test can show me the error if this logic is flawed. The new test that verifies the non negative size of the page store for the topic works as expected with this fix.

clebertsuconic:
@gtully I saw this already .. and LGTM... I'm just waiting on a full test suite run.

I share the same CI private box with you and I know the test is running. I will post my review after the test is finished.

if full pass I will merge it straight away.

clebertsuconic:
@gtully test failure:


5.2 sec | 1
-- | --
org.apache.activemq.artemis.tests.integration.cluster.distribution.MessageRedistributionTest.testRedistributionWithPagingOnTarget

clebertsuconic:
The test is not valid.. if the source is paging, the message will be stored on the source page, and not in the SNF queue.

gtully:
@Thanks clebertsuconic - that is what I need to find. those tests in particular, I will need to investigate more, when the address actually pages the continue at: https://github.com/apache/activemq-artemis/blob/b7669691876db45e11c929ebece425d6166beb70/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java#L1468

means that the queue dispatch does not happen and I guess the page cursor will only deal with the store and forward queue. It may require a different type fix.

also, the startPaging approach is very deterministic from a test perspective, noted!

clebertsuconic:
we probably need some double addSize on that case to discount where it came from? I dunno.

gtully:
the routing context is lost when we page, just left to the cursor. but once there are multiple cursors, there is contention on who gets to ack first.
back to the drawing board in any event.

clebertsuconic:
lets keep this open until we figure out the proper fix.. at least the test is very nice.. but we have to figure to a better fix.

gtully:
this fix directly tackles the negative address size problem of incrementing/decrementing the correct page store, by recording the owner as a transient message attribute. When the reference count goes to 0, the message size is decremented from the incrementing store.

gtully:
What is odd with the current implementation is that a local queue and a remote/cluster queue can both reference a message, and the first to see it will track the message body size on its page store. If that store is acked first, it will still appear to hold messages, till the second reference is decremented. 
There is a case to be made for only tracking in memory body sizes on the global page size, and leaving all pageStores will just references. But that leaves a pageStore max size a multiple of references rather than bytes.

Tracking the full body size on each pageStore will force global paging to early. But that is another approach that could work.

In both cases, there would be no need to track the owner in the message b/c each reference is equivalent.

@clebertsuconic what do you think?

clebertsuconic:
this counting thing on paging needs some rework. my fault pretty much. 

I tried to reuse the counters for large message references and the transaction, when I needed a separate counter.


there are some other factors... I will be working on this tomorrow.

clebertsuconic:
Here's what I think we have to do:

at least for paging, have a generic counter for both persistent and non persistent.

When sending the first message always go to the sending address. the reference will go to the routing queue.

I will need to play with this tomorrow. As I am also having the same issue on a branch I"m working on.

gtully:
here is what I have now. 
each message reference holds a reference to the sending address pageStore, which ever queue gets to first increment does the size increase, which ever is the last decrements. the refUp is being used for page store size adjustments, the durable up just tracking a count. Hense every queue that has a reference, gets to do refUp and the durable count logic just handles persistence/deletion.
With this change, the sending address page store always tracks the message size, every other reference just a 64byte reference. This leads to a cluster connection store and forward queue only paging on references exceeding a threshold or what appears to be small 64byte messages.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -970,36 +970,37 @@ public long getID() {
...
    }
 
    @Override
-   public void refDown(Message message) {
-      int count = message.refDown();
+   public void refDown(MessageReference messageReference) {
+      int count = messageReference.getMessage().refDown();
+      if (count == 0) {
+         if (messageReference.getOwner() != null) {
+            messageReference.getOwner().addSize(-messageReference.getMessageMemoryEstimate());
```
clebertsuconic, line 999: I would keep an else here, and just use this if null?

Some unit tests may break without this?

you're running the tests.. lets see.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1455,6 +1456,7 @@ public void processRoute(final Message message,
          deliveryTime = message.getScheduledDeliveryTime();
       }
 
+      PagingStore owningStore = pagingManager.getPageStore(message.getAddressSimpleString());
       for (Map.Entry<SimpleString, RouteContextList> entry : context.getContexListing().entrySet()) {
          PagingStore store = pagingManager.getPageStore(entry.getKey());
```
clebertsuconic, line 1461: may I suggest a simple optimization?

this is one simple optimization:

```java
    PagingStore store;
    if (entry.getKey() == message.getAddressSimpleString || entry.getKey().equals(message.getAddressSimpleString()) {
          store = owningStore;
   } else {
       store = pagingManager.getPageStore(entry.getKey());
   }
```

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: it seems only notification/management addresses have a null page store, however the reference count need to be independent of any particular queue having a page store

clebertsuconic: I think for those it should just use the one from the queue where you are then.. I think it would make sense.

clebertsuconic: tests went well. just leave this one as is then...

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1211,6 +1211,7 @@ public MessageReference reload(final Message message, final Queue queue, final T
          }
       }
 
+      queue.refUp(reference);
```
gtully, line 1214: This causes org.apache.activemq.artemis.tests.integration.client.LargeMessageTest#testFilePersistenceBlockedPreACK to fail. 
the large message file does not get removed.
Without it, it seems to work but the ref count is -1 on delete, which seems wrong. it may be an ordering issue

clebertsuconic: did you see my comment.. I sent a fix on another PR.

PR #3284, opened 2020-10-01
NO-JIRA Fixing the paths for the Docker examples

chrisbier:


jbertram:
Can you squash your 2 commits down to just 1?

jbertram:
And also add "NO-JIRA" to the beginning of the commit message?

chrisbier:
Sure thing

artemis-docker/Dockerfile-centos
```diff
@@ -57,7 +57,7 @@ USER root
 
 RUN mkdir /var/lib/artemis-instance && chown -R artemis.artemis /var/lib/artemis-instance
 
-COPY ./docker/docker-run.sh /
+COPY ./artemis-docker/docker-run.sh /
```
brusdev, line 60: This path don't match the path created by the `prepare-docker.sh` script: https://github.com/apache/activemq-artemis/blob/2.15.0/artemis-docker/prepare-docker.sh#L54

chrisbier: When I ran it on my machine, I had to fix the path in order for the Dockerfile to find the file.
```Step 15/20 : COPY ./docker/docker-run.sh /
COPY failed: stat /var/lib/docker/tmp/docker-builder613788742/docker/docker-run.sh: no such file or directory
```

chrisbier: Ah. Sorry! My fault for not reading closely.

PR #3286, opened 2020-10-02
ARTEMIS-2928 blocking CallbackCache can be replaced with a JCTools lock-free queue

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-2928

### DIFF
--- a/artemis-journal/pom.xml
+++ b/artemis-journal/pom.xml
@@ -69,6 +69,10 @@
          <version>${project.version}</version>
       </dependency>
       <dependency>
+         <groupId>org.jctools</groupId>
+         <artifactId>jctools-core</artifactId>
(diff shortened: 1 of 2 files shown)


michaelandrepearce:
Whilst we expect this to have a perf improvement, have we got some stats on this?

franz1981:
@michaelandrepearce 
I haven't measured it yet (with a disk), but:
* The original code was reinventing the wheel with a suboptimal solution
* Using the new version should allow to remove the cache from the native project, given that is not necessary

I think that even with the same performance (nearly impossible) it worth to be replaced...

franz1981:
I've got a green run on the whole test suite, going to merge this :+1:

PR #3287, opened 2020-10-02
ARTEMIS-2926 Scheduled task executions are skipped randomly

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-2926

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
@@ -20,11 +20,12 @@ package org.apache.activemq.artemis.core.server;
 import java.security.AccessController;
 import java.security.PrivilegedAction;
 import java.util.concurrent.Executor;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.activemq.artemis.utils.ActiveMQThreadFactory;
 import org.jboss.logging.Logger;
@@ -41,15 +42,13 @@ public abstract class ActiveMQScheduledComponent implements ActiveMQComponent, R
    /** initialDelay < 0 would mean no initial delay, use the period instead */
    private long initialDelay;
    private long period;
-   private long millisecondsPeriod;
    private TimeUnit timeUnit;
    private final Executor executor;
-   private volatile ScheduledFuture future;
+   private volatile boolean isStarted;
+   private ScheduledFuture future;
    private final boolean onDemand;
-
-   long lastTime = 0;
-
-   private final AtomicInteger delayed = new AtomicInteger(0);
+   // The start/stop actions shouldn't interact concurrently with delay so it doesn't need to be volatile
+   private AtomicBoolean bookedForRunning;
 
    /**
     * It creates a scheduled component that can trigger {@link #run()} with a fixed {@code checkPeriod} on a configured {@code executor}.
@@ -73,6 +72,8 @@ public abstract class ActiveMQScheduledComponent implements ActiveMQComponent, R
       this.period = checkPeriod;
       this.timeUnit = timeUnit;
       this.onDemand = onDemand;
+      this.bookedForRunning = new AtomicBoolean(false);
+      this.isStarted = false;
    }
 
    /**
@@ -89,12 +90,7 @@ public abstract class ActiveMQScheduledComponent implements ActiveMQComponent, R
                                      long checkPeriod,
                                      TimeUnit timeUnit,
                                      boolean onDemand) {
-      this.executor = null;
-      this.scheduledExecutorService = scheduledExecutorService;
-      this.initialDelay = initialDelay;
-      this.period = checkPeriod;
-      this.timeUnit = timeUnit;
-      this.onDemand = onDemand;
+      this(scheduledExecutorService, null, initialDelay, checkPeriod, timeUnit, onDemand);
    }
 
    /**
@@ -150,11 +146,11 @@ public abstract class ActiveMQScheduledComponent implements ActiveMQComponent, R
 
    @Override
    public synchronized void start() {
-      if (future != null) {
+      if (isStarted) {
          // already started
          return;
       }
-
+      isStarted = true;
       if (scheduledExecutorService == null) {
          scheduledExecutorService = new ScheduledThreadPoolExecutor(1, getThreadFactory());
          startedOwnScheduler = true;
@@ -165,10 +161,9 @@ public abstract class ActiveMQScheduledComponent implements ActiveMQComponent, R
          return;
       }
 
-      this.millisecondsPeriod = timeUnit.convert(period, TimeUnit.MILLISECONDS);
-
       if (period >= 0) {
-         future = scheduledExecutorService.scheduleWithFixedDelay(runForScheduler, initialDelay >= 0 ? initialDelay : period, period, timeUnit);
+         final AtomicBoolean booked = this.bookedForRunning;
+         future = scheduledExecutorService.scheduleWithFixedDelay(() -> runForScheduler(booked), initialDelay >= 0 ? initialDelay : period, period, timeUnit);
       } else {
          logger.tracef("did not start scheduled executor on %s because period was configured as %d", this, period);
       }
@@ -188,15 +183,26 @@ public abstract class ActiveMQScheduledComponent implements ActiveMQComponent, R
 
    }
 
-   public void delay() {
-      int value = delayed.incrementAndGet();
-      if (value > 10) {
-         delayed.decrementAndGet();
-      } else {
-         // We only schedule up to 10 periods upfront.
-         // this is to avoid a window where a current one would be running and a next one is coming.
-         // in theory just 2 would be enough. I'm using 10 as a precaution here.
-         scheduledExecutorService.schedule(runForScheduler, Math.min(period, period * value), timeUnit);
+   /**
+    * A delay request can succeed only if:
+    * <ul>
+    *    <li>there is no other pending delay request
+    *    <li>there is no pending execution request
+    * </ul>
+    * <p>
+    * When a delay request succeed it schedule a new execution to happen in {@link #getPeriod()}.<br>
+    */
+   public boolean delay() {
+      final AtomicBoolean booked = this.bookedForRunning;
+      if (!booked.compareAndSet(false, true)) {
+         return false;
+      }
+      try {
+         scheduledExecutorService.schedule(() -> bookedRunForScheduler(booked), period, timeUnit);
+         return true;
+      } catch (RejectedExecutionException e) {
+         booked.set(false);
+         throw e;
       }
    }
 
@@ -261,7 +267,14 @@ public abstract class ActiveMQScheduledComponent implements ActiveMQComponent, R
    }
 
    @Override
-   public void stop() {
+   public synchronized void stop() {
+      if (!isStarted) {
+         return;
+      }
+      isStarted = false;
+      // Replace the existing one: a new periodic task or any new delay after stop
+      // won't interact with the previously running ones
+      this.bookedForRunning = new AtomicBoolean(false);
       if (future != null) {
          future.cancel(false);
          future = null;
@@ -275,8 +288,8 @@ public abstract class ActiveMQScheduledComponent implements ActiveMQComponent, R
    }
 
    @Override
-   public synchronized boolean isStarted() {
-      return future != null;
+   public boolean isStarted() {
+      return isStarted;
    }
 
    // this will restart the scheduled component upon changes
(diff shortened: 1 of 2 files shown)


gemmellr:
I think the changes seem ok, but I think perhaps the PR overlooks another simpler and more important behaviour that may be leading to the observed issue?

The given period at construction of the scheduled tasks is documented as "the delay between the termination of one execution and the start of the next". Thats unsurprisingly consistent with the behaviour of scheduledExecutorService.scheduleWithFixedDelay(), which is whats used by the 'not onDemand' instances of the scheduled tasks. However, the tasks dont actually run in the scheduledExecutorService thread if the additional executor is given during construction. If the second executor is given, the scheduled task is just offloaded by the scheduledExecutorService for execution on the provided executor and entirely forgotten about. That seems like it could be the core of the observed issue to me?

The above means there is no further tracking by the scheduler of when the task actually runs or how long the given task takes, meaning the periodic contract is somewhat lost at that point forward. Consider a situation:
1.  Say there is a backlog of existing (related or unrelated) things for the executor still to run, so a new 'scheduled offloaded' task may not run for a little while until that is processed. Or instead say that thread scheduling means the second executor doesnt immediately get to executing the task. Whatever the reason, something means there is a small delay, but eventually the task does run.
2. A second task instance comes along from the scheduledExecutorService at some point, very closely after the configured period since it isnt affected by actual execution of the task, which gets offloaded. Maybe now there isnt any or as much backlog on the second executor, or theres a better thread scheduling environment, and this second task may get run relatively quicker than the prior instance actualy did.
3. Due to the 'lastTime' tracking occuring within the task itself, on the second executor, this second task instance which was offloaded by the scheduledExecutorService at its precise period, will now be observed to have occurred within the configured period of the previous tasks 'lstTime' and so get skipped.
4. This means nothing happened, and wont until the scheduledExecutorService comes along after a 3rd period and offloads the task another time, by which point approx double the expected period has elapsed and the task actually executes. Rinse and repeat this process over and over.

If the second executor is provided, its actual execution + 'lastTime' period checks are essentially happening independently of the scheduling, and it seems like the scheduledExecutorService is trying to somewhat blindly throw tasks over a wall such that they land at the right time and actually get to run as opposed to skipping and waiting for next time.

franz1981:
@gemmellr 

A couple of observations on the changes on be3c193 on https://github.com/apache/activemq-artemis/blob/be3c193f73775b961f855e6d935c442e28a14875/artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java#L324-L331
I've chosen on purpose to allow a racing booking request (can be either from a `delay` or the next periodic task submission on `executor`) to happen before instead of on completion of `ActiveMQScheduledComponent.this.run()`:

- because of the expected behaviour of `delay`: 
eg `JDBCJournalImpl` submit some records and call `syncTimer.delay()`; it means that if a pending task hasn't yet hit`bookedForRunning.compareAndSet(true, false)`, the delay request will fail, leaving some records to not be picked up and processed later
- because of the uselessness (due to how a JVM work?) to detect a "slow" (ie which execution time is > period) pending task (see https://issues.apache.org/jira/browse/ARTEMIS-2926?focusedCommentId=17212908&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17212908): TLDR if `bookedForRunning.compareAndSet(true, false)` takes long time to be hit it doesn't mean that the task hasn't been prompty executed/is slow -> skipping the next period task just means we're slowing down rate!

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -265,6 +291,14 @@ public void stop() {
       if (future != null) {
          future.cancel(false);
          future = null;
+         final AtomicBoolean bookedForRunning = this.bookedForRunning;
+         if (bookedForRunning != null) {
+            // this is trying to fast-fail any running task or pending one:
+            bookedForRunning.compareAndSet(false, true);
```
gemmellr, line 297: This goes at odds with the prior behaviour, where no attempt was made to stop the tasks already handed to the executor or the final scheduling already being in the process of either handing off another to the executor or running it directly on the scheduledExecutorService. The latter is still true to an extent, the future.cancel(false) remains above, but this new bit will ensure it doesnt actually do anything unless it already passed the gate and starts the 'ActiveMQScheduledComponent.this.run();'.

Have you checked if there is there any impact of this behaviour change? If not, might it be worth omitting this bit, and just leaving the reassignment below for now?

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -287,35 +321,63 @@ private void restartIfNeeded() {
...
-      public void run() {
-         if (executor != null) {
-            executor.execute(runForExecutor);
-         } else {
-            runForExecutor.run();
+   }
+
+   private void runForScheduler(AtomicBoolean bookedForRunning) {
+      if (executor != null) {
+         if (bookedForRunning != null && !bookedForRunning.compareAndSet(false, true)) {
```
gemmellr, line 357: Both of the 'if executor != null' legs do this check first, I think it would be simpler to move it outside the if and do it once atthe start. The descriptions dont differ enough to justify both IMO, so its more rather than less complicated/confusing having both.

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -188,41 +202,53 @@ public ClassLoader run() {
...
-         // We only schedule up to 10 periods upfront.
-         // this is to avoid a window where a current one would be running and a next one is coming.
-         // in theory just 2 would be enough. I'm using 10 as a precaution here.
-         scheduledExecutorService.schedule(runForScheduler, Math.min(period, period * value), timeUnit);
+      final AtomicBoolean bookedForRunning = this.bookedForRunning;
+      if (bookedForRunning == null || bookedForRunning.compareAndSet(false, true)) {
+         try {
+            scheduledExecutorService.schedule(() -> delayedRunForScheduler(bookedForRunning), nsPeriod, TimeUnit.NANOSECONDS);
+         } catch (RejectedExecutionException e) {
+            if (bookedForRunning == null) {
```
gemmellr, line 221: Guessing that was meant to be != null? EDIT: as later commenting, think the null check should be eliminated through simplification.

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -188,41 +202,53 @@ public ClassLoader run() {
...
+               bookedForRunning.set(false);
+            }
+            throw e;
+         }
       }
    }
 
    public long getPeriod() {
-      return period;
+      return timeUnit.convert(nsPeriod, TimeUnit.NANOSECONDS);
```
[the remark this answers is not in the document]
franz1981: Good idea: I remember I've checked every possible caller of it, but...without tests we can easily get regressions and I could have missed something :+1:

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -60,41 +58,58 @@
...
+    * @param overBooking              if {@code true} {@link #delay()} executions and period ones are allowed to accumulate, {@code false} otherwise. Default is {@code false}.
     */
    public ActiveMQScheduledComponent(ScheduledExecutorService scheduledExecutorService,
                                      Executor executor,
                                      long initialDelay,
                                      long checkPeriod,
                                      TimeUnit timeUnit,
-                                     boolean onDemand) {
+                                     boolean onDemand,
+                                     boolean overBooking) {
```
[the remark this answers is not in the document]
franz1981: > I would remove the overbooking, eliminate the possibility of null for the atomic, and simplify the rest of the code accordingly. It will still be plenty complicated :)

+100 agree: I was thinking that was needed for the JDBC lease lock, but effectively isn't needed, I'm removing it

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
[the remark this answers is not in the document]
gemmellr: Actually, I see there now is a getTimeUnit, so I think this change needs effectively unwound...it should return the same value as passed in.

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
franz1981: > Have you checked if there is there any impact of this behaviour change? If not, might it be worth omitting this bit, and just leaving the reassignment below for now?

It's the safer choice here so will do as you suggest and simplify the rest of the code (although "simplify" is rather optimistic here XD )

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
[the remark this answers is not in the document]
franz1981: TBH given that we don't measure time anymore it's fine :)

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -188,41 +202,53 @@ public ClassLoader run() {
...
    public void delay() {
-      int value = delayed.incrementAndGet();
-      if (value > 10) {
-         delayed.decrementAndGet();
-      } else {
-         // We only schedule up to 10 periods upfront.
-         // this is to avoid a window where a current one would be running and a next one is coming.
-         // in theory just 2 would be enough. I'm using 10 as a precaution here.
-         scheduledExecutorService.schedule(runForScheduler, Math.min(period, period * value), timeUnit);
+      final AtomicBoolean bookedForRunning = this.bookedForRunning;
```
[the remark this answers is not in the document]
franz1981: Sorry, bad habit of mine to force using the local variable; they're funcionally the same so always complex to find an appropriate name (!= `localBookedForRunning`)

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
franz1981: And indeed it was, good eyes!

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -188,15 +181,27 @@ public ClassLoader run() {
...
-         scheduledExecutorService.schedule(runForScheduler, Math.min(period, period * value), timeUnit);
+   /**
+    * If {@code overBooking} is {@code true} a delay request can succeed only if:
+    * <ul>
+    *    <li>there is no other pending delay request
+    *    <li>there is no pending execution request
+    * </ul>
+    * <p>
+    * If {@code overBooking} is {@code false} a delay request must always succeed.<br>
+    * When a delay request succeed it schedule a new execution to happen in {@link #getPeriod()}.<br>
```
gemmellr, line 192: Comment needs updated

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -265,6 +270,9 @@ public void stop() {
       if (future != null) {
          future.cancel(false);
          future = null;
+         // Replace te existing one: a new periodic task or any new delay after stop
```
gemmellr, line 273: "the"

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -265,6 +270,9 @@ public void stop() {
...
+         // won't interact with the previously running ones
+         this.bookedForRunning = new AtomicBoolean(false);
```
gemmellr, line 275: The comment doesnt entirely fit if this is inside the null check, does it? It wont apply for a 'new delay' since the future doesnt exist for such 'on demand' usage.

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -287,35 +295,47 @@ private void restartIfNeeded() {
...
+      ActiveMQScheduledComponent.this.run();
+   }
 
-         if (!onDemand && lastTime > 0) {
-            if (System.currentTimeMillis() - lastTime < millisecondsPeriod) {
-               logger.trace("Execution ignored due to too many simultaneous executions, probably a previous delayed execution");
-               return;
+   private void bookedRunForScheduler(AtomicBoolean booked) {
+      if (executor != null) {
+         assert booked.get();
```
gemmellr, line 316: assert is in both legs, move outside if

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -287,35 +295,47 @@ private void restartIfNeeded() {
...
+      // It unblocks:
+      // - a new delay request
+      // - next periodic run request (in case of executor != null)
+      // Although tempting, don't move this one after ActiveMQScheduledComponent.this.run():
+      // - it can cause "delay" to change semantic ie a racing delay while finished executing the task, won't succeed
+      // - it won't prevent "slow tasks" to accumulate, because slowness cannot be measured inside running method;
+      //   it just cause skipping runs for perfectly timed executions too
+      // Re the code as it is: if the next periodic run has been able to request a new round while
+      // the previous one hasn't yet started, it makes sense to skip it, because its job will be executed by
+      // the old one (yet to start).
```
gemmellr, line 308: Not sure this bit of comment is in the right place, or if so could do with updating. This code here doesnt skip anything, as the comment reads like. It instead resets the flag used by other areas to skip.

franz1981: Yep effectively is not the right place for this bits of code, but is more the overall logic...I can put it in the header of the class docs!

clebertsuconic:
@franz1981 I have seen the results of the test suite on our CI, and in both cases the test suite did not complete.

I don't think it is safe to merge this ATM.

franz1981:
@clebertsuconic I'm now investigating, but is strange: the failures seems not related to my change, indeed I'm getting failures on tests that are not using the scheduled component...
I'm trying removing the latest commits before mine to check if that's the cause of the failure...

franz1981:
I've added a new commit but seems to create some trouble: I'll work today to fix what's wrong here...

franz1981:
@clebertsuconic It seems that I was missing to correctly stop :) 
The CI now is green :+1:

PR #3289, opened 2020-10-06
ARTEMIS-2918 Support Server restart on I/O errors

franz1981:
This PR include several changes:

- refactor node managers to allow a unified handling of lost lock events and I/O errors while dealing with the shared state
- improve granularity of the file lock acquisition timeout
- introduce the `restart-allowed` configuration parameter (`false` by default`) to allow the broker to restart itself in case of I/O critical failure

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
@@ -495,6 +495,8 @@ public final class ActiveMQDefaultConfiguration {
 
    public static final int DEFAULT_MAX_DISK_USAGE;
 
+   public static final boolean DEFAULT_RESTART_ALLOWED = false;
+
    static {
       int maxDisk;
       try {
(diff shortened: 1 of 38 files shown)


franz1981:
@jbertram @clebertsuconic Please add the "don't merge it yet" tag (seems that I can't do it :( )
@jbertram I know you've worked around error handling on activation, so I would love some help/review time from you if you can; we can discuss both here or on chat if needed :+1: 

@uomik This one could be useful for you as well :)

gtully:
I wonder if this feature could be recast as "keep the broker alive on JDBC io errors", a much narrower focus.

Keeping a general broker alive in a jvm without a jvm restart will be very difficult, b/c errors can occur at any point during start and stop!

If the focus is just JDBC, an essentially atomic storage, it could be possible to report JDBC errors to the clients and keep the broker alive.
A client would see an exception and/or closed connection and retry in the normal way.
The broker would resume normal service when it again was able to get a pooled JDBC connection.

Note: the lease locker will need to deal with the potential of a jdbc pool to block, it will need to track the lease duration locally such that it can expire and stop if the lease is lost. In other words, the lease will only survive jdbc errors for the remainder of its lease.

franz1981:
After a chat with @gtully I'm going to close this in favour of something more contained and simpler to improve the JDBC resiliency on network glitches and/or temporary failover(s)

franz1981:
This PR has been replaced by https://github.com/apache/activemq-artemis/pull/3301, so I'm gonna close it :+1:

PR #3290, opened 2020-10-06
ARTEMIS-2933 Upgrade com.sun.winsw to 2.7.0-net4

brusdev:
The only executable included in the broker bin package that depends on .NET Framework 3.5 is artemis-service.exe/winsw. Using winsw for .NET Framework 4 we should remove any dependency on .NET Framework 3.5.

This would simplify the installation on Windows because .NET Framework 4 is installed by default on the last versions of Windows.

### DIFF
--- a/artemis-cli/pom.xml
+++ b/artemis-cli/pom.xml
@@ -192,7 +192,7 @@
                    <groupId>com.sun.winsw</groupId>
                    <artifactId>winsw</artifactId>
                    <version>${winsw.version}</version>
-                   <classifier>bin</classifier>
+                   <classifier>net4</classifier>
                    <type>exe</type>
                    <outputDirectory>${basedir}/target/classes/org/apache/activemq/artemis/cli/commands/bin/</outputDirectory>
                    <destFileName>artemis-service.exe</destFileName>


jbertram:
Is this backwards compatible with 3.5? For example, if a user had .NET Framework 3.5 installed would the new winsw still work?

brusdev:
It will require .NET Framework 4 is installed by default on the last versions of Windows while .NET Framework 3.5 not.

jbertram:
If my understanding is correct this may break users running .NET Framework 3.5. If so I don't think we can do this in a minor release. I think we'd want to ship versions that work with either 3.5 or 4.

jbertram:
Perhaps the `pom.xml` could be modified so that 2 different versions of `artemis-service.exe` are created. The new one could have a new name to reflect its usage (e.g. `artemis-service-net4.exe` or perhaps with the same name but in a new `net4` directory) and the documentation could be updated accordingly (including deprecating the old one).

brusdev:
@jbertram `artemis-service-net4.exe` LGTM, I'll update the PR.

gemmellr:
I would maybe avoid adding simply '-net4' in the name/patch at this point given 6 is already in RC phase?

The support pages https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies suggest all versions of windows back to 8/server2012 have some .Net 4 framework version installed by default, and all versions back to 7/server2008 can get one if not installed.

Given that it doesnt seem unreasonable to require it at this point, so what about the reverse possibility? Make an alternative 'new' "-net3" version that ships for a limited time as a fallback option if we think it necessary (given above, I probably wouldnt), and then anyone actually affected by the change can use it instead if they dont want to use one of the many 4+ versions already available to them. 

I wouldn't say dropping support of a .Net version in a minor release is all that different than dropping support of a Java version in a minor release, which has been done before and is about to be done again.

gemmellr:
Given that last bit of my previous comment, maybe changing to requiring .Net 4 [by default, if also adding the fallback option] in the same release as changing to requiring Java 11 would be a nice alignment of support changes.

brusdev:
@gemmellr I'm investigating on the possibility to run executable built for .NET 2 on .NET 4 using the `App.config file`
https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-configure-an-app-to-support-net-framework-4-or-4-5

gemmellr:
If thats possible, great. I dont see a great need at this point personally. We are about to require a 4 year old JVM to run the broker...it doesnt seem unreasonable to require a ~10? year old version of .Net thats only optionally used to start/stop it.

brusdev:
@jbertram @gemmellr thanks for your feedback

PR #3291, opened 2020-10-07
ARTEMIS-2886 Optimize security auth

luisalves00:
As discussed in https://github.com/apache/activemq-artemis/pull/3254 I've added FQQN to ActiveMQSecurityManager5 when is available. Also had to do the same for the cache. So now the cache is working with the FQQN. I think there's no side effects as FQQN is more specific.
Please review an check if I forgot something.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
@@ -253,24 +253,22 @@ public class SecurityStoreImpl implements SecurityStore, HierarchicalRepositoryC
          /*
           * If a valid queue is passed in and there's an exact match for the FQQN then use the FQQN instead of the address
           */
-         boolean isFullyQualified = false;
          SimpleString fqqn = null;
          if (bareQueue != null) {
             fqqn = CompositeAddress.toFullyQualified(bareAddress, bareQueue);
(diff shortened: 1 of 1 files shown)


jbertram (commented):
I don't understand this commit. There's no semantic change as far as I can tell. The only real difference is that you're invoking `authorizationCache.put` *every* time instead of just when a new entry needs to be added.

The commit message says, "google uses immutables," but I don't see any reference to that in the Guava cache documentation. Furthermore, running rudimentary unit tests demonstrates that the `ConcurrentHashSet` in the cache is being updated as expected.

jbertram (changes requested):
I don't see any actual semantic difference from before. If there's a use-case which is broken without this fix then a test-case should be added to reproduce the problem and validate the fix. If not then either the commit messages needs to better justify the change or the commit should be omitted.

luisalves00:
@jbertram you are correct. Did a test https://github.com/luisalves00/activemq-artemis/blob/VerifyAuthorizationCache/artemis-server/src/test/java/org/apache/activemq/artemis/core/security/impl/AuthorizationCacheTest.java and cache works fine. I think the element was being evicted while I was debugging, as the default ttl is 10sec. I was thinking it was a copy on write kind of cache. So I'll revert everything except the ffqn != null. Thanks for the review.

jbertram (commented):
Please squash all your commits into one. Once that's done I'll merge it.

luisalves00:
@jbertram I think I did it well, but please verify.

PR #3293, opened 2020-10-07
ARTEMIS-2935 Fix JMX ObjectNames

brusdev:
Fix invalid object names setting a default broker name if it cannot be used.


gaohoward:
looking good

PR #3295, opened 2020-10-09
ARTEMIS-2931: Resolving potential NPEs detected by eclipse

pekuz:
WARNING: the eclipse static analyser is pretty limited and it cannot cope with cases like: correlated variables, exception paths, .... 70% of eclipse warnings on artemis codebase is a false alarm.

Anyway some variable correlations can be eliminated and code becomes more readable for humans too.

For "never-throws", the assumption is made explicit. If you disagree with the reverse-engineered assumption then it is likely an indication of a true potential NPE.

Last but not least, copy&paste is a common source of bugs. I suspect eclipse indirectly detected one such case.

Hope it helps

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
@@ -565,11 +565,12 @@ public class PagingStoreImpl implements PagingStore {
       SequentialFileFactory factory = null;
       try {
          factory = checkFileFactory();
+         SequentialFile file = factory.createSequentialFile(fileName);
+         return file.exists();
       } catch (Exception ignored) {
+         assert false : "PagingStoreFactory::newFileFactory never-throws assumption failed.";
+         return false;
       }
-
-      SequentialFile file = factory.createSequentialFile(fileName);
-      return file.exists();
    }
 
    @Override
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/replication/ReplicationEndpoint.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/replication/ReplicationEndpoint.java
@@ -405,7 +405,7 @@ public final class ReplicationEndpoint implements ChannelHandler, ActiveMQCompon
          }
       }
 
-      if (this.channel != null && outgoingInterceptors != null) {
+      if (channel != null && outgoingInterceptors != null) {
          if (channel.getConnection() instanceof RemotingConnectionImpl)  {
             try {
                RemotingConnectionImpl impl = (RemotingConnectionImpl) channel.getConnection();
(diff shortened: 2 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
```diff
@@ -565,11 +565,12 @@ public boolean checkPageFileExists(final int pageNumber) {
       SequentialFileFactory factory = null;
       try {
          factory = checkFileFactory();
+         SequentialFile file = factory.createSequentialFile(fileName);
+         return file.exists();
       } catch (Exception ignored) {
+         assert false : "PagingStoreFactory::newFileFactory never-throws assumption failed.";
```
jbertram, line 571: I'm not sure about this `assert` here. First, it references "newFileFactory" which I think is incorrect. Second, I think just returning `false` would be sufficient.

pekuz:
Hi Justin,

the "PagingStoreFactory::newFileFactory" is a reference to the interface method that is, transitively, the root of the problem, here.  

Its signature reads:

   SequentialFileFactory newFileFactory(SimpleString address) throws Exception;

Mere 

``  return false; 

could be an option but I hesitate until sure that exceptions can be silently swallowed. That's why I rather suggest to combine with the assert: to fail fast.  

  Hope it helps

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
pekuz: See the ticket, I received link to the ticket rather than to this change set so I reacted in the ticket.

michaelandrepearce: I agree with @jbertram having assert and rasing errors is not nice. Its much cleaner to return boolean with true false if it was successful

michaelandrepearce:
-1 adding assert it raises runtime exceptions. Either return a boolean if success or not or throw a normal exception that forces all calling code to handle. I would think returning boolean here is safer and cleaner imo

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
pekuz: the assert gives flexibility:
   in prod environments it can erased (like it would not exist in the sources)
   in dev environments it stops so one have proof out of contact situation actually happened

I went for assert as the contract is not evident. Returning mere false means that contract it clear and understood.  

If anybody helps to clarify contract then removing assert is super simple change to accept. 

  Hope it helps

pekuz:
Added clarification what motivates the assert and conditions for its removal.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
michaelandrepearce: To be clear i am -1 this as assert raises a runtime exception, you may as well just leave it with npe if youre doing that.  Other options have been chaired by myself and @jbertram and im sure there is alternatives too

michaelandrepearce:
My -1 remains

jbertram:
FWIW, I merged this, but I replaced the `assert` with a debug log statement.

PR #3296, opened 2020-10-12
ARTEMIS-2942 Disable hawtio proxy

brusdev:


michaelandrepearce:
What why?

brusdev:
hawtio comes with a proxy servlet so you can connect to all remote nodes in the whitelist (the whitelist is configured via the hawtio.proxyWhitelist system property). By default, hatwio versions >= 1.5.0 have a whitelist which only allows connections to 127.0.0.1. So you could only connect to a broker on the same node.
Given this default restriction, I don't think users are using this servlet with the default configuration and leaving it enabled could cause issues. With this PR I would disable the proxy servlet by default adding `-Dhawtio.disableProxy=true` to artemis.profile.
If a user needs this feature can remove `-Dhawtio.disableProxy=true` from artemis.profile.

PR #3297, opened 2020-10-12
[ARTEMIS-2939]: Artemis should not delete corrupt log files.

ehsavoie:
* Moving corrupted journal files to the attic folder.

Jira: https://issues.apache.org/jira/browse/ARTEMIS-2939

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
@@ -291,6 +291,9 @@ public final class ActiveMQDefaultConfiguration {
    // The minimal number of data files before we can start compacting
    private static int DEFAULT_JOURNAL_COMPACT_MIN_FILES = 10;
 
+   // The maximal number of data files before we can start deleting corrupted files instead of moving them to attic.
+   private static int DEFAULT_JOURNAL_MAX_ATTIC_FILES = 10;
+
    // Interval to log server specific information (e.g. memory usage etc)
    private static long DEFAULT_SERVER_DUMP_INTERVAL = -1;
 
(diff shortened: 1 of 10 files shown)


brusdev:
Limiting the corrupt log files moved to the attic folder could mitigate a future issue due to a disk full. I'm thinking at ActiveMQ Artemis running in a pod using a volume with a limited capacity.

ehsavoie:
In the case of a pod, wouldn't the restart of the broker be a restart of the pod itself ?

michaelandrepearce:
Currently with delete means theres no risk of disk filling up. With this it seems it now introduces that risk. -1 from me, until there is a process to limit the number. Also this feature should be configurable so that default behaviour remains as was, and you have to enable this to turn on. So users dont get a shock change of behaviour on upgrade.

ehsavoie:
@clebertsuconic WDYT ? ^^

clebertsuconic:
@michaelandrepearce / @ehsavoie I reviewed Emmanuel's branch before he sent the PR, and I asked him to remove the option and always move the file instead of removing.

This is something that should not happen under regular circumstances... the original intent of the feature was on misconfigured sizes.. the user configured at 1M then reconfigured the journal at 10M.. instead of reusing the file the file would be deleted.

As of a damaged file it would be better to move the file for further inspection.

Besides we have plenty of options.. for simplicity I asked to not add another option.


If you're concerned on eventually being disk full.. we could add a limit... or perhaps not start the broker until the attic is cleared... (although I feel a bit concerned on adding such thing when users have automatic tooling to restart the broker).

What about just having a max number of files... if the limit is reached we then start to remove them?


We could have a property to configure the max on the attic.

@ehsavoie my bad I made you remove the property.. but I really want to have the default behavior as moving the files always... a sensible default on the max attic size would be ok


@michaelandrepearce WDYT?

clebertsuconic:
@ehsavoie can you add a max-attic-size configuration parameter, defaulted at 10?


that way we can have it always active.... protect disk full.. etc... etc...

PR #3300, opened 2020-10-13
ARTEMIS-2947 Implement SecurityManager that supports replication

jbertram:


artemis-cli/src/test/java/org/apache/activemq/cli/test/ArtemisTest.java
```diff
@@ -105,6 +105,8 @@
 public class ArtemisTest extends CliTestBase {
    private static final Logger log = Logger.getLogger(ArtemisTest.class);
```
clebertsuconic, line 107: Just for the record.. I'm adding a comment here, and amending as your change:

```java
   // some tests will set this, as some security methods will need to know the server the CLI started
   private ActiveMQServer server;
```

PR #3301, opened 2020-10-13
ARTEMIS-2941 Improve JDBC HA connection resiliency

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-2941


franz1981:
I've run the CI tests on this and it's not causing regressions :+1:

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -377,7 +415,7 @@ protected FileLock lock(final int lockPosition) throws Exception {
                   return null;
```
brusdev, line 415: why are we catching the InterruptedException here and throwing it at the end ?

franz1981: Good point, let me check why I've made this change so I can comment :+1:

franz1981: I see why now, the original loop looked not correct...
the while (!interrupted) loop can end only if `interrupted == true`  and by contract we should throw `InterruptedException` for any caller there...but maybe the caller already take care of this, so better that I double check if I've changed semantic here, thanks!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/jdbc/ActiveMQScheduledLeaseLock.java
```diff
@@ -84,37 +89,55 @@ public synchronized void stop() {
...
+               lockLost = false;
+            }
+         }
+      }
+      // a failed attempt to renew is treated as a lost lock
+      if (lockLost) {
+         try {
+            lockListener.lostLock();
+         } catch (Throwable t) {
+            LOGGER.warnf(t, "Errored while notifying %s lock listener", lockName);
```
brusdev, line 116: Can an error on handle the lost lock lead to a split brain? Is it a critical error in this case?

franz1981: That's a good question: I've added this one to track OOM with a meaningful msg, but in theory a lost lock should just close forcibly the Node manager to allow any pending use of it to stop as well

franz1981: We should always stop on this, but if we get OOM there is a risk to split brain because we cannot create a new thread to stop the broker as expected

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
franz1981: I've decided to not change the original loop here :+1:

franz1981:
I've removed a byteman test that was relying on something that was already broken before this PR ie the number of tryLocks on FileLockNodeManager.

PR #3302, opened 2020-10-14
ARTEMIS-2948 & ARTEMIS-1187 use negotiated compatible version

hezhou66:
ARTEMIS-2948

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
@@ -1146,7 +1146,7 @@ public final class ClientSessionImpl implements ClientSessionInternal, FailureLi
 
    @Override
    public int getVersion() {
-      return sessionContext.getServerVersion();
+      return sessionContext.getCompatibleVersion();
    }
 
    @Override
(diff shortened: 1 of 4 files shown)


clebertsuconic:
What are you trying to fix here ?


There are no tests.   Take a look at the compatibility tests where I play with many versions as part of class loading.

hezhou66:
> What are you trying to fix here ?
> 
> There are no tests. Take a look at the compatibility tests where I play with many versions as part of class loading.

Thanks for reply . The tests are complicated and difficult at one project.  The process trace detail as follow ：

The client version is 2.4.x whitch has not SessionSendMessage_V2, sends SessionSendMessage packet to the server

The server version is 2.12.x , use ServerPacketDecoder.decodeSessionPacket method to decode message packet, but the method decode it to SessionSendMessage or SessionSendMessage_V2 depending on the connection version whitch has been set to server version(2.12.x) after failover reattach.

So the SessionSendMessage was decoded as SessionSendMessage_V2, losing the last Long correlationID, causing CoreMessage.decode error.

ServerPacketDecoder decodeSessionPacket
ActiveMQPacketHandler handleCreateSession
ActiveMQClientProtocolManager newSessionContext
ActiveMQSessionContext newCreateSession

clebertsuconic:
@hezhou66 you know that if you had added a test on the ./testsuite/compatibility-tests this would been merged already, right?


I was planning to it myself, however I'm swamped with my own PR as you can probably see.


Please consider adding the test.. or try to recruit someone else to do it.

jbertram:
@hezhou66, I'm looking to create a test-case for this to validate your fix and prevent potential regressions in the future. Can you elaborate on the full use-case? What is the minimum steps required to reproduce this?

hezhou66:
I'm sorry to get back so late.    I found  it was my own mistake.    I used standalone -- "live-only" policy, and I rewrote the client code by removing toplogy subscriber whitch I thought not useless in standalone model. But in witch  I  found the  version negotiation. Although the version number sored in the server  is not just the compatible versions, it is functional.

Once again, I'd like to apologize to you all！

clebertsuconic:
@hezhou66 no problem.. I'm glad it's ok...

If you find an issue around this or anywhere else.. a test case would help a lot... 

We can sometimes do with descriptions and do our best. but a test case always improves speed.

I'm glad this was not an issue. (as I understand).

PR #3303, opened 2020-10-14
ARTEMIS-2949 Reduce GC on OperationContext::checkTasks

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-2949

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
@@ -16,9 +16,7 @@
  */
 package org.apache.activemq.artemis.core.persistence.impl.journal;
 
-import java.util.Iterator;
 import java.util.LinkedList;
-import java.util.List;
 import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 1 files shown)


franz1981:
This PR reduce dramatically the amount of GC produced on the libaio blocked poll and the CPU time with it, see 
`master`
image
while this PR
image
That's 10 seconds CPU flamegraphs of the libaio poller thread for 1000 producer/consumer/queues sending 100 bytes persistent msg using Core JMS. 
In violet there are the calls to `LinkedList::*` that are now disappeared

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
```diff
@@ -191,41 +192,75 @@ public void executeOnCompletion(final IOCallback completion, final boolean store
...
-               // If set, we use an executor to avoid the server being single threaded
-               execute(holder.task);
-
-               iter.remove();
-            } else {
-               // End of list here. No other task will be completed after this
-               break;
-            }
-         }
+         checkGenericTasks();
```
clebertsuconic, line 263: rename this to checkCompleteContext please

franz1981: :+1:

franz1981: checkGenericTasks ->  checkCompleteContext ok?

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
```diff
@@ -304,6 +339,22 @@ public String toString() {
       }
    }
 
+   static final class StoreTaskHolder {
```
clebertsuconic, line 342: Rename this to StoreOnlyTaskHolder

and add a comment please stating why this is separate from TaskHolder? (just a single line is fine, nothing big)

clebertsuconic: Also add a comment on TaskHolder staying that we are not extending StoreTaskHolder because of memory optimizations.. as it would use more memory and this is on the hot path

PR #3305, opened 2020-10-15
ARTEMIS-2953 - drop the use of non deterministic finalize

gtully:
watch the PageFile tests in the build jobs.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
@@ -583,17 +583,6 @@ public final class Page implements Comparable<Page> {
    }
 
    @Override
-   protected void finalize() {
-      try {
-         if (file != null && file.isOpen()) {
-            file.close(false);
-         }
-      } catch (Exception e) {
-         ActiveMQServerLogger.LOGGER.pageFinaliseError(e);
-      }
-   }
-
-   @Override
    public int hashCode() {
       final int prime = 31;
       int result = 1;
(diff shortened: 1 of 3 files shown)


gtully:
Adding dome debugging to the finalise, I am not finding any failures in the integration tests but I see that lots of the unit tests leave pages dangling that get cleaned up via finalize.
I will push a debug build to see what fails in the real broker tests.

gtully:
fmi: a case where the finalize kicks in from the 14 build, that test still needs some clean up:
java.lang.Error: Finalize is doing close work on: Page::pageNr=129, file=FakeSequentialFile:000000129.page
	at org.apache.activemq.artemis.core.paging.impl.Page.finalize(Page.java:594)
	at java.base/java.lang.System$2.invokeFinalize(System.java:2190)
	at java.base/java.lang.ref.Finalizer.runFinalizer(Finalizer.java:87)
	at java.base/java.lang.ref.Finalizer$FinalizerThread.run(Finalizer.java:171)
Caused by: java.lang.Throwable: Thread[Thread-6121,5,main]
	at org.apache.activemq.artemis.core.paging.impl.Page.open(Page.java:485)
	at org.apache.activemq.artemis.core.paging.impl.PagingStoreImpl.openNewPage(PagingStoreImpl.java:1122)
	at org.apache.activemq.artemis.core.paging.impl.PagingStoreImpl.page(PagingStoreImpl.java:859)
	at org.apache.activemq.artemis.tests.unit.core.paging.impl.PagingStoreImplTest$1WriterThread.run(PagingStoreImplTest.java:426)

gtully:
I found a few tests that caused the finalize to kick in and resolve the leaks. I think this is good to go now.

PR #3306, opened 2020-10-19
ARTEMIS-2888 ARTEMIS-2859 ARTEMIS-2768 - revert new page-store-name a…

gtully:
…ddressSetting, when the page store respects the target address and the size is tallied on the target address store, it is no longer neecessary


clebertsuconic:
can you close the relevant JIRAs as rejected?

PR #3307, opened 2020-10-19
[ARTEMIS-2954]: RA doesn't use the RA specified prefix when setting up a destination.

ehsavoie:
* When creating a destination it should rely on the RA configuration if
   none is provided instead of relying directly on the default
configuration.

Issue: https://issues.apache.org/jira/browse/ARTEMIS-2954


clebertsuconic:
Our test suite should cover this.. and I know you have other tests on wildly for this


so I'm merging it

PR #3257, opened 2020-09-11
ARTEMIS-2838 - migrate to HawtIO 2

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-2838

The main reason of upgrading to HawtIO 2 is that HawtIO 1 is no longer supported for CVE's etc.

Although this is a rewrite all the screens perform the same functionally, although there are improvements and additions.

Refer to the Jira for more info.

Since this is a big change Id be grateful if anyone has the time to review and kick the tyres.

PR #3309, opened 2020-10-20
ARTEMIS-2838 - forgot to add a send message tab for addresses.

andytaylor:
plus removed some erronious logging

https://issues.apache.org/jira/browse/ARTEMIS-2838


michaelpearce-gain:
@andytaylor  / @clebertsuconic  have the images in docs been updates since the hawtio upgrade? I flagged this on https://github.com/apache/activemq-artemis/pull/3257 as well but no response.

PR #3311, opened 2020-10-21
ARTEMIS-2957 ManagementContext is started twice

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-2957

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Run.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Run.java
@@ -109,7 +109,6 @@ public class Run extends LockAbstract {
 
          server = BrokerFactory.createServer(broker.server, securityManager, activateCallback);
 
-         managementContext.start();
          server.createComponents();
          AtomicBoolean serverActivationFailed = new AtomicBoolean(false);
          server.getServer().registerActivationFailureListener(exception -> serverActivationFailed.set(true));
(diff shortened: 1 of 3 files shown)


franz1981:
@andytaylor it's safe to remove the first `managementContext.start()` on  https://github.com/apache/activemq-artemis/pull/3311/files#diff-517541877ba25124d5cbedf79e343779f2fb16a256dd7da4f30dd10055ac06afL112?

Just checking for the `ManagementContext` to not being started twice is fixing this: I don't want introduce a further bug for no reason :)

franz1981:
I'm getting a different failure occasionally now:
```
[main] 08:38:30,022 INFO  [org.apache.activemq.artemis.tests.util.ActiveMQTestBase] **** end #test testFailbackOnJMX() ***
[ERROR] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 11.331 s <<< FAILURE! - in org.apache.activemq.artemis.tests.smoke.jmxfailback.JmxFailbackTest
[ERROR] testFailbackOnJMX(org.apache.activemq.artemis.tests.smoke.jmxfailback.JmxFailbackTest)  Time elapsed: 11.214 s  <<< FAILURE!
```
That seems that the backup JMX connection isn't avaialble yet

franz1981:
It seems that removing the double start of `ManagementContext` will make the test to fail: I'm investigating if I can left the double start to be as it is given that now `ManagementContext` start/stop are thread-safe

PR #3312, opened 2020-10-22
ARTEMIS-2958 Timed out waiting pool stop on backup restart

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-2958

BackupManager::activated should correctly unblock
unauthorized blocking requests to allow a backup broker
to prompty restart in case of live failback

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/BackupManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/BackupManager.java
@@ -192,9 +192,9 @@ public class BackupManager implements ActiveMQComponent {
       private TransportConfiguration connector;
       protected long retryInterval;
       private ClusterManager clusterManager;
-      private boolean stopping = false;
-      private boolean announcingBackup;
-      private boolean backupAnnounced = false;
+      private volatile boolean stopping = false;
+      private volatile boolean announcingBackup;
+      private volatile boolean backupAnnounced = false;
 
       @Override
       public String toString() {
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
@@ -643,6 +643,10 @@ public final class ChannelImpl implements Channel {
          unlock();
       }
       closed = true;
+      // unblock any blocked call:
+      // don't move this one before closed = true, because
+      // unblocked calls need to check if (closed) to "gracefully" shutdown.
+      returnBlocking();
    }
 
    @Override


franz1981:
@clebertsuconic Please don't merge it yet: I'm introducing a change into a core class so would like to be sure everything is fine first...

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/BackupManager.java
```diff
@@ -313,6 +316,15 @@ public void informTopology() {
       * */
       public void close() {
          stopping = true;
```
gtully, line 318: looks like stopping needs to be volatile also

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/BackupManager.java
```diff
@@ -267,7 +268,8 @@ public void run() {
                   ClientSessionFactoryInternal backupSessionFactory = localBackupLocator.connect();
```
gtully, line 268: Can this call also block?

gtully: stopping the executor really should be bound, first stop and wait for completion, then whack it via shutdown now. I guess with agressive default tolerances.
keeping track of outstanding blocking commands as they are executed will be tedious :-)

gtully (approved):
this looks sensible to me, but I just wonder if other calls could block there. Is not the key ensuring that the executor stops in a timely manner?
Give it a chance to complete to do graceful shutdown, then whack it!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/BackupManager.java
franz1981: > Can this call also block?

It has the logic to retry connecting in a loop but I'm now aware if is happening: i suppose yes given that the old code was tracking it with `announcingBackup = true;`

> keeping track of outstanding blocking commands as they are executed will be tedious :-)
That's correct, the problem here is who has ownership on `ClusterControl`: if is `BackupManager` it should be in charge of closing a leaking one and is up to others (an existing connection), to unblock any blocked call.

franz1981:
@clebertsuconic the CI has run fine with this change, just not sure that's ok for `ClusterControl::close` to be called *only* if blocked on authorization or it should be eventually closed anyway

franz1981:
I've found a simple way to fix this without impacting the current logic 
The real problem seems that `BackupManager::activated` that's called when the backup is becoming live (much before the failback) was calling `BackupConnector::close` that was in charge to `closeLocator(backupServerLocator)`, that was closing any underlying connection to the live without unblocking any pending authorization.
Just allowing a `Channel::close` to correctly awake any pending operation is enough to fix this, preventing a task to block the thread pool stop *if* the backup is going to stop to let the live failback.

gtully:
that looks a lot simpler. nice!

PR #3313, opened 2020-10-22
[ARTEMIS-2954]: RA doesn't use the RA specified prefix when setting u…

ehsavoie:
…p a destination.

* Fixing the fix for double prefix issue as the destiantion name wasn't
  appended when the old prefix wasn't present.

Issue: https://issues.apache.org/jira/browse/ARTEMIS-2954


artemis-ra/src/main/java/org/apache/activemq/artemis/ra/inflow/ActiveMQActivation.java
```diff
@@ -597,12 +597,12 @@ private String getTopicWithPrefix(String topic) {
       if (spec.getTopicPrefix() == null) {
          if (spec.isEnable1xPrefixes() == null) {
             if (ra.isEnable1xPrefixes() != null && ra.isEnable1xPrefixes() && !topic.startsWith(PacketImpl.OLD_TOPIC_PREFIX.toString())) {
-               return PacketImpl.OLD_TOPIC_PREFIX.toString();
+               return PacketImpl.OLD_TOPIC_PREFIX.toString() + topic;
```
clebertsuconic, line 600: ouch!!! :)

glad you picked this one up

PR #3315, opened 2020-10-27
ARTEMIS-2964 - fire advisory messages via post office, independent of…

gtully:
… connection state. consume via regular auth

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
@@ -50,6 +50,7 @@ import org.apache.activemq.artemis.api.core.RoutingType;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.client.ActiveMQClientLogger;
 import org.apache.activemq.artemis.core.io.IOCallback;
+import org.apache.activemq.artemis.core.persistence.CoreMessageObjectPools;
 import org.apache.activemq.artemis.core.persistence.OperationContext;
 import org.apache.activemq.artemis.core.postoffice.Binding;
 import org.apache.activemq.artemis.core.postoffice.Bindings;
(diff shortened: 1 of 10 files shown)


gtully:
@franz1981 the coreMessageObjectPools has now moved to be shared across AMQSession. I think that makes sense, it can be reused for advisories that will likely share the addresses. I wonder about the locality of an object pool, is it something you have thought about?

franz1981:
@gtully The pool itself can pool just things until certain limits and it really depends how much sharing affect temporal locality vs the capacity of the pool. In term of threading model, the pool is perfectly thread-safe, because isn't precise, so it should be fine

clebertsuconic:
you had some real failures on this PR:


Test Name | Duration | Age
-- | -- | --
org.apache.activemq.artemis.tests.integration.security.SecurityPerAcceptorJmsTest.testJAASSecurityManagerAuthorizationNegative[protocol=OPENWIRE] | 0.12 sec | 1
org.apache.activemq.artemis.tests.integration.security.SecurityTest.testJAASSecurityManagerOpenWireNegative | 0.43 sec | 1
org.apache.activemq.artemis.tests.integration.security.SecurityTest.testJAASSecurityManagerAuthenticationWithCertsAndOpenWire | 0.24 sec | 1
org.apache.activemq.artemis.tests.integration.server.SecureConfigurationTest.testSecureDurableSubscriber[2: protocol=OPENWIRE] | 0.24 sec | 1
org.apache.activemq.artemis.tests.integration.server.SecureConfigurationTest.testSecureQueue[2: protocol=OPENWIRE]

clebertsuconic:
also on org.apache.activemq.artemis.tests.smoke.jmxmultiplefailback.ReplicatedMultipleFailbackTest.testMultipleFailback

clebertsuconic:
they seem test issues only... it seems the tests need to be adapted due to the changes you're making...


but still if you could handle it please?

gtully:
Thanks.

Good to see those tests, they nicely demonstrate the need for permissions on Advisories (ActiveMQ.Advisory.#) or avoiding their use via:
openwire connectionFactory.setWatchTopicAdvisories(false) or url parameter jms.watchTopicAdvisories=false

sorted with examples of each approach.

PR #3317, opened 2020-10-28
ARTEMIS-2937 Fixing Tests and some review

clebertsuconic:


clebertsuconic:
I still have some review to add here.

clebertsuconic:
@gemmellr I need to merge this as I have another non related fix that is depending on this. (If you send a message to an ANYCAST queue with a distinct name you would have a failure and an auto-created queue). 

We can send more commits to review on this.

PR #3322, opened 2020-10-30
ARTEMIS-2969 / ARTEMIS-2937 Dealing with Connection Timeout properly

clebertsuconic:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPBridgeTest.java
```diff
@@ -54,17 +54,6 @@ protected ActiveMQServer createServer() throws Exception {
       return createServer(AMQP_PORT, false);
    }
 
-   @Test
-   public void testsSimpleConnect() throws Exception {
```
clebertsuconic, line 58: This test was not doing anything at this point..

I had use for it during dev, but as things progressed it is not needed any longer.

clebertsuconic:
I will add a test on QpidDispatchPeerTest for now.. and I will develop something more complete on next week.

PR #3324, opened 2020-10-30
ARTEMIS-2823 Apply default datasource configs if not overridden

franz1981:


franz1981:
@uomik ;) was still missing something this one before the release

PR #3330, opened 2020-11-05
ARTEMIS-2971: add ANONYMOUS and EXTERNAL SASL support

gemmellr:
add ANONYMOUS and EXTERNAL SASL mechanism support for outgoing AMQP server connections.

Uses a vertx-proton based test server to validate the connection behaviour. This could be be similarly used to better test a lot of the connection behaviour of the broker-connection functionality. E.g checking the source/target details of connecting senders or receivers, or validating expected mirror-related messages get sent on activity, or testing disconnect handling, or checking the waypoint related config is added while using the 'peer' connection type without needing a Dispatch server.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPConnectSaslTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPConnectSaslTest.java
@@ -1,0 +1,299 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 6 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPConnectSaslTest.java
```diff
@@ -0,0 +1,299 @@
...
+
+      String amqpServerConnectionURI = "tcp://localhost:" + mockServer.actualPort() +
+            "?sslEnabled=true;trustStorePath=" + TRUSTSTORE_NAME + ";trustStorePassword=" + TRUSTSTORE_PASSWORD;
+      if (requireClientCert) {
+         amqpServerConnectionURI += ";keyStorePath=" + CLIENT_KEYSTORE_NAME + ";keyStorePassword=" + CLIENT_KEYSTORE_PASSWORD;
+      }
+
+      AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration("testSimpleConnect", amqpServerConnectionURI);
+      amqpConnection.setReconnectAttempts(0);// No reconnects
+      amqpConnection.setUser(USER); // Wont matter if EXTERNAL is offered and a client-certificate is provided, but will otherwise.
```
clebertsuconic, line 201: shouldn't we add actual java properties supporting this?

I mean, adding properties on the .XSD? 

I know you may say you could go all the way to a lot of stuff that's not he URI, but this seems a nice use case to support through distinct properties on the AMQPBrokerConnection configuration.

gemmellr: The test was just checking it uses external if offered, even in the face of other options, rather than because its the only option. There isnt any config needed for it, just basically 'it is offered, I have a cert'.

I did think about adding the ability to restrict which mechanisms it can pick, which would make sense, but I decided its a seperate change. Also, its more likely to make the StackOverflowError occur so I think fixing that would need to occur concurrently or before.

PR #3333, opened 2020-11-12
Fix unecessary closes

lcc:
Meaningless Close: In several classes, close(), specified in Closeable interface, has no effect and other methods in those classes can be called after close() without IOException.


michaelandrepearce:
-1 even if underlying implementation does nothing with the close its good practice to still call the close method if the class implements closable at the point you may want to close the resources as there is nothing stopping the implementations to in future have logic behind the close method to free up internal resource.

lcc:
You do have a point! I guess future-proofing in this case is more important than a slightly cleaner code. Thanks for the feedback, would you like me to close this PR?

PR #3334, opened 2020-11-12
ARTEMIS-2984 Compressed large messages can leak native resources

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-2984

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerImpl.java
@@ -19,9 +19,9 @@ package org.apache.activemq.artemis.core.client.impl;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
-import java.util.concurrent.atomic.AtomicLong;
 
 import org.apache.activemq.artemis.api.core.ActiveMQException;
+import org.apache.activemq.artemis.api.core.ActiveMQLargeMessageException;
 import org.apache.activemq.artemis.api.core.ICoreMessage;
 import org.apache.activemq.artemis.api.core.Message;
 import org.apache.activemq.artemis.api.core.SimpleString;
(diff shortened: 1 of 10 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/LargeMessageController.java
```diff
@@ -18,10 +18,9 @@
 
 import java.io.OutputStream;
 
-import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
 
-public interface LargeMessageController extends ActiveMQBuffer {
```
clebertsuconic, line 24: this is an API change, isn't?

it's part of the Core Client API when dealing with large messages...

it's a client API.. I don't think. you can guarantee there are no users using this API.

artemis-core-client/src/test/java/org/apache/activemq/artemis/util/CompressionUtilTest.java
```diff
@@ -104,42 +103,6 @@ public void testDeflaterReader2() throws Exception {
       compareByteArray(allCompressed, output, compressedDataLength);
    }
 
-   @Test
```
clebertsuconic, line 106: what happened to this test?

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/client/impl/ReadableLargeMessageController.java
```diff
@@ -0,0 +1,877 @@
...
+import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
+import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.core.client.ActiveMQClientLogger;
+import org.apache.activemq.artemis.core.client.impl.ClientConsumerInternal;
+import org.apache.activemq.artemis.core.client.impl.LargeMessageControllerImpl;
+import org.apache.activemq.artemis.utils.ByteUtil;
+import org.apache.activemq.artemis.utils.DataConstants;
+import org.apache.activemq.artemis.utils.UTF8Util;
+
+final class ReadableLargeMessageController extends LargeMessageControllerImpl implements ActiveMQBuffer {
```
clebertsuconic, line 37: This was part of the API Contract. and you moved it to tests only.

artemis-core-client/src/test/java/org/apache/activemq/artemis/util/CompressionUtilTest.java
franz1981: `InflaterReader` was no longer needed, so could be removed (need to be sure of this one yet)!

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/client/impl/ReadableLargeMessageController.java
franz1981: I see that we expose `LargeMessageController` just on:
```java
public interface ClientLargeMessageInternal extends ClientMessageInternal {

   void setLargeMessageController(LargeMessageController controller);

   LargeMessageController getLargeMessageController();
```
So, if users are not allowed to use internal classes this shouldn't be a big deal (yet to be verified)

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/LargeMessageController.java
franz1981: answered on https://github.com/apache/activemq-artemis/pull/3334#discussion_r522248647

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/client/impl/ReadableLargeMessageController.java
clebertsuconic: Not allowed is a strong word.. Users are discouraged to do it..

the original intention of the controller was to be exposed, however this was never achieved.

I think we should have the class you moved to the Testsuite sitting on the same package, in case this breaks someone usage they have the option to decorate the controller with the class you moved, with a big warning they should avoid doing it.

franz1981:
@clebertsuconic I'm reworking on this to save users to break their expectations but I see some interesting things that could improve large message streaming, see:
```java
   private byte getByte(final long index) {
      checkForPacket(index);

      if (fileCache != null && index < packetPosition) {
         return fileCache.getByteFromCache(index);
      } else {
         return currentPacket.getChunk()[(int) (index - packetPosition)];
      }
   }
```
This method is the only one using `FileCache`, part of the `ActiveMQBuffer` API on large messages, that's internal.
Large message streaming is slowed down (and get some additional memory footprint increase too) by the `FileCache` presence, because it needs to save each incoming chunk into it as well, but if users won't read it it's just useless work...

clebertsuconic:
@franz1981 I'm lost on what you mean on your last comment.. you mean we should remove usage, or you have already reworked the usage?

franz1981:
@clebertsuconic 
> you mean we should remove usage, or you have already reworked the usage?

I couldn't do it because of the `ActiveMQBuffer` API on `LargeMessageController` we expose to users: if we drop it, we could save using `FileCache` while receiving chunks of the large message, both speeding it up and reducing its memory footprint.
Now we have a `jmh` module and I can work on a benchmark that can show the speedup I'm talking about

clebertsuconic:
@franz1981 what about the previous propose? to make the usage separate? we break the API but allow an alternative?

you had a class that you had moved to tests.. all I asked you was to move that class to the API.. so if someone is broken by this they could just change their code to use this new class.

they are using an internal method.. they would be broken by it.. but at least they would have an alternative.

if that's not possible then we can talk about breaking it for sure.

michaelandrepearce:
Do we really expose the controller to users? Yes its public but is it exposed in the client apis?

clebertsuconic:
we don't expose it...

Franz had moved the controller to a separate class on a test..

all I asked was to have that class as part of the main jar. 

But if that's a big deal.. just take it out then.

franz1981:
@michaelandrepearce @clebertsuconic If we take the step to remove it I'm sure large message streaming can just drop any file cache and both code and perf will benefit there

> what about the previous propose? to make the usage separate? we break the API but allow an alternative?

If the usage wasn't exposed to remove the whole code paths that makes large msg streaming complex, saving both perf and leaky methods to exists as a whole. If removing isn't an option I can just prepare the code for a future removal, when we decide to do that, but we won't get any perf and stability benefits from removing it.
Adding a separate class to perform the read still won't allow me to remove the FileCache, so no perf benefits...

michaelandrepearce:
@franz1981 i think we established end users dont see this class we can do with it as we want. Removal, refactor, give it cups of tea.

franz1981:
Please don't merge it, I'm still trying to be 100% sure that file cache isn't used..but maybe it's used somehow...

clebertsuconic:
If it's easy. Keep the class available like before. 

If not.  Just take it out. 

Keep it simple.

franz1981:
As it is now is just packed in a different class to make it more readeable :) 
Hence this version on this PR should be fine, just need to run the CI on it

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerImpl.java
```diff
@@ -410,108 +410,117 @@ private void largeMessageSendBuffered(final boolean sendBlocking,
...
+            throw ActiveMQClientMessageBundle.BUNDLE.errorReadingBody(e);
+         }
+         if (read == -1) {
+            // bytesRead can be 0 if the stream return -1 after 0-length reads
+            return -bytesRead;
+         }
+         bytesRead += read;
+      }
+      while (bytesRead < chunkLimit);
+      assert bytesRead == chunkLimit;
```
clebertsuconic, line 522: is this assert correct? what if you are reading a large message that Is not a multiple of chunkLimit?

clebertsuconic:
@franz1981 what's the status on this PR? is this ready to merge?

I saw briefly one assert (that I added a comment) that I wasn't sure at a first look.

michaelandrepearce:
@clebertsuconic / @franz1981 whats occuring with this, wanting to start clearing down old / stagnant PR's a bit of spring cleaning so to say - so we focus on stuff thats actively working on / relevant .

franz1981:
Closing this because large messages changed a bit recently: need to be reworked

PR #3336, opened 2020-11-12
ARTEMIS-2985 - don't block netty threads for mqtt protocol actions

gtully:
I am still working on an intermittent test failure, more eyes welcome
MQTTQueueCleanTest.testQueueCleanWhenConnectionSynExeConnectAndDisconnect


gtully:
the test failure was the result of adding a consumer without being fully initialised.
server == null
see: https://github.com/apache/activemq-artemis/pull/3336/files#diff-b56a3b568eccf7ca40814304ea1776eae27b1a4a2e5273cc2403470eae6414a3R257

PR #3338, opened 2020-11-13
ARTEMIS-2990 - avoid auto boxing on binding id used as map key

gtully:


franz1981:
One suggestion: given that we need some reliable way to measure improvementsbe it's time to bring into Artemis something like https://issues.apache.org/jira/plugins/servlet/mobile#issue/PROTON-1690
Wdyt?
I can help doing it and writing the bench

gtully:
@franz1981 I think that would be great. I have a poor mans ignored test that I was using to root around to see some low hanging fruit. For sure a proper bench mark would be better. What proton has done seems perfect, and we can start small.
go fo it.

PR #3339, opened 2020-11-13
ARTEMIS-2991 Removing HttpAcceptorHandler from HttpKeepAliveRunnable when upgrade to websocket connection

ldebello:
Sending this small fix to resolve the ticket https://issues.apache.org/jira/browse/ARTEMIS-2991

# Implementation

The idea is to remove the HttpAcceptorHandler from the list of handlers when the upgrade is perform

# Test

1) Do you know a good way of testing this? I was thinking using JMSWebSocketConnectionTest to later get the NettyAcceptor but not sure about how to get it, I can check other approaches but I would like to know your opinion before creating the test

2) Also I need to test the the websocket connection keeps opens without the need of that handler.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/ProtocolHandler.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/ProtocolHandler.java
@@ -39,7 +39,6 @@ import io.netty.handler.codec.http.HttpRequestDecoder;
 import io.netty.handler.codec.http.HttpResponseEncoder;
 import org.apache.activemq.artemis.api.core.client.ActiveMQClient;
 import org.apache.activemq.artemis.core.buffers.impl.ChannelBufferWrapper;
-import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
 import org.apache.activemq.artemis.core.remoting.impl.netty.ConnectionCreator;
 import org.apache.activemq.artemis.core.remoting.impl.netty.HttpAcceptorHandler;
 import org.apache.activemq.artemis.core.remoting.impl.netty.HttpKeepAliveRunnable;
@@ -47,6 +46,7 @@ import org.apache.activemq.artemis.core.remoting.impl.netty.NettyAcceptor;
 import org.apache.activemq.artemis.core.remoting.impl.netty.NettyConnector;
 import org.apache.activemq.artemis.core.remoting.impl.netty.NettyServerConnection;
 import org.apache.activemq.artemis.core.remoting.impl.netty.TransportConstants;
+import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
 import org.apache.activemq.artemis.core.server.protocol.websocket.WebSocketServerHandler;
 import org.apache.activemq.artemis.spi.core.protocol.ProtocolManager;
 import org.apache.activemq.artemis.utils.ConfigurationHelper;
@@ -85,6 +85,10 @@ public class ProtocolHandler {
       return new ProtocolDecoder(true, false);
    }
 
+   public HttpKeepAliveRunnable getHttpKeepAliveRunnable() {
+      return httpKeepAliveRunnable;
+   }
+
    public void close() {
       if (httpKeepAliveRunnable != null) {
          httpKeepAliveRunnable.close();
@@ -96,6 +100,8 @@ public class ProtocolHandler {
    }
 
    class ProtocolDecoder extends ByteToMessageDecoder {
+
+      private static final String HTTP_HANDLER = "http-handler";
 
       private final boolean http;
 
@@ -141,7 +147,7 @@ public class ProtocolHandler {
                ctx.pipeline().addLast("websocket-handler", new WebSocketServerHandler(websocketSubprotocolIds, ConfigurationHelper.getIntProperty(TransportConstants.STOMP_MAX_FRAME_PAYLOAD_LENGTH, TransportConstants.DEFAULT_STOMP_MAX_FRAME_PAYLOAD_LENGTH, nettyAcceptor.getConfiguration())));
                ctx.pipeline().addLast(new ProtocolDecoder(false, false));
                ctx.pipeline().remove(this);
-               ctx.pipeline().remove("http-handler");
+               ctx.pipeline().remove(HTTP_HANDLER);
                ctx.fireChannelRead(msg);
             } else if (upgrade != null && upgrade.equalsIgnoreCase(NettyConnector.ACTIVEMQ_REMOTING)) { // HORNETQ-1391
                // Send the response and close the connection if necessary.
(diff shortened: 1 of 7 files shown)


jbertram:
If you're going to pursue this solution you should probably just implement the `handlerRemoved` method on `org.apache.activemq.artemis.core.remoting.impl.netty.HttpAcceptorHandler`, e.g.:

```java
   @Override
   public void handlerRemoved(final ChannelHandlerContext ctx) throws Exception {
      super.handlerRemoved(ctx);
      httpKeepAliveTask.unregisterKeepAliveHandler(this);
   }
```

jbertram:
Also, for what it's worth, I don't think the HTTP keep-alive is necessary for the websocket connection. Websockets already have a ping/pong mechanism.

ldebello:
> If you're going to pursue this solution you should probably just implement the `handlerRemoved` method on `org.apache.activemq.artemis.core.remoting.impl.netty.HttpAcceptorHandler`, e.g.:
> 
> ```java
>    @Override
>    public void handlerRemoved(final ChannelHandlerContext ctx) throws Exception {
>       super.handlerRemoved(ctx);
>       httpKeepAliveTask.unregisterKeepAliveHandler(this);
>    }
> ```

Awesome I didn't see that method that would be better. I will change it. Do you think is the correct fix or I need to something better?

ldebello:
> Also, for what it's worth, I don't think the HTTP keep-alive is necessary for the websocket connection. Websockets already have a ping/pong mechanism.

Yes I think the same but I need to verify that

jbertram:
Using `handlerRemoved` seems like a legitimate fix. Only testing will confirm, though.

ldebello:
I have added a test to verify the unregister is called

ldebello:
Let me know if I am missing something or if you have some comments

ldebello:
One extra question I saw there is a tag for 2.16.0 not sure if possible to include this for that version or we will need to wait until 2.17.0 or 2.16.x currently we are restarting our broker every two weeks to avoid issues from the leak handlers

jbertram:
@ldebello, 2.16.0 has already been released. The release announcement was sent to the ActiveMQ User List. You can find the announcement in the [list archive](http://mail-archives.apache.org/mod_mbox/activemq-users/202011.mbox/%3CCAKF%2BbsrKmN%3D0WbfQgUcbPz5558WiaBPZJe9bHUF%3D2Gh5t%3DbTAg%40mail.gmail.com%3E). The download is available [here](http://activemq.apache.org/components/artemis/download/).

If you really want/need this fix you can always download the tag of 2.16.0, apply this fix, and either build a full release (e.g. using `mvn clean install -Prelease` or just build the jar for the `activemq-server` module and drop that in to your existing environment.

As far as this PR goes you need to squash your commits and then reference ARTEMIS-2991 in the commit message as described in the [hacking guide](https://github.com/apache/activemq-artemis/blob/master/docs/hacking-guide/en/code.md#commitMessageDetails).

ldebello:
> @ldebello, 2.16.0 has already been released. The release announcement was sent to the ActiveMQ User List. You can find the announcement in the [list archive](http://mail-archives.apache.org/mod_mbox/activemq-users/202011.mbox/%3CCAKF%2BbsrKmN%3D0WbfQgUcbPz5558WiaBPZJe9bHUF%3D2Gh5t%3DbTAg%40mail.gmail.com%3E). The download is available [here](http://activemq.apache.org/components/artemis/download/).
> 
> If you really want/need this fix you can always download the tag of 2.16.0, apply this fix, and either build a full release (e.g. using `mvn clean install -Prelease` or just build the jar for the `activemq-server` module and drop that in to your existing environment.
> 
> As far as this PR goes you need to squash your commits and then reference ARTEMIS-2991 in the commit message as described in the [hacking guide](https://github.com/apache/activemq-artemis/blob/master/docs/hacking-guide/en/code.md#commitMessageDetails).

Thanks for the response, I have already squash the commits and change the message, thanks for the help

jbertram:
This looks good for the most part, but I think the test is a bit weak. I would like to see an integration test that would fail before your fix and succeed after it which tests the actual number of handlers (i.e. the size of `org.apache.activemq.artemis.core.remoting.impl.netty.HttpKeepAliveRunnable#handlers`). Let me know if you need help implementing this.

ldebello:
> This looks good for the most part, but I think the test is a bit weak. I would like to see an integration test that would fail before your fix and succeed after it which tests the actual number of handlers (i.e. the size of `org.apache.activemq.artemis.core.remoting.impl.netty.HttpKeepAliveRunnable#handlers`). Let me know if you need help implementing this.

I totally agree with you, I tested manually because I didnt find a good way of accesing to the handler list, because the ProtocolHandler which creates the Runnables lives is created by the NettyAcceptor in the constructor. One option could be reflection in each instance until reach the handler seems a little dirty but possible, mocking everything seems kind of hard, what do you think is the best wait to test this?

jbertram:
It's common to add getters or other helper methods when writing tests. You'll need to add a method to get the `ProtocolHandler` from the `NettyAcceptor` as well as a method to get the `HttpKeepAliveRunnable` from the `ProtocolHandler` and finally a method to get the `List<HttpAcceptorHandler>` from the `HttpKeepAliveRunnable`.

Once you create/configure/start the embedded server for your test you can use the `org.apache.activemq.artemis.core.server.ActiveMQServer#getRemotingService` method and then `org.apache.activemq.artemis.core.remoting.server.RemotingService#getAcceptor`. Once you have the right `Acceptor` instance you can use the methods you added to get the list of handlers. Your test should create a bunch of websocket connections and make sure the size of the handler list doesn't keep growing.

ldebello:
> It's common to add getters or other helper methods when writing tests. You'll need to add a method to get the `ProtocolHandler` from the `NettyAcceptor` as well as a method to get the `HttpKeepAliveRunnable` from the `ProtocolHandler` and finally a method to get the `List<HttpAcceptorHandler>` from the `HttpKeepAliveRunnable`.
> 
> Once you create/configure/start the embedded server for your test you can use the `org.apache.activemq.artemis.core.server.ActiveMQServer#getRemotingService` method and then `org.apache.activemq.artemis.core.remoting.server.RemotingService#getAcceptor`. Once you have the right `Acceptor` instance you can use the methods you added to get the list of handlers. Your test should create a bunch of websocket connections and make sure the size of the handler list doesn't keep growing.

I am not a very fond of adding getters for testing but if that is ok I will try to go on path, I will include additional tests to validate the fix, thanks for the help

jbertram:
I don't see any problem with adding getters for testing in this kind of situation. None of these classes is part of a "public" API. Testing is a critical part of development and tests should be able to get all the state they need from a class in order to test it effectively. It's much simpler and straight-forward to just add the getters instead of using something ugly and brittle like reflection. Also, mocking is good for a lot things, but we need an actual integration test here.

ldebello:
> I don't see any problem with adding getters for testing in this kind of situation. None of these classes is part of a "public" API. Testing is a critical part of development and tests should be able to get all the state they need from a class in order to test it effectively. It's much simpler and straight-forward to just add the getters instead of using something ugly and brittle like reflection. Also, mocking is good for a lot things, but we need an actual integration test here.

I have added the get and I created a small interface for the keep alive handler, let me know if you are ok with the changes or if you think another test should be added

jbertram:
@ldebello, was there a specific reason for adding a new interface? It seems unnecessary to me.

ldebello:
> @ldebello, was there a specific reason for adding a new interface? It seems unnecessary to me.

I thought it was better, to later try to provide a new implement for the runnable. I saw a PR including JMH tests so I thought about waiting for that PR to be merged to try to change the current implementation providing some tests.

If you think that is not necessary I can change it, to include the getter and return an inmutable list of handlers.

jbertram:
I don't think the interface is necessary until there is a clear need for a different implementation. I would remove it to simplify the code. In any event, the name of the interface was confusing.

ldebello:
> I don't think the interface is necessary until there is a clear need for a different implementation. I would remove it to simplify the code. In any event, the name of the interface was confusing.

np, I can remove it :). I will do the change, thanks for the help

ldebello:
@jbertram  I have done the change, let me know your thoughts

clebertsuconic:
@jbertram I'm out of context on this PR.. if you could merge this one ?

PR #3342, opened 2020-11-13
NO-JIRA: remove duplicate consecutive words

uroesch:
Removes duplicate consecutives words from markdown
documentation files.


jbertram:
Thanks for the contribution!

PR #3343, opened 2020-11-14
ARTEMIS-2996 Provide JMH Benchmarks for Artemis

franz1981:


franz1981:
That's a draft PR aiming to provide some reliable numbers to check various Artemis component performance.
In particular for https://github.com/apache/activemq-artemis/pull/3338

So @gtully if curious try to check with me if what I've written make sense for your use case ;)

franz1981:
@gtully To run the tests from the IDE should download https://plugins.jetbrains.com/plugin/7529-jmh-java-microbenchmark-harness 
And I can help with the various setting (enabling profiler, gc logs etc etc)

franz1981:
@gtully FYI sometime I'm getting a concurrent modification exception while iterating on the linked addresses of AddressImpl, so probably there is some missing sync somewhere...
And the `addBinding` method seems that is not properly unit-tested too, so I'm adding a commit with the missing test + fix

franz1981:
Had some fun with this and found an easy low hanging fruit that really improve performance a lot...
Using 4 threads for `testAddRemoveNewBindingWhilePublish` and 4 on `testPublishWhileAddRemoveNewBinding`:

`master`
```
Benchmark                                                                (topicsLog2)   Mode  Cnt          Score          Error  Units
WildcardAddressManagerPerfTest.both:testAddRemoveNewBindingWhilePublish            16  thrpt   16      95647.902 ±     5049.284  ops/s
WildcardAddressManagerPerfTest.both:testPublishWhileAddRemoveNewBinding            16  thrpt   16  137072955.875 ± 21611759.467  ops/s
```
This pr, after making `SEPARATOR` `static final`:
Benchmark                                                                (topicsLog2)   Mode  Cnt          Score          Error  Units
WildcardAddressManagerPerfTest.both:testAddRemoveNewBindingWhilePublish            16  thrpt   16      94823.178 ±     4462.129  ops/s
WildcardAddressManagerPerfTest.both:testPublishWhileAddRemoveNewBinding            16  thrpt   16  174427240.181 ± 14332414.006  ops/s
```

**Comment by franz1981** [comment]
For `getBindingsForRoutingAddress` with already existing bindings the major bottleneck seems
image
But luckly its concurrent performance scale linearly with the number of threads, because there are no locks

**Comment by michaelandrepearce** [comment]
> For `getBindingsForRoutingAddress` with already existing bindings the major bottleneck seems `CompositeAddress.extractAddressName(address)` (in violet)
> 
> image
> 
> But luckly its concurrent performance scale linearly with the number of threads, because there are no locks

Is it worth having a cache/map between the composite name and the binding so that this isnt done every time.

**Comment by franz1981** [comment]
@michaelandrepearce Yep, it seems something good, unless I can find a very simple and obvious way to improve what we have eg by using some better `indexOf(String)` algorithm -> something like https://github.com/netty/netty/pull/10737

**Comment by michaelandrepearce** [comment]
Could actually just make the extra composite address entry at time of queue creation and remove at queue deletion into the binding lookup map, then you don't need to worry if you get composite or not its going to be a simple lookup in the same maps

**Comment by franz1981** [comment]
@michaelandrepearce I've made a little math mistake and now results have a different shape...address lookup now is bottlenecked by CHM::get, see

image

**Comment by franz1981** [comment]
While the add/remove binding case has several bottlenecks, let me see if they can addressed "easily"

**Comment by franz1981** [comment]
FYI the motivation here is to improve scalability with > 100 subscribers :+1:

**Comment by michaelandrepearce** [comment]
Doing it at queue create and delete time i think will be cleaner. Whilst appreciate youre wanting a quick win, we have to think about longer term

**Comment by franz1981** [comment]
> Doing it at queue create and delete time i think will be cleaner. Whilst appreciate youre wanting a quick win, we have to think about longer term. E.g. i dont think we should be doing add updates at message send. This really should be with queue lifecycle

Yep, it's correct...the thing is that querying seems already quite fast...just FYI:
```
Benchmark                                                          (topicsLog2)   Mode  Cnt         Score         Error  Units
WildcardAddressManagerPerfTest.testJustAddRemoveNewBinding                    8  thrpt   16      1131.959 ±      67.676  ops/s
WildcardAddressManagerPerfTest.testJustPublish                    	      8  thrpt   16  91773172.028 ± 2440317.871  ops/s
```
So add and remove a new binding with 1 << 8 ie 256 subscribers has a very bad performance

**Comment by franz1981** [comment]
add binding seems to be slowed down by `AddressImpl::addLinkedAddress` that's calling `ArrayList::indexOf`:

image

**Comment by franz1981** [comment]
remove binding is slowed down due to traversing CHM values (the bindings, that's ok) and `CompositeAddress::extractQueueName`:

image

**Comment by franz1981** [comment]
ATM remove binding is far more expensive then adding it

**Comment by michaelandrepearce** [comment]
Thats not hot path though. Sending messages is. Youre not expecting add and remove of queues every ms

**Comment by franz1981** [comment]
@michaelandrepearce depends, there are few cases with mqtt that need some love there as well :)
That's why I am working on this ;)

**Comment by michaelandrepearce** [comment]
But mqtt souldnt be adding and removing queues at a rate. It can have 100s of subscribers but they shouldn't be coming an going at a rates like we have message flow.

**Comment by michaelandrepearce** [comment]
As in i get optimising for routing to 100s of queues, but the add remove of a queue not so. As such this is why i suggested that we create the link between composite and binding at queue create and delete

**Comment by franz1981** [comment]
I believe your suggestion is good to get the best from routing (that's ok for othes and more common/important use cases): re mqtt, I am aware of some cases that can stress the same behaviour in have implemented on the JMH benchmark ie given a big amount of topics, concurrent adding and removing few new ones (routing few messages) while routing on the other ones.
I can add some fake routing between add/remove but the problem is always the same: locking and linear (double) traversal of matching bindings (see the linked addresses behaviour) are the bottlenecks.
Improving this use case won't hurt existing ones (that's my purpose here) and that's why this benchmark can help to detect.

Tomorrow I will let @gtully continue on this; I know he was working on something related

**Comment by michaelandrepearce** [comment]
If the issue is readpath being blocked as add removes are occuring. Why not look at copy on write structure. As such read aka lookup on message handling will be lock free at cost of write path add remove queue.

**Comment by franz1981** [comment]
@gtully I believe I've mostly fixed the scalability of binding removal (that's more then 4X improved now, gong from O(n) to O(1)).
I'm going to fix the addition as said on https://github.com/apache/activemq-artemis/pull/3343#issuecomment-727621137

**Comment by franz1981** [comment]
These are the results.

This pr:
```
Benchmark                                                                (topicsLog2)   Mode  Cnt          Score          Error  Units
WildcardAddressManagerPerfTest.both:testAddRemoveNewBindingWhilePublish             2  thrpt   16     299899.123 ±     9376.689  ops/s
WildcardAddressManagerPerfTest.both:testPublishWhileAddRemoveNewBinding             2  thrpt   16   90076643.252 ± 21148545.456  ops/s

WildcardAddressManagerPerfTest.both:testAddRemoveNewBindingWhilePublish             8  thrpt   16       7615.982 ±      123.914  ops/s
WildcardAddressManagerPerfTest.both:testPublishWhileAddRemoveNewBinding             8  thrpt   16   83680564.528 ±   781632.643  ops/s

WildcardAddressManagerPerfTest.both:testAddRemoveNewBindingWhilePublish            10  thrpt   16       1061.621 ±      128.786  ops/s
WildcardAddressManagerPerfTest.both:testPublishWhileAddRemoveNewBinding            10  thrpt   16   48221531.411 ±  2804251.587  ops/s

WildcardAddressManagerPerfTest.testJustAddRemoveNewBinding                          2  thrpt   16     465088.099 ±    34561.615  ops/s
WildcardAddressManagerPerfTest.testJustAddRemoveNewBinding                          8  thrpt   16      11778.483 ±      173.307  ops/s
WildcardAddressManagerPerfTest.testJustAddRemoveNewBinding                         10  thrpt   16       2109.041 ±      186.689  ops/s

WildcardAddressManagerPerfTest.testJustPublish                                      2  thrpt   16  274036118.300 ±  4554573.431  ops/s
WildcardAddressManagerPerfTest.testJustPublish                                      8  thrpt   16  187202367.351 ±  2023585.447  ops/s
WildcardAddressManagerPerfTest.testJustPublish                                     10  thrpt   16  101350692.610 ±  2772091.382  ops/s
```
`master`:
```
**

Benchmark                                                                (topicsLog2)   Mode  Cnt          Score         Error  Units
WildcardAddressManagerPerfTest.both:testAddRemoveNewBindingWhilePublish             2  thrpt   16     295434.319 ±   10580.352  ops/s
WildcardAddressManagerPerfTest.both:testPublishWhileAddRemoveNewBinding             2  thrpt   16  100350741.463 ± 7166844.441  ops/s

WildcardAddressManagerPerfTest.both:testAddRemoveNewBindingWhilePublish             8  thrpt   16       1181.538 ±      36.217  ops/s
WildcardAddressManagerPerfTest.both:testPublishWhileAddRemoveNewBinding             8  thrpt   16   86594664.222 ± 1040794.968  ops/s

WildcardAddressManagerPerfTest.both:testAddRemoveNewBindingWhilePublish            10  thrpt   16         67.873 ±       9.049  ops/s
WildcardAddressManagerPerfTest.both:testPublishWhileAddRemoveNewBinding            10  thrpt   16   49433517.093 ±  596134.766  ops/s

WildcardAddressManagerPerfTest.testJustAddRemoveNewBinding                          2  thrpt   16     442642.438 ±    9116.475  ops/s
WildcardAddressManagerPerfTest.testJustAddRemoveNewBinding                          8  thrpt   16       2145.601 ±      95.437  ops/s
WildcardAddressManagerPerfTest.testJustAddRemoveNewBinding                         10  thrpt   16        156.090 ±      10.565  ops/s

WildcardAddressManagerPerfTest.testJustPublish                                      2  thrpt   16  262489608.447 ± 2390608.018  ops/s
WildcardAddressManagerPerfTest.testJustPublish                                      8  thrpt   16  170683082.929 ± 4646044.717  ops/s
WildcardAddressManagerPerfTest.testJustPublish                                     10  thrpt   16   97865949.150 ±  955235.418  ops/s
```

add/remove bindings is improved ten-fold with many subscribers (1024), while querying (useful on common routing) is a bit improved (see `testJustPublish` results).

From now on I'm going to fix the routing bottlenecks

**Comment by gtully** [comment]
there are some nice wins here. good work.
the jmh having a non permissive license, means we cannot ship code with reference. There is a nice example in protonj, a pom that does not deploy anything and makes a nice executable jar.
see: https://github.com/apache/qpid-proton-j/blob/1e6c7a3c8a89e8d9d7c4e36eeee623a3b5f657fd/tests/performance-jmh/pom.xml
that is the pattern we need to follow for the bench test code. a new module that is disabled by default

**Comment by gtully** [comment]
I put the jmh tests into a new tests/performance-jmh module that is not build by default.

File: artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/AddressImpl.java [hunk 1]
```diff
@@ -67,17 +69,23 @@ public boolean containsWildCard() {
    }
 
    @Override
-   public List<Address> getLinkedAddresses() {
-      return linkedAddresses;
+   public Collection<Address> getLinkedAddresses() {
+      return linkedAddresses == null ? Collections.emptySet() : linkedAddresses;
```
gtully, line 73: There are a bunch of places where getLinkedAddresses are iterated over, and the collection is not thread safe. We need to double check those are covered by an external sync such that we don't get concurrent mod exceptions.

gtully: hmm, I think org.apache.activemq.artemis.core.postoffice.PostOffice#updateMessageLoadBalancingTypeForAddress will cause trouble.

clebertsuconic:
nice job on this one @franz1981 .. I just have one question which I will ask inline.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/WildcardAddressManager.java
```diff
@@ -221,19 +221,21 @@ private void addAddress(final SimpleString address, final Address actualAddress)
...
-      if (bindings == null || bindings.getBindings().size() == 0) {
-         List<Address> addresses = address.getLinkedAddresses();
-         for (Address address1 : addresses) {
-            address1.removeLinkedAddress(address);
-            Bindings linkedBindings = super.getBindingsForRoutingAddress(address1.getAddress());
-            if (linkedBindings == null || linkedBindings.getBindings().size() == 0) {
-               removeAddress(address1);
+      if (bindings == null || bindings.getBindings().isEmpty()) {
+         synchronized (this) {
+            List<Address> addresses = address.getLinkedAddresses();
```
clebertsuconic: I'm confused here.. it seems to me that you can either move the synchronize to something specific to the address (synchronize (address)  { // the one you returned from getLinkedAddresses();


or the whole thing should be synchronized...

Are you sure this is the right thing?

clebertsuconic: I'm also concerned with the dependency added.. can we move it to extra-tests?

franz1981: > it seems to me that you can either move the synchronize to something specific to the address (synchronize (address) { // the one you returned from getLinkedAddresses();
or the whole thing should be synchronized...

I've no idea TBH: consider that in other methods eg `addAndUpdateAddressMap` we treat changes in the 2 directions (the addresses referenced by `address::getLinkedAddresses` and `address`) as a whole/atomically, so I wasn't sure to further narrow the lock scope to something different.
Instead, I already have a little concern on the correctness as it is now, but I need to write a concurrent test to verify this, so will update this comment later.

> Are you sure this is the right thing?

Never :D 
From a performance perspective this is already a nice step beyond, because in the hot path of binding removal we're not anymore causing additions to be blocked. The ideal thing would be to NOT have any sparse linked address list, but an internal data structure on `WildcardAddressManager` (maybe a concurrent/or not BiMap or something similar?)

franz1981: > I'm also concerned with the dependency added.. can we move it to extra-tests?

Sure, I'm going to do it now and will write a test to verify that I haven't break any concurrent behaviour...if not, the `synchronize` changes are a minor if compared to the other improvements and we can address them later, given that we now have the right tool (the benchmark) to spot regressions (on CI as well ;) )

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/AddressImpl.java
franz1981: I see that as well and I'm a bit concern about it too, but I don't think that the initialization will make any difference here...
There is a synchronization problem here. period...
It's weird we've never had issues on this, so I'm going to write a test and check if I can break it. If not, putting a synchronize will just kill performance and the same if we copy the whole linked addresses list (or we use a copy on write data structure too, but on addBinding/removeBinding side)....

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/WildcardAddressManager.java
franz1981: About the `extra-tests` thing...
I'm not sure a `test` thing is the right place, because JMH benchs can (and should) use an annotation processor to create a `jar` that can be used on a CI to run benchmarks...JMH benchs are not junit tests (and shoudn't given that JUnit is not the proper env...)

franz1981: @clebertsuconic The optional jmh module has been created

gtully: it is now in its own module, and can set the manifest main in its benchmark.jar and that module needs to be manually built and won't deploy.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/AddressImpl.java
gtully: I added a test for the concurrency issue with updateMessageLoadBalancingTypeForAddress.  The simplest solution is a concurrent set.

tests/performance-jmh/README.md
```diff
@@ -0,0 +1,41 @@
+Apache Artemix JMH Benchmarks
+-------
+This module contains optional [JMH](http://openjdk.java.net/projects/code-tools/jmh/) performance tests.
+
+Note that this module is an optional part of the overall project build and does not deploy anything, due to its use
+of JMH which is not permissively licensed. The module must either be built directly.
```
gemmellr, line 6: "The module must either be built directly." Sentence fragment that doesnt make any sense after the 'or enabled with a profile' followup was dropped from the original proton-j source inspiration.

tests/performance-jmh/README.md
```diff
@@ -0,0 +1,41 @@
...
+To run all JMH benchmarks:
+
+    java -jar target/benchmark.jar
+
+To list available benchmarks:
+
+    java -jar target/proton-j-performance-jmh.jar -l
+Some JMH help:
+
+    java -jar target/proton-j-performance-jmh.jar -h
```
gemmellr, line 34: Missed update for the jar names (assuming the module was actually renamed ;)).

tests/performance-jmh/pom.xml
```diff
@@ -0,0 +1,94 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <parent>
+        <groupId>org.apache.activemq.tests</groupId>
+        <artifactId>artemis-tests-pom</artifactId>
+        <version>2.17.0-SNAPSHOT</version>
```
gemmellr, line 23: Since this module isn't part of the main build, this presumably wont be updated during the release process, meaning it wont be possible to build the module from source once released (and the matching snapshot is deleted) without modification or also building the older snapshot. That is, assuming the test source actually compiles at all at that point, which it may not if it hasn't been used in a bit. Not necessarily a blocker, but something to note.

(Handling this is part of what the profile trickery in the proton-j source inspiration does, ensuring the module is enabled and re-versioned during release process, such that it can actually be used later from the produced source release, and keeps a check that the module actually compiles).

gemmellr (commented):
The changes to the actual server etc modules seem like they should be in/against their own PR/JIRA rather than one explicitly indicated for adding benchmarks. Maybe it could be renamed to reflect those changes/discussion, and the benchmark bits perhaps moved to their own referencing new one.

tests/performance-jmh/pom.xml
clebertsuconic: You have to make sure the source build would not include it.

gemmellr: You really dont. You can decide to.

franz1981:
That's the results against the very last version (using concurrent hash sets/maps for the linked addresses): 
```
Benchmark                                                                (topicsLog2)   Mode  Cnt          Score         Error  Units
WildcardAddressManagerPerfTest.both:testAddRemoveNewBindingWhilePublish             2  thrpt   16     308471.410 ±    5716.540  ops/s
WildcardAddressManagerPerfTest.both:testPublishWhileAddRemoveNewBinding             2  thrpt   16  105066755.032 ± 4830432.429  ops/s

WildcardAddressManagerPerfTest.both:testAddRemoveNewBindingWhilePublish             8  thrpt   16       6029.859 ±     298.996  ops/s
WildcardAddressManagerPerfTest.both:testPublishWhileAddRemoveNewBinding             8  thrpt   16   84996842.522 ± 2890128.047  ops/s

WildcardAddressManagerPerfTest.both:testAddRemoveNewBindingWhilePublish            10  thrpt   16       1127.679 ±      94.390  ops/s
WildcardAddressManagerPerfTest.both:testPublishWhileAddRemoveNewBinding            10  thrpt   16   48979269.260 ±  657710.616  ops/s

WildcardAddressManagerPerfTest.testJustAddRemoveNewBinding                          2  thrpt   16     508689.125 ±    7910.394  ops/s
WildcardAddressManagerPerfTest.testJustAddRemoveNewBinding                          8  thrpt   16      11218.972 ±     307.024  ops/s
WildcardAddressManagerPerfTest.testJustAddRemoveNewBinding                         10  thrpt   16       1918.918 ±     143.462  ops/s

WildcardAddressManagerPerfTest.testJustPublish                                      2  thrpt   16  298489735.550 ± 3336624.283  ops/s
WildcardAddressManagerPerfTest.testJustPublish                                      8  thrpt   16  186897808.588 ± 2513969.124  ops/s
WildcardAddressManagerPerfTest.testJustPublish                                     10  thrpt   16  101008224.759 ± 2169930.719  ops/s
```
we're within the margin of errors but I still need to check how the memory footprint has changed

tests/performance-jmh/pom.xml
clebertsuconic: @gemmellr We had people making a big deal about the extra-tests as part of the source tree years ago.

perhaps we didn't need that removed?

PR #3344, opened 2020-11-16
ARTEMIS-2990 - update wildcard address map on creation only, avoid du…

gtully:
…plicates and duplicate checks

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/AddressImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/AddressImpl.java
@@ -17,7 +17,6 @@
 package org.apache.activemq.artemis.core.postoffice.impl;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 
 import org.apache.activemq.artemis.api.core.SimpleString;
@@ -37,7 +36,7 @@ public class AddressImpl implements Address {
 
    private final boolean containsWildCard;
 
-   private List<Address> linkedAddresses = null;
+   private final List<Address> linkedAddresses = new ArrayList<>();
 
    private final WildcardConfiguration wildcardConfiguration;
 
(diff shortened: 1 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/AddressImpl.java
```diff
@@ -69,24 +68,16 @@ public boolean containsWildCard() {
...
 
    @Override
    public void addLinkedAddress(final Address address) {
-      if (linkedAddresses == null) {
-         linkedAddresses = new ArrayList<>(1);
-      }
-      if (!linkedAddresses.contains(address)) {
-         linkedAddresses.add(address);
-      }
+      linkedAddresses.add(address);
```
franz1981, line 76: I would add an `assert !linkedAddress,contains(address)` here as an invariant on the tests

michaelandrepearce:
@gtully polite ask. Can you please let prs open for at least a day or two for people to have time to review before merging. This pr was raised and merged in only a few ours not allowing anyone time to review

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/AddressImpl.java
```diff
@@ -37,7 +36,7 @@
 
    private final boolean containsWildCard;
 
-   private List<Address> linkedAddresses = null;
+   private final List<Address> linkedAddresses = new ArrayList<>();
```
michaelandrepearce, line 39: Whats the impact here for those not using wildcards

gtully: there is an impact, it is reverted to lazy init via https://github.com/apache/activemq-artemis/pull/3343 where a set works better for deletions.

gtully:
@michaelandrepearce polite ack.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/AddressImpl.java
michaelandrepearce: Dont add assert into runtime code.

michaelandrepearce: It is better to follow the pattern from set interface and return true / false if operation actually actioned

PR #3345, opened 2020-11-17
ARTEMIS-2997 NPE with JMS queue/topic toString()

jbertram:


artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQDestination.java
```diff
@@ -464,7 +464,7 @@ public boolean isQueue() {
    // Public --------------------------------------------------------
 
    public String getAddress() {
-      return simpleAddress.toString();
+      return simpleAddress != null ? simpleAddress.toString() : null;
```
clebertsuconic, line 467: toString should always return something not null IMO. Can you instead do:

```java
return simpleAddress != null ? simpleAddress.toString() : "<null address>";
```

or simply

```java
   return "" + simpleAddress;
```

jbertram: `toString()` itself *won't* return `null` since it's returning `"ActiveMQQueue[" + getAddress() + "]"`

jbertram: `getAddress()` is returning `null` not `toString()`

michaelandrepearce: Theres a static toString method on simple string that handles null from memory surely just use that for cleaness like in other code areas

jbertram: @michaelandrepearce, I don't see any such method on `SimpleString`.

michaelandrepearce: E.g. you could do Objects.toString(simpleAddress, null)

jbertram: FWIW, the pattern `.* != null ? .*\.toString\(\) : null` is used in a handful of places throughout the code-base (including elsewhere in `ActiveMQDestination`) whereas `Objects.toString\(.*\)` is used just once.

jbertram: I think we're getting close to a bike shed here.

clebertsuconic: @jbertram when I read your changes here:

```java
   public String getAddress() {
      return simpleAddress != null ? simpleAddress.toString() : null;
```

I thought this was the toString()


my bad!

sorry for the confusion!

PR #3348, opened 2020-11-17
NO-JIRA Update Travis and github builds to include jmh as part of the builds

clebertsuconic:
This will make sure PR checks will always check for issues on the new module


clebertsuconic:
@franz1981 can you review this please?

PR #3350, opened 2020-11-18
ARTEMIS-2987 improve message body display

jbertram:


artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/browse.js
```diff
@@ -80,25 +80,26 @@ var Artemis;
             </div>
             <div class="form-group" ng-show="$ctrl.showMessageDetails">
                 <button class="btn btn-primary" ng-click="$ctrl.showMessageDetails = false">Back</button>
-                <h2>MessageID: {{$ctrl.currentMessage.messageID}}</h2>
-                <h2>Headers</h2>
+                <h4>Message ID: {{$ctrl.currentMessage.messageID}}</h4>
+
+                <h4>Displaying body as <span ng-bind="$ctrl.currentMessage.textMode"></span></h4>
```
clebertsuconic, line 85: did you test this on both AMQP and Core?

jbertram: I did a quick test for this with core, but this change is really just superficial. There's no semantic change here. The same information is being displayed as always, just in a slightly different way.

clebertsuconic: I see. I haven't fully read the whole PR at the time I posted this.. I thought you implemented some visutalzation of the body, in which case you would have needed to change some java code.. 

it's fine then.. thanks

PR #3352, opened 2020-11-20
ARTEMIS-2990 - alway be getBiased and only publish complete records a…

gtully:
…nd only calculate linked addresses once ARTEMIS-2990

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/WildcardAddressManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/WildcardAddressManager.java
@@ -63,7 +63,7 @@ public class WildcardAddressManager extends SimpleAddressManager {
 
       // this should only happen if we're routing to an address that has no mappings when we're running checkAllowable
       if (bindings == null && !wildCardAddresses.isEmpty()) {
-         Address add = addAndUpdateAddressMap(address, true);
+         Address add = addAndUpdateAddressMap(address);
          if (!add.containsWildCard()) {
             for (Address destAdd : add.getLinkedAddresses()) {
                Bindings b = super.getBindingsForRoutingAddress(destAdd.getAddress());
@@ -94,7 +94,7 @@ public class WildcardAddressManager extends SimpleAddressManager {
    public boolean addBinding(final Binding binding) throws Exception {
       boolean exists = super.addBinding(binding);
       if (!exists) {
-         Address add = addAndUpdateAddressMap(binding.getAddress(), false);
+         Address add = addAndUpdateAddressMap(binding.getAddress());
          if (add.containsWildCard()) {
             for (Address destAdd : add.getLinkedAddresses()) {
                super.addMappingInternal(destAdd.getAddress(), binding);
@@ -113,7 +113,7 @@ public class WildcardAddressManager extends SimpleAddressManager {
 
    @Override
    public void updateMessageLoadBalancingTypeForAddress(SimpleString address, MessageLoadBalancingType messageLoadBalancingType) throws Exception {
-      Address add = addAndUpdateAddressMap(address, true);
+      Address add = addAndUpdateAddressMap(address);
       Bindings bindingsForRoutingAddress = super.getBindingsForRoutingAddress(address);
       if (bindingsForRoutingAddress != null) {
          bindingsForRoutingAddress.setMessageLoadBalancingType(messageLoadBalancingType);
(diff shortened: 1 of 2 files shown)


gtully:
@franz1981 the little assert you mentioned on the linked address collection would have caught this. Am wiser now. Apart from the duplicate work on an existing (thing add/remove binding), I think we were also publishing too early. That is now sorted.

franz1981:
I would add an assert to verify the addressMap.put always return null, if I have correctly understood your change.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/WildcardAddressManager.java
```diff
@@ -175,36 +175,38 @@ public void clear() {
...
+   private Address addAndUpdateAddressMap(final SimpleString address) {
       final boolean containsWildCard = address.containsEitherOf(wildcardConfiguration.getAnyWords(), wildcardConfiguration.getSingleWord());
       final Map<SimpleString, Address> addressMap = containsWildCard ? wildCardAddresses : addresses;
-      Address actualAddress = null;
-      if (getBiased) {
-         // CHM::get doesn't need to synchronize anything, so it's to be preferred for getBiased cases
-         actualAddress = addressMap.get(address);
-      }
+      Address actualAddress = addressMap.get(address);
       if (actualAddress == null) {
```
franz1981, line 182: it's just a personal thing with code: 
```java
if (actualAddress != null) {
    return actualAddress;
}
// from now one, slow path
```
It's helpful for the brain to not further reading the rest of code, if not needed 
and it allows to pack the slow path code into a specific (`synchronized`) method too

franz1981 (approved):

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/WildcardAddressManager.java
```diff
@@ -175,36 +175,38 @@ public void clear() {
...
-                  actualAddress.addLinkedAddress(destAdd);
-               }
-            }
-         } else {
-            for (Address destAdd : wildCardAddresses.values()) {
-               if (actualAddress.matches(destAdd)) {
-                  destAdd.addLinkedAddress(actualAddress);
-                  actualAddress.addLinkedAddress(destAdd);
+         synchronized (this) {
+            actualAddress = addressMap.get(address);
```
franz1981, line 184: Same here: https://github.com/apache/activemq-artemis/pull/3352/files#r527734942

franz1981 (approved):

franz1981:
I like the idea, so the load (fast) path is now safe, because the linked addresses are correctly published already

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/AddressImpl.java
```diff
@@ -84,6 +84,7 @@ public void addLinkedAddress(final Address address) {
       if (linkedAddresses == null) {
          linkedAddresses = PlatformDependent.hasUnsafe() ? new NonBlockingHashSet<>() : new ConcurrentHashSet<>();
       }
+      assert !linkedAddresses.contains(address);
```
michaelandrepearce, line 87: Previously if it called twice it wont error but simply proceed. Is this really a good idea, risk breaking or erroring

michaelandrepearce: Im not feeling comfortable about this.

michaelandrepearce: Better to change return as boolean and simply return true or false if succeeded and if you care if it succeeded use that to subsequently handle.

franz1981: I like this solution!

gtully: I am not a fan of asserts generally, but in this case, if I added that assert with my original change, this PR would not be necessary. I think it serves a purpose. It forces a test failure on bad usage.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/WildcardAddressManager.java
gtully: true, that is better.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/AddressImpl.java
gtully: see the original pr: https://github.com/apache/activemq-artemis/pull/3344#discussion_r524344331

michaelandrepearce: But that is at risk at causing a runtime issue in real systems. If you want to validate it, as noted and as per general collection kind of behaviour its normal to return true false if a change occured not error

jbertram: I'm not advocating one way or the other here, but aren't assertions disabled by the JVM by default and must be enabled using `-ea`?

franz1981: In this case what could be the proper action in case we see the linked address to be added twice? A log? An exception?
I suppose both are poor to capture a performance need (more then a correctness thing). I am open to proposals here but I am not sure asserts are not good if is to guide specific invariant to be checked with ki cost at runtime, especially for performance reasons.

franz1981: Said that I think is not bad to return a Boolean but we still need to under what action to take in case of false branch: If not, I don't see a reason to add it to the API. Differently from a collection, that's a generic reusable piece of code, this one is domain code and isn't there to be reusable.

franz1981: @jbertram yep and iirc surefire add it by default while running tests, that's why I have suggested that option

michaelandrepearce: We shouldnt be relying on code that only does something in tests as then at runtime we have different behaviour.

michaelandrepearce: If you need something just for tests we would be better extracting interface and then in test injecting a mock or interception that checks the call count.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/WildcardAddressManager.java
gtully: I looked at changing this, if I follow the same pattern on the get check inside the sync I end up with a method with 3 return statements. At the moment it has a single return, which i like.
I thin the single return statement is better than 2 fast paths and one slow :-)

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/WildcardAddressManager.java
gtully: that is the crux, one fast return path is good. 2 now gives us in total 3 return paths. I think it is more readable with a single return path.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/AddressImpl.java
gtully: in this case, the different behaviour is correct, just a waste. The code now does the add/remove of linked addresses once, and if someone changes that, the assert will fire. When I made that optimisation without the assert, it was dropped in a few commits and none the wiser. an assert is a good win here imho. It is suggesting the contract but it is not a fatal flaw to insert the same value twice, just a waste.
@michaelandrepearce are you -1 on asserts or just this assert ?

gtully: to progress this, the assert is gone!

PR #3357, opened 2020-11-23
ARTEMIS-3009 - reference remoteBinding loadbalancing type on add bind…

gtully:
…ing, rather then doing a second pass via the po and address manager. rework ARTEMIS-1680


gtully:
rebased

PR #3363, opened 2020-11-26
ARTEMIS-3016 Reduce DuplicateIDCache memory footprint

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3016


franz1981:
@gtully I'm going to add a JMH benchmark to measure that performance hasn't regressed too.

First tests using 100 caches with 20K elements and `persist` shows a big improvement; I think this could be further improved by playin' a bit dirty with Object Oriented Programming :yum: 

Test:
```java
   public static void main(String[] args) throws Exception {
      final int caches = 100;
      final int capacity = 20_000;
      final byte[][] ids = new byte[capacity][];

      for (int c = 0; c < capacity; c++) {
         ids[c] = java.util.UUID.randomUUID().toString().getBytes();
      }

      DuplicateIDCache[] cc = new DuplicateIDCache[caches];
      for (int i = 0; i < caches; i++) {
         DuplicateIDCache c = new DuplicateIDCacheImpl(new SimpleString("test"), capacity, new NullStorageManager(), true);
         for (int j = 0; j < capacity; j++) {
            c.addToCache(ids[j], null, true);
         }
         cc[i] = c;
      }
      System.out.println("Press a key after performing an heap dump to stop...");
      System.in.read();
      for (DuplicateIDCache cache : cc) {
         cache.clear();
      }
   }
```
 
` master` is using `280 MB` vs this PR `190 MB`

franz1981:
First results of allocation rate from this PR:
```
Benchmark                                                   (persist)  (size)   Mode  Cnt        Score        Error   Units
DuplicateIDCacheBenchmark.atomicVerify                          false    2000  thrpt   16  5023355.078 ±  36609.582   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate.norm      false    2000  thrpt   16       80.000 ±      0.001    B/op
DuplicateIDCacheBenchmark.atomicVerify:·gc.count                false    2000  thrpt   16          ≈ 0               counts
DuplicateIDCacheBenchmark.atomicVerify                          false   20000  thrpt   16  4401455.813 ± 160390.631   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate.norm      false   20000  thrpt   16       80.000 ±      0.001    B/op
DuplicateIDCacheBenchmark.atomicVerify:·gc.count                false   20000  thrpt   16          ≈ 0               counts
DuplicateIDCacheBenchmark.atomicVerify                           true    2000  thrpt   16  4734734.651 ±  82617.669   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate.norm       true    2000  thrpt   16       80.000 ±      0.001    B/op
DuplicateIDCacheBenchmark.atomicVerify:·gc.count                 true    2000  thrpt   16          ≈ 0               counts
DuplicateIDCacheBenchmark.atomicVerify                           true   20000  thrpt   16  3957606.687 ± 543564.163   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate.norm       true   20000  thrpt   16       80.000 ±      0.001    B/op
DuplicateIDCacheBenchmark.atomicVerify:·gc.count                 true   20000  thrpt   16          ≈ 0               counts
```

franz1981:
I've added a further commit to reduce the allocation rate and improve a bit performance:
```
Benchmark                                                   (persist)  (size)   Mode  Cnt        Score        Error   Units
DuplicateIDCacheBenchmark.atomicVerify                          false    2000  thrpt   16  6421235.726 ±  67792.082   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate.norm      false    2000  thrpt   16       56.000 ±      0.001    B/op
DuplicateIDCacheBenchmark.atomicVerify                          false   20000  thrpt   16  5237169.779 ± 231236.770   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate.norm      false   20000  thrpt   16       56.000 ±      0.001    B/op
DuplicateIDCacheBenchmark.atomicVerify                           true    2000  thrpt   16  5968040.463 ±  82664.101   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate.norm       true    2000  thrpt   16       56.000 ±      0.001    B/op
DuplicateIDCacheBenchmark.atomicVerify                           true   20000  thrpt   16  4898354.808 ± 305894.713   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate.norm       true   20000  thrpt   16       56.000 ±      0.001    B/op
```

I'll keep the commit separated because from code quality is a bit regressed.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/DuplicateIDCacheImpl.java
```diff
@@ -69,11 +93,24 @@ public DuplicateIDCacheImpl(final SimpleString address,
 
       ids = new ArrayList<>(size);
 
+      cachedBoxedInts = boxedInts(size);
```
gtully, line 96: Is this necessary if  -XX:AutoBoxCacheMax=Integer.Max and Integer.valueOf is used?

franz1981: IBM JVMs doesn't had it IIRC. Ideally Integer shouldn't be necessary...
I am preparing a further pr that won't use boxed types ;)

gtully: nice

franz1981: I've implemented another version using a bespoke single-writer concurrent primitive valued hash map but...performance are a bit regressed (`ConcurrentHashMap` is still a nice piece of engineering, but I still need to understand why...)

And memory footprint isn't that great too (because of the small load-factor to save collisions): ~186 MB (that's similar to optimized boxed types)

franz1981: I've removed the last commit including the bespoke primitive map: seems that I cannot beat CHM (not a surprise) due to open-addressing nature of the version I've built. Big sized maps, almost full, work best as closed addressing and that's exactly the use case of CHM.

franz1981:
I've further improved the performance of this by implementing a batch hashCode on `byte[]`, results below:
```
Benchmark                                                   (persist)  (size)   Mode  Cnt        Score        Error   Units
DuplicateIDCacheBenchmark.atomicVerify                          false    2000  thrpt   16  7302759.425 ± 143209.346   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate.norm      false    2000  thrpt   16       56.000 ±      0.001    B/op
DuplicateIDCacheBenchmark.atomicVerify                          false   20000  thrpt   16  6139804.607 ± 293313.296   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate           false   20000  thrpt   16      218.550 ±     10.440  MB/sec
DuplicateIDCacheBenchmark.atomicVerify                           true    2000  thrpt   16  6982708.801 ±  37226.679   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate.norm       true    2000  thrpt   16       56.000 ±      0.001    B/op
DuplicateIDCacheBenchmark.atomicVerify                           true   20000  thrpt   16  5804655.241 ± 317214.721   ops/s
DuplicateIDCacheBenchmark.atomicVerify:·gc.alloc.rate.norm       true   20000  thrpt   16       56.000 ±      0.001    B/op
```

franz1981:
The new hash code implementation should help in other places as well, but probably will send a new PR to address it, FYI
```
Benchmark                                      (length)  (logPermutations)  (seed)  (unsafe)  Mode  Cnt   Score   Error  Units
ByteArrayHashCodeBenchamark.arraysHashCode            7                  4       0      true  avgt   16  10.790 ± 0.089  ns/op
ByteArrayHashCodeBenchamark.arraysHashCode            7                 10       0      true  avgt   16  11.155 ± 0.284  ns/op
ByteArrayHashCodeBenchamark.arraysHashCode            8                  4       0      true  avgt   16  11.501 ± 0.037  ns/op
ByteArrayHashCodeBenchamark.arraysHashCode            8                 10       0      true  avgt   16  11.901 ± 0.380  ns/op
ByteArrayHashCodeBenchamark.arraysHashCode           36                  4       0      true  avgt   16  37.772 ± 0.202  ns/op
ByteArrayHashCodeBenchamark.arraysHashCode           36                 10       0      true  avgt   16  38.580 ± 1.283  ns/op
ByteArrayHashCodeBenchamark.bytesUtilHashCode         7                  4       0      true  avgt   16   8.337 ± 0.088  ns/op
ByteArrayHashCodeBenchamark.bytesUtilHashCode         7                 10       0      true  avgt   16   8.393 ± 0.056  ns/op
ByteArrayHashCodeBenchamark.bytesUtilHashCode         8                  4       0      true  avgt   16   7.066 ± 0.022  ns/op
ByteArrayHashCodeBenchamark.bytesUtilHashCode         8                 10       0      true  avgt   16   7.151 ± 0.096  ns/op
ByteArrayHashCodeBenchamark.bytesUtilHashCode        36                  4       0      true  avgt   16  13.721 ± 0.842  ns/op
ByteArrayHashCodeBenchamark.bytesUtilHashCode        36                 10       0      true  avgt   16  13.437 ± 0.078  ns/op
```
The 8 bytes case is in case duplicate ids are 64 bit sequence ids while the 36 bytes case in case of a random UUID,
It's a lot more faster then the OOTB version of `Arrays::hashCode` (on JDK 8).

franz1981:
Just FYI, the bespoke primitive map impl was https://github.com/apache/activemq-artemis/blob/8d86e19198b431e67fb37ad1580ac7b246bf4a59/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SingleWriterObjIntHashMap.java

clebertsuconic:
@franz1981 / @gtully  I had the impression that the duplicateCache would be broken with Wildcard destinations.. being the cache at the individual address, not the wildcard. Did you check on that?

franz1981:
@clebertsuconic nope, I am just addressing the memory footprint issue, worth a PR to check/fix it in case

franz1981:
@gtully @clebertsuconic 
I see instead other issues re this cache eg what happen if a duplicate cache, configured to not persist, is loading duplicate ids from the journal? 

Handling this case correctly would allow to further improve memory footprint, saving a `!persist` dup cache to use any object-long pair to track recordIDs, given that no record is going to be written.

franz1981:
By separating in memory and persistent ids caches I've further improved the memory footprint of the in-memory case, getting ~140 MB ie ~50 MB less then the persistent case

franz1981:
@clebertsuconic One not: the in-memory id cache load won't do anything because that method isn't supposed to be used, see `master` code on `PostOfficeJournalLoader`:
```java
   @Override
   public void handleDuplicateIds(Map<SimpleString, List<Pair<byte[], Long>>> duplicateIDMap) throws Exception {
      for (Map.Entry<SimpleString, List<Pair<byte[], Long>>> entry : duplicateIDMap.entrySet()) {
         SimpleString address = entry.getKey();

         DuplicateIDCache cache = postOffice.getDuplicateIDCache(address);

         if (configuration.isPersistIDCache()) {
            cache.load(entry.getValue());
         }
      }
   }
```

franz1981:
I've prepared a summary of the improvements of this PR by refactoring the JMH benchmark to keep it as a separate commit to ease comparisons (using https://jmh.morethan.io/ to produce the graphic):
`gc.alloc.rate.norm`:
image
troughput:
image

TLDR:
the allocation rate (normalized) shows a huge inprovement and the troughput is highly improved as well.

franz1981:
CI seems fine with this PR: @clebertsuconic @gtully let me know if we're ready to go and you agree on the changes made :+1: 

IMO there is still a giant room of improvement on cache ids loading during journal load, because it could use the newly introduced `ObjLongPair`, saving `Long` to be used given that `Long` is the record ID and on journal load is never `null`...
but I can send a follow-up PR on this: this one seems already fat enough :)

franz1981:
Just one note for further investigations on this: I'm not sure it's a benchmark issue or not, but probably it isn't.
I see that both master and this new version share a weird GC behaviour with G1GC, see GC behavior on master:
image

It shows a considerable amount of cpu time spent on copy_to_survivor_space that it seems we're getting lot of promotions using the default collector of java 8 ie parallel 

The same behaviour isn't present if I use the primitive Obj-int concurrent map implemented on https://github.com/apache/activemq-artemis/blob/8d86e19198b431e67fb37ad1580ac7b246bf4a59/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SingleWriterObjIntHashMap.java

image

Instead, we get lot of GC compaction activity (probably on the large ObjIntEntry[] and entries insteance backing the primitive map). I would run the same benchs with G1GC to mimic what a running broker would do, but I see that the first case (`master` and this PR using `ConcurrentHashMap`s really have a lot of "hidden" GC activity despit getting an apparent higher throughput advantage.

clebertsuconic:
@franz1981 Franz I need a lecture on the BoxedInt logic thing you did.. I can't put my head around it...

also... isn't there a way to reuse some logic between Persistent and InMemory duplicate cache? they seem a copy paste of one another with added logic to persist.

franz1981:
@clebertsuconic 
> Franz I need a lecture on the BoxedInt logic thing you did.. I can't put my head around it...

Yep, we can have a meet so I can explain the JMH + flamegraphs bits as well, If you like (opened to the rest of community if anyone want to attend)

> isn't there a way to reuse some logic between Persistent and InMemory duplicate cache?

I have tried hard, and I can try even harder too, but I see some benefits from this: is more easy to be understood and the dependencies and data types used clearly reflect the needs eg no storage manager is needed for the in memory case and same for recordID on the ObjLong pair. This both make the code simpler to be understood and more performant & with much less memory footprint.

I can try collect some of the shared functionalities into some static method or abstract class if it's worthy.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PersistentDuplicateIDCache.java
```diff
@@ -0,0 +1,390 @@
...
+import org.apache.activemq.artemis.core.transaction.Transaction;
+import org.apache.activemq.artemis.core.transaction.TransactionOperationAbstract;
+import org.apache.activemq.artemis.utils.ByteUtil;
+import org.jboss.logging.Logger;
+
+import static org.apache.activemq.artemis.api.core.ObjLongPair.NIL;
+import static org.apache.activemq.artemis.core.postoffice.impl.IntegerCache.boxedInts;
+
+/**
+ * A fixed size rotating cache of last X duplicate ids.
```
clebertsuconic, line 43: Can you add a comment here on why we did not use Inheritance and reusability?

I spoke with you on this.. I understand the issue.. but if we document with a Javadoc may prevent others from trying to use a single base class.

clebertsuconic:
@michaelandrepearce there's no new feature on this.. we already had the ability to disable persistence of duplicate cache. Franz just did a refactoring to separate implementations. at least as I understand it. (I'm still slow after the holidays break :) )

michaelandrepearce:
> @michaelandrepearce there's no new feature on this.. we already had the ability to disable persistence of duplicate cache. Franz just did a refactoring to separate implementations. at least as I understand it. (I'm still slow after the holidays break :) )

You're right, its my holiday brain, https://github.com/apache/activemq-artemis/pull/3363/files#diff-fca1c7c5162ab4d798afd50e53bc7ca1e655037ec5032052596ada678f84779fR1318 was what made me think its a new option with the if statement, didn't notice it used to be passed in, and its just refactor.

PR #3370, opened 2020-12-05
ARTEMIS-3021 OOM due to wrong CORE message memory estimation

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3021

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java
@@ -389,6 +389,16 @@ public class ByteUtil {
       return true;
    }
 
+   /**
+    * This ensure a more exact resizing then {@link ByteBuf#ensureWritable(int)}, if needed.<br>
+    * It won't try to trim a large enough buffer.
+    */
+   public static void ensureExactWritable(ByteBuf buffer, int minWritableBytes) {
+      if (buffer.maxFastWritableBytes() < minWritableBytes) {
+         buffer.capacity(buffer.writerIndex() + minWritableBytes);
+      }
+   }
+
 
    /**
     * Returns {@code true} if  the {@link SimpleString} encoded content into {@code bytes} is equals to {@code s},
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
@@ -60,13 +60,15 @@ import org.apache.activemq.artemis.utils.UUID;
 import org.apache.activemq.artemis.utils.collections.TypedProperties;
 import org.jboss.logging.Logger;
 
+import static org.apache.activemq.artemis.utils.ByteUtil.ensureExactWritable;
+
 /** Note: you shouldn't change properties using multi-threads. Change your properties before you can send it to multiple
  *  consumers */
 public class CoreMessage extends RefCountMessage implements ICoreMessage {
 
    public static final int BUFFER_HEADER_SPACE = PacketImpl.PACKET_HEADERS_SIZE;
 
-   private volatile int memoryEstimate = -1;
+   protected volatile int memoryEstimate = -1;
 
    private static final Logger logger = Logger.getLogger(CoreMessage.class);
 
@@ -80,8 +82,6 @@ public class CoreMessage extends RefCountMessage implements ICoreMessage {
    private volatile boolean validBuffer = false;
 
    protected volatile ResetLimitWrappedActiveMQBuffer writableBuffer;
-
-   Object body;
 
    protected int endOfBodyPosition = -1;
 
@@ -445,7 +445,6 @@ public class CoreMessage extends RefCountMessage implements ICoreMessage {
       // with getEncodedBuffer(), otherwise can introduce race condition when delivering concurrently to
       // many subscriptions and bridging to other nodes in a cluster
       synchronized (other) {
-         this.body = other.body;
          this.endOfBodyPosition = other.endOfBodyPosition;
          internalSetMessageID(other.messageID);
          this.address = other.address;
@@ -641,6 +640,15 @@ public class CoreMessage extends RefCountMessage implements ICoreMessage {
    @Override
    public int getMemoryEstimate() {
       if (memoryEstimate == -1) {
+         if (buffer != null && !isLargeMessage()) {
+            if (!validBuffer) {
+               // this can happen if a message is modified
+               // eg clustered messages get additional routing information
+               // that need to be correctly accounted in memory
+               checkEncode();
+            }
+         }
+         final TypedProperties properties = this.properties;
          memoryEstimate = memoryOffset +
             (buffer != null ? buffer.capacity() : 0) +
             (properties != null ? properties.getMemoryOffset() : 0);
@@ -733,11 +741,9 @@ public class CoreMessage extends RefCountMessage implements ICoreMessage {
          endOfBodyPosition = BUFFER_HEADER_SPACE + DataConstants.SIZE_INT;
       }
 
-      buffer.setIndex(0, 0);
-      buffer.writeInt(endOfBodyPosition);
-
+      buffer.setInt(0, endOfBodyPosition);
       // The end of body position
-      buffer.writerIndex(endOfBodyPosition - BUFFER_HEADER_SPACE + DataConstants.SIZE_INT);
+      buffer.setIndex(0, endOfBodyPosition - BUFFER_HEADER_SPACE + DataConstants.SIZE_INT);
 
       encodeHeadersAndProperties(buffer);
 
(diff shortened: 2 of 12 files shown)


franz1981:
I need to check yet how the CI would behave on this and add some tests.
I'll next check if large messages and AMQP are affected the same in the same way

franz1981:
@gemmellr I suppose that the Netty ByteBuf behaviour in case the buffer is enlarged can (and should) be fixed for AMQP as well:
TLDR if you got a 32K msg and write a single byte into it, Netty will enlarge it to be the next power of 2 capacity ie 64K, wasting ~50% of its space in case the message will sit on the queue awaiting it to be sent.

I've "solved" this by forcing Netty to know upfront the desired required capacity and use https://netty.io/4.0/api/io/netty/buffer/ByteBuf.html#capacity-int- to size it upfront.

franz1981:
@gemmellr it seems that AMQP isn't affected by this because a pooled Netty (heap) buffer is used to encode stuff and then copy an exact sized byte[] where needed but I see another issue and is exactly the polled Netty heap buffer.
That stealthy makes the broker memory footprint on heap much larger because Netty would create large byte[] arenas. I suggest to not use pooled heap buffer or it can stealthy cause OOM as well.
This means either:
- use off heap pooled Netty buffers to temporarily encode things 
- use unpooled heap smartly sizes ones to save both the additional copy and pooling as a whole

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
```diff
@@ -622,6 +619,12 @@ private RuntimeException onCheckPropertiesError(Throwable e) {
    @Override
    public int getMemoryEstimate() {
       if (memoryEstimate == -1) {
+         // durable messages can enlarge buffer::capacity while persisted:
+         // better to encode it to properly account its max capacity
+         if (isDurable()) {
+            checkEncode();
```
franz1981, line 625: That means that a durable message would get encoded some time before then expected: @clebertsuconic it's something we can tolerate? 
if not the ideal thing would be to estimate it based on its state, but I'm not sure is feasable.

franz1981:
This is breaking many tests, need to understand what's going on

franz1981:
@clebertsuconic @jbertram 
I see that we've some concurrency issue on `TypedProperties`: 
```
org.apache.activemq.artemis.core.protocol.hornetq.PropertiesConversionTest.testMultiThreadChanges
org.apache.activemq.artemis.tests.compatibility.HornetQSoakTest.testSoakHornetQ on testSoakHornetQ(org.apache.activemq.artemis.tests.compatibility.HornetQSoakTest)
org.apache.activemq.artemis.tests.integration.cluster.failover.ReplicatedMultipleServerFailoverExtraBackupsTest.testStartLiveFirst
```
These tests were failing with my first version of PR that was enabling some strict assertion to check for concurrent modification of `TypedProperty` during `CoreMessage::encode` calls.
I believe the change I've made here to replaced the assert with a debug log is masking a real (rare) issue. wdyt?

franz1981:
I still need to write a test that would show how clustered messages were getting a too wrong estimation without this change.

franz1981:
the CI seems fine with this, so I'm just going to prepare a test for the specific change and we're ready to go

franz1981:
Seems that writing a proper reproducer for this is hard: I can reproduce it by hand by using a 2 nodes cluster. (upstream and downstream), a disconnected durable subscriber on the downstream broker, 40K messages and a producer that keep running vs the upstream one until getting the downstream broker to go OOM. On this scenario a dump on the upstream broker shows 64KiB buffer sized core messages ie the next available power of 2 sized one.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
```diff
@@ -426,7 +426,6 @@ protected CoreMessage(CoreMessage other, TypedProperties copyProperties) {
       // with getEncodedBuffer(), otherwise can introduce race condition when delivering concurrently to
       // many subscriptions and bridging to other nodes in a cluster
       synchronized (other) {
-         this.body = other.body;
```
michaelandrepearce, line 429: I might be a plank, but surely this avoids using up memory as the body object is copied by reference here, rather than later needing to deserialise again and creating a new object in memory.

michaelandrepearce: as well as a perf improvement as serialisation is a time burn.

franz1981: That body param seems was unused, I just clean it up as dead code

franz1981:
This need some time to be ready, but it's worth to be fixed IMO, so I'll leave it opened as draft

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java
```diff
@@ -389,6 +389,16 @@ private static boolean equalsUnsafe(final byte[] left,
       return true;
    }
 
+   /**
+    * This ensure a more exact resizing then {@link ByteBuf#ensureWritable(int), if needed.<br>
```
gtully, line 393: this seems problematic:
/home/runner/work/activemq-artemis/activemq-artemis/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java:393: error: unexpected text after parenthesis
Error:      * This ensure a more exact resizing then {@link ByteBuf#ensureWritable(int), if needed.<br>
Error:                                               ^

franz1981:
I see that we change copied message address while moving messages and that is causing re-encoding: it would be better to save it to happen as a whole...adding a commit to address it

franz1981:
CI was green but I would like to add a benchmark + a test to verify it won't regress in the future

franz1981:
There's a real failure on test `org.apache.activemq.artemis.tests.compatibility.HornetQSoakTest.testSoakHornetQ on testSoakHornetQ(org.apache.activemq.artemis.tests.compatibility.HornetQSoakTest)`:

```
[Thread-22 (ActiveMQ-server-org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl$6@48247149)] 11:30:12,040 WARN  [org.apache.activemq.artemis.core.server] AMQ222151: removing consumer which did not handle a message, consumer=ServerConsumerImpl [id=0, filter=null, binding=LocalQueueBinding [address=topic, queue=QueueImpl[name=ARTEMIS8.test, postOffice=PostOfficeImpl [server=ActiveMQServerImpl::name=localhost], temp=false]@cd27f06, filter=null, name=ARTEMIS8.test, clusterName=ARTEMIS8.test77120279-46fa-11ec-a1ac-ecf4bbced624]], message=Reference[397]:RELIABLE:CoreMessage[messageID=397,durable=true,userID=78923a2e-46fa-11ec-a1ac-ecf4bbced624,priority=4, timestamp=Tue Nov 16 11:30:12 EST 2021,expiration=0, durable=true, address=jms.topic.topic,size=282,properties=TypedProperties[__AMQ_CID=78764db3-46fa-11ec-a1ac-ecf4bbced624,_AMQ_ROUTING_TYPE=0]]@350453798: java.lang.AssertionError: Bad Headers encode size estimation
	at org.apache.activemq.artemis.core.message.impl.CoreMessage.encodeHeadersAndProperties(CoreMessage.java:785)
	at org.apache.activemq.artemis.core.message.impl.CoreMessage.encode(CoreMessage.java:748)
	at org.apache.activemq.artemis.core.message.impl.CoreMessage.checkEncode(CoreMessage.java:354)
	at org.apache.activemq.artemis.core.message.impl.CoreMessage.getEncodeSize(CoreMessage.java:828)
	at org.apache.activemq.artemis.core.protocol.core.impl.wireformat.SessionReceiveMessage.expectedEncodeSize(SessionReceiveMessage.java:54)
	at org.apache.activemq.artemis.core.protocol.core.impl.PacketImpl.createPacket(PacketImpl.java:356)
	at org.apache.activemq.artemis.core.protocol.core.impl.PacketImpl.encode(PacketImpl.java:327)
	at org.apache.activemq.artemis.core.protocol.core.impl.ChannelImpl.send(ChannelImpl.java:379)
	at org.apache.activemq.artemis.core.protocol.core.impl.ChannelImpl.sendBatched(ChannelImpl.java:333)
	at org.apache.activemq.artemis.core.protocol.core.impl.CoreSessionCallback.sendMessage(CoreSessionCallback.java:128)
	at org.apache.activemq.artemis.core.server.impl.ServerConsumerImpl.deliverStandardMessage(ServerConsumerImpl.java:1165)
	at org.apache.activemq.artemis.core.server.impl.ServerConsumerImpl.proceedDeliver(ServerConsumerImpl.java:510)
	at org.apache.activemq.artemis.core.server.impl.QueueImpl.proceedDeliver(QueueImpl.java:3840)
	at org.apache.activemq.artemis.core.server.impl.QueueImpl.deliver(QueueImpl.java:3782)
	at org.apache.activemq.artemis.core.server.impl.QueueImpl.deliverDirect(QueueImpl.java:3741)
	at org.apache.activemq.artemis.core.server.impl.QueueImpl.addTail(QueueImpl.java:1251)
	at org.apache.activemq.artemis.core.server.impl.RoutingContextImpl.internalprocessReferences(RoutingContextImpl.java:197)
	at org.apache.activemq.artemis.core.server.impl.RoutingContextImpl.processReferences(RoutingContextImpl.java:192)
	at org.apache.activemq.artemis.core.postoffice.impl.PostOfficeImpl$1.done(PostOfficeImpl.java:1563)
	at org.apache.activemq.artemis.core.persistence.impl.journal.OperationContextImpl$1.run(OperationContextImpl.java:284)
	at org.apache.activemq.artemis.utils.actors.OrderedExecutor.doTask(OrderedExecutor.java:42)
	at org.apache.activemq.artemis.utils.actors.OrderedExecutor.doTask(OrderedExecutor.java:31)
	at org.apache.activemq.artemis.utils.actors.ProcessorBase.executePendingTasks(ProcessorBase.java:65)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at org.apache.activemq.artemis.utils.ActiveMQThreadFactory$1.run(ActiveMQThreadFactory.java:118)
	```
I see that using `assert` for invariant has some value :P @gtully

**Comment by franz1981** [comment]
`org.apache.activemq.artemis.tests.compatibility.HornetQSoakTest.testSoakHornetQ` is now fixed: waiting if CI is green and we're ready to go :)

File: artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java [hunk 4]
```diff
@@ -748,21 +754,42 @@ public synchronized CoreMessage encode() {
...
+      final int headersSize =
+         Long.BYTES +                                    // messageID
+         addressEncodedBytes +                           // address
+         userIDEncodedSize +                             // userID
+         Byte.BYTES +                                    // type
+         Byte.BYTES +                                    // durable
+         Long.BYTES +                                    // expiration
+         Long.BYTES +                                    // timestamp
+         Byte.BYTES;                                     // priority
+      synchronized (properties) {
```
franz1981, line 772: @clebertsuconic this is needed because there are tests/cases where `properties`'s content can change concurrently and the presized underlying  buffer using `properties::getEncodeSize` is no longer valid.

I think we shouldn't allow `properties` to change concurrently while encoding is happening: is it a valid use case?

franz1981:
Given that CI is green and I got reviewed by @clebertsuconic about it, I think this is ready to go...merging soon today if none complain :+1:

PR #3372, opened 2020-12-07
NO-JIRA: Correct misspellings in documentation

uroesch:
Correct a few words with misspellings in the
documentation.


clebertsuconic:
Awesome.. thanks a lot!

PR #3373, opened 2020-12-07
ARTEMIS-3023 Avoid opening AMQP Large Message for final parsing

clebertsuconic:


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
```diff
@@ -362,6 +348,62 @@ public void addBytes(ReadableBuffer data) throws Exception {
...
+            parsingBuffer = new CompositeReadableBuffer();
+         }
+         parsingBuffer.append(copy);
+         if (!initialHeader) {
+            genericPargeLargeMessage();
+         }
+      }
+   }
+
+   private void genericPargeLargeMessage() {
```
tabish121, line 383: Thinking there might be a spelling error here

clebertsuconic: yeah.. thanks

clebertsuconic: fixed

PR #3374, opened 2020-12-08
ARTEMIS-3024 Expiry or DLQ Retry not working with AMQP

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2140,6 +2140,7 @@ private int iterQueue(final int flushLimit,
                                       QueueIterateAction messageAction) throws Exception {
       int count = 0;
       int txCount = 0;
+      Integer expectedHits = messageAction.expectedHits();
```
michaelandrepearce, line 2143: Whats the purpose of this?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Queue.java
```diff
@@ -361,6 +361,10 @@ int moveReferences(int flushLimit,
 
    int retryMessages(Filter filter) throws Exception;
 
+   default int retryMessages(Filter filter, Integer expectedHits) throws Exception {
```
michaelandrepearce, line 364: Why default method? Its been implemented in queueimpl

clebertsuconic: there are a bunch of Fakes that don't require implementing anything.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
clebertsuconic: QueueControl.expire with a messageID will only expire a single record, however the iteration will go until the end. it's sort of an optimization.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Queue.java
clebertsuconic: @michaelandrepearce I could go without the default.. but having the default here minimized the number of changes on Fakes.. etc... It was just for simplicity and minimize the number of lines changed.

PR #3377, opened 2020-12-10
ARTEMIS-3026 Allow "re-encode" of amqp large messages

clebertsuconic:
notice the quotes on "re-encode", as this is just replacing the set of application properties, properties and headers by a new set
if a flag reEncoded is set to true on AMQPLargeMessage

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
@@ -87,6 +87,8 @@ public class AMQPLargeMessage extends AMQPMessage implements LargeServerMessage 
       }
    }
 
+   private boolean reencoded = false;
+
    /**
     * AMQPLargeMessagePersister will save the buffer here.
     * */
@@ -622,7 +624,15 @@ public class AMQPLargeMessage extends AMQPMessage implements LargeServerMessage 
 
    @Override
    public void reencode() {
+      reencoded = true;
+   }
 
+   public void setReencoded(boolean reencoded) {
+      this.reencoded = reencoded;
+   }
+
+   public boolean isReencoded() {
+      return reencoded;
    }
 
    @Override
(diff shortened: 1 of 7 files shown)


clebertsuconic:
@jbertram this would allow you to use re-encode on large messages at your new feature.

Notice there will be an additional cost to deliver large messages (encoding the application properties again), but this is in my opinion an expected cost.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
```diff
@@ -622,7 +624,15 @@ public long getPersistentSize() throws ActiveMQException {
 
    @Override
    public void reencode() {
+      reEncoded = true;
+   }
+
+   public void setReEncoded(boolean reEncoded) {
+      this.reEncoded = reEncoded;
```
michaelandrepearce, line 631: Nit. Why is the E capitalised. Reencoded is a single word. https://en.m.wiktionary.org/wiki/reencoded

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
```diff
@@ -622,7 +624,15 @@ public long getPersistentSize() throws ActiveMQException {
 
    @Override
    public void reencode() {
+      reEncoded = true;
```
michaelandrepearce, line 627: Why isnt the reencoding happening immediately like for other messages and expected behaviour, e.g. other areas of code and plugins are expecting the message to rencoded by calling this without doing extra work.

clebertsuconic: It would be fairly difficult to re-encode the message on the large message file.
For AMQP Large messages, I'm instead just flagging the message as re-encoded, and I would replace the stream during the delivery.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
clebertsuconic: right.. fixed it. thanks

PR #3383, opened 2020-12-16
ARTEMIS-3034 CLI Transfer Option

clebertsuconic:


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
```diff
@@ -21,28 +21,18 @@
...
-import org.apache.activemq.artemis.utils.ConfigurationHelper;
-import org.apache.activemq.artemis.utils.uri.SchemaConstants;
 import org.apache.qpid.jms.JmsConnectionFactory;
 
 public class ConnectionAbstract extends InputAbstract {
 
    private static final String DEFAULT_BROKER_URL = "tcp://localhost:61616";
 
-   @Option(name = "--url", description = "URL towards the broker. (default: tcp://localhost:61616)")
+   @Option(name = "--source-url", description = "URL towards the broker. (default: Read from current broker.xml or tcp://localhost:61616 if the default cannot be parsed)")
```
brusdev, line 35: @clebertsuconic is it possible to revert the option name to `--url`?

clebertsuconic: Thank you.   This was an accident. I didn't mean to.

PR #3386, opened 2020-12-21
ARTEMIS-3040 Fix reloading large messages paged before 2.12

brusdev:


franz1981:
Not sure but probably LargeMessageType should be changed according, well done!

franz1981:
I would reference ARTEMIS-2676 as the cause of this :)
Well done Dom!

franz1981 (approved):

PR #3387, opened 2020-12-21
ARTEMIS-3027 Fixing AMQP persister encoding

brusdev:


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessagePersisterV2.java
```diff
@@ -71,8 +70,6 @@ public void encode(ActiveMQBuffer buffer, Message record) {
          buffer.writeInt(properties.getEncodeSize());
          properties.encode(buffer.byteBuf());
       }
-
```
clebertsuconic, line 74: why did you need a new persister? you could just check if you have more bytes on the ByteBuffer.

brusdev: @clebertsuconic I also would avoid a new persister but the trailer of the paged messages has no fixed size[1].
So I can't figure out how to check the remaining bytes to understand if it includes or not the expiration. Any suggestion?

[1] https://github.com/apache/activemq-artemis/blob/2.16.0/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagedMessageImpl.java#L225

PR #3388, opened 2020-12-22
ARTEMIS-3041 Allow encrypted data source property values

brusdev:


fvaleri:
It would probably make sense to do this only for the password key, as the documentation says: "Apache ActiveMQ Artemis can be configured to use 'masked' passwords in its configuration files."

brusdev:
Many data source properties contain sensitive values, like the username and the value encryption for the data source properties is optional so I would not limit what properties the user can encrypt.

fvaleri:
> Many data source properties contain sensitive values, like the username and the value encryption for the data source properties is optional so I would not limit what properties the user can encrypt.

Agree, but then we should also update the doc, adding a note about this.

clebertsuconic:
@brusdev / @fvaleri I will merge this now.. feel free to follow up with another PR for doc changes if you guys decide to improve docs on this.

brusdev:
@fvaleri I have just created the #3391 to improve the doc about JDBC connection pooling.

fvaleri:
Thanks @brusdev , that's great.

PR #3393, opened 2021-01-04
ARTEMIS-3049 Reduce live page lookup cost

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3049

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
@@ -468,16 +468,11 @@ public class PagingStoreImpl implements PagingStore {
       Page page = createPage(pageId);
       page.open();
 
-      List<PagedMessage> messages = page.read(storageManager);
+      final List<PagedMessage> messages = page.read(storageManager);
 
-      LivePageCache pageCache = new LivePageCacheImpl(pageId);
+      final PagedMessage[] initialMessages = messages.toArray(new PagedMessage[messages.size()]);
 
-      for (PagedMessage msg : messages) {
-         pageCache.addLiveMessage(msg);
-         // As we add back to the live page,
-         // we have to discount one when we read the page
-         msg.getMessage().usageDown();
-      }
+      final LivePageCache pageCache = new LivePageCacheImpl(pageId, initialMessages);
 
       page.setLiveCache(pageCache);
 
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
@@ -480,7 +480,7 @@ public class PagingStoreImpl implements PagingStore {
 
       currentPage = page;
 
-      cursorProvider.addPageCache(pageCache);
+      cursorProvider.addLivePageCache(pageCache);
 
       /**
        * The page file might be incomplete in the cases: 1) last message incomplete 2) disk damaged.
@@ -1110,7 +1110,7 @@ public class PagingStoreImpl implements PagingStore {
 
          newPage.setLiveCache(pageCache);
 
-         cursorProvider.addPageCache(pageCache);
+         cursorProvider.addLivePageCache(pageCache);
 
          currentPageSize = 0;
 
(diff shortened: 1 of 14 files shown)


franz1981:
@qihongxu Hi!! 
I've tried to address/improve what you've shown in your comment on https://github.com/apache/activemq-artemis/pull/2494#issuecomment-455086939
Any chances you can test it with that same original test?
I'm going to provide a micro-bench just on the append only list, but only a end-2-end test can shows if the improvements are effective as expected

franz1981:
Despite my expectations, the results are NOT promising...

This PR
1 thread:
```
Benchmark                                            (chunkSize)  (size)   Mode  Cnt          Score         Error  Units
ConcurrentAppendOnlyChunkListLookupBenchmark.lookup           32    1000  thrpt   16  108968555.072 ±  135438.271  ops/s
ConcurrentAppendOnlyChunkListLookupBenchmark.lookup           32   10000  thrpt   16  105613840.728 ±  475604.647  ops/s
```
3 threads:
```
Benchmark                                            (chunkSize)  (size)   Mode  Cnt         Score          Error  Units
ConcurrentAppendOnlyChunkListLookupBenchmark.lookup           32    1000  thrpt   16  91693844.669 ± 29399345.167  ops/s
ConcurrentAppendOnlyChunkListLookupBenchmark.lookup           32   10000  thrpt   16  78520201.713 ±  3428526.469  ops/s
```

`master`

1 thread:
```
Benchmark                                            (chunkSize)  (size)   Mode  Cnt         Score         Error  Units
ConcurrentAppendOnlyChunkListLookupBenchmark.lookup           32    1000  thrpt   16  59747234.340 ± 1117878.317  ops/s
ConcurrentAppendOnlyChunkListLookupBenchmark.lookup           32   10000  thrpt   16   7699916.690 ±   36448.254  ops/s
```
3 threads:
```
Benchmark                                            (chunkSize)  (size)   Mode  Cnt          Score         Error  Units
ConcurrentAppendOnlyChunkListLookupBenchmark.lookup           32    1000  thrpt   16  181201564.143 ± 2457478.235  ops/s
ConcurrentAppendOnlyChunkListLookupBenchmark.lookup           32   10000  thrpt   16   23011863.725 ±  110276.468  ops/s
```

It shows that 3 consumers won't get the expected speedup and I suppose that's due to the last accessed buffer miss.

franz1981:
I've added @Contended to some fields of the list + `--jvmArgs="-XX:-RestrictContended"` and now it seems to work better: it seems that the modifications on cached last accessed buffer have a false sharing with lastIndex accesses.

franz1981:
I've chosen to remove the caching of the last accessed buffer because multiple consumers would scale negatively due to this and instead, let a reloaded live cache to benefit from O(1) accesses instead, while getting a more compacted data representation instead.

The PR is providing a JMH bench to track perf regression on lookup scaling.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
```diff
@@ -468,16 +468,11 @@ protected void reloadLivePage(int pageId) throws Exception {
...
-      LivePageCache pageCache = new LivePageCacheImpl(pageId);
+      final PagedMessage[] initialMessages = messages.toArray(new PagedMessage[messages.size()]);
 
-      for (PagedMessage msg : messages) {
-         pageCache.addLiveMessage(msg);
-         // As we add back to the live page,
-         // we have to discount one when we read the page
-         msg.getMessage().usageDown();
-      }
+      final LivePageCache pageCache = new LivePageCacheImpl(pageId, initialMessages);
```
franz1981, line 475: @clebertsuconic 
the original `addLiveMessage` was calling `usageUp` right before `usageDown`: it means that they can both be saved as the ref count will remain the same, correct?

michaelandrepearce:
> I've added `@Contended` to some fields of the list + `--jvmArgs="-XX:-RestrictContended"` and now it seems to work better: it seems that the modifications on cached last accessed buffer have a false sharing with lastIndex accesses.
> 

Im not seeing this flag being altered/added at all in the commits?

franz1981:
> Im not seeing this flag being altered/added at all in the commits?

Correct: I have used it in a run with JMH while I was trying to improve the page lookup cost of the chunked list, but I have decided that such change was too invasive, see https://github.com/apache/activemq-artemis/pull/3393#issuecomment-754085723

franz1981:
The PR should be in a good fit, any other concern?

clebertsuconic:
I'm running a complete test suite on a rebased branch.. if everything is fine I will merge it.

PR #3394, opened 2021-01-04
do not override 'java.io.tmpdir' if set

vhdirk:
Fixes https://issues.apache.org/jira/browse/ARTEMIS-2026

### DIFF
--- a/artemis-boot/src/main/java/org/apache/activemq/artemis/boot/Artemis.java
+++ b/artemis-boot/src/main/java/org/apache/activemq/artemis/boot/Artemis.java
@@ -112,7 +112,7 @@ public class Artemis {
          }
       }
 
-      if (fileInstance != null) {
+      if (System.getProperty("java.io.tmpdir") == null && fileInstance != null) {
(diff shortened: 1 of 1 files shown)


artemis-boot/src/main/java/org/apache/activemq/artemis/boot/Artemis.java
```diff
@@ -112,7 +112,7 @@ public int compare(File file, File file1) {
          }
       }
 
-      if (fileInstance != null) {
+      if (System.getProperty("java.io.tmpdir") == null) {
```
jbertram, line 115: I think the `fileInstance != null` should still be there as it's still valid.

vhdirk: Ah, yes, you're right. Updated.

brusdev:
The commit message isn't formatted as described at https://github.com/apache/activemq-artemis/blob/master/docs/hacking-guide/en/code.md#typical-development-cycle

PR #3396, opened 2021-01-04
ARTEMIS-3031 add new callback for broker deactivation

jbertram:
The existing deactivation callback happens *after* several important
services are shutdown (e.g. the remoting service which allows client
connectivity). This commit adds a new callback which is invoked *before*
any services are stopped. This is useful for embedded use-cases where
applications want to stop gracefully before any part of the broker is
stopped.

A default, empty method implementation is provided so that existing
callback implementations don't need to change.


franz1981 (approved):
LGTM very useful!

PR #3397, opened 2021-01-05
ARTEMIS-3050 Reduce PagedReferenceImpl memory footprint

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3050

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/PagedReferenceImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/PagedReferenceImpl.java
@@ -30,10 +30,9 @@ import org.apache.activemq.artemis.core.server.Queue;
 import org.apache.activemq.artemis.core.server.ServerConsumer;
 import org.apache.activemq.artemis.core.server.impl.AckReason;
 import org.apache.activemq.artemis.core.transaction.Transaction;
-import org.apache.activemq.artemis.utils.collections.LinkedListImpl;
 import org.jboss.logging.Logger;
 
-public class PagedReferenceImpl extends LinkedListImpl.Node<PagedReferenceImpl> implements PagedReference, Runnable {
+public class PagedReferenceImpl implements PagedReference, Runnable {
 
    private static final Logger logger = Logger.getLogger(PagedReferenceImpl.class);
 


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/PagedReferenceImpl.java
```diff
@@ -30,10 +30,9 @@
 import org.apache.activemq.artemis.core.server.ServerConsumer;
 import org.apache.activemq.artemis.core.server.impl.AckReason;
 import org.apache.activemq.artemis.core.transaction.Transaction;
-import org.apache.activemq.artemis.utils.collections.LinkedListImpl;
 import org.jboss.logging.Logger;
 
-public class PagedReferenceImpl extends LinkedListImpl.Node<PagedReferenceImpl> implements PagedReference, Runnable {
```
clebertsuconic, line 36: This was an optimization you performed some time ago, to speed up things on the delivering... Queue.references...

Are you sure about this change? Did you validate perf on it?

franz1981: Yes but i was young :)
I didn't know that pages references are never placed on the queue, hence they don't need to extends node

michaelandrepearce: <= He had hair back then and care free:P

franz1981:
This one should be safe to be merged

clebertsuconic:
lets postpone this PR.. and think about the WeakREference on PagedReference

perhaps this will be ok, as the messages will be sent through the LInkedList on Queue in certain occasions. I guess if we make the WeakReference on PagedReference a strong reference, it willl probably be ok.

clebertsuconic:
postponing after the next release... as discussed.

michaelandrepearce:
@clebertsuconic  / @franz1981  whats occuring with this, wanting to start clearing down old / stagnant PR's a bit of spring cleaning so to say - so we focus on stuff thats actively working on / relevant .

franz1981:
@michaelandrepearce I was waiting the comment of @clebertsuconic 
```
postponing after the next release... as discussed.
```
So don't know yet what are the replacement/proposal to this yet

jbertram:
I'm converting this to a draft and removing the `DO-NOT-MERGE-YET` label. Using the draft functionality is a bit more elegant. That said, it would be great to either get this merged or just close it. What do you think @clebertsuconic, @franz1981?

jbertram:
@clebertsuconic, what do you think about this PR now that you've eliminated the soft cache? Should we merge it or close it?

PR #3398, opened 2021-01-05
ARTEMIS-3051 Fix MessageReferenceImpl::getMemoryEstimate

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3051

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/MessageReferenceImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/MessageReferenceImpl.java
@@ -86,6 +86,8 @@ public class MessageReferenceImpl extends LinkedListImpl.Node<MessageReferenceIm
 
    // Static --------------------------------------------------------
 
+   // This value has been computed by using https://github.com/openjdk/jol
+   // on HotSpot 64-bit VM COOPS, 8-byte alignment
    private static final int memoryOffset = 72;
 
    // Constructors --------------------------------------------------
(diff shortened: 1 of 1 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/MessageReferenceImpl.java
```diff
@@ -86,7 +86,7 @@ public int compare(MessageReference o1, MessageReference o2) {
 
    // Static --------------------------------------------------------
 
-   private static final int memoryOffset = 64;
+   private static final int memoryOffset = 72;
```
michaelandrepearce, line 89: Can we have some doc or something around this very magical number (its magical without any context), how do we derive 72?

franz1981: Good point, we have it already in other parts of the source code (see TypedProperties)

michaelandrepearce: Should it be a static constant then, where the constant is kept close to where if something changes the constant is. And a small javadoc regarding how it's calculated so that should something change people know how to recalc to update it

franz1981: > Should it be a static constant then, where the constant is kept close to where if something changes the constant is

The issue with TypedProperty is much subtle and complex to be fixed: I'm going to send a separate PR for that one

> And a small javadoc regarding how it's calculated so that should something change people know how to recalc to update it

+1, that's what I'm going to add

PR #3401, opened 2021-01-08
ARTEMIS-3058 improper AddressSettings merge

jbertram:
The merge method in AddressSettings should *not* use any getters. It
should reference the relevant variables directly. Using any getters will
return default values in the underlying value is null. This can cause
problems for hierarchical settings.

Also fixed a few potential NPEs exposed by the test-case.


michaelandrepearce:
nice find. LGTM

PR #3403, opened 2021-01-11
ARTEMIS-3059 AMQP message reencoding should save creating Netty heap arenas

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3059


franz1981:
@gemmellr and @clebertsuconic need your review here: I see that we have several parts on the broker that are using pooled heap buffers...

franz1981:
It doesn't seem enough to save Netty heap arenas to be allocated, see
Screenshot from 2021-01-11 14-17-08

gemmellr:
I dont really know much about the brokers pooled netty buffer usages, most of my usage (elsewhere) avoids them. As you say there do look to be others uses besides the one changed here.

franz1981:
@gemmellr it seems that Netty pooled heap arenas are still allocated, but are empty (that's whye they have such small footprint on the screenshot).
This PR is delivering the improvement I was seeking.

> I dont really know much about the brokers pooled netty buffer usages, most of my usage (elsewhere) avoids them. As you say there do look to be others uses besides the one changed here.

I see that `AMQPMessage::createDeliveryCopy` is using it, but probably @clebertsuconic knows about it

clebertsuconic:
@franz1981 we should probably review and manage those PooledBuffers usage as whole?

PR #3404, opened 2021-01-11
ARTEMIS-3061 AMQPMessage::getDuplicateProperty can save key comparisons and class checks

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3061

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
@@ -1018,7 +1018,7 @@ public abstract class AMQPMessage extends RefCountMessage implements org.apache.
             return null;
          }
       }
-      return getObjectProperty(org.apache.activemq.artemis.api.core.Message.HDR_DUPLICATE_DETECTION_ID);
+      return getApplicationObjectProperty(org.apache.activemq.artemis.api.core.Message.HDR_DUPLICATE_DETECTION_ID.toString());
    }
 
    @Override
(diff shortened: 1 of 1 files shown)


franz1981:
@gemmellr 
Just a qq about this:

```java
         if (value instanceof UnsignedInteger ||
            value instanceof UnsignedByte ||
            value instanceof UnsignedLong ||
            value instanceof UnsignedShort) {
            return ((Number) value).longValue();
         }
```
I've improved things a bit on this PR, but still...is it necessary?

tabish121:
> @gemmellr
> Just a qq about this:
> 
> ```java
>          if (value instanceof UnsignedInteger ||
>             value instanceof UnsignedByte ||
>             value instanceof UnsignedLong ||
>             value instanceof UnsignedShort) {
>             return ((Number) value).longValue();
>          }
> ```
> 
> I've improved things a bit on this PR, but still...is it necessary?

Unless the calling code is able / willing to deal with AMQP primitive types like these unsigned values then yes you still need it.  It was added for that reason as these types escaped into the AMQP to Core conversion stuff and it didn't deal with them correctly leading to issues.  They are valid types for ApplicationProperties so they need to be accounted for in this context.

gemmellr:
I'd presume so. The code calling it either needs to handle those types, or as its doing convert it to something else if it doesnt.

gemmellr:
Yeah, what Tim said lol.

franz1981:
Thanks guys, so let me know if the change I did seems fine, I am just trying to separate the slow/fast path ms save some strong comparison and class check

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1382,20 +1382,26 @@ public final Object getObjectProperty(String key) {
...
-         }
+         return getApplicationObjectProperty(key);
       }
 
       return null;
    }
 
+   private Object getApplicationObjectProperty(String key) {
+      Object value = getApplicationPropertiesMap(false).get(key);
+      if (value instanceof Number) {
```
clebertsuconic, line 1393: Can't we just do the ultimate simplification? I know this would be a semantic change but looking at the usages I only see this being used on filter.

I think it would be safe to change this to:

```java
      if (value instanceof Number) {
             return ((Number) value).longValue();
       }
       return value;
```

michaelandrepearce: What if number is float or double

clebertsuconic: hmmm.. yeah.. you're right.. ok!

PR #3405, opened 2021-01-12
ARTEMIS-3063 AMQPMessage get property can use a String switch

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3063


franz1981:
This one is an easy one: using `switch` enable a nice `String` optimization allowing `javac` to use  a `lookupswitch` that has a log(n) cost (with base 2) to lookup at the correct hash code (and I've verified that the hashcode of the cases in the switch have no collisions) + O(n) - n the smaller string to compare to check the specific string equality.

Now the thing is: the binary search used by `lookupswitch` is fast enough? 
I would say yes, but probably only a JMH bench can help here.

PR #3406, opened 2021-01-12
ARTEMIS-3065 AMQP Anonymous producer would eventually block

clebertsuconic:


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/logger/ActiveMQAMQPProtocolLogger.java
```diff
@@ -66,4 +66,9 @@
       "\nSuccess on Server AMQP Connection {0} on {1} after {2} retries" +
       "\n*******************************************************************************************************************************\n", format = Message.Format.MESSAGE_FORMAT)
    void successReconnect(String name, String hostAndPort, int currentRetry);
+
+   @LogMessage(level = Logger.Level.WARN)
+   @Message(id = 111004, value = "An anonymous producer is sending messages on destination {0} and {1} while their AddressFullPolicies are clashing with each other. This could lead to inconsistencies between blocking and paging at your client sender. Notice you could have other occurencies of this scenario but this message is printed only once per anonymous client sender.",
```
clebertsuconic, line 71: I'm looking for feedback on this error message here. any edit to make it more understandable would be appreciated.

michaelandrepearce:
I assume this either can occur in core and openwire also? Or is it already solved there? If already solved can you point us to how so can compare solutions in review. If it isnt shouldnt the solution  be generic for all protocols? Not just amqp

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/paging/AmqpPagingTest.java
```diff
@@ -118,4 +128,49 @@ public void testPaging() throws Exception {
...
+      final String data = builder.toString();
+      final int MSG_COUNT = 3_000;
+
+      // sending size to explod max size
+      server.getPagingManager().addSize((int) ((PagingManagerImpl) server.getPagingManager()).getMaxSize());
+      server.getPagingManager().addSize(100_000);
+
+      server.getAddressSettingsRepository().addMatch("blockedQueue", new AddressSettings().setAddressFullMessagePolicy(AddressFullMessagePolicy.BLOCK));
+
+      ConnectionFactory factory = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:5672");
```
michaelandrepearce, line 148: Can we make this test more generic as its already using just jms api and test the other protocols switching the connection factories to test them for this problem as we do in some other tests.

clebertsuconic: this is a problem exclusive to AMQP. in AMQP the credit is per link, while in openwork and core it's per address..

Do you mean to run the same test with other protocols? is that what you asked?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/logger/ActiveMQAMQPProtocolLogger.java
clebertsuconic: @michaelandrepearce as I have your attention, can you review this error message? or is it fine as is?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/paging/AmqpPagingTest.java
michaelandrepearce: Yes. Just to be sure.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/logger/ActiveMQAMQPProtocolLogger.java
michaelandrepearce: It made sense to me when i read it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/paging/AmqpPagingTest.java
clebertsuconic: ok, will do later today.. thanks for the review

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/logger/ActiveMQAMQPProtocolLogger.java
```diff
@@ -66,4 +66,9 @@
...
+      format = Message.Format.MESSAGE_FORMAT)
+   void incompatibleAddressSettingsOnAnonymousProducer(String oldAddress, String newAddress);
```
gemmellr, line 73: The method name mentions  'incompatibleAddressSettings' but its specific to AddressFullPolicy so perhaps it should refer to only AddressFullPolicy rather than Settings.

Also its called for any difference, but some differences aren't really incompatible/problematic, really the main issue is between BLOCK and anything else. Maybe it shouldnt always be called as it is now, or should be named 'different..' rather than 'incompatible..'.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
```diff
@@ -184,6 +193,19 @@ protected void actualDelivery(AMQPMessage message, Delivery delivery, Receiver r
       }
    }
 
+   private void validateAddressOnAnymousLink(AMQPMessage message) throws Exception {
+      SimpleString newAddress = message.getAddressSimpleString();
+      if (newAddress != null && !newAddress.equals(lastAddress)) {
+         PagingStore newPageStore = sessionSPI.getProtocolManager().getServer().getPagingManager().getPageStore(newAddress);
+         if (!addressAlreadyClashed && lastPageStore != null && lastPageStore.getAddressFullMessagePolicy() != newPageStore.getAddressFullMessagePolicy()) {
+            addressAlreadyClashed = true; // print the warning only once
```
gemmellr, line 201: Nitpicking, but its the policies differing that was important rather than the address not being the same, so maybe addressFullPolicyAlreadyDiffered?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
```diff
@@ -174,6 +179,10 @@ private RoutingType getRoutingType(Symbol[] symbols, SimpleString address) {
    protected void actualDelivery(AMQPMessage message, Delivery delivery, Receiver receiver, Transaction tx) {
       try {
          if (sessionSPI != null) {
+            // message could be null on unit tests (Mocking from ProtonServerReceiverContextTest).
+            if (address == null && message != null) {
+               validateAddressOnAnymousLink(message);
```
gemmellr, line 184: Typo in method name, Anonymous.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
```diff
@@ -184,6 +193,19 @@ protected void actualDelivery(AMQPMessage message, Delivery delivery, Receiver r
       }
    }
 
+   private void validateAddressOnAnymousLink(AMQPMessage message) throws Exception {
+      SimpleString newAddress = message.getAddressSimpleString();
+      if (newAddress != null && !newAddress.equals(lastAddress)) {
+         PagingStore newPageStore = sessionSPI.getProtocolManager().getServer().getPagingManager().getPageStore(newAddress);
+         if (!addressAlreadyClashed && lastPageStore != null && lastPageStore.getAddressFullMessagePolicy() != newPageStore.getAddressFullMessagePolicy()) {
```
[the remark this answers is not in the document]
clebertsuconic: I thought about eventually extending the check into some of the sizes.. that's why I kept PageStore originally.

As the idea evolved I agree I don't need the pageStore any more.. I will take it out.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
clebertsuconic: I'm storing the addresses so we can print them properly in case there's a clash. An architect would have a lot of trouble finding the proper producer without knowing the name where the issue occurred.

clebertsuconic:
@gemmellr Thanks as usual!

PR #3407, opened 2021-01-14
ARTEMIS-3067 - track application properties in memory estimate and ch…

gtully:
…eck for modification after potential filter execution

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPStandardMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPStandardMessage.java
@@ -124,10 +124,22 @@ public class AMQPStandardMessage extends AMQPMessage {
    @Override
    public int getMemoryEstimate() {
       if (memoryEstimate == -1) {
-         memoryEstimate = memoryOffset + (data != null ? data.capacity() : 0);
+         memoryEstimate = memoryOffset + (data != null ? data.capacity() + unmarshalledApplicationPropertiesMemoryEstimateFromData() : 0);
       }
 
       return memoryEstimate;
+   }
+
+   private int unmarshalledApplicationPropertiesMemoryEstimateFromData() {
+      if (applicationProperties != null) {
+         // they have been unmarshalled, estimate memory usage based on their encoded size
+         if (remainingBodyPosition != VALUE_NOT_PRESENT) {
+            return remainingBodyPosition - applicationPropertiesPosition;
+         } else {
+            return data.capacity() - applicationPropertiesPosition;
+         }
+      }
+      return 0;
    }
 
    @Override
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -2935,6 +2935,9 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
          MessageReference ref;
 
+         // filter evaluation or transformation may cause properties to be lazyDecoded, we need to reflect that
+         int existingMemoryEstimate = 0;
+
          Consumer handledconsumer = null;
 
          synchronized (this) {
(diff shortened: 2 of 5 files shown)


gtully:
This fix seems a little indirect to me, but I did not see a better way. Am open to suggestions!

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPStandardMessage.java
```diff
@@ -124,12 +124,24 @@ protected ReadableBuffer getData() {
...
       }
 
       return memoryEstimate;
    }
 
+   private int unmarshalledApplicationPropertiesMemoryEstimateFromData() {
+      if (applicationProperties != null) {
+         // they have been unmarshalled, estimate memory usage based on their encoded size
+         if (remainingBodyPosition != VALUE_NOT_PRESENT) {
+            return remainingBodyPosition - applicationPropertiesPosition;
```
clebertsuconic, line 137: it's important to always return the same size...

When the message enters the system, you're adding the size of the message, when leaving you're returning the size after parsing the property.

You need to cache the value, otherwise you would get inconsistencies found by the test suite here:


 org.apache.activemq.artemis.tests.integration.client.ConsumerTest.testSendReceiveAMQP[isNetty=true, persistent=true]	5.4 sec	1
 org.apache.activemq.artemis.tests.integration.client.ConsumerTest.testSendAMQPReceiveCore[isNetty=true, persistent=true]	5.2 sec	1
 org.apache.activemq.artemis.tests.integration.client.ConsumerTest.testSendAMQPReceiveOpenWire[isNetty=true, persistent=true]	5.3 sec	1
 org.apache.activemq.artemis.tests.integration.client.ConsumerTest.testSendReceiveAMQP[isNetty=true, persistent=false]	5.3 sec	1
 org.apache.activemq.artemis.tests.integration.client.ConsumerTest.testSendAMQPReceiveCore[isNetty=true, persistent=false]	5.2 sec	1
 org.apache.activemq.artemis.tests.integration.client.ConsumerTest.testSendAMQPReceiveOpenWire[isNetty=true, persistent=false]	5.3 sec	1
 org.apache.activemq.artemis.tests.integration.cluster.failover.ReplicatedManyMultipleServerFailoverNoNodeGroupNameTest.testStartLiveFirst

gtully: the intent is to compensate for any change in the estimate, in the knowledge that it may happen for filter evaluation or for transformation. But I must not have all the possibilities covered, and it may be tricky. There is deliver and directDeliver etc. 
The consumerTest is now sorted, but I need to investigate next org.apache.activemq.artemis.tests.integration.amqp.connect.AMQPReplicaTest#testReplicaCatchupOnQueueCreates

gtully: hmm. org.apache.activemq.artemis.tests.integration.amqp.connect.AMQPReplicaTest#testReplicaCatchupOnQueueCreates looks to be timing dependent,  unrelated to the changes as far as i can see.

gtully: pushed a rebase to see if there are any more failures in the mix

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3716,6 +3730,16 @@ private boolean deliver(final MessageReference ref) {
       }
    }
 
+   private void accountForChangeInMemoryEstimate(final MessageReference ref, final int existingMemoryEstimate) {
+      final int delta = ref.getMessageMemoryEstimate() - existingMemoryEstimate;
+      if (delta > 0) {
+         PagingStore pageStore = ref.getOwner();
+         if (pageStore != null) {
+            pageStore.addSize(delta);
```
franz1981, line 3738: 2 observations:

1. what happen to core messages here?
2. this method could become static

gtully: thanks franz,
a core message won't have any delta that I am aware of, it's memory estimate does not change, but you were in that space recently, does it change?

gtully: yes to the static, let me fix that, thanks!

PR #3410, opened 2021-01-18
More console fixes

andytaylor:

PR #3412, opened 2021-01-19
ARTEMIS-3074 Add ActiveMQServerControl#createBridge() method variant …

TomasHofman:
…which accepts the callTimeout attribute

https://issues.apache.org/jira/browse/ARTEMIS-3074

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/BridgeConfiguration.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/BridgeConfiguration.java
@@ -16,16 +16,50 @@
  */
 package org.apache.activemq.artemis.core.config;
 
+import javax.json.JsonArrayBuilder;
+import javax.json.JsonObject;
+import javax.json.JsonObjectBuilder;
+import javax.json.JsonString;
+import javax.json.JsonValue;
 import java.io.Serializable;
+import java.io.StringReader;
 import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
 import org.apache.activemq.artemis.api.core.client.ActiveMQClient;
 import org.apache.activemq.artemis.core.server.ComponentConfigurationRoutingType;
+import org.apache.activemq.artemis.utils.JsonLoader;
 
 public final class BridgeConfiguration implements Serializable {
 
    private static final long serialVersionUID = -1057244274380572226L;
+
+   public static String NAME = "name";
+   public static String QUEUE_NAME = "queue-name";
+   public static String FORWARDING_ADDRESS = "forwarding-address";
+   public static String FILTER_STRING = "filter-string";
+   public static String STATIC_CONNECTORS = "static-connectors";
+   public static String DISCOVERY_GROUP_NAME = "discovery-group-name";
+   public static String HA = "ha";
+   public static String TRANSFORMER_CONFIGURATION = "transformer-configuration";
+   public static String RETRY_INTERVAL = "retry-interval";
+   public static String RETRY_INTERVAL_MULTIPLIER = "retry-interval-multiplier";
+   public static String INITIAL_CONNECT_ATTEMPTS = "initial-connect-attempts";
+   public static String RECONNECT_ATTEMPTS = "reconnect-attempts";
+   public static String RECONNECT_ATTEMPTS_ON_SAME_NODE = "reconnect-attempts-on-same-node";
+   public static String USE_DUPLICATE_DETECTION = "use-duplicate-detection";
+   public static String CONFIRMATION_WINDOW_SIZE = "confirmation-window-size";
+   public static String PRODUCER_WINDOW_SIZE = "producer-window-size";
+   public static String CLIENT_FAILURE_CHECK_PERIOD = "client-failure-check-period";
+   public static String USER = "user";
+   public static String PASSWORD = "password";
+   public static String CONNECTION_TTL = "connection-ttl";
+   public static String MAX_RETRY_INTERVAL = "max-retry-interval";
+   public static String MIN_LARGE_MESSAGE_SIZE = "min-large-message-size";
+   public static String CALL_TIMEOUT = "call-timeout";
+   public static String ROUTING_TYPE = "routing-type";
 
    private String name = null;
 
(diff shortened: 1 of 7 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -3796,19 +3796,58 @@ public void createBridge(final String name,
...
                   reconnectAttempts, useDuplicateDetection, confirmationWindowSize, clientFailureCheckPeriod,
-                  staticConnectorsOrDiscoveryGroup, useDiscoveryGroup, ha, user, "****");
+                  staticConnectorsOrDiscoveryGroup, useDiscoveryGroup, ha, user, "****", callTimeout);
       }
       checkStarted();
 
       clearIO();
 
       try {
          TransformerConfiguration transformerConfiguration = transformerClassName == null ? null : new TransformerConfiguration(transformerClassName);
```
brusdev, line 3849: You should check if `transformerClassName` isEmpty to work properly with hawtio2, see ARTEMIS-3072 / #3410 for further details.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -3796,19 +3796,58 @@ public void createBridge(final String name,
...
+              reconnectAttempts,
+              useDuplicateDetection,
+              confirmationWindowSize,
+              clientFailureCheckPeriod,
+              staticConnectorsOrDiscoveryGroup,
+              useDiscoveryGroup,
+              ha,
+              user,
+              password,
+              ActiveMQClient.DEFAULT_CALL_TIMEOUT);
```
michaelpearce-gain, line 3816: I would suggest this isnt defaulted here. 

Should keep original behaviour of not setting it in the configuration, and in new method setting it.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -3796,19 +3796,58 @@ public void createBridge(final String name,
...
+                            final int reconnectAttempts,
+                            final boolean useDuplicateDetection,
+                            final int confirmationWindowSize,
+                            final long clientFailureCheckPeriod,
+                            final String staticConnectorsOrDiscoveryGroup,
+                            boolean useDiscoveryGroup,
+                            final boolean ha,
+                            final String user,
+                            final String password,
+                            final long callTimeout) throws Exception {
```
michaelpearce-gain, line 3837: Make this Long, then you can use check on null if to set in config or not. shouldnt be defaulting it at this layer, thats for lower downs to handle, else in future if we change how config item works, we will forget all about this stuff.

jbertram:
Given the proliferation of overloaded management methods it was determined awhile back that we'd move to a more de-coupled approach using flexible JSON-based input. See, for example, the method `org.apache.activemq.artemis.api.core.management.ActiveMQServerControl#createQueue(java.lang.String)` which replaces over 10 overloaded variants. The JSON input `String` for this method can be created manually simply based on the parameter names or generated from the `org.apache.activemq.artemis.api.core.QueueConfiguration` object using the `toJSON` method. I recommend you implement a similar method for `createBridge`. The class `org.apache.activemq.artemis.core.config.BridgeConfiguration` already exists and you can use that for these changes (although you'll likely need to move it out of the `activemq-server` module and into the `activemq-commons` module).

I realize the change I'm describing is much larger than the one you already submitted in this PR, but the new method in this PR is just going to be deprecated in a bit so you'll have to change your integration code anyway. I think it's probably best to do it "right" the first time.

jbertram:
If you would prefer not to do the change I suggested I could probably send a PR relatively soon.

TomasHofman:
@jbertram no problem, I will make the change as you suggested.

Thanks for all the input!

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
TomasHofman: Fixed in the master branch in the mean time. :+1:

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
TomasHofman: Resolved, thanks for input!

TomasHofman:
@jbertram I implemented the `createBridge(jsonString)` method, can you check if it is what you imagined?

I didn't move the `BridgeConfiguration` class into `activemq-commons` module as it wasn't necessary in the scope of work I did. It rather looked problematic as I would have to move other classes with it (`ActiveMQDefaultConfiguration`, `ActiveMQClient`, `ComponentConfigurationRoutingType`) which `BridgeConfiguration` depends on.

If we were to move it out of `activemq-server` module, would the `artemis-core-client` module be a suitable place? This seems easier judging from how the classes are coupled, and it would allow also us to create method `ActiveMQServerControl#createBridge(BridgeConfiguration)` :), which may be nice type safe addition to the JSON string version.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -3822,6 +3822,30 @@ public void createBridge(final String name,
       }
    }
 
+   @Override
+   public String createBridge(String bridgeConfigurationAsJson) throws Exception {
```
jbertram, line 3826: This can return `void`. The only reason the `createQueue` JSON method returns a `String` is because the broker applies address settings to the queue on the fly to end up at the final configuration and that's something the client might like to know. In this case the broker just takes what it's given.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/BridgeConfiguration.java
```diff
@@ -360,6 +498,98 @@ public BridgeConfiguration setCallTimeout(long callTimeout) {
       return this;
    }
 
+   /**
+    * This method returns a JSON-formatted {@code String} representation of this {@code QueueConfiguration}. It is a
+    * simple collection of key/value pairs. The keys used are referenced in {@link #set(String, String)}.
+    *
+    * @return a JSON-formatted {@code String} representation of this {@code QueueConfiguration}
```
jbertram, line 505: The JavaDoc references `QueueConfiguration`.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/BridgeConfiguration.java
```diff
@@ -360,6 +498,98 @@ public BridgeConfiguration setCallTimeout(long callTimeout) {
...
+      builder.add(MAX_RETRY_INTERVAL, getMaxRetryInterval());
+      builder.add(MIN_LARGE_MESSAGE_SIZE, getMinLargeMessageSize());
+      builder.add(CALL_TIMEOUT, getCallTimeout());
+      builder.add(ROUTING_TYPE, getRoutingType().name());
+
+      return builder.build().toString();
+   }
+
+   /**
+    * This method returns a {@code QueueConfiguration} created from the JSON-formatted input {@code String}. The input
```
jbertram, line 567: The JavaDoc references `QueueConfiguration`.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -3822,6 +3822,30 @@ public void createBridge(final String name,
...
+      if (AuditLogger.isEnabled()) {
+         AuditLogger.createBridge(this.server, bridgeConfigurationAsJson);
+      }
+      checkStarted();
+
+      clearIO();
+
+      try {
+         // when the QueueConfiguration is passed through createQueue all of its defaults get set which we return to the caller
```
jbertram, line 3835: The comment references `QueueConfiguration`.

jbertram:
Aside from my inline comments I've got a few others...

- Please deprecate all the other `createBridge` management methods.
- Add another test like `org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlTest#testCreateAndDestroyBridge` using the new method.

> ...it would allow also us to create method `ActiveMQServerControl#createBridge(BridgeConfiguration)` :), which may be nice type safe addition to the JSON string version.

Management methods should only take simple types so they can be invoked from JMX tools as well as HTTP via Jolokia. The only time a complex object (e.g. `BridgeConfiguration`) would be used is if the invoking client was in the same JVM but in that case they could use the embedded broker's configuration API directly and won't need the management method.

TomasHofman:
Sorry for the long delays, I'm in the middle of a release...

I realized there are further complications regarding default values. E.g. user and password fields default to `"ACTIVEMQ.CLUSTER.ADMIN.USER"` and `"CHANGE ME!!"` and we need to be able to set them to `null`, which means null values have to be deserialized. I chose following behavior:

* All non-null values are always serialized to JSON, even if they are default (this could be changed to serialize only if value is non-default, but to me that seems it would be confusing),
* if JSON string doesn't contain a field, during deserialization it will be instantiated with its' default value,
* if JSON string contains a field with `null`, it will be instantiated to null, regardless of it's default value (unless it's a primitive type field which would fail).

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
TomasHofman: Done.

jbertram:
The PR looks good and I'm ready to merge, but I had one additional question...

As I was reviewing the PR I noticed that while the bridge configuration references a connector there isn't actually any way to add a new connector via the management API which means every new bridge will have to use a connector which *already exists*. This is a gap in functionality that has been there for a long time, but I just now noticed it. Is this gap going to impact your use-case? If not, I'll go ahead and merge and then send a PR later for a new management method for adding a connector.

TomasHofman:
Hi @jbertram, connectors are not important to me, I only need a way to set bridge call-timeout :). Thank you for your reviews!

rvais:
I reviewed this PR as well. The tests which are included should be should be sufficient for the purpose. Feel free to reopen this for future Artemis version.

PR #3413, opened 2021-01-19
ARTEMIS-2937 - tidy up shutdown, resolve intermittent failure in org.…

gtully:
…apache.activemq.artemis.tests.integration.amqp.connect.AMQPReplicaTest#testReplicaCatchupOnQueueCreates


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -1223,6 +1223,7 @@ private void stop(boolean failoverOnServerShutdown,
          for (ActiveMQComponent component : this.protocolServices) {
             stopComponent(component);
          }
+         protocolServices.clear();
```
clebertsuconic, line 1226: nice catch!

PR #3415, opened 2021-01-20
NO-JIRA Minor code improvement in NettyAcceptor class

sebthom:
Make public static fields final and remove an unused private class field.


franz1981 (approved):
LGTM!

PR #3418, opened 2021-01-22
ARTEMIS-3064 - ensure useTopologyForLoadBalancing is disabled when ha…

gtully:
…=false

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederationConnection.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederationConnection.java
@@ -69,6 +69,10 @@ public class FederationConnection {
          }
       }
 
+      if (!config.isHA()) {
+         serverLocator.setUseTopologyForLoadBalancing(false);
+      }
+
       serverLocator.setConnectionTTL(config.getConnectionTTL());
       serverLocator.setClientFailureCheckPeriod(config.getClientFailureCheckPeriod());
       serverLocator.setReconnectAttempts(config.getReconnectAttempts());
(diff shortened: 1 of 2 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederationConnection.java
```diff
@@ -69,6 +69,10 @@ public FederationConnection(Configuration configuration, String name, Federation
          }
       }
 
+      if (!config.isHA()) {
+         serverLocator.setUseTopologyForLoadBalancing(false);
```
michaelandrepearce, line 73: Should we maybe just like for others look to make this pick up from url configuration?

What do bridges do?

clebertsuconic: @gtully can you run the whole test suite and post results?

gtully: interesting, https://issues.apache.org/jira/browse/ARTEMIS-347 introduces a cluster-connection-url in place of a connector-ref. The configuration is now a choice.
Maybe we can deprecate all of the server locator config for federation and just use a url? Or do we need to retain a choice? I think the server locator url parser is smart enough to cover all options w.r.t discovery groups etc.

gtully: the tests look good. I think even is we provide an alternative url config, we will need this fix for the existing configuration.

I can raise an issue to bypass the connector-refs and have a connector-uri option where the rich url has all that is needed

gtully: essentially a version of ARTEMIS-347 for federation

PR #3423, opened 2021-01-31
added space to separate 2 words that otherwise appear joined

erwindon:
Error message 222147 does not have a space between the word "queue" and the actual queue name. this simple PR fixes that.

I noticed other small opportunities for improvement with the messages:
* there are 3(!) spellings of cannot/can not/can't. none of then wrong, just inconsistent;
* some message are a full sentence and still do not start with a capital;
* the last sentence does not always end with '.' when there are multiple sentences in the message. (and how should a single sentence end?);
* a few messages have multiple consecutive spaces;
* some messages have leading/trailing spaces.

shall I change (some/all) of these in a next PR?


clebertsuconic:
thanks!

PR #3424, opened 2021-02-01
ARTEMIS-3089 direct delivery can break LVQ+non-destructive

jbertram:


michaelandrepearce:
Why not in LVQQueue impl override isDirectDeliver() logic and get it to return false if non destructive, so that all the code around LVQ is in just the queue code. Rather than leaching it out to consumer or other code parts.

jbertram:
@michaelandrepearce , that's a fair point. Although, the `isDirectDeliver()` method is only used by _other_ classes and not the queue itself to determine whether or not delivery should be direct. Therefore, the changes will almost certainly need to be more invasive than the simple solution you've suggested.

michaelandrepearce:
I think we should somehow find a way so its self contained still

jbertram:
@michaelandrepearce, agreed. I'm already working on a solution.

jbertram:
@michaelandrepearce, I isolated the changes to `org.apache.activemq.artemis.core.server.impl.LastValueQueue`. Let me know what you think.

clebertsuconic:
@jbertram would make sense LastValueQueue and direct deliver? 

I saw a previous version (before your interaction with @michaelandrepearce )... and I understand that.. but on an overall case ... would make sense direct deliver and LastValue Queue? shouldn't we really return false on Issupportdirectdeliver method for LVQ?

jbertram:
As noted previously changing the return value of `isDirectDeliver()` in and of itself won't change the behavior of the queue implementation. The only things which use `isDirectDeliver()` are tests and management.

michaelandrepearce:
> @michaelandrepearce, I isolated the changes to `org.apache.activemq.artemis.core.server.impl.LastValueQueue`. Let me know what you think.

Just saw it. Nice and clean change. And only two lines!! +1 from me

PR #3425, opened 2021-02-01
ARTEMIS-3091 Fix AMQP message headers export from the journal

brusdev:


franz1981 (approved):
LGTM: just a qq..
what happen to the scanned data after using `toCore` on bridges?
My concern is re an increased memory footprint on AMQP message that pass unnoticed (to the owner paging store)...

Given that `toCore` is a fire-and-forget action, maybe worth to duplicate the message (without performing any copy) and use `toCore` on the duplicate instance, to save the original message to increase its memory footprint.

brusdev:
@franz1981 thanks for the review. I have checked the bridge implementation with the AmqpBridgeClusterRedistributionTest. It copies the AMQPMessage before to forward it so it should be fine.

PR #3431, opened 2021-02-04
removed the single quotes around a title on the Status screen

erwindon:
The title "Address Memory Used" on the Status page has superfluous single-quotes around it.
This PR removes those.
No other title uses visible quotes (I search both visually and in source-code); so it was an exception.
The source code does not mention any special reason to have these quotes and they were not separately introduced earlier.


clebertsuconic:
@andytaylor can you review this one please?

jbertram:
@clebertsuconic, I reviewed and merged it.

@erwindon, thanks for the contribution!

PR #3435, opened 2021-02-06
ARTEMIS-3105 large message file not closed on backup side

CNNJYB:
We send a large number of large messages to the cluster, on the backup side, we found error message of too many open files. 
We think that large message is released on live side but not released on the backup side, when receive the last chunk.


franz1981 (approved):
LGTM in general, just thinking how to release it given that it introduce a new encoding/decoding form for the `ReplicationLargeMessageEndMessage`

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationLargeMessageEndMessage.java
```diff
@@ -57,6 +64,7 @@ public void decodeRest(final ActiveMQBuffer buffer) {
       if (buffer.readableBytes() >= DataConstants.SIZE_LONG) {
          pendingRecordId = buffer.readLong();
       }
+      isDelete = buffer.readBoolean();
```
clebertsuconic, line 67: can you check hasRemaining() on the buffer here, and return false by default?

We intend to keep live and replica the same version, but users do weird things... just in case please?

clebertsuconic:
LGTM.. only one small request on verifying the buffer having available bytes.. other than that I will run the whole test suite on Monday and if it's good I will merge it (after your amend please?)

PR #3436, opened 2021-02-06
ARTEMIS-2876: Port artemis-junit @Rule to junit5 extensions

luisalves00:
Please review a possible solution for ARTEMIS-2876. The module name allow backwards compatibility.

artemis-junit-all - hold the core logic that can be used by both junit4 `(@Rule)` and junit5 (`@RegisterExtension`). Possibly used by TestNG if needed. (don't depend on junit)
artemis-junit - keep only the junit4 @Rule implementation  (depend only junit4)
artemis-junit-extensions - same as artemis-junit but for Junit5.  (depend only junit5) - will allow junit5 only projects to include only junit5.

If accepted I still need to add proper javadoc, some method encapsulation and update the documentation. Also, better names for classes and modules are welcomed.


jbertram:
I think that a better approach might be to create 3 sub-modules under `artemis-junit`. You could have `artemis-junit-4`, `artemis-junit-5`, and then `artemis-junit-commons` for all the bits they share. I'm not sure what this would mean for backwards compatibility since the Maven coordinates would change for existing users. Perhaps there's some Maven magic we could do to get around that. Or perhaps we could do this change in a major version.

luisalves00:
I like the sub-modules names proposal, in fact they are similar to the ones I originally had planned, but renamed them to make it retro-compatible. If it's okay to do the breaking change I can modify the modules. Using this sub-module strategy it might even be compatible, but I think junit5 API will be pulled...I'll submit another proposal next weekend. Thanks.

eidottermihi:
@luisalves00 @jbertram  I'm really looking forward to get JUnit 5 support as it is proposed here. Since there is not very much activity since over a year - are you still actively working on this PR?

If not - I could try to take it further from here if you don't mind!

jbertram:
@eidottermihi, go for it!

luisalves00:
Thanks @eidottermihi. Currently on vacation...glad you took care of it.

jbertram:
Fixed via #4181.

PR #3448, opened 2021-02-10
[ARTEMIS-3113]: Artemis AMQP shouldn't depend on JMS.

ehsavoie:
* removing the  JMS dependency on AMQP module
* fixing destinations usage.

Jira: https://issues.apache.org/jira/browse/ARTEMIS-3113

### DIFF
--- a/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQDestination.java
+++ b/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQDestination.java
@@ -31,6 +31,7 @@ import org.apache.activemq.artemis.api.core.RoutingType;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.protocol.core.impl.PacketImpl;
 import org.apache.activemq.artemis.jndi.JNDIStorable;
+import org.apache.activemq.artemis.utils.DestinationUtil;
 
 /**
  * ActiveMQ Artemis implementation of a JMS Destination.
@@ -42,6 +43,8 @@ public class ActiveMQDestination extends JNDIStorable implements Destination, Se
 
    private static final long serialVersionUID = 5027962425462382883L;
 
+   // INFO: These variables are duplicated as part of AMQPMessageSupport in artemis-amqp-protocols
+   //       The duplication there is to avoid a dependency on this module from a server's module
    public static final String QUEUE_QUALIFIED_PREFIX = "queue://";
    public static final String TOPIC_QUALIFIED_PREFIX = "topic://";
    public static final String TEMP_QUEUE_QUALIFED_PREFIX = "temp-queue://";
@@ -167,25 +170,7 @@ public class ActiveMQDestination extends JNDIStorable implements Destination, Se
    public static SimpleString createQueueNameForSubscription(final boolean isDurable,
                                                        final String clientID,
                                                        final String subscriptionName) {
-      final String queueName;
-      if (clientID != null) {
-         if (isDurable) {
-            queueName = ActiveMQDestination.escape(clientID) + SEPARATOR +
-               ActiveMQDestination.escape(subscriptionName);
-         } else {
-            queueName = "nonDurable" + SEPARATOR +
-               ActiveMQDestination.escape(clientID) + SEPARATOR +
-               ActiveMQDestination.escape(subscriptionName);
-         }
-      } else {
-         if (isDurable) {
-            queueName = ActiveMQDestination.escape(subscriptionName);
-         } else {
-            queueName = "nonDurable" + SEPARATOR +
-               ActiveMQDestination.escape(subscriptionName);
-         }
-      }
-      return SimpleString.toSimpleString(queueName);
+      return DestinationUtil.createQueueNameForSubscription(isDurable, clientID, subscriptionName);
    }
 
    public static String createQueueNameForSharedSubscription(final boolean isDurable,
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
@@ -28,21 +28,9 @@ import java.util.Arrays;
 import java.util.Map;
 import java.util.Set;
 
-import javax.jms.Destination;
-import javax.jms.JMSException;
-import javax.jms.Queue;
-import javax.jms.TemporaryQueue;
-import javax.jms.TemporaryTopic;
-import javax.jms.Topic;
-
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.message.impl.CoreMessage;
 import org.apache.activemq.artemis.core.persistence.CoreMessageObjectPools;
-import org.apache.activemq.artemis.jms.client.ActiveMQDestination;
-import org.apache.activemq.artemis.jms.client.ActiveMQQueue;
-import org.apache.activemq.artemis.jms.client.ActiveMQTemporaryQueue;
-import org.apache.activemq.artemis.jms.client.ActiveMQTemporaryTopic;
-import org.apache.activemq.artemis.jms.client.ActiveMQTopic;
 import org.apache.activemq.artemis.protocol.amqp.converter.jms.ServerJMSBytesMessage;
 import org.apache.activemq.artemis.protocol.amqp.converter.jms.ServerJMSMapMessage;
 import org.apache.activemq.artemis.protocol.amqp.converter.jms.ServerJMSMessage;
@@ -54,7 +42,6 @@ import org.apache.qpid.proton.amqp.Binary;
 import org.apache.qpid.proton.amqp.Symbol;
 import org.apache.qpid.proton.amqp.messaging.Data;
 import org.apache.qpid.proton.message.Message;
-import org.jboss.logging.Logger;
 
 /**
  * Support class containing constant values and static methods that are used to map to / from
@@ -62,7 +49,11 @@ import org.jboss.logging.Logger;
  */
 public final class AMQPMessageSupport {
 
-   private static final Logger logger = Logger.getLogger(AMQPMessageSupport.class);
+   // These are duplicates from ActiveMQDestination on the jms module.
+   public static final String QUEUE_QUALIFIED_PREFIX = "queue://";
+   public static final String TOPIC_QUALIFIED_PREFIX = "topic://";
+   public static final String TEMP_QUEUE_QUALIFED_PREFIX = "temp-queue://";
+   public static final String TEMP_TOPIC_QUALIFED_PREFIX = "temp-topic://";
 
 
    public static SimpleString HDR_ORIGINAL_ADDRESS_ANNOTATION = SimpleString.toSimpleString("x-opt-ORIG-ADDRESS");
@@ -303,34 +294,11 @@ public final class AMQPMessageSupport {
       return  key;
    }
 
-
-   public static String toAddress(Destination destination) {
-      try {
-         if (destination instanceof ActiveMQDestination) {
-            return ((ActiveMQDestination) destination).getAddress();
-         } else {
-            if (destination instanceof Queue) {
-               return ((Queue) destination).getQueueName();
-            } else if (destination instanceof Topic) {
-
-               return ((Topic) destination).getTopicName();
-            }
-         }
-      } catch (JMSException e) {
-         // ActiveMQDestination (and most JMS implementations I know) will never throw an Exception here
-         // this is here for compilation support (as JMS declares it), and I don't want to propagate exceptions into
-         // the converter...
-         // and for the possibility of who knows in the future!!!
-         logger.warn(e.getMessage(), e);
-      }
-      return null;
-   }
-
    public static ServerJMSBytesMessage createBytesMessage(long id, CoreMessageObjectPools coreMessageObjectPools) {
       return new ServerJMSBytesMessage(newMessage(id, BYTES_TYPE, coreMessageObjectPools));
    }
 
-   public static ServerJMSBytesMessage createBytesMessage(long id, byte[] array, int arrayOffset, int length, CoreMessageObjectPools coreMessageObjectPools) throws JMSException {
+   public static ServerJMSBytesMessage createBytesMessage(long id, byte[] array, int arrayOffset, int length, CoreMessageObjectPools coreMessageObjectPools) throws Exception {
       ServerJMSBytesMessage message = createBytesMessage(id, coreMessageObjectPools);
       message.writeBytes(array, arrayOffset, length);
       return message;
@@ -348,7 +316,7 @@ public final class AMQPMessageSupport {
       return new ServerJMSTextMessage(newMessage(id, TEXT_TYPE, coreMessageObjectPools));
    }
 
-   public static ServerJMSTextMessage createTextMessage(long id, String text, CoreMessageObjectPools coreMessageObjectPools) throws JMSException {
+   public static ServerJMSTextMessage createTextMessage(long id, String text, CoreMessageObjectPools coreMessageObjectPools) throws Exception {
       ServerJMSTextMessage message = createTextMessage(id, coreMessageObjectPools);
       message.setText(text);
       return message;
@@ -358,13 +326,13 @@ public final class AMQPMessageSupport {
       return new ServerJMSObjectMessage(newMessage(id, OBJECT_TYPE, coreMessageObjectPools));
    }
 
-   public static ServerJMSMessage createObjectMessage(long id, Binary serializedForm, CoreMessageObjectPools coreMessageObjectPools) throws JMSException {
+   public static ServerJMSMessage createObjectMessage(long id, Binary serializedForm, CoreMessageObjectPools coreMessageObjectPools) throws Exception {
       ServerJMSObjectMessage message = createObjectMessage(id, coreMessageObjectPools);
       message.setSerializedForm(serializedForm);
       return message;
    }
 
-   public static ServerJMSMessage createObjectMessage(long id, byte[] array, int offset, int length, CoreMessageObjectPools coreMessageObjectPools) throws JMSException {
+   public static ServerJMSMessage createObjectMessage(long id, byte[] array, int offset, int length, CoreMessageObjectPools coreMessageObjectPools) throws Exception {
       ServerJMSObjectMessage message = createObjectMessage(id, coreMessageObjectPools);
       message.setSerializedForm(new Binary(array, offset, length));
       return message;
@@ -374,7 +342,7 @@ public final class AMQPMessageSupport {
       return new ServerJMSMapMessage(newMessage(id, MAP_TYPE, coreMessageObjectPools));
    }
 
-   public static ServerJMSMapMessage createMapMessage(long id, Map<String, Object> content, CoreMessageObjectPools coreMessageObjectPools) throws JMSException {
+   public static ServerJMSMapMessage createMapMessage(long id, Map<String, Object> content, CoreMessageObjectPools coreMessageObjectPools) throws Exception {
       ServerJMSMapMessage message = createMapMessage(id, coreMessageObjectPools);
       final Set<Map.Entry<String, Object>> set = content.entrySet();
       for (Map.Entry<String, Object> entry : set) {
@@ -396,7 +364,8 @@ public final class AMQPMessageSupport {
    }
 
 
-   public static byte destinationType(Destination destination) {
+   // IMPORTANT-TODO: HOW TO GET THIS?
+   /*public static byte destinationType(Destination destination) {
       if (destination instanceof Queue) {
          if (destination instanceof TemporaryQueue) {
             return TEMP_QUEUE_TYPE;
@@ -412,21 +381,26 @@ public final class AMQPMessageSupport {
       }
 
       return QUEUE_TYPE;
+   } */
+
+   public static byte destinationType(String destination) {
+      if (destination.startsWith(QUEUE_QUALIFIED_PREFIX)) {
+         return QUEUE_TYPE;
+      }
+      if (destination.startsWith(TOPIC_QUALIFIED_PREFIX)) {
+         return TOPIC_TYPE;
+      }
+      if (destination.startsWith(TEMP_QUEUE_QUALIFED_PREFIX)) {
+         return TEMP_QUEUE_TYPE;
+      }
+      if (destination.startsWith(TEMP_TOPIC_QUALIFED_PREFIX)) {
+         return TEMP_TOPIC_TYPE;
+      }
+      return QUEUE_TYPE;
    }
 
-   public static Destination destination(byte destinationType, String address) {
-      switch (destinationType) {
-         case TEMP_QUEUE_TYPE:
-            return new ActiveMQTemporaryQueue(address, null);
-         case TEMP_TOPIC_TYPE:
-            return new ActiveMQTemporaryTopic(address, null);
-         case TOPIC_TYPE:
-            return new ActiveMQTopic(address);
-         case QUEUE_TYPE:
-            return new ActiveMQQueue(address);
-         default:
-            return new ActiveMQQueue(address);
-      }
+   public static String destination(byte destinationType, String address) {
+      return address;
    }
 
 }
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
@@ -39,7 +39,11 @@ import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSup
 import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.JMS_AMQP_ORIGINAL_ENCODING;
 import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.JMS_AMQP_REPLYTO_GROUP_ID;
 import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.OCTET_STREAM_CONTENT_TYPE;
+import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.QUEUE_QUALIFIED_PREFIX;
 import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.SERIALIZED_JAVA_OBJECT_CONTENT_TYPE;
+import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.TEMP_QUEUE_QUALIFED_PREFIX;
+import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.TEMP_TOPIC_QUALIFED_PREFIX;
+import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.TOPIC_QUALIFIED_PREFIX;
 import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.createBytesMessage;
 import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.createMapMessage;
 import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.createMessage;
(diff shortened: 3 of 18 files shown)


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
```diff
@@ -396,7 +364,8 @@ private static CoreMessage newMessage(long id, byte messageType, CoreMessageObje
    }
 
 
-   public static byte destinationType(Destination destination) {
+   // IMPORTANT-TODO: HOW TO GET THIS?
```
tabish121, line 367: I'm confused, if this is important than why commit the change before having figured that bit out, if it isn't then why leave this whole bit of commented out code here ?

michaelandrepearce:
@clebertsuconic can you please let PRs be open for a day or so to give people time to review. This was opened 2hrs ago and merged under an hour of being opened

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
```diff
@@ -303,34 +294,11 @@ public static String toAnnotationName(String key) {
       return  key;
    }
 
-
-   public static String toAddress(Destination destination) {
```
michaelandrepearce, line 307: Wheres this logic gone to?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
```diff
@@ -303,34 +294,11 @@ public static String toAnnotationName(String key) {
...
-      }
-      return null;
-   }
-
    public static ServerJMSBytesMessage createBytesMessage(long id, CoreMessageObjectPools coreMessageObjectPools) {
       return new ServerJMSBytesMessage(newMessage(id, BYTES_TYPE, coreMessageObjectPools));
    }
 
-   public static ServerJMSBytesMessage createBytesMessage(long id, byte[] array, int arrayOffset, int length, CoreMessageObjectPools coreMessageObjectPools) throws JMSException {
+   public static ServerJMSBytesMessage createBytesMessage(long id, byte[] array, int arrayOffset, int length, CoreMessageObjectPools coreMessageObjectPools) throws Exception {
```
michaelandrepearce, line 301: Can we avoid throwing raw exception

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -264,8 +266,8 @@ protected static ServerJMSMessage processHeader(ServerJMSMessage jms, Header hea
             jms.setLongProperty("JMSXDeliveryCount", header.getDeliveryCount().longValue() + 1);
          }
       } else {
-         jms.setJMSPriority((byte) javax.jms.Message.DEFAULT_PRIORITY);
-         jms.setJMSDeliveryMode(DeliveryMode.NON_PERSISTENT);
+         jms.setJMSPriority((byte) MESSAGE_DEFAULT_PRIORITY);
```
michaelandrepearce, line 269: I dont understand the driver here. This is handling JMS specific stuff, so it just makes sense to use the api values and not be making our own

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
```diff
@@ -444,7 +446,7 @@ private static void encodeUnsupportedMessagePropertyType(ServerJMSMessage jms, S
       }
    }
 
-   private static void setProperty(javax.jms.Message msg, String key, Object value) throws JMSException {
+   private static void setProperty(ServerJMSMessage msg, String key, Object value) throws Exception {
```
michaelandrepearce, line 449: -1 throwing raw exception

michaelandrepearce:
Im tempted to just revert this on master tbh. This is a massive change, that cannot be reviewed quickly. Ive just spent 15 mins and still got loads of classes to go and wont get through till tomorrow. 

Can you revert this on master @clebertsuconic and let everyone have time to discuss/review

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQDestination.java
```diff
@@ -42,6 +43,8 @@
 
    private static final long serialVersionUID = 5027962425462382883L;
 
+   // INFO: These variables are duplicated as part of AMQPMessageSupport in artemis-amqp-protocols
+   //       The duplication there is to avoid a dependency on this module from a server's module
```
michaelandrepearce, line 47: Then move them to somewhere unified. Lets not duplicate stuff

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
```diff
@@ -54,15 +42,18 @@
...
-import org.jboss.logging.Logger;
 
 /**
  * Support class containing constant values and static methods that are used to map to / from
  * AMQP Message types being sent or received.
  */
 public final class AMQPMessageSupport {
 
-   private static final Logger logger = Logger.getLogger(AMQPMessageSupport.class);
+   // These are duplicates from ActiveMQDestination on the jms module.
```
michaelandrepearce, line 52: -1 please dont duplicate stuff. If needed place into some common place

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
```diff
@@ -171,15 +165,15 @@ public static AMQPMessage fromCore(ICoreMessage coreMessage, StorageManager stor
             properties.setMessageId("ID:" + message.getInnerMessage().getUserID().toString());
          }
       }
-      Destination destination = message.getJMSDestination();
+      SimpleString destination = message.getJMSDestination();
```
michaelandrepearce, line 168: I expect get JMSDestination to return a JMSDestination object not a string

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMapMessage.java
```diff
@@ -137,84 +123,75 @@ public void setObject(final String name, final Object value) throws JMSException
...
    }
 
-   @Override
-   public byte getByte(final String name) throws JMSException {
+   public byte getByte(final String name) throws Exception {
       try {
          return map.getByteProperty(new SimpleString(name));
       } catch (ActiveMQPropertyConversionException e) {
-         throw new MessageFormatException(e.getMessage());
+         throw new RuntimeException(e.getMessage());
```
michaelandrepearce, line 142: -1 just throwing a runtime exception. This loses all context about the exception type. Ditto this on all same in this pr

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
```diff
@@ -137,226 +131,178 @@ public final String getJMSCorrelationID() throws JMSException {
...
+   public final void setJMSDeliveryMode(int deliveryMode) throws Exception {
+      switch (deliveryMode) {
+         case DeliveryMode_PERSISTENT:
+            message.setDurable(true);
+            break;
+         case DeliveryMode_NON_PERSISTENT:
+            message.setDurable(false);
+            break;
+         default:
+            throw new Exception("Invalid mode " + deliveryMode);
```
michaelandrepearce, line 167: -1 throwing raw exception.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
```diff
@@ -137,226 +131,178 @@ public final String getJMSCorrelationID() throws JMSException {
...
 
-   @Override
-   public final void setJMSDestination(Destination destination) throws JMSException {
-      if (destination == null) {
-         message.setAddress((SimpleString) null);
-      } else {
-         message.setAddress(((ActiveMQDestination) destination).getSimpleAddress());
-      }
-
+   public final void setJMSDestination(String destination) throws Exception {
```
michaelandrepearce, line 150: -1 the name of this suggests this should be a jms object.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
```diff
@@ -137,226 +131,178 @@ public final String getJMSCorrelationID() throws JMSException {
...
-   public final void setJMSDeliveryMode(int deliveryMode) throws JMSException {
-      if (deliveryMode == DeliveryMode.PERSISTENT) {
-         message.setDurable(true);
-      } else if (deliveryMode == DeliveryMode.NON_PERSISTENT) {
-         message.setDurable(false);
-      } else {
-         throw new JMSException("Invalid mode " + deliveryMode);
+   public final void setJMSDeliveryMode(int deliveryMode) throws Exception {
+      switch (deliveryMode) {
+         case DeliveryMode_PERSISTENT:
```
michaelandrepearce, line 160: Can we make a class with the constants inside so its a bit like before. This just is getting messy now

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
```diff
@@ -171,15 +165,15 @@ public static AMQPMessage fromCore(ICoreMessage coreMessage, StorageManager stor
...
       if (destination != null) {
-         properties.setTo(toAddress(destination));
-         maMap.put(JMS_DEST_TYPE_MSG_ANNOTATION, AMQPMessageSupport.destinationType(destination));
+         properties.setTo(destination.toString());
```
michaelandrepearce, line 170: Wheres the original logic that was being done inside toAddress

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
```diff
@@ -425,7 +419,7 @@ private static Section convertBody(ServerJMSMessage message, Map<Symbol, Object>
                break;
          }
       } else if (message instanceof ServerJMSTextMessage) {
-         String text = (((TextMessage) message).getText());
+         String text = (((ServerJMSTextMessage) message).getText());
```
michaelandrepearce, line 422: Can we use an interface here. And not concrete classes

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
```diff
@@ -456,7 +450,7 @@ private static Section convertBody(ServerJMSMessage message, Map<Symbol, Object>
             while (true) {
                list.add(m.readObject());
             }
-         } catch (MessageEOFException e) {
+         } catch (Exception e) {
```
michaelandrepearce, line 453: This should only be catching/ doing this for EOF exceptions not for all!

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
clebertsuconic: I thought we had fixed this.. I will take a look on this on.. thanks

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
clebertsuconic: This is only used on conversions.. there's no need for anything specifically here...  This is just throwing whatever happens underneath.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: This is for Conversion only..  there's no JMSDestination.. we would need to depend on artemis-jms-impl that depends on JMS.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
clebertsuconic: There is no more JMS on the conversion module.. 

This is the AMQP Converter.. not the JMS implementation

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: Same that I said before...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: No.. we can't... we need to remove the JMS API implementation from the converter.

Right now the converter is depending on the JMS Client API.

Wildfly for instance is not JMS any longer. It is just simpler to remove the dependency.

This is again just for conversion.

michaelandrepearce: I got that you want to remove the current api. But still can have an interface even if an internal one

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: I dont follow

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
michaelandrepearce: Ok then rename methods, and dont call it JMSxxxxxx

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
clebertsuconic: what's the different on throwing something like ArtemisException here?


This is used just for the converters, all that matters is that we capture eventual exceptions and log them.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: But then rename the method. Leaving it named JMSxxxxx means you expect a JMS thing

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
clebertsuconic: The AMQP Protocol implementation is depending on the JMS implementation. Going forward we have Jakarta and JMS (at the point they are a transformation). There makes no sense to force a direct dependency on it.

Wildfly is having to import JMS just because of this.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
clebertsuconic: ok, it will be a RuntimeException.. it can't be anything else on the conversion.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
clebertsuconic: you're nit-picking here.. this is going after the JMS implementation.. but without having the actual implementation. It would be easier to keep it the same way.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: There's nothing else that could happen here...

I don't see the need to duplicate the entire JMS Exception just for these...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMapMessage.java
clebertsuconic: This should not really happen at the Converters.

Creating our own Exception here wouldn't make it more useful.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: there's a bug on this part, I'm fixing it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: I would then duplicate the whole JMS API internally for the conversion. 

Concrete types is all we need here. The only reason I added these classes in the first place was to do the conversion.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: Can we revert master? And then review this properly?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
michaelandrepearce: But thats even worse

michaelandrepearce: We really should have some exception hierarchy

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMapMessage.java
michaelandrepearce: Should not happen doesnt mean it cannot...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
clebertsuconic: Why do you suggest here? I don't want to copy, paste and duplicate all the Message Interfaces and all the Exceptions as part of the converter. That would be a waste.

Message APIs and Exception types is pretty much what I use from the JMS API. If you want a concrete exception, you are pretty much opposing to remove the dependency? I don't see a need here really.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: @michaelandrepearce I did not want to depend on JMS at the first time I implemented the converter. I did it back then because it was the easy route. (copy and paste the client API, depend on the interfaces and get done with it).

Now I need to remove the dependency. Using concrete classes is the way to go here.. I don't think there's much to discuss. .unless everybody is opposing to this change.. which I can't see it as a bad thing.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: So either then move away from it being JMS this and that. E.g. to convert from/to Core which core is not JMS. Lets just drop JMS properly

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMapMessage.java
clebertsuconic: It can... a RuntimeException that's captured on the converter is all we need.

What good would make if I created a generic ConversionException and replaced all the JMSException by ConversionException here?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
michaelandrepearce: Im really not. Im suggesting drop trying to be JMS at all if you want to remove JMS api stuff here.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMapMessage.java
michaelandrepearce: You may as well just let the original exception bubble out.

michaelandrepearce: Least that way youd get something a little more useful and specific...in this sample "ActiveMQPropertyConversionException"

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
michaelandrepearce: Thats worse. Surely we should have a specific exception

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: Or even why not just move over to being the Jakarta EE interfaces that replace the older oracle / geronimo ones. As i understand its 1..1 atm

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
michaelandrepearce: Or if ita about the move to Jarkata. Why not use the Jarkata interfaces

clebertsuconic: why the converter needs JMS at all? it just needs to know to convert it.. nothing else.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
clebertsuconic: the AMQP Protocol Head should not depend on JMS at all.. no need for it just for the interfaces definition that are not required.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
michaelandrepearce: Then maybe have a single base exception for these.  At least that way can still differentiate between those and other possible unexpected exceptions that could arise from a bug.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
michaelandrepearce: So why not simply use the jakata interfaces in server going forwards

michaelandrepearce: Its 1..1 afaik atm

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
michaelandrepearce: Differentiate between an exception thats unexpected entirely due to a bug like concurrent modification or NPE from expected ones

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
michaelandrepearce: Then lets drop the JMS constructs here then not this halfway house

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/jms/ServerJMSMessage.java
clebertsuconic: will do that

clebertsuconic:
I reverted the commit. sending a new one tomorrow

clebertsuconic:
followed up with #3450

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java
ehsavoie: I'm not sure that we should use JMS or Jakarta Messaging API to convert between core and amqp messsages

PR #3450, opened 2021-02-11
ARTEMIS-3113 - Artemis AMQP shouldn't depend on JMS.

clebertsuconic:
* removing the  JMS dependency on AMQP module
* fixing destinations usage.
* refactoring to remove some JMS usage and make exceptions a bit better

Jira: https://issues.apache.org/jira/browse/ARTEMIS-3113


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/coreWrapper/CoreMapMessageWrapper.java
```diff
@@ -137,84 +130,75 @@ public void setObject(final String name, final Object value) throws JMSException
          }
          TypedProperties.setObjectProperty(new SimpleString(name), val, map);
       } catch (ActiveMQPropertyConversionException e) {
-         throw new MessageFormatException(e.getMessage());
+         throw new RuntimeException(e.getMessage());
```
clebertsuconic, line 133: I am fixing these...

clebertsuconic: done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
```diff
@@ -396,36 +371,54 @@ private static CoreMessage newMessage(long id, byte messageType, CoreMessageObje
...
-            return new ActiveMQTemporaryQueue(address, null);
-         case TEMP_TOPIC_TYPE:
-            return new ActiveMQTemporaryTopic(address, null);
-         case TOPIC_TYPE:
-            return new ActiveMQTopic(address);
-         case QUEUE_TYPE:
-            return new ActiveMQQueue(address);
-         default:
-            return new ActiveMQQueue(address);
+         case ANYCAST: prefix = QUEUE_QUALIFIED_PREFIX; break;
```
michaelandrepearce, line 413: This looks like it no longer handles temporaries

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/coreWrapper/CoreMapMessageWrapper.java
```diff
@@ -0,0 +1,231 @@
...
+   public void setFloat(final String name, final float value) {
+      map.putFloatProperty(new SimpleString(name), value);
+   }
+
+   public void setDouble(final String name, final double value) {
+      map.putDoubleProperty(new SimpleString(name), value);
+   }
+
+   public void setString(final String name, final String value) {
+      map.putSimpleStringProperty(new SimpleString(name), value == null ? null : new SimpleString(value));
```
michaelandrepearce, line 121: Isnt there a static method on SimpleString that returns null if string null, so code can be cleaner

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/coreWrapper/CoreMessageWrapper.java
```diff
@@ -137,247 +175,187 @@ public final String getJMSCorrelationID() throws JMSException {
...
-   public void decode() throws Exception {
+   public void decode() {
       if (!message.isLargeMessage()) {
          message.getBodyBuffer().resetReaderIndex();
       }
    }
 
-   @Override
-   public final boolean isBodyAssignableTo(Class c) throws JMSException {
+   public final boolean isBodyAssignableTo(Class c)  {
```
michaelandrepearce, line 358: If this isnt used and no longer impl interface...would be better to remove. No point having a no op method in a concrete class with no interface contract

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/coreWrapper/CoreMessageWrapper.java
```diff
@@ -137,247 +175,187 @@ public final String getJMSCorrelationID() throws JMSException {
...
 
-   @Override
-   public void clearBody() throws JMSException {
+   public void clearBody() {
       message.getBodyBuffer().clear();
    }
 
-   @Override
-   public final <T> T getBody(Class<T> c) throws JMSException {
+   public final <T> T getBody(Class<T> c)  {
```
michaelandrepearce, line 338: Any point of this noop method. Cleaner to remove it

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/coreWrapper/CoreMessageWrapper.java
```diff
@@ -137,247 +175,187 @@ public final String getJMSCorrelationID() throws JMSException {
...
 
-   @Override
-   public final void setObjectProperty(String name, Object value) throws JMSException {
+   public final void setObjectProperty(String name, Object value) {
       MessageUtil.setObjectProperty(message, name, value);
    }
 
-   @Override
-   public final void acknowledge() throws JMSException {
+   public final void acknowledge() {
```
michaelandrepearce, line 330: See other comments on noops

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/coreWrapper/CoreMessageWrapper.java
```diff
@@ -137,247 +175,187 @@ public final String getJMSCorrelationID() throws JMSException {
...
    }
 
-   @Override
-   public final void setJMSExpiration(long expiration) throws JMSException {
+   public final void setJMSExpiration(long expiration)  {
       message.setExpiration(expiration);
    }
 
-   @Override
-   public final long getJMSDeliveryTime() throws JMSException {
```
michaelandrepearce, line 222: Have we a possible bug here in use case of delivery delay in jms2.0 and message gets converted from amqp to core / vice versa.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
clebertsuconic: The method that uses routing type has the same semantics here.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/coreWrapper/CoreMapMessageWrapper.java
clebertsuconic: not part of my change but ok

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/coreWrapper/CoreMessageWrapper.java
clebertsuconic: TTL is handled directly at the converter. I think this method was not being used.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageSupport.java
clebertsuconic: The original method was creating a Queue type (TemporaryQueue, TemporaryTopic, Queue, Topic), to set on the JMS Object, to then only use the simple name inside the object.

The Queue type was being set somewhere else, and this was not being used.

This method now is being use to replace ActiveMQDEstiation.createQueue

There is a test that was already validating this and it is passing.

clebertsuconic:
@michaelandrepearce I think this is ready to merge now...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/coreWrapper/CoreMessageWrapper.java
michaelandrepearce: I was talking about DeliveryDelay (feature in jms 2.0) not TimeToLive

michaelandrepearce: getJMSDeliveryTime
long getJMSDeliveryTime()
                 throws JMSException
Gets the message's delivery time value.
When a message is sent, the JMSDeliveryTime header field is left unassigned. After completion of the send or publish method, it holds the delivery time of the message. This is the the difference, measured in milliseconds, between the delivery time and midnight, January 1, 1970 UTC.

A message's delivery time is the earliest time when a JMS provider may deliver the message to a consumer. The provider must not deliver messages before the delivery time has been reached.

Returns:
the message's delivery time value
Throws:
JMSException - if the JMS provider fails to get the delivery time due to some internal error.
Since:
JMS 2.0
See Also:
setJMSDeliveryTime(long)

michaelandrepearce:
@clebertsuconic see comment inline

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/coreWrapper/CoreMessageWrapper.java
clebertsuconic: This is handled by the converter directly. The "JMS" facade was not using it in any way.

There are open wire scheduled delivery tests for this already.


On AMQP to Core Converter:

https://github.com/apache/activemq-artemis/blob/master/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AmqpCoreConverter.java#L279-L300

On Core to AMQP Converter, the Converter will remove the Scheduled Delivery on Purpose, as part of:

https://github.com/apache/activemq-artemis/blob/master/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java#L300-L302



The only reason Actually AMQP To Core converter is treating scheduling is because we used to always convert to Core. On the recent versions of the broker the message will stay on its original format until the actuall delivery occurred.

If there is a bug it would not been cause by this change anyways but I think we are safe here.

clebertsuconic: I will add some testing as a separate task from this. There's one possibility actually, when you send Core Message with a scheduled delivery through the AMQP Mirror the delivery delay may been stripped on that case. This would be a side effect of what you mentioned. I will test it and fix it if it's an issue. (not related to this change but I will validate it).

michaelandrepearce: Totally agree it's not related to this change. E.g. if its a bug was preexisting, just that on reviewing this change in detail it flagged to me this looks like a bug we may not know about, better to discuss it now than someone hit it in prod.

clebertsuconic: @michaelandrepearce PR coming soon: https://issues.apache.org/jira/browse/ARTEMIS-3116

It was an issue indeed. I believe you would need AMQP Mirror to hit the issue.

PR coming soon.

PR #3451, opened 2021-02-11
Add a note about the `jms-client-id` session metadata to the docs

jsmucr:
I've added a note about how to identify an open core session via the management console by setting the `jmx-client-id` session metadata entry.

### DIFF
--- a/docs/user-manual/en/core.md
+++ b/docs/user-manual/en/core.md
@@ -159,6 +159,19 @@ end asynchronous systems which is not possible using the standard JMS API. For
 more information on this advanced feature please see the section [Guarantees of
 sends and commits](send-guarantees.md).
 
+> **Identifying your session**
+>
+> Assigning IDs to your core sessions can save you from a bit of headache
+> when monitoring and debugging the cluster using the [management console](management-console.md).
+> ```java
+> ClientSession session;
+> // ...
+> session.addMetaData(ClientSession.JMS_SESSION_IDENTIFIER_PROPERTY, "jms-client-id");
+> session.addMetaData("jms-client-id", "my-session");
+> ```
+> Such ID will then appear in the **Client ID** column under the
+> **Connections**, **Consumers** and **Producers** tabs.
+
 ### ClientConsumer
 
 Clients use `ClientConsumer` instances to consume messages from a queue. Core


docs/user-manual/en/core.md
```diff
@@ -159,6 +159,19 @@ end asynchronous systems which is not possible using the standard JMS API. For
 more information on this advanced feature please see the section [Guarantees of
 sends and commits](send-guarantees.md).
 
+> **Identifying your session**
```
clebertsuconic, line 162: these are all commented out.

docs/user-manual/en/core.md
```diff
@@ -159,6 +159,19 @@ end asynchronous systems which is not possible using the standard JMS API. For
...
+>
+> Assigning IDs to your core sessions can save you from a bit of headache
+> when monitoring and debugging the cluster using the [management console](management-console.md).
+> ```java
+> ClientSession session;
+> // ...
+> session.addMetaData(ClientSession.JMS_SESSION_IDENTIFIER_PROPERTY, "jms-client-id");
```
clebertsuconic, line 169: setClientID from JMS would do the same here.

jbertram:
FWIW, I think we should add a core-specific client identifier meta-data key so users don't need to use a JMS-specific one. It should be really easy to add. I'll open a Jira for that and send a PR soon (which will update this documentation).

docs/user-manual/en/core.md
jsmucr: What if I don't use JMS though?

docs/user-manual/en/core.md
jsmucr: ? I don't understand.

jsmucr:
I'm sorry that you don't find this useful. But please note that it took me a lot of googling before I found this approach somewhere in the depths of Stack Overflow. I'm pretty sure it should be written *somewhere*. So for those desperately searching, once again:

```java
session.addMetaData(ClientSession.JMS_SESSION_IDENTIFIER_PROPERTY, "jms-client-id"); // Or just ""
session.addMetaData("jms-client-id", "my-session");
```

Any other approach didn't work for me.

jbertram:
@jsmucr, it's not that the documentation isn't *useful*. It's simply that if core clients need functionality that they don't have (e.g. a way to uniquely identify themselves in the web console) then we should add that ability rather than using JMS implementation details in the core client. My goal is to make this ability more intuitive and less reliant on a hack. I apologize for any misunderstanding.

I'm curious...Did you ask about this subject yourself on the ActiveMQ user mailing list or Stack Overflow? If not, you might have saved yourself some trouble.

jsmucr:
@jbertram I see, thank you. 

I didn't ask there. It seems that I've got banned for some reason (not sure how that happened). I asked a couple of times to get unbanned via the form that the page offered to me but nobody responded.

PR #3455, opened 2021-02-16
  ARTEMIS-3121 Avoid creation of unnecessary StringBuilder instance in Netty…

sebthom:
…Acceptor#getProtocols(Map)

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
@@ -862,18 +862,12 @@ public class NettyAcceptor extends AbstractAcceptor {
       return this;
    }
 
-   private static String getProtocols(Map<String, ProtocolManager> protocolManager) {
-      StringBuilder sb = new StringBuilder();
-      if (protocolManager != null) {
-         Set<String> strings = protocolManager.keySet();
-         for (String string : strings) {
-            if (sb.length() > 0) {
-               sb.append(",");
-            }
-            sb.append(string);
-         }
+   private static String getProtocols(final Map<String, ProtocolManager> protocolManagers) {
+      if (protocolManagers == null || protocolManagers.isEmpty()) {
+         return "";
       }
-      return sb.toString();
+
+      return String.join(",", protocolManagers.keySet());
    }
 
    // Inner classes -----------------------------------------------------------------------------


artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
```diff
@@ -863,17 +863,19 @@ public NettyAcceptor setShutdownTimeout(int shutdownTimeout) {
    }
 
    private static String getProtocols(Map<String, ProtocolManager> protocolManager) {
-      StringBuilder sb = new StringBuilder();
-      if (protocolManager != null) {
+      if (protocolManager == null) {
```
franz1981, line 866: how many times this is supposed to happen?

sebthom: I don't know, but code like this just hurts my eyes.

sebthom: I would actually prefer a early exit using `if(protocolManager == null) return "";` but that would result in more lines being touched.

sebthom: For me it is also about reducing the cognitive load when reading source code. Now you know directly when a null was specified you get an empty string.

franz1981: it's fine as long is more readable...
> but code like this just hurts my eyes.

Agree, but I would focus on things that hurt/benefit common usage, not just our eyes :P 
Although I agree is nice to clean up stuff, to make it more readable :)

sebthom: @franz1981 I just came across it while trying to understand/analyze the whole class to better understand what may cause https://issues.apache.org/jira/browse/ARTEMIS-3117. Maybe you can have a look there too. :-D

franz1981: taking a look although i suggest to run async-profiler on it, disabling Dyantrace; that would help finding native hotspots too (if SSL is involved on Netty side).

michaelandrepearce:
This isnt NO Jira work. Can a JIRA item be raised

clebertsuconic:
> This isnt NO Jira work. Can a JIRA item be raised

@michaelandrepearce this is just code cleanup. Nothing that would bring any value to a release notes...

The commit itself would be enough record of the change here.

If someone is creating a JIRA, it would be a task, as the is not an improvement, not a feature, not a bug fx.. no value on the release notes.

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
```diff
@@ -863,15 +863,16 @@ public NettyAcceptor setShutdownTimeout(int shutdownTimeout) {
    }
 
    private static String getProtocols(Map<String, ProtocolManager> protocolManager) {
+      if (protocolManager == null || protocolManager.isEmpty())
```
gemmellr, line 866: If we are aiding readability here, what about making the variable name "protocolManagers" since it is a map of them, not a single instance.

If statements are nicer with braces I find, less screwups later.

I have no idea where this is called from but I'm also initially surprised it doesnt just throw if the map is null/empty. When is it expected/valid to have no protocols?

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
```diff
@@ -863,15 +863,16 @@ public NettyAcceptor setShutdownTimeout(int shutdownTimeout) {
...
       StringBuilder sb = new StringBuilder();
-      if (protocolManager != null) {
-         Set<String> strings = protocolManager.keySet();
-         for (String string : strings) {
-            if (sb.length() > 0) {
-               sb.append(",");
-            }
-            sb.append(string);
+      Set<String> strings = protocolManager.keySet();
+      for (String string : strings) {
```
gemmellr, line 871: Again if the aim is readability, using protocolManager(s).forEach(..) would seem nicer than seperate .keySet() and for-each loop, and perhaps also more efficient depending on impl.

gemmellr (commented):
I would tend to agree with this not being a NO-JIRA change. Changing code almost never is.

It arguably is an improvement, albeit a trivial code-only one with no intent of messaging behavioural change expected. Changes can always have unintended causes though (e.g we came across one such issue very recently) and while that doesnt appear to be the case here it rarely tends to at the time, its good to have the JIRA for reference and/or update later. Change is change. We've each already spent longer discussing this than it would have taken to create a JIRA.

sebthom:
@gemmellr Based on your feedback I cleanedup/simplified the method further.

I have no problem in opening a corresponding JIRA if required.

PR #3459, opened 2021-02-17
ARTEMIS-3117 Provide CachingOpenSSLContextFactory

sebthom:
to mitigate performance degradation in JDK 11 during TLS connection initialization.

See https://issues.apache.org/jira/browse/ARTEMIS-3117

This is an alternative to https://github.com/apache/activemq-artemis/pull/3456 based on feedback by @franz1981

This PR introduces an additional `SSLContextConfig` object that can be passed around instead of having to provide 6-7 parameters on each method invocation throughout the call chain.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnector.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnector.java
@@ -104,6 +104,7 @@ import io.netty.util.ResourceLeakDetector;
 import io.netty.util.ResourceLeakDetector.Level;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GlobalEventExecutor;
+import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
 import org.apache.activemq.artemis.core.client.ActiveMQClientLogger;
 import org.apache.activemq.artemis.core.client.ActiveMQClientMessageBundle;
@@ -116,15 +117,15 @@ import org.apache.activemq.artemis.spi.core.remoting.BufferHandler;
 import org.apache.activemq.artemis.spi.core.remoting.ClientConnectionLifeCycleListener;
 import org.apache.activemq.artemis.spi.core.remoting.ClientProtocolManager;
 import org.apache.activemq.artemis.spi.core.remoting.Connection;
+import org.apache.activemq.artemis.spi.core.remoting.ssl.OpenSSLContextFactoryProvider;
+import org.apache.activemq.artemis.spi.core.remoting.ssl.SSLContextConfig;
+import org.apache.activemq.artemis.spi.core.remoting.ssl.SSLContextFactoryProvider;
 import org.apache.activemq.artemis.utils.ConfigurationHelper;
 import org.apache.activemq.artemis.utils.FutureLatch;
 import org.apache.activemq.artemis.utils.IPV6Util;
 import org.jboss.logging.Logger;
 
 import static org.apache.activemq.artemis.utils.Base64.encodeBytes;
-
-import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
-import org.apache.activemq.artemis.spi.core.remoting.ssl.SSLContextFactoryProvider;
 
 public class NettyConnector extends AbstractConnector {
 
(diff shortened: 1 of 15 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/remoting/ssl/SSLContextConfig.java
```diff
@@ -0,0 +1,215 @@
...
+
+import org.apache.activemq.artemis.core.remoting.impl.netty.TransportConstants;
+
+/**
+ * This class holds configuration parameters for SSL context initialization.
+ * To be used with {@link SSLContextFactory} and {@link OpenSSLContextFactory}.
+ */
+public final class SSLContextConfig {
+
+   public static final class Builder {
```
jbertram, line 28: This `Builder` class seems a bit redundant to me. It has all the same properties as the parent `SSLContextConfig`. Why not just make `SSLContextConfig` fluent by having normal setter methods which return `this` (as is done for other configuration objects in the code-base, e.g. `org.apache.activemq.artemis.core.config.DivertConfiguration`)?

sebthom: This way SSLContext is immutable and can be directly used as cache key in the caching factories.

jbertram: Makes sense, @sebthom. Thanks for the explanation.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnector.java
```diff
@@ -621,11 +622,23 @@ public void initChannel(Channel channel) throws Exception {
 
             if (sslEnabled && !useServlet) {
 
-               SSLEngine engine;
+               final SSLContextConfig sslContextConfig = SSLContextConfig.builder()
+                  .withKeystoreProvider(realKeyStoreProvider)
```
jbertram, line 626: Maybe this is a nit-pick, but I prefer not having the "with" on the method names here.

jbertram:
@ehsavoie, can you review this please to make sure it's not going to break any of the integration you implemented originally?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnector.java
sebthom: I can change that, no problem.

sebthom:
@jbertram @ehsavoie I have two questions:
1. why was the `SSLContextFactory` lookup implemented via a ServiceLoader and using a priority-based selection algo. This seems unnecessarily complicated. why was it not implemented in a way that e.g. the fully qualified class of the context factory to be used can be specified in the transport settings of the `broker.xml` instead?
1. why is `CachingSSLContextFactory` holding the cache in a static map and not an instance field? https://github.com/apache/activemq-artemis/blob/52263663c48082227916cc3477f8892d9f10134b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/CachingSSLContextFactory.java#L35

docs/user-manual/en/configuring-transports.md
```diff
@@ -499,22 +499,32 @@ for more details.
   [broker's classpath](using-server.md#adding-runtime-dependencies).
 
 
-#### Configuring a SSLContextFactory
+#### Configuring an SSLContextFactory
```
ehsavoie, line 502: Why 'an' ?

ehsavoie (commented):
1. Because I wanted to be able to override the factory (which I do in my current integration code) and I didn't want another configuration entry.
2. This was a quick poc but as we didn't want to introduce regression this was not used :(. I agree that static is not required.
3. Because I wanted to be able to reuse SSLContext coming from some other provider, so the other parameters wouldn't be used. Also note that you could have several nodes with different paths to the stores but using the same cachekey.

docs/user-manual/en/configuring-transports.md
sebthom: Because `SSL` is an abbreviation and you speak it with the vowel "e" first, i.e. "an Es Es El"

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnector.java
sebthom: Done.

sebthom:
Is there any good reason not enable SSL context caching by default? Esp. since Artemis will require Java 11 soon where SSL context is significantly more expensive.

jbertram:
The rationale for not using caching by default was simply to avoid changing the default behavior as users can be sensitive to that. Moving to Java 11 may change that. At the very least we should call this out in the upgrade documentation.

jbertram:
This looks good to me. Do you have any objections @ehsavoie? If not, I'll merge it.

ehsavoie (approved):

jbertram:
Great work! Thanks again, @sebthom.

PR #3463, opened 2021-02-22
ARTEMIS-3093 Ordering on multiple consumers

clebertsuconic:


clebertsuconic:
I am fixing and running tests... please, do not merge this yet.

michaelandrepearce:
Im getting a bit of de ja vu here. Didnt alot of these changes being removed you added for something similar with ordering before?

clebertsuconic:
We did, yeah...

What I'm doing now, is I'm always using addSorted on close / rollback.

While scheduling will still use addHead... so we won't need scheduling argument on addSorted...

I did some cleanup then on the API around it.

Back then I thought we should have kept core as it was. but I see this change on core would be better as well.

clebertsuconic:
I will minimize the changes though... I will keep the scheduling for now

clebertsuconic:
fixed it!

michaelandrepearce:
Ah ok. So basically youre no longer making distinction between amqp and other clients and applying the same logic regardless. Gotcha nice. +1

clebertsuconic:
I will wait the test results on a 3 hours run I have with the whole test suite before I merge this.

PR #3464, opened 2021-02-22
ARTEMIS-3135 - track possible change in memory estimate when messages…

gtully:
… are converted to maps for JMX or UI display, follows up from ARTEMIS-3067

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -3072,7 +3072,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
          }
 
          if (existingMemoryEstimate > 0 ) {
-            accountForChangeInMemoryEstimate(ref, existingMemoryEstimate);
+            MessageReferenceImpl.accountForChangeInMemoryEstimate(ref, existingMemoryEstimate);
          }
       }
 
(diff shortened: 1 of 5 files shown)


gtully:
full suite looks good.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3733,16 +3733,6 @@ private boolean deliver(final MessageReference ref) {
       }
    }
 
-   private static void accountForChangeInMemoryEstimate(final MessageReference ref, final int existingMemoryEstimate) {
-      final int delta = ref.getMessageMemoryEstimate() - existingMemoryEstimate;
-      if (delta > 0) {
```
clebertsuconic, line 3738: what if the delta was lower?

shouldn't you set messageMemoryEstimate, in case there are further changes on the message...

(just to keep it ready for future usages, say a transformation)

gtully: it is quite specific to the case of decoding application properties which will always increase the usage.  it is currently the only operation that will cause the memoryEstimate to change.

gtully: re worked fix to be less error prone via a direct reference to the owning page store in the message in stead of the message reference. https://github.com/apache/activemq-artemis/pull/3467

PR #3465, opened 2021-02-23
ARTEMIS-2802: Add a null check when checking matching metaData inside Federated Queue

cshannon:
Other protocols besides CORE may not have a metadata map so we need to
check for null before passing to the filter matcher

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/queue/FederatedQueue.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/queue/FederatedQueue.java
@@ -143,7 +143,8 @@ public class FederatedQueue extends FederatedAbstract implements ActiveMQServerC
       //We check the session meta data to see if its a federation session, if so by default we ignore these.
       //To not ignore these, set include-federated to true, which will mean no meta data filter.
       ServerSession serverSession = server.getSessionByID(consumer.getSessionID());
-      if (metaDataFilter != null && serverSession != null && metaDataFilter.match(serverSession.getMetaData())) {
+      if (metaDataFilter != null && serverSession != null && serverSession.getMetaData() != null &&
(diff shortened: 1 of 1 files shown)


jbertram:
Nice catch, @cshannon. Any chance you could add a test to avoid potential regressions in the future?

cshannon:
@jbertram - I can try, the current tests are only set up for CORE. I can see if I can add a quick test for OpenWire to FederatedQueue before merging.

cshannon:
@jbertram - So it looks like it would take some work to convert the tests over to use a real port and to then use OpenWire, etc. I tried to do it real fast but the RetryRule is giving me a lot of problems with port binding issues and I don't really have more time this week to work on it unfortunately as I have a bunch of other stuff to do. If it's not urgent I can try and get back to it when I have time if no one else wants to pick it up and try and do a test.

cshannon:
@jbertram - I took one more quick look at writing a test because I realized @michaelandrepearce uploaded some modified test files that work with OpenWire. After looking at it i'm not really sure we can write a good test.

I ran those test files and I do see the log but as pointed out everything still passes. The reason why those tests pass are because when the exception occurs inside of FilterImpl it just returns false which is what the proper result should be anyways. See: https://github.com/apache/activemq-artemis/blob/36a771150b3ce7a0d889f7df3b7c7a6cd37769f7/artemis-server/src/main/java/org/apache/activemq/artemis/core/filter/impl/FilterImpl.java#L128

So, I'm not really sure of a way (at least with integration tests) to write a test to verify the fix as the correct thing already happens either way thanks to the error handling in the exception block. The only thing I can think of would be to maybe mock some stuff and add a test to try and verify that the createRemoteConsumer method checks for null properly before passing to the metaDataFilter but that would probably require re-working the FederatedQueue class to be able to get access to test as everything is private and non-static now.

jbertram:
@cshannon, that's fair enough. Thanks for looking into it. I've merged the PR.

PR #3468, opened 2021-02-24
ARTEMIS-3138 Shared Nothing Live broker shouldn't try to connect to itself

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3138

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
@@ -1546,20 +1546,22 @@ public final class ServerLocatorImpl implements ServerLocatorInternal, Discovery
       if (receivedTopology) {
          return;
       }
-      TransportConfiguration[] newInitialconnectors = (TransportConfiguration[]) Array.newInstance(TransportConfiguration.class, newConnectors.size());
 
-      int count = 0;
+      final List<TransportConfiguration> newInitialconnectors = new ArrayList<>(newConnectors.size());
+
       for (DiscoveryEntry entry : newConnectors) {
-         newInitialconnectors[count++] = entry.getConnector();
-
          if (ha && topology.getMember(entry.getNodeID()) == null) {
             TopologyMemberImpl member = new TopologyMemberImpl(entry.getNodeID(), null, null, entry.getConnector(), null);
             // on this case we set it as zero as any update coming from server should be accepted
             topology.updateMember(0, entry.getNodeID(), member);
          }
+         // ignore its own transport connector
+         if (!entry.getConnector().equals(clusterTransportConfiguration)) {
+            newInitialconnectors.add(entry.getConnector());
+         }
       }
 
-      this.initialConnectors = newInitialconnectors.length == 0 ? null : newInitialconnectors;
+      this.initialConnectors = newInitialconnectors.toArray(new TransportConfiguration[newInitialconnectors.size()]);
 
       if (clusterConnection && !receivedTopology && this.getNumInitialConnectors() > 0) {
          // The node is alone in the cluster. We create a connection to the new node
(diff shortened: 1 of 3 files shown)


franz1981:
Please don't merge it yet: I'm sure this is the right approach to solve the problem without causing other issues

franz1981:
@clebertsuconic @gaohoward 
I've added a second commit with a different approach to fix this, but I'm hitting https://issues.apache.org/jira/browse/ARTEMIS-3138?focusedCommentId=17290060&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17290060

ie multiple exceptions thrown (although with silent logs) with 
```
AMQ219024: Could not select a TransportConfiguration to create SessionFactory
```
Because the connector received via broadcast is the same one of the cluster configuration, hence is ignored.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
```diff
@@ -1546,20 +1546,22 @@ public synchronized void connectorsChanged(List<DiscoveryEntry> newConnectors) {
...
+      final List<TransportConfiguration> newInitialconnectors = new ArrayList<>(newConnectors.size());
 
+      for (DiscoveryEntry entry : newConnectors) {
          if (ha && topology.getMember(entry.getNodeID()) == null) {
             TopologyMemberImpl member = new TopologyMemberImpl(entry.getNodeID(), null, null, entry.getConnector(), null);
             // on this case we set it as zero as any update coming from server should be accepted
             topology.updateMember(0, entry.getNodeID(), member);
          }
+         // ignore its own transport connector
+         if (!entry.getConnector().equals(clusterTransportConfiguration)) {
```
clebertsuconic, line 1559: I think a core connection would get the topology sent to clients from the ClusterConnections's ServerLocator.

As long as your own connection is getting into the Topology for the clients, this is ok.

I know the test suite is passing, but did you check if the topology is being updated somewhere else? 


just wanted to make sure the test suite passed intentionally and not by an accident like other nodes would give you the update to the clients.

franz1981: > As long as your own connection is getting into the Topology for the clients, this is ok.

That's why the check `entry.getConnector().equals(clusterTransportConfiguration)` is placed after updating `topology`: the new `entry` should affect the topology. but not the `initialConnectors` list, because the latter would make it create a connection to itself ie is not a valid connector.

I'm just concerned that the we throw multiple times `ActiveMQClientMessageBundle.BUNDLE.noTCForSessionFactory()` that's a `ActiveMQException` and it will cause the `ConnectRunnable` to re-schedule itself on `retry interval`, see:
```java
   /**
    * used for making the initial connection in the cluster
    */
   private final class ConnectRunnable implements Runnable {

      private final ServerLocatorInternal serverLocator;

      private ConnectRunnable(ServerLocatorInternal serverLocator) {
         this.serverLocator = serverLocator;
      }

      @Override
      public void run() {
         try {
            if (started) {
               serverLocator.connect();
               if (serverLocator == replicationLocator) {
                  replicationClusterConnectedLatch.countDown();
               }
            }
         } catch (ActiveMQException e) {
            if (!started) {
               return;
            }

               logger.debug("retry on Cluster Controller " + System.identityHashCode(ClusterController.this) + " server = " + server);
            }
            server.getScheduledPool().schedule(this, serverLocator.getRetryInterval(), TimeUnit.MILLISECONDS);
         }
      }
   }
```
Throwing many exceptions isn't nice and I see that it will keep on doing it during the whole life-cycle of a live server *if* there is a single pair, because the backup won't have any acceptor opened to allow the live to connect to it...

clebertsuconic:
I agree with the changes, the test suite on your branch ran 2 or 3 times successfully (which is a good rate)..

I'm going to merge this.. if you find other changes you need on this, you could apply a new PR, or revert this if we ever find it wrong. I think this is a good reason to merge this already.

clebertsuconic:
I intended to squash the commits before I merge.. I did some wrong git-fu, but it's ok.

PR #3473, opened 2021-03-02
ARTEMIS-3154 - Add support for navigation/deleting messages

andytaylor:
ARTEMIS-3154 - while being browsed in the details

https://issues.apache.org/jira/browse/ARTEMIS-3154

### DIFF
--- a/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/browse.js
+++ b/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/browse.js
@@ -84,7 +84,15 @@ var Artemis;
 
             </div>
             <div class="form-group" ng-show="$ctrl.showMessageDetails">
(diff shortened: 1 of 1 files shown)


artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/browse.js
```diff
@@ -343,7 +352,7 @@ var Artemis;
             origQueue = {
                 itemField: 'StringProperties',
                 header: 'Original Queue',
-                templateFn: function(value) {
+                templateFn: function(value) {next
```
brusdev, line 355: Is the `next` word on the 355 line a typo?

andytaylor: It is Ive fixed it, cheers Dom.

PR #3475, opened 2021-03-03
ARTEMIS-3133 Just Encapsulating ObjectPool into a small utility

clebertsuconic:


clebertsuconic:
@franz1981 for your review

clebertsuconic:
if we intend to use object pools.. perhaps we could expand on this idea.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ObjectPool.java
```diff
@@ -0,0 +1,70 @@
...
+         internalPool = PlatformDependent.newFixedMpscQueue(maxSize);
+      } else {
+         internalPool = null;
+      }
+      this.cleaner = cleaner;
+      this.supplier = supplier;
+   }
+
+   /** Use this when you are getting a new Object. In a sense of loaning something, or check in into a hotel */
+   public T checkIn() {
```
tabish121, line 48: The naming here for the pooling APIs is pretty much the exact opposite of what you normally see in pool APIs (the fact that you had to add a note here on what checkIn means it was already unclear to the writer I'd guess).  Most of the pooling APIs tend to use terminology like "borrowObject" for the fetch from the pool API and something along the lines of "returnObject" for the return.  Or if you prefer treat it more like a queue and use "poll" and "offer" etc.  Or just swap the two you have now because anyone using this is going to need proper API docs and will have to read them each time to realize that checkIn is actually taking something out of the pool and vice versa.

clebertsuconic: @tabish121 thanks...

funny story actually, m first choice was poll and offer,

which I then renamed to borrowObject and returnObject...

Which in the end I renamed to this! :)


I will switch to borrowObject and returnObject


Thanks for the review!

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ObjectPool.java
```diff
@@ -0,0 +1,70 @@
...
+public class ObjectPool<T> {
+
+   private final Queue<T> internalPool;
+
+   private final Consumer<T> cleaner;
+   private final Supplier<T> supplier;
+
+   public ObjectPool(int maxSize, Consumer<T> cleaner, Supplier<T> supplier) {
+      if (maxSize > 0) {
+         internalPool = PlatformDependent.newFixedMpscQueue(maxSize);
```
franz1981, line 39: This is the complex part: this queue work correctly only if there are multiple producers and a single consumer ie it breaks with many parallel consumers.
I am not sure I would create a general purpose utility for this reason, because the use case is very narrow and difficult to be enforced...
In addition, it seems that's used in case no pool is needed, that means 

1) we pay several additional costs just to create the other object
2) we always create the object pool itself

We can use a JMH bench so I can show better what I mean

franz1981: In short:
- If you still want to create the utility I would use different factory method and config parameters to cover all the use cases (spsc, mpmc,mpsc,spmc) but probably there are not enouf JCTools queues and just using mpmc is not a good idea because the performance and memory footprint are not negligible
- I wouldn't create any object pool in case is not needed, to save additional methods in the stack, lambdas and creation checks

franz1981: In addition: I see that probably I've been too conservative and mpsc is too much, just using spsc should be enough, please take a look where we're going to release the used packet instances -> it should always happen in a "as single-thread fashion" ie not the same thread, but there are no concurrent access from more then 1 thread each time to release it...but I could be wrong :D

clebertsuconic: @franz1981 this PR is pretty much doing the same thing you did.. I'm just encapsulating MPSC into an object.

This is just to be clearer to read.

When I asked you to review the code here was to review the API.

Now if we should pool or not it's an orthogonal discussion. The code right now is pooling and I would like it (tried it) to be more readable on its intention.

clebertsuconic: I would still create the Utility. document its usage...

I would rather have the pool.borrow in the code with the logic encapsulated. It is prettier.

clebertsuconic: @franz1981 I renamed the ObjectPool class, and documented the usage.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/objectPools/ObjectPool.java
```diff
@@ -0,0 +1,68 @@
...
+   public ObjectPool(int maxSize, Consumer<T> cleaner, Supplier<T> supplier) {
+      internalPool = createQueue(maxSize);
+      this.cleaner = cleaner;
+      this.supplier = supplier;
+   }
+
+   abstract Queue<T> createQueue(int maxSize);
+
+   /** Use this to instantiate or return objects from the pool */
+   public final T borrowObject() {
```
franz1981, line 46: I would just use `borrow`/`release` or `borrow`/`free`

franz1981 (approved):
LGTM!

franz1981 (approved):
LGTM

PR #3476, opened 2021-03-04
NO-JIRA: GHA workflow improvements

gemmellr:
- Builds the examples using the release profile on JDK8, as the Travis job does.
- Build the examples on JDK11+ using the -Pexamples,noRun profiles for equivalent build (release profile requires JDK8), which the Travis build doesnt do.
- Adds a workflow_dispatch trigger, allowing manual start of a new test runs (on any branch), without overwriting all the previous run logs like using the 'rerun all jobs' button within a particular run result.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: "Build"
 
-on: [push, pull_request]
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   build:
@@ -32,8 +32,15 @@ jobs:
       - name: Build Main
         run: |
           mvn -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Djdk8-errorprone -Pfast-tests -Pextra-tests -Ptests-CI -Pjmh install
-      - name: Build Examples
+
+      - name: Build Examples (JDK8 / -Prelease)
+        if: matrix.java == '8'
         run: |
           cd examples
-          mvn install -PnoRun
+          mvn install -Prelease
 
+      - name: Build Examples (JDK 11+ / -Pexamples,noRun)
+        if: matrix.java != '8'
+        run: |
+          cd examples
+          mvn install -Pexamples,noRun


.github/workflows/build.yml
```diff
@@ -32,8 +32,15 @@ jobs:
       - name: Build Main
         run: |
           mvn -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Djdk8-errorprone -Pfast-tests -Pextra-tests -Ptests-CI -Pjmh install
-      - name: Build Examples
+
+      - name: Build Examples (JDK8 / -Prelease)
+        if: matrix.java == '8'
```
clebertsuconic, line 37: @gemmellr before last release, I had to fix javadoc, however currently javadoc only builds on JDK 1.8

I tried to do an if jdk=8 at the time but I did not know how to do it?

Since you figured it out now, can we add after this line?

```mvn -Prelease javadoc:javadoc```


WDYT?

clebertsuconic: I wanted to prevent people breaking javadocs again.. or at least be aware when that happens.

gemmellr:
Essentially yep... If you mean for the main build you would probably add a
new separate named step for Javadoc, before the examples steps, i.e copy
the block rather than just add that line to it.

Also, though the release profile needs JDK8, could you just do
javadoc:javadoc on all JDKs without the profile? We certainly had that
building on 8-14 when we first introduced ability to build with >8, I fixed
lots of javadoc errors at least (warnings still there, but don't fail the
build).

Either way, I'd expect the javadoc to build on them all eventually, and as
the newer ones are typically stricter it's probably those that most need
checked. So even if it were needed for now, an 'if JDK8' restriction on
that should probably be removed in time.

I also actually thought it always built at least a good chunk of the
overall javadoc, since it often looks to be one of the things really
slowing the Artemis build down overall. I've certainly wanted to suggest
disabling it in the default build before as it's so slow. Maybe it doesn't
build it all though, while at release it does?

gemmellr:
Annoying that email replies dont actually show up in the 'comment thread' that caused them hehe.

gemmellr:
> I also actually thought it always built at least a good chunk of the
overall javadoc, since it often looks to be one of the things really
slowing the Artemis build down overall. I've certainly wanted to suggest
disabling it in the default build before as it's so slow. Maybe it doesn't
build it all though, while at release it does?

That seems to be the case, the artemis-website module spends a chunk of time (~1.5min in CI) triggering building javadoc for select modules: artemis-core-client, artemis-jms-client, artemis-server, artemis-jms-server, artemis-journal, artemis-selector. Presumably the release does every module or builds them slightly differently, such that you encountered an error in an area that wasnt seen previously during the above artemis-website bits.

The artemis-website module is always building those javadoc bits currently, even though it isnt normally building the actual docs for the site, which it only does in the release profile. I dont see any we wouldn't just move the javadoc builds into the release profile too (or add another related profile to trigger it). Does anyone typically look at that output before release, given its available in the IDEs already if you want it during dev, say while you are editing it? After that you can then set up a either nightly build that verifies it (GHA has cron from scheduling), or something triggered by each commit. I tend to do the nightly, given it doesnt change much and is slow.

PR #3479, opened 2021-03-07
ARTEMIS-3163 Experimental support for Netty IO_URING incubator

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3163


artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
```diff
@@ -401,6 +406,21 @@ public ActiveMQThreadFactory run() {
             acceptorType = KQUEUE_ACCEPTOR_TYPE;
 
             logger.debug("Acceptor using native kqueue");
+         } else if (useIoUring && CheckDependencies.isIoUringAvailable()) {
```
michaelandrepearce, line 409: I would expect io uring to be primary option later with fallback to epoll and then nio, as such maybe this should be first in order of preference. Its off by default but this way going forwards correct preference order would exist.

franz1981:
These are my results by using a single single threaded acceptor for both clients and replication (on the live broker) to fairly compare epoll vs io_uring.
The test is similar to the one on https://issues.apache.org/jira/browse/ARTEMIS-2852 with 32 JMS core clients, 100 persistent bytes messages and IO_URING transport has been used *only* on the live server, leaving the rest as it is ie backup + clients

NOTE: These are just preliminary results, so I won't share HW configuration or anything to make this reproducible, but it should give the magnitude of improvement offered by io_uring.

`master`:
```
**************
EndToEnd Throughput: 22582 ops/sec
**************
mean               1410.83
min                 333.82
50.00%             1368.06
90.00%             1679.36
99.00%             2293.76
99.90%             3489.79
99.99%            13107.20
max               16187.39
count               320000
```
`this pr`:
```
**************
EndToEnd Throughput: 30540 ops/sec
**************
mean               1052.52
min                 329.73
50.00%             1007.62
90.00%             1286.14
99.00%             1736.70
99.90%             4653.06
99.99%            13893.63
max               16711.68
count               320000
```
The profile data collected with https://github.com/jvm-profiling-tools/async-profiler/ are attached on https://issues.apache.org/jira/browse/ARTEMIS-3163

But the important bits are:

- Replication event loop thread: 935 (epoll) vs 775 (io_uring) samples -> ~94% cpu usage vs 78% cpu usage
- SYSCALLs samples: 
`epoll`: ~61% samples
image
`io_uring`: ~31% samples
image

The io_uring version is far more efficient while using resources then epoll despite our replication process already try to batch writes as much as possible to amortize syscall cost: would be interesting to compare epoll with some OpenOnLoad kernel bypass driver vs io_uring :P

*IMPORTANT*:
Why I've chosen to use a single thread for everything?
Don't be tempted to leave the default configuration, because it uses 3 * available cores for the replication/client acceptors: the io_uring version is that much efficient then epoll then the Netty event loops tends to go idle most of the time and need to be awaken, causing application threads to always pay the cost to wakeup event loop threads...this can make the io_uring version to look worse then epoll, while is right the opposite(!!)

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
franz1981: Yes :) correct

franz1981: Damn karaf and its feature checks :O it works for reflection too? :O

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/CheckDependencies.java
```diff
@@ -54,4 +54,17 @@ public static final boolean isKQueueAvailable() {
          return false;
       }
    }
+
+   public static boolean isIoUringAvailable() {
+      try {
+         return Env.isLinuxOs() && (boolean) (Class.forName("io.netty.incubator.channel.uring.IOUring")
```
michaelandrepearce, line 60: Should just call IOUring.isAvailable() like we do with kqueue and epoll, dont do this via reflection

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/ActiveMQClientLogger.java
```diff
@@ -420,6 +425,11 @@
            format = Message.Format.MESSAGE_FORMAT)
    void unableToCheckEpollAvailabilitynoClass();
 
+   @LogMessage(level = Logger.Level.WARN)
+   @Message(id = 212079, value = "IoUring is not available, please add to the classpath or configure useIoUring=false to remove this warning",
```
michaelandrepearce, line 429: Please put in correct place in order or the code. So people know what the high watermark for next code is without needing to hunt. All others are in asc order, please follow

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/ActiveMQClientLogger.java
```diff
@@ -395,6 +395,11 @@
            format = Message.Format.MESSAGE_FORMAT)
    void unableToCheckEpollAvailability(@Cause Throwable e);
 
+   @LogMessage(level = Logger.Level.WARN)
+   @Message(id = 212080, value = "Unable to check IoUring availability ",
```
michaelandrepearce, line 399: Please put in correct order place in file based on id. See other comment for why

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
```diff
@@ -112,6 +113,7 @@
    public static final String NIO_ACCEPTOR_TYPE = "NIO";
    public static final String EPOLL_ACCEPTOR_TYPE = "EPOLL";
    public static final String KQUEUE_ACCEPTOR_TYPE = "KQUEUE";
+   public static final String IOURING_ACCEPTOR_TYPE = "IO_URING";
```
michaelandrepearce, line 116: Why not just IOURING we dont add under score seperator for KQueue or EPoll. Just also avoids any escaping issues on urls or other things in config urls or admin console

michaelandrepearce:
Docs need updating.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/CheckDependencies.java
franz1981: It's not available via maven afaik so it's up to the users to compile it from source and use it

michaelandrepearce: <dependency> <groupId>io.netty.incubator</groupId> <artifactId>netty-incubator-transport-native-io_uring</artifactId> <version>0.0.3.Final</version> <classifier>linux-x86_64</classifier> </dependency>

michaelandrepearce: Its available. These is actually in the docs on the netty github

michaelandrepearce: In fact 0.0.4 is latest

https://mvnrepository.com/artifact/io.netty.incubator/netty-incubator-transport-native-io_uring/0.0.4.Final

franz1981: Wow ok, will remove the reflection then, thanks to have checked!

michaelandrepearce:
@franz1981 whats status on this one? Im keen to merge it, happy to help contribute any last bits like slight code re-org on if statements, and docs that are needed myself, if needed next week.

franz1981:
>  whats status on this one? Im keen to merge it, happy to help contribute any last bits like slight code re-org on if statements, and docs that are needed myself, if needed next week.

That seems a nice addition: i just would like to perform some better testing to help users to know which kernel version (and incubator version) to use. Let it parks here for a week and then we can move on and maybe decide with a public vote if people are interested and would like to know what it is/its purpose. We can have a call too with the community to explain it :)

jbertram:
@franz1981, just a friendly reminder that it's been a week. This looks like a nice feature so it would be nice to get it out to the community if it's ready.

franz1981:
@jbertram 
> just a friendly reminder that it's been a week. This looks like a nice feature so it would be nice to get it out to the community if it's ready.

I don't know yet if we should bring this in without any reflection usage (as @michaelandrepearce suggested) and we need to add a doc paragraph that  state we have no idea (yet) if SSL or other features works as expected ie EXPERIMENTAL tag on everything.
If any of the community people would like to raise the appropriate issues *after* this will be merged I can happily work on Netty side too, to fix things.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/TransportConstants.java
```diff
@@ -61,6 +61,8 @@
 
    public static final String USE_KQUEUE_PROP_NAME = "useKQueue";
 
+   public static final String USE_IOURING_PROP_NAME = "useIoUring";
```
gemmellr, line 64: Your recent comments Franz noted flagging everything with caution that its experimental...this might be a simple spot to trivially bang the point home to a user, making the option name reflect it...e.g "useIoUringExperimental"

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
gemmellr: You can probably add some metadata to artemis-server-osgi let it know the io.netty.incubator.channel.uring package usages are optional.

clebertsuconic:
I think we should run the extended suite using io_uring... if it's pass.. we should merge it upstream.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/TransportConstants.java
michaelandrepearce: I disagree, that should just be clear in the docs, else we end up with mangled config.

michaelandrepearce:
@franz1981 my intent was to send you PR with changes, had a disaster work week, im hoping to get to it next week for you.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/TransportConstants.java
gemmellr: Having experimental bits using explicit non-final options or even APIs prior to completion to signal their status isnt mangled config IMO. Its fairly common, e.g the JDK has many instances of this.

michaelandrepearce:
@franz1981 i sent to your branch which this is PR from, a PR with my proposed changes to address the comments i had raised in this PR, see: https://github.com/franz1981/activemq-artemis/pull/15 , if you agree with those then feel free to merge to your branch, and i have no further comments.

jbertram:
@franz1981, @clebertsuconic, @michaelandrepearce, where are we on this? Is this something we still want to do? Netty's io_uring lib is still in their incubator, but it's up to 0.16 now.

michaelandrepearce:
@jbertram i would like to still see this in, as long as we mark it experimental or in incubation in docs, i personally support this.

jbertram:
@michaelandrepearce, given the silence from @franz1981 I'd say you should go ahead and send a PR of your own with the changes you recommended (plus resolving the existing conflicts). Then we can move forward. Thanks!

PR #3482, opened 2021-03-08
ARTEMIS-3168 - add PrincipalConversionLoginModule feature

gtully:


michaelandrepearce:
Looks like a nice feature but not 100% how to use. It be great if an example was added.

gtully:
what configuration doc section are you thinking about? 
It is covered in the JAAS login module section in security.md and the test demonstrates configuration.

Fair point on the example. 
I came across this need when playing with org.keycloak.adapters.jaas.DirectAccessGrantsLoginModule
Let me try and wrangle that into an example and update this PR.

gtully:
updated with an example that pulls it together using keycloak. I hope the readme.md is informative and reads well, feedback appreciated. Thanks in advance!

gtully:
fixed up some check style and json excluded from rat check.

gtully:
updated to only download keycloak on run

gtully:
any other feedback, I would appreciate a read of the example readme.md - i tweaked a few typos just now.

PR #3484, opened 2021-03-10
ARTEMIS-3141 - respect the browse page limit on all queue controll/jm…

gtully:
…x operations that use a queue browser


clebertsuconic:
nice one

PR #3486, opened 2021-03-11
ARTEMIS-3174: Set new connection on ServerSession during reattachment

meierhofer08:
During session reattachment, also set the new connection on the
"session" member of the ServerSessionPacketHandler. Until now,
the connected ServerSessionImpl instance still referenced the old
connection although it had already been transferred on the new connection.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
@@ -1018,6 +1018,8 @@ public class ServerSessionPacketHandler implements ChannelHandler {
 
       newConnection.syncIDGeneratorSequence(remotingConnection.getIDGeneratorSequence());
 
+      session.transferConnection(newConnection);
+
       Connection oldTransportConnection = remotingConnection.getTransportConnection();
 
       remotingConnection = newConnection;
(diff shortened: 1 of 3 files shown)


jbertram:
Could you possibly add a test to validate this fix and mitigate any regressions in the future?

meierhofer08:
That would be good of course. The only problem is that I'm not familiar with the testing structure of Artemis/how to set up mocks for testing etc. It would help if you could point me into a direction for where to place the test and how to achieve what I want to test.

The test would need to:
- Set up an Artemis core client connection to a broker with a confirmationWindowSize != -1
- Simulate connection loss to broker, the client has to detect the connection failure ("Connection failure to ... has been detected")
- After less than the connectionTTL (default 60 seconds), allow connection to broker again -> client fails over
- Client tries to reattach all the sessions, server transfers the sessions
- Afterwards, check if the remotingConnection is set correctly (has the new connectionID) on every object connected to the session

It could even be simplified to a "component test" only on the broker where you have a session with an "old" connection, and then do a "ServerSessionPacketHandler.transferConnection" suppyling a new connection. Afterwards, check if the all the objects related to the sessionHandler (itself, the channel, the serverSession) have the new connection assigned.

If you provide me with enough details, I might be able to write the test. Otherwise, it would maybe be easier if you write the test as you know the code in more detail.

jbertram:
Like most functionality in the broker there was an existing test that was simple to modify so I added a test and merged this PR. Thanks for the contribution, @meierhofer08!

meierhofer08:
Ok perfect, thank you for merging and ensuring that there is a test for the change!

PR #3488, opened 2021-03-11
ARTEMIS-3176 don't log JDBC datasource 'password' property

jbertram:


artemis-jdbc-store/src/main/java/org/apache/activemq/artemis/jdbc/store/drivers/JDBCDataSourceUtils.java
```diff
@@ -29,7 +29,7 @@ public static DataSource getDataSource(String dataSourceClassName, Map<String, O
       ActiveMQJournalLogger.LOGGER.initializingJdbcDataSource(dataSourceClassName, dataSourceProperties
          .keySet()
          .stream()
-         .map(key -> key + "=" + dataSourceProperties.get(key))
+         .map(key -> key + "=" + (key.equals("password") ? "****" : dataSourceProperties.get(key)))
```
brusdev, line 32: Could the key be uppercase?

jbertram: It would only be uppercase if [`org.apache.activemq.artemis.core.config.storage.DatabaseStorageConfiguration#getDataSource`](https://github.com/apache/activemq-artemis/blob/master/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/storage/DatabaseStorageConfiguration.java#L170) changes. It's hard-coded to use `password`. 

I thought about making it a constant, but to be able to use it from `org.apache.activemq.artemis.jdbc.store.drivers.JDBCDataSourceUtils` I'd have to move it (and `org.apache.activemq.artemis.core.config.StoreConfiguration`) out of the `artemis-server` module and into the `artemis-commons` module.

brusdev: Could a user set the password using a `data-source-property` with an uppercase property?
<data-source-property key="Password" value="artemis" />

jbertram: @brusdev, yes, I would assume that's theoretically possible. I will update the PR to use `equalsIgnoreCase` instead of `equals`.

brusdev (approved):
LGTM +1

PR #3490, opened 2021-03-12
ARTEMIS-3175 - implement address setting management-message-attribute…

gtully:
…-size-limit to sensibly limit data returned by list/browse/filter management ops

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/ICoreMessage.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/ICoreMessage.java
@@ -102,10 +102,11 @@ public interface ICoreMessage extends Message {
 
    /**
     * @return Returns the message in Map form, useful when encoding to JSON
+    * @param valueSizeLimit
     */
    @Override
-   default Map<String, Object> toMap() {
-      Map map = toPropertyMap();
+   default Map<String, Object> toMap(int valueSizeLimit) {
+      Map map = toPropertyMap(valueSizeLimit);
       map.put("messageID", getMessageID());
       Object userID = getUserID();
       if (getUserID() != null) {
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
@@ -708,7 +708,15 @@ public interface Message {
     * @return Returns the message in Map form, useful when encoding to JSON
     */
    default Map<String, Object> toMap() {
-      Map map = toPropertyMap();
+      return toMap(-1);
+   }
+
+   /**
+    * @return Returns the message in Map form, useful when encoding to JSON
+    * @param valueSizeLimit that limits [] map values
+    */
+   default Map<String, Object> toMap(int valueSizeLimit) {
+      Map map = toPropertyMap(valueSizeLimit);
       map.put("messageID", getMessageID());
       Object userID = getUserID();
       if (getUserID() != null) {
@@ -728,6 +736,14 @@ public interface Message {
     * @return Returns the message properties in Map form, useful when encoding to JSON
     */
    default Map<String, Object> toPropertyMap() {
+      return toPropertyMap(-1);
+   }
+
+   /**
+    * @return Returns the message properties in Map form, useful when encoding to JSON
+    * @param valueSizeLimit that limits [] map values
+    */
+   default Map<String, Object> toPropertyMap(int valueSizeLimit) {
       Map map = new HashMap<>();
       for (SimpleString name : getPropertyNames()) {
          Object value = getObjectProperty(name.toString());
@@ -735,11 +751,11 @@ public interface Message {
          if (value instanceof SimpleString) {
             value = value.toString();
          }
+         value = JsonUtil.truncate(value, valueSizeLimit);
          map.put(name.toString(), value);
       }
       return map;
    }
-
 
    /** This should make you convert your message into Core format. */
    ICoreMessage toCore();
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
@@ -127,6 +127,8 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
 
    public static final boolean DEFAULT_ENABLE_METRICS = true;
 
+   public static final int MANAGEMENT_MESSAGE_ATTRIBUTE_SIZE_LIMIT = 256;
+
    private AddressFullMessagePolicy addressFullMessagePolicy = null;
 
    private Long maxSizeBytes = null;
@@ -253,6 +255,8 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
 
    private Boolean enableMetrics = null;
 
+   private Integer managementMessageAttributeSizeLimit = null;
+
    //from amq5
    //make it transient
    private transient Integer queuePrefetch = null;
@@ -318,6 +322,7 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
       this.defaultGroupFirstKey = other.defaultGroupFirstKey;
       this.defaultRingSize = other.defaultRingSize;
       this.enableMetrics = other.enableMetrics;
+      this.managementMessageAttributeSizeLimit = other.managementMessageAttributeSizeLimit;
    }
 
    public AddressSettings() {
@@ -914,6 +919,15 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
       return this;
    }
 
+   public int getManagementMessageAttributeSizeLimit() {
+      return managementMessageAttributeSizeLimit != null ? managementMessageAttributeSizeLimit : AddressSettings.MANAGEMENT_MESSAGE_ATTRIBUTE_SIZE_LIMIT;
+   }
+
+   public AddressSettings setManagementMessageAttributeSizeLimit(int managementMessageAttributeSizeLimit) {
+      this.managementMessageAttributeSizeLimit = managementMessageAttributeSizeLimit;
+      return this;
+   }
+
    /**
     * merge 2 objects in to 1
     *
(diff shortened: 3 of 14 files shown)


gtully:
along with the feature, there were a few little bug fixes, like byte[] properties not being parsed to json or small message bodies on text messages being null. Those scenarios are covered in the tests.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
```diff
@@ -726,20 +728,44 @@ default long getWholeMessageSize() {
...
          }
+         value = truncate(value, valueSizeLimit);
          map.put(name.toString(), value);
       }
       return map;
    }
 
+   static Object truncate(final Object value, final int valueSizeLimit) {
+      Object result = value;
+      if (valueSizeLimit > 0) {
```
michaelandrepearce, line 749: I would expect truncate 0 to give me zero, normally expect to not truncate or unlimited to be denoted by -1

artemis-jms-client/src/test/java/org/apache/activemq/artemis/jms/client/ConversionTest.java
```diff
@@ -36,7 +36,7 @@ public void testCoreToJMSConversion() {
       clientMessage.setDurable(true)
               .setPriority((byte) 9)
               .setExpiration(123456);
-      Map<String, Object> messageMap = clientMessage.toMap();
+      Map<String, Object> messageMap = clientMessage.toMap(0);
```
michaelandrepearce, line 39: See other comment but semantically i would 0 to truncate to 0, unlimited i would expect to call -1, likewise typically i would use toMap() to keep interface unchanged.

artemis-server/src/main/java/org/apache/activemq/artemis/core/transaction/impl/CoreTransactionDetail.java
```diff
@@ -56,6 +56,6 @@ public String decodeMessageType(Message msg) {
 
    @Override
    public Map<String, Object> decodeMessageProperties(Message msg) {
-      return msg.toMap();
+      return msg.toMap(256);
```
michaelandrepearce, line 59: Surely this could break existing case. I would expect this to remain unlimited

michaelandrepearce: Or at least configurable with default of unlimited

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -127,6 +127,8 @@
 
    public static final boolean DEFAULT_ENABLE_METRICS = true;
 
+   public static final int MANAGEMENT_MESSAGE_ATTRIBUTE_SIZE_LIMIT = 256;
```
michaelandrepearce, line 130: Surely want existing behaviour of unlimited so users dont get shocks.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -253,6 +255,8 @@
 
    private Boolean enableMetrics = null;
 
+   private Integer managementMessageAttributeSizeLimit = AddressSettings.MANAGEMENT_MESSAGE_ATTRIBUTE_SIZE_LIMIT;
```
michaelandrepearce, line 258: Should be null. So things like merge of address settings dont merge unset values.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
```diff
@@ -706,9 +707,10 @@ default long getWholeMessageSize() {
 
    /**
     * @return Returns the message in Map form, useful when encoding to JSON
+    * @param valueSizeLimit
     */
-   default Map<String, Object> toMap() {
```
michaelandrepearce, line 710: This is a breaking api change on core. This is exposed to users

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
```diff
@@ -726,20 +728,44 @@ default long getWholeMessageSize() {
 
    /**
     * @return Returns the message properties in Map form, useful when encoding to JSON
+    * @param valueSizeLimit
     */
-   default Map<String, Object> toPropertyMap() {
+   default Map<String, Object> toPropertyMap(int valueSizeLimit) {
```
michaelandrepearce, line 733: This is a breaking api change to users of core

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
```diff
@@ -726,20 +728,44 @@ default long getWholeMessageSize() {
...
          if (value instanceof SimpleString) {
             value = value.toString();
          }
+         value = truncate(value, valueSizeLimit);
          map.put(name.toString(), value);
       }
       return map;
    }
 
+   static Object truncate(final Object value, final int valueSizeLimit) {
```
michaelandrepearce, line 747: This is private logic. Currently by doing this it be exposed on api class. Do we really want to expose this? I would make this in an internal class and call that method from default methods

michaelandrepearce:
This looks quite risky in change of core messages etc, if scope is meant just for management console/cli etc,  Not sure we need to change such core internals just for management api layer. Surely it can be handled at a layer above.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/ICoreMessage.java
```diff
@@ -102,10 +102,11 @@
 
    /**
     * @return Returns the message in Map form, useful when encoding to JSON
+    * @param valueSizeLimit
     */
    @Override
-   default Map<String, Object> toMap() {
-      Map map = toPropertyMap();
+   default Map<String, Object> toMap(int valueSizeLimit) {
```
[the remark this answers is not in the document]
gtully: makes sense.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
gtully: ok, -1 it will be.

artemis-jms-client/src/test/java/org/apache/activemq/artemis/jms/client/ConversionTest.java
gtully: agree

artemis-server/src/main/java/org/apache/activemq/artemis/core/transaction/impl/CoreTransactionDetail.java
gtully: on the default. I want the broker to be constrained by default but yes the focus is the management api b/c that returns all the data. I want to fix it such that the broker returns 'some' reasonable amount of data  to ui/management/control by default.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
gtully: nope. If there is a 1k property, we don't want that 1k on the browser or on the jmx command line as ascii. The fact that we currently display the full values is a bug imho.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
gtully: the override will sort that and split usage between those who care and those who don't.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
gtully: fair, will fix.
thanks for all the feedback, it is fantastic.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
gtully: ok, that makes sense. thanks

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
michaelpearce-gain: i cannot resolve my comment but treat this as resolved.

artemis-server/src/main/java/org/apache/activemq/artemis/core/transaction/impl/CoreTransactionDetail.java
michaelpearce-gain: Yes but this is decode and isnt just for ui....

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
michaelpearce-gain: Yes but this is in interface, we have to keep toMap()

artemis-server/src/main/java/org/apache/activemq/artemis/core/transaction/impl/CoreTransactionDetail.java
gtully: yep. that is back to the original

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
gtully: yep. sorted. toMap is now untouched.

michaelpearce-gain: oops miss-read sorry, re read it and youre sorting it, cool.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
gtully: providing overrides with the limit leaves existing api as now. that is better. thanks!

gtully:
I think this is good to go. the core api remains intact and the overrides with limits are used from the jmx/control points to enforce the limits that can protect the broker and UI's. Thanks for the feedback!

PR #3492, opened 2021-03-15
ARTEMIS-3180 - fix multiple path match case in wildcard address map

gtully:


sebthom:
I can confirm that the changes solve the issue we are facing with Artemis 2.0.17.

gtully:
thanks for verification. I should have a clean test run in a few hours and can push this fix.

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/postoffice/impl/AddressMapUnitTest.java
```diff
@@ -638,7 +638,7 @@ public void testHashA() throws Exception {
       assertFalse(aABCA.matches(aHashA));
       assertFalse(aHashA.matches(aABCA));
 
-      assertEquals(0, countMatchingWildcards(abcaS));
+      assertEquals(1, countMatchingWildcards(abcaS));
```
michaelandrepearce, line 641: This doesnt seem right. I would expect existing test case to pass the same without modification

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/postoffice/impl/AddressMapUnitTest.java
```diff
@@ -650,7 +650,7 @@ public void testHashA() throws Exception {
       SimpleString AHashA = new SimpleString("a.#.a");
       underTest.put(AHashA, AHashA);
 
-      assertEquals(1, countMatchingWildcards(new SimpleString("a.b.c.a")));
+      assertEquals(2, countMatchingWildcards(new SimpleString("a.b.c.a")));
```
michaelandrepearce, line 653: Ditto comment i would expect existing test case to return the same

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/postoffice/impl/AddressMapUnitTest.java
```diff
@@ -875,7 +936,7 @@ public void testMax() throws Exception {
       underTest.put(new SimpleString("#.a"), new SimpleString("#.a"));
 
       assertEquals(3, countMatchingWildcards(new SimpleString("test.a")));
-      assertEquals(1, countMatchingWildcards(new SimpleString("test.a.a")));
+      assertEquals(3, countMatchingWildcards(new SimpleString("test.a.a")));
```
michaelandrepearce, line 939: Again this shouldn't change

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/postoffice/impl/AddressMapUnitTest.java
sebthom: Isn't this test essentially about if "#.a" matches "a.b.c.a"? Which I think it should.

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/postoffice/impl/AddressMapUnitTest.java
sebthom: Here "#.a" and "a.#.a" are matched against "a.b.c.a". Both should match, given that as per doc "# match any sequence of zero or more words"

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/postoffice/impl/AddressMapUnitTest.java
michaelandrepearce: My issue is that would be a change of behaviour, that some users may have got used to and built systems around. As such just changing that may break many users.

If theres a change in behaviour it needs to be behind a toggle and then by default kept as original.

gtully: @michaelandrepearce this is a bug fix for a regression, the tests verify the behaviour of the existing code but the semantic was wrong in the tests.

sebthom (approved):

PR #3493, opened 2021-03-16
ARTEMIS-3182 remove defunct repo

jbertram:
This repo is no longer operative. See
https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/
for more info. All dependencies are in Maven Central.


gemmellr:
Merged. The commit log is very slightly inaccurate but it will become spot on eventually hehe (repo is still around for a bit yet).

I ran the build locally with a fresh local repo to double check things since the CI build will have things cached, it was all good.

PR #3495, opened 2021-03-16
spacing in logmessage

erwindon:
a specific log message lists a few terms. 2 of these terms do not have any spacing between them as I noticed in actual output,
2 additional spacing adjustments for consistency.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/TopologyMemberImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/TopologyMemberImpl.java
@@ -154,7 +154,7 @@ public final class TopologyMemberImpl implements TopologyMember {
 
    @Override
    public String toString() {
-      return "TopologyMember[id = " + nodeId + ", connector=" + connector + ", backupGroupName=" + backupGroupName + ", scaleDownGroupName=" + scaleDownGroupName + "]";
+      return "TopologyMember[id=" + nodeId + ", connector=" + connector + ", backupGroupName=" + backupGroupName + ", scaleDownGroupName=" + scaleDownGroupName + "]";
    }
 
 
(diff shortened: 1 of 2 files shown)


artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/services/toolbar.js
```diff
@@ -25,7 +25,7 @@ var Artemis;
                     <form class="toolbar-pf-actions">
                         <div class="form-group toolbar-pf-filter">
                             <div class="input-group">
-                                <div class="input-group-btn">
+                                <div class="input-group-btn" style="padding-left: 10px">
```
clebertsuconic, line 28: @andytaylor can you review this?

brusdev:
If you are applying the same style `padding-left: 10px` to all divs with the `input-group-btn` class you could define that class in the https://github.com/apache/activemq-artemis/blob/master/artemis-hawtio/activemq-branding/src/main/webapp/plugin/css/activemq.css

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/TopologyMemberImpl.java
```diff
@@ -154,7 +154,7 @@ public URI toBackupURI() {
 
    @Override
    public String toString() {
-      return "TopologyMember[id = " + nodeId + ", connector=" + connector + ", backupGroupName=" + backupGroupName + ", scaleDownGroupName=" + scaleDownGroupName + "]";
+      return "TopologyMember[id=" + nodeId + ", connector=" + connector + ", backupGroupName=" + backupGroupName + ", scaleDownGroupName=" + scaleDownGroupName + "]";
```
brusdev, line 157: Is this change on purpose?

erwindon:
I have to apologize deeply here!
Due to a faulty rebase, this PR contained also code from another PR.
None of the HTML changes were supposed to be here.
That has now been corrected.

artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/services/toolbar.js
erwindon: sorry, the HTML code should not have been here, see comment below

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/TopologyMemberImpl.java
erwindon: yes, though cosmetic. the other items in the same message do not have spacing around the "="

PR #3496, opened 2021-03-16
visually separate the controls in groups

erwindon:
The "filter" toolbar above many pages has all controls without spacing.
This PR inserts a little bit of spacing to make the groups of controls more obvious.

old situation:
afbeelding

new situation:
afbeelding

spacing is applied 3 times:
1) at the start of the row to separate it from the left margin
2) at "sort order" to separate the "filter" group from the "ordering" group
3) at "columns" to separate it from "reset"

since the "looking glass" being an icon, no extra spacing was needed there

note: I tried to use "1em" as spacing, but that did not work, so this PR uses "10px"

### DIFF
--- a/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/services/toolbar.js
+++ b/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/services/toolbar.js
@@ -25,7 +25,7 @@ var Artemis;
                     <form class="toolbar-pf-actions">
                         <div class="form-group toolbar-pf-filter">
                             <div class="input-group">
-                                <div class="input-group-btn">
+                                <div class="input-group-btn" style="padding-left: 10px">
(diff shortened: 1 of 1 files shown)


brusdev:
If you are applying the same style `padding-left: 10px` to all divs with the `input-group-btn` class you could define that class in the https://github.com/apache/activemq-artemis/blob/master/artemis-hawtio/activemq-branding/src/main/webapp/plugin/css/activemq.css

erwindon:
> If you are applying the same style `padding-left: 10px` to all divs with the `input-group-btn` class you could define that class in the https://github.com/apache/activemq-artemis/blob/master/artemis-hawtio/activemq-branding/src/main/webapp/plugin/css/activemq.css

the spacing is only applied to 3 of this class instances, not to all. there is no other information available that defines the groups. so the spacing is applied to 3 instances that are group "leaders" to give them a slight offset from the previous control.

jbertram:
Thanks for the contribution, @erwindon!

PR #3498, opened 2021-03-18
ARTEMIS-2870: Transfer connection close/failure listeners one by one …

meierhofer08:
…on reattachment

Previously, when a session was reattached, all the close/failure listeners
were removed from the old connection and set onto the new connection.
This only worked when at most 1 session of the old connection was
transferred: When the second session was transferred, the old
connection already didn't contain any close/failure listeners anymore,
and therefore the list of close/failure listeners was overwritten by
an empty list for the new connection.

Now, when a session is being transferred, it only transfers the
close/failure listeners that belong to it, which are the session itself
+ the TempQueueCleanerUppers.

Modified a test to check whether the sessions are failure listeners of
the new connection after reattachment.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/reattach/ReattachTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/reattach/ReattachTest.java
@@ -17,6 +17,7 @@
 package org.apache.activemq.artemis.tests.integration.cluster.reattach;
 
 import java.util.Objects;
+import java.util.Set;
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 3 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/reattach/ReattachTest.java
```diff
@@ -133,18 +134,31 @@ public void testReattachTransferConnectionOnSession() throws Exception {
       locator.setRetryInterval(retryInterval).setRetryIntervalMultiplier(retryMultiplier).setReconnectAttempts(reconnectAttempts).setConfirmationWindowSize(1024 * 1024);
       ClientSessionFactoryInternal sf = (ClientSessionFactoryInternal) createSessionFactory(locator);
       ClientSession session = sf.createSession(false, true, true);
+      ClientSession secondSession = sf.createSession(false, true, true);
```
clebertsuconic, line 137: no need to make this change... but I'm creating a second test on top of this one

Your change is going into testReattachTransferConnectionOnSession2().



I will do this as part of the merge and amending this as you.. you will still be the author of the commit.

meierhofer08: Ah ok, a separate test also makes sense, I wasn't sure as they both effectively test "transferConnection" in "ServerSessionImpl".
Thank you for checking and merging so quickly!

PR #3499, opened 2021-03-18
NO-JIRA Remove unused private logger field

sebthom:


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/Base64.java
```diff
@@ -152,8 +151,6 @@
     */
    private static final byte NEW_LINE = (byte) '\n';
 
-   private static final Logger logger = Logger.getLogger(Base64.class);
```
clebertsuconic, line 155: I wish Java was like Go.. something private like this without being used should been a compilation error.

Sure, you can use errorProne.. etc... but not as clear!

clebertsuconic: thanks!

PR #3502, opened 2021-03-21
chore: replace JDK 14 with JDK 16

sullis:


clebertsuconic:
should we replace, or add?

michaelandrepearce:
Why do we miss 15? After all 15 is actually a MTS version with some jdk providers. E.g. Zulu from Azul

clebertsuconic:
@michaelandrepearce that also...

shouldn't we add 15 and 16 instead.. or is there a reason to actually replace 14 by 16?

gemmellr:
The main reason is 16 is the 'current' non-LTS release and 15(+14) is now out of support in many (but not necessarily all, as noted) cases.

I think it makes most sense to only test LTS + Current. We should also bear in mind that a good reason not test every JDK is simpy to not use the limited shared CI resources (CI service resources being shared by the entire apache org) for things of potentially less vale such as checking [semi-]unsupported JDKs.

I'd drop 14 and skip testing 15 at this point personally (already have elsewhere) but if you wanted to test it then I'd at least consider e.g swapping Travis jobs to target such 'side cases', or better yet use the Jenkins nodes where resources aren't as hotly contested to begin with. I wouldn't test any 12-16 versions once 17 comes out in September personally, which is another reason I wouldn't really bother testing <16 at this point either.

That said, the discussion on 16 itself is perhaps rather academic currently given it doesnt actually build on it, note the job failure. Its possible it could need a (long overdue?) build overhaul of plugin versions etc, as its complaining about the enforcer plugin, which I think I already upgraded previously when doing the work to build on > JDK8 and then had to downgrade it again as there was some intermittent interaction with other plugins from doing so that caused the build (not tests) to sporadically fail with newer ones.

gemmellr:
As Justin referenced above, I have raised another PR #3578 which updates the GHA (and additionally Travis) build job to use Java 16 rather than 14, along with changes to enable it to actually do such a build and test (mostly).

PR #3509, opened 2021-03-22
ARTEMIS-3198 Add concurrency option on core bridges

AntonRoskvist:
This commit adds a concurrency-option for the core bridges to handle high latency networks better

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/ClusterManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/ClusterManager.java
@@ -475,16 +475,21 @@ public final class ClusterManager implements ActiveMQComponent {
 
       clusterLocators.add(serverLocator);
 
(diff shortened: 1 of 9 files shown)


franz1981:
Nice contribution!!
Just curious, as we have made with ReplicationManager on https://github.com/apache/activemq-artemis/pull/3392 probably we are not correctly filling enough packets to maximize the TCP packet size and amortize network latencies: have you tried by disabling TCP no delay too?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/ClusterManager.java
```diff
@@ -475,16 +475,21 @@ public synchronized void deployBridge(final BridgeConfiguration config) throws E
...
-      managementService.registerBridge(bridge, config);
-
-      bridge.start();
-
-      if (server.hasBrokerBridgePlugins()) {
-         server.callBrokerBridgePlugins(plugin -> plugin.afterDeployBridge(bridge));
+      for (int i = 0; i < config.getConcurrency(); i++) {
+         Bridge bridge = new BridgeImpl(serverLocator, config.getInitialConnectAttempts(), config.getReconnectAttempts(),
+               config.getReconnectAttemptsOnSameNode(), config.getRetryInterval(), config.getRetryIntervalMultiplier(),
+               config.getMaxRetryInterval(), nodeManager.getUUID(), new SimpleString(config.getName()), queue,
```
jbertram, line 481: You need to change the name of each bridge (e.g. `config.getName() + "-" + i`) so that it's unique otherwise you won't be able to manage them (if necessary).

AntonRoskvist: Added unique naming for all bridges as you recommended, hope it looks good

AntonRoskvist:
> Nice contribution!!
> Just curious, as we have made with ReplicationManager on #3392 probably we are not correctly filling enough packets to maximize the TCP packet size and amortize network latencies: have you tried by disabling TCP no delay too?

Thanks,

No I haven't tried that and it looks like a real nice optimization, but for my use case I am looking to increase the throughput I am currently seeing by some 10-15 times at least. I will definitely try that out as well though!

PR #3512, opened 2021-03-23
NO-JIRA Save double property lookup on common case

franz1981:


clebertsuconic:
Looks good, but I think you need a JIRA on this one... you even wrote a test.. it doesn't seem a NO-JIRA to me.

franz1981:
Leaving this as a draft until a proper JIRA is created

jbertram:
@franz1981, is the only thing holding this back the lack of a Jira? If so, I can create one, update the commit message, and merge this. Let me know.

franz1981:
@jbertram 
>  is the only thing holding this back the lack of a Jira..

Nope, I still need to rebase and run the whole test suite on this

PR #3513, opened 2021-03-23
ARTEMIS-2877 ReplicationEndpoint can flush responses based on writability

franz1981:


franz1981:
Still performing some perf test on this :+1:

clebertsuconic:
@franz1981 why you closed it? I thought this had a significant improvement?

franz1981:
@clebertsuconic sadly not :(
Tests that were bottlenecked by replication cost will probably get a much better improvement by further batching packets on sender side instead, but I will send a separate PR for that after next release

PR #3515, opened 2021-03-23
Update activmq5 version to fix CVE-2021-26117

trevorlinton:
Fixes CVE-2021-26117 by upgrading activemq5 client to 5.16.1


brusdev:
this is a server's side issue that does not affect ActiveMQ Artemis

PR #3516, opened 2021-03-23
ARTEMIS-3204 Fixing NPE on Counting Queue for Resource Limit

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQMessageBundle.java
```diff
@@ -363,7 +363,7 @@
    ActiveMQSessionCreationException sessionLimitReached(String username, int limit);
 
    @Message(id = 229111, value = "Too many queues created by user ''{0}''. Queues allowed: {1}.", format = Message.Format.MESSAGE_FORMAT)
-   ActiveMQSessionCreationException queueLimitReached(String username, int limit);
+   ActiveMQSecurityException queueLimitReached(String username, int limit);
```
clebertsuconic, line 366: before anyone asks.. this was supposed to be a SecurityException :)

jbonofre (approved):
I confirm it works the issue (tested locally). Thanks !

PR #3519, opened 2021-03-25
ARTEMIS-3198 - Fix order of concurrency parameter...

AntonRoskvist:
...in schema file so it matches documentation

### DIFF
--- a/artemis-server/src/main/resources/schema/artemis-configuration.xsd
+++ b/artemis-server/src/main/resources/schema/artemis-configuration.xsd
@@ -1654,6 +1654,15 @@
             </xsd:annotation>
          </xsd:element>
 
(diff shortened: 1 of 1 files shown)


michaelandrepearce:
I would suggest rather than changing the order which will break existing installs the docs are simply updated.

AntonRoskvist:
It is not released yet, just sent that PR a few days back

clebertsuconic:
This has not been released yet.  What is the issue ?

michaelandrepearce:
@clebertsuconic there is no issue if it hasnt been released yet. As i said, disregard my comment if it hasn't.

clebertsuconic:
I'm completely lost on what was the issue Michael. I'm on the iPhone. I will review this later :)

clebertsuconic:
I even thought this was the original PR. So disregard anything I said also.

michaelandrepearce:
LGTM tbh.

PR #3526, opened 2021-04-02
ARTEMIS-3219 Improve FQQN message routing

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3219


franz1981:
This is both improving performance of FQQN routing by reducing the amount of map lookups to compute the next binding search and it's improving normal routing by removing from the hot path less common or slow path methods eg duplicate cache lookups.

I need to run an end 2 end test with FQQN before merging this + running the full CI, because it contains a big number of changes.

PR #3527, opened 2021-04-06
ARTEMIS-3221 Migrating to Jakarta EE 8 artifacts

brusdev:


gemmellr:
I havent actually properly looked at the changes, but skimming them it seems like you are going to be missing an update to the LICENCE file (and perhaps NOTICE file) that goes in the binary convenience distribution, given the aim here is to change a bunch of deps in ways which make them all go from using ALv2 to something else.

brusdev:
@gemmellr thank you for pointing me to LICENSE file, I have just updated it but I'm not sure how I should update the NOTICE file, do you have any suggestions?

artemis-distribution/src/main/resources/licenses/bin/LICENSE
```diff
@@ -432,4 +432,11 @@ For Underscore:
...
-https://github.com/jashkenas/underscore/blob/master/LICENSE
\ No newline at end of file
+https://github.com/jashkenas/underscore/blob/master/LICENSE
+
+==============================================================================
+For Jakarta EE 8:
+==============================================================================
+This product bundles Jakarta EE 8, which is available under a
+"EPL 2.0" license.  For details, see
+https://github.com/eclipse-ee4j/jakartaee-api/blob/8.0.0/LICENSE.txt
```
gemmellr, line 442: The actual licence needs to be in this file, or linked from it to another file elsewhere in the distribution. Links to external sites arent sufficient as they can come and go or change (say if the entire API stack moves to a different organisation)

(Yes this problem applies to the one right above it, which is also wrong, and further doesnt even link to a version as you try to)

We dont really use "jakartaee-api"/"Jakarta EE 8" either, only specific APIs, they should each be specifically called out that this applies to them. If the individual licence text of multiple items matches they can reuse the same definition.

gemmellr:
> thank you for pointing me to LICENSE file, I have just updated it but I'm not sure how I should update the NOTICE file, do you have any suggestions?

The NOTICE file should only convey 'required notices', and distributing the license text itself often already satisfies most 'must include a copy of <statement> type requirements, so changing NOTICE is often unnecessary, especially for using things that dont themselves have a NOTICE file. In this case, im not sure anything will need added, but its worth checking nothing needs removed, and just generally checking the file seems correct and up to date.

artemis-distribution/src/main/resources/licenses/bin/LICENSE
brusdev: I have just pushed a commit to add a `licenses` folder with the required files and to specify the Jakarta EE 8 APIs bundled.

artemis-distribution/src/main/resources/licenses/bin/LICENSE
```diff
@@ -432,4 +432,46 @@ For Underscore:
 ==============================================================================
 This product bundles Underscore, which is available under a
 "MIT" license.  For details, see
-https://github.com/jashkenas/underscore/blob/master/LICENSE
\ No newline at end of file
+https://github.com/jashkenas/underscore/blob/master/LICENSE
+
+==============================================================================
```
gemmellr, line 437: Looking at the existing file I would suggest adding all these additions above the previous 'grouping' of console bits to be clearer that they are separate, up at https://github.com/apache/activemq-artemis/blob/2.17.0/artemis-distribution/src/main/resources/licenses/bin/LICENSE#L247

artemis-distribution/src/main/resources/licenses/bin/LICENSE
```diff
@@ -432,4 +432,46 @@ For Underscore:
...
+For jakarta.activation-api:
+==============================================================================
+This product bundles Underscore, which is available under the
```
gemmellr, line 440: The entry is for jakarta.activation-api but it mentions including "Underscore", which was the entry above.

Same issue for all the entries that follow.

artemis-distribution/src/main/resources/licenses/bin/LICENSE
```diff
@@ -432,4 +432,46 @@ For Underscore:
...
+Eclipse Distribution License (EDL) 1.0. For details, see licenses/EDL-1.0.txt
+
+==============================================================================
+For jakarta.inject-api:
+==============================================================================
+This product bundles Underscore, which is available under the
+Apache License Version 2.0. For details, see licenses/EPL-2.0.txt
```
gemmellr, line 447: This says "Apache License Version 2.0" but then links to EPL-2.0. Which is it?

Note if it is ALv2 you dont need to mention it at all since its already covered.

artemis-distribution/src/main/resources/licenses/bin/LICENSE
brusdev: fixed

artemis-distribution/src/main/resources/licenses/bin/LICENSE
brusdev: removed

artemis-distribution/src/main/resources/licenses/bin/LICENSE
brusdev: I'm not sure we should move above `Apache ActiveMQ Artemis Subcomponents` https://github.com/apache/activemq-artemis/blob/2.17.0/artemis-distribution/src/main/resources/licenses/bin/LICENSE#L247
maybe above `For Angular` https://github.com/apache/activemq-artemis/blob/2.17.0/artemis-distribution/src/main/resources/licenses/bin/LICENSE#L264
WDYT?

gemmellr: I dont think it should go there as that would still be suggesting they are part of the 'management console component' demarcated by https://github.com/apache/activemq-artemis/blob/2.17.0/artemis-distribution/src/main/resources/licenses/bin/LICENSE#L256 when they aren't. I think above 247 makes sense.

brusdev: fixed

tests/integration-tests/pom.xml
```diff
@@ -370,12 +370,6 @@
          <artifactId>woodstox-core-asl</artifactId>
          <version>4.4.0</version>
          <scope>test</scope>
-         <exclusions>
```
ehsavoie, line 373: Why did you remove that exclusion ?

pom.xml
```diff
@@ -136,8 +137,7 @@
       <version.micrometer>1.6.1</version.micrometer>
       <hamcrest.version>2.1</hamcrest.version>
       <surefire.version>2.22.2</surefire.version>
-      <version.jaxb.api>2.3.1</version.jaxb.api>
-      <version.jaxb.runtime>2.3.3</version.jaxb.runtime>
+      <jaxb.runtime.version>2.3.3</jaxb.runtime.version>
```
ehsavoie, line 140: Why did you rename *version.jaxb.runtime* into *jaxb.runtime.version* ?

tests/integration-tests/pom.xml
brusdev: I have reverted it. I was missing that dependency in the dependency tree

pom.xml
brusdev: it was just to use same convection of other version properties but I don't need it in this commit so I have reverted it

clebertsuconic:
The commit message here says.. ARTEMIS-3215 Upgrade hawtio version to 2.13.2"


I guess you had some git-fu issue when you squashed the wrong comment?

clebertsuconic:
I will do a couple tests before merging it

clebertsuconic:
if you do mvn dependency:tree from integration-tests you still see Geronimo being pulled down. Is that expected?

clebertsuconic:
I still see a Geronimo usage from openwire server's protocol, and openwire clients...

if I look a mvn dependency:tree I still see this a couple times... but it seems to be the last one now?


[INFO] +- org.apache.activemq:artemis-junit:jar:2.18.0-SNAPSHOT:test
[INFO] +- org.apache.activemq:activemq-client:jar:5.16.0:test

brusdev:
I pushed a commit to fix it.

PR #3528, opened 2021-04-07
ARTEMIS-3219 Save allocating map entries during bindings iteration

franz1981:
That's a follow-up PR to complete the work done on https://issues.apache.org/jira/browse/ARTEMIS-3219


franz1981:
Merging this because sitting here completed from long time (already run CI on this).

PR #3530, opened 2021-04-08
ARTEMIS-3227 Web Console could be shutdown after certain failures

clebertsuconic:


clebertsuconic:
@michaelandrepearce  ^^

michaelpearce-gain:
@clebertsuconic nice thanks! appreciated as always

PR #3537, opened 2021-04-12
ARTEMIS-3236 Preserve managed queues on removing MQTT subscription

brusdev:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/imported/MQTTQueueCleanTest.java
```diff
@@ -31,6 +34,59 @@
...
+      server.addAddressInfo(new AddressInfo(address)
+                               .addRoutingType(RoutingType.MULTICAST)
+                               .setAutoCreated(false));
+
+      server.createQueue(new QueueConfiguration(queueName)
+                            .setAddress(address)
+                            .setRoutingType(RoutingType.MULTICAST)
+                            .setAutoCreated(false)
+                            .setAutoCreateAddress(false)
+                            .setAutoDelete(false));
```
clebertsuconic, line 52: Wouldn't the subscription be created by MQTT itself?

I understand the fix here would work if createQueue was called explicitly. But how a subscribe and disconnect would work with MQTT? Perhaps we should add an MQTT test as well?


Also: I would combine the two tests with a sub-module passing an argument (auto-create true or false).

That means, you create an internal method, and pass an argument with the difference in semantics.

brusdev: I have changed it, now the testsuite LGTM

PR #3541, opened 2021-04-13
ARTEMIS-3240 - ensure acked messages in transactions are returned to …

gtully:
…the sessions on connection failure, prior to rollback processing. Fix and test

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
@@ -682,6 +682,10 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
    @Override
    public void fail(ActiveMQException me, String message) {
 
+      for (Transaction tx : txMap.values()) {
+         tx.rollbackIfPossible();
+      }
+
       if (me != null) {
          //filter it like the other protocols
          if (!(me instanceof ActiveMQRemoteDisconnectException)) {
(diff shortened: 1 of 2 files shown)


gtully:
@clebertsuconic I was surprised that the scanreferences removed the messages as part of the ack, but it looks like this is expected as rollback puts them back. I added some logic to put them back via the failure handler, such that the transaction rollback has work to do. The test captures the use case, an ack in play before an server connection dies with a socket closed exception.

clebertsuconic:
@gtully can we talk over a google hangout tomorrow or some video call? I would expect also the rollback to handle these.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -698,6 +700,21 @@ public void fail(ActiveMQException me, String message) {
       shutdown(true);
    }
 
+   private void returnReferencesFromAnyPendingLocalTransactions() {
+      for (Transaction tx : txMap.values()) {
+         final AMQSession amqSession = (AMQSession) tx.getProtocolData();
+         if (amqSession != null) {
+            amqSession.getCoreSession().resetTX(tx);
+            try {
+               commandProcessorInstance.returnReferences(tx, amqSession);
```
clebertsuconic, line 709: this whole code here would be simply:

```java
   private void returnReferencesFromAnyPendingLocalTransactions() {
      for (Transaction tx : txMap.values()) {
         try {
            tx.rollback();
         } catch (Exception e) {
            logger.warn(e.getMessage());
         }
      }
   }
```

I don't know the reason for for returnReferences command? I think it's because open wire can redeliver messages locally without a roundtrip to the server, hence you need to make a distinction.

but on this case, a simple rollback on the transaction would suffice.

gtully: nice, that is simpler, I found rollbackIfPossible that will ignore pending xa. It fixes the issue. thanks. Will push an update with the change in a bit.

PR #3543, opened 2021-04-14
ARTEMIS-3106 Fixing javadoc build for artemis-amqp-protocol

clebertsuconic:
The reference between AMQP Protocol and Server was Indirect.
Not having the dependency explictly set ill break Javadoc.

I am reapplying the commit reverted to fix the javadoc build.

Revert "ARTEMIS-3106: remove duplicate artemis-server dependency added in 5313a800a37a4174a43289457a8934266b50f96b"

This reverts commit 3c90c695f3ac1a77f7ec981980bcd8d50bd72f25.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/pom.xml
+++ b/artemis-protocols/artemis-amqp-protocol/pom.xml
@@ -39,6 +39,11 @@
       </dependency>
       <dependency>
          <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-server</artifactId>
+         <version>${project.version}</version>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
          <artifactId>artemis-core-client</artifactId>
          <version>${project.version}</version>
       </dependency>


clebertsuconic:
@gemmellr you reverted this commit (that I am reapplying on this PR).

the dependency was added to fix the javadoc build, as the dependency to Artemis-server between AMQP and server was indirect.

Not having the dependency here will break javadoc build.


if we cannot have the dependency, we would need another approach to fix javadoc here.

gemmellr:
Thats odd. The dependency is already there. Maven even complains about it being duplicated and threatens it may break the build in future.

clebertsuconic:
It is odd actually...
I didn't realize it was already there.. all I was doing was fixing javadoc, and I thought it was not there.. hmm...

perhaps it's an ordering thing for javadoc build...?


I guess I could remove the last one, and keep the first.

gemmellr:
I guess its an ordering issue. The dependency was already there and is still there at: https://github.com/apache/activemq-artemis/blob/3c90c695f3ac1a77f7ec981980bcd8d50bd72f25/artemis-protocols/artemis-amqp-protocol/pom.xml#L85-L89 

Perhaps moving it up to where the one removed is would be the actual change. Though I have no idea why that should matter, I'd guess something else is arguably broken for that to matter.

gemmellr:
Cross post :)

clebertsuconic:
let me double check with and without the option.. to make sure I'm not screwing up things

gemmellr:
Actually, the build passed with my commit...are you sure its broken?

clebertsuconic:
not any more! :)

clebertsuconic:
let me double check.. sorry for the hassle

clebertsuconic:
@gemmellr can you pretend you never saw this PR? I don't know what happened... sorry about that...

gemmellr:
Hehe sure I can do that :)

laeubi:
> @gemmellr can you pretend you never saw this PR? I don't know what happened... sorry about that...

You should better make ASF Github / Jira bot more silent then as it sends out a mail for every comment here :-D

PR #3548, opened 2021-04-17
ARTEMIS-3256 internalqueue and internaladdress get their own color in the broker diagram

erwindon:
In the broker diagram, the icons for InternalAddress and InternalQueue are the same, as they each have the default style.
This PR explicitly assigns the same border colour as Address/Queue, but still leaves the fill colour white/transparent.

the new result:
broker-diagram
as produced by the code from this branch (and configuring a cluster so that there are actually internal objects)

previously both were same colour:
afbeelding

See also https://issues.apache.org/jira/browse/ARTEMIS-3256

### DIFF
--- a/artemis-hawtio/activemq-branding/src/main/webapp/plugin/css/activemq.css
+++ b/artemis-hawtio/activemq-branding/src/main/webapp/plugin/css/activemq.css
@@ -69,6 +69,14 @@ svg text {
    fill: #78932c;
 }
 
+.pf-topology-svg g.InternalAddress circle {
+   stroke:  #2b326e;
+}
+
+.pf-topology-svg g.InternalQueue circle {
+   stroke:  #50621d;
+}
+
 /*Adds a border to top of page*/
 .pf-c-page__header {
     border-top: 3px solid #B21054;


brusdev:
@erwindon nice enanchment, can you create an [issue](https://issues.apache.org/jira/projects/ARTEMIS/issues) for this enhancement and prepend the commit comment with the jira key (see [Typical development cycle](https://github.com/apache/activemq-artemis/blob/main/docs/hacking-guide/en/code.md#typical-development-cycle))?

clebertsuconic:
@erwindon you changed the title for the PR...

what you should instead is to change the title of the commit. (which is actually what matters).


git commit --amend

(on the editor.. change the name)


git push origin -f



@erwindon if you have some difficulty on doing that, I can use my gif-fu to amend the commit while merging.. let me know if you need help.

erwindon:
didn't I do both?
my middle name is "rebase", so that should not be a problem :-)
...
ah, I did that for the other PR that I raised but missed this one
...
one second...

clebertsuconic:
thank you

PR #3549, opened 2021-04-17
NO-JIRA fixed spelling error in code of broker diagram

erwindon:
while reading through the code for the broker diagram, I noticed 2 small inconsistencies. this PR fixes them.

note that this is based on a code review only. the fix does not seem to solve any real-world problems.


brusdev:
@erwindon good catch, can you prepend your commit comment with the `NO-JIRA` prefix?

clebertsuconic:
thanks for this @erwindon

PR #3550, opened 2021-04-19
Fix description of DiskStoreUsage

middagj:
In 73c4e399d9c2e8401f2034ffb0c5e42b4a70b053 a description is added to DiskStoreUsage. It incorrectly describes the diskStoreUsage as a percentage. This commit changes it to a fraction which it is (also before the description change). A percentage would be better, since MaxDiskUsage is also specified as percentage.


clebertsuconic:
I'm adding your comment to the commit message... thanks

jbertram:
In every other place where we report this metric (e.g. in the logs) we multiply it by 100 to make it a true percentage. The `max-disk-usage` is also configured as a percentage. I think it makes sense to just change `org.apache.activemq.artemis.core.server.files.FileStoreMonitor#calculateUsage` to multiply the result by `100` and then stop multiplying it everywhere else by `100`. We'll also need to *not* divide max-disk-usage by `100` when instantiating `FileStoreMonitor`. This will simplify the code by making the usage consistent throughout.

middagj:
I agree, but it is a backwards incompatible change. I was trying to add a monitor to it and can imagine other people also have something like in place. Although most users probably will be notified via their own monitoring if it is changed, since almost any disk usage in percentage is bigger than what they are currently using.

An idea is to introduce a `DiskUsage` that returns a percentage and deprecate `DiskStoreUsage`. Then the naming is also better aligned with `MaxDiskUsage`.

jbertram:
That's a fair point. If folks have alerts or something configured (e.g. in Prometheus or Grafana) then such a change would break those alerts.

PR #3555, opened 2021-04-26
ARTEMIS-2716 Implements pluggable Quorum Vote

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-2716

### DIFF
--- a/artemis-distribution/pom.xml
+++ b/artemis-distribution/pom.xml
@@ -231,6 +231,17 @@
          <version>${project.version}</version>
          <classifier>javadoc</classifier>
       </dependency>
+      <!-- quorum -->
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-quorum-api</artifactId>
+         <version>${project.version}</version>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-quorum-ri</artifactId>
+         <version>${project.version}</version>
+      </dependency>
       <dependency>
          <groupId>io.netty</groupId>
          <artifactId>netty-all</artifactId>
--- a/artemis-distribution/src/main/assembly/dep.xml
+++ b/artemis-distribution/src/main/assembly/dep.xml
@@ -62,6 +62,9 @@
             <include>org.apache.activemq.rest:artemis-rest</include>
             <include>org.apache.qpid:qpid-jms-client</include>
             <include>io.micrometer:micrometer-core</include>
+            <!-- quorum -->
+            <include>org.apache.activemq:artemis-quorum-api</include>
+            <include>org.apache.activemq:artemis-quorum-ri</include>
 
             <!-- dependencies -->
             <include>jakarta.jms:jakarta.jms-api</include>
@@ -97,6 +100,12 @@
             <include>com.sun.xml.bind:jaxb-impl</include>
             <include>jakarta.activation:jakarta.activation-api</include>
             <include>jakarta.security.auth.message:jakarta.security.auth.message-api</include>
+            <!-- quorum -->
+            <include>org.apache.curator:curator-recipes</include>
+            <include>org.apache.curator:curator-client</include>
+            <include>org.apache.curator:curator-framework</include>
+            <include>org.apache.zookeeper:zookeeper</include>
+            <include>org.apache.zookeeper:zookeeper-jute</include>
          </includes>
          <!--excludes>
             <exclude>org.apache.activemq:artemis-website</exclude>
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationBackupActivation.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationBackupActivation.java
@@ -1,0 +1,599 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.core.server.impl;
+
+import javax.annotation.concurrent.GuardedBy;
+
+import java.util.Objects;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Consumer;
+
+import org.apache.activemq.artemis.api.core.ActiveMQException;
+import org.apache.activemq.artemis.api.core.ActiveMQIllegalStateException;
+import org.apache.activemq.artemis.api.core.Pair;
+import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.api.core.TransportConfiguration;
+import org.apache.activemq.artemis.core.protocol.core.Channel;
+import org.apache.activemq.artemis.core.replication.ReplicationEndpoint;
+import org.apache.activemq.artemis.core.server.ActiveMQServer;
+import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
+import org.apache.activemq.artemis.core.server.LiveNodeLocator;
+import org.apache.activemq.artemis.core.server.NodeManager;;
+import org.apache.activemq.artemis.core.server.cluster.ClusterControl;
+import org.apache.activemq.artemis.core.server.cluster.ClusterController;
+import org.apache.activemq.artemis.core.server.cluster.ha.ReplicationBackupPolicy;
+import org.apache.activemq.artemis.quorum.DistributedLock;
+import org.apache.activemq.artemis.quorum.DistributedPrimitiveManager;
+import org.apache.activemq.artemis.quorum.UnavailableStateException;
+import org.jboss.logging.Logger;
+
+import static org.apache.activemq.artemis.core.server.impl.ReplicationObserver.ReplicationFailure;
+
+/**
+ * This activation can be used by a primary while trying to fail-back ie {@code failback == true} or
+ * by a natural-born backup ie {@code failback == false}.<br>
+ */
+public final class ReplicationBackupActivation extends Activation implements DistributedPrimitiveManager.UnavailableManagerListener {
+
+   private static final Logger LOGGER = Logger.getLogger(ReplicationBackupActivation.class);
+
+   private final boolean wasLive;
+   private final ReplicationBackupPolicy policy;
+   private final ActiveMQServerImpl activeMQServer;
+   // This field is != null iff this node is a primary during a fail-back ie acting as a backup in order to become live again.
+   private final String expectedNodeID;
+   @GuardedBy("this")
+   private boolean closed;
+   private final DistributedPrimitiveManager distributedManager;
+   // Used for monitoring purposes
+   private volatile ReplicationObserver replicationObserver;
+   // Used for testing purposes
+   private volatile ReplicationEndpoint replicationEndpoint;
+   // Used for testing purposes
+   private Consumer<ReplicationEndpoint> onReplicationEndpointCreation;
+   // Used to arbiter one-shot server stop/restart
+   private final AtomicBoolean stopping;
+
+   public ReplicationBackupActivation(final ActiveMQServerImpl activeMQServer,
+                                      final boolean wasLive,
+                                      final DistributedPrimitiveManager distributedManager,
+                                      final ReplicationBackupPolicy policy) {
+      this.wasLive = wasLive;
+      this.activeMQServer = activeMQServer;
+      if (policy.isTryFailback()) {
+         final SimpleString serverNodeID = activeMQServer.getNodeID();
+         if (serverNodeID == null || serverNodeID.isEmpty()) {
+            throw new IllegalStateException("A failback activation must be biased around a specific NodeID");
+         }
+         this.expectedNodeID = serverNodeID.toString();
+      } else {
+         this.expectedNodeID = null;
+      }
+      this.distributedManager = distributedManager;
+      this.policy = policy;
+      this.replicationObserver = null;
+      this.replicationEndpoint = null;
+      this.stopping = new AtomicBoolean(false);
+   }
+
+   /**
+    * used for testing purposes.
+    */
+   public DistributedPrimitiveManager getDistributedManager() {
+      return distributedManager;
+   }
+
+   @Override
+   public void onUnavailableManagerEvent() {
+      synchronized (this) {
+         if (closed) {
+            return;
+         }
+      }
+      LOGGER.info("Unavailable quorum service detected: try restart server");
+      asyncRestartServer(activeMQServer, true);
+   }
+
+   /**
+    * This util class exists because {@link LiveNodeLocator} need a {@link LiveNodeLocator.BackupRegistrationListener}
+    * to forward backup registration failure events: this is used to switch on/off backup registration event listening
+    * on an existing locator.
+    */
+   private static final class RegistrationFailureForwarder implements LiveNodeLocator.BackupRegistrationListener, AutoCloseable {
+
+      private static final LiveNodeLocator.BackupRegistrationListener NOOP_LISTENER = ignore -> {
+      };
+      private volatile LiveNodeLocator.BackupRegistrationListener listener = NOOP_LISTENER;
+
+      public RegistrationFailureForwarder to(LiveNodeLocator.BackupRegistrationListener listener) {
+         this.listener = listener;
+         return this;
+      }
+
+      @Override
+      public void onBackupRegistrationFailed(boolean alreadyReplicating) {
+         listener.onBackupRegistrationFailed(alreadyReplicating);
+      }
+
+      @Override
+      public void close() {
+         listener = NOOP_LISTENER;
+      }
+   }
+
+   @Override
+   public void run() {
+      synchronized (this) {
+         if (closed) {
+            return;
+         }
+      }
+      try {
+         LOGGER.info("Trying to reach majority of quorum service nodes");
+         distributedManager.start();
+         LOGGER.info("Quorum service available: starting broker");
+         distributedManager.addUnavailableManagerListener(this);
+         // Stop the previous node manager and create a new one with NodeManager::replicatedBackup == true:
+         // NodeManager::start skip setup lock file with NodeID, until NodeManager::stopBackup is called.
+         activeMQServer.resetNodeManager();
+         activeMQServer.getNodeManager().stop();
+         // A primary need to preserve NodeID across runs
+         activeMQServer.moveServerData(policy.getMaxSavedReplicatedJournalsSize(), policy.isTryFailback());
+         activeMQServer.getNodeManager().start();
+         if (!activeMQServer.initialisePart1(false)) {
+            return;
+         }
+         synchronized (this) {
+            if (closed)
+               return;
+         }
+         final ClusterController clusterController = activeMQServer.getClusterManager().getClusterController();
+         clusterController.awaitConnectionToReplicationCluster();
+         activeMQServer.getBackupManager().start();
+         ActiveMQServerLogger.LOGGER.backupServerStarted(activeMQServer.getVersion().getFullVersion(),
+                                                         activeMQServer.getNodeManager().getNodeId());
+         activeMQServer.setState(ActiveMQServerImpl.SERVER_STATE.STARTED);
+         final DistributedLock liveLock = replicateAndFailover(clusterController);
+         if (liveLock == null) {
+            return;
+         }
+         startAsLive(liveLock);
+      } catch (Exception e) {
+         if ((e instanceof InterruptedException || e instanceof IllegalStateException) && !activeMQServer.isStarted()) {
+            // do not log these errors if the server is being stopped.
+            return;
+         }
+         ActiveMQServerLogger.LOGGER.initializationError(e);
+      }
+   }
+
+   private void startAsLive(final DistributedLock liveLock) throws Exception {
+      activeMQServer.setHAPolicy(policy.getLivePolicy());
+
+      synchronized (activeMQServer) {
+         if (!activeMQServer.isStarted()) {
+            liveLock.close();
+            return;
+         }
+         ActiveMQServerLogger.LOGGER.becomingLive(activeMQServer);
+         // stopBackup is going to write the NodeID previously set on the NodeManager,
+         // because activeMQServer.resetNodeManager() has created a NodeManager with replicatedBackup == true.
+         activeMQServer.getNodeManager().stopBackup();
+         activeMQServer.getStorageManager().start();
+         activeMQServer.getBackupManager().activated();
+         // IMPORTANT:
+         // we're setting this activation JUST because it would allow the server to use its
+         // getActivationChannelHandler to handle replication
+         final ReplicationPrimaryActivation primaryActivation = new ReplicationPrimaryActivation(activeMQServer, distributedManager, policy.getLivePolicy());
+         liveLock.addListener(primaryActivation);
+         activeMQServer.setActivation(primaryActivation);
+         activeMQServer.initialisePart2(false);
+         // calling primaryActivation.stateChanged !isHelByCaller is necessary in case the lock was unavailable
+         // before liveLock.addListener: just throwing an exception won't stop the broker.
+         final boolean stillLive;
+         try {
+            stillLive = liveLock.isHeldByCaller();
+         } catch (UnavailableStateException e) {
+            LOGGER.warn(e);
+            primaryActivation.onUnavailableLockEvent();
+            throw new ActiveMQIllegalStateException("This server cannot check its role as a live: activation is failed");
+         }
+         if (!stillLive) {
+            primaryActivation.onUnavailableLockEvent();
+            throw new ActiveMQIllegalStateException("This server is not live anymore: activation is failed");
+         }
+         if (activeMQServer.getIdentity() != null) {
+            ActiveMQServerLogger.LOGGER.serverIsLive(activeMQServer.getIdentity());
+         } else {
+            ActiveMQServerLogger.LOGGER.serverIsLive();
+         }
+         activeMQServer.completeActivation(true);
+      }
+   }
+
+   private LiveNodeLocator createLiveNodeLocator(final LiveNodeLocator.BackupRegistrationListener registrationListener) {
+      if (expectedNodeID != null) {
+         assert policy.isTryFailback();
+         return new NamedLiveNodeIdLocatorForReplication(expectedNodeID, registrationListener, policy.getRetryReplicationWait());
+      }
+      return policy.getGroupName() == null ?
+         new AnyLiveNodeLocatorForReplication(registrationListener, activeMQServer, policy.getRetryReplicationWait()) :
+         new NamedLiveNodeLocatorForReplication(policy.getGroupName(), registrationListener, policy.getRetryReplicationWait());
+   }
+
+   private DistributedLock replicateAndFailover(final ClusterController clusterController) throws ActiveMQException, InterruptedException {
+      final RegistrationFailureForwarder registrationFailureForwarder = new RegistrationFailureForwarder();
+      // node locator isn't stateless and contains a live-list of candidate nodes to connect too, hence
+      // it MUST be reused for each replicateLive attempt
+      final LiveNodeLocator nodeLocator = createLiveNodeLocator(registrationFailureForwarder);
+      clusterController.addClusterTopologyListenerForReplication(nodeLocator);
+      try {
+         while (true) {
+            synchronized (this) {
+               if (closed) {
+                  return null;
+               }
+            }
+            final ReplicationFailure failure = replicateLive(clusterController, nodeLocator, registrationFailureForwarder);
+            if (failure == null) {
+               Thread.sleep(clusterController.getRetryIntervalForReplicatedCluster());
+               continue;
+            }
+            if (!activeMQServer.isStarted()) {
+               return null;
+            }
+            LOGGER.debugf("ReplicationFailure = %s", failure);
+            boolean voluntaryFailOver = false;
+            switch (failure) {
+               case VoluntaryFailOver:
+                  voluntaryFailOver = true;
+               case NonVoluntaryFailover:
+                  final DistributedLock liveLock = tryAcquireLiveLock();
+                  // from now on we're meant to stop:
+                  // - due to failover
+                  // - due to restart/stop
+                  assert stopping.get();
+                  if (liveLock != null) {
+                     return liveLock;
+                  }
+                  boolean restart = true;
+                  if (voluntaryFailOver && isFirstFailbackAttempt()) {
+                     restart = false;
+                     LOGGER.error("Failed to fail-back: stopping broker based on quorum results");
+                  } else {
+                     ActiveMQServerLogger.LOGGER.restartingAsBackupBasedOnQuorumVoteResults();
+                  }
+                  // let's ignore the stopping flag here, we're in control of it
+                  asyncRestartServer(activeMQServer, restart, false);
+                  return null;
+               case RegistrationError:
+                  LOGGER.error("Stopping broker because of critical registration error");
+                  asyncRestartServer(activeMQServer, false);
+                  return null;
+               case AlreadyReplicating:
+                  // can just retry here, data should be clean and nodeLocator
+                  // should remove the live node that has answered this
+                  LOGGER.info("Live broker was already replicating: retry sync with another live");
+                  continue;
+               case ClosedObserver:
+                  return null;
+               case BackupNotInSync:
+                  LOGGER.info("Replication failure while initial sync not yet completed: restart as backup");
+                  asyncRestartServer(activeMQServer, true);
+                  return null;
+               case WrongNodeId:
+                  LOGGER.error("Stopping broker because of wrong node ID communication from live: maybe a misbehaving live?");
+                  asyncRestartServer(activeMQServer, false);
+                  return null;
+               default:
+                  throw new AssertionError("Unsupported failure " + failure);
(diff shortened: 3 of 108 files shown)


artemis-distribution/pom.xml
```diff
@@ -183,6 +183,73 @@
            <artifactId>tomcat-servlet-api</artifactId>
        </dependency>
 
+      <!-- for artemis-quorum-ri -->
```
michaelpearce-gain, line 186: wheres the ZK alternative option dependencies?

franz1981: eagle eyes :D adding them now, together with features

michaelandrepearce: I still dont spot them. Unresolving. I could be being blind

artemis-distribution/src/main/assembly/dep.xml
```diff
@@ -106,6 +110,27 @@
             <include>com.sun.xml.bind:jaxb-core</include>
             <include>jakarta.activation:jakarta.activation-api</include>
             <include>jakarta.security.auth.message:jakarta.security.auth.message-api</include>
+            <!-- quorum -->
```
michaelandrepearce, line 113: Where are the zk deps here too?

artemis-distribution/src/main/assembly/dep.xml
```diff
@@ -125,6 +125,12 @@
             <include>org.objenesis:objenesis</include>
             <include>com.typesafe:config</include>
             <include>io.github.classgraph:classgraph</include>
+
+            <include>org.apache.curator:curator-recipes</include>
+            <include>org.apache.curator:curator-client</include>
+            <include>org.apache.curator:curator-framework</include>
```
franz1981, line 131: @michaelandrepearce these are the ZK deps

franz1981: Now out of office eh, still need to rebase vs master and include other deps for logging  and force 3.6.1 ZK client version

pom.xml
```diff
@@ -846,6 +853,106 @@
...
+            <groupId>com.typesafe</groupId>
+            <artifactId>config</artifactId>
+            <version>1.3.2</version>
+         </dependency>
+         <dependency>
+            <groupId>io.github.classgraph</groupId>
+            <artifactId>classgraph</artifactId>
+            <version>4.2.3</version>
+         </dependency>
+         <!-- Curator Zookeeper RI -->
```
franz1981, line 930: @michaelandrepearce here! I still lack logging deps and config

artemis-distribution/pom.xml
franz1981: see https://github.com/apache/activemq-artemis/pull/3555/files#r625567290 and https://github.com/apache/activemq-artemis/pull/3555/files#r625528887

artemis-distribution/src/main/assembly/dep.xml
franz1981: https://github.com/apache/activemq-artemis/pull/3555/files#r625528887

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationBackupActivation.java
```diff
@@ -0,0 +1,489 @@
...
+               throw new AssertionError("Unsupported failure " + failure);
+         }
+      }
+   }
+
+   private DistributedLock tryAcquireLiveLock() throws InterruptedException {
+      assert activeMQServer.getNodeManager().getNodeId() != null;
+      final String liveID = activeMQServer.getNodeManager().getNodeId().toString();
+      final int voteRetries = policy.getVoteRetries();
+      final long maxAttempts = voteRetries >= 0 ? (voteRetries + 1) : -1;
```
franz1981, line 239: @michaelandrepearce
@jbertram 
 It seems that `vote-retries` can be set to be `-1` (unbounded) and that it was used with the meaning that `vote-retries` == 0 is still an option...
I've decided to change the semantic to better match its name ie "retries" it seems to me (I'm not native eng, so please correct me if I'm wrong eh!) that are "additional" to the first mandatory "try".

michaelpearce-gain: so 

-1 normally means to me "unlimited" e.g. never stop

0 means dont do any retry, e.g. fail/error on first error, and do not re-attempt.

michaelpearce-gain: in regards to replication, i would typically expect.

-1 normally means wait until all members have acked.
0 means do not wait
1 would wait till 1 member acked.

franz1981: Not sure about the "members" count: I believe it's the amount of retrying to get the majority of votes (or the majority of the maximum seen members? Not sure) that would agree that the previous live is down for real - for the "old" replication. 
Here the semantic is similar, but how a completed/successfull vote is going to happen is an implementation detail of the "plugged" quorum vote algorithm aka the distributed lock and it's not just a check on the existing live role, but an atomic check-then-act to allow this backup to start failover too.
In order to make this a bit more resistant I have decided that both a failed quorum vote due to loss of quorum and a failed attempt to become live (because the existing live is still stopping or it's an impl detail of the quorum vote provider eg raft/zk leader re-election) are seen as retryable instead of failing fast. To fail fast the user must set the retries to 0.
According to the existing failover test this seems to play nice with the existing user experience.

franz1981: I see this doc related `vote-retries`:

> vote-retries
This property controls how many times the backup broker retries the quorum vote in order to receive a majority result that allows the backup broker to start up.

franz1981:
Sadly `canAcquireLockOnMajorityRestart` can still get some failures because of https://issues.apache.org/jira/browse/CURATOR-594, I'm awaiting the PR I've sent on https://github.com/apache/curator/pull/383 to be merged

franz1981:
I see that https://issues.apache.org/jira/browse/ARTEMIS-2297 can cause `ReplicationManager::sendSynchronizationDone` to behave differently with the pluggable quorum vote: 
the original JIRA was meant to force an isolated `master` to suicide to save split brain to happen; while the new quorum this is not needed because network isolation is already handled by the distributed lock listener, that should force the broker to suicide.
I probably need to modify `ReplicationManager` to ignore the quorum members count or just save `QuorumManager` to be created.

franz1981:
Some notes on the existing test failures:

- `org.apache.activemq.artemis.tests.smoke.quorum.PluggableQuorumSinglePairTest#testBackupFailoverAndPrimaryFailback` can get a failing test using `forceKill=false` (that's using a gentle SIGTERM to stop live primary): it's likely a bug/feature of `Atomix`while using Raft, because if a killed live was a Raft leader too, a soft kill would make it able to release its live lock and then crash, forcing a re-election and causing the failing over backup to "reconsider" the ownership of the live lock just obtained:
```
[main] 17:47:51,694 INFO  [org.apache.activemq.artemis.tests.smoke.quorum.PluggableQuorumSinglePairTest] killing primary
**********************************
Killing server java.lang.UNIXProcess@674658f7
**********************************
atomixReplicationBackup-out:2021-05-26 17:47:51,706 WARN  [org.apache.activemq.artemis.core.client] AMQ212037: Connection failure to localhost/127.0.0.1:61616 has been detected: AMQ219015: The connection was disconnected because of server shutdown [code=DISCONNECTED]
atomixReplicationBackup-out:2021-05-26 17:47:51,706 WARN  [org.apache.activemq.artemis.core.client] AMQ212037: Connection failure to localhost/127.0.0.1:61616 has been detected: AMQ219015: The connection was disconnected because of server shutdown [code=DISCONNECTED]
atomixReplicationBackup-out:2021-05-26 17:47:51,765 INFO  [org.apache.activemq.artemis.quorum.atomix.AtomixDistributedLock] Failed to acquire lock ac6f986d-be39-11eb-981a-8cc68169c75b
atomixReplicationBackup-out:2021-05-26 17:47:51,861 INFO  [org.apache.activemq.artemis.quorum.atomix.AtomixDistributedLock] Acquired lock ac6f986d-be39-11eb-981a-8cc68169c75b with version Version{version=10}
atomixReplicationBackup-out:2021-05-26 17:47:51,863 INFO  [org.apache.activemq.artemis.core.server] AMQ221037: ActiveMQServerImpl::serverUUID=ac6f986d-be39-11eb-981a-8cc68169c75b to become 'live'
atomixReplicationBackup-out:2021-05-26 17:47:51,872 WARN  [org.apache.activemq.artemis.core.client] AMQ212004: Failed to connect to server.
atomixReplicationBackup-out:2021-05-26 17:47:51,974 INFO  [org.apache.activemq.artemis.core.server] AMQ221080: Deploying address exampleTopic supporting [MULTICAST]
atomixReplicationBackup-out:2021-05-26 17:47:51,974 INFO  [org.apache.activemq.artemis.core.server] AMQ221080: Deploying address exampleQueue supporting [ANYCAST]
atomixReplicationBackup-out:2021-05-26 17:47:51,975 INFO  [org.apache.activemq.artemis.core.server] AMQ221003: Deploying ANYCAST queue exampleQueue on address exampleQueue
atomixReplicationBackup-out:2021-05-26 17:47:52,164 WARN  [org.apache.activemq.artemis.core.server.impl.ReplicationBackupActivation] org.apache.activemq.artemis.quorum.UnavailableStateException: io.atomix.primitive.PrimitiveException: java.net.ConnectException: Failed to connect to the cluster
atomixReplicationBackup-out:2021-05-26 17:47:52,164 ERROR [org.apache.activemq.artemis.core.server] AMQ224000: Failure in initialisation: ActiveMQIllegalStateException[errorType=ILLEGAL_STATE message=This server cannot check its role as a live: activation is failed]
atomixReplicationBackup-out:	at org.apache.activemq.artemis.core.server.impl.ReplicationBackupActivation.startAsLive(ReplicationBackupActivation.java:194) [artemis-server-2.18.0-SNAPSHOT.jar:2.18.0-SNAPSHOT]
atomixReplicationBackup-out:	at org.apache.activemq.artemis.core.server.impl.ReplicationBackupActivation.run(ReplicationBackupActivation.java:157) [artemis-server-2.18.0-SNAPSHOT.jar:2.18.0-SNAPSHOT]
atomixReplicationBackup-out:	at org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl$ActivationThread.run(ActiveMQServerImpl.java:4240) [artemis-server-2.18.0-SNAPSHOT.jar:2.18.0-SNAPSHOT]
atomixReplicationBackup-out:
atomixReplicationBackup-out:2021-05-26 17:47:55,018 WARN  [io.atomix.protocols.raft.roles.FollowerRole] RaftServer{data-partition-1}{role=FOLLOWER} - io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: localhost/127.0.0.1:7777
atomixReplicationBackup-out:2021-05-26 17:47:55,019 INFO  [io.atomix.protocols.raft.impl.RaftContext] RaftServer{data-partition-1} - Transitioning to CANDIDATE
atomixReplicationBackup-out:2021-05-26 17:47:55,021 INFO  [io.atomix.protocols.raft.roles.CandidateRole] RaftServer{data-partition-1}{role=CANDIDATE} - Starting election
atomixReplicationBackup-out:2021-05-26 17:47:55,032 WARN  [io.atomix.protocols.raft.roles.CandidateRole] RaftServer{data-partition-1}{role=CANDIDATE} - io.netty.channel.AbstractChannel$AnnotatedConnectException: finishConnect(..) failed: Connection refused: localhost/127.0.0.1:7777
atomixReplicationBackup-out:2021-05-26 17:47:55,034 INFO  [io.atomix.protocols.raft.impl.RaftContext] RaftServer{data-partition-1} - Transitioning to LEADER
atomixReplicationBackup-out:2021-05-26 17:47:55,041 INFO  [io.atomix.protocols.raft.impl.RaftContext] RaftServer{data-partition-1} - Found leader backup
[raft-server-data-partition-1] 17:47:55,052 INFO  [io.atomix.protocols.raft.impl.RaftContext] RaftServer{data-partition-1} - Found leader backup
[main] 17:47:55,283 INFO  [org.apache.activemq.artemis.tests.smoke.quorum.PluggableQuorumSinglePairTest] killed primary
```
It shows that 
```
atomixReplicationBackup-out:2021-05-26 17:47:51,765 INFO  [org.apache.activemq.artemis.quorum.atomix.AtomixDistributedLock] Failed to acquire lock ac6f986d-be39-11eb-981a-8cc68169c75b
atomixReplicationBackup-out:2021-05-26 17:47:51,861 INFO  [org.apache.activemq.artemis.quorum.atomix.AtomixDistributedLock] Acquired lock ac6f986d-be39-11eb-981a-8cc68169c75b with version Version{version=10}
```
live lock is acquired by the failing over backup, but
```
atomixReplicationBackup-out:2021-05-26 17:47:52,164 ERROR [org.apache.activemq.artemis.core.server] AMQ224000: Failure in initialisation: ActiveMQIllegalStateException[errorType=ILLEGAL_STATE message=This server cannot check its role as a live: activation is failed]
```
The lock is not considered available and the failover process is stopped.
The critical part seems:
```
atomixReplicationBackup-out:2021-05-26 17:47:55,034 INFO  [io.atomix.protocols.raft.impl.RaftContext] RaftServer{data-partition-1} - Transitioning to LEADER
```
It shows that backup has become the RAFT leader (given that live isn't around and that's ok, but: why backup hasn't suicide itself given that it's not able to start as a live? This maybe need some attention.

- `org.apache.activemq.artemis.tests.smoke.quorum.ZookeeperPluggableQuorumSinglePairTest` is hitting some thread leaks on `ListenerHandler` threads 
eg
```
Thread Thread[ListenerHandler-/127.0.0.1:38567,5,main] is still alive with the following stackTrace:
sun.misc.Unsafe.park(Native Method)
java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
java.util.concurrent.locks.AbstractQueuedSynchronizer$ConditionObject.await(AbstractQueuedSynchronizer.java:2039)
java.util.concurrent.LinkedBlockingQueue.take(LinkedBlockingQueue.java:442)
java.util.concurrent.ThreadPoolExecutor.getTask(ThreadPoolExecutor.java:1074)
java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1134)
java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
java.lang.Thread.run(Thread.java:748)
...

**Comment by michaelpearce-gain** [comment]
@franz1981 i think thats a blocker issue, you should be able to safely force kill, and also soft kill. I like the phrasing "bug/feature" LOL arent all bugs just features to be implemented :P

**Comment by franz1981** [comment]
@michaelpearce-gain agree. Just would like to understand if I can just force witness node to always be a raft leader, making it similar to a zookeeper external node. But seems a bit too limiting, maybe.
The other important bit, maybe my fault is
> why backup hasn't suicided itself given that it's not able to start as a live?

Tomorrow I am going to check this one instead.

**Comment by franz1981** [comment]
Update on 
> why backup hasn't committed suicide given that it's not able to start as a live?

Assuming a correct `Atomix` behaviour It seems related to a bug on my side:
```java
   private void startAsLive(final DistributedLock liveLock) throws Exception {
         // ...
         // IMPORTANT:
         // we're setting this activation JUST because it would allow the server to use its
         // getActivationChannelHandler to handle replication
         final ReplicationPrimaryActivation primaryActivation = new ReplicationPrimaryActivation(activeMQServer, distributedManager, policy.getLivePolicy());
         liveLock.addListener(primaryActivation);
         activeMQServer.setActivation(primaryActivation);
         activeMQServer.initialisePart2(false);
         final boolean stillLive;
         try {
            stillLive = liveLock.isHeldByCaller();
         } catch (UnavailableStateException e) {
            LOGGER.warn(e);
            throw new ActiveMQIllegalStateException("This server cannot check its role as a live: activation is failed");
         }
         if (!stillLive) {
            throw new ActiveMQIllegalStateException("This server is not live anymore: activation is failed");
         }
         // ...    
```
If the quorum is lost before `liveLock.addListener(primaryActivation)`, the current `AtomixDistributedLock::onStateChanged`:
```java
   private void onStateChanged(PrimitiveState state) {
      LOGGER.info(state);
      switch (state) {
         case SUSPENDED:
         case EXPIRED:
         case CLOSED:
            for (LockListener listener : listeners) {
               listener.stateChanged(LockListener.EventType.UNAVAILABLE);
            }
            break;
      }
   }
```  
It's going to find empty `listeners` and `ReplicationPrimaryActivation::stateChanged` won't be called to async stop the server.
The late check `liveLock.isHeldByCaller` should fail and throw an exception, but it won't cause the server to stop.

In short, the issue is that `liveLock.isHeldByCaller` must be able to stop the server and it should call ``AtomixDistributedLock::onStateChanged` in case of a lost lock, before throwing the exception.
This same timing issue could happen with Zookeeper too, so it worth to be fixed.

Re the Atomix issue instead, I'm going to investigate a bit more what's going on, because it doesn't seem that `AtomixDistributedLock::onStateChanged` has been called while ` liveLock.isHeldByCaller` has failed, that seems a bug.

franz1981:
I believe to have ported enough integration tests: I'm now going to add some more smoke tests, improve logs/configs and maybe prepare some doc

franz1981:
Re the thread leak on smoke tests on
> org.apache.activemq.artemis.tests.smoke.quorum.ZookeeperPluggableQuorumSinglePairTest is hitting some thread leaks on ListenerHandler threads

I see that https://github.com/apache/zookeeper/blob/c74658d398cdc1d207aa296cb6e20de00faec03e/zookeeper-server/src/main/java/org/apache/zookeeper/server/quorum/QuorumCnxManager.java#L915 seems a `Thread` created by Zookeeper: despite usiug a leak-free zk client version (see https://issues.apache.org/jira/browse/CURATOR-595) this seems a separate issue: going to take a better look why/what it is.

franz1981:
waiting https://issues.apache.org/jira/browse/ZOOKEEPER-4309 to be fixed and ported to any zk version

franz1981:
I decide 2 important things re this PR:

- going to drop Atomix support -community is not reactive and the solution itself doesn't seem solid as expected -
- there are some loop in the primary state diagram that can be dangerous:
*on primary start*: we loop forever searching some broker with the same Node ID to pair with (to failback) or we try acquire the live lock -> both should be attempted just once
*during failback*: if the failback process is going to fail, the primary is restarting as primary, getting into the loop mentioned above -> it should just stop

@jbertram @gtully Any opinons?

gtully:
I think you are right.
I would go further, and drop the fail-back feature in favour of a shutdown with delayed (after lock release sleep(epoch)) jvm exit.

Have a broker role be determined by the shared lock, the first to acquire becomes primary, the others potential backups.

The existing roles flow from the possibility of unilateral decisions, two primary brokers could be two live brokers in error. However with a reliable distributed lock and no unilateral decisions (exit once state is unknown) multiple competing brokers are fine.

This is a simpler model, easier to reason about and totally deterministic. Availability is tied to the quorum. To have more HA in the event of a lock service outage (ie quorum nodes fail), introduce another cluster of nodes.

michaelpearce-gain:
I disagree failback is very important, feature relied on by many, i dont think its something we can simply just drop.

gtully:
sorry, I need to give some more context, dropping failback b/c it is no longer necessary is what I am thinking and I am considering/assuming some additional state from the lock. 

Let me elaborate and see if this makes sense/holds water!

The bottom line here is: we must avoid unilateral decisions b/c our journal cannot reconcile concurrent writes on different nodes. ie: we cannot deal with split brain, so we must avoid it.
With a simple distributed lock, there is an implicit timeout or epoch. If we track the current epoch we can coordinate our local state transitions and avoid unilateral decisions.

in a simpler state model, a broker transitions from:
 BROKER - PRIMARY - REPLICATED
or:
 BROKER - BACKUP - IN_SYNC_REPLICA

On restart:
For the next 'in sequence' epoch the REPLICATED and IN_SYNC_REPLICA are identical, either can become PRIMARY and race. The winner becomes PRIMARY and the loser BACKUP.
In this case, delaying the exit of one can replicate or force 'failback'.

The more interesting case, if an epoch is skipped, our local state is stale, revert to BROKER and exit in the hope/knowledge/expectation that PRIMARY will restart. PRIMARY is no longer REPLICATED and the IN_SYNC_REPLICA is stale. PRIMARY must restart.

Here is the crux, we trade off availability for consistency. The only failover that can happen is from REPLICATED to an IN_SYNC_REPLICA

gtully:
in short, with a reliable quorum (distributed lock with epoch), we use it:

1. to handle the transition that we can currently support, one failover from primary to backup. 
2. to completely avoid split brain (that term is odd!) and trade availability.

If we can do this with 3nodes, replicating that setup (2 more nodes that share a witness node) can give us better HA

michaelpearce-gain:
So to avoid split brain is simple with distributed lock and ZK, only one process can hold that lock, and to be active you must have the lock, application with the lock should constantly be checking it still has the lock and if not disabling itself. You should never become active unless you get the lock, there should be no epoch about it, if you do that, you're asking for disaster.

With ZK how many artemis nodes you have makes no longer any difference to the avoiding split, you should be able to run just 2 artemis nodes for HA  (for ZK yes it requires 3 minimum) but those scale differently you're not scaling those for data plane use case.

michaelpearce-gain:
Maybe this is something better to be discussed with whiteboards and a community call?

gtully:
> You should never become active unless you get the lock, there should be no epoch about it, if you do that, you're asking for disaster.

what you want is to have the lock and the current data. The lock alone is not sufficient.

Consider this: 
primary starts, backup starts and becomes an in sync replica. Both die. primary comes back up for a while, and dies. backup restarts, It gets the lock but has stale data. boom!

The distributed lock only guards access to the journal, but something else (the epoch counter) has to guard the singular valid transition from primary to in sync backup. It is only valid if the lock is **next** held by the backup.

michaelpearce-gain:
So the instance shouldn't even be trying to even get the lock until it is back "back-in-sync", it should only be trying to get the lock in case it "could" take over.

franz1981:
I believe the epoch/version that should both mark the journal (or replicated journal) and a live role acquisition (hence should be something on ZK, indexed on the NodeID?) to be key to save any journal misalignment to happen, that's something even more important then classic split-brain: as @michaelpearce-gain already mentioned, the distributed lock should already prevent it to happen. Happy that both have raised a similar/same idea: this make me more convinced that something like that is the way to go.

Re this PR: I wanted to provide the above "enhancement" in a later phase really, as a separate improvement, while here I'm "just" trying to abstract away the quorum algorithm + providing a RI that allow a single pair to work fine + addressing some dark corners of the state diagram in the "classic" native quorum replication.

The loops I've mentioned above are my fault while trying to fill some of the gaps in the existing replication state transitions.
A brief summary here:

- *classic*: master's start by searching for existing live node with the same node ID to pair with and, if none are around, just starting straight as live -> this one is an easy cause of split brain with network partition
- *now*: primary's start involves a loop, first searching for existing live node with the same node ID to pair with and, if none is around, try just once to acquire the live lock before starting as live: if the live lock acquisition fail, retry to search existing live nodes etc etc

I think that this loop isn't needed, but could be performed just once ie search for other lives, if not found, try acquire live lock (for some time) and if failed, just exit.
It saves a completely isolated broker to keep on retrying to acquire the lock and/or search other lives forever. wdyt?

Another "hidden" loop:
- *classic*: during a master fail-back failure (see `SharedNothingBackupQuorum.BACKUP_ACTIVATION.FAILURE_REPLICATING` on the code) the broker is restarted as pure backup (no able to failback anymore!) deleting it's data folder too, including the node ID. 
That's dangerous because deleting the data folder will cause the master to loose it's identity (NodeID is going to change or restart) and if the node is manually restarted it will search for an non-existing node ID and able to start (SPLIT BRAIN served!).
- *new*: during a primary fail-back failure the broker is restarted as primary with its node ID preserved (so, looping to search for existing lives to pair with or trying to acquire live lock if not found any, in the loop mentioned above). 

The new approach is better because node ID is preserved and there's no risk a primary node restart with a backup acting as a live will ever cause a split brain to happen, but still, restarting as a primary while isolated can still cause the start process to last forever, while a failback should be a all-or-nothing mechanism IMO: the primary should try to failback to the existing live and just failing if not, without "restarting", wdyt?

michaelpearce-gain:
If the primary was restarted whilst isolated, it shouldnt be able to get the lock from ZK....... im clearly missing something.

mattrpav:
> community call

I'd be interested in attending if a community call on this is held.  Thanks

franz1981:
@mattrpav Happy to hear that let me schedule something for the next week, everyone interested can join :) 
@gtully @michaelpearce-gain @jbertram @clebertsuconic are invited too, just need to check my calendar and decide which tech use for the call ;)

franz1981:
@gtully @michaelpearce-gain @clebertsuconic @jbertram @mattrpav The proposal for the community call is for next Wed (09) on 15 UK time: will write here or on slack a link for the call.

Some of the subjects of the call are about:
- key differences from "classic" replication process
- improvements to handle on this PR or delay as separate enhancements 
eg Ideas about journal epoch/versioning: worth doing it now or later? Discuss about issue and solutions
- some key decision/discussions on existing behaviours

The last one is very important to move on with this PR, and the behaviour I was thinking are:

1. classic replication: single pair with pinger vs multiple pairs using quorum -> former deactivate/activate master based on network availability, latter just stop master on cluster connection drop (see https://github.com/apache/activemq-artemis/blob/0b95cb7754e1ae732f1ae9a3ddef657ea13ac90d/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/SharedNothingLiveActivation.java#L263) -> which behaviour to apply on the pluggable quorum vote? impl both and make it configurable?
2. classic replication: failure during backup failover (including inability to become live due to network split/majority agreement) cause it to restart as a pure backup (it includes the failback case) -> what's I've impl now and what should be better to do instead?

**If new things will popup on my head I'm going to drop them here, but first replies if the time/day of the call is fine ;)**

michaelpearce-gain:
@franz1981 thanks, yes this time slot works fine for me, i have reserved it out in my calendar, please let us know dial in link once youve sorted.

franz1981:
@michaelpearce-gain @gtully @jbertram @clebertsuconic @brusdev @mattrpav 
*10th Wed 15 UK time*
The meeting link is: https://meet.google.com/eve-hnos-ikg
The calendar link: https://calendar.google.com/event?action=TEMPLATE&tmeid=Mzg5bGM3MXNmZXVpZ2Jta3VsNjllN3FuNHAgbmlncm8uZnJhQG0&tmsrc=nigro.fra%40gmail.com

franz1981:
hi @mattrpav can I share on youtube our community video? I'm going to link it in a dev Apache post too

mattrpav:
> hi @mattrpav can I share on youtube our community video? I'm going to link it in a dev Apache post too

Sure thing

franz1981:
To summarize the current status of this PR, re

### Comparison with classic replication
Same as classic replication:
- fail-fast (JVM-wise) live on quorum loss: retrying to acquire lock as soon as quorum is restored can cause journal misalignments, see https://issues.apache.org/jira/browse/ARTEMIS-3340 - network pinger act differently

Differently from classic replication:
- failing-back primary preserve its node ID during journal rotation on activation start
- failing-back primary sync with live brokers with its same node ID
- fail-fast (JVM-wise) failing-back primary due to unavailable quorum/live still alive: classic replication cause it to restart as backup, loosing its node ID. On restart, it will cause it to activate as live (!) although with a different node ID.

gtully:
it may make sense to categorise those difference as existing bugs and give them jiras.
on the third one, it looks like a very risky operation, without some coordination and ARTEMIS-3340. It is impossible for the primary to know if the backup has shutdown if it does not get a reply.

franz1981:
@gtully 
Good idea about the bug JIRAs!
> on the third one, it looks like a very risky operation

You mean the classic behaviour ie restarting as backup while forgetting node ID, or,
You mean the new behaviour ie fail fast while preserving node ID instead of retrying?

> It is impossible for the primary to know if the backup has shutdown if it does not get a reply to the shutdown request.

These failures I have mentioned are just while the primary failing-back broker believe to have an up to date journal, is proceeding to fail-over, and failing, because of missing consensus or no majority of nodes around (both classic and new replication).

In the original version, it was "safely" restarting as backup, but forgetting it's node ID:
- If a user restart it, it's going to become a live broker with a different node ID (clients can connect to it)

In the new version it's restarted by user, it's preserving its node ID:
- If the backup it's still around as live with the same node ID, the primary will search for it and try to failback again
- if the backup is not around, it will start as live but with a misaligned journal

My change on new aim to reduce the chance of a misbehaviour, but is not solving completely the misalignment issue, as said.

franz1981:
Still WIP, I'm trying to add a new mechanism to list for unavailable managers and see if It improves backup reliability @gtully

franz1981:
@gtully @michaelandrepearce 
Next week I'm going 2 weeks on PTO, so I would like to move this one first.
Today I'm going to add some other bits of doc and smoke tests: other improvements on both doc, tests and changes in the state diagram will be pushed after this.
The failback case has been made as much as possible able to fail-fast, but I see that this should cover several other parts we didn't cover with classic replication too eg

- while awaiting the live to appear: is it supposed to wait forever? IMO,nope
- while trying to connect to its designated live: it's supposed to retry forever? IMO, nope

These 2 points requires to define the failback operation a bit differently from what was before, but it's something that could be addressed later, given that what I'm already delivering is on par (or better) then classic replication.

franz1981:
@gtully 
I've added some docs and this is ready for review and eventually merged, assuming there's still a lot of work to do on this, as the linked JIRAs already explain.

michaelandrepearce:
Well why shouldn't it try for ever to connect. In both situations. All you are doing by it not is making it manual intervention which you operationally want to not do. And what is an operator to do. Restart it.... so your not getting any win.

franz1981:
> All you are doing by it not is making it manual intervention which you operationally want to not do. And what is an operator to do. Restart it.... so your not getting any win.

Are you talking about primary start loop, now missing, or about the automatic activation/deactivation of primary because of unavailable quorum service?

The latter has been discussed during the community call and the reason why it hasn't been implemented was because of missing journal versioning: without journal versioning there is no point to quickly restart, because it just increase the chance of journal misalignments.

We can safely introduce it later, when the version allow to check if the broker is not just live, but with the most up to date journal.
Right now, it's not suggested to quickly restart a primary without checking who was the last broker alive: the same with classic replication.

I'm not against self-healing but I believe this should be introduced with the right feature to make it reliable, or ca became self-misaligning.

michaelandrepearce:
Im confused here because i thought we agreed on the call that versioned journal was needed and the assumption i took there was that would then need to be implemented before we move forwards

franz1981:
During the call we have agreed that I would have raised the JIRAs for the missing features to enhance it and making it more robust in front of journal misalignments scenario, if compared to che classic replication.

The already delivered values here are:
- support for single pair too
- many fixes on some dark corners of the original state diagram
- easy migration path (configuration-wise)
- pluggable quorum service

Users can already use what we have here getting same or better value then classic replication.

In the meantime we can:
- collect feedbacks on it as it is
- design journal versioning as an enhancement
- any other improvements we need 

Agree about not been in hurry, but given the magnitude of this feature I believe delivering it in chunks/iteratively is the way to go (assuming each chunk to have its value).

I was sure I have explained it during the call

@gtully any comment?

michaelandrepearce:
So i thought we had a data state issue and thus a solution was must before we went ahead. As such the need for the journal versioning to address that. Im very cautious about us delivering this to main and to users without a solution to that issue as its asking for production issues

gtully:
I think what we have here is a very good line in the sand. the problem that a lock version solves for a computer can more easily be solved by a capable human. What we have here is a clear definition of the problem. This is a nice incremental improvement.

michaelandrepearce:
And here lies the biggest issue, the human. This is what and why this was something to be solved, because too often than not, the human cannot make the right decision, starts accidentally the wrong node, so forth and so on.
And thus why we had the call to find the solution. Now the agreed solution from call was the journal solution proposed by youself, but we do need something. If we dont think journal solution now work, then we should probably have a follow up call. but i suspect thats not the case here.

gtully:
this PR is just a core part of the solution, https://issues.apache.org/jira/browse/ARTEMIS-3340 can layer over the versioning.

michaelandrepearce:
Indeed, but i think we need both parts before we goto main, id be concerned about this going into main before that as then we could accidentally release just one part and people getting themselves into an issue in prod, without both parts being done, and protecting people from themselves.

gtully:
@michaelandrepearce I hear you. there is more work to do for sure, but it won't be in any way easier to shoot ones self in the foot with these changes. 
It is all about collaboration, it is difficult to get folks to commit to code on a PR or even a feature branch. 

The reality is that If it is not on main, it does not exist. 

This feature is ripe for collaboration because it pulls in a distributed lock service, getting more users of that would be great to shape the API. The journal exclusive lock is just one use case (all be it an important one).

michaelandrepearce:
Right now in testing this, yes i have been building and running it, i have started the wrong node after taking it all down, and thus ended up running the older journal, right now with leader follower (mstr, slv) setups, because of the leader (mstr) concept you always know which node should be taken down last and backup first. And on restart if you do accidentally start slave you're safe, because it wont start till master is there.

michaelandrepearce:
If it does go to main, then it should have docs written with very explicitly this is experimental not for production and work in progress.

franz1981:
> If it does go to main, then it should have docs written with very explicit statements (big red/orange banner like) that this is experimental not for production and work in progress.

+100 agree
It would ease the pain of using a separate branch chasing main and would ease collaboration IMO (it's just personal).
I would like to add a comment on this one in the doc + a couple of other minor changes ie I would prefix all ZK live locks with `namespace/locks/` to prepare for the journal version change (that will likely use something like `namespace/longs/` prefixes

@gtully 
I believe we should doc somewhere the data misalignments issue of classic replication too (not just the JIRA with the fix using a journal version) and what are the expected mitigation actions to help: the pluggable version will going to see a fix for this, but for classic probably not, hence it worths to help users to be aware of it IMO.

michaelandrepearce:
Def want to be able to give artemis zk a base i the tree (configurable) after all many places will share a zk with other stuff or one zk for multiple brokers

franz1981:
Yep, that's already implemented using the "namespace" parameter of the manager, but I would like to already separate locks from "others" (journal timestamps or any other primitives we will need in the future)

franz1981:
@michaelandrepearce @gtully 
I hove both added the ZK default  locks path (but I'm playing with ZK to see if it satisfies me) and the warning for users mentioning the upstream JIRA that explain the data misalignment issue.
I'm going to add a doc JIRA to produce some meaningful mitigation actions for replication in general, given that `classic` replication is affected by this, but is very unlikely that's gonna be fixed.

gtully:
I have pulled all of the above into a new PR and added in the activation sequencing. I think we can move forward on that.

https://github.com/apache/activemq-artemis/pull/3646

franz1981:
Closing as being superseded by https://github.com/apache/activemq-artemis/pull/3646

PR #3557, opened 2021-04-27
ARTEMIS-3271 Improve Critical Analyzer to sample single therads

clebertsuconic:


gtully:
looks good. the test captures the failure nicely. I guess the sampling is not unlike the current behaviour where multiple threads can modify the state. At least now there won't be any false positives. 
also good that we are down to a single value to update for a single thread. nice.

gtully:
Thinking out loud, if we want to improve on sampling, and check every thread enter/exit, some sort of xor of the thread id and the analyser doing mark (to capture the current bit mask) and sweep if it has not changed since the mark, could work without recourse to tracking nanno time. Something for the future maybe.

clebertsuconic:
I don't think we need to check on every thread. I was just checking around, and most uses are OK for this.

PR #3561, opened 2021-04-30
ARTEMIS-3258 added warning for url parameters that are explicitly ignored

erwindon:
When setting up a downstream-federation, several ssl-related parameters are silently stripped from the provided URL.
This PR causes a warning to be shown for each parameter that was removed.

Only parameter name is mentioned, but not the given value as that is not needed and even insecure since that information then ends up in the logfile.

This helps to solve connection problems in this area; and it encourages a cleaner configuration file.

a test gave the following output:
```
2021-05-01 00:49:09,971 WARN  [org.apache.activemq.artemis.core.client] AMQ212079: The upstream connector from the downstream federation will ignore url parameter sslEnabled
2021-05-01 00:49:09,971 WARN  [org.apache.activemq.artemis.core.client] AMQ212079: The upstream connector from the downstream federation will ignore url parameter trustStorePath
2021-05-01 00:49:09,971 WARN  [org.apache.activemq.artemis.core.client] AMQ212079: The upstream connector from the downstream federation will ignore url parameter trustStorePassword
```

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/config/federation/FederationDownstreamConfiguration.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/config/federation/FederationDownstreamConfiguration.java
@@ -21,6 +21,7 @@ import java.util.Map;
 import java.util.Objects;
 import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
 import org.apache.activemq.artemis.api.core.TransportConfiguration;
+import org.apache.activemq.artemis.core.client.ActiveMQClientLogger;
 import org.apache.activemq.artemis.core.remoting.impl.netty.TransportConstants;
 
 public class FederationDownstreamConfiguration extends FederationStreamConfiguration<FederationDownstreamConfiguration> {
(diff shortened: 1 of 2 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/config/federation/FederationDownstreamConfiguration.java
```diff
@@ -40,23 +41,30 @@ public TransportConfiguration getUpstreamConfiguration() {
       return upstreamConfiguration;
    }
 
+   private void stripParam(Map<String, Object> params, String name) {
+	   Object oldValue = params.remove(name);
+	   if (oldValue != null) {
+		   ActiveMQClientLogger.LOGGER.ignoredParameterForDownstreamFederation(name);
+	   }
```
brusdev, line 48: there are 8 checkstyle violations because lines 45-48 contain tab characters

erwindon: I just saw the build report too. I'll soon fix that...

erwindon:
@brusdev all checks now passed...

PR #3565, opened 2021-05-04
ARTEMIS-3280 Netty Pool micrometer metric plugin

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3280

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
@@ -630,6 +630,9 @@ public final class ActiveMQDefaultConfiguration {
    // Number of concurrent workers for a core bridge
    public static int DEFAULT_BRIDGE_CONCURRENCY = 1;
 
+   // Whether or not to report Netty pool metrics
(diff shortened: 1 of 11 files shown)


franz1981:
This PR is including the many tiny details exposed by the Netty metrics, but common users will probably just need `netty.pooled.used.memory` of type `direct` and `heap`

franz1981:
@jbertram Do you think it worths to expose `netty.pooled.used.memory` with JMX as well? maybe in a separate PR?

jbertram (approved):

jbertram:
@franz1981, for the sake of flexibility it would be good to support these metrics via JMX as well. JMX is still heavily used, and many users have existing infrastructure for it.

jbertram:
Are there any Netty docs that we could link to from our docs that explain what these metrics are in more detail?

franz1981:
@jbertram 
> Are there any Netty docs that we could link to from our docs that explain what these metrics are in more detail?
yess, on https://netty.io/4.1/api/io/netty/buffer/PooledByteBufAllocatorMetric.html

michaelpearce-gain:
I would agree anything metrics wise, we should always seek to ensure its in JMX as then its also visible in admin console.

jbertram:
I went ahead and merged this. We can add corresponding JMX metrics in another PR. Nice work, @franz1981!

PR #3567, opened 2021-05-04
ARTEMIS-3283 Added SlowConsumerThreshold unit configuration option

mtaylor:


mtaylor:
retest

jbertram:
@mtaylor, I was about to merge this and noticed there were no documentation updates. Could you add something to `address-model.md` and `slow-consumers.md`? Thanks!

mtaylor:
Sure no problem.  Thanks Justin

mtaylor:
@jbertram Hey Justin.  I added a section to address-model.md.  The slow-consumers.md did not seem appropriate.  Could you take a look.

Thank you.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -3638,12 +3638,29 @@
...
             </xsd:element>
 
+            <xsd:element name="slow-consumer-threshold-measurement-unit" maxOccurs="1" minOccurs="0">
+               <xsd:annotation>
+                  <xsd:documentation>
+                     The units used to measure the slow consumer threshold.  Default is messages-per-second.
+                  </xsd:documentation>
+               </xsd:annotation>
+               <xsd:simpleType>
+                  <xsd:restriction base="xsd:string">
```
michaelpearce-gain, line 3655: @clebertsuconic / @mtaylor this looks very similar to the period unit that is in the journal history, i assume we're just using TimeUnit under the covers, should we maybe extract the type and simply share it....

michaelpearce-gain: e.g. could just have a shared type in the xsd that defines timeUnit and simple share that between Cleberts journal history time unit / period and yours.

    <xsd:simpleType name="timeUnit">
        <xsd:restriction base="xsd:string">
            <xsd:enumeration value="DAYS"/>
            <xsd:enumeration value="HOURS"/>
            <xsd:enumeration value="MINUTES"/>
            <xsd:enumeration value="SECONDS"/>
        </xsd:restriction>
    </xsd:simpleType>

mtaylor: I did consider using a time unit.  However, the measurement unit is really messages-per-second, messages-per-minute, I wanted to make it clear hence went with the more verbose option.  

I suppose it could be changed to mean "messages-per-<TimeUnit>" but imo that's not as intuitive.  It doesn't use TimeUnit under the covers here it instead uses a "MeasurementUnit" enum (as the calculations were very simple).

PR #3573, opened 2021-05-11
ARTEMIS-3292 PageSyncTimer can reduce garbage and sync while batching syncs

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3292

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PageSyncTimer.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PageSyncTimer.java
@@ -16,12 +16,13 @@
  */
 package org.apache.activemq.artemis.core.paging.impl;
 
-import java.util.LinkedList;
-import java.util.List;
+import java.util.ArrayDeque;
+import java.util.Queue;
 import java.util.concurrent.Executor;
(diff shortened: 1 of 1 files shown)


franz1981:
Would be better to use a batching strategy that won't cause tick to copy N pending syncs ie O(N) cost, but just a O(1) cost using a "double buffering" mechanism a-la video game computer graphic, used in order prj eg https://github.com/HdrHistogram/HdrHistogram/blob/master/src/main/java/org/HdrHistogram/SingleWriterRecorder.java#L134
is using a `WriterReaderPhaser` to switch active histograms in order to save the producer threads from being blocked while flushing.

franz1981:
TLDR on this change (still need to run CI on this): 

- lock-free sync requests (no interference with background sync):  it affects `PagingStoreImpl::page`, that's performing `sync` on
```java
            if (tx == null && syncNonTransactional && message.isDurable()) {
               sync();
            }
```
- performing background sync no longer allocate ` OperationContext[]`: it affects paging with fast disks ie with a very short page sync timeout

No hurry to get it but I think is a nice improvement @clebertsuconic @michaelandrepearce wdyt?

franz1981:
Still waiting approval so I can move on with CI test round (after rebase) and can brought in

PR #3576, opened 2021-05-12
NO-JIRA artemis.data system property 

cenitfish:
Within the artemis.xml the locations for storing the data can be configured (paging-directory, bindings-directory, journal-directory, large-messages-directory). Would it be possible to additionally support the system property artemis.data? The configuration would then be much easier.

### DIFF
--- a/artemis-server-osgi/src/main/java/org/apache/activemq/artemis/osgi/OsgiBroker.java
+++ b/artemis-server-osgi/src/main/java/org/apache/activemq/artemis/osgi/OsgiBroker.java
@@ -75,10 +75,17 @@ public class OsgiBroker {
          security.setRolePrincipalClass(rolePrincipalClass);
       }
       String brokerInstance = null;
-      String karafDataDir = System.getProperty("karaf.data");
-      if (karafDataDir != null) {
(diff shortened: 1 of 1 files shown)


clebertsuconic:
Sure we can do that. 

But artemis.instance was meant for that already.  What would be the use case?

clebertsuconic:
this is just renaming karat.data property as artemis.data right?

I just realized that the context of your change is OSGI only?

cenitfish:
Simply renaming karaf.data to artemis.data would also suffice for our purposes. In the meantime I tried the system property artemis.instance, but without success.

cenitfish:
And yes we work with Apache Karaf OSGi environment.

clebertsuconic:
ok.. fine with me...

Can you create a JIRA, and then:

```
git commit -a --amend
```

edit your NO-JIRA with the JIRA

```
git push -f
```

?

cenitfish:
I did and thank you!

michaelpearce-gain:
JIRA?

michaelpearce-gain:
Ok ignore my last comment i see commit is with Jira, maybe just update the PR name too ;) as this is what i picked up on quickly reviewing backlog of  PR's i havent put my eyes on.

PR #3577, opened 2021-05-12
ARTEMIS-3294 Fix conversion of messages with a text body

brusdev:


franz1981 (approved):

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/openmbean/OpenTypeSupport.java
```diff
@@ -293,7 +293,7 @@ protected void init() throws OpenDataException {
             if (m.containsProperty(Message.HDR_LARGE_COMPRESSED)) {
                rc.put(CompositeDataConstants.TEXT_BODY, "[compressed]");
             } else {
-               final String text = m.getReadOnlyBodyBuffer().readString();
+               final String text = m.getReadOnlyBodyBuffer().readNullableSimpleString().toString();
```
franz1981, line 296: This shouldn't be
```java
final SimpleString nullableText = m.getReadOnlyBodyBuffer().readNullableSimpleString();
fina String text = nullableText == null? null : nullableText.toString();
// ...
```
because of
```java
   public static SimpleString readNullableSimpleString(ByteBuf buffer) {
      int b = buffer.readByte();
      if (b == DataConstants.NULL) {
         return null;
      }
      return readSimpleString(buffer);
   }
```

brusdev: @franz1981 thanks, just fixed

gtully:
that looks correct, good catch. I recall trying to get rid of the simpleString,  learning :-)

gtully (approved):

PR #3578, opened 2021-05-13
ARTEMIS-3296: get things building on Java 16 (and 17EA+21)

gemmellr:
These initial changes get things building on Java 16 (and 17EA+21)

Updates the parent pom and various plugins or deps, tidies up some elements of the poms and consolidates to inherit plugin versions from the parent where possible.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 8, 11, 14 ]
+        java: [ 8, 11, 16 ]
 
     steps:
       - uses: actions/checkout@v2
@@ -31,7 +31,7 @@ jobs:
       # By setting anything to org.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED we are disabling libaio loading on the testsuite
       - name: Build Main
         run: |
-          mvn -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Djdk8-errorprone -Pfast-tests -Pextra-tests -Ptests-CI -Pjmh install
+          mvn -Dorg.apache.activemq.artemis.core.io.aio.AIOSequentialFileFactory.DISABLED=AnythingNotNull -Derrorprone -Pfast-tests -Pextra-tests -Ptests-CI -Pjmh install
 
       - name: Build Examples (JDK8 / -Prelease)
         if: matrix.java == '8'
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
       env:
         - EXAMPLES_PROFILE="noRun"
     - os: linux
-      jdk: openjdk14
+      jdk: openjdk16
       env:
         - EXAMPLES_PROFILE="noRun"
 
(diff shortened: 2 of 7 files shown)


gemmellr:
Although they build, the tests in the cdi and compatibility-tests modules are disabled when running on Java 16 as they dont work yet. More work can perhaps get them going later. 

For the compatibility-tests, upgrading groovy to a version got them into a more functional state, but they then got into a scenario where they appeared to be doing an infinite reconnect loop (mismatching client and server version failure was noted) and they repeatedly threw OOM errors also.

jbertram:
This should replace #3502, no?

gemmellr:
Yes, I originally meant to mention it and forgot to.

gemmellr:
Irony..part of the changes made in #3579 (which I just merged) to get some Kerberos tests working on JDK11+, does not work on 16 and an updated dep has also stopped other tests running on JDK16 that are in the PR build..

I shall begin investigating relevant module access config or test exclusion workarounds now :)

gemmellr:
Added another commit with changes to hopefully address that.

I dropped the Apache DS dep back 1 version (meaning it went from M15 originally up to 25 now, instead of 26) to remove the more general issue it added that was affecting the multiple existing tests. That issue has been fixed towards the next DS release but they have just removed [1] other functionality the test is using from it, so that release isnt going to be an option and the test will need removed/redone in future.

With that problem resolved, there was indeed a test-specific issue remaining with that second kerberos on JDK16 needing addressed, with its access of classes now guarded since JDK16, so I added an export for that to let it run and it now passed locally.

[1]https://github.com/apache/directory-server/commit/bb6db861681951c930941ed4bb92fd358fa333ec

gemmellr:
Looks like the GitHub Actions build itself is a bit broken (also doing same with a different build, https://github.com/apache/activemq-artemis/actions/runs/842401037). The subjobs were 'starting' (rather than just queued) since I pushed and the 16 job eventually failed seemingly without doing anything.

Travis passed for the main repo, and both Travis and GHA and Travis passed for my fork when the the commits were pushed:
https://github.com/gemmellr/activemq-artemis/actions/runs/842347976
https://travis-ci.org/github/gemmellr/activemq-artemis/builds/771137254?utm_source=github_status&utm_medium=notification

gemmellr:
FWIW, I did a _"mvn clean install -Derrorprone -Ptests -Ptests-CI -Pjmh"_ run on JDK8 with these changes the other day which passed

gemmellr:
Rebased to current main and squashed the 2 commits.

clebertsuconic:
@gemmellr thanks as usual Robbie

gemmellr:
Small fixup made in https://github.com/apache/activemq-artemis/commit/7733a76649403193e0e0e7e1a6279095fdbf426c around script permissions in the binary assembly, after issue was raised via #3585

brusdev:
@gemmellr the version 23 of the org.apache parent pom includes an execution of the maven-enforcer-plugin with an id different from the id defined in the artemis pom and this is causing duplicate runs

gemmellr:
Yep, as discussed elsewhere it would be good to rename the maven and java version enforcer checks to add '-version' to the execution id, will be clearer and consistent with the parent pom run, avoiding the effectively duplicate run.

gemmellr:
@brusdev already pushed it ;)

PR #3579, opened 2021-05-13
ARTEMIS-2813 Update MiniKDC and fix Kerberos tests on JDK 11+

tabish121:
Updates the MiniKDC dependency and adds another now needed dep in order
to get the Kerberos tests working on JDK 11+ builds.


gemmellr:
Looks good, ran the tests on 8 and 11 without issue. Merged.

PR #3580, opened 2021-05-13
ARTEMIS-3297 Historic Journal Backup

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/config/Configuration.java
```diff
@@ -668,6 +668,25 @@ Configuration addDiscoveryGroupConfiguration(String key,
...
+   File getJournalHistoryLocation();
+
+   /**
+    * Sets the file system directory used to store historical backup journal.
+    */
+   Configuration setJournalHistory(String dir);
+
+   int getJournalHistoryMaxDays();
+
+   Configuration  setJournalHistoryMaxDays(int days);
```
gtully, line 682: This is a neat feature. I think I would call it something else from a configuration perspective.

journalRetention or just retention

configured with a:
 
int retentionLimit > 0
enum retentionUnit (seconds/days/numFiles)

artemis-server/src/test/resources/ConfigurationTest-full-config.xml
```diff
@@ -406,6 +406,9 @@
       <page-max-concurrent-io>17</page-max-concurrent-io>
       <read-whole-page>true</read-whole-page>
       <journal-directory>somedir2</journal-directory>
+      <journal-history-directory>history</journal-history-directory>
```
gtully, line 409: this config looks neat, but I think "journal-retention" is a better name, and I think in high volume scenarios second granularity will be necessary to capture use cases less than days.

gtully (commented):
this looks great. But I vote for naming this feature "retention" and the cli can replay historical data from what has been retained. Ie: the cli can be extended with wildcard matches etc but the broker will always just be "retaining" pages/journal files.

artemis-server/src/test/resources/ConfigurationTest-full-config.xml
clebertsuconic: @gtully I thought about using max-files for such cases.. Humans are usually slow to react... ..

perhaps we could have this

<journal-retention period-type="days|files|hours" period="Number" folder="data/history"/>

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/Configuration.java
clebertsuconic: Oooohhhh... I like the name retention!!! Really nice!!!!

artemis-server/src/test/resources/ConfigurationTest-full-config.xml
gtully: files is good if you care about you max disk usage. But I think for time, hours is too coarse. and seconds covers everything up to hours nicely.

gtully: period does not work for files: maybe:

<journal-retention limit-type="files|days|hours|seconds" limit="number > 0" folder="data/history"/>

gtully: <journal-retention  limit-type="files|days|hours|seconds" limit="number > 0" folder="data/history"/>

michaelandrepearce: What if you want to limit by files and time at same time.

michaelandrepearce: I would suggest a timebased config item and a files/size based config item

gtully: yep, I agree, time and space, probably need both, to limit disk usage in case the time limits is too large.

michaelandrepearce: ``

michaelandrepearce: > files is good if you care about you max disk usage. But I think for time, hours is too coarse. and seconds covers everything up to hours nicely.

Initial idea was for it to be coarse as it wood be days/weeks, intent is to time machine back after say an incident that gets found out after the fact and work out what could have happened. I’m not sure how useful seconds here would be tbh.

clebertsuconic: @michaelandrepearce some time ago, I talked to an Architect at my job who wanted to get diffs from the broker and send over to another site from time to time. 

This kind of feature would be useful for such case maybe?


It wouldn't hurt to have it though.. it's just an additional option.


I am thinking to have a commented out xml on the broker.xml that will have 7 days, 10G as the config. So if users just uncomment that they will have the 7 days and 10G max disk.

clebertsuconic: Hmmm.. this feature could also be used for Disaster and Recovery options.

clebertsuconic: @michaelandrepearce / @gtully actually.. I take that back...


I will do days and hours...


I don't have time to test all the options now.. having more code means I would have to test it...


If we ever cross a use case of a user needing more granularity we can add it later.

clebertsuconic: I was going to take it back, but I actually wrote a test to validate the options on the XML.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -675,6 +675,82 @@
...
+                  <xsd:annotation>
+                     <xsd:documentation>
+                        The location on your file system where the journal-retention will take place.
+                     </xsd:documentation>
+                  </xsd:annotation>
+               </xsd:attribute>
+               <xsd:attribute name="period" type="xsd:integer" use="required">
+                  <xsd:annotation>
+                     <xsd:documentation>
+                        The amount of time used to keep files.
```
michaelandrepearce, line 709: What if someone wants to limit just by size?

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -675,6 +675,82 @@
...
+            </xsd:annotation>
+            <xsd:complexType>
+               <xsd:attribute name="unit" use="required">
+                  <xsd:annotation>
+                     <xsd:documentation>
+                        This configures the period type to use on limit. By default it is DAYS.
+                     </xsd:documentation>
+                  </xsd:annotation>
+                  <xsd:simpleType>
+                     <xsd:restriction base="xsd:string">
```
michaelandrepearce, line 692: Same comment as below what if someone wants to limit just on size

artemis-server/src/main/resources/schema/artemis-configuration.xsd
michaelandrepearce: maybuscan use xsd assert to handle more complex checks that at least size or time is set and if time that period is set

artemis-server/src/main/resources/schema/artemis-configuration.xsd
michaelandrepearce: Also why seperate the unit and the period why not as a single attribute string like for storage where its value and scaler

E.g. 2040594944ms    360s  4h 7d   and if no unit default to ms

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -675,6 +675,82 @@
...
+                        <xsd:enumeration value="DAYS"/>
+                        <xsd:enumeration value="HOURS"/>
+                        <xsd:enumeration value="SECONDS"/>
+                     </xsd:restriction>
+                  </xsd:simpleType>
+               </xsd:attribute>
+               <xsd:attribute name="directory" type="xsd:string" use="required">
```
michaelandrepearce, line 699: Why is this an attribute but for all other data directories its the element value? Just a consistency thing.

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -768,6 +772,45 @@ public void parseMainConfig(final Element e, final Configuration config) throws
...
+               break;
+            case "SECONDS":
+               unit = TimeUnit.SECONDS;
+               break;
+            default:
+               unit = TimeUnit.DAYS;
+         }
+
+         config.setJournalRetentionDirectory(directory);
+         config.setJournalRetentionTimeUnit(unit);
```
michaelandrepearce, line 807: So that two args throughout code base dont have to deal with time conversion and also in more critical code it doesnt have to be converted later why not convert to ms at point of config load

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/dataformat/JournalAddRecord.java
```diff
@@ -30,32 +30,37 @@
 
    protected final byte recordType;
 
-   protected final boolean add;
+   protected final byte journalType;
```
michaelandrepearce, line 33: Is this back compatible?

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -2245,6 +2379,155 @@ private synchronized JournalLoadInformation load(final LoaderCallback loadManage
...
+            }
+         };
+
+
+         if (journalRetentionPeriod > 0) {
+            String[] fileNames = journalRetentionFolder.list(fnf);
+            Arrays.sort(fileNames);
+
+            GregorianCalendar calendar = this.calendarThreadLocal.get();
+            calendar.setTimeInMillis(System.currentTimeMillis() - journalRetentionUnit.toMillis(journalRetentionPeriod));
```
michaelandrepearce, line 2400: Why doing the conversion to ms every time, why not convert to ms just once all the way up at point of reading config

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -2245,6 +2379,155 @@ private synchronized JournalLoadInformation load(final LoaderCallback loadManage
...
+      if (journalRetentionFolder != null) {
+         ArrayList<JournalFile> filesToMove;
+         synchronized (historyPendingFiles) {
+            filesToMove = new ArrayList<>(historyPendingFiles.size());
+            filesToMove.addAll(historyPendingFiles);
+            historyPendingFiles.clear();
+         }
+
+
+         Calendar calendar = calendarThreadLocal.get();
```
michaelandrepearce, line 2449: Why in some places using gregorian and her just plain calendar

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -1858,6 +1988,8 @@ public synchronized void compact() throws Exception {
          journalLock.writeLock().unlock();
       }
 
+      processBackup();
```
michaelandrepearce, line 1991: I thought during our discussions we want the backup before compaction so its literally as raw as it gets.

michaelandrepearce: Or am i miss understanding whats going on here?

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -108,6 +115,7 @@
     *
     * */
    public static final double UPDATE_FACTOR;
+   public static final String BKP = ".bkp";
```
michaelandrepearce, line 118: If its going to another directory do we really need or even want a suffix on it?

michaelandrepearce:
@clebertsuconic looks good left some comments / queries. On top of that just for my own benefit as i couldnt figure it out, where is it that data that would page is now still written to journal so that we get a pure journal? I see large message was dealt with but couldn't spot records that paged

artemis-server/src/main/resources/schema/artemis-configuration.xsd
clebertsuconic: I think we should require a period configured. make it large if you wish (365 days e.g)

artemis-server/src/main/resources/schema/artemis-configuration.xsd
clebertsuconic: it wouldn't be as clear. the XSD is pretty standard.. if you edit the XML with an editor you get auto-completion and self documented. As if you use the schema you use on bytes it would be free form... more error prone.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
clebertsuconic: the element is journal-retention.. you simply enable it or disable it...

I think it is very neat the way it's done now.


you just comment this from the xml, and you have it working:

```xml
<journal-retention unit="DAYS" directory="data/retention" period="7" storage-limit="10G"/>
```

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/dataformat/JournalAddRecord.java
clebertsuconic: It is... Compatibility tests would be failing if it wasn't.


addBytes will write 1 for true, 0 for false.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
clebertsuconic: ok.. fair enough.. I can do that

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
clebertsuconic: It doesn't make a difference really? the threadLocal is still a GregorianCalendar.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
clebertsuconic: processBackup() will copy the BKP files to the retention folder.

Before I do the computation I need any BKP files cleared so they can be treated. So, the backup is before compacting, hence I'm calling processBackup before compacting here.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
clebertsuconic: This is for the current journal file.

I rename it to BKP, in case the node crashes before the copy is finished, when restarted the copy is finished and the journal name restored. This is for the pending files to be copied from the journal to the BKP area.

clebertsuconic:
@michaelandrepearce that's happening on the JournalstorageManager::pageWrite

https://github.com/clebertsuconic/activemq-artemis/blob/backup-journal/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java#L410-L427


I'm using the storeAddEvent, using the same record type as the Add Message and addRef.

Basically if you recover a message from the "paging area" it will come back to the journal. 



(at least for now).

We could use the same to write to Paging.. or simply do the route to whatever is the best at the time of recovering. But at this version this is always coming back to the journal no matter where it came from.

clebertsuconic:
@michaelandrepearce RecoverTest as I added is dealing with Transaction, non TX, Paging, non Paging, large message, non large message, NIO, MAPPED, AIO ... all with combinations.. trying to capture all possible scenarios.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
michaelandrepearce: Yes but other dirs the element value is the directory name... not saying to move attributes for the retention parts. Just saying shouldnt the directory attribute be the element value as like other data dirs

michaelandrepearce: Eg

`<journal-retention unit="DAYS" period="7" storage-limit="10G">data/retention</journal-retention>`

clebertsuconic: you mean this?

```
<journal-retention unit="DAYS" period="7" storage-limit="10G">data/retention</journal-retention>
```

I would have to call it journal-retention-directory.. it wouldn't look as nice IMO. I would probably have to split it into separate elements... I would rather keep the way it is now.

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
clebertsuconic: I will think about this between tomorrow (Saturday) and Monday.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
michaelandrepearce: because of the change i was thinking there was maybe more to it.  I guess just consistency so in future people not double thinking like i just did why the difference

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
michaelandrepearce: Understood

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
michaelandrepearce: Thanks

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/dataformat/JournalAddRecord.java
michaelandrepearce: Cool just checking

artemis-server/src/main/resources/schema/artemis-configuration.xsd
michaelandrepearce: Im one for consistency. Especially as a user mindset. Like why all other dirs the element value yet here its an empty element value but its an attribute.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
clebertsuconic: I will allow unlimited.. and log a warn if no value is provided.

clebertsuconic: I will actually require one filled.. either period or storage and throw an error if not passed.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
clebertsuconic: It is not possible to have both attributes and default children on a xml type. I will go with what I had before... and I need to move on on this.

michaelandrepearce: I dont follow. Im sure if its enabled by declaring the xml were ensuring that directory is NOT optional.

clebertsuconic: the other data dirs don't have attributes. I don't want to get stuck over this TBH.

clebertsuconic: I can't make it this on the xsd:


```xml
<journal-retention-directory period="x">directory</journal-retention-directory
```


There are other complex types on the XSD for the configuration.. I don't see an issue with this TBH... And I really need to finish this task as I need to finish the release.

michaelandrepearce: Its about consistency. And the issue is once released we cant easily change it.

We actually already have such type on xsd/xml. connectors and acceptors which have a value as well as attributes.

michaelandrepearce: I believe sample type in the xsd is transportType

michaelpearce-gain:
```
            <xsd:element name="journal-retention-directory" maxOccurs="1" minOccurs="0">
                <xsd:annotation>
                    <xsd:documentation>
                        the directory to store journal-retention message in and rention configuraion.
                    </xsd:documentation>
                </xsd:annotation>
                <xsd:complexType>
                    <xsd:simpleContent>
                        <xsd:extension base="xsd:string">
                            <xsd:attribute name="unit" use="optional">
                                <xsd:annotation>
                                    <xsd:documentation>
                                        This configures the period type to use on limit. By default it is DAYS.
                                    </xsd:documentation>
                                </xsd:annotation>
                                <xsd:simpleType>
                                    <xsd:restriction base="xsd:string">
                                        <xsd:enumeration value="DAYS"/>
                                        <xsd:enumeration value="HOURS"/>
                                        <xsd:enumeration value="SECONDS"/>
                                    </xsd:restriction>
                                </xsd:simpleType>
                            </xsd:attribute>
                            <xsd:attribute name="period" type="xsd:integer" use="optional">
                                <xsd:annotation>
                                    <xsd:documentation>
                                        The amount of time used to keep files.
                                    </xsd:documentation>
                                </xsd:annotation>
                            </xsd:attribute>
                            <xsd:attribute name="storage-limit" type="xsd:string" use="optional">
                                <xsd:annotation>
                                    <xsd:documentation>
                                        Size (in bytes) before we starting removing files from the retention area.
                                        this is an extra protection on top of the period.
                                        Notice we first remove files based on period and if you're using more storage then you
                                        configured we start removing older files.
                                        By default this is unlimited (not filled).
                                        Supports byte notation like "K", "Mb", "GB", etc.
                                    </xsd:documentation>
                                </xsd:annotation>
                            </xsd:attribute>
                        </xsd:extension>
                    </xsd:simpleContent>
                </xsd:complexType>
            </xsd:element>```

Comment (michaelpearce-gain) [reply artemis-server/src/main/resources/schema/artemis-configuration.xsd:699]: dont forget to add    <xsd:attributeGroup ref="xml:specialAttrs"/>

File: artemis-server/src/main/resources/schema/artemis-configuration.xsd [hunk 12]
```diff
@@ -675,6 +675,82 @@
...
+
+         <xsd:element name="journal-history-directory" type="xsd:string" default="" maxOccurs="1" minOccurs="0">
+            <xsd:annotation>
+               <xsd:documentation>
+                  The directory to store historical information about the data on the broker.
+               </xsd:documentation>
+            </xsd:annotation>
+         </xsd:element>
+
+         <xsd:element name="journal-history-max-days" type="xsd:integer" default="-1" maxOccurs="1" minOccurs="0">
```
michaelpearce-gain, line 737: is this dead now?

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -675,6 +675,82 @@
...
+            <xsd:annotation>
+               <xsd:documentation>
+                  The maximum number of days to keep on the history journal folder backup.
+               </xsd:documentation>
+            </xsd:annotation>
+         </xsd:element>
+
+         <xsd:element name="journal-history-max-files" type="xsd:integer" default="-1" maxOccurs="1" minOccurs="0">
```
michaelpearce-gain, line 745: is this dead now?

clebertsuconic: yes... max-files will be removed.

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -768,6 +772,63 @@ public void parseMainConfig(final Element e, final Configuration config) throws
...
+         TimeUnit unit;
+
+         switch (unitStr) {
+            case "DAYS":
+               unit = TimeUnit.DAYS;
+               break;
+            case "HOURS":
+               unit = TimeUnit.HOURS;
+               break;
+            case "MINUTES":
```
michaelpearce-gain, line 807: Here we support minutes but schema only has hours, days, seconds, i guess we want to add minutes to the schema / xsd

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -675,6 +675,82 @@
...
+                  <xsd:annotation>
+                     <xsd:documentation>
+                        This configures the period type to use on limit. By default it is DAYS.
+                     </xsd:documentation>
+                  </xsd:annotation>
+                  <xsd:simpleType>
+                     <xsd:restriction base="xsd:string">
+                        <xsd:enumeration value="DAYS"/>
+                        <xsd:enumeration value="HOURS"/>
+                        <xsd:enumeration value="SECONDS"/>
```
michaelpearce-gain, line 695: In artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java seems we support also Minutes.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
michaelpearce-gain: <xsd:element name="journal-retention-directory" maxOccurs="1" minOccurs="0">
                <xsd:annotation>
                    <xsd:documentation>
                        the directory to store e journal-retention will take place.
                    </xsd:documentation>
                </xsd:annotation>
                <xsd:complexType>
                    <xsd:simpleContent>
                        <xsd:extension base="xsd:string">
                            <xsd:attribute name="unit" use="optional" type="timeUnit">
                                <xsd:annotation>
                                    <xsd:documentation>
                                        This configures the period type to use on limit. By default it is DAYS.
                                    </xsd:documentation>
                                </xsd:annotation>
                            </xsd:attribute>
                            <xsd:attribute name="period" type="xsd:integer" use="optional">
                                <xsd:annotation>
                                    <xsd:documentation>
                                        The amount of time used to keep files.
                                    </xsd:documentation>
                                </xsd:annotation>
                            </xsd:attribute>
                            <xsd:attribute name="storage-limit" type="xsd:string" use="optional">
                                <xsd:annotation>
                                    <xsd:documentation>
                                        Size (in bytes) before we starting removing files from the retention area.
                                        this is an extra protection on top of the period.
                                        Notice we first remove files based on period and if you're using more storage then you
                                        configured we start removing older files.
                                        By default this is unlimited (not filled).
                                        Supports byte notation like "K", "Mb", "GB", etc.
                                    </xsd:documentation>
                                </xsd:annotation>
                            </xsd:attribute>
                        </xsd:extension>
                    </xsd:simpleContent>
                </xsd:complexType>
            </xsd:element>


And also define timeUnit so its re-usable going forwards, after all its java TimeUnit that we use to convert.

    <xsd:simpleType name="timeUnit">
        <xsd:restriction base="xsd:string">
            <xsd:enumeration value="DAYS"/>
            <xsd:enumeration value="HOURS"/>
            <xsd:enumeration value="MINUTES"/>
            <xsd:enumeration value="SECONDS"/>
        </xsd:restriction>
    </xsd:simpleType>

clebertsuconic: ok.. thanks.. that works

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -675,6 +675,52 @@
...
+                     <xsd:attribute name="unit" use="optional">
+                        <xsd:annotation>
+                           <xsd:documentation>
+                              This configures the period type to use on limit. By default it is DAYS.
+                           </xsd:documentation>
+                        </xsd:annotation>
+                        <xsd:simpleType>
+                           <xsd:restriction base="xsd:string">
+                              <xsd:enumeration value="DAYS"/>
+                              <xsd:enumeration value="HOURS"/>
```
michaelpearce-gain, line 696: we probably want to add MINUTES as the artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java added, seems to handle it.

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -768,6 +772,64 @@ public void parseMainConfig(final Element e, final Configuration config) throws
...
+         int period = getAttributeInteger(node, "period", -1, Validators.GT_ZERO);
+         String unitStr = getAttributeValue(node, "unit");
+
+         if (unitStr == null) {
+            unitStr = "DAYS";
+         }
+
+         TimeUnit unit;
+
+         switch (unitStr) {
```
michaelpearce-gain, line 801: Why not use TimeUnit.valueOf its already an enum?

michaelpearce-gain: e.g. this switch could be simplified down to:

unit = TimeUnit.valueOf(unitStr.toUpperCase());

clebertsuconic: Nice.. idea.. done!

michaelpearce-gain:
One more query, im trying to work out here, is the copy done whilst and journal is blocked for write, or is it being done after like with compaction and release not blocking the journal writes on the current file, but simply before we do do the release or compaction.

clebertsuconic:
@michaelpearce-gain it should be done after...

I'm renaming the file since the lock... but the actual copy should be done outside of any locking.


there might be some tweaking to be done into it though. but that was my intention at least.

clebertsuconic:
This should be ready to be merged now.. I'm running one final round of tests.. and if everything goes well I will merge it tomorrow... further changes can be followed as with any module, but I think this is in good shape now.

PR #3581, opened 2021-05-14
ARTEMIS-3295 - do not cluster advisories

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-3295


clebertsuconic:
I am sending a small tweak into this.

Instead of passing the List as a reference everywhere, it would be simpler to just read from ClusterManager.

The list is passed as a reference and if you change the source you also change on the passed reference. I think it could eventually be confusing. Instead I am just adding a getter on ClusterManager and reading from the source every time is needed.

clebertsuconic:
superseded by #3582 

I did a small tweak to the code instead of asking you to do it.. 
@andytaylor you're still the author of the commit.

PR #3582, opened 2021-05-14
ARTEMIS-3295 - do not cluster advisories

clebertsuconic:
https://issues.apache.org/jira/browse/ARTEMIS-3295


clebertsuconic:
This is a tweak on top of #3581

PR #3583, opened 2021-05-17
ARTEMIS-3285 potential duplicate messages with LVQ + non-destructive

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LastValueQueue.java
```diff
@@ -259,6 +251,8 @@ public QueueConfiguration getQueueConfiguration() {
    private void replaceLVQMessage(MessageReference ref, HolderReference hr) {
       MessageReference oldRef = hr.getReference();
 
+//      logger.info("Replacing " + oldRef + " with " + ref);
```
michaelandrepearce, line 254: Commented code

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -133,7 +133,7 @@
    protected static final int CRITICAL_CONSUMER = 3;
    protected static final int CRITICAL_CHECK_DEPAGE = 4;
 
-   private static final Logger logger = Logger.getLogger(QueueImpl.class);
+   protected static final Logger logger = Logger.getLogger(QueueImpl.class);
```
michaelandrepearce, line 136: Should keep private and if a subclass wants to log then it makes its own logger , so its logged against the right class

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LastValueQueue.java
jbertram: fixed

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
jbertram: fixed

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSNonDestructiveTest.java
```diff
@@ -588,4 +599,87 @@ private void sendLVQTombstone(ConnectionSupplier producerConnectionSupplier, Str
...
+      if (dups.size() > 0) {
+         StringBuffer sb = new StringBuffer();
+         for (Map.Entry<String, Integer> stringIntegerEntry : dups.entrySet()) {
+            sb.append(stringIntegerEntry.getKey() + "(" + stringIntegerEntry.getValue() + "),");
+         }
+         Assert.fail("Duplicate messages received " + sb);
+      }
+
+      if (persistenceEnabled) {
+         Wait.assertTrue(() -> {
```
clebertsuconic, line 679: why the check on persistenceEnabled? Is the test running more than once?

tests/config/logging.properties
```diff
@@ -68,4 +68,5 @@ handler.TEST.formatter=PATTERN
 # Formatter pattern configuration
 formatter.PATTERN=org.jboss.logmanager.formatters.PatternFormatter
 formatter.PATTERN.properties=pattern
-formatter.PATTERN.pattern=[%t] %d{HH:mm:ss,SSS} %-5p [%c] %s%E%n
+#formatter.PATTERN.pattern=[%t] %d{HH:mm:ss,SSS} %-5p [%c] %s%E%n
+formatter.PATTERN.pattern=%d{HH:mm:ss,SSS} %-5p %s%E%n
```
clebertsuconic, line 72: @jbertram was this change intentional?

jbertram: No. I just fixed it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSNonDestructiveTest.java
jbertram: The test class is using `Parameterized` and if persistence is disabled then `countJournal` throws a bunch of exceptions.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -918,7 +917,7 @@ public Queue getQueue() {
             }
 
             if (ref == null) {
-               ActiveMQIllegalStateException ils = ActiveMQMessageBundle.BUNDLE.consumerNoReference(id, messageID, messageQueue.getName());
+               ActiveMQIllegalStateException ils = new ActiveMQIllegalStateException("Cannot find ref to ack " + messageID);
```
clebertsuconic: Was this on purpose? The former seemed more complete to me.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -928,7 +927,7 @@ public Queue getQueue() {
 
             acks++;
          }
-         while (ref.getMessageID() != messageID);
+         while (ref != null && ref.getMessageID() != messageID);
```
clebertsuconic: was this on purpose? it seems a left over from a debug, since it was never supposed to hit null here (the throw exception is done when ref==null)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LastValueQueue.java
```diff
@@ -151,25 +171,20 @@ public synchronized void addTail(final MessageReference ref, final boolean direc
...
+            if (isNonDestructive() && hr.isInDelivery()) {
+               // if the ref is already being delivered we'll do the replace in the postAcknowledge
+               hr.setReplacementRef(ref);
+            } else {
+               // We need to overwrite the old ref with the new one and ack the old one
+               replaceLVQMessage(ref, hr);
 
+               if (isNonDestructive() && hr.isDelivered()) {
+                  hr.resetDelivered();
+                  // since we're replacing a ref that was already delivered we want to trigger a "special" delivery for this new replacement
```
clebertsuconic, line 183: I - stale comment... the "special" word...

ii - just an idea... what about rename repeatNextDelivery(messageReference) as nextDelivery(messageReference)

and call it here instead of using the collection directly? 

just a suggestion!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LastValueQueue.java
```diff
@@ -56,9 +58,27 @@
 @SuppressWarnings("ALL")
 public class LastValueQueue extends QueueImpl {
 
+   private static final Logger logger = Logger.getLogger(LastValueQueue.class);
    private final Map<SimpleString, HolderReference> map = new ConcurrentHashMap<>();
    private final SimpleString lastValueKey;
 
+   // only use this within synchronized methods or synchronized(this) blocks
+   protected volatile LinkedList<MessageReference> nextDeliveries = new LinkedList<>();
```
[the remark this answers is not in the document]
clebertsuconic: I almost forgot.. replace volatile by final here please?

it used to be an element on your logic. .but as it's a collection now it needs to be a final linkedList (at least on the way it is implemented now.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
jbertram: This was not on purpose.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
jbertram: This was also not on purpose.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LastValueQueue.java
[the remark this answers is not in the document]
jbertram: Should be `final` now.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/LastValueQueue.java
jbertram: I fixed "special" in the comment.

I didn't change `repeatNextDelivery`. I used "repeat" in the name because it serves the same purpose as `repeat` on the `LinkedListIterator` in `QueueImpl`. I wanted that to be clear so its easy to see that if the consumer is "busy" we just repeat the delivery no matter what.

clebertsuconic:
nice job on this one... merging it.

PR #3584, opened 2021-05-17
ARTEMIS-3303 Default thread pool size is too generous

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3303

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
@@ -146,7 +146,7 @@ public final class ActiveMQDefaultConfiguration {
    private static int DEFAULT_SCHEDULED_THREAD_POOL_MAX_SIZE = 5;
 
    // Maximum number of threads to use for the thread pool. -1 means 'no limits'.
-   private static int DEFAULT_THREAD_POOL_MAX_SIZE = 30;
+   private static int DEFAULT_THREAD_POOL_MAX_SIZE = Math.max(2, Runtime.getRuntime().availableProcessors() / 2);
 
    // true means that security is enabled
    private static boolean DEFAULT_SECURITY_ENABLED = true;
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyAcceptor.java
@@ -404,8 +404,8 @@ public class NettyAcceptor extends AbstractAcceptor {
       } else {
 
          if (remotingThreads == -1) {
-            // Default to number of cores * 3
-            remotingThreads = Runtime.getRuntime().availableProcessors() * 3;
+            // Default to half of cores
+            remotingThreads = Math.max(1, Runtime.getRuntime().availableProcessors() / 2);
          }
 
          if (useEpoll && CheckDependencies.isEpollAvailable()) {
--- a/artemis-server/src/main/resources/schema/artemis-configuration.xsd
+++ b/artemis-server/src/main/resources/schema/artemis-configuration.xsd
@@ -99,7 +99,7 @@
             </xsd:annotation>
          </xsd:element>
 
-         <xsd:element name="thread-pool-max-size" type="xsd:int" default="30" maxOccurs="1" minOccurs="0">
+         <xsd:element name="thread-pool-max-size" type="xsd:int" default="-1" maxOccurs="1" minOccurs="0">
             <xsd:annotation>
                <xsd:documentation>
                   Maximum number of threads to use for the thread pool. -1 means 'no limits'.
--- a/docs/user-manual/en/configuration-index.md
+++ b/docs/user-manual/en/configuration-index.md
@@ -180,7 +180,7 @@ system-property-prefix | Prefix for replacing configuration settings using Bean 
 internal-naming-prefix | the prefix used when naming the internal queues and addresses required for implementing certain behaviours. | `$.activemq.internal`
 [populate-validated-user](security.md#tracking-the-validated-user)| whether or not to add the name of the validated user to the messages that user sends. | `false`
 [security-settings](security.md#role-based-security-for-addresses) | [a list of security-setting](#security-setting-type). | n/a
-[thread-pool-max-size](thread-pooling.md#thread-management) | Maximum number of threads to use for the thread pool. -1 means 'no limits'. | 30
+[thread-pool-max-size](thread-pooling.md#thread-management) | Maximum number of threads to use for the thread pool. -1 means 'no limits'. | `available cores/2`
 [transaction-timeout](transaction-config.md) | how long (in ms) before a transaction can be removed from the resource manager after create time. | 300000
 [transaction-timeout-scan-period](transaction-config.md) | how often (in ms) to scan for timeout transactions. | 1000
 [wild-card-routing-enabled](wildcard-routing.md) | true means that the server supports wild card routing. | `true`
--- a/docs/user-manual/en/configuring-transports.md
+++ b/docs/user-manual/en/configuring-transports.md
@@ -224,12 +224,12 @@ simple TCP:
   if you are using this please update your configuration
 
 - `remotingThreads`. Apache ActiveMQ Artemis will, by default, use a number of
-  threads equal to three times the number of cores (or hyper-threads) as
+  threads equal to half the number of cores (or hyper-threads) as
   reported by `Runtime.getRuntime().availableProcessors()` for processing
   incoming packets. If you want to override this value, you can set the number of
   threads by specifying this parameter. The default value for this parameter is
-  `-1` which means use the value from
-  `Runtime.getRuntime().availableProcessors()` \* 3.
+  `-1` which means use half the value from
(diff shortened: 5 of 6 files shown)


franz1981:
Need some discussion about:

- min pool size
- default pool size

What happen when different acceptors are being used?
This behaviour is still a bit static.

michaelpearce-gain:
Im always cautious about such changes, after all one default for one org, is different to anothers use case in another org.

franz1981:
All fair points, and indeed I believe this should be a cautious and more conservative change but still, there are some historical motivations and experimental facts that can prove that what we set by default is no longer valid/usefull and that it was optimizing for context switches here, nor for throughout or latencies trafe-offs: there are some historical motivations and experimental facts that can prove this.

1. Re the Netty event loop sizing

- historical facts: HornetQ and earlier versions of Artemis was blocking Netty threads, but that's no longer true. We can even choose to use `Blockhound` to enforce/check it on our CI, see https://github.com/netty/netty/pull/9687
- experimental facts: generating a uniformly distributed load with clients >= cores using Core clients shown that the default configuration of Netty thread pool (3X number of cores) prevent scaling and is both hitting troughput and latencies. See https://github.com/apache/activemq-artemis/pull/3572#issuecomment-841788187 for some more details about it.

The motivation re the experimental facts seems related how Netty event loop group works: 
- Netty assign client connections in round-robin fashion to the configured Netty threads
- each client connection can issue write/read events on the event loop (single) selector to wakeup for any work to do
- if the number of Netty threads exceed the number of cores and the number of clients is <= Netty threads, each time such notification happen they have some chance (2/3 possibilities) the thread that's going to handle it won't be on cpu (because they exceed the amount of cores) and the OS is forced to deschedule some (random) thread in order to run the Netty thread responsible to handle the interrupt, causing un-necessary context-switches.

The netty default is of 2X the amount of cores for applications that heavily relies just on event loop processing, but Artemis it's not: even AMQP use I/O threads and need GC, compiler threads and sometime global threads to perform its job. Just using 3X is a waste of resources for the current Artemis version.

2. Re the global thread pool sizing

That's a bit more complex and depends by how `ActiveMQThreadPoolExecutor` works.
Just writing a simple program can help to spot what's the problem with it (very similar to the Netty one, but not the same).
```java
   public static void main(String[] args) throws InterruptedException {
      ThreadPoolExecutor executor = new ActiveMQThreadPoolExecutor(0, 30, 60L, TimeUnit.SECONDS, new ThreadFactory() {
         @Override
         public Thread newThread(Runnable r) {
            Thread t = new Thread(r);
            System.err.println("created new thread: " + t);
            return t;
         }
      });
      ExecutorFactory factory = new OrderedExecutorFactory(executor);
      final int clients = 30;
      int bursts = 100;
      ConcurrentHashSet[] executingThreads = new ConcurrentHashSet[clients];
      ArtemisExecutor[] artemisExecutor = new ArtemisExecutor[clients];
      for (int i = 0; i< clients; i++) {
         artemisExecutor[i] = factory.getExecutor();
         executingThreads[i] = new ConcurrentHashSet();
      }
      ConcurrentMap<Thread, AtomicLong> executingT = new ConcurrentHashMap<>();
      for (int j = 0; j< bursts;j++) {
         for (int i = 0; i < clients; i++) {
            ConcurrentHashSet threadsSeen =executingThreads[i];
            artemisExecutor[i].execute(() -> {
               try {
                  TimeUnit.MILLISECONDS.sleep(1);
               } catch (InterruptedException e) {
                  e.printStackTrace();
               }
               threadsSeen.add(Thread.currentThread());
               AtomicLong counter = executingT.get(Thread.currentThread());
               if (counter == null) {
                  executingT.put(Thread.currentThread(), new AtomicLong(1));
               } else {
                  counter.lazySet(counter.get() + 1);
               }
            });
         }
         System.out.println("GC pause");
         Thread.sleep(100);
      }
      for (int i = 0; i< clients; i++) {
         artemisExecutor[i].flush(60, TimeUnit.SECONDS);
      }
      executor.shutdown();
      executor.awaitTermination(70, TimeUnit.SECONDS);
      System.out.println("Executing threads: " + executingT);
      System.out.println("Workload distribution per artemis executor:");
      for (int i = 0; i < clients; i++) {
         System.out.println("[" + (i + 1) + "] - " + executingThreads[i].size());
      }
   }
```
On my machine (12 cores with HT - 6 real cores) it prints 30 times 
```created new thread: ...```
and 
```
Executing threads: 
{Thread[Thread-1,5,]=103, 
Thread[Thread-20,5,]=99, 
Thread[Thread-17,5,]=99, 
Thread[Thread-11,5,]=101, 
Thread[Thread-18,5,]=99, 
Thread[Thread-14,5,]=100, 
Thread[Thread-13,5,]=100, 
Thread[Thread-21,5,]=99, 
Thread[Thread-24,5,]=98, 
Thread[Thread-28,5,]=98, 
Thread[Thread-5,5,]=103, 
Thread[Thread-30,5,]=97, 
Thread[Thread-27,5,]=97, 
Thread[Thread-6,5,]=103, 
Thread[Thread-4,5,]=102, 
Thread[Thread-23,5,]=98, 
Thread[Thread-25,5,]=98, 
Thread[Thread-8,5,]=102, 
Thread[Thread-7,5,]=102,
Thread[Thread-3,5,]=103,
Thread[Thread-9,5,]=101, 
Thread[Thread-10,5,]=102, 
Thread[Thread-19,5,]=99, 
Thread[Thread-12,5,]=101, 
Thread[Thread-15,5,]=100, 
Thread[Thread-26,5,]=97, 
Thread[Thread-29,5,]=97, 
Thread[Thread-2,5,]=103,
Thread[Thread-16,5,]=100, 
Thread[Thread-22,5,]=99}
Workload distribution per artemis executor:
[1] - 17
[2] - 18
[3] - 17
[4] - 15
[5] - 13
[6] - 13
[7] - 17
[8] - 17
[9] - 18
[10] - 14
[11] - 13
[12] - 17
[13] - 15
[14] - 14
[15] - 17
[16] - 18
[17] - 16
[18] - 12
[19] - 17
[20] - 16
[21] - 16
[22] - 14
[23] - 17
[24] - 17
[25] - 17
[26] - 21
[27] - 20
[28] - 19
[29] - 22
[30] - 18
```
It gives some important info to understand how this thread pool works.
with small enough burst of tasks (but not that small, ~1 ms), issued by several core clients (30 for this test) with some pauses (100 ms is the g1gc default pause target): 

- the load is spread among all threads ie each thread is getting ~100 tasks each
- each executor (client) is getting it's tasks executed by different threads (12->22 on 30 available)
- the number of created threads depends how busy existing ones are

In short, if the global thread executor is going to perform mostly non-blocking operations (NOTE: the I/O executor is responsible for I/O blocking ops), with enough clients (clients > available cores) we're going to use the whole number of threads configured on the pool. 
But if the max pool size exceed the available cores we will end up, similartly to the Netty case, to deschedule some at random, just to wake-up the next one in charge to handle a specific task.

There are few assumptions to be verified (what if `ArtemisExecutor` kept busy for too much time a specific Thread, global thread pool tasks cannot block? etc etc) and more tests to be performed, but this shouldn't stop from searching for better adapative (based on the machine spec) default IMO.

franz1981:
In short, I see no harm into reducing a bit the requirements for both thread pools, at least getting:

- netty to use N cores
- global thread pool to use N cores

Leaving the OS to handle, via context switches, the fairness of task executions among both.
I've decided (in this PR) to use half the core for both because of GC and background OS threads that can handle disk/networking tasks too.

michaelpearce-gain:
I do indeed trust 100% your intent and belief that this is a great improvement, im just saying we need some stats and testing to back it all up, in theory yes perfect, but as we've learnt before the best of intentions caused issues for others.

gtully:
I too would take the conservative approach on this change. I think it would make a great documentation update, and we can recommend to explicitly set values to balance io and workers across cores. It makes good sense.
Following happy user feedback, we can work to change the defaults in 2.19 or 3, ie: some point in the future.

franz1981:
@gtully @michaelandrepearce Agree with both of you: need some stats and proper use case load tests to back these new config although Netty thread pool config just doesn't sound "right" to me, but a workaround for the way HornetQ was using Netty event loops. 
I'm going to prepare some deeper investigation on this right after the quorum vote work has completed + doing some historic research on the default config we use: I'm sure @clebertsuconic got some pointers about it.
Thanks both for the feedbacks, always useful and wise :)

clebertsuconic:
can't you change ./artemis create to take the correct counting, and apply the new defaults on new created servers?

users moving older tested configurations wouldn't be affected by anything... new brokers would then use the "new defaults" and it would be up to the user to decide on what to do...

we can always log.info (" We recommend new values now.. please change your settings");

michaelpearce-gain:
This sounds like a good suggestion to me, it avoids existing users getting any surprises of defaults changing beneath their feet, but for new servers, they get the new sparkly settings.

Like wise, I like the idea that if someone is using default, e.g. not explicitly setting, then an info or warn could alert a user to this tuning, so they can at their own will, update their settings to the new sparkly settings.

michaelandrepearce:
@franz1981 whats occuring with this, i think an agree-able way forward was suggested by @clebertsuconic , but since didnt see any further updates, wanting to start clearing down old / stagnant PR's a bit of spring cleaning so to say - so we focus on stuff thats actively working on / relevant .

franz1981:
I'm keen to re-implement it as suggested by @clebertsuconic , but after some extensive performance testing to be sure is just the right choice for most users. Closing this, to be reopened in the future

PR #3585, opened 2021-05-17
NO-JIRA Specifying Maven Deploy Plugin as 2.4

clebertsuconic:
Using the latest (3.2.0) here will break the executables authorization for some reason

### DIFF
--- a/artemis-distribution/pom.xml
+++ b/artemis-distribution/pom.xml
@@ -263,6 +263,8 @@
          </plugin>
          <plugin>
             <artifactId>maven-assembly-plugin</artifactId>
+            <!-- if I don't specify version 2.4, the deploy is not setting the executables as expected -->
+            <version>2.4</version>
             <executions>
               <execution>
                   <id>source</id>


clebertsuconic:
@gemmellr the latest version of the maven deployment plugin is not reading the executor's argument on dep.xml, as a result a binary distribution would be broken.

Do you want to figure out why that happened? or can we just keep 2.4 here?

gemmellr:
*assembly plugin, rather than deploy.

I'll look into the issue first. 2.4 is about heading for 9 years old now, we should try to stay more current so that upgrades don't need to be as wide ranging or extreme as this one was. Essentially the issue is just that I spent so long on the rest that I obviously didn't run the binary again in the end, which I should have, and have broken it along the way without noticing.

gemmellr:
I have dropped the version back to 3.1.1 in https://github.com/apache/activemq-artemis/commit/7733a76649403193e0e0e7e1a6279095fdbf426c which got things working for me.

It seems that changes in 3.2.0 around reproducible builds changed some potentially undefined behaviour. The new behaviour exists in 3.3.0 as well. It seems perhaps some of the files in the assembly are relying on the old plugin behaviour, and so coupled with the descriptor not fully defining the permissions it wants this lead to the breakage on newer versions.

I think we should try ensuring the descriptor is updated to fully reflect what it wants and then update again, but for now the slightly-older but-not-entirely-ancient plugin version seems to be doing the trick.

PR #3586, opened 2021-05-18
NO-JIRA Fix maven compiler version for java 1.8

brusdev:


gemmellr:
Can you elaborate what is 'broken', is it just that '8 != 1.8'? As I understood when Emmanuel made the change, the compiler plugin understands the '8' and so it works. I'd expect it would fail before now otherwise since its been like this for around a year now.

If there is a real issue it probably warrants a JIRA.

gemmellr:
I should note I'm not against changing this, I use the original '1.8' style syntax elsewhere, its just '8' isnt really wrong.

brusdev:
Other poms could use the maven compiler version as default value for other purposes ie the enforcer `requireJavaVersion` rule. While the maven-compiler-plugin supports both `1.8` and `8` values, other plugin as `maven-enforcer-plugin` supports just the `1.8` value.

PR #3589, opened 2021-05-19
ARTEMIS-3294 Fix testTextMessageAttributeLimits

brusdev:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/QueueControlTest.java
```diff
@@ -579,7 +579,8 @@ public void testTextMessageAttributeLimits() throws Exception {
 
       final String twentyBytes = new String(new char[20]).replace("\0", "#");
 
-      ClientMessage clientMessage = createTextMessage(session, twentyBytes, true);
```
gtully, line 582: is there any value in the existing test?

I guess it is possible to put anything into a core message, it is up to the reader to know that it will be possibly nullable, no? Or was the assumption just wrong. It is an error to encode a plain string?

if it is not an error, then maybe we need to include both test variants.

brusdev: I would use a nullable string to write the body of TEXT_TYPE core messages to stay consistent with the JMS client. Maybe could be worth adding some checks to enforce the body format.

PR #3590, opened 2021-05-20
ARTEMIS-3309 Add a parameter to limit messages to move or transfer

brusdev:


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/Transfer.java
```diff
@@ -105,10 +105,207 @@
    @Option(name = "--target-topic", description = "Destination to be used. It can be prefixed with queue:// or topic:// and can be an FQQN in the form of <address>::<queue>. (Default: queue://TEST)")
    String targetTopic;
 
-   boolean isCopy() {
+   @Option(name = "--max-count", description = "The maximum number of messages to transfer.")
```
clebertsuconic, line 108: Can we make this consistent to ./artemis consumer... so this would come as --message-count ?

Can you check others...

if you want to expand towards those options.. isn't this something you could share a common base class with consumer?

brusdev: The `max-count` parameter has a different name from `message-count` because it has  a different semantic. The first one defines the maximum number of messages to transfer while the second one would define the exact number of messages to transfer waiting if the available messages are less than the `message-count` value (as the consumer does).

clebertsuconic: Seems pretty much the same to me.

./artemis consumer --message-count means.. the name of messages to consume...

While 

./artemis transfer --message-count would mean the maximum number of messages to transfer.


it would keep CLI clean.  things would get easier to remember when you move from one tool to another.

brusdev: I see, I have just pushed a commit to fix the parameter name.

PR #3591, opened 2021-05-20
ARTEMIS-3310 Paging could lose AMQPmessage if match didn't work for any reason

clebertsuconic:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/ParseAppMultiThread.java
```diff
@@ -0,0 +1,99 @@
...
+         persister.encode(buffer, originalMessage);
+         buffer.readerIndex(1);
+
+         AMQPStandardMessage amqpStandardMessage = (AMQPStandardMessage) persister.decode(buffer, null, null);
+
+
+         if (rep == 0) {
+            // it is enough to check the first time only
+            // this is to make sure the message does not have application properties parsed
+            Field field = AMQPMessage.class.getDeclaredField("applicationProperties");
```
franz1981, line 63: @clebertsuconic there is the state of parsing as an enum exposed for testing purposes

PR #3597, opened 2021-05-26
ARTEMIS-3314: fix remaining warnings during assembly build

gemmellr:
Carrying on after #3593, these changes fix all the remaining warnings emitted while building the assemblies, simplifying picking up on any future warnings issued.

### DIFF
--- a/artemis-distribution/pom.xml
+++ b/artemis-distribution/pom.xml
@@ -281,7 +281,21 @@
                   <id>bin</id>
                   <configuration>
                      <descriptors>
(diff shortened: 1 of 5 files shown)


gtully (commented):
great addition. thanks for taking the time to sort these. the assembly is really important and often neglected.

brusdev (approved):
brilliant, no more warnings on building artemis-distribution

clebertsuconic:
@gemmellr thanks a lot man

PR #3601, opened 2021-06-01
ARTEMIS-3323 - ensure openwire message id is unique and consistent fo…

gtully:
…r the life of a broker when converted from core

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/interop/GeneralInteropTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/interop/GeneralInteropTest.java
@@ -250,10 +250,51 @@ public class GeneralInteropTest extends BasicOpenWireTest {
       }
    }
 
+
+   @Test
+   public void testReceiveTwiceTheSameCoreMessage() throws Exception {
+
+      final String text = "HelloAgain";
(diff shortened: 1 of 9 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/interop/GeneralInteropTest.java
```diff
@@ -243,17 +243,60 @@ public void testFailoverReceivingFromCore() throws Exception {
          assertEquals(text + 1, ((TextMessage)message).getText());
          message.acknowledge();
 
+         System.err.println("id: " + message.getJMSMessageID());
```
clebertsuconic, line 246: logger.debug

is this by accident?

I had recently (an year or so ago) removed all System.outs and System.errs from the test suite.

clebertsuconic:
other than the System.err that seems an accident. (If you intended for it, it's best to use logger.debug() anyways) it LGTM.

As long as the test suite is green and if you fix the System.err.println() this is good to go.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/interop/GeneralInteropTest.java
gtully: good catch. removed that. thanks!

PR #3604, opened 2021-06-02
ARTEMIS-3326 - fix state visibility between netty and actor thread af…

gtully:
…ter initial connection info processing. fix and test


clebertsuconic:
LGTM if tests are good

PR #3605, opened 2021-06-02
ARTEMIS-3327 removing unecessary blocking operations on update and de…

clebertsuconic:


clebertsuconic:
@franz1981  ^

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -1075,14 +1067,27 @@ private void internalAppendUpdateRecord(long id,
...
                                            IOCompletion callback) throws InterruptedException, java.util.concurrent.ExecutionException {
-      final SimpleFuture<Boolean> result = newSyncAndCallbackResult(sync, callback);
       appendExecutor.execute(new Runnable() {
          @Override
          public void run() {
             journalLock.readLock().lock();
             try {
                JournalRecord jrnRecord = records.get(id);
+               if (jrnRecord == null) {
+                  if (compactor == null || (!compactor.containsRecord(id))) {
```
franz1981, line 1079: Save the compactor in a local final variable to save it load it each time

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -1075,14 +1067,27 @@ private void internalAppendUpdateRecord(long id,
...
+                     if (updateCallback != null) {
+                        updateCallback.accept(id, false);
+                     }
+                     logger.debug("Record " + id + " had not been found");
+
```
franz1981, line 1084: Debug if is debug level to save string creation

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -1075,14 +1067,27 @@ private void internalAppendUpdateRecord(long id,
...
       appendExecutor.execute(new Runnable() {
          @Override
          public void run() {
             journalLock.readLock().lock();
             try {
                JournalRecord jrnRecord = records.get(id);
+               if (jrnRecord == null) {
+                  if (compactor == null || (!compactor.containsRecord(id))) {
+                     if (updateCallback != null) {
+                        updateCallback.accept(id, false);
```
franz1981, line 1081: Michael P spent some time to save boxed Long instances to be allocated and I think we should continue here: let's create a specific interface that can use primitive long and boolean

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -1097,26 +1102,32 @@ public void run() {
                // record==null here could only mean there is a compactor
                // computing the delete should be done after compacting is done
                if (jrnRecord == null) {
-                  compactor.addCommandUpdate(id, usedFile, updateRecord.getEncodeSize());
+                  if (compactor != null) {
```
franz1981, line 1105: Save compactor in a local variable to save it to be loaded more then one time

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -1129,15 +1140,17 @@ public void appendDeleteRecord(final long id, final boolean sync, final IOComple
 
       checkJournalIsLoaded();
       lineUpContext(callback);
-      checkKnownRecordID(id, true);
-
-      internalAppendDeleteRecord(id, sync, callback);
+      SimpleFuture<Boolean> future = new SimpleFutureImpl<>();
```
franz1981, line 1143: I would create a specific future that can act as a callback too here: it would save the lambda allocation

franz1981 (changes requested):
Lovely change! And I like we have saved the map on the journal, is both nice for perf and memory footprint

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
clebertsuconic: the compactor is only changed on writeLock. it is safe to just use the compactor directly here.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
clebertsuconic: compactor is only changed on writeLock(). it is safe to use it here.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
clebertsuconic: this is not used at all.. it is just for tests.

Besides there's no change here.. the previous version was doing exactly the same.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
franz1981: Yes but it still an unnecessary cost: each time we use it, the jit will load it and null check it implicitly

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
clebertsuconic: done

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
clebertsuconic: I don't think that's a big deal but I will do it

clebertsuconic:
@franz1981 I am merging this.. if you have any further changes.. we can always send new commits...

Tests are solid here.

PR #3606, opened 2021-06-03
ARTEMIS-3328: assembly related improvements

gemmellr:
https://issues.apache.org/jira/browse/ARTEMIS-3328

- Fix assembly descriptors and stop downgrading the assembly plugin.
- Only build the source-release assembly during release (or on explicit request).
- Suppress creating and deploying (then deleting) unused assembly file configured by parent pom.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -140,18 +140,11 @@ scm.tag=1.4.0
 ```
 
 
-## Removing additional files
+## Closing the staging repository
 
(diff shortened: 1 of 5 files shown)


gemmellr:
I meant to remove that last commit before I pushed the rebase...but since its pure deletion of redundant bits and easy to distinguish from the rest regardless of being in its own commit, I'll just leave it be since it was a NO-JIRA effort anyway.

PR #3610, opened 2021-06-04
ARTEMIS-3261 Updating logic to use only replaceable records on compacting verification

clebertsuconic:


artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -323,20 +325,26 @@ public JournalImpl setHistoryFolder(File historyFolder, long maxBytes, long peri
    private final ReadWriteLock journalLock = new ReentrantReadWriteLock();
    private final ReadWriteLock compactorLock = new ReentrantReadWriteLock();
 
-   HashSet<Integer> replaceableRecords;
+   ByteObjectHashMap<Byte> replaceableRecords;
```
franz1981, line 328: I would use a BitSet: it would save tons of memory while being faster too.
As an alternative, if we know upfront all the record types I would enum them, using the enum as parameter instead of byte and using an EnumSet too.
IMO the solution with the BitSet is the best one in term of footprint/performance: it always need just a single long[] of capacity 1 if record type is >=0

franz1981 (changes requested):

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
franz1981: Iteration could be performed with 
```
for (int recordType = bitSet.nextSetBit(0); recordType != -1; recordType = bitSet.nextSetBit(recordType + 1)) {
    // do something with record type
}
```

franz1981: Given that I don't know if it's an hot path I am already happy with the PR as it is: this one is just an improvement, given that we are already changing collection.

franz1981 (approved):

PR #3612, opened 2021-06-05
ARTEMIS-3332 - fix search on consumer count in console

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-3332


brusdev (approved):
LGTM

PR #3614, opened 2021-06-07
ARTEMIS-3329 ability to purge all queues on address

jbertram:


artemis-commons/src/main/java/org/apache/activemq/artemis/logs/AuditLogger.java
```diff
@@ -2833,4 +2833,30 @@ static void isAutoCreated(Object source) {
...
+   @LogMessage(level = Logger.Level.INFO)
+   @Message(id = 601749, value = "User {0} has purged address {1}", format = Message.Format.MESSAGE_FORMAT)
+   void purgeAddressSuccess(String user, String queueName);
+
+
+   static void purgeAddressFailure(String queueName) {
+      RESOURCE_LOGGER.purgeAddressFailure(getCaller(), queueName);
+   }
+
+   @LogMessage(level = Logger.Level.INFO)
```
brusdev, line 2859: log the purgeAddressFailure with a a warning with the exception details could help investigate on failures because retrieving the exception details from console isn't immediate.

jbertram: I'm not sure that's appropriate for the audit log. There are *lots* of places where failures are logged in the audit log and there are no debugging details. This is just following that same pattern (e.g. from `pause`, `resume`, etc.).

clebertsuconic:
I saw the situation that needed this functionality.


it would be best if we added a method on PagingSTore to remove all the pages, and take it out of paging.

The situation that made this happen had pages stuck, although they were consumed already.


If you could as part of this task, when purgeAll is called, remove all the pages instead it would be a great addition. it would solve situations where paging is stuck due to whatever is happening underneath.

jbertram:
@clebertsuconic, I'm not sure I'm familiar enough with the paging implementation to do this. I had a look and the it appeared pretty nuanced.

michaelpearce-gain:
@clebertsuconic is this making it into the next release? as you know this is from a production operational issue you were inv with me.

PR #3618, opened 2021-06-10
ARTEMIS-3341: update checkstyle, use consistent plugin version

gemmellr:


jbertram:
Why not use this in `checkstyle.xml`:
```xml
      <module name="RightCurly">
         <property name="option" value="alone_or_singleline"/>
         <property name="tokens" value="METHOD_DEF, CTOR_DEF"/>
      </module>
```
Then you could drop your changes from `TypedProperties` and `AbstractJDBCDriver` and everything would be consistent.

As it stands now the change looks a bit strange because you're eliminating a newline in `SensitiveDataCodec` and adding one in the other two aforementioned classes.

gemmellr:
I actually prefer class definitions having the newline so I just went with the default. The change in SensitiveDataCodec was about originally about eliminating the semi colon which already failed the 'alone' check, and I just made it consistent with all the other default and empty methods that would have needed updated if not for config tweak.

jbertram:
> I actually prefer class definitions having the newline so I just went with the default.

By "class" do you mean "constructor"? The checkstyle `CLASS_DEF` and `CTOR_DEF` are different things, and I was talking about the latter.

> ...I just made it consistent with all the other default and empty methods that would have needed updated if not for config tweak.

There doesn't appear to be a consistent style for right curly brackets for `default` method definitions. A quick search through the code-base with a regular expression indicates that the newline version is far more prevalent than the single-line version. Only a few files use the single-line version with `org.apache.activemq.artemis.protocol.amqp.proton.handler.EventHandler` being the main culprit.

I don't really care *what* the style is. I just think it should be consistent across all method definitions, whether those be constructors, `default` methods, or normal methods. Given that and your preference for newlines for constructors (assuming that's what you meant) then I'd say you should just revert your change to `checkstyle.xml` and update the classes that violate the style as it is enforced with the new version. I believe that was what the original style intended.

gemmellr:
I did mean constructor, per the context. I actually disagree that methods and constuctors need to be treated the same, I consider them different things and I think having the constructor with the newline makes them stand out better and is more readable. For empty methods I don't see the need to enforce that, they often look better without. That there are only 2 constructors affected that way (but more methods affected the other) also suggest that was the prevailing preference up to this point, as clearly the old checkstyle wasn't enforcing it either way whereas the new one does. So all round I prefer what I did.

Ultimately I don't care enough about this to argue any more over it though, but I also won't invest any more time on adding newlines to all the methods affected, as I don't actually think that is needed, better looking, or more consistent. 

I was only interested in getting rid of the CVE warning so that other warnings can be more easily seen as they arise.

This proved way way more of a headache than expected and really is not worth the hassle it took to even get it to this point.

gemmellr:
I changed the checkstyle config per your original comment to allow the existing constructors and methods on single lines and giving the resulting minimal change here. I have now pushed that, further changes desired can go in separately.

jbertram:
OK, that's fair. Thanks, Robbie!

After reviewing this a bit more I've sent #3619 as a follow-up.

PR #3620, opened 2021-06-12
ARTEMIS-3343 throw exception on duplicate address-setting match

jbertram:


artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationParserTest.java
```diff
@@ -95,6 +95,22 @@ public void testDuplicateQueue() throws Exception {
...
    }
 
+   @Test
+   public void testDuplicateAddressSettings() throws Exception {
+      String filename = "FileConfigurationParser-duplicateAddressSettings.xml";
+      FileConfiguration fc = new FileConfiguration();
+      FileDeploymentManager deploymentManager = new FileDeploymentManager(filename);
+      deploymentManager.addDeployable(fc);
+      try {
+         deploymentManager.readConfiguration();
```
clebertsuconic, line 105: I bet we will get questions from users on why their server's are lot loading any longer?

Perhaps we should log.warn and just ignore the second entry?

michaelandrepearce:
What happens if i have foo and foo#...

jbertram:
@michaelandrepearce since `"foo".equals("foo#")` evaluates to `false` there would be no error on that case.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationParserTest.java
jbertram: I think logging and ignoring is viable, and it will preserve the current semantics. I originally decided to throw the exception here to match how we treat other invalid configurations (e.g. bad XML, `Validator` failures, etc.).

erwindon: @clebertsuconic I believe the current behaviour is that all but the last entry is ignored as the entries are first stored in a map
@jbertram I personally would not mind a fatal error on startup as I'm always testing new software versions. but totally acceptable when it is 'only' a warning.

clebertsuconic: @jbertram 

I think the right thing would be to throw the error like you're doing.. by the other hand.. I have seen some installations where the user has 100s of destinations configured.. and it will be quite possible to have a duplicate on those... 

Considering what I have seen some users doing, I think it would be wise to log.warn instead...

I have been using a pattern with ("****** 127 stars *****\nMYMESSAGE-GOES-HERE\n****127 stars ****") on places I really want to call the user's attention... (don't know if that's cheesy or not.. but it certainly catches an eye on logging).. I think we could do something like this...


WDYT?

erwindon: My list of address-settings is a bit more modest. The main reason why I fell into this trap is that I did not even realise that the name patterns should be unique.
The startup of Artemis does not produce that much output that one should fear that a warning would be missed.
I would not create a separate warning style for this.
When there is doubt between warning or error, just go for warning. It is the safest choice and still helps to solve the problem much more than the old situation (where I needed git-bisect to find the root cause, as I was clueless)

jbertram: @clebertsuconic, I've updated the code to just log a WARN and ignore the settings from the duplicate match.

PR #3621, opened 2021-06-14
ARTEMIS-3313 Fix import of anycast messages placed on DLA

AntonRoskvist:


jbertram:
This only fixes the issue for import and not retry, correct? Ideally the fix would resolve the issue for both use-cases. It may be better to simply unset the routing type when the message is first routed to the dead-letter or expiry address.

AntonRoskvist:
Yes, this only fixes the import. I should clarify the Jira, because these issues seem to be unrelated, at least as far as I can tell....
The first issue, which this PR is for, are messages "lost" when running an export/import. The messages are still in the export file but are destined for an address which have no queues of the same RoutingType as the messages have.

The other issue is unaffected by RoutingType and instead have to do with auto-deleted queues. When running auto-created resources for DLA messages and the original queue gets auto-removed, then retrying those messages sends them to nowhere. This issue also needs to be addressed I think but is not in scope for this PR (I don't think). Jira mentions them both because I ran into them at the same time.

brusdev:
Maybe we could preserve the original routing type in another property, ie
SimpleString HDR_ORIG_ROUTING_TYPE = new SimpleString("_AMQ_ORIG_ROUTING_TYPE");

AntonRoskvist:
@brusdev Added the HDR_ORIG_ROUTING_TYPE property

AntonRoskvist:
@jbertram Seems the retry was a simple fix, I added that now.
@brusdev Now also setting the original routingType when retrying messages from DLA and HDR_ORIG_ROUTING_TYPE is set

jbertram:
I'm still doing some investigating to see if this is the optimal solution.

AntonRoskvist:
Hello @jbertram, have you found any alternate solution for this or is there a special part of this PR you are uncertain about and want me to rewrite?

jbertram:
See #3782 for a better solution.

PR #3624, opened 2021-06-14
ARTEMIS-3155 support better backwads compatibility

jbertram:


clebertsuconic:
nit-pick.. commit message.. support better backwards compatibility on SSL properties


I can only read that context if I open the JIRA.. it would be better to explain at the message.

PR #3627, opened 2021-06-16
ARTEMIS-3286 do not use comma in queue name as it upsets the web-console

erwindon:
solve ARTEMIS-3286 by replacing the comma with a semicolon
(as a comma upsets the webgui)

tested as follows:
`artemis consumer --destination topic://foobar --durable --clientID myclientid`

afbeelding

See https://issues.apache.org/jira/browse/ARTEMIS-3286 for original result.


brusdev:
@erwindon thanks for this contribution but it is already available hawtio 2.13.5 with the fix for ARTEMIS-3286. I have created #3637 to upgrade the hawtio version to 2.13.5 and I would close this PR.

erwindon:
as long as either one of the fixes reached the next Artemis edition, no problem!

PR #3628, opened 2021-06-17
ARTEMIS-3354 journal-max-io cannot be set to 1 with ASYNCIO journal

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3354


brusdev (approved):
LGTM

PR #3629, opened 2021-06-17
ARTEMIS-3356: fix debug message arguments

michpetrov:
Issue: [ARTEMIS-3356](https://issues.apache.org/jira/browse/ARTEMIS-3356)

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
@@ -253,8 +253,8 @@ public final class Page implements Comparable<Page> {
 
    public synchronized List<PagedMessage> read(StorageManager storage, boolean onlyLargeMessages) throws Exception {
       if (logger.isDebugEnabled()) {
-         logger.debugf("reading page %d on address = %s onlyLargeMessages = %b", storeName, pageId,
-                       storage, onlyLargeMessages);
+         logger.debugf("reading page %d on address = %s onlyLargeMessages = %b",
+            new Object[] {pageId, storeName, onlyLargeMessages});
       }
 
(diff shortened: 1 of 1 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/Page.java
```diff
@@ -253,8 +253,8 @@ public synchronized PagedMessage readMessage(int startOffset,
 
    public synchronized List<PagedMessage> read(StorageManager storage, boolean onlyLargeMessages) throws Exception {
       if (logger.isDebugEnabled()) {
-         logger.debugf("reading page %d on address = %s onlyLargeMessages = %b", storeName, pageId,
-                       storage, onlyLargeMessages);
+         logger.debugf("reading page %d on address = %s onlyLargeMessages = %b", new Object[] {pageId,
```
brusdev, line 256: why the Object array?

jbertram: I had the same question. :smile:

clebertsuconic: that's like changing 6 for half dozen...
It's the same thing.. I don't understand what is the point?

michpetrov: Otherwise I get a conflict because there are two methods with similar signatures:
`debugf(String, int, Object, Object)`
`debugf(String, Object, Object, Object)`

brusdev: I see thanks

brusdev (approved):
LGTM +1

PR #3631, opened 2021-06-21
ARTEMIS-3362 - fix Session.CLIENT_ACKNOWLEDGE behaviour for core clie…

gtully:
…nt when consumer.receive is in play


gtully:
hmm, I had to revert to the specs, I think what I am expecting is JMS 1.1 semantics, and the openwire client gives me that. With the jms 2.0 core client, the session has to be closed, not just the consumer. that explains to some extent the assertions in the failing tests.

PR #3632, opened 2021-06-21
ARTEMIS-3177 allow a NOT-CONTAINS filter

erwindon:
As a solution to quickly filter out internal queues; but implemented as a generic solution.
To filter out internal queues: "[Name] [Does Not Contain] [$]".

tested in all relevant web gui pages.
tested in commandline: `artemis queue stat --operation NOT_CONTAINS --field NAME --value '$'`

considerations:
I could have added NOT-EQUALS; but I did not find this useful (as is just removes 1 item from the list)

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
@@ -63,7 +63,7 @@ public class StatQueue extends AbstractAction {
    }
 
    public enum OPERATION {
-      CONTAINS, EQUALS, GREATER_THAN, LESS_THAN
+      CONTAINS, NOT_CONTAINS, EQUALS, GREATER_THAN, LESS_THAN
    }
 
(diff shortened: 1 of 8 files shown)


michaelpearce-gain:
Nice! Just for completeness as touching this area anyhow, and so the option is there, is it worth adding not equals?

erwindon:
@michaelandrepearce This PR started its life as a solution to hide internal queues in an easy way. when applied to a "name", the not-equals operation would hide only one row and you would still need to enter that name. when I now look through the pages, I only found "[Message Count] [Not Equals] [0]" to be useful, but that can just as easily be done as "[Message Count] [Greater Than] [0]". When you (or someone else) provides me with a useful usecase for [Not Equals], I'll add it.

brusdev:
It LGTM, cloud you add a test, ie org.apache.activemq.cli.test.ArtemisTest#testQstat ?

erwindon:
@brusdev the testcase is added

brusdev (approved):

erwindon:
rebase in progress...

PR #3562, opened 2021-04-30
ARTEMIS-3275 Lock CORE client communication during failover retries

brusdev:

PR #3634, opened 2021-06-23
ARTEMIS-3365 Broker Balancers

brusdev:
This feature adds broker balancers to distribute the incoming client connections across multiple brokers. It provides a native redirection for supported clients and a new management API for other clients. The native redirection can be enabled per acceptor and is supported only for CORE and AMQP clients. See [ARTEMIS-3365](https://issues.apache.org/jira/browse/ARTEMIS-3365)[1] and the [draft documentation ](https://github.com/brusdev/activemq-artemis/blob/broker_balancers/docs/user-manual/en/broker-balancers.md)[2] for further details. 

[1] https://issues.apache.org/jira/browse/ARTEMIS-3365
[2] https://github.com/brusdev/activemq-artemis/blob/broker_balancers/docs/user-manual/en/broker-balancers.md


docs/user-manual/en/broker-balancers.md
```diff
@@ -0,0 +1,156 @@
+# Broker Balancers
+Apache ActiveMQ Artemis broker balancers allow incoming client connections to be distributed across multiple [target brokers](target-brokers).
+The target brokers are grouped in [pools](#pools) and the broker balancers use a [target key](#target-key)
+to select a target broker from a pool of brokers according to a [policy](#policies).
+
+## Target Broker
+Target broker is a broker that can accept incoming client connections and is local or remote.
+The local target is a special target that represents the same broker hosting the broker balancer.
+The remote target is another broker reachable by the broker that hosts the broker balancer.
```
gtully, line 9: it is just reachable. full stop. Does it really need to host a balancer?

docs/user-manual/en/broker-balancers.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Target Key
+The broker balancer uses a target key to select a target broker.
+It is a string retrieved from an incoming client connections, the supported values are:
```
gtully, line 13: client connection (not plural)

docs/user-manual/en/broker-balancers.md
```diff
@@ -0,0 +1,156 @@
...
+* `SNI_HOST` is the hostname indicated by the client in the SNI extension of the TLS protocol;
+* `SOURCE_IP` is the source IP address of the client;
+* `USER_NAME` is the username indicated by the client.
+
+## Pools
+The pool is a group of brokers and checks periodically their state.
+Including the local target broker in the pool allows broker hosting the broker balancer to accept incoming client connections. 
+By default, a pool doesn't include the local target broker, to include it the `local-target-enabled` parameter must be `true`.
+A pool becomes ready when the minimum number of ready targets defined by the `quorum-size` parameter is reached.
```
gtully, line 22: is it just quorum (the size is implicit, as in quorum is the minimum number of members that make the group valid.

docs/user-manual/en/broker-balancers.md
```diff
@@ -0,0 +1,156 @@
...
+This behaviour avoids weird distribution at startup or after a restart.
```
gtully, line 23: what happens when the quorum is set and not reached? Does the connection just get rejected? expecting a retry?

docs/user-manual/en/broker-balancers.md
```diff
@@ -0,0 +1,156 @@
...
+        <connector-ref>connector1</connector-ref>
+        <connector-ref>connector2</connector-ref>
+        <connector-ref>connector3</connector-ref>
+    </static-connectors>
+</pool>
+```
+
+## Policies
+The policy define how to select a broker from a pool. The included policies are:
+* `FIRST_ELEMENT` to select the first target broker from a pool;
```
gtully, line 74: the _ELEMENT seems odd? FIRST on it's own or HEAD (but that implies a list). Can you add a one liner when FIRST would make sense, ROUND_ROBIN with a single entry in a pool would behave the same.

docs/user-manual/en/broker-balancers.md
```diff
@@ -0,0 +1,156 @@
...
+* `LEAST_CONNECTIONS` to select the targets with the fewest active connections. This policy helps you maintain an equal distribution of active connections with the target brokers.
+
+A policy is defined by the `policy` element. Let's take a look at a policy example from broker.xml:
+```xml
+<policy name="FIRST_ELEMENT"/>
+```
+
+## Cache
+The broker balancer provides a cache with a timeout to improve the stickiness of the target broker selected.
+The broker balancer returns the same target broker for a target key until the target key is present in the cache and is ready.
```
gtully, line 86: until? should that be "as long as"

docs/user-manual/en/broker-balancers.md
```diff
@@ -0,0 +1,156 @@
...
+Let's take a look at flowchart of the broker balancer workflow:
+![Broker Balancer Workflow](images/broker_balancer_workflow.png)
+
+
+## Redirection
+Apache ActiveMQ Artemis provides a native redirection for supported clients and a new management API for other clients.
+The native redirection can be enabled per acceptor and is supported only for CORE and AMQP clients.
+The acceptor with the `redirect-to` url parameter will redirect the incoming connections.
+The `redirect-to` url parameter specifies the name of the broker balancer to use,
+ie the following acceptor will redirect the incoming CORE client connections using the broker balancer with the name `bb1`:
```
gtully, line 147: maybe it makes sense to reuse the name from above. in place of bb1 use 'consistent-hash-balancer' just to tie the url parameter to the name element of the xml. If I have that correct!

docs/user-manual/en/broker-balancers.md
gtully: that comment is for line 9.

michaelandrepearce:
Is it possible that this can be used to shard an address across brokers and have clients send to different brokers based on mgroups via a consistent hash? E.g. client has smarts to send mgroup A to broker 1 mgroup B to broker 3 and mgroup C to broker 2 for example where the mgroup to broker is by consistent hash

brusdev:
@michaelandrepearce 
> Is it possible that this can be used to shard an address across brokers and have clients send to different brokers based on mgroups via a consistent hash? E.g. client has smarts to send mgroup A to broker 1 mgroup B to broker 3 and mgroup C to broker 2 for example where the mgroup to broker is by consistent hash

The client connections can be distributed across brokers using a consistent hash based on USER_NAME. So supposed to match the client users with the mgroups then each client is redirected to the broker with the hash code closest to the hash code of client user/mgroup.

docs/user-manual/en/broker-balancers.md
brusdev: if the quorum is not reached and the target key is not present in the cache then the connection is rejected

brusdev: I'll add further details to the redirection section

docs/user-manual/en/broker-balancers.md
brusdev: I can't find a better name so I used `FIRST_ELEMENT` to be consistent with the similar client load balancing policy: org.apache.activemq.artemis.api.core.client.loadbalance.FirstElementConnectionLoadBalancingPolicy

michaelandrepearce:
Ok so thats a no then. ;) needs to be mgroups has to be flexible as typically grouping is done by a business key. Ill continue with the thing ive been working on then.

docs/user-manual/en/broker-balancers.md
brusdev: I used the `quorum-size` name just to be consistent with other parameter names, see replicatedPolicyType and replicaPolicyType.

gtully:
> Ok so thats a no then. ;) mgroups has to be flexible as typically grouping is done by a business key. Ill continue with the thing ive been working on then.

If I understand correctly, that is about sharding producers that will set a groupId on messages? I guess this could be layered over a component of the clientId, using the group as a clientId prefix. Then sharding would be on some component of the clientID. key=clientId filter=<some regexp that extracts the groupd Id shard component>

would that work for your use case?

gtully:
having a key-filter or target-key-filter applied to the raw key would be a nice addition. Even for the client-IP policy, being able to shard on just some of the components of an IP is powerful when IP ranges are carefully managed.

michaelandrepearce:
@gtully not really. The keys are business keys, akin to why you may want to use kafka partition keys. Theres actually a nice implementation in the old fabric8 that im looking at pulling across, update and enhance.

brusdev:
I have just pushed a commit to add the `CLIENT_ID` target key. Now the broker balancer can use the jms client ID to distribute the incoming client connections across multiple brokers.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -469,7 +473,45 @@ public void onRemoteOpen(Connection connection) throws Exception {
...
+               connection.setCondition(error);
+
+               redirected = true;
+            }
+         } else {
+            ActiveMQServerLogger.LOGGER.cannotRedirectClientConnection(transportConnection);
+
+            ErrorCondition error = new ErrorCondition();
+            error.setCondition(ConnectionError.CONNECTION_FORCED);
+            error.setDescription(ConnectionError.CONNECTION_FORCED.toString());
```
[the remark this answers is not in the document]
brusdev: @gemmellr yes, the redirect to the local target is a pass through. Indeed if target.isLocal() then redirected = false and the connection is managed like any old connection would without the balancer, ie the `RedirectTest#testSymmetricRedirect` test uses the local targets:
https://github.com/brusdev/activemq-artemis/blob/broker_balancers/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/RedirectTest.java#L221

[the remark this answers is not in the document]
gemmellr: Oops misread diff, that else bit is when target==null. General sentiment of 'give a more useful error description' still applies, i.e give some detail so poor user can perhaps understand and maybe help themselves if they did something wrong. As is a client is likely to just retry, possibly in a loop if this case repeats, with no real indication whats going on.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -469,7 +474,46 @@ public void onRemoteOpen(Connection connection) throws Exception {
...
+         if (target != null) {
+            ActiveMQServerLogger.LOGGER.redirectClientConnection(transportConnection, target);
+
+            if (!target.isLocal()) {
+               String host = ConfigurationHelper.getStringProperty(TransportConstants.HOST_PROP_NAME, TransportConstants.DEFAULT_HOST, target.getConnector().getParams());
+               int port = ConfigurationHelper.getIntProperty(TransportConstants.PORT_PROP_NAME, TransportConstants.DEFAULT_PORT, target.getConnector().getParams());
+
+               ErrorCondition error = new ErrorCondition();
+               error.setCondition(ConnectionError.REDIRECT);
+               error.setDescription(String.format("Connection redirected to %s by broker balancer %s", brokerBalancer.getName(), target));
```
gemmellr, line 495: Are the args possibly reversed? Havent tried it, but the format string seems to read more like the balancer name is expected to be second/last?

Either way, the toString() on Target seems like this might result in including a bunch of config/names some might consider 'internal' to the broker setup into the description. Or perhaps just not entirely relevant to an AMQP client/user that doesnt use them. 

It might be simpler and clearer just to use the target host:port, which are already available here.

brusdev: yes the args are reversed. I'll replace the target.toString() with host:port

clebertsuconic:
@brusdev rebase it please?

brusdev:
@clebertsuconic I have just rebased this PR but it depends on #3562 PR.

examples/features/broker-balancer/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+   <name>ActiveMQ Artemis Broker Balancer Examples</name>
+
+   <!-- Properties -->
+   <properties>
+      <!--
+      Explicitly declaring the source encoding eliminates the following
+      message: [WARNING] Using platform encoding (UTF-8 actually) to copy
+      filtered resources, i.e. build is platform dependent!
+      -->
+      <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```
gemmellr, line 42: Dont think this is needed, the apache and artemis parent poms both look to be setting it already. Certainly the comment can go.

examples/features/broker-balancer/evenly-redirect/src/main/java/org/apache/activemq/artemis/jms/example/EvenlyRedirectExample.java
```diff
@@ -0,0 +1,102 @@
...
+
+         if (connectionProducer1 != null) {
+            connectionProducer1.close();
+         }
+      }
+
+      /**
+       * Step 2. create a connection for consumer0 and consumer1, and receive a few messages.
+       * the server0 will redirect the connection to the same target broker of the respective producer
+       * because the consumer and the producer share the same clientID.
```
gemmellr, line 73:
```suggestion
       * the server0 will redirect the connection to the same target broker of the respective producer
       * from earlier as the new consumer connection uses the same ClientID.
```

Avoiding saying 'share the same ClientID' since only one connection at a time can use a ClientID

gemmellr: Actually, as this uses 2 different ClientIDs but is doing 'least connections', is the ClientID relevant? When the second set of connections happens...what guarantee is it that they go to the same place as the producer if it is only doing 'least connections' balance and both servers have 0? Would it work as described if I changed the order of connection creation on the second round?

examples/features/broker-balancer/evenly-redirect/readme.md
```diff
@@ -0,0 +1,6 @@
+# AMQP Broker Connection with Receivers
```
gemmellr, line 1: Title needs fixed

docs/user-manual/en/broker-balancers.md
```diff
@@ -0,0 +1,176 @@
...
+* `SOURCE_IP` is the source IP address of the client;
+* `USER_NAME` is the username indicated by the client.
+
+## Pools
+The pool is a group of target brokers and checks periodically their state.
+It provides a list of ready target brokers to distribute incoming client connections only when it is active.
+A pool becomes active when the minimum number of ready target brokers defined by the `quorum-size` parameter is reached.
+When it is not active, it doesn't provide any target avoiding weird distribution at startup or after a restart.
+Including the local target broker in the pool allows broker hosting the broker balancer to accept incoming client connections. 
+By default, a pool doesn't include the local target broker, to include it the `local-target-enabled` parameter must be `true`.
```
gemmellr, line 27:
```suggestion
By default, a pool doesn't include the local broker, to include it as a target the `local-target-enabled` parameter must be `true`.
```
Moving 'target' for clarity, as it isnt a 'target broker' yet if it isnt in the pool, which is whats explained next.

docs/user-manual/en/broker-balancers.md
```diff
@@ -0,0 +1,176 @@
...
+        <connector-ref>connector3</connector-ref>
+    </static-connectors>
+</pool>
+```
+
+## Policies
+The policy define how to select a broker from a pool. The included policies are:
+* `FIRST_ELEMENT` to select the first target broker from a pool. It is useful to select the ready target brokers
+  according to the priority defined with their sequence order, ie supposing to have 2 target brokers
+  this policy selects the second target broker only when the first target broker isn't ready.
```
gemmellr, line 81:
```suggestion
* `FIRST_ELEMENT` to select the first target broker from the pool which is ready. It is useful to select the ready target brokers
  according to the priority defined with their sequence order, ie supposing there are 2 target brokers
  this policy selects the second target broker only when the first target broker isn't ready.
```

docs/user-manual/en/broker-balancers.md
```diff
@@ -0,0 +1,176 @@
...
+
+## Cache
+The broker balancer provides a cache with a timeout to improve the stickiness of the target broker selected.
+The broker balancer returns the same target broker for a target key as long as it is present in the cache and is ready.
+By default, the cache is enabled, to disable the cache the `cache-timeout` parameter must be `0`.
+
+## Defining broker balancers
+A broker balancer is defined by `broker-balancer` element, it includes the following items:
+* the `name` attribute defines the name of the broker balancer;
+* the `target-key` element defines what key to select a target broker, the supported values are: `SNI_HOST`, `SOURCE_IP`, `USER_NAME`.
```
gemmellr, line 99:
```suggestion
* the `target-key` element defines what key to select a target broker, the supported values are: `SNI_HOST`, `SOURCE_IP`, `USER_NAME`, `CLIENT_ID`.
```
Missing CLIENT_ID

examples/features/broker-balancer/symmetric-redirect/src/main/java/org/apache/activemq/artemis/jms/example/SymmetricRedirectExample.java
```diff
@@ -0,0 +1,106 @@
...
+
+         if (connectionProducer1 != null) {
+            connectionProducer1.close();
+         }
+      }
+
+      /**
+       * Step 2. create a connection for consumer0 and consumer1, and receive a few messages.
+       * the server0 will redirect the connection to the same target broker of the respective producer
+       * because the consumer and the producer share the same clientID.
```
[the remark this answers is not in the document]
brusdev: I meant clientID prefix, I'll fix it.

examples/features/broker-balancer/evenly-redirect/src/main/java/org/apache/activemq/artemis/jms/example/EvenlyRedirectExample.java
brusdev: The broker balancer defined int the evenly-redirect example uses a cache based on the CLIENT_ID key to provide affinity. I'll change the example to prove it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/TargetKeyTest.java
```diff
@@ -0,0 +1,162 @@
...
+   public void testSNIHostKey() throws Exception {
+      setupLiveServerWithDiscovery(0, GROUP_ADDRESS, GROUP_PORT, true, true, false);
+      getDefaultServerAcceptor(0).getParams().put(TransportConstants.SSL_ENABLED_PROP_NAME, true);
+      getDefaultServerAcceptor(0).getParams().put(TransportConstants.KEYSTORE_PATH_PROP_NAME, "verified-localdomain-keystore.jks");
+      getDefaultServerAcceptor(0).getParams().put(TransportConstants.KEYSTORE_PASSWORD_PROP_NAME, "secureexample");
+
+      setupBalancerServerWithDiscovery(0, TargetKey.SNI_HOST, MOCK_POLICY_NAME, null, true, null, 1);
+      startServers(0);
+
+      ConnectionFactory connectionFactory = createFactory(protocol, true, "localhost.localdomain",
```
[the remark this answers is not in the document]
gemmellr: You cant for Qpid JMS, dont know for openwire.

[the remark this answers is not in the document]
brusdev: I have just pushed a commit to skip this test if `localhost.localdomain` is not valid.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/TargetKeyTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+      Assert.assertEquals(1, keys.size());
+      Assert.assertEquals("test", keys.get(0));
+   }
+
+   @Test
+   public void testSNIHostKey() throws Exception {
+      setupLiveServerWithDiscovery(0, GROUP_ADDRESS, GROUP_PORT, true, true, false);
+      getDefaultServerAcceptor(0).getParams().put(TransportConstants.SSL_ENABLED_PROP_NAME, true);
+      getDefaultServerAcceptor(0).getParams().put(TransportConstants.KEYSTORE_PATH_PROP_NAME, "verified-localdomain-keystore.jks");
```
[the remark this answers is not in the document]
brusdev: keystore replaced

examples/features/broker-balancer/symmetric-redirect/readme.md
```diff
@@ -0,0 +1,5 @@
+# AMQP Broker Connection with Receivers
```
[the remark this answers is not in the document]
brusdev: title fixed

examples/features/broker-balancer/symmetric-redirect/readme.md
```diff
@@ -0,0 +1,5 @@
...
+
+To run the example, simply type **mvn verify** from this directory, or **mvn -PnoServer verify** if you want to create and start the broker manually.
+
+This example demonstrates how you can distribute incoming client connections across two brokers using a symmetric architecture.
```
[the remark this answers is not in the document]
brusdev: I have added a few words about the 'symmetric architecture'

examples/features/broker-balancer/pom.xml
brusdev: removed

examples/features/broker-balancer/evenly-redirect/src/main/java/org/apache/activemq/artemis/jms/example/EvenlyRedirectExample.java
brusdev: I missed to commit the updated example. Now it is updated and it uses `<target-key-filter>^.{3}</target-key-filter>`

examples/features/broker-balancer/symmetric-redirect/src/main/java/org/apache/activemq/artemis/jms/example/SymmetricRedirectExample.java
```diff
@@ -0,0 +1,107 @@
...
+public class SymmetricRedirectExample {
+
+   public static void main(final String[] args) throws Exception {
+
+      /**
+       * Step 1. Create a connection for producer0 and producer1, and send a few messages.
+       * the server0 will redirect the connection of each producer to a different target brokers.
+       */
+      ConnectionFactory connectionFactory0Server0 = new ActiveMQConnectionFactory("tcp://localhost:61616?ha=true&reconnectAttempts=30&clientID=FOO_PRODUCER");
+      ConnectionFactory connectionFactory1Server0 = new ActiveMQConnectionFactory("tcp://localhost:61616?ha=true&reconnectAttempts=30&clientID=TOO_PRODUCER");
```
gemmellr, line 42:
```suggestion
       * the server0 will redirect the connection of each producer to a different target broker.
       */
      ConnectionFactory connectionFactory0Server0 = new ActiveMQConnectionFactory("tcp://localhost:61616?ha=true&reconnectAttempts=30&clientID=FOO_PRODUCER");
      ConnectionFactory connectionFactory1Server0 = new ActiveMQConnectionFactory("tcp://localhost:61616?ha=true&reconnectAttempts=30&clientID=BAR_PRODUCER");
```

examples/features/broker-balancer/symmetric-redirect/src/main/java/org/apache/activemq/artemis/jms/example/SymmetricRedirectExample.java
```diff
@@ -0,0 +1,107 @@
...
+         }
+      }
+
+      /**
+       * Step 2. create a connection for consumer0 and consumer1, and receive a few messages.
+       * the server1 will redirect the connection to the same target broker of the respective producer
+       * from earlier as the new consumer connection uses the same ClientID prefix.
+       */
+      ConnectionFactory connectionFactory0Server1 = new ActiveMQConnectionFactory("tcp://localhost:61617?ha=true&reconnectAttempts=30&clientID=TOO_CONSUMER");
+      ConnectionFactory connectionFactory1Server1 = new ActiveMQConnectionFactory("tcp://localhost:61617?ha=true&reconnectAttempts=30&clientID=FOO_CONSUMER");
```
gemmellr, line 78:
```suggestion
      ConnectionFactory connectionFactory0Server1 = new ActiveMQConnectionFactory("tcp://localhost:61617?ha=true&reconnectAttempts=30&clientID=BAR_CONSUMER");
      ConnectionFactory connectionFactory1Server1 = new ActiveMQConnectionFactory("tcp://localhost:61617?ha=true&reconnectAttempts=30&clientID=FOO_CONSUMER");
```

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/TargetKeyTest.java
```diff
@@ -0,0 +1,180 @@
...
+
+      Assert.assertEquals(1, keys.size());
+      Assert.assertEquals("test", keys.get(0));
+   }
+
+   @Test
+   public void testSNIHostKey() throws Exception {
+      final String localHostname = "localhost.localdomain";
+
+      assumeLocalHostname(localHostname);
```
gemmellr, line 114: Personally I'd still prefer a test that ran everywhere connecting to localhost (with non-localhost SNI with Core client possible), over a test that needs checking if it ran. A unit test could probably also be used to ensure the connection is passed to the validator, and that it sees non-"localhost" values pass through from the connection.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/TargetKeyTest.java
```diff
@@ -0,0 +1,180 @@
...
+      startServers(0);
+
+      ConnectionFactory connectionFactory = createFactory(protocol, true, localHostname,
+         TransportConstants.DEFAULT_PORT + 0, null, null, null);
+
+      try (Connection connection = connectionFactory.createConnection()) {
+         connection.start();
+
+         connection.setExceptionListener(exception -> {
+            return;
```
gemmellr, line 131: Why was this no-op just added? Did you mean to add a check it didnt fire?

docs/user-manual/en/broker-balancers.md
```diff
@@ -0,0 +1,176 @@
...
+            <discovery-group-ref discovery-group-name="dg1"/>
+        </pool>
+    <policy name="CONSISTENT_HASH"/>
+    </broker-balancer>
+    <broker-balancer name="least-connections-balancer">
+        <policy name="LEAST_CONNECTIONS"/>
+        <pool>
+            <discovery-group-ref discovery-group-name="dg2"/>
+        </pool>
+    </broker-balancer>
```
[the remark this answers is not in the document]
brusdev: By default, the target key is SOURCE_IP and cache timeout is infinite.

examples/features/broker-balancer/evenly-redirect/readme.md
```diff
@@ -0,0 +1,6 @@
+# Evenly Redirect Example
+
+To run the example, simply type **mvn verify** from this directory, or **mvn -PnoServer verify** if you want to create and start the broker manually.
+
+This example demonstrates how incoming client connections are evenly redirected across two brokers
+using a third broker with a broker balancer to redirect incoming client connections.
```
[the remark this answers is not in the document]
brusdev: yes I would need a persistent cache to get Durable Subscription after a restart

examples/features/broker-balancer/evenly-redirect/src/main/java/org/apache/activemq/artemis/jms/example/EvenlyRedirectExample.java
```diff
@@ -0,0 +1,105 @@
...
+public class EvenlyRedirectExample {
+
+   public static void main(final String[] args) throws Exception {
+
+      /**
+       * Step 1. Create a connection for producer0 and producer1, and send a few messages.
+       * the server0 will redirect the connection of each producer to a different target brokers.
+       */
+      ConnectionFactory connectionFactoryProducer0 = new ActiveMQConnectionFactory("tcp://localhost:61616?ha=true&reconnectAttempts=30&clientID=FOO_PRODUCER");
+      ConnectionFactory connectionFactoryProducer1 = new ActiveMQConnectionFactory("tcp://localhost:61616?ha=true&reconnectAttempts=30&clientID=TOO_PRODUCER");
```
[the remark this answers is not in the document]
brusdev: updated

examples/features/broker-balancer/evenly-redirect/src/main/java/org/apache/activemq/artemis/jms/example/EvenlyRedirectExample.java
```diff
@@ -0,0 +1,105 @@
...
+
+         if (connectionProducer1 != null) {
+            connectionProducer1.close();
+         }
+      }
+
+      /**
+       * Step 2. create a connection for consumer0 and consumer1, and receive a few messages.
+       * the server0 will redirect the connection to the same target broker of the respective producer
+       * because the consumer and the producer share the same clientID prefix.
```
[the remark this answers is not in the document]
brusdev: updated

examples/features/broker-balancer/evenly-redirect/src/main/java/org/apache/activemq/artemis/jms/example/EvenlyRedirectExample.java
```diff
@@ -0,0 +1,105 @@
...
+       */
+      ConnectionFactory connectionFactoryConsumer0 = new ActiveMQConnectionFactory("tcp://localhost:61616?ha=true&reconnectAttempts=30&clientID=TOO_CONSUMER");
+      ConnectionFactory connectionFactoryConsumer1 = new ActiveMQConnectionFactory("tcp://localhost:61616?ha=true&reconnectAttempts=30&clientID=FOO_CONSUMER");
```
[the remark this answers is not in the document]
brusdev: updated

examples/features/broker-balancer/symmetric-redirect/src/main/java/org/apache/activemq/artemis/jms/example/SymmetricRedirectExample.java
brusdev: updated

examples/features/broker-balancer/symmetric-redirect/src/main/java/org/apache/activemq/artemis/jms/example/SymmetricRedirectExample.java
brusdev: updated

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/TargetKeyTest.java
brusdev: Now the test is executed at least with CORE protocol

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/TargetKeyTest.java
brusdev: removed

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/TargetKeyTest.java
gemmellr: Not quite what I was suggesting but hopefully does the trick with the fallbacks :)

(I was suggesting, have the core client connect to "localhost" but override the SNI to something else - like "localhost.localdomain" as the cert already matches that too)

docs/user-manual/en/broker-balancers.md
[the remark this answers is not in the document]
gemmellr: I don't recall seeing those values mentioned in the docs, might be worth adding, or making them more obvious if there.

examples/features/broker-balancer/evenly-redirect/readme.md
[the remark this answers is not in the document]
gemmellr: Yep, and I think its fine not having that, just thought it might be worth noting (maybe in the code comment instead) as the caching may catch some folks out if they dont consider that.

clebertsuconic:
I only have two questions:

I - I see a few trust stores changed as part of this PR? why?
II - I didn't like the fact that you have to instantiate an AMQP Redirector for every connection created. even if there is no Load Balancer in place. Is there a way to pass the connection as an argument and do not instantiate it? Or check before you create the instance.. like only perform the check if a redirector is configured?   that will give it a hit on the regular path for AMQP Connections. It seems a small thing but I have seen perf issues associated with new instances in a lot of circunstances.

docs/user-manual/en/broker-balancers.md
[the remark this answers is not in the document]
brusdev: doc updated

examples/features/broker-balancer/evenly-redirect/readme.md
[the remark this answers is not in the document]
brusdev: doc updated

brusdev:
@clebertsuconic I have regenerate security resources after adding other server names to the `build.sh` script. I have just pushed a commit to avoid creating new redirect objects if not required.

gemmellr:
I dont think the keystore changes should using NO-JIRA in their log, at least some are being made directly for this JIRA so it makes sense for it to be reflected in the commit log.

I think Clebert was looking to avoid creating the redirect handler objects entirely, which the updated version doesnt appear to. It still looks to affect every connection, but instead of with the throwaway handler, by adding a new field and creating the handler up front, then just possibly not using it. In some ways this is actually worse than previous version.

I think there is a nicer route in-between, but nearer the earlier version 2212d02f206bbdf9916db225e876eef4583c1326 [1], where it simply checks the redirect-to value on the connection first, and only then proceeds to create and use the handler if there was a need. E.g changes from the previous version for AMQP could maybe have been something like this instead (+add similar for other protocols):

```
@@ -528,6 +528,5 @@ public class AMQPConnectionContext extends ProtonInitializable implements EventH
 
-      AMQPRedirectHandler redirectHandler = new AMQPRedirectHandler(protocolManager.getServer(),
-         connectionCallback, connection, handler.getSASLResult() != null ? handler.getSASLResult().getUser() : null);
+      final boolean tryRedirect = connectionCallback.getTransportConnection().getRedirectTo() != null;
 
-      if (redirectHandler.redirect() || !validateConnection(connection)) {
+      if (tryRedirect && redirect(connection) || !validateConnection(connection)) {
          connection.close();
@@ -557,2 +556,9 @@ public class AMQPConnectionContext extends ProtonInitializable implements EventH
 
+   private final boolean redirect(Connection connection) throws Exception {
+      AMQPRedirectHandler redirectHandler = new AMQPRedirectHandler(protocolManager.getServer(),
+            connectionCallback, connection, handler.getSASLResult() != null ? handler.getSASLResult().getUser() : null);
+
+      return redirectHandler.redirect();
+   }
+
    class ScheduleOperator implements UnaryOperator<Future<?>> {
```


[1]  I still had the the 2212d02f206bbdf9916db225e876eef4583c1326 version locally, I put it at https://github.com/gemmellr/activemq-artemis/tree/balancer-pr3634-old1 in case you should want it.

gemmellr:
Ah, so @brusdev nudged me toward the fact the protocol manager isnt created per-connection, so there is indeed now less of the handlers being made with this version, with it now being the new contexts created per-connection and passed around to the existing handler, only after the check of 'redirect-to' presence.

I'd say I still prefer the simpler code from before just creating the handler, but only when its needed per the above suggestion.

clebertsuconic:
@gemmellr I think it's pretty simple... the redirectHandler will do such check, and then create a new instance on the context when needed.

You could make use of some static classes to avoid an initial creation, but it does not make a difference on this case since it's only one instance per acceptor.

I will merge this now


(However I will  amend the JIRA on the security artifacts change)

gemmellr:
The redirect handler doesnt do the important check though (it did do it before your original comment), its now already done outside it before you even get it, the same way I was suggesting, which meant everything else changed was unnecessary complexity.

Its basically still instantiating the same amount of stuff now (most of it its just not done unless there is a redirect-to), but just split it up into different places, making it a more complicated and verbose, and meaning a bunch of unrelated code needed updated.

brusdev:
@gemmellr I see your point, I'll try to reduce complexity (maybe not this release). Thanks a lot for your reviews and suggestions.

PR #3635, opened 2021-06-24
ARTEMIS-2007 - allow redistribution if there are unmatched messages p…

gtully:
…ending on a queue and there is new remote demand

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -113,7 +113,6 @@ import org.apache.activemq.artemis.utils.collections.NodeStore;
 import org.apache.activemq.artemis.utils.collections.LinkedListIterator;
 import org.apache.activemq.artemis.utils.collections.PriorityLinkedList;
 import org.apache.activemq.artemis.utils.collections.PriorityLinkedListImpl;
-import org.apache.activemq.artemis.utils.collections.SingletonIterator;
 import org.apache.activemq.artemis.utils.collections.TypedProperties;
 import org.apache.activemq.artemis.utils.critical.CriticalComponentImpl;
 import org.apache.activemq.artemis.utils.critical.EmptyCriticalAnalyzer;
@@ -188,6 +187,8 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
    private boolean mirrorController;
 
+   private volatile boolean hasUnMatchedPending = false;
+
    // Messages will first enter intermediateMessageReferences
    // Before they are added to messageReferences
    // This is to avoid locking the queue on the producer
@@ -260,7 +261,8 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
    private final SimpleString address;
 
-   private ConsumerHolder<Redistributor> redistributor;
+   // redistributor goes in the consumers list, this signals its presence and allows for easy comparison/check
+   private volatile ConsumerHolder<Redistributor> redistributor;
 
    private ScheduledFuture<?> redistributorFuture;
 
@@ -1326,12 +1328,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
       getExecutor().execute(new Runnable() {
          @Override
          public void run() {
-            try {
-               cancelRedistributor();
-            } catch (Exception e) {
-               // nothing that could be done anyway.. just logging
-               ActiveMQServerLogger.LOGGER.unableToCancelRedistributor(e);
-            }
+            cancelRedistributor();
          }
       });
 
@@ -1391,7 +1388,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
             return false;
          }
 
-         if (consumers.size() >= consumersBeforeDispatch) {
+         if (getConsumerCount() >= consumersBeforeDispatch) {
             if (dispatchingUpdater.compareAndSet(this, BooleanUtil.toInt(false), BooleanUtil.toInt(true))) {
                dispatchStartTimeUpdater.set(this, System.currentTimeMillis());
             }
@@ -1419,7 +1416,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
       try (ArtemisCloseable metric = measureCritical(CRITICAL_CONSUMER)) {
          synchronized (this) {
-            if (maxConsumers != MAX_CONSUMERS_UNLIMITED && consumers.size() >= maxConsumers) {
+            if (maxConsumers != MAX_CONSUMERS_UNLIMITED && getConsumerCount() >= maxConsumers) {
                throw ActiveMQMessageBundle.BUNDLE.maxConsumerLimitReachedForQueue(address, name);
             }
 
@@ -1488,7 +1485,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
             if (consumerRemoved) {
                consumerRemovedTimestampUpdater.set(this, System.currentTimeMillis());
-               if (consumers.size() == 0) {
+               if (getConsumerCount() == 0) {
                   stopDispatch();
                }
             }
@@ -1525,11 +1522,6 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
             supports = false;
          }
       }
-      if (redistributor != null) {
-         if (!redistributor.consumer.supportsDirectDelivery()) {
-            supports = false;
-         }
-      }
       return supports;
    }
 
@@ -1540,10 +1532,11 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
       if (redistributor != null) {
          // Just prompt delivery
          deliverAsync();
+         return;
       }
 
       if (delay > 0) {
-         if (consumers.isEmpty()) {
+         if (consumers.isEmpty() || hasUnMatchedPending) {
             DelayedAddRedistributor dar = new DelayedAddRedistributor(executor);
 
             redistributorFuture = scheduledExecutor.schedule(dar, delay, TimeUnit.MILLISECONDS);
@@ -1562,17 +1555,34 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    }
 
    @Override
-   public synchronized void cancelRedistributor() throws Exception {
-      if (redistributor != null) {
-         redistributor.consumer.stop();
-         redistributor = null;
-      }
-
+   public synchronized void cancelRedistributor() {
       clearRedistributorFuture();
+
+      if (redistributor != null) {
+         try {
+            redistributor.consumer.stop();
+         } catch (Exception e) {
+            ActiveMQServerLogger.LOGGER.unableToCancelRedistributor(e);
+         } finally {
+            consumers.remove(redistributor);
+            redistributor = null;
+         }
+      }
    }
 
    @Override
    public int getConsumerCount() {
+      // we don't want to count the redistributor, it is an internal transient entry in the consumer list
+      if (redistributor != null) {
+         synchronized (this) {
+            final int size = consumers.size();
+            if (size > 0 && redistributor != null) {
+               return size - 1;
+            } else {
+               return size;
+            }
+         }
+      }
       return consumers.size();
    }
 
@@ -1775,7 +1785,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    public Map<String, List<MessageReference>> getDeliveringMessages() {
       final Iterator<ConsumerHolder<? extends Consumer>> consumerHolderIterator;
       synchronized (this) {
-         consumerHolderIterator = redistributor == null ? consumers.iterator() : SingletonIterator.newInstance(redistributor);
+         consumerHolderIterator = consumers.iterator();
       }
 
       Map<String, List<MessageReference>> mapReturn = new HashMap<>();
@@ -2767,9 +2777,6 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
       for (ConsumerHolder holder : this.consumers) {
          holder.resetIterator();
       }
-      if (redistributor != null) {
-         redistributor.resetIterator();
-      }
    }
 
    @Override
@@ -2972,6 +2979,10 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
       // Either the iterator is empty or the consumer is busy
       int noDelivery = 0;
 
+      // track filters not matching, used to track when all consumers can't match, redistribution is then an option
+      int numNoMatch = 0;
+      int numAttempts = 0;
+
       int handled = 0;
 
       long timeout = System.nanoTime() + TimeUnit.MILLISECONDS.toNanos(DELIVERY_TIMEOUT);
@@ -2998,9 +3009,6 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
          MessageReference ref;
 
-         // filter evaluation or transformation may cause properties to be lazyDecoded, we need to reflect that
-         int existingMemoryEstimate = 0;
-
          Consumer handledconsumer = null;
 
          synchronized (this) {
@@ -3015,14 +3023,10 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
             }
 
             ConsumerHolder<? extends Consumer> holder;
-            if (redistributor == null) {
-               if (consumers.hasNext()) {
-                  holder = consumers.next();
-               } else {
-                  break;
-               }
+            if (consumers.hasNext()) {
+               holder = consumers.next();
             } else {
-               holder = redistributor;
+               break;
             }
 
             Consumer consumer = holder.consumer;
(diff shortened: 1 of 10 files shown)


gtully:
this is a relatively small change to support the scenario called out in the doc as being unsupported. The trigger is still remote demand, the difference is that when we see remote demand "and" existing consumers that don't match messages "or" we have no local consumers, we will attempt to redistribute.

clebertsuconic:
@gtully this is good. however there's one thing I'm not 100% sure is when to set hasUnMatchedPending to false?

if we add a new consumer for instance, we might have to remove the redistribution? did you add a test validating the opposite side of the issue?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3055,6 +3058,10 @@ private boolean deliver() {
                } else if (status == HandleStatus.NO_MATCH) {
                   // nothing to be done on this case, the iterators will just jump next
                   consumers.reset();
+                  numNoMatch++;
+                  if (numNoMatch == consumerCount) {
```
gemmellr, line 3062: The JIRA and code seem like it would be trying for 'no consumers matched message', but since the consumers iterator is reset just before this, aren't many/most of the attempts leading to this count incrementing going to be performed against the same consumer(s, but mostly first one), rather than checking all the consumers?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3066,7 +3073,7 @@ private boolean deliver() {
             } else if (!consumers.hasNext()) {
                // Round robin'd all
 
-               if (noDelivery == this.consumers.size()) {
+               if (noDelivery == consumerCount) {
```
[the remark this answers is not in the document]
gtully: I had the same worries because my motivation was just to avoid a second related call to size() b/c the logic of seeing noDelivery every where is what I need, it depends on consumers not changing. I can peek some more to get a better answer.
thanks for the shared concern :-)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: that is exactly what I want to achieve, I will get back to you.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2920,6 +2922,10 @@ private boolean deliver() {
       // Either the iterator is empty or the consumer is busy
       int noDelivery = 0;
 
+      // track filters not matching, used to track when all consumers can't match, redistribution is then an option
+      int numNoMatch = 0;
+      final int consumerCount = this.consumers.size();
```
michaelandrepearce, line 2927: Between here and the sync block a consumer could be added or removed and thus the count and logic within sync block would be broke. This probably should be in the sync block to avoid this

michaelandrepearce: Or just call consumers.size when needed to be safe. Rather than causing race issues

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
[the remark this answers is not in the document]
michaelandrepearce: I would leave this with call to size() its a int held number internally of the collection so should have no perf concerns. Is safer than having count being got outside the sync and then consumer added or removed and then its wrong and then theres an introduced concurrency issue

michaelandrepearce (changes requested):
comments left inline.

gtully:
thanks for all the feedback. I have sorted the potential race but in addition the logic was off. I need two counters to track in order deliver attempts that fail with no match.
however I think there is more to it, deliver treats the redistributor as a separate thing, it is not one of the consumers list, this may be ok if the redistributor is short lived, but I think it will mean that the redistributor starves local consumers b/c of the existing asumptions that it is only present when there are no local consumers!. I need to verify that. hense Draft status.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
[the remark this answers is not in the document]
gtully: leaving the call to size in place

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: I needed to add a second counter to track attempts such that I can track a bounded linear sequence of noMatch attempts that match the size of the consumers list.

gtully:
final piece, b/c the redistributor can now co exist with local consumers, it makes sense to have it in the consumers list rather than a special case of deliver.
this ensures that local consumer are not starved by the presence of a transient redistributor

gtully:
full suite of tests looks good.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1512,11 +1514,6 @@ private boolean checkConsumerDirectDeliver() {
             supports = false;
          }
       }
-      if (redistributor != null) {
```
michaelandrepearce, line 1515: Why is this being removed? New feature shouldnt be needing to remove this.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3007,23 +3009,25 @@ private boolean deliver() {
                   logger.trace("Queue " + this.getName() + " is delivering reference " + ref);
                }
 
-               existingMemoryEstimate = ref.getMessageMemoryEstimate();
```
michaelandrepearce, line 3010: Wheres this gone

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1567,7 +1565,16 @@ protected void finalize() throws Throwable {
 
    @Override
    public int getConsumerCount() {
-      return consumers.size();
+      int size = consumers.size();
+      // we don't want to count the redistributor, it is an internal transient entry in the consumer list
+      if (size > 0) {
+         synchronized (this) {
```
michaelandrepearce, line 1571: -1 this method is hit alot on metric scraping it shouldnt be taking the lock else will cause perf issue, as metrics will be stealing the delivery lock.

michaelandrepearce (changes requested):
Left comments inline

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
michaelandrepearce: Maybe make the redistributor volatile and use a static atomic field updater to update and get its value

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: that is just dead code removal, value is not used

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: fair enough... will sort

gtully: implemented with the updater. thanks for the feedback :-)

gtully:
@michaelandrepearce  do you still have concerns here?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -140,6 +139,7 @@
    private static final AtomicLongFieldUpdater<QueueImpl> dispatchStartTimeUpdater = AtomicLongFieldUpdater.newUpdater(QueueImpl.class, "dispatchStartTime");
    private static final AtomicLongFieldUpdater<QueueImpl> consumerRemovedTimestampUpdater = AtomicLongFieldUpdater.newUpdater(QueueImpl.class, "consumerRemovedTimestamp");
    private static final AtomicReferenceFieldUpdater<QueueImpl, Filter> filterUpdater = AtomicReferenceFieldUpdater.newUpdater(QueueImpl.class, Filter.class, "filter");
+   private static final AtomicReferenceFieldUpdater<QueueImpl, ConsumerHolder> redistributorUpdater = AtomicReferenceFieldUpdater.newUpdater(QueueImpl.class, ConsumerHolder.class, "redistributor");
```
gemmellr, line 142: Is this needed? Seems like there are no CAS etc operations done on it, just straight sets, and all actual uses of the redistributor field except one are done under synchronized methods/blocks, except the use in getConsumerCount where it just does a get with this updater vs using the volatile-anyway field. (The subsequent usage there is perhaps not so safe as it isnt under lock)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1567,7 +1566,14 @@ protected void finalize() throws Throwable {
...
    public int getConsumerCount() {
-      return consumers.size();
+      // we don't want to count the redistributor, it is an internal transient entry in the consumer list
+      // check for presence before checking consumers to align with use of set()
+      final boolean compensateForPresenceOfRedistributor = redistributorUpdater.get(this) != null;
+      int size = consumers.size();
+      if (size > 0 && compensateForPresenceOfRedistributor) {
+         size--;
+      }
+      return size;
```
gemmellr, line 1576: Sine this isnt done under lock and both the redistributor and consumers list could be changing independently, it seems quite racey? Seems to be used in some important places so I'd wonder if that is a problem.

Given that 'consumers' is actually a custom structure, and updates to its underlying structure are done under lock, it coudl perhaps be made to handle this internally (and e.g add a specific compensating size method)?

michaelpearce-gain:
@gtully yeah i have similar concerns to gemmellr seems we still have some possible race, tbh im not 100% if the decision to treat redistributor as a normal consumer is best idea tbh for the scope of this PR anyhow.

I wonder if the original requirement can be still me/implement without that larger change, so at least you get what you need in, without the larger changes/risks of the bigger change of starting to treat redistributor as almost a normal consumer, it almost feels like the redistributor change should be a separated thing in its own right.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: @gemmellr I like that suggestion. Thanks! I will peek at doing the compensation in size() and have consumers track the redistributor that it contains, it should be a lot cleaner.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: it was really just for consistency, but this could be internal to the custom consumers structure as you suggested below, which would be simpler

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gemmellr: Just to be clear, I was suggesting adding another method than size. I would expect size to continue to report it as being there (like it does just now) and only things that care about the distinction to call the new one.

gtully:
b/c the redistributor must now coexist with other consumers, it makes sense to treat it as one.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: I think that is all users, the redistributor is managed by the broker, to an observer it does not exist.

gtully: but I see your point, it may be better in a different method

gtully:
I have pushed an update to fix the race. having queueConsumersImpl deal with the redistributor was getting too involved b/c the queue still wants to access the redistributor for some checks,  so I am leaving getConsumerCount as the only api that knows about the presence of the redistributor and consumers.size includes it. Some uses of consumers.size have been replaced with getConsumerCount where appropriate.
There is some evidence of the redistributor having been in the consumers list at some point, b/c all uses of getConsumers use instanceOf to only deal with serverConsumer. It makes sense to me, and when a redistributor must co exist, the dispatch logic is much simpler.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: with the latest update, the distinction is now between use of consumers.size when we don't care and getConsumerCount when we only want user consumers.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: I have reverted to just accessing the volatile and doing the logic inside a sync if necessary. The updater is no more.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1573,6 +1572,17 @@ public synchronized void cancelRedistributor() throws Exception {
 
    @Override
    public int getConsumerCount() {
+      // we don't want to count the redistributor, it is an internal transient entry in the consumer list
+      if (redistributor != null) {
```
michaelpearce-gain, line 1576: With this new code, we are adding still a perf penalty, as now with volatile reads we are going to main memory not cpu cache, this affects not just here, but where redistributor is used in many code places.

michaelpearce-gain: @franz1981 can you review this?

michaelpearce-gain: i guess it is needed, else it wouldnt be visible to other thread... problem is that's a perf change as all the reads will now be from main memory for it rather than just local cpu cache

michaelpearce-gain: im shutting up, i dont think its that bad.....looks fine.

michaelpearce-gain: yes im typing as im thinking apologies....

michaelpearce-gain (approved):

michaelandrepearce (approved):

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: no worries. your concerns and views are valid, there is a cost here. I will explore a little more if it is possible to encapsulate in queueConsumersImpl to avoid the volatile read.

michaelpearce-gain: no its fine, you need it. i approved.

gtully:
any further comments, @gemmellr @franz1981 ? Thanks!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1564,6 +1562,7 @@ private void clearRedistributorFuture() {
    @Override
    public synchronized void cancelRedistributor() throws Exception {
       if (redistributor != null) {
+         consumers.remove(redistributor);
```
gemmellr, line 1565: The redistributor.consumer.stop() below this removal can throw, which would prevent the "redistributor = null" below that. Now that the getConsumerSize() return  is contingent on both the consumers map size and the redistributor nulling, perhaps this potential for them to get out of alignment should be handled now?

gtully: good point and good catch, it should be handled. thanks!

gtully: sorted, but going to tidy up the catch of that potential throw... it is not consisten

gtully: should be sorted now, thanks for the second set of eyes. turns out the log message was a cut and paste!

PR #3638, opened 2021-06-28
ARTEMIS-3369 only remove destination prefixes that are recognized

erwindon:


michaelandrepearce:
What is this fixing, is there a test case demonstrating the issue, and proving the fix?

erwindon:
a short description is available in ARTEMIS-3369, I'll first repeat it here:

> The Artemis CLI variations producer, consumer, browse and transfer, all prepare a destination using function getDestination. However, this function also removes a prefix that it was not looking for. It tests for a topic prefix (topic://) and then removes any prefix that was present.
> It should explicitly test for the 2 prefixes it cares about (topic:// and queue://) and then remove only a recognized prefix when it was present.

erwindon:
my test cases are command-line based (as this is a fix for the cli).

BEFORE:
`artemis consumer --protocol amqp --url amqp://localhost:5672 --destination topic://dest1` --> ok, server sees topic "dest1"
`artemis consumer --protocol amqp --url amqp://localhost:5672 --destination queue://dest2` --> ok, server sees queue "dest2"
`artemis consumer --protocol amqp --url amqp://localhost:5672 --destination xyz://dest3` --> FAIL, server sees queue "dest3", the prefix "xyz://" was not recognized by the CLI and was still removed.

AFTER:
`artemis consumer --protocol amqp --url amqp://localhost:5672 --destination topic://dest1` --> ok, server sees topic "dest1", same behaviour as before
`artemis consumer --protocol amqp --url amqp://localhost:5672 --destination queue://dest2` --> ok, server sees topic "dest2", same behaviour as before
`artemis consumer --protocol amqp --url amqp://localhost:5672 --destination xyz://dest3` --> ok, server sees queue "xyz://dest3", the prefix "xyz://" was not recognized by the CLI and was now not removed. this allows the free use of prefixes that are recognized by the server in the anycastPrefix or multicastPrefix as configured with an acceptor.

See also https://activemq.apache.org/components/artemis/documentation/latest/address-model.html#configuring-an-anycast-prefix for the documentation on these prefixes.

jbertram:
These tests should be in the test-suite for 2 reasons:
- Validate the fix
- Prevent regressions in the future

See `org.apache.activemq.cli.test.ArtemisTest` for other automated CLI tests. It should be pretty straight-forward to add your tests in there.

Additional thoughts:
 - The `anycastPrefix` and `multicastPrefix` are useful mainly for protocols that have no semantic distinctions for their destinations (e.g. STOMP). The CLI uses JMS which *does* have semantic distinctions for its destinations. That is why it has the built-in `queue://` and `topic://` prefixes for determining semantics. Therefore, I'm not sure it makes a lot of sense to use custom prefixes on the CLI when built-in prefixes already exist to do the same work.
 - The broker will *strip* any configured prefix. So, for example, if you had `anycastPrefix=foo://` and you used `foo://myQueue` from a client then the broker would strip `foo://` from the name and just use `myQueue` with anycast semantics. In your example, if you actually had `anycastPrefix=xyz://` the second screenshot would look exactly like the first.

erwindon:
@jbertram I'll add the regression tests

On your additional thoughts:
> The anycastPrefix and multicastPrefix are useful mainly for protocols that have no semantic distinctions for their destinations (e.g. STOMP). The CLI uses JMS which does have semantic distinctions for its destinations. That is why it has the built-in queue:// and topic:// prefixes for determining semantics. Therefore, I'm not sure it makes a lot of sense to use custom prefixes on the CLI when built-in prefixes already exist to do the same work.

I was setting up some basic tests using the `artemis producer` and `artemis consumer` to show how a client application can use the prefixes to get what it wants. In this case these are low-level amqp clients that do not (all) have the opportunity to set a capabilities value "queue" or "topic"; thus they rely on using prefixes.

> The broker will strip any configured prefix. So, for example, if you had anycastPrefix=foo:// and you used foo://myQueue from a client then the broker would strip foo:// from the name and just use myQueue with anycast semantics. In your example, if you actually had anycastPrefix=xyz:// the second screenshot would look exactly like the first.

I used "xyz://" specifically in the above tests to have a prefix that was _not_ recognized by the broker. That way I could easily verify whether the prefix reaches the server, because it would show up in the gui. In reality of course the true anycastPrefix value or multicastPrefix value would be used. Sorry for the confusion. I did not to want to use addresses/queues that really start with anything that looks like a prefix :-)

michaelandrepearce:
I we wanted to expose custom queue configs not supported natively by protocol on the destination i would be keen on repeating the solution used by core and openwire using suffixed parameters. So we have consistency

jbertram:
> I was setting up some basic tests using the `artemis producer` and `artemis consumer` to show how a client application can use the prefixes to get what it wants. In this case these are low-level amqp clients...

The problem here is that the `artemis producer` and `artemis consumer` are *not* low-level AMQP clients. They use the JMS API (which is why they can easily be switched between core and AMQP protocols). Therefore, they are not suited to demonstrating prefix functionality for low-level clients. They are just simple tools to producer and consume messages using the core or AMQP protocol via the JMS API.

I realize that your change is relatively small and potentially viewed as trivial, but the fundamental use and design of these commands is worth noting to encourage proper usage and mitigate scope creep.

erwindon:
My assumption was that the `anycast://` and `multicast://` prefixes would also work from the `artemis` commandline, bt then with server-side logic.
When that is not the case (or not intended to be so), then my PR is indeed useless.
With your comments in mind, is it better to close this?

jbertram:
@erwindon, there are no default prefixes on the broker. You'd have to configure those on the `acceptor` in `broker.xml` using `anycastPrefix=anycast://` and `multicastPrefix=multicast://`. That, combined with your change, would give the same semantics as `queue://` and `topic://` respectively from the CLI.

erwindon:
@jbertram, sorry for that, I'm already using exactly those settings for `anycastPrefix`/`multicastPrefix`. I just did not realise they were non-default as I applied them long ago, so I did not mention that information again.
I wanted to use the result of this PR to create a repeater for ARTEMIS-3153, initially as a shell script using "artemis consumer" and "artemis producer".

jbertram:
@erwindon, given our discussion I think this PR should likely be closed. WDYT?

erwindon:
everyone should be happy with a PR, so agreed
(PR https://github.com/apache/activemq-artemis/pull/3643 is nice example)

PR #3640, opened 2021-07-02
ARTEMIS-3372 - disconnect consumers on queue deletion event

gtully:


michaelandrepearce (approved):
LGTM

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/management/OpenWireDeleteQueueTest.java
```diff
@@ -0,0 +1,136 @@
...
+         if ("netty".equals(tconfig.getName())) {
+            Map<String, Object> params = tconfig.getExtraParams();
+            params.put("supportAdvisory", false);
+            params.put("suppressInternalManagementObjects", false);
+         }
+      }
+
+   }
+   @Test
+   public void testDestroyQueueClosingConsumers() throws Exception {
```
clebertsuconic, line 81: Can't you use this test for AMQP and Core as well? it looks a nice idea.

or this is something specific to OpenWire and I missed it?

gtully: not sure, this problem is specific to openwire, the method was TODO and the expectation from an openwire perspective is related to the auto queue creation semantic and the shared view of both producers/consumers. Hense only openwire needs to get a full disconnect

PR #3642, opened 2021-07-02
ARTEMIS-3374 config managed queue can be deleted by durable subscriber

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSMessageConsumerTest.java
```diff
@@ -217,6 +220,30 @@ private void doTestSelector(boolean topic) throws Exception {
       }
    }
 
+   @Test(timeout = 60000)
+   public void testDurableSubscriptionWithConfigurationManagedQueue() throws Exception {
```
michaelandrepearce, line 224: Should maybe this same test be extended and checked for openwire and core

michaelandrepearce:
Should ensure same behaviour on core and openwire.

jbertram:
@michaelandrepearce, good call. I updated the test and the fix.

michaelandrepearce (approved):
Lgtm. Nice stuff @jbertram as always

PR #3644, opened 2021-07-06
ARTEMIS-3327 Removing unecessary block operation on journal append record

clebertsuconic:


clebertsuconic:
this is simply an optimization. There are no semantic changes on this commit.

jbertram:
My reading here is that the `result` wasn't actually used for anything so there was no reason to keep it. Is that how you see it?

clebertsuconic:
@jbertram that's right

brusdev:
@clebertsuconic was the result used to wait the appendExecutor execution?

gemmellr:
Depends on the provided values of 'sync' and 'callback'. The called _newSyncAndCallbackResult(sync, callback)_ method creating the future did this:
`return (sync && callback == null) ? new SimpleFutureImpl<>() : SimpleFuture.dumb();`

If asked to sync and there was no callback, it returns a 'proper future' that could be used to actually wait for the execution, and detect failures from thrown exceptions passed via the fail method.

If not asked to sync, or a callback is given, it returns a 'dumb' future that doesnt wait at all and just returns null immediately, ignores completions/failures etc.

I dont know what the sync and callback values actually are for this situation but I guess from that they were eitehr always false or non-null. Might be worth adding a check to that effect though.

clebertsuconic:
@gemmellr / @brusdev / @jbertram I was thinking only about the artemis use case here, where we always use this method with sync=false.

I know some people is using the journal directly (who actually requested me to make the journal an independent project a few times).. and they actually will use the sync version.


I will just revert this method. thanks for the comments.

gemmellr:
Might be worth adding a comment, and test, to avoid a recurrence.

gemmellr:
Revert and test addition were done via #3647

PR #3646, opened 2021-07-07
ARTEMIS-3340 layered over ARTEMIS-2716 - activation sequence tracking to protect the journal

gtully:


gtully:
leaving as draft pending complete test run and some conversation. these are new policies so the behaviour is more restrictive as indicated in the tests. I think activationSequence captures the intent nicely but am open to suggestions for better naming.

there are some nodeManager behaviours that may need a revisit.

gtully:
added some more tests to validate activation tracking and extended 'primary' policy to allow explicitly setting the nodeID. Not sure why that is a UUID and not human readable, there is already a UUID identifier. In any event,  this allows 'peer' primary brokers to coordinate with each other on a shared identity which can support multi primary or peer mode. The activation sequence ensures that they don't step in each other as only the server with the current activation sequence will activate/go live. In the event of both crashing while replicated, any restarted server can activate.

gtully:
all the tests are good on this branch :-)

gtully:
I don't want to squash, I think all of the commits franz has are mostly isolated.

gtully:
@franz1981  @michaelandrepearce bits I would like to improve:
- have the nodeID be human readable, just a string. it is currently a UUID b/c it it typically auto generated. However to facilitate coordination among peers it needs to be preconfigured/predetermined and it makes sense to have that human readable. using NodeID ha-server-01, state in zk could then be: 
```
       /prod/activemq-artemis/ha-server-01/lock
       /prod/activemq-artemis/ha-server-01/activation-squence

in place of: 
       /prod/activemq-artemis/locks/ha-server-01/lock
       /prod/activemq-artemis/activation-squences/ha-server-01/activation-sequence

```
If the values are collocated is is trivial to check and revert an activation-sequence. I think this will help with maintainability but it may have other ramifications. It may need some consideration.

- the RepicationBackupActivation is shared between roles and does two things... Replicate and Replicate for failback. It would be good to separate out these two jobs and have two activations.  That would really simplify the peer case.

franz1981:
Not a biggie, but I see that data rotation on backup start need to (re)set activation sequence to 0 or during a fail-back the broker risk to believe to have *any* meaningful data (because sequence > 0), while it's not; it should look like an empty backup (with activation sequence == 0).

IN addition (but I can work on this as a separate PR) I would like to provide logs or some folder naming strategy to relate activation sequence with the data version, to help recovering crazy bugs (we haven't yet discovered really :P).

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationPrimaryActivation.java
```diff
@@ -0,0 +1,489 @@
...
+            }
+            try {
+               TimeUnit.MILLISECONDS.sleep(100);
+            } catch (InterruptedException ignored) {
+            }
+         }
+         while (done < System.currentTimeMillis());
+
+         if (coordinatedActivationSequence.get() == inSyncReplicaActivation) {
+            LOGGER.warnf("Timed out waiting for failback server activation with NodeID = %s: and sequence > %d: after %dms",
```
franz1981, line 362: what happen if the other primary hasn't been able to failback?

If this broker is going to restart as backup it will skip checking any activation sequence and will rotate the data (while preserving the NodeID), is it correct?
In the worst case scenario the other primary has died and there is no live around AND this broker will rotate the most up to date data, that seems a bad thing. Let's speak on monday about it: maybe makes sense to NOT skip checking the activation sequence on server restart during a failback, in short...
```java
         if (policy.isTryFailback()) {
            // we are replicating to overwrite our data, transient backup state while trying to be the primary
         } else {
            // we may be a valid insync_replica (backup) if our activation sequence is largest for a nodeId
            // verify that before removing data..
            final DistributedLock liveLockWithInSyncReplica = checkForInSyncReplica();
```
can become
```java
            // we may be a valid insync_replica (backup) if our activation sequence is largest for a nodeId
            // verify that before removing data..
            final DistributedLock liveLockWithInSyncReplica = checkForInSyncReplica();
```
because `checkForInSyncReplica` won't give up on becoming backup if the data is the most up to date.
wdyt?

franz1981 (changes requested):

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationPrimaryActivation.java
franz1981: Just a side note: on primary's start it search in the topology for a live server and if found, it restart as a (ready to failback) backup: if we `checkForInSyncReplica` it should be fine, because if there's another live server with its acceptor already opened, it means that it already own the live lock and has successfully increased activation sequence, hence `checkForInSyncReplica` should be able to fail fast thanks to the dirty read introduced recently and just make the failback process to start as expected.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationBackupActivation.java
```diff
@@ -0,0 +1,692 @@
...
+            if (!activeMQServer.isStarted()) {
+               return null;
+            }
+            LOGGER.debugf("ReplicationFailure = %s", failure);
+            boolean voluntaryFailOver = false;
+            switch (failure) {
+               case VoluntaryFailOver:
+                  voluntaryFailOver = true;
+               case NonVoluntaryFailover:
+                  final DistributedLock liveLock = tryAcquireLiveLock();
```
franz1981, line 362: tryAcquireLiveLock should return null if the activation sequence is not a match (and can perform a dirty read to fail fast): startAsLive is going to throw an exception if the activation sequence change before activating, but this would prevent it to restart as an empty backup and replicate a live again

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationBackupActivation.java
```diff
@@ -0,0 +1,692 @@
...
+      } catch (Exception e) {
+         if ((e instanceof InterruptedException || e instanceof IllegalStateException) && !activeMQServer.isStarted()) {
+            // do not log these errors if the server is being stopped.
+            return;
+         }
+         ActiveMQServerLogger.LOGGER.initializationError(e);
+      }
+   }
+
+   private DistributedLock checkForInSyncReplica() throws InterruptedException, ExecutionException, TimeoutException, UnavailableStateException {
```
franz1981, line 208:
```suggestion
   private DistributedLock tryActivate() throws InterruptedException, ExecutionException, TimeoutException, UnavailableStateException {
```

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationBackupActivation.java
```diff
@@ -0,0 +1,692 @@
...
+         LOGGER.infof("Trying to reach the majority of quorum nodes");
+         distributedManager.start();
+         LOGGER.debug("Quorum service available");
+
+         if (policy.isTryFailback()) {
+            // we are replicating to overwrite our data, transient backup state while trying to be the primary
+         } else {
+            // we may be a valid insync_replica (backup) if our activation sequence is largest for a nodeId
+            // verify that before removing data..
+            final DistributedLock liveLockWithInSyncReplica = checkForInSyncReplica();
```
franz1981, line 159:
```suggestion
            final DistributedLock liveLockWithInSyncReplica = tryActivate();
```

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationBackupActivation.java
franz1981, line 362:
```suggestion
                  // disable quorum service unavailability handling and just treat this imperatively
                  if (!stopping.compareAndSet(false, true)) {
                     return null;
                  }
                  // from now on we're meant to stop:
                  // - due to failover
                  // - due to restart/stop
                  DistributedLock liveLockWithInSyncReplica;
                  try {
                     liveLockWithInSyncReplica = tryActivate();
                  } catch (Throwable error) {
                     LOGGER.debug("Errored while attempting activation", error);
                     liveLockWithInSyncReplica = null;
                  }
                  assert stopping.get();
                  if (liveLockWithInSyncReplica != null) {
                     return liveLockWithInSyncReplica;
                  }
                  boolean restart = true;
                  if (voluntaryFailOver && isFirstFailbackAttempt()) {
                     restart = false;
                     LOGGER.error("Failed to fail-back: stopping broker based on quorum results");
                  } else {
                     ActiveMQServerLogger.LOGGER.restartingAsBackupBasedOnQuorumVoteResults();
                  }
                  // let's ignore the stopping flag here, we're already in control of it
                  asyncRestartServer(activeMQServer, restart, false);
```

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationBackupActivation.java
```diff
@@ -0,0 +1,692 @@
...
+    * <p>
+    * In case of fail-back, any event that's going to restart this broker as backup (eg quorum service unavailable
+    * or some replication failures) will cause {@code wasLive} to be {@code false}, because the HA policy set isn't
+    * a primary anymore.
+    */
+   private boolean isFirstFailbackAttempt() {
+      return wasLive && policy.isTryFailback();
+   }
+
+   private DistributedLock tryAcquireLiveLock() throws InterruptedException {
```
franz1981, line 422: This can just be removed

franz1981 (changes requested):

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationBackupActivation.java
```diff
@@ -0,0 +1,692 @@
...
+                  }
+                  boolean restart = true;
+                  if (voluntaryFailOver && isFirstFailbackAttempt()) {
+                     restart = false;
+                     LOGGER.error("Failed to fail-back: stopping broker based on quorum results");
+                  } else {
+                     ActiveMQServerLogger.LOGGER.restartingAsBackupBasedOnQuorumVoteResults();
+                  }
+                  // let's ignore the stopping flag here, we're in control of it
+                  asyncRestartServer(activeMQServer, restart, false);
```
franz1981, line 378:
```suggestion
                  // disable quorum service unavailability handling and just treat this imperatively
                  if (!stopping.compareAndSet(false, true)) {
                     return null;
                  }
                  // from now on we're meant to stop:
                  // - due to failover
                  // - due to restart/stop
                  DistributedLock liveLockWithInSyncReplica;
                  try {
                     liveLockWithInSyncReplica = tryActivate();
                  } catch (Throwable error) {
                     LOGGER.debug("Errored while attempting activation", error);
                     liveLockWithInSyncReplica = null;
                  }
                  assert stopping.get();
                  if (liveLockWithInSyncReplica != null) {
                     return liveLockWithInSyncReplica;
                  }
                  boolean restart = true;
                  if (voluntaryFailOver && isFirstFailbackAttempt()) {
                     restart = false;
                     LOGGER.error("Failed to fail-back: stopping broker based on quorum results");
                  } else {
                     ActiveMQServerLogger.LOGGER.restartingAsBackupBasedOnQuorumVoteResults();
                  }
                  // let's ignore the stopping flag here, we're already in control of it
                  asyncRestartServer(activeMQServer, restart, false);
```

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationBackupActivation.java
```diff
@@ -0,0 +1,692 @@
...
+         } catch (UnavailableStateException e) {
+            LOGGER.warnf(e, "Failed to acquire live lock %s because of unavailable quorum service: stop trying", liveID);
+            distributedManager.stop();
+            return null;
+         }
+      }
+      LOGGER.warnf("Failed to acquire live lock %s after %d tries", liveID, maxAttempts);
+      distributedManager.stop();
+      return null;
+   }
```
franz1981, line 456:
```suggestion
```

franz1981 (changes requested):

franz1981:
Going to send a PR to the @gtully branch with:

- refactored common methods on activations
- fixed tests/config/docs to deprecate useless configs

These changes are related to https://github.com/apache/activemq-artemis/pull/3646#discussion_r676187972

franz1981:
I'm going to send a new PR after squashing all commits of the @gtully's branch

franz1981:
@clebertsuconic https://github.com/apache/activemq-artemis/pull/3680 replace this PR, that can be marked as not to be merged

clebertsuconic:
this is supeceeded by Franz's PR.

clebertsuconic:
#3680 replaces this one.

PR #3649, opened 2021-07-08
ARTEMIS-3374 & ARTEMIS-3381

jbertram:
A couple of tests were explicitly written to delete the
configuration-managed queue. Since this is no longer allowed these
tests had to change.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/SecureConfigurationTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/SecureConfigurationTest.java
@@ -120,7 +120,7 @@ public class SecureConfigurationTest extends ActiveMQTestBase {
    }
 
    @Test
-   public void testSecureDurableSubscriber() throws Exception {
+   public void testCreateSecureDurableSubscriber() throws Exception {
       ConnectionFactory connectionFactory = getConnectionFactory("b", "b");
       String message = "blah";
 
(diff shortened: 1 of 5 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/SecureConfigurationTest.java
```diff
@@ -148,20 +134,6 @@ public void testSecureDurableSubscriber() throws Exception {
       } catch (JMSSecurityException j) {
          //Expected exception
       }
-
-      try {
```
clebertsuconic, line 152: why is this being removed? why the test is no longer valid?

jbertram: The explanation is in the commit message.

clebertsuconic: shouldn't you change the test to use a destination that is not managed by the configuration?

perhaps play with both options through a parameterized test and validate both options?

clebertsuconic: the test is about validating permissions on removing queues.

removeQueue is still a valid operation, isn't? on that case I think you have to change the test in  a way it still validating removeQueue over security.

clebertsuconic: @jbertram isn't the test valid anyways?

This is about creating a queue for the subscription. Meaning the topic is managed, but not the subscription. it should still throw an error or manage the creation of the queue correct.

jbertram: The address and subscription queue are both managed in the configuration (i.e. they are configured in the XML). If you remove either of these from the XML then the test will fail because the user doesn't have permission to create the subscription queue. This appears to be the main point of the test, although it tests the security on removal as well (which is no longer valid).

jbertram: FWIW, we already have tests for authorization on removing a queue.

clebertsuconic: as long as we are still testing removing topic subscriptions and security elsewhere?

jbertram: I created a new test to verify an unauthorized user cannot, in fact, delete a durable subscription queue by re-subscribing with a different selector and it failed which means that the original test was *not* testing this. I believe this PR can be merged as-is. I will send another PR to fix the insecure deletion problem.

jbertram: Actually, don't merge it yet. I'm investigating some other failures in the test-suite.

jbertram: I re-ran the test-suite and the previous failures are gone. I ran the failing tests locally and none failed (even after many repeated runs) except `org.apache.activemq.artemis.tests.integration.openwire.management.OpenWireDeleteQueueTest#testDestroyQueueClosingConsumers` which was added recently and only after after > 300 runs. It looks like it has a small race-condition, but it's not related to this commit from what I can tell.

PR #3651, opened 2021-07-13
NO-JIRA Adding Core Bridge Transfer Test

clebertsuconic:


clebertsuconic:
I cannot make these test to fail. 

Even though I made the bridge to cause duplicate if I killed both servers and restarted them.

@brusdev  perhaps you could improve the test from here?

brusdev:
@clebertsuconic thanks, I also can not get your test failing.

I have created another test to simulate what is causing duplicates using the Core Bridge:
https://github.com/brusdev/activemq-artemis/tree/fix_core_bridge_duplicates

PR #3652, opened 2021-07-14
ARTEMIS-3384 Fix bridge duplicate messages after reconnection

brusdev:


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PersistentDuplicateIDCache.java
```diff
@@ -240,6 +242,7 @@ private synchronized void addToCache(final ByteArray holder,
          }
 
          if (instantAdd) {
+            addToCacheInMemory(holder, recordID);
```
clebertsuconic, line 245: I don't get it.. wouldn't this be either in TX or not?

brusdev: yes but it isn't executed by TX if instantAdd = true

PR #3653, opened 2021-07-14
ARTEMIS-3385 management changes can be reverted by XML update

jbertram:
Durable changes made via the management API (e.g. adding
security-settings, adding address-settings, adding diverts) can be
reverted when reloading the XML at runtime.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/RedeployTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/RedeployTest.java
@@ -198,6 +198,237 @@ public class RedeployTest extends ActiveMQTestBase {
    }
 
    @Test
+   public void testRedeploySecuritySettings() throws Exception {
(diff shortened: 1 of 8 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/RedeployTest.java
```diff
@@ -197,6 +197,237 @@ public void run() {
...
+      final ReusableLatch latch = new ReusableLatch(1);
+
+      Runnable tick = new Runnable() {
+         @Override
+         public void run() {
+            latch.countDown();
+         }
+      };
+
+      embeddedActiveMQ.getActiveMQServer().getReloadManager().setTick(tick);
```
clebertsuconic, line 220: setTick(latch::countDown) would been nicer..  no need to make the change though as it's just a test. (just an optional comment)

clebertsuconic: never mind.. all the other tests are doing the same.. it's just a copy :)

PR #3656, opened 2021-07-15
ARTEMIS-3389 Duplicate Detection Fix with AMQP and Rollback

clebertsuconic:


clebertsuconic:
tests failing.. converted to draft

clebertsuconic:
I think I have to delete this PR. There are tests validating that the DuplicateCache is only persisted after commit in the case of transactions. Now I have De Ja Voux that I think Wildfly will probably have tests validating the semantic I am changing here.. Better to leave it out.

clebertsuconic:
@brusdev I am closing this PR. Ibased on my previous comment.

gemmellr:
I expect the tests are there because it makes sense for it to work the way it does. If you roll back a transaction with a send, then the send 'hasnt really had effect' yet. If you were to detect a subsequent send of that message in another transaction as a duplicate then when committed then you would end up with no message to consume despite only ever committing one send of it.

PR #3657, opened 2021-07-15
ARTEMIS-3388 URI query values decoded twice

jbertram:


artemis-commons/src/test/java/org/apache/activemq/artemis/utils/URIParserTest.java
```diff
@@ -58,9 +58,9 @@ public void testSchemaFruit() throws Throwable {
    public void testGenerateWithEncoding() throws Throwable {
       FruitParser parser = new FruitParser();
       Fruit myFruit = new Fruit("tomato&fruit");
-      myFruit.setHost("somehost&uui");
```
brusdev, line 61: can we check both `&` and `%26` ?

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/uri/BeanSupport.java
```diff
@@ -148,7 +148,8 @@ public static String getData(List<String> ignored, Object... beans) throws Excep
                            sb.append("&");
                         }
                         empty = false;
-                        sb.append(descriptor.getName()).append("=").append(encodeURI(value));
+//                        sb.append(descriptor.getName()).append("=").append(encodeURI(value));
```
michaelpearce-gain, line 151: nit: commented out code (remove)

PR #3661, opened 2021-07-20
ARTEMIS-3392 Scale down would fail if target queue's id greater than max int

gaohoward:


clebertsuconic:
nice one Howard. thanks

PR #3666, opened 2021-07-21
ARTEMIS-3397 remove queue rate metric from web console

jbertram:
This is a follow-up from ARTEMIS-2322.

The changes related to expired message are only there because
QueueFilterPredicate had a bug where the rate was correlated to expired
messages. When I fixed that I noticed that expired messages was actually
missing so I added it.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/QueueField.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/QueueField.java
@@ -31,7 +31,7 @@ public enum QueueField {
    DELIVERING_COUNT("deliveringCount"),
    MESSAGES_ADDED("messagesAdded"),
    MESSAGES_ACKED("messagesAcked"),
-   RATE("rate"),
+   MESSAGES_EXPIRED("messagesExpired"),
    ROUTING_TYPE("routingType"),
    USER("user"),
    AUTO_CREATED("autoCreated"),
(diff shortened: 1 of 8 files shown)


michaelandrepearce:
Ill be honest i dont understand the removal of this. If theres a bug fix it. But dont remove

jbertram:
@michaelandrepearce, check the commit message for 246bf083914c7acbb05a7fe0904c471331242c39 from ARTEMIS-2322. Hopefully that explains the issue.

michaelandrepearce:
This is quite a public breaking change though.

jbertram:
A few points...

1. The "rate" was never meant to be user facing. It was designed to be used exclusively for slow-consumer detection (i.e. `org.apache.activemq.artemis.core.server.impl.QueueImpl.SlowConsumerReaperRunnable`). Calling it from other places can break slow-consumer detection and result in false negatives. I made a mistake by adding `getRate()` to the `org.apache.activemq.artemis.core.server.Queue` interface. It should always have been `private`. I fixed that in this PR.
2. The "rate" was exposed originally on the web console by mistake via [ARTEMIS-1349](https://issues.apache.org/jira/browse/ARTEMIS-1349) in 2.3.0. It was likely included in `org.apache.activemq.artemis.core.management.impl.view.QueueView` because `getRate()` was on the `Queue` interface.
3. A corresponding JMX attribute was mistakenly added in 2.10.0 via [ARTEMIS-2322](https://issues.apache.org/jira/browse/ARTEMIS-2322) and then removed in 2.16.0 via [ARTEMIS-2909](https://issues.apache.org/jira/browse/ARTEMIS-2909). I've heard no complaints since then. I didn't even realize it was exposed on the web console until recently otherwise I would have removed it when I removed the JMX attribute.
4. I think it's unlikely that "rate" is widely used given it was never properly documented and exhibits strange behavior (e.g. consecutive calls to `getRate()` can return widely varying results unless the rate of production is extremely consistent over time, especially if any other user or system is also polling it).
5. Better alternatives exists as outlined in 246bf083914c7acbb05a7fe0904c471331242c39.

michaelandrepearce:
fair enough.

PR #3667, opened 2021-07-22
ARTEMIS-3367 Set verifyHost true by default

brusdev:


clebertsuconic:
@brusdev the DO-NOT-MERGE-YET tag it now a moot thing. you can convert the PR as a draft.


there is a link on the GitHub interface now:

Still in progress? Convert to draft

.project
```diff
@@ -14,4 +14,15 @@
 	<natures>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 	</natures>
+	<filteredResources>
```
clebertsuconic, line 17: why did you add this? was this on purpose?

brusdev: no, it wasn't on purpose. I'll remove it, thanks

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpFailoverEndpointDiscoveryTest.java
```diff
@@ -90,9 +93,9 @@ private JmsConnectionFactory getJmsConnectionFactory() {
       if (protocol == 0) {
          return new JmsConnectionFactory("failover:(amqp://localhost:61616)");
       } else {
-         String keystore = this.getClass().getClassLoader().getResource("client-side-keystore.jks").getFile();
-         String truststore = this.getClass().getClassLoader().getResource("client-side-truststore.jks").getFile();
-         return new JmsConnectionFactory("failover:(amqps://localhost:61616?transport.keyStoreLocation=" + keystore + "&transport.keyStorePassword=secureexample&transport.trustStoreLocation=" + truststore + "&transport.trustStorePassword=secureexample&transport.verifyHost=false)");
+         String keystore = this.getClass().getClassLoader().getResource("examples/features/standard/jmx-ssl/src/main/resources/activemq/server0/client-keystore.jks").getFile();
+         String truststore = this.getClass().getClassLoader().getResource("examples/features/standard/jmx-ssl/src/main/resources/activemq/server0/server-ca-truststore.jks").getFile();
```
jbertram, line 97: Why not reference the resources in `tests/security-resources` like the comment at the beginning of the class suggests? Are the example resources unique in some way?

brusdev: @jbertram thanks, it was due a refactoring issue of my IDE. I have just pushed a commit to fix.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/ssl/CoreClientOverOneWaySSLTest.java
```diff
@@ -82,60 +85,18 @@ public CoreClientOverOneWaySSLTest(String storeProvider, String storeType, boole
       if (suffix.equalsIgnoreCase("PKCS12")) {
          suffix = "p12";
       }
-      SERVER_SIDE_KEYSTORE = "server-side-keystore." + suffix;
-      CLIENT_SIDE_TRUSTSTORE = "client-side-truststore." + suffix;
+      SERVER_SIDE_KEYSTORE = "server-keystore." + suffix;
+      CLIENT_SIDE_TRUSTSTORE = "server-ca-truststore." + suffix;
```
gemmellr, line 89: Related to earlier comment, you then get somewhat less obvious things like this seemingly (but not actually) contradictory constant definition.

tests/security-resources/build.sh
```diff
@@ -0,0 +1,156 @@
...
+
+keytool -storetype pkcs12 -keystore other-client-keystore.p12 -storepass $STORE_PASS -alias other-client -certreq -file other-client.csr
+keytool -storetype pkcs12 -keystore client-ca-keystore.p12 -storepass $STORE_PASS -alias client-ca -gencert -rfc -infile other-client.csr -outfile other-client.crt -validity $VALIDITY -ext bc=ca:false -ext eku=cA -ext san=dns:localhost,ip:127.0.0.1
+
+keytool -storetype pkcs12 -keystore other-client-keystore.p12 -storepass $STORE_PASS -keypass $KEY_PASS -importcert -alias client-ca -file client-ca.crt -noprompt
+keytool -storetype pkcs12 -keystore other-client-keystore.p12 -storepass $STORE_PASS -keypass $KEY_PASS -importcert -alias other-client -file other-client.crt
+
+keytool -importkeystore -srckeystore other-client-keystore.p12 -destkeystore other-client-keystore.jceks -srcstoretype pkcs12 -deststoretype jceks -srcstorepass securepass -deststorepass securepass
+keytool -importkeystore -srckeystore other-client-keystore.p12 -destkeystore other-client-keystore.jks -srcstoretype pkcs12 -deststoretype jks -srcstorepass securepass -deststorepass securepass
```
gemmellr, line 132: Not really convinced 'other client'  (EDIT and now that I look, 'other server', 'unknown-client' and 'unknown-server') really needs to have a keystore and truststore of every type. Makes sense that the main set have the various different types to verify they all work (even though that has very little to do with the broker code in the end) but having them all for the various others seems like needless complexity.

examples/features/standard/jmx-ssl/readme.md
```diff
@@ -16,12 +16,54 @@ With these properties, ActiveMQ Artemis broker will be manageable remotely using
...
-* `keytool -export -keystore client-side-keystore.jks -file client-side-cert.cer -storepass secureexample`
-* `keytool -import -keystore server-side-truststore.jks -file client-side-cert.cer -storepass secureexample -keypass secureexample -noprompt`
+```shell
+#!/bin/bash
+set -e
+
+KEY_PASS=securepass
+STORE_PASS=securepass
+CA_VALIDITY=365000
+VALIDITY=36500
```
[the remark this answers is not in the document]
brusdev: I used a multiple of 365 because it is the number of **days in a year**, this was supposed to make it easier to evaluate how many years is the validity of the certificates.

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/remoting/impl/netty/NettyConnectorTest.java
```diff
@@ -56,10 +59,10 @@ public void setUp() throws Exception {
...
       params.put(TransportConstants.SSL_ENABLED_PROP_NAME, true);
       params.put(TransportConstants.SSL_PROVIDER, TransportConstants.OPENSSL_PROVIDER);
-      params.put(TransportConstants.KEYSTORE_PATH_PROP_NAME, "openssl-server-side-keystore.jks");
-      params.put(TransportConstants.KEYSTORE_PASSWORD_PROP_NAME, "secureexample");
-      params.put(TransportConstants.TRUSTSTORE_PATH_PROP_NAME,  "openssl-server-side-truststore.jks");
-      params.put(TransportConstants.TRUSTSTORE_PASSWORD_PROP_NAME, "secureexample");
+      params.put(TransportConstants.KEYSTORE_PATH_PROP_NAME, "server-keystore.jks");
+      params.put(TransportConstants.KEYSTORE_PASSWORD_PROP_NAME, "securepass");
+      params.put(TransportConstants.TRUSTSTORE_PATH_PROP_NAME,  "client-ca-truststore.jks");
+      params.put(TransportConstants.TRUSTSTORE_PASSWORD_PROP_NAME, "securepass");
```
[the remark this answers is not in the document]
brusdev: I used names for security resources that denote their content and not the purpose to avoid confusion if they were to be used in different contexts.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/ssl/CoreClientOverOneWaySSLTest.java
brusdev: I see, this could not be obvious but it should clarify what the test is doing:
The `SERVER_SIDE_KEYSTORE` points to a keystore with the server key.
The `CLIENT_SIDE_TRUSTSTORE` points to a truststore with the server ca certificate.

tests/security-resources/build.sh
brusdev: The `CoreClientOverOneWaySSLTest` and `CoreClientOverTwoWaySSLTest` tests use most of them because they are parameterized tests. ATM only 2 truststores for the `other-client` and 2 truststores for the `unknown-client` are not used but we might need them in the future if they are used in a parametric test.

examples/features/standard/jmx-ssl/readme.md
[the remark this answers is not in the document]
gemmellr: I did get that, but I think the value is just unnecessarily large. So much so that people who dont know (and dont spot there is a 0 difference at the end, which I didnt initially) are likely to believe its something else.

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/remoting/impl/netty/NettyConnectorTest.java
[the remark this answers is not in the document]
gemmellr: The problem is one of them still reads like it mainly conveys the purpose rather than the content, which makes the other one seem like it does too, leading to the combination being confusing, more so when historically the name was historically where it was being used.

I didnt particularly see need for the '-side' expansion in the old names, but together the old names were far clearer to me than the replacements. I fully expect people not copying an existing test to mess up using these, I know I likely will.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/ssl/CoreClientOverOneWaySSLTest.java
gemmellr: I would hope the comments showed that I actually do understand what it is doing. I'm the one who argued that the tests should use CAs. I just think the different naming chosen adds scope for confusion, and is likely to trip people up who are used to the previous (and I would say more typical) naming format where the file name pointed to where it is used.

tests/security-resources/build.sh
gemmellr: I would actually change the tests then, theres basically no need to run all those combinations. If the full suite wasted a little less time where easily possible (like here), it might not take so ridiculously long to run that important changes are made without running them (as clearly just happened on another PR).

Files can be added later when they are actually needed, no need to keep around unused files. Though again, I see no reason the fully-unused ones would ever actually be needed, using them implies time wasting. However, being able to easily regenerate things when needed is part of the argument to use a script to create them (Though this case is so trivial you wouldnt even need to regenerate the rest, you could import the same keys to the new store with the command being added to the script)

PR #3668, opened 2021-07-22
ARTEMIS-3398 - fix filtering in console with filter

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-3398


clebertsuconic:
I don't know how to verify this, so I will just trust you.

PR #3670, opened 2021-07-24
ARTEMIS-3401 use regex on web console for DLQ & expiry

jbertram:
Currently dead-letter and expiry addresses can only be identified on the
web console if they perfectly match a configurable string value (i.e.
"DLQ" and "ExpiryQueue" by default respectively). However, lots of users
have multiple dead-letter and expiry addresses which follow the same
naming pattern. The web console should support using a regular
expression so all of these addresses can be properly identified.

### DIFF
--- a/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/html/preferences.html
+++ b/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/html/preferences.html
@@ -39,8 +39,8 @@
 
   <div class="form-group">
     <label class="col-md-2 control-label" for="artemisDLQ">
-      The DLQ of the Broker
-      <span class="pficon pficon-info" data-toggle="tooltip" data-placement="top" title="The Dead Letter Queue of the Broker"></span>
+      Dead-letter address regex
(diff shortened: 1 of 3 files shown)


brusdev:
Could the retry button be enabled for all messages with the `Message.HDR_ORIGINAL_ADDRESS` property or for all addresses matching with the `DLQ`/`expiryAddress` value of another address getAddressSettingsAsJSON ?

jbertram:
Inspecting the messages for `Message.HDR_ORIGINAL_ADDRESS` *might* work. I'm not sure if that header is stripped when the message is actually retried so you might encounter a weird situation where messages on the original queue can be retried.

I'm not sure how using `getAddressSettingsAsJSON` would work here as there's no "reverse lookup" where you can submit a `dead-letter-address` value and get back the address names which match. I think you'd have to do some kind of brute force search by first invoking `getAddressNames` and then taking every individual address name and using `getAddressSettingsAsJSON` to see if `dead-letter-address` or `expiry-address` matched the value you're looking for. Also, there's the complication of auto-created dead-letter and expiry resources which use a prefix and/or suffix.

clebertsuconic:
I have no way to verify this. I will just trust you..

It would be great if a test was added using the framework done by @brusdev some time ago.. but it could be a separate PR.

PR #3672, opened 2021-07-26
ARTEMIS-3402 Split Brain detection should reject bad member updates

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3402

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
@@ -1706,6 +1706,12 @@ public interface ActiveMQServerLogger extends BasicLogger {
                                  "**************************************************************************************************************************************************************************************************************************************************************", format = Message.Format.MESSAGE_FORMAT)
    void possibleSplitBrain(String nodeID, String connectionPairInformation);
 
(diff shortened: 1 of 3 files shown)


franz1981:
@clebertsuconic @brusdev  that's a conservative version of the fix (still lack a test): I still don't understand
```java
         memberInput.setLive(connector);
```
code path that's going to be taken if the proposed live is `null`

clebertsuconic:
If you tested (running the test suite) LGTM.

I feel like I missed the return on the code.

thanks for the fix.

franz1981:
@clebertsuconic CI seems fine with this, although I've left the 
```
memberInput.setLive(connector);
```
and I would like to understand what's its purpose

franz1981:
This is read to be reviewed now :+1:

franz1981:
Given the comment on https://github.com/apache/activemq-artemis/pull/3672#issuecomment-887184813 I'm going to merge it after few hours: the CI is already happy with it

clebertsuconic:
I don't remember the exact conditions I have seen on DNSSwitchTest. That was to fix the condition bought by the split brain.


I would have to debug again to give you some understanding. You could add some log.debug into that and view the information there.


as I said I'm happy with this if tests are good. Make sure you run DNSSwitchTest a couple of times manually on your laptop please.

PR #1782, opened 2018-01-17
ARTEMIS-1612 Fix message redistribution for prefixed addresses

mtaylor:

PR #3673, opened 2021-07-26
ARTEMIS-3223 - ensure distribution uses the address from the message,…

gtully:
… rather than the address from the queue which may be a wildcard sub and not valid for publishng on, fix and test

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -1289,29 +1289,14 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
    public Pair<RoutingContext, Message> redistribute(final Message message,
                                                      final Queue originatingQueue,
                                                      final Transaction tx) throws Exception {
-      Bindings bindings = addressManager.getBindingsForRoutingAddress(originatingQueue.getAddress());
+      Bindings bindings = addressManager.getBindingsForRoutingAddress(message.getAddressSimpleString());
 
       if (bindings != null && bindings.allowRedistribute()) {
          // We have to copy the message and store it separately, otherwise we may lose remote bindings in case of restart before the message
          // arrived the target node
          // as described on https://issues.jboss.org/browse/JBPAPP-6130
          Message copyRedistribute = message.copy(storageManager.generateID());
-         copyRedistribute.setAddress(originatingQueue.getAddress());
-
-         if (tx != null) {
-            tx.addOperation(new TransactionOperationAbstract() {
-               @Override
-               public void afterRollback(Transaction tx) {
-                  try {
-                     //this will cause large message file to be
-                     //cleaned up
-                     // copyRedistribute.refDown();
-                  } catch (Exception e) {
-                     logger.warn("Failed to clean up message: " + copyRedistribute);
-                  }
-               }
-            });
-         }
+         copyRedistribute.setAddress(message.getAddress());
 
          RoutingContext context = new RoutingContextImpl(tx);
 
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
@@ -1755,6 +1755,10 @@ public interface ActiveMQServerLogger extends BasicLogger {
    @Message(id = 222302, value = "Failed to deal with property {0} when converting message from core to OpenWire: {1}", format = Message.Format.MESSAGE_FORMAT)
    void failedToDealWithObjectProperty(SimpleString property, String exceptionMessage);
 
+   @LogMessage(level = Logger.Level.WARN)
+   @Message(id = 222303, value = "Redistribution by {0} of messageID = {1} failed", format = Message.Format.MESSAGE_FORMAT)
+   void errorRedistributing(@Cause Throwable t, String queueName, long m);
+
    @LogMessage(level = Logger.Level.ERROR)
    @Message(id = 224000, value = "Failure in initialisation", format = Message.Format.MESSAGE_FORMAT)
    void initializationError(@Cause Throwable e);
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/Redistributor.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/Redistributor.java
@@ -178,6 +178,7 @@ public class Redistributor implements Consumer {
                      queue.deliverAsync();
                   }
                } catch (Exception e) {
+                  ActiveMQServerLogger.LOGGER.errorRedistributing(e, toManagementString(), reference.getMessageID());
                   try {
                      tx.rollback();
                   } catch (Exception e2) {
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
@@ -2166,7 +2166,12 @@ public class ServerSessionImpl implements ServerSession, FailureListener {
             }
          } */
 
-      AddressInfo art = getAddressAndRoutingType(new AddressInfo(msg.getAddressSimpleString(), routingType));
+      final AddressInfo targetFromMessage = new AddressInfo(msg.getAddressSimpleString(), routingType);
+      AddressInfo art = getAddressAndRoutingType(targetFromMessage);
+      if (art != targetFromMessage) {
+         // remove the prefix from the message, with the address model change, only non prefixed addresses exist on the broker
+         msg.setAddress(art.getName());
+      }
 
       // check the user has write access to this address.
       try {
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpBridgeClusterRedistributionTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpBridgeClusterRedistributionTest.java
@@ -136,8 +136,8 @@ public class AmqpBridgeClusterRedistributionTest extends AmqpClientTestSupport {
       bridgeNotificationsQueue = SimpleString.toSimpleString("BridgeNotifications");
       notificationsQueue = SimpleString.toSimpleString("Notifications");
 
-      setupClusterConnection("cluster0", "", MessageLoadBalancingType.ON_DEMAND, 1, true, 1, 2);
-      setupClusterConnection("cluster1", "", MessageLoadBalancingType.ON_DEMAND, 1, true, 2, 1);
+      setupClusterConnection("cluster-1->2", "", MessageLoadBalancingType.ON_DEMAND, 1, true, 1, 2);
+      setupClusterConnection("cluster-2->1", "", MessageLoadBalancingType.ON_DEMAND, 1, true, 2, 1);
 
       server0.start();
 
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/CoreClientTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/CoreClientTest.java
@@ -258,6 +258,8 @@ public class CoreClientTest extends ActiveMQTestBase {
             for (int i = 0; i < numMessages / anycastPrefixes.size(); i++) {
                ClientMessage message = consumer.receive(1000);
                assertNotNull(message);
+               // this seems to be the only assert of this non requirement
+               assertFalse(message.getAddress().contains(queuePrefix));
                message.acknowledge();
             }
             assertNull(consumer.receiveImmediate());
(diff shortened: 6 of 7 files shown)


gtully:
I am thinking there must be some tests that depend on the old behaviour, lets see. sanity checks on the ones that appear relevant don't show a culprit so I will await a full suite run to validate.

gtully:
there was one failure in redistribution with prefixes, which pointed to the cause of this problem. https://github.com/apache/activemq-artemis/pull/1782

re working the fix to update the message address once the routing is transformed seems correct. It keeps the prefix logic confined to the acceptor/session that produces the message. I notice that there is already some mangling going on for legacy clients in relation to prefixes.
Of note in the test on redistribution and prefixes, the consumer is on a "non prefixed" queue, the old behaviour would have an address with a prefix on messages on that queue, which seems odd.

Elyytscha:
if you can link the docs where its explained how to build artemis from your fork, i would be willing to test this on my local machine, we have written a docker-compose based test suite which spins up an artemis cluster and which is able to reproduce this error. so we would just need a compiled artemis.jar from your fork to test this out on our side i think.

best regards and thanks for working on this :)

gtully:
clone https://github.com/gtully/activemq-artemis/tree/ARTEMIS-3223 
set JAVA_HOME=..
mvn clean install -Pdev -DfailIfNoTests=false -Dtest=MqttClusterWildcardTest

in a few minutes you will have a distribution in:
artemis-distribution/target/apache-artemis-2.18.0-SNAPSHOT-bin

there are more instructions at: https://activemq.apache.org/components/artemis/documentation/2.0.0/hacking-guide/tests.html but I think you will be good to go.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1289,29 +1289,13 @@ public MessageReference reload(final Message message, final Queue queue, final T
...
+      Bindings bindings = addressManager.getBindingsForRoutingAddress(SimpleString.toSimpleString(message.getAddress()));
 
       if (bindings != null && bindings.allowRedistribute()) {
          // We have to copy the message and store it separately, otherwise we may lose remote bindings in case of restart before the message
          // arrived the target node
          // as described on https://issues.jboss.org/browse/JBPAPP-6130
          Message copyRedistribute = message.copy(storageManager.generateID());
-         copyRedistribute.setAddress(originatingQueue.getAddress());
-
-         if (tx != null) {
```
michaelandrepearce, line 1301: Isnt this re-introducing an old bug that left large messages hanging around when transactional.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1289,29 +1289,13 @@ public MessageReference reload(final Message message, final Queue queue, final T
    public Pair<RoutingContext, Message> redistribute(final Message message,
                                                      final Queue originatingQueue,
                                                      final Transaction tx) throws Exception {
-      Bindings bindings = addressManager.getBindingsForRoutingAddress(originatingQueue.getAddress());
+      Bindings bindings = addressManager.getBindingsForRoutingAddress(SimpleString.toSimpleString(message.getAddress()));
```
michaelandrepearce, line 1292: Why creating simple string of the address why not call - getAddressSimpleString

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1289,29 +1289,13 @@ public MessageReference reload(final Message message, final Queue queue, final T
    public Pair<RoutingContext, Message> redistribute(final Message message,
                                                      final Queue originatingQueue,
                                                      final Transaction tx) throws Exception {
-      Bindings bindings = addressManager.getBindingsForRoutingAddress(originatingQueue.getAddress());
```
michaelandrepearce, line 1292: Isnt this change breaking use case where a message is moved from its original address to another queue and address, from which it then needs to re-distribute.....

michaelandrepearce (changes requested):
see inline comments

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
michaelandrepearce: This is why for redistribution the queue is used, not the message's address.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
gtully: no. that is dead code.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
gtully: good call, will fix

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
gtully: I don't think so, redistribution (when enabled via a redistribution-delay != -1 is redistributing the message. Where it was stuck is just an artefact of where it is stuck. The message is still the message.
the wrangling of the address seems to come from a bad fix for a problem with prefixes left in messages.

gtully:
there may be some issue though, I see an intermittent failure in org.apache.activemq.artemis.tests.integration.amqp.AmqpBridgeClusterRedistributionTest#testSendMessageToBroker0GetFromBroker2 that I need to investigate. It fails most of the time.

Elyytscha:
> there may be some issue though, I see an intermittent failure in org.apache.activemq.artemis.tests.integration.amqp.AmqpBridgeClusterRedistributionTest#testSendMessageToBroker0GetFromBroker2 that I need to investigate. It fails most of the time.

i have a compiled artemis.jar with the latest commit from this branch, does it make sense to test already or should i wait until you had time to investigate and maybe push some changes?

gtully:
@Elyytscha I think go ahead and verify on your end. From what I have been able to understand so far the intermittent failure in org.apache.activemq.artemis.tests.integration.amqp.AmqpBridgeClusterRedistributionTest#testSendMessageToBroker0GetFromBroker2 seems to be timing related. thanks for the help :-)

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
michaelandrepearce: But the point is the address could be different if the message has been diverted or moved within the broker. Redistribution should use the queue its being Redistributed from not the messages set value. As Redistribution is at the queue level.

gtully: the divert case is ok, a divert copies and sets the target address on the copy. What is the move case?
The scenario where it breaks is where the queue is a wildcard queue, not a routable address. I am really reverting a change from 3years ago that looks like it was done for the wrong reason, prefixes being left on a message.
see: https://github.com/apache/activemq-artemis/commit/f09bde07dfa56e2c65ccf04adf8e158003fd673b

clebertsuconic: as far as I know, the redistribution will send to a specific queue.. not a reroute, so Divert should not be in play.

gtully: the crux of the problem here is that the queue address contains wildcards and those are not valid in for finding matching bindings or as the message address. you cannot send to a.#, you can only consume a.b from a.#

Elyytscha:
Hello,

we tested already, but i'm not sure if we did everything correct.

redistribution on the same node is now okay with mqtt and clean session false,

when you subscribe to a wildcard topic with clean session false, then disconnect, then send 2 messages on node a and node b to a topic in this wildcard range with different publisher, then reconnect the client, the messages are appearing with their original topic, not with a wildcard anymore.

but we have some weird behavior with the management console and with same steps above but connecting to node b instead of node a where subscriber was connected before, it does not work, there are generally no messages arriving. the weird error we see and why we cant check the settings in the webconsole is actually:


```
artemis1_1   | 2021-07-29 20:07:33,496 WARN  [io.hawt.system.Authenticator] Login failed due to: No LoginModules configured for activemq-management-management-management-management-management-management-management-management
artemis1_1   | 2021-07-29 20:07:34,026 WARN  [io.hawt.system.Authenticator] Login failed due to: No LoginModules configured for activemq-management-management-management-management-management-management-management-management
artemis1_1   | 2021-07-29 20:07:35,454 WARN  [io.hawt.system.Authenticator] Login failed due to: No LoginModules configured for activemq-management-management-management-management-management-management-management-management
```
with this login.config:
```
activemq {
  org.apache.activemq.artemis.spi.core.security.jaas.GuestLoginModule sufficient
      debug=true
      org.apache.activemq.jaas.guest.user="guest"
      org.apache.activemq.jaas.guest.role="guest";
};

activemq-management {
  org.apache.activemq.artemis.spi.core.security.jaas.PropertiesLoginModule sufficient
      debug=true
      org.apache.activemq.jaas.properties.user="artemis-users.properties"
      org.apache.activemq.jaas.properties.role="artemis-roles.properties";
};
```
but i can't imagine how that should be related to this pullrequest, so i think we have some errors in the setup or something has changed since activemq 2.16 which we used before.. so i think we have some configuration error, we are investigating this..

best regards

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
michaelandrepearce: move case is where you move a message using admin function. Or in case of DLQ or Expiry Queues, where message is auto moved into corresponding defined dlq or expiry queue.

michaelandrepearce: would better solution here be then to look up binding by name? not by address. As redistribution is at queue level anyhow?

Elyytscha:
Hello

we can confirm that with this pullrequest our problem is solved:

we can now do the following in an artemis cluster with mqtt with clean session false in mosquitto-client:
subscribe suscriber1 to node1, disconnect subscriber1 from node1 , publish message test1 on node1 with publisher1 and test2 on node2 with publisher2, connect subscriber1 to node2, topic arrives correctly:
```
Client subscriber1 sending CONNECT
Client subscriber1 received CONNACK
Client subscriber1 sending SUBSCRIBE (Mid: 1, Topic: /drivers/+/services, QoS: 0)
Client subscriber1 received PUBLISH (d0, q0, r0, m0, '/drivers/user-x/services', ... (5 bytes))
test1
Client subscriber1 received PUBLISH (d0, q0, r0, m0, '/drivers/user-x/services', ... (5 bytes))
test2
Client subscriber1 received SUBACK
Subscribed (mid: 1): 0
```

without this pullrequest (plain artemis 2.18) the same test looks like this:

```
Client subscriber1 sending CONNECT
Client subscriber1 received CONNACK
Client subscriber1 sending SUBSCRIBE (Mid: 1, Topic: /drivers/+/services, QoS: 0)
Client subscriber1 received PUBLISH (d0, q0, r0, m0, '/drivers/+services', ... (5 bytes))
test1
Client subscriber1 received PUBLISH (d0, q0, r0, m0, '/drivers/+/services', ... (5 bytes))
test2
Client subscriber1 received SUBACK
Subscribed (mid: 1): 0
```

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
michaelpearce-gain: i didnt see any comment on this....

michaelandrepearce (approved):

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
gtully: I don't think that will work with wildcards, b/c the queue can be bound multiple times, depending on what addresses it matches. Also a remote a/+ is a match for a local a/#, both match a message published to address a.b

gtully: I just peeked at queueImpl.move (used by dlq and expiry), it will reset the message address. so a DLQ address can be redistributed ok. Is that your concern?

gtully:
finally got to the source of the failure in org.apache.activemq.artemis.tests.integration.amqp.AmqpBridgeClusterRedistributionTest#testSendMessageToBroker0GetFromBroker2
the test sometimes uses redistribution, depending if the bridge gets setup on time. When it did the forward errored out silently... fixed that... but the root case was the need to keep the setAddress on the copy.

gtully:
rebased

PR #3676, opened 2021-07-27
ARTEMIS-1925 - allow redistribution with loadbalance type of OFF to e…

gtully:
…nsure local consumers get priority, we only optionally redistribute when messages are stuck

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/MessageRedistributionTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/MessageRedistributionTest.java
@@ -731,6 +731,40 @@ public class MessageRedistributionTest extends ClusterTestBase {
    }
 
    @Test
+   public void testRedistributionWhenRemoteConsumerIsAddedLbOffWithRedistribution() throws Exception {
+      setupCluster(MessageLoadBalancingType.OFF_WITH_REDISTRIBUTION);
+
+      startServers(0, 1, 2);
+
+      setupSessionFactory(0, isNetty());
+      setupSessionFactory(1, isNetty());
+      setupSessionFactory(2, isNetty());
+
+      createQueue(0, "queues.testaddress", "queue0", null, false);
+      createQueue(1, "queues.testaddress", "queue0", null, false);
+      createQueue(2, "queues.testaddress", "queue0", null, false);
+
+      addConsumer(0, 0, "queue0", null);
+
+      waitForBindings(0, "queues.testaddress", 1, 1, true);
+      waitForBindings(1, "queues.testaddress", 1, 0, true);
+      waitForBindings(2, "queues.testaddress", 1, 0, true);
+
+      waitForBindings(0, "queues.testaddress", 2, 0, false);
+      waitForBindings(1, "queues.testaddress", 2, 1, false);
+      waitForBindings(2, "queues.testaddress", 2, 1, false);
+
+      send(0, "queues.testaddress", 20, false, null);
+
+      removeConsumer(0);
+
+      addConsumer(1, 1, "queue0", null);
+
+      verifyReceiveAll(20, 1);
+      verifyNotReceive(1);
+   }
+
+   @Test
    public void testBackAndForth() throws Exception {
       for (int i = 0; i < 10; i++) {
          setupCluster(MessageLoadBalancingType.ON_DEMAND);
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
@@ -175,7 +175,7 @@ public class Create extends InputAbstract {
    @Option(name = "--max-hops", description = "Number of hops on the cluster configuration")
    private int maxHops = 0;
 
-   @Option(name = "--message-load-balancing", description = "Load balancing policy on cluster. [ON_DEMAND (default) | STRICT | OFF]")
+   @Option(name = "--message-load-balancing", description = "Load balancing policy on cluster. [ON_DEMAND (default) | STRICT | OFF | OFF_WITH_REDISTRIBUTION]")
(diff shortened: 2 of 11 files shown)


michaelandrepearce:
-1 This is Symantec breaking change for an existing setting and behavour, e.g. im using OFF to ensure redistribute never occurs, regardless. it should never redistribute.

michaelandrepearce (changes requested):
see comment

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/MessageRedistributionTest.java
```diff
@@ -730,6 +730,40 @@ public void testRedistributionWhenRemoteConsumerIsAdded() throws Exception {
       verifyNotReceive(1);
    }
 
+   @Test
+   public void testRedistributionWhenRemoteConsumerIsAddedLbOff() throws Exception {
+      setupCluster(MessageLoadBalancingType.OFF);
```
michaelandrepearce, line 735: This test is actually proving we're introducing a breaking change...

gtully: there are two features in play, the initial load balancing on routing, and redistribution. The fact that they are currently intertwined is a bug, they do different things, this is clear in the code and in the existing doc.

This change separates them further, and what breaks was already broken.
This is my read of what breaks:
enabling redistribution with redistribution-delay != -1 and then using LoadBalanceType.OFF to disable redistribution!.

clearly, to disable redistribution, leave  redistribution-delay at the default of -1. ie: don't configure it in the first place.

The 'broken' case, which is just wrong, can be called out in the release note.

this is a new test that captures the 'now' more clear separation of the features, they are independent and useful when combined.

gtully:
> e.g. im using OFF to ensure redistribute never occurs, regardless. it should never redistribute.
in my reading of the code and doc, this is just a broken assumption.
disable redistribution by not enabling redistribution.

redistribution is a solution to the problem of stuck messages due to early load balancing at routing time.
what is nice is that it is separate and it can have a binary behaviour.
5.x has deceaseNetworkConsumerPriority that biases towards local consumers. enabling redistribution can make such a feature binary (and better) when combined with no loadbalancing on initial routing. 

ON_DEMAND is like STRICT when there are remote consumers, if we only want to go remote when we have stuck messages, then redistribution is the nice answer.
this change makes that possible by extending what is there in a natural way.

the change in semantic, that needs to be communicated for sure, but it need not be a blocker.
the real semantics are in the code and test and git history, all the tests are fine with this change and the new test verifies it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/MessageRedistributionTest.java
michaelandrepearce: Can we simply just introduce a new setting it is far far safer and better. I see no need to have this as a release note and have users to have breaks when it can be sooo easily avoided with a different setting value

michaelandrepearce:
I think it should be a blocker. Especially when theres an alternative option thats easily to implement and avoids a break change that is simply adding an additional config option for the new semantic/behaviour

gtully:
wha do you suggest in terms of configuration? to my mind more configuration in this area is more confusion.

gtully:
as config: redistribution-delay-when-load-balance-type-off with default -1, it just does not make sense. It is mixing two things. I just don't think the breaking case is worth protecting, just don't enable redistribution-delay if you don't want redistribution.
is there more to it than that?

michaelandrepearce:
the issue is loadbalancing and redistribution was intertwined, e.g. most of the time the reason for loadbalancing was in cluster setup and in that scenario you wanted redistribution to occure, its why the other settings such as ON DEMAND also have an effect

AntonRoskvist:
The same basic thing was attempted here:
https://github.com/apache/activemq-artemis/pull/2893

I for one would really welcome this change, weather it's a separate option or the default behavior. I see a lot of excessive "forwarding" caused by ON_DEMAND when the feature I am really after is just the redistribution.

Br,
Anton

gtully:
@michaelandrepearce re: "the issue is loadbalancing and redistribution was intertwined" is exactly the problem, they should be separate and I want to fix that. Not make it more complicated, simplify. They are already separately configured but intertwined in code. This is a perfect release note . If it is a problem for some existing broken config, then there is existing config that can sort the problem.

michaelpearce-gain:
I understand you want to do something here, but i still hold strongly especially as no need to break existing setups, can start brining new config values in and mark old as deprecated, and add your new logic onto that new config.

gtully:
the "break existing setups" is actually "point out existing broken setups", and I think the existence of such configs is likely to be small. I say that b/c redistribution is enabled (non default) but won't be in play due to the load balance policy. With redistribution being configured, if it now kicks in it will be just a 'fix'.  
I agree in general, but in this specific case, new config won't make thinks better. The existing config being mixed up *is* the problem.
the problem I have is what to use for new config names, it just does not make sense to me. What do you suggest?

michaelpearce-gain:
OFF_DEMAND? It then follow the ON_DEMAND, where LB is on, and distribution is demand based.

gtully:
I really think this is a step in the wrong direction, but I will go with OFF_WITH_DISTRIBUTION, which make redistribution conditional on loadbalance type ON_DEMAND or OFF_WITH_DISTRIBUTION && redistribution-delay > 0

To my mind, load balancing happens at routing time,  redistribution on a queue with stuck messages. They should be independent. Maybe for 3.0 we can resolve this :-)

gtully:
There is a new load balance type - OFF_WITH_REDISTRIBUTION that ensures local consumers get priority, ie: no forwarding on initial routing, but the potential to redistribute if configured

gtully:
@michaelandrepearce can you close the loop on this one, thanks!

michaelandrepearce (approved):

michaelandrepearce:
@gtully should be ok to merge now.

PR #3677, opened 2021-07-27
ARTEMIS-3405 Log the end of paging mode for a destination

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3405


franz1981:
@clebertsuconic This change is going to produce log messages even if there's a forced started/stop paging

franz1981:
Got a failure on 
```
Error:  Failures: 
Error:    PagingStoreImplTest.testLogStartStopPaging:886->Assert.assertTrue:53->Assert.assertTrue:42->Assert.fail:87
```
probalby some async behaviour while paging/depaging, fixing it and converting to draft in the meantime

franz1981:
@clebertsuconic getting some test failures I cannot reproduce locally, checking

franz1981:
The CI is failing (no idea why) but If I run locally the failing test with
```
$ mvn -Dtest=PagingStoreImplTest#testLogStopPaging -U surefire:test -Ptests
```
it just work, I'm now trying on a remote CI as well

PR #3680, opened 2021-07-30
ARTEMIS-2716 + ARTEMIS-3340 Pluggable Quorum + Sequential activation tracking

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-2716

and 

https://issues.apache.org/jira/browse/ARTEMIS-3340


franz1981:
Some of the test failures are due to Zookeeper URL/port assignment (there must be something happening at random while starting testable ZK server), but nothing that prevent the feature to work: I'm now giving another round of doc/code/test check and this can be ready to be merged :+1:

clebertsuconic:
@franz1981 can you rebase it please?

franz1981:
@clebertsuconic running CI again after rebased

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/ClusterTestBase.java
```diff
@@ -85,16 +89,22 @@
...
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.ClassRule;
+import org.junit.Rule;
+import org.junit.rules.TemporaryFolder;
 
 public abstract class ClusterTestBase extends ActiveMQTestBase {
 
+   @Rule
+   public TemporaryFolder tmpFolder = new TemporaryFolder();
```
clebertsuconic, line 106: don't create a new TemporaryFolder please. ActiveMQTestBase already have one... now when we run tests we will need to deal with two temporary folders.. which is not easier for debugging.


(I will send you a PR removing this one)

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/file/FileDistributedLockTest.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.activemq.artemis.quorum.DistributedPrimitiveManager;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+
+public class FileDistributedLockTest extends DistributedLockTest {
+
+   @Rule
+   public TemporaryFolder tmpFolder = new TemporaryFolder();
```
clebertsuconic, line 34: another temporary folder? please take it out? use the one from the top.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ReplicationPrimaryActivation.java
```diff
@@ -75,7 +73,7 @@
 
    private final DistributedPrimitiveManager distributedManager;
 
-   private final AtomicBoolean stoppingServer;
+   private volatile boolean stoppingServer;
```
clebertsuconic, line 76: I didn't mean to make this change

artemis-commons/src/test/java/org/apache/activemq/artemis/utils/ThreadLeakCheckRule.java
```diff
@@ -270,6 +270,18 @@ private boolean isExpectedThread(Thread thread) {
       } else if (threadName.contains("ObjectCleanerThread")) {
          // Required since upgrade to Netty 4.1.22 maybe because https://github.com/netty/netty/commit/739e70398ccb6b11ffa97c6b5f8d55e455a2165e
          return true;
+      } else if (threadName.contains("RMI TCP")) {
+         return true;
+      } else if (threadName.contains("RMI Scheduler")) {
+         return true;
+      } else if (threadName.contains("RMI RenewClean")) {
```
clebertsuconic, line 277: Are you sure you're not hiding a thread leakage? can't you just stop the RMI Stub on this case?

artemis-quorum-api/src/main/java/org/apache/activemq/artemis/quorum/MutableLong.java
```diff
@@ -0,0 +1,51 @@
...
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.activemq.artemis.quorum;
+
+public interface MutableLong extends AutoCloseable {
```
clebertsuconic, line 21: Can you called this as MutableSequence, since it seems to be used on Sequence Management on Distributed Lock?


or if you don't want to change it.. can you add some javadoc on how this is used please? Perhaps both?

artemis-quorum-ri/src/main/java/org/apache/activemq/artemis/quorum/file/FileBasedPrimitiveManager.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.activemq.artemis.quorum.DistributedLock;
+import org.apache.activemq.artemis.quorum.DistributedPrimitiveManager;
+import org.apache.activemq.artemis.quorum.MutableLong;
+
+/**
+ * This is an implementation suitable to be used just on unit tests and it won't attempt
+ * to manage nor purge existing stale locks files. It's part of the tests life-cycle to properly
+ * set-up and tear-down the environment.
+ */
+public class FileBasedPrimitiveManager implements DistributedPrimitiveManager {
```
clebertsuconic, line 39: Could this be used for SharedStorage as well? did you make any changes into that part?

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/file/FileDistributedLockTest.java
clebertsuconic: I sent a PR removing these

artemis-commons/src/test/java/org/apache/activemq/artemis/utils/ThreadLeakCheckRule.java
franz1981: Nope, already checked in the first PR on the quorum vote

artemis-quorum-api/src/main/java/org/apache/activemq/artemis/quorum/MutableLong.java
franz1981: It's a general purpose counter that allow `long` values: on Curator is called AtomicDistributedLong, but here I allow non-atomic ops, that's why I've called it MutableLong.

artemis-quorum-ri/src/main/java/org/apache/activemq/artemis/quorum/file/FileBasedPrimitiveManager.java
franz1981: It's being used just for testing purposes and it's not meant to be used elsewhere (yet)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/ClusterTestBase.java
franz1981: thanks

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/file/FileDistributedLockTest.java
franz1981: thanks

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/quorum/ZookeeperPluggableQuorumPeerTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+      // alive as unreplicated, it has configured node id
+      assertTrue(Wait.waitFor(() -> 1L == backup.getActivationSequence().orElse(Long.MAX_VALUE).longValue()));
+
+      final String nodeID = backup.getNodeID().get();
+      Assert.assertNotNull(nodeID);
+      LOGGER.infof("NodeID: %s", nodeID);
+
+      LOGGER.info("starting peer a primary");
+      Process primaryInstance = primary.startServer(this, 0);
```
clebertsuconic, line 75: this variable is not used.

gemmellr:
The huge amount of log output generated by zookeeper bits during the tests caused the Travis CI jobs to exceed the allowed log size and abort mid run (e.g https://app.travis-ci.com/github/apache/activemq-artemis/builds/234627453 for the merge).

I worked around it by the above commit https://github.com/apache/activemq-artemis/commit/b8c779c9fa21aa45fcbdabf50fbdde168d7aabe2 redirecting the output to file, and also https://github.com/apache/activemq-artemis/commit/9baa9a4c2b6adfa543689131e0aee63e02bac729 to stop maven only outputting error details (as that then meant there wasnt output for so long it thought the build died).

I would wonder if the amount of output can be reduced though, e.g there are lots of warnings about not finding Jetty that seem like they could be removed one way or another (e.g suppressing them, or adding it). Presumably the many many connection refused etc errors could be harder.

PR #3682, opened 2021-08-04
ARTEMIS-3302 fix regression with OpenSSL

jbertram:
When using the OpenSSL provider on the broker the getPeerCertificates()
method does *not* return a X509Certificate[] so we need to convert the
Certificate[] that is returned. This code is inspired by Tomcat's
org.apache.tomcat.util.net.jsse.JSSESupport class.


gemmellr:
I was also about to merge this, tried it out and it resolved the failures. (I was waiting until after trying it out together with some CI related changes I'm working on that would have caught this problem in the PR, and separately the Java 15+ breakage and deprecation fixes from the original changes).

PR #3684, opened 2021-08-04
ARTEMIS-3302: enable some integration tests in the fast-tests profile

gemmellr:
Enables some integration tests in the PR build, to help catch issues for important areas such as TLS and security handling without needing the full 3hr+ build. I tried it out with some reverts to ensure this would have meant PRs would have caught both the JDK15+ SSL runtime breakage (throwing UOE for previously deprecated-for-removal method), and the OpenSSL related issue with the changes made to remove use of those deprecated methods. It would also have caught the karaf tests not working on JDK 11.


clebertsuconic:
LGTM

PR #3685, opened 2021-08-05
NO-JIRA: make errorprone consistently opt-in on all JDKs

gemmellr:
The use of errorprone compiler plugin is opt-in on JDK8 and JDK16 already (due to the very slow compiler forking needed there) but is on by default in JDK11 etc with no easy way to disable it (still adds 2 minutes of build overhead running on Java 11 without forking). It would be good to make it consistently opt-in, making it easier to see the core build detail locally without it present (it emits a lot of added output) and not incurring the build overhead unless actually requested.

It will still always be run in CI as that always sets -Derrorprone when running.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -947,6 +947,9 @@
           <id>jdk11to15-errorprone</id>
           <activation>
               <jdk>[11,16)</jdk>
+              <property>
+                  <name>errorprone</name>
+              </property>
           </activation>
           <build>
               <plugins>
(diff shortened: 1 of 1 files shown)


clebertsuconic:
@gemmellr would you know how to make -Pdev setting error prone by default?

I have tried doing it before and I couldn't figure it out...

In my Mind, if I compiled with -Pdev I would expect to have error prone by default.

(I'm simply asking here.. not asking you to do it of course).

gemmellr:
I dont think you can, one profile cant enable another, and there are multiple different configs needed for errorprone depending on JDK so it needs to be in separate profiles.

PR #3688, opened 2021-08-06
ARTEMIS-3414 Disable OpenWire advisory support as default created configuration on CLI

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3414

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
@@ -283,6 +283,12 @@ public class Create extends InputAbstract {
    @Option(name = "--staticCluster", description = "Cluster node connectors list, separated by comma: Example \"tcp://server:61616,tcp://server2:61616,tcp://server3:61616\"")
    String staticNode;
 
+   @Option(name = "--support-advisory", description = "If set, the generated configuration will allow advisories for the openwire protocol")
+   boolean supportAdvisory = false;
+
+   @Option(name = "--suppress-internal-management-objects", description = "If set, the generated configuration will register any advisory addresses/queues to management services for the openwire protocol")
+   boolean suppressInternalManagementObjects = false;
+
    public String[] getStaticNodes() {
       if (staticNode == null) {
          return new String[0];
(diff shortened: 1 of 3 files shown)


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -283,6 +283,12 @@
    @Option(name = "--staticCluster", description = "Cluster node connectors list, separated by comma: Example \"tcp://server:61616,tcp://server2:61616,tcp://server3:61616\"")
    String staticNode;
 
+   @Option(name = "--support-advisory", description = "Set supportAdvisory=true for the main 'artemis' acceptor on OPENWIRE protocol (Default: false)")
```
clebertsuconic, line 286: On Boolean values, the defaults are always false.

instead, please change the text as... 


description = "If you use --support-advisory, the generated configuration will allow advisories for the openwire protocol"

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -283,6 +283,12 @@
...
+   boolean supportAdvisory = false;
+
+   @Option(name = "--suppress-internal-management-objects", description = "Set suppressInternalManagementObjects=true for the main 'artemis' acceptor on OPENWIRE protocol (Default: false)")
+   boolean suppressInternalManagementObjects = false;
```
clebertsuconic, line 290: Same, here,

description = "If --supress-internal-management-objects is used the generated configuration will allow Internal Management Objects for the OpenWire protocol"



if you could find a better description for these objects it would be even better.

clebertsuconic: shouldn't these by the same attribute used?

you create a single CLI attribute such as --openwire-advisory


and use the same attribute on both places?

franz1981: By default people won't even notice, but I have left them to be separated as the acceptor params are (can be tuned singularly)

clebertsuconic:
LGTM as long as it's passing check style.. etc...

PR #3689, opened 2021-08-06
ARTEMIS-3413: clean up mirror tests of create behaviour

gemmellr:
2ee22912f07d4418e1f71bd496e016ad3123fede just added a test called "testDeleteQueueWithRemoveFalse" (just abvoe the context of this change). That in itself seems fine, but what was odd is the diff context for the change ended with context showing an existing test "testDoNotSendDelete" that appeared to want to test the same thing.

However it doesnt actually do that at all,  doing exactly the same thing as a third "testDoNotSendCreate" test, due to an argument not being used. This third test also doesnt do what its name suggests, with both of them instead verifying that a create command _is_ being sent.

This cleans things up so one test verifies the create is sent and the other that it isnt when configured, and renames them accordingly.

(Setting up a test server and monitoring that no mirroring command was sent when configured not to would probably be a better and faster test. But for another time..)


clebertsuconic:
thanks Robbie

clebertsuconic:
sorry I didn't mean to close it.. I meant to comment.. and then Merge...

PR #3691, opened 2021-08-06
ARTEMIS-3417 AutoDelete false by default on the created broker.xml

clebertsuconic:


artemis-cli/src/test/java/org/apache/activemq/cli/test/ArtemisTest.java
```diff
@@ -749,7 +751,6 @@ public void testMissingUserFileViaManagement() throws Exception {
 
       File userFile = new File(instance1.getAbsolutePath() + "/etc/artemis-users.properties");
       userFile.delete();
-      //      File roleFile = new File(instance1.getAbsolutePath() + "/etc/artemis-roles.properties");
```
clebertsuconic, line 752: non related to my change but I want to use the opportunity to remove this.

brusdev (approved):
LGTM +1

PR #3694, opened 2021-08-11
ARTEMIS-3429-3430-3446 Fixes for pluggable quorum vote activation sequence

franz1981:
It contains several fixes for https://issues.apache.org/jira/browse/ARTEMIS-3340 especially for the peer/multi-primary use case:
- patching node manager NodeID using `coordination-id` while acting as backup: after a quorum loss while backup, the "right" NodeId is lost
- fixes some JMX accesses while acting as empty backup: some JMX states relies on a `STARTED` broker, so need to set it when possible (can affect a "common" backup too)

There's some work around the NodeManager persistent state handling; given that the pluggable replication is reusing some bits of existing APIs, there's a tech debt that this PR aim to fix.

One "important" backup change is that NodeID and activation sequence's NodeManager are both set in-memory right before backup announce, without storing any of them yet (similarly to what `classic` replication does).
Both are "flushed" into disk if:
1. backup stop while in-sync: if the live die while backup is stopped, restarting the backup makes it a live candidate
2. during failover, right before starting as live (but before increasing coordinated and local activation sequence): if backup crash, if restarted, it can be a live candidate
3. on quorum loss while awaiting live to fail

I'm not sure yet about this behaviour, although as said, is the same as `classic` replication (with network pinger): maybe it worth to persist node manager state (node ID + received activation sequence) as soon as backup is announced (after received the initial sync files), because the existing replication mechanism fail to clients any replicated data activity if the replica won't ack it back to live ie as long as the activation sequence isn't incremented by the live due to running un-replicated again, the backup data is "potentially" a good one. 
TLDR **if both server crash at the same time I expect both, if restarted, to be good live candidate, while with the current implementation, given that backup node manager state isn't persisted, only the original live can become live again**.

Will leave this in a draft state until decided what's the best approach.

### DIFF
--- a/docs/user-manual/en/activation-tools.md
+++ b/docs/user-manual/en/activation-tools.md
@@ -1,0 +1,46 @@
+# Activation Sequence Tools
+
+You can use the Artemis CLI to execute activation sequence maintenance/recovery tools
+for [Pluggable Quorum Replication](ha.md).
+
+The 2 main commands are `activation list` and `activation set`, that can be used together to recover some disaster
+happened to local/coordinated activation sequences.
+
+Here is a disaster scenario built around the RI (using [Apache Zookeeper](https://zookeeper.apache.org/)
+and [Apache curator](https://curator.apache.org/)) to demonstrate the usage of such commands.
+
+## Troubleshooting Case: Zookeeper Cluster disaster
+
+A proper Zookeeper cluster should use at least 3 nodes, but what happens if all these nodes crash loosing any activation
+state information required to run an Artemis replication cluster?
+
+During the disaster ie Zookeeper nodes no longer reachable, brokers:
+
+- live ones shutdown (and if restarted by a script, should hang awaiting to connect to the Zookeeper cluster again)
+- replicas become passive, awaiting to connect to the Zookeeper cluster again
+
+Admin should:
+
+1. stop all Artemis brokers
+2. restart Zookeeper cluster
+3. search brokers with the highest local activation sequence for their `NodeID`, by running this command from the `bin`
+   folder of the broker:
+
+```bash
+$ ./artemis activation list --local
+Local activation sequence for NodeID=7debb3d1-0d4b-11ec-9704-ae9213b68ac4: 1
+```
+
+4. from the `bin` folder of the brokers with the highest local activation sequence
+
+```bash
+# assuming 1 to be the highest local activation sequence obtained at the previous step 
+# for NodeID 7debb3d1-0d4b-11ec-9704-ae9213b68ac4
+$ ./artemis activation set --remote --to 1
+Forced coordinated activation sequence for NodeID=7debb3d1-0d4b-11ec-9704-ae9213b68ac4 from 0 to 1
+```
+
+5. restart all brokers: previously live ones should be able to be live again
+
+The higher the number of Zookeeper nodes are, the less the chance are that a disaster like this requires Admin
+intervention, because it allows the Zookeeper cluster to tolerate more failures.
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/Artemis.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/Artemis.java
@@ -32,6 +32,7 @@ import org.apache.activemq.artemis.cli.commands.InvalidOptionsError;
 import org.apache.activemq.artemis.cli.commands.Kill;
 import org.apache.activemq.artemis.cli.commands.Mask;
 import org.apache.activemq.artemis.cli.commands.PrintVersion;
+import org.apache.activemq.artemis.cli.commands.activation.ActivationSequenceSet;
 import org.apache.activemq.artemis.cli.commands.check.HelpCheck;
 import org.apache.activemq.artemis.cli.commands.check.NodeCheck;
 import org.apache.activemq.artemis.cli.commands.check.QueueCheck;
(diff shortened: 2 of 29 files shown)


franz1981:
This pr still contains lot of failing tests using in memory node manager that was relying on the old node manager behaviour for the activation sequence updates ie always using write/read on actual files (!).
I need to fix them and will proceed to unmaking this as draft.

franz1981:
Re the behaviour explained above

> TLDR if both server crash at the same time I expect both, if restarted, to be good live candidate, while with the current implementation, given that backup node manager state isn't persisted, only the original live can become live again.

The reason why the `classic` replication was a bit conservative (not persisting the replica NodeID until backup was stopped or after failover was happening (after successfull quorum vote), was because a `classic` slave were not able to become live if restarted, while the pluggable quorum vote allows it, so maybe it worths to persist node ID and activation sequence a bit ASAP ie right before backup announce.

clebertsuconic:
ARTEMIS-3340 is already release.

On that case you need a new JIRA. with an affect version as 2.18.0 fix version 2.19.0

Since this is released, this becomes a new issue.

franz1981:
thanks @clebertsuconic , I'm going to create a new JIRA for this :+1: 
Now the CI is green with this: when I come back from vacation this is ready to be reviewed/merged, but I still need to decide when makes sense for a replica to persist NodeID/activation sequence info of live, let's speak about it next week ;)

franz1981:
I'm still waiting @gtully to come back and we can quickly review this before merging. CI is already good...
The only point to decide is when to persist local replica NodeID and activation sequence:

- I've decided to persist them right after the initial sync happen, because the replication process should already take care (on the replicated server) to await replica response before answering back to the client ie data delta from initial sync shouldn't diverge between live-backup while the coordinated activation sequence is still the same (will need @clebertsuconic opinion here too)
- `classic` replication instead, persist them only if backup is stopped or if backup is successfully failing-over

The implication of using these strategy is that after a simultaneous crash of both brokers and restart of just the backup:
- the former allow it to start as live because its data is in-sync (that's correct)
- the latter prevent it to start as live, because it didn't store nodeID and local activation sequence, hence it's still appear as an empty backup

To me, using the former strategy increase HA in case of crashes, assuming that the replication process is correctly syncing data (dealing correctly with delta/in-flight changes after the initial sync).

gtully (approved):
some nice tests there, looks good.

gtully:
for the cli, it needs to be very scenario and play book driven. For example:

Scenario: 
 1. zk has burned, restarted and is empty!
 2. broker fails to start with activation sequence X and coordinated sequence 0 (presumably)

Play book:
in this case, we need check each brokers activation sequence and push the largest to zk.
nodeA> artemis activation list
nodeA> X
nodeB> artemis activation list
nodeB> Y
nodeA|B> artemis activation push  


I guess list could be show.
if the return code from artemis activation list is an int (of the local activation sequence), this can be scripted nicely.

maybe: artemis activation list --remote is the value from zk.

just some thoughts :-)

another scenario: 
 live has burned
 backup has archived stale data that is now all we have.

play book:
 move the backup archive to current data
 push the current (stale) activation sequence: nodeX> artemis activation push  --force (b/c it will be stale)

franz1981:
@gtully 
The not obvious thing is that CLI commands still use Artemis loggers, hence we have lot of garbage around output (including ZK/curator/node manager logs): let me check if I can do something like this

docs/user-manual/en/activation-tools.md
```diff
@@ -0,0 +1,46 @@
+# Activation Sequence Tools
+
+You can use the Artemis CLI to execute activation sequence maintenance/recovery tools
+for [Pluggable Quorum Replication](ha.md).
+
+The 2 main commands are `activation list` and `activation set`, that can be used together to recover some disaster
+happened to local/coordinated activation sequences.
+
+Here a disaster scenario built around the RI (using [Apache Zookeeper](https://zookeeper.apache.org/)
```
gtully, line 9: here is a

docs/user-manual/en/activation-tools.md
```diff
@@ -0,0 +1,46 @@
...
+and [Apache curator](https://curator.apache.org/)) to demonstrate the usage of such commands.
+
+## Troubleshooting Case: Zookeeper Cluster disaster
+
+A proper Zookeeper cluster should use at least 3 nodes, but what happen if all these nodes crash loosing any activation
```
gtully, line 14: happens

docs/user-manual/en/activation-tools.md
```diff
@@ -0,0 +1,46 @@
...
+You can use the Artemis CLI to execute activation sequence maintenance/recovery tools
+for [Pluggable Quorum Replication](ha.md).
+
+The 2 main commands are `activation list` and `activation set`, that can be used together to recover some disaster
+happened to local/coordinated activation sequences.
+
+Here a disaster scenario built around the RI (using [Apache Zookeeper](https://zookeeper.apache.org/)
+and [Apache curator](https://curator.apache.org/)) to demonstrate the usage of such commands.
+
+## Troubleshooting Case: Zookeeper Cluster disaster
```
gtully, line 12: This is great, nice scenario and usage example. 👍

gtully (approved):

docs/user-manual/en/activation-tools.md
gtully: or 'will happen'

franz1981:
CI is green and review are fine, merging it :+1:

PR #3696, opened 2021-08-17
ARTEMIS-3420: Target Java 11

gemmellr:
https://issues.apache.org/jira/browse/ARTEMIS-3420

I have been extracting and dropping related isolated changes onto main for a while already that have just released with 2.18.0, since they improved various Java 11+ related issues without actually requiring it. These are my remaining changes to actually require and target Java 11, resolve new issues that creates, and clean up some bits that become stale as a result.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 8, 11, 17 ]
+        java: [ 11, 17 ]
 
     steps:
       - uses: actions/checkout@v2
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 8, 11, 17 ]
+        java: [ 11, 17 ]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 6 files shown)


gemmellr:
Rebased again and updated to account for other subsequent build config changes I made.

gemmellr:
Yep figured as much, that is the reason I haven't sent a <tumblweed> comment to the mail thread already ;)

michaelandrepearce:
@gemmellr  / @clebertsuconic  with the next Java LTS due this month, with the dropping of java 8, i assume we want to add support of 17, at same time, e.g. we keep support of two current java LTS versions.

gemmellr:
You may have seen me actually reference 17ea at various points in JIRAs and emails, as I do personally think it should supported quickly though I'd say its somewhat separate from requiring 11.

That said, I have actually also been ensuring the GHA build works on 17ea with these changes though (and also 2.18.0 with those I extracted before then), I just didnt include it here yet as its wasnt out yet and Temurin inst in the GHA build runner tool cache yet. I planned to add it to the actual build as soon as 17 released though (and hopefully Temurin might actually be in the tool cache by then also).

michaelandrepearce:
@gemmellr of course 17 is separate than this PR it was more a question of did you have something in the pipeline...and it sounds like you do. +100.

clebertsuconic:
yayyyy!!!!

gemmellr:
:)

PR #3697, opened 2021-08-18
ARTEMIS-3423 - create correct queue when durable subs recreated via AMQP

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-3423

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSSharedDurableConsumerTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSSharedDurableConsumerTest.java
@@ -29,7 +29,9 @@ import javax.jms.Topic;
 import java.util.Arrays;
 import java.util.Collection;
 
+import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
+import org.apache.activemq.artemis.core.server.impl.QueueImpl;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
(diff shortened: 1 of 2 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSSharedDurableConsumerTest.java
```diff
@@ -138,4 +140,46 @@ public void testSharedDurableConsumerWithArtemisClientAndAMQPClient() throws Exc
       testSharedDurableConsumer(connection, connection2);
 
    }
+
+   @Test(timeout = 30000)
+   public void testSharedDurableConsumerWithSelector() throws JMSException {
```
gemmellr, line 145:
```suggestion
   public void testSharedDurableConsumerWithSelectorChange() throws JMSException {
```
Clarify it doesnt test the selector behaviour itself, only the effect of changing it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSSharedDurableConsumerTest.java
```diff
@@ -138,4 +140,46 @@ public void testSharedDurableConsumerWithArtemisClientAndAMQPClient() throws Exc
...
+
+         Topic topic = session.createTopic(getTopicName());
+
+         MessageConsumer consumer = session.createSharedDurableConsumer(topic, "SharedConsumer", "a=b");
+         QueueImpl queue = (QueueImpl) server.getPostOffice().getBinding(qName).getBindable();
+         assertEquals(-1, queue.getMaxConsumers());
+         consumer.close();
+         MessageConsumer consumer2 = session.createSharedDurableConsumer(topic, "SharedConsumer", "a=b and b=c");
+         queue = (QueueImpl) server.getPostOffice().getBinding(qName).getBindable();
+         assertEquals(-1, queue.getMaxConsumers());
```
gemmellr, line 159: This only asserts the getMaxConsumers return didnt change, it would be better to verify the wider expected behaviour when the selector changed as well, i.e that the queue was deleted or cleared. As is this test would presumably still pass if we instead just removed the entire selector change handling it aims to test.

E.g assert the queue objects differ, or send a message before creating the new consumer and validate the message count changes.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSSharedDurableConsumerTest.java
```diff
@@ -138,4 +140,46 @@ public void testSharedDurableConsumerWithArtemisClientAndAMQPClient() throws Exc
...
+      } finally {
+         connection.close();
+      }
+   }
+
+   @Test(timeout = 30000)
+   public void testUnSharedDurableConsumerWithSelector() throws JMSException {
```
gemmellr, line 166:
```suggestion
   public void testDurableConsumerWithSelectorChange() throws JMSException {
```
Shouldn't be in the JMSSharedDurableConsumerTest class, but rather the JMSDurableConsumerTest class. Can then go with just testDurableConsumerWithSelectorChange.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSSharedDurableConsumerTest.java
```diff
@@ -138,4 +140,46 @@ public void testSharedDurableConsumerWithArtemisClientAndAMQPClient() throws Exc
...
+
+         Topic topic = session.createTopic(getTopicName());
+
+         MessageConsumer consumer = session.createDurableConsumer(topic, "SharedConsumer", "a=b", false);
+         QueueImpl queue = (QueueImpl) server.getPostOffice().getBinding(qName).getBindable();
+         assertEquals(1, queue.getMaxConsumers());
+         consumer.close();
+         MessageConsumer consumer2 = session.createDurableConsumer(topic, "SharedConsumer", "a=b and b=c", false);
+         queue = (QueueImpl) server.getPostOffice().getBinding(qName).getBindable();
+         assertEquals(1, queue.getMaxConsumers());
```
gemmellr, line 180: Same comment around better verifying behaviour as in the other test.

clebertsuconic: I have already merged it... but I will send a commit with your suggestions

gemmellr: I've put in a followup for the remaining issue

PR #3701, opened 2021-08-19
[ARTEMIS-3431]: PrintData is missing a retro-compatible method.

ehsavoie:
* Adding public static void printData(File bindingsDirectory, File messagesDirectory, File pagingDirectory, PrintStream out, boolean secret) throws Exception

Issue: https://issues.apache.org/jira/browse/ARTEMIS-3431

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/PrintData.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/PrintData.java
@@ -124,6 +124,10 @@ public class PrintData extends DBOption {
       printData(bindingsDirectory, messagesDirectory, pagingDirectory, System.out, secret, false);
    }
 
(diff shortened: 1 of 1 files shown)


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/PrintData.java
```diff
@@ -124,6 +124,10 @@ public static void printData(File bindingsDirectory, File messagesDirectory, Fil
       printData(bindingsDirectory, messagesDirectory, pagingDirectory, System.out, secret, false);
    }
 
+   public static void printData(File bindingsDirectory, File messagesDirectory, File pagingDirectory, PrintStream out, boolean secret) throws Exception {
```
clebertsuconic, line 127: I did not know this method was in use at wildly. sorry about that.. I though it was only used internally.

PR #3704, opened 2021-08-20
ARTEMIS-3436 Journal Replay operation

clebertsuconic:


artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -2891,6 +2891,21 @@ public void forceMoveNextFile() throws Exception {
       }
    }
 
+   @Override
+   public void forceBackup() throws Exception {
```
clebertsuconic, line 2895: Doing my own review here, I'm going to send a parameter to the wait latch...

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -4438,5 +4440,24 @@ private PropertiesLoginModuleConfigurator getPropertiesLoginModuleConfigurator()
...
+
+   @Override
+   public void replay(String startScan, String endScan, String address, String target, String filter) throws Exception {
+
+      SimpleDateFormat format = ReplayManager.newRetentionSimpleDateFormat();
+
+      Date startScanDate = format.parse(startScan);
+      Date endScanDate = format.parse(endScan);
+
+      System.out.println("Start date :: " + startScanDate + " formatted as " + format.format(startScanDate));
```
clebertsuconic, line 4457: forgot some system.out here.. I will remove it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -4048,6 +4053,109 @@ public void testResetUser() throws Exception {
...
+         Assert.assertNotNull(consumer.receive(5000));
+         Assert.assertNull(consumer.receiveNoWait());
+
+         if (useDate) {
+            serverControl.replay("dontexist", "dontexist", null); // just to force a move next file, and copy stuff into place
+            SimpleDateFormat format = new SimpleDateFormat("yyyyMMddHHmmss");
+            Thread.sleep(1000); // waiting a second just to have the timestamp change
+            String dateEnd = format.format(new Date());
+            Thread.sleep(1000); // waiting a second just to have the timestamp change
+            System.out.println("Date End " + dateEnd);
```
clebertsuconic, line 4094: remove this...

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -480,8 +481,14 @@ public NetworkHealthCheck getNetworkHealthCheck() {
       return networkHealthCheck;
    }
 
-   // life-cycle methods
-   // ----------------------------------------------------------------
+
+   @Override
+   public void replay(Date start, Date end, String address, String target, String filter) throws Exception {
+      if (replayManager == null) {
+         throw new NullPointerException("replayManager not set");
```
clebertsuconic, line 488: I will use the logger...

michaelandrepearce:
+1 

One quick thought could this also replace/compliment the side queue replay on subscription feature implemented last year as now wouldnt need to have a queue to the side can just use the journal

michaelandrepearce:
I was talking re retroactive address.

clebertsuconic:
@michaelandrepearce it could possibly be done I guess. although the replay operation is not meant to be multi-threaded... (as it will require reading a lot of IO files).

it would be ok as long it's not something heavily used (like too many requests per minute or something like that).

PR #3709, opened 2021-08-25
ARTEMIS-3442 Configure journal retention through cli create

clebertsuconic:


brusdev:
@clebertsuconic why is the `journal-retention-days` option mandatory?

PR #3711, opened 2021-08-29
ARTEMIS-3449 Speedup AMQP large message streaming

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3449

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
@@ -27,6 +27,7 @@ import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
 import org.apache.activemq.artemis.api.core.ICoreMessage;
 import org.apache.activemq.artemis.api.core.Message;
+import org.apache.activemq.artemis.core.buffers.impl.ChannelBufferWrapper;
 import org.apache.activemq.artemis.core.io.SequentialFile;
 import org.apache.activemq.artemis.core.message.LargeBodyReader;
 import org.apache.activemq.artemis.core.persistence.CoreMessageObjectPools;
@@ -107,7 +108,7 @@ public class AMQPLargeMessage extends AMQPMessage implements LargeServerMessage 
    private StorageManager storageManager;
 
    /** this is used to parse the initial packets from the buffer */
-   CompositeReadableBuffer parsingBuffer;
+   private CompositeReadableBuffer parsingBuffer;
 
    public AMQPLargeMessage(long id,
                            long messageFormat,
@@ -144,16 +145,6 @@ public class AMQPLargeMessage extends AMQPMessage implements LargeServerMessage 
       this.storageManager = copy.largeBody.getStorageManager();
       this.reencoded = copy.reencoded;
       setMessageID(newID);
-   }
-
-   public void openLargeMessage() throws Exception {
-      this.parsingData = new AmqpReadableBuffer(largeBody.map());
-   }
-
-   public void closeLargeMessage() throws Exception {
-      largeBody.releaseResources(false, true);
-      parsingData.freeDirectBuffer();
-      parsingData = null;
    }
 
    public void releaseEncodedBuffer() {
@@ -347,14 +338,14 @@ public class AMQPLargeMessage extends AMQPMessage implements LargeServerMessage 
    public void addBytes(ReadableBuffer data) throws Exception {
       parseLargeMessage(data);
 
-      if (data.hasArray() && data.remaining() == data.array().length) {
-         //System.out.println("Received " + data.array().length + "::" + ByteUtil.formatGroup(ByteUtil.bytesToHex(data.array()), 8, 16));
-         largeBody.addBytes(data.array());
-      } else {
-         byte[] bytes = new byte[data.remaining()];
-         data.get(bytes);
-         //System.out.println("Finishing " + bytes.length + ByteUtil.formatGroup(ByteUtil.bytesToHex(bytes), 8, 16));
-         largeBody.addBytes(bytes);
+      final int remaining = data.remaining();
+      final ByteBuf writeBuffer = PooledByteBufAllocator.DEFAULT.directBuffer(remaining, remaining);
+      try {
+         // perform copy of data
+         data.get(new NettyWritable(writeBuffer));
+         largeBody.addBytes(new ChannelBufferWrapper(writeBuffer, true, true));
+      } finally {
+         writeBuffer.release();
       }
    }
 
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
@@ -16,7 +16,6 @@
  */
 package org.apache.activemq.artemis.protocol.amqp.proton;
 
-import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -572,7 +571,7 @@ public class ProtonServerSenderContext extends ProtonInitializable implements Pr
       void deliver() {
 
          // This is discounting some bytes due to Transfer payload
-         int frameSize = protonSession.session.getConnection().getTransport().getOutboundFrameSizeLimit() - 50 - (delivery.getTag() != null ? delivery.getTag().length : 0);
+         final int frameSize = protonSession.session.getConnection().getTransport().getOutboundFrameSizeLimit() - 50 - (delivery.getTag() != null ? delivery.getTag().length : 0);
 
          DeliveryAnnotations deliveryAnnotationsToEncode;
 
@@ -584,48 +583,37 @@ public class ProtonServerSenderContext extends ProtonInitializable implements Pr
             deliveryAnnotationsToEncode = null;
          }
 
-         LargeBodyReader context = message.getLargeBodyReader();
          try {
-            context.open();
-            try {
+            final ByteBuf frameBuffer = PooledByteBufAllocator.DEFAULT.directBuffer(frameSize, frameSize);
+            final NettyReadable frameView = new NettyReadable(frameBuffer);
+            try (LargeBodyReader context = message.getLargeBodyReader()) {
+               context.open();
                context.position(position);
                long bodySize = context.getSize();
+               // materialize it so we can use its internal NIO buffer
+               frameBuffer.ensureWritable(frameSize);
 
-               ByteBuffer buf = ByteBuffer.allocate(frameSize);
+               if (position == 0 && sender.getLocalState() != EndpointState.CLOSED && position < bodySize) {
+                  if (!deliverInitialPacket(context, deliveryAnnotationsToEncode, frameBuffer)) {
+                     return;
+                  }
+               }
 
                for (; sender.getLocalState() != EndpointState.CLOSED && position < bodySize; ) {
                   if (!connection.flowControl(this::resume)) {
-                     context.close();
                      return;
                   }
-                  buf.clear();
-                  int size = 0;
+                  frameBuffer.clear();
 
-                  try {
-                     if (position == 0) {
-                        replaceInitialHeader(deliveryAnnotationsToEncode, context, WritableBuffer.ByteBufferWrapper.wrap(buf));
-                     }
-                     size = context.readInto(buf);
+                  final int readSize = context.readInto(frameBuffer.internalNioBuffer(0, frameSize));
 
-                     sender.send(new ReadableBuffer.ByteBufferReader(buf));
-                     position += size;
-                  } catch (java.nio.BufferOverflowException overflowException) {
-                     if (position == 0) {
-                        if (log.isDebugEnabled()) {
-                           log.debug("Delivery of message failed with an overFlowException, retrying again with expandable buffer");
-                        }
-                        // on the very first packet, if the initial header was replaced with a much bigger header (re-encoding)
-                        // we could recover the situation with a retry using an expandable buffer.
-                        // this is tested on org.apache.activemq.artemis.tests.integration.amqp.AmqpMessageDivertsTest
-                        size = retryInitialPacketWithExpandableBuffer(deliveryAnnotationsToEncode, context, buf);
-                     } else {
-                        // if this is not the position 0, something is going on
-                        // we just forward the exception as this is not supposed to happen
-                        throw overflowException;
-                     }
-                  }
+                  frameBuffer.writerIndex(readSize);
 
-                  if (size > 0) {
+                  sender.send(frameView);
+
+                  position += readSize;
+
+                  if (readSize > 0) {
 
                      if (position < bodySize) {
                         connection.instantFlush();
@@ -633,7 +621,7 @@ public class ProtonServerSenderContext extends ProtonInitializable implements Pr
                   }
                }
             } finally {
-               context.close();
+               frameBuffer.release();
             }
 
             if (preSettle) {
@@ -661,35 +649,59 @@ public class ProtonServerSenderContext extends ProtonInitializable implements Pr
          }
       }
 
+      private boolean deliverInitialPacket(final LargeBodyReader context,
+                                           final DeliveryAnnotations deliveryAnnotationsToEncode,
+                                           final ByteBuf frameBuffer) throws Exception {
+         assert position == 0 && context.position() == 0;
+         if (!connection.flowControl(this::resume)) {
+            return false;
+         }
+         frameBuffer.clear();
+         try {
+            replaceInitialHeader(deliveryAnnotationsToEncode, context, new NettyWritable(frameBuffer));
+         } catch (IndexOutOfBoundsException indexOutOfBoundsException) {
+            assert position == 0 : "this shouldn't happen unless replaceInitialHeader is updating position before modifying frameBuffer";
+            if (log.isDebugEnabled()) {
+               log.debug("Delivery of message failed with an overFlowException, retrying again with expandable buffer");
+            }
+            // on the very first packet, if the initial header was replaced with a much bigger header (re-encoding)
+            // we could recover the situation with a retry using an expandable buffer.
+            // this is tested on org.apache.activemq.artemis.tests.integration.amqp.AmqpMessageDivertsTest
+            sendAndFlushInitialPacket(deliveryAnnotationsToEncode, context);
+            return true;
+         }
+         final int writableBytes = frameBuffer.writableBytes();
+         if (writableBytes == 0) {
+            sender.send(new NettyReadable(frameBuffer));
+            connection.instantFlush();
+            return true;
+         }
+         final int writtenBytes = frameBuffer.writerIndex();
+         final int readSize = context.readInto(frameBuffer.internalNioBuffer(writtenBytes, writableBytes));
+         frameBuffer.writerIndex(writtenBytes + readSize);
+         sender.send(new NettyReadable(frameBuffer));
+         position += readSize;
+         connection.instantFlush();
+         return true;
+      }
+
       /**
-       * This is a retry logic when either the delivery annotations or re-encoded buffer is bigger than the frame size
-       * This will create one expandable buffer.
-       * It will then let Proton to do the framing correctly
+       * This must be used when either the delivery annotations or re-encoded buffer is bigger than the frame size.<br>
+       * This will create one expandable buffer, send and flush it.
        */
-      private int retryInitialPacketWithExpandableBuffer(DeliveryAnnotations deliveryAnnotationsToEncode,
-                                                         LargeBodyReader context,
-                                                         ByteBuffer buf) throws Exception {
-         int size;
-         buf.clear();
+      private void sendAndFlushInitialPacket(DeliveryAnnotations deliveryAnnotationsToEncode,
+                                             LargeBodyReader context) throws Exception {
          // if the buffer overflow happened during the initial position
          // this means the replaced headers are bigger then the frame size
          // on this case we do with an expandable netty buffer
-         ByteBuf nettyBuffer = PooledByteBufAllocator.DEFAULT.buffer(AMQPMessageBrokerAccessor.getRemainingBodyPosition(message) * 2);
+         final ByteBuf nettyBuffer = PooledByteBufAllocator.DEFAULT.directBuffer(AMQPMessageBrokerAccessor.getRemainingBodyPosition(message) * 2);
          try {
             replaceInitialHeader(deliveryAnnotationsToEncode, context, new NettyWritable(nettyBuffer));
-            size = context.readInto(buf);
-            position += size;
-
-            nettyBuffer.writeBytes(buf);
-
-            ByteBuffer nioBuffer = nettyBuffer.nioBuffer();
-            nioBuffer.position(nettyBuffer.writerIndex());
-            nioBuffer = (ByteBuffer) nioBuffer.flip();
-            sender.send(new ReadableBuffer.ByteBufferReader(nioBuffer));
+            sender.send(new NettyReadable(nettyBuffer));
          } finally {
             nettyBuffer.release();
+            connection.instantFlush();
          }
-         return size;
       }
 
       private int replaceInitialHeader(DeliveryAnnotations deliveryAnnotationsToEncode,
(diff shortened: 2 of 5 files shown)


franz1981:
@clebertsuconic this is addressing https://issues.apache.org/jira/browse/ARTEMIS-3026 without any exceptional code path, although I would like to log some warning, given that handling the exception case of header > frameSize can cause a severe performance degradation

franz1981:
Still need to run the whole test suite :)

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
```diff
@@ -754,23 +759,11 @@ private int writeHeaderAndAnnotations(LargeBodyReader context,
    }
 
    private void finishLargeMessage() {
-      lmUsageDown();
```
clebertsuconic, line 757: why did you remove this.. this was the whole point of the fix I recently added.

clebertsuconic:
I don't understand why you removed my recent fix on usageUp.

I don't see any equivalent calls in your call

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
franz1981: Sorry I have removed by accident while reading of main, will fix this tomorrow

franz1981: fixed!

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
```diff
@@ -587,43 +587,48 @@ void deliver() {
...
+                  final int writtenBytes;
+                  if (position == 0) {
+                     // no need to cache NettyWritable: position should be 0 just once per large message file
+                     replaceInitialHeader(deliveryAnnotationsToEncode, context, new NettyWritable(tmpFrameBuf));
+                     writtenBytes = tmpFrameBuf.writerIndex();
+                     // tested on org.apache.activemq.artemis.tests.integration.amqp.AmqpMessageDivertsTest:
+                     // tmpFrameBuf can grow over the initial capacity
+                     if (nioBuffer.remaining() < writtenBytes) {
+                        // ensure reading at least frameSize from the file
+                        tmpFrameBuf.ensureWritable(frameSize);
```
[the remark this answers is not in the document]
franz1981: This has to be done because of how netty works with internalNioBuffer: it forces Netty buffer to be materialized from EMPTY singleton to a proper one. I still need to be sure that's still valid TBH.

> This change seems like it will specifically undo that by making the buffer perpetually the wrong size and make every subsequent write for the message use an off-sized buffer leading to oddball and less efficient transfer framing behaviour.

I'm not sure about this really: this change is going to send `frameSize` (or less, depending by the remaining data in the large file) data and, if  https://issues.apache.org/jira/browse/ARTEMIS-3026 is going to happen, just a single initial bigger one which length is `encoded header + frameSize`. The subsequent frames will be `frameSize`-sized as ever.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
```diff
@@ -587,43 +587,48 @@ void deliver() {
...
+                        // refresh internal NIO buffer: the previous one is no longer valid
+                        nioBuffer = tmpFrameBuf.internalNioBuffer(0, writtenBytes + frameSize);
+                        bufPosition = nioBuffer.position();
```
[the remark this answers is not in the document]
franz1981: Let me see if I can simplify the code path here to make it simpler then, I see that looks too tricky and can be easily changed causing bugs

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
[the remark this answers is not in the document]
franz1981: mmmm now that I read my own code I see that's not doing exactly what I was expecting to do, let me turn this into draft so I can fix it and better match the existing logic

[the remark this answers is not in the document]
franz1981: Let's try again: I'm now simplified the code path, hope now it looks a bit more simpler to follow, with less checks

franz1981:
I see I didn't state clearly on https://issues.apache.org/jira/browse/ARTEMIS-3449 that the major benefit of this change is not just speeding up file reading, but it reducing the chances of OOM due to lack of native resources too (see https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/native/java/io/io_util.c#L95-L99 for reference).

In the original code, byte[] was used to perform read vs `SequentialFile`, using a `RandomAccessFile` under the hood as explained on https://github.com/apache/activemq-artemis/pull/2844#issuecomment-533606931.
Sadly, `RandomAccessFile::read` is using platform `malloc/free` (for writing/reading > [8192](https://github.com/frohoff/jdk8u-jdk/blob/da0da73ab82ed714dc5be94acd2f0d00fbdfe2e9/src/share/native/java/io/io_util.c#L58)) under the hood that can massively degrade over time due to fragmentation (if using the default system memory allocator) until been unable to allocate anything, eventually causing the broker to OOM. WIth the current approach of this PR we both save this to happen while benefitting from the speedup mentioned in https://github.com/apache/activemq-artemis/pull/2844#issuecomment-533906809 and reported here:

```
Benchmark                                  (chunkSize)    (fileName)  (fileSize)  (ioSize)  Mode  Cnt     Score     Error  Units
ChunkedFileBenchmark.readFullyFile               32768  /tmp/raf.tmp    10485760     32768  avgt    4  2153.527 ± 715.641  us/op
ChunkedFileBenchmark.readFullyFileWithNio        32768  /tmp/raf.tmp    10485760     32768  avgt    4  1775.999 ± 831.715  us/op
```
`readFullyFile` is using `byte[]` , while `readFullyFileWithNio` is using direct `ByteBuffer`

franz1981:
I've just added a commit to address the write side too, manually performing the copy of the data into a pool Netty buffer

franz1981:
I see that I'm relying (write side) on the knowledge that incoming data is using a composite readable buffer, while the whole complete version should account of Netty/NIO ones too, to save the copy to happen, looking more as

```java
   public void addBytes(ReadableBuffer data) throws Exception {
      parseLargeMessage(data);
      boolean releaseBuffer = true;
      ByteBuf writeBuffer = null;
      try {
         if (!data.hasArray()) {
            // try access some direct buffer in
            if (data instanceof ReadableBuffer.ByteBufferReader) {
               final ByteBuffer nioBuffer = data.byteBuffer();
               if (nioBuffer.isDirect()) {
                  writeBuffer = Unpooled.wrappedBuffer(nioBuffer);
                  releaseBuffer = false;
               }
            } else if (data instanceof NettyReadable) {
               final ByteBuf nettyBuffer = ((NettyReadable) data).getByteBuf();
               if (nettyBuffer.isDirect()) {
                  writeBuffer = nettyBuffer;
                  releaseBuffer = false;
               }
            }
         }
         if (writeBuffer == null) {
            // perform copy of data
            final int bytes = data.remaining();
            writeBuffer = PooledByteBufAllocator.DEFAULT.directBuffer(bytes, bytes);
            data.get(new NettyWritable(writeBuffer));
         }
         largeBody.addBytes(new ChannelBufferWrapper(writeBuffer, true, true));
      } finally {
         if (releaseBuffer) {
            if (writeBuffer != null) {
               writeBuffer.release();
            }
         }
      }
   }
```
It seems very complicate to me, so I prefer to leave it as it is, @gemmellr wdyt?

While I need your advice on the small change I've made on https://github.com/apache/activemq-artemis/pull/3711/commits/7f56ab813f8eb1e5ceb7dba4d214ee6507d9f386: is it safe?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
```diff
@@ -587,52 +586,48 @@ void deliver() {
...
-                        // this is tested on org.apache.activemq.artemis.tests.integration.amqp.AmqpMessageDivertsTest
-                        size = retryInitialPacketWithExpandableBuffer(deliveryAnnotationsToEncode, context, buf);
+                  tmpFrameBuf.clear();
+                  int headerSize = 0;
+                  int readLimit = frameSize;
+                  if (position == 0) {
+                     replaceInitialHeader(deliveryAnnotationsToEncode, context, new NettyWritable(tmpFrameBuf));
+                     headerSize = tmpFrameBuf.writerIndex();
+                     if (headerSize <= frameSize) {
+                        readLimit = frameSize - headerSize;
```
gemmellr, line 610: This still feels unnecessarily complicated. It complicates the whole remainder of the method, for a corner case thats actually quite unlikely to even arise. The variable is also confusingly named since it is absolutely not the 'header size', its only non-header stuff than can ever get it into this situation.

Same comment as yesterdays prior version, why not just send the generated first payload at this point (as the old version did when the initial buffer wasnt big enough, in retryInitialPacketWithExpandableBuffer()) and then replace the buffer with an appropriately initial-sized one, and continue the loop (assuming there is more to be sent). That would consolidate all the complexity inside this sub-part of the 'if position=0' handling, and leave the rest of the handling as the most trivial 'read data, send data' as it can be.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
```diff
@@ -146,16 +147,6 @@ private AMQPLargeMessage(final AMQPLargeMessage copy,
       setMessageID(newID);
    }
 
-   public void openLargeMessage() throws Exception {
```
gemmellr, line 149: The 'getData()' method appears to depend on this method having created "parsingData". Though I admit this method does seem ununused, getData() seems to be used a lot. I thus wonder why this seems unused

gemmellr:
> In the current PR I see that I'm relying (write side) on the knowledge that incoming data is using a composite readable buffer, while the whole complete version should account of Netty/NIO ones too, to save the copy to happen, looking more as
> 
> ```java
>    public void addBytes(ReadableBuffer data) throws Exception {
>       parseLargeMessage(data);
>       boolean releaseBuffer = true;
>       ByteBuf writeBuffer = null;
>       try {
>          if (!data.hasArray()) {
>             // try access some direct buffer in
>             if (data instanceof ReadableBuffer.ByteBufferReader) {
>                final ByteBuffer nioBuffer = data.byteBuffer();
>                if (nioBuffer.isDirect()) {
>                   writeBuffer = Unpooled.wrappedBuffer(nioBuffer);
>                   releaseBuffer = false;
>                }
>             } else if (data instanceof NettyReadable) {
>                final ByteBuf nettyBuffer = ((NettyReadable) data).getByteBuf();
>                if (nettyBuffer.isDirect()) {
>                   writeBuffer = nettyBuffer;
>                   releaseBuffer = false;
>                }
>             }
>          }
>          if (writeBuffer == null) {
>             // perform copy of data
>             final int bytes = data.remaining();
>             writeBuffer = PooledByteBufAllocator.DEFAULT.directBuffer(bytes, bytes);
>             data.get(new NettyWritable(writeBuffer));
>          }
>          largeBody.addBytes(new ChannelBufferWrapper(writeBuffer, true, true));
>       } finally {
>          if (releaseBuffer) {
>             if (writeBuffer != null) {
>                writeBuffer.release();
>             }
>          }
>       }
>    }
> ```
> 
> It seems very complicate to me, so I prefer to leave it as it is, @gemmellr wdyt?
> 

Im a bit confused by the above bits of the comment. I'm not seeing where you actually rely on the knowledge that incoming data is using a CRB (which it currently is, but that could change at any time..it didnt used to for example). In any case I would agree its complicated enough without adding all the above.

> While I need your advice on the small change I've made on [7f56ab8](https://github.com/apache/activemq-artemis/commit/7f56ab813f8eb1e5ceb7dba4d214ee6507d9f386): is it safe?

I dont really know what the original is really doing or how this updates it, so...dont know yet, need to look further.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
franz1981: >  The variable is also confusingly named since it is absolutely not the 'header size', its only non-header stuff than can ever get it into this situation.

I can rework on naming to better capture what's happening here

> why not just send the generated first payload at this point (as the old version did when the initial buffer wasnt big enough, in retryInitialPacketWithExpandableBuffer())

I think it can be done, indeed, meaning that we will skip reading the file at this round, let me rework to do it, I like it

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
franz1981: @clebertsuconic any thoughts here?

franz1981:
> Im a bit confused by the above bits of the comment. I'm not seeing where you actually rely on the knowledge that incoming data is using a CRB (which it currently is, but that could change at any time..it didnt used to for example)

Given that I know (as a programmer, not in the code :P ) that what's used is a CRB, I didn't take time to implement optimized versions for the other cases eg Netty/NIO buffer

> I dont really know what the original is really doing or how this updates it, so...dont know yet, need to look further.

thanks, let me know :+1:

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
clebertsuconic: @gemmellr / @franz1981 I remember using this at some point.. I don't know where it stopped using it. 

I didn't find any usage of it for sure.

clebertsuconic: at some point I started using getWholeBodyReader on Core to AMQP conversion (from LargeBody), and BodyReader on the AMQP Message. 

If the test suite is passing, this is fine as the test suite will cover conversion of large messages between core and AMQP.

gemmellr:
> 
> Given that I know (as a programmer, not in the code :P ) that what's used is a CRB, I didn't take time to implement optimized versions for the other cases eg Netty/NIO buffer
> 

Ok, your original comment read to me as if it were saying it wouldnt work otherwise, i.e it relied on the CRB, which I  just wasnt seeing. I think its fine the simple way it is.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
gemmellr: There is a buffer that this opens. That buffer is still being referenced in getData(), and throws but only if it isnt set, which it never will be if this isnt used. One of these situations seems like it has to be wrong.

franz1981:
I'm going to run the CI on this and if there are no other comments, I'll undraft it to move on with reviews :+1:

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
franz1981: Waiting @clebertsuconic answer to know what to remove here

gemmellr:
Think your branch needs fixing @franz1981 , there seem to be a lot of unrelated changes mixed in here now, inc a conflict.

Also, "NO-JIRA turning LargeBodyReader into AutoCloseable", is clearly not a NO-JIRA change....either its in the same JIRA or it gets its own.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
```diff
@@ -661,43 +649,67 @@ void deliver() {
...
+            connection.instantFlush();
+            return true;
+         }
+         final int writtenBytes = frameBuffer.writerIndex();
+         final int readSize = context.readInto(frameBuffer.internalNioBuffer(writtenBytes, writableBytes));
+         frameBuffer.writerIndex(writtenBytes + readSize);
+         sender.send(new NettyReadable(frameBuffer));
+         position += readSize;
+         connection.instantFlush();
+         return true;
```
gemmellr, line 685: I think perhaps this can be simplified.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
```diff
@@ -572,7 +571,7 @@ void resume() {
       void deliver() {
 
          // This is discounting some bytes due to Transfer payload
-         int frameSize = protonSession.session.getConnection().getTransport().getOutboundFrameSizeLimit() - 50 - (delivery.getTag() != null ? delivery.getTag().length : 0);
+         final int frameSize = protonSession.session.getConnection().getTransport().getOutboundFrameSizeLimit() - 50 - (delivery.getTag() != null ? delivery.getTag().length : 0);
```
gemmellr, line 574: More than that, this will be the same every time for a given message so theres no point calculating it here.

gemmellr:
I'm going to try and put together a spin on your changes Franz, I'll chat to you in the morning about it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
```diff
@@ -584,56 +583,45 @@ void deliver() {
...
+            final NettyReadable frameView = new NettyReadable(frameBuffer);
+            try (LargeBodyReader context = message.getLargeBodyReader()) {
+               context.open();
                context.position(position);
                long bodySize = context.getSize();
+               // materialize it so we can use its internal NIO buffer
+               frameBuffer.ensureWritable(frameSize);
 
-               ByteBuffer buf = ByteBuffer.allocate(frameSize);
+               if (position == 0 && sender.getLocalState() != EndpointState.CLOSED && position < bodySize) {
```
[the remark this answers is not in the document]
franz1981: > "position = 0..." is a pretty awkward way of saying 'have we finished doing the required initial packet work before'. Adding a specific boolean for this might be better

Agree but I understand why @clebertsuconic used position here, to save adding another field member (ie another state variable to preserve/maintain) to track something that position validation could track (more on this later).

> Also it doesnt necessarily actually matter if position <bodySize() 

Good point, this is not a useful check there

> e.g there are no delivery-annotations or header in the original message to skip past, or the subsequently generated ones precisely fit in the buffer with room for nothing else, or the connection was flow controlled before sending anything else

If there are no delivery annotations or header, the deliveryInitialPacket takes care filling the rest of the frame with some file content (and that should be at least one byte, but this is a missing check yet, indeed), increasing position over 0.
If flow control kick in, it would happen before doing it, hence only close or available Netty outbound buffer would resume it (with position still == 0).
Assuming that files contains at least a single byte, using position == 0 should be safe, logically, despite not very intuitive.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
franz1981: The fail fast path can be saved, but that means leaving 0 bytes available to the nii buffer and actually not reading anything on the file (that's ok).

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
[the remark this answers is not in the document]
gemmellr: > > e.g there are no delivery-annotations or header in the original message to skip past, or the subsequently generated ones precisely fit in the buffer with room for nothing else, or the connection was flow controlled before sending anything else
> 
> If there are no delivery annotations or header, the deliveryInitialPacket takes care filling the rest of the frame with some file content (and that should be at least one byte, but this is a missing check yet, indeed), increasing position over 0.

True, forgot that.

> If flow control kick in, it would happen before doing it, hence only close or available Netty outbound buffer would resume it (with position still == 0).
> Assuming that files contains at least a single byte, using position == 0 should be safe, logically, despite not very intuitive.

I meant flow control after sending the initial packet, in the loop. There are 2 routes that it is possible (but unlikely, I admit) to get into the loop without incrementing the position field, which would mean it could go through the initial stuff again and generate the initial packet again if it flow controls.

gemmellr:
> I'm going to try and put together a spin on your changes Franz, I'll chat to you in the morning about it.

I made some further changes on top of yours with ideas, at https://github.com/gemmellr/activemq-artemis/commit/1ce7a71856e4d90c995de4491d30222459cc4e23

franz1981:
@gemmellr @clebertsuconic 
Going to merge the PR of Robbie on this branch, thanks

Side note: `AMQPLargeMessage` contains some dead code eg

```java
   public void openLargeMessage() throws Exception {
      this.parsingData = new AmqpReadableBuffer(largeBody.map());
   }

   public void closeLargeMessage() throws Exception {
      largeBody.releaseResources(false, true);
      parsingData.freeDirectBuffer();
      parsingData = null;
   }
```
and `private volatile AmqpReadableBuffer parsingData;` as well, given that isn't used anymore, can be removed from
```java
   @Override
   public ReadableBuffer getData() {
      if (parsingData == null) {
         throw new RuntimeException("AMQP Large Message is not open");
      }

      return parsingData;
   }
   ```

But `getData` seems heavily used by `AMQPStandardMessage`: going to get rid of the dead code soon in a separate commit, but need some feedback to not break anything

gemmellr:
Even if it visibly isnt called, the remaining strangs seem to suggest like it should be and presumably was before, so I would leave the existing code for now and track down why/when it became unused and determine what the actual next step should be before just removing it, since the getData() usage appears to be pointing to a potentially significant issue.

gemmellr:
This wasnt finished as many of the previous comments show and so shouldnt have been merged yet.

I raised another PR to include some of the additional changes that were to be included. Alas the ship has sailed on the comment about not yet removing code that perhaps should still be used and is seemingly pointing out a problem exists.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
clebertsuconic: getData is used on conversion.. if you send an AMQP Large Message and receive it on core.. getData will be used to read the whole thing into memory.

clebertsuconic: but getData() is not opening the largeMessage as far as I know... it's just using the parsing data... let me take a look

gemmellr: It isnt opening it, true. It relied on the the 'now unused' openLargeMessage() and closeLargeMessage() methods just removed by this PR to open and close the buffer it uses. As that no longer happens, nothing will, and getData() will throw if ever actually called.

PR #3712, opened 2021-08-31
ARTEMIS-3453: exclude log4j transitive dep from zookeeper

gemmellr:
https://issues.apache.org/jira/browse/ARTEMIS-3453

Didnt test this at all yet other than seeing its effect on the main build. Seems the way to go though so raising for a PR run and any input from others.

In particular, by excluding sl4fj-log4j and log4j across the board, it may be necessary to add other dependencies to some modules (examples?) if they lack a another means of output such as another slf4j bridge.

### DIFF
--- a/artemis-quorum-ri/pom.xml
+++ b/artemis-quorum-ri/pom.xml
@@ -43,12 +43,6 @@
       <dependency>
          <groupId>org.apache.zookeeper</groupId>
          <artifactId>zookeeper</artifactId>
-         <exclusions>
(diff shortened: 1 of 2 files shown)


gtully:
+1 it is the way to go

franz1981:
Will give a shot to see what can happen here

gemmellr:
I dont see any examples depending on zookeeper directly, so I'm going to push the change as-is, can always follow up if needed.

PR #3714, opened 2021-09-01
ARTEMIS-3445 automatically clean-up abandoned MQTT subscriptions

jbertram:


artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManagerFactory.java
```diff
@@ -64,4 +74,46 @@ public ProtocolManager createProtocolManager(ActiveMQServer server,
...
       return MODULE_NAME;
    }
+
+
+   @Override
+   public void loadProtocolServices(ActiveMQServer server, List<ActiveMQComponent> services) {
+      services.add(new MQTTPeriodicTasks(server.getScheduledPool()));
+   }
+
+   public class MQTTPeriodicTasks implements ActiveMQComponent {
```
clebertsuconic, line 84: extend ActiveMQScheduledComponent instead.



provide a tick method.. like other components...

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManagerFactory.java
```diff
@@ -64,4 +74,46 @@ public ProtocolManager createProtocolManager(ActiveMQServer server,
...
+               }
+            }
+         };
+         pool.scheduleAtFixedRate(runnable, 0, 5, TimeUnit.SECONDS);
+      }
+
+      @Override
+      public void stop() throws Exception {
+         if (pool instanceof ThreadPoolExecutor) {
+            ((ThreadPoolExecutor)pool).remove(runnable);
```
clebertsuconic, line 110: the best would been to cancel on the future...

but if you extend the ScheduledComponent.. it would been easier to manage this.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManagerFactory.java
```diff
@@ -64,4 +74,46 @@ public ProtocolManager createProtocolManager(ActiveMQServer server,
...
+      public MQTTPeriodicTasks(ScheduledExecutorService pool) {
+         this.pool = pool;
+      }
+
+      @Override
+      public void start() throws Exception {
+         runnable = () -> {
+            for (Map.Entry<String, MQTTSessionState> entry : sessionStates.entrySet()) {
+               MQTTSessionState state = entry.getValue();
+               MQTTLogger.LOGGER.debug("Inspecting session state: " + state);
```
clebertsuconic, line 97: We are not using this type of logger any more for debug


just get an individual logger on the class you're at...

and just use logger.debug(....)

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManagerFactory.java
```diff
@@ -41,11 +46,16 @@
...
 
+   private long defaultMqttSessionExiryInterval = TransportConstants.DEFAULT_DEFAULT_MQTT_SESSION_EXPIRY_INTERVAL;
+
    @Override
    public ProtocolManager createProtocolManager(ActiveMQServer server,
                                                 final Map<String, Object> parameters,
                                                 List<BaseInterceptor> incomingInterceptors,
                                                 List<BaseInterceptor> outgoingInterceptors) throws Exception {
+      if (parameters.containsKey(TransportConstants.DEFAULT_MQTT_SESSION_EXPIRY_INTERVAL)) {
+         defaultMqttSessionExiryInterval = Long.parseLong((String)parameters.get(TransportConstants.DEFAULT_MQTT_SESSION_EXPIRY_INTERVAL));
```
clebertsuconic, line 57: What happens when you have multiple Protocol managers? which one will take the expiryInterval?

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManagerFactory.java
```diff
@@ -64,4 +74,46 @@ public ProtocolManager createProtocolManager(ActiveMQServer server,
...
+   @Override
+   public void loadProtocolServices(ActiveMQServer server, List<ActiveMQComponent> services) {
+      services.add(new MQTTPeriodicTasks(server.getScheduledPool()));
+   }
+
+   public class MQTTPeriodicTasks implements ActiveMQComponent {
+      ScheduledExecutorService pool;
+      Runnable runnable;
+
+      public MQTTPeriodicTasks(ScheduledExecutorService pool) {
```
clebertsuconic, line 88:
```suggestion
   public class MQTTPeriodicTasks extends ActiveMQScheduledComponent {
      public MQTTPeriodicTasks(ScheduledExecutorService scheduledExecutorService) {
         super(scheduledExecutorService, null, 5, TimeUnit.SECONDS, false);
      }

      @Override
      public void run() {
         for (Map.Entry<String, MQTTSessionState> entry : sessionStates.entrySet()) {
            MQTTSessionState state = entry.getValue();
            if (logger.isDebugEnabled()) {
               logger.debug("Inspecting session state: " + state);
            }
            if (!state.getAttached() && defaultMqttSessionExiryInterval != -1 && state.getDisconnectedTime() + (defaultMqttSessionExiryInterval * 1000) < System.currentTimeMillis()) {
               if (logger.isDebugEnabled()) {
                  logger.debug("Removing expired session state: " + state);
               }
               sessionStates.remove(entry.getKey());
            }
         }
      }
   }
```

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManagerFactory.java
clebertsuconic: the method is actually a runnable... I have sent you a PR

PR #3716, opened 2021-09-01
NO-JIRA Fix test security resources copy

brusdev:


clebertsuconic:
thanks for this.. I was having issues with Intelij and this security resources... thanks

PR #3720, opened 2021-09-02
NO-JIRA Fix broker connection AMQP SSL example

brusdev:


gtully (commented):
perfect, thanks!

gemmellr:
@gtully @brusdev

To me it looks like this change makes the example totally inconsistent with its own documentation (see readme.md and store-generation.txt, which no longer match what exists). It also seemed to be working fine for me before the change.

examples/features/broker-connection/amqp-sending-overssl/src/main/java/org/apache/activemq/artemis/jms/example/BrokerConnectionSenderSSL.java
```diff
@@ -58,7 +58,7 @@ public static void main(final String[] args) throws Exception {
       // Step 2. create a connection on server1, and receive a few messages.
       //          the sender on the broker conneciton will take care of the transfer.
       Connection connectionOnServer1 = null;
-      ConnectionFactory connectionFactoryServer1 = new JmsConnectionFactory("amqps://localhost:5771?transport.trustStoreLocation=target/server1/etc/activemq.example.truststore&transport.trustStorePassword=activemqexample");
+      ConnectionFactory connectionFactoryServer1 = new JmsConnectionFactory("amqps://localhost:5771?transport.trustStoreLocation=target/server0/etc/server-ca-truststore.jks&transport.trustStorePassword=securepass");
```
gemmellr, line 61: Why change it to using the 'server0' ca trust store file when it is connecting to server1 ? (I expect they are copies of the same file currently, but regardless...if they both exist it should use the 'matching' one)

brusdev: good catch, I'll fix it

brusdev:
@gemmellr I have just created #3722 PR to fix the inconsistency

PR #3723, opened 2021-09-03
NO-JIRA: undo addition of module-specific resources-plugin version

gemmellr:
Commit 97e84bb0d12cd2ce51571032ec3a946a97e6589d added in some explicit uses of the maven-resources plugin, and specifically configured use of version 2.6 for these. The build is otherwise already using version 3.1.0 of the resources-plugin, via pluginManagement from the apache parent pom. Its not clear the specific older version was needed, so this would remove it and keep the version config consistent and consolidated in the parents.

@brusdev ^

(Original change should probably have a JIRA that this could have gone against also)

### DIFF
--- a/tests/integration-tests/pom.xml
+++ b/tests/integration-tests/pom.xml
@@ -478,7 +478,11 @@
       <plugins>
          <plugin>
             <artifactId>maven-resources-plugin</artifactId>
-            <version>${maven-resources-plugin-version}</version>
+            <configuration>
+               <nonFilteredFileExtensions>
+                  <nonFilteredFileExtension>jks</nonFilteredFileExtension>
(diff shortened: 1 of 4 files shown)


gemmellr:
Also fixes the build warnings introduced from having multiple version definitions, e.g:
> [WARNING] 'build.plugins.plugin.(groupId:artifactId)' must be unique but found duplicate declaration of plugin org.apache.maven.plugins:maven-resources-plugin @ line 572, column 18

gemmellr:
So yeah, _now_ it should hopefully also fix the build warning :)

clebertsuconic:
thanks as usual @gemmellr

PR #3726, opened 2021-09-07
NO-JIRA Adding quorum paging & large msg BackupSyncJournalTest

franz1981:
@clebertsuconic This is just adding support to other 2 scenarios, but given that replication process is 1:1 with the `classic` replication I'm not sure yet if makes sense to have both tests in (pluggable quorum + classic); maybe I can get rid of the original one and replace with this? 
I'm opened to suggestions


clebertsuconic:
we definitely don't need both.. that test was intended to test the sync process....


I would make sure we only run only one of them.. make the pluggable one.. and get rid of the other, please?

clebertsuconic:
@franz1981 actually.. I take that back.. I thought those were long running tests... (taking minutes to run)... and they take seconds...

I don't mind if we keep the extra tests on this case.. if they were taking a lot of time I would make a case for them

clebertsuconic:
@franz1981 this LGTM... can you make it ready for review and merge it yourself?

PR #3728, opened 2021-09-08
ARTEMIS-3464 Missing ACKs on Page and Mirror

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -112,6 +112,141 @@ public AtomicInteger getScheduledCleanupCount() {
...
+                     r.run();
+                  } catch (Throwable e) {
+                     logger.warn(e.getMessage(), e);
+                  }
+               }
+            }
+         });
+         PageIterator iterator = this.iterator(true);
+         boolean hasNext = iterator.hasNext();
+         System.out.println("hasNext = " + hasNext);
```
clebertsuconic, line 190: note to myself: removing system.out

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -112,6 +112,139 @@ public AtomicInteger getScheduledCleanupCount() {
...
+      }
+
+      public Runnable getNotfound() {
+         return notfound;
+      }
+
+      PageScan(Comparable<PagedReference> scanFunction, Runnable found, Runnable notfound) {
+         this.scanFunction = scanFunction;
+         this.found = found;
+         this.notfound = notfound;
```
gemmellr, line 139: notFound (plus all the other places similarly)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPReplicaTest.java
```diff
@@ -845,23 +852,31 @@ private void consumeMessages(boolean largeMessage,
                                 int LAST_ID,
                                 int port,
                                 boolean assertNull) throws JMSException {
-      ConnectionFactory cf = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + port);
+      ConnectionFactory cf = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + port + "?jms.prefetchPolicy.all=0");
```
gemmellr, line 855: Is this needed when there is only one consumer at a time?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPReplicaTest.java
```diff
@@ -640,25 +640,32 @@ private void replicaTest(boolean largeMessage,
          server_2.startBrokerConnection(brokerConnectionName);
       }
 
+      snfreplica = server_2.locateQueue(replica.getMirrorSNF());
```
gemmellr, line 643: Why do you need to repopulate the variable, shouldnt it still be the same?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/PagedMirrorTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+   @Test
+   public void testPaged() throws Throwable {
+      String sendURI = "tcp://localhost:61616";
+      String consumeURI = "tcp://localhost:61616";
+      String secondConsumeURI = "tcp://localhost:61617";
+
+      Wait.waitFor(() -> server1.locateQueue("$ACTIVEMQ_ARTEMIS_MIRROR_other") != null);
+
+      org.apache.activemq.artemis.core.server.Queue snf1 = server2.locateQueue("$ACTIVEMQ_ARTEMIS_MIRROR_other");
```
gemmellr, line 81: May have been clearer not to name both connections "other". 

Also seems odd to wait for the server1 queue to appear and then locate the server2 queue first. Seems like you would wait on and locate each in turn.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/PagedMirrorTest.java
```diff
@@ -0,0 +1,161 @@
...
+      try (Connection consumeConnection = secondConsumeCF.createConnection()) {
+         Session consumeSession = consumeConnection.createSession(true, Session.SESSION_TRANSACTED);
+         Queue jmsQueue = consumeSession.createQueue("someQueue");
+         MessageConsumer consumer = consumeSession.createConsumer(jmsQueue);
+         consumeConnection.start();
+
+         for (int i = 0; i < NUMBER_OF_MESSAGES - 1; i++) {
+            TextMessage message = (TextMessage) consumer.receive(6000);
+            Assert.assertNotNull(message);
+            Assert.assertNotEquals(ACK_I, message.getIntProperty("i"));
```
gemmellr, line 149: Would be nicer to assert the others are as expected as opposed to just not-this-one. You could make a list and use one of the assertion/matcher libs that would compare an expected list at the end, e.g hamcrest is already in the build.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPReplicaTest.java
clebertsuconic: the broker restarted in some of the tests.. so it's better to repopulate it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
```diff
@@ -380,10 +394,7 @@ private void performAck(String nodeID, long messageID, Queue targetQueue, ACKMes
             logger.warn(e.getMessage(), e);
          }
       } else {
-         if (logger.isDebugEnabled()) {
-            logger.debug("Post ack Server " + server + " could not find messageID = " + messageID +
-                            " representing nodeID=" + nodeID);
-         }
+         performAckOnPage(nodeID, messageID, targetQueue, ackMessageOperation);
```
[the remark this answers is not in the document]
clebertsuconic: this is using the same thread that would been used on the page reading process (the scan is doing that).

in most of the cases the message should been on the queue by the depage process. in case things are out of balance, the scan will read on the paging... using the same thread that would been used on reading from page. it will then issue the pageACK.  I don't think you could get messages orphaned on this case.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPReplicaTest.java
gemmellr: But not this test? Its populated 11 lines earlier, just before the connection is stopped and started for "making sure we wouldn't duplicate the mirror". I couple maybe see adding an assertion here that doing a new locate returns the exact same queue object, i.e that the connection stop/start process didnt disturb the queue, but I dont really see reason to just reset the variable.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
```diff
@@ -476,4 +487,48 @@ public void postAcknowledge(MessageReference ref, AckReason reason) {
    public void sendMessage(Message message, RoutingContext context, List<MessageReference> refs) {
    }
 
+   // I need a supress warning annotation here
+   // because errorProne is issuing an error her, however I really intend to compare PageACK against PagedReference
+   @SuppressWarnings("ComparableType")
```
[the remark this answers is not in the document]
clebertsuconic: I thought it was nice to use Comparable.. which was what I wanted.. I didn't anticipate the warnings...

I will move to IntFunction.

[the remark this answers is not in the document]
clebertsuconic: I need the opposite of IntFunction though... I need a Function<PageReference,Integer>

I will create my own to avoid the boxing. at least until I could find an equivalent on JDK.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPReplicaTest.java
clebertsuconic: It would be my preference. especially because of cancelling deliveries in paging.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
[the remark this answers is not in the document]
gemmellr: The scan is done on the same thread that reads the pages, yes...but I dont see where there is any guarantee or coordination that the message hasnt already depaged by that thread before we eventually get round to asking it to do the scan for us, after trying and failing to ack the message on the queue itself from the other IO thread?

To me it seems like the connection thread can try to ack on the queue, fail to do so as it isnt there yet...decide to pop a scan onto the paging thread, only for that thread to successfully depage the message while we are preparing the request, and then be asked to remove it from the paging store and fail at that because its no longer there anymore by the time it runs, but is now on the queue we wont check again.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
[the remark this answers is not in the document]
gemmellr: Oops. I was thinking of what ToIntFunction does.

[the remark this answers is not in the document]
clebertsuconic: ahh.. yes.. ToIntFunction.. that's what I was looking for.. :)

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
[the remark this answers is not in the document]
clebertsuconic: hmmm... let me think about that.. I will address this shortly.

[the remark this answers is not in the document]
clebertsuconic: #3737 is addressing this

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/PagedMirrorTest.java
clebertsuconic: My intention was to make sure this message was ack. We have other tests to validate the other cases.

gemmellr: I disagree as the other tests dont appear to ack an individual message in the middle of the queue, and if thats the case then no other test actually can be verifying what I suggested to (essentially that behaviour for the rest of the queued messages is as expected ad not in some way broken).

All this checks is that the 199 messages then received arent the one it acked. It would still pass if the broker had sent completely the wrong messages, even the same message over and over, etc...its just not as robust a test as it could trivialy be.

PR #3729, opened 2021-09-09
three pool types?

pwright:


pwright:
@brusdev is this correct?

brusdev:
@pwright yes, this is correct, thanks for fixing it

PR #3730, opened 2021-09-09
ARTEMIS-3449: futher fixes and simplifications

gemmellr:
Changes that were going to be included in #3711 before it was prematurely merged. Resolves some issues as discussed and makes some simplifications to both the changes that were made and other existing code.


franz1981 (approved):
LGTM as said, it looks more clear to me :+1:

PR #3737, opened 2021-09-10
ARTEMIS-3464 Improving ACK reliability on Paging and code improvements

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -153,7 +157,7 @@ public void performScanAck() {
       // we should only have a max of 2 scheduled tasks
       // one that's might still be currently running, and another one lined up
       // no need for more than that
-      if (scheduledScanCount.incrementAndGet() < 2) {
+      if (scheduledScanCount.incrementAndGet() <= 2) {
```
gemmellr, line 160: This was obviously quite broken before the change, but even afterwards it still seems brittle and could leave scans hanging around to get executed until another request comes along (assuming it does). 

Its still racey because the count is only decremented by the worker thread after the the scans are actually performed, but it prepares the list of scans to operate on at the start before doing any, whilst these increment-then-decrement adjustments by the performScanAck requester threads can overlap with the whole process. This could mean that the tasks to operate on by the last executing task have already been prepared, but no task might get scheduled here by a new requester when it should have, as the number of requesters incrementing the value means that it seems there is a scheduled job that will handle it when there isnt (and they then subsequently decrement the value accordingly).

I think the worker needs to update the value before it prepares its local scan list to operate on, so this counter speaks only to future scheduled scan work (it could probably be a boolean at that point) or the 'add + perform' steps collapsed into one that isnt susceptible to this issue.

clebertsuconic: The idea is to have at least 2 scheduled threads...
one that might be running, one that's about to be executed..and I don't see that logic broken here.

Before with < 2, I would have only up to most 1 thread scheduled. 


If I was using a boolean then I agree I would have to get the count before. I think this is safe here.

gemmellr: It isnt safe because it fails to ensure that there are 2 scheduled, since the incrementing and decrementing here isnt suitably coordinated with the worker threads task preparation and decrementing after the work is done. Increments here can cause subsequent requesters not to schedule anything, if they arrive before the effect of those increments is removed, but after the effect of the worker thread task prep or decrement take place. As a result, they may not schedule when actually they needed to.

E.g: If more than 2 threads are trying to add scans and call performScanAck, while simultaneously a prior executing scan is just about to mark it that it already completed its work by decrementing and the other scheduled scan can then begin being executed, a 3rd+ requester thread at that point may not schedule anything here due to a prior 2 incrementing the count (whilst it was already 2+ from the prior about-to-complete and scheduled tasks, making it 4+). However the previously-executing scan may just complete (knocking 1 from the probably-still-4+ scheduledScanCount value, thus not taking it below the point another will yet be scheduled) and causing the other already scheduled task to begin and read all the tasks at the time...but perhaps occurring before it might including those of a 3rd+ new requester, and before the 1st+2nd new requesters then eventually decrement the count as they didnt schedule anything. Nothing prevents those requester decrements only happening after the scan list was already read by the second executing worker task, but before the 3rd+ new requesters adds their scans to the intermediate list, meaning they then dont schedule an execution because the count is still 2+. If so their scans would stay in the intermediate list and it never get executed until a later requester comes along to add a further request after enough decrements have occurred to make it then actually schedule something.

clebertsuconic: The scheduler here is  single thread... 
If you like I can make it synchronized just in case the method is ever used outside of the mirror scope.

clebertsuconic: @gemmellr https://github.com/apache/activemq-artemis/pull/3743

gemmellr: Ah, I thought it was the IO threads calling it, I see now it is the Queue thread. I do still think it makes sense for it to protect itself though rather than rely on callers to though, certainly as it _appears_ to try to currently but doesnt entirely.

PR #3741, opened 2021-09-13
ARTEMIS-3474 - replace whitelist with allowlist in management.xml

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-3474


andytaylor:
I will update with your feedback

andytaylor:
Updated as requested

PR #3742, opened 2021-09-13
ARTEMIS-2007 - refactor to make use of existing refCountForConsumers …

gtully:
…for tracking consumer count and remove need for volatile redistributor

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -179,7 +179,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
    protected final PageSubscription pageSubscription;
 
-   private ReferenceCounter refCountForConsumers;
+   private final ReferenceCounter refCountForConsumers;
 
    private final PageIterator pageIterator;
 
@@ -218,17 +218,17 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
    protected final ScheduledDeliveryHandler scheduledDeliveryHandler;
 
-   private AtomicLong messagesAdded = new AtomicLong(0);
+   private final AtomicLong messagesAdded = new AtomicLong(0);
 
-   private AtomicLong messagesAcknowledged = new AtomicLong(0);
+   private final AtomicLong messagesAcknowledged = new AtomicLong(0);
 
-   private AtomicLong ackAttempts = new AtomicLong(0);
+   private final AtomicLong ackAttempts = new AtomicLong(0);
 
-   private AtomicLong messagesExpired = new AtomicLong(0);
+   private final AtomicLong messagesExpired = new AtomicLong(0);
 
-   private AtomicLong messagesKilled = new AtomicLong(0);
+   private final AtomicLong messagesKilled = new AtomicLong(0);
 
-   private AtomicLong messagesReplaced = new AtomicLong(0);
+   private final AtomicLong messagesReplaced = new AtomicLong(0);
 
    private boolean paused;
 
@@ -261,8 +261,8 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
    private final SimpleString address;
 
-   // redistributor goes in the consumers list, this signals its presence and allows for easy comparison/check
-   private volatile ConsumerHolder<Redistributor> redistributor;
+   // redistributor singleton goes in the consumers list
+   private ConsumerHolder<Redistributor> redistributor;
 
    private ScheduledFuture<?> redistributorFuture;
 
@@ -634,6 +634,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
       this.id = queueConfiguration.getId();
 
       this.address = queueConfiguration.getAddress();
+      this.refCountForConsumers = queueConfiguration.isTransient() ? new TransientQueueManagerImpl(server, queueConfiguration.getName()) : new QueueManagerImpl(server, queueConfiguration.getName());
 
       this.addressInfo = postOffice == null ? null : postOffice.getAddressInfo(address);
 
@@ -861,13 +862,6 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    }
 
    // Queue implementation ----------------------------------------------------------------------------------------
-   @Override
-   public synchronized void setConsumersRefCount(final ReferenceCounter referenceCounter) {
-      if (refCountForConsumers == null) {
-         this.refCountForConsumers = referenceCounter;
-      }
-   }
-
    @Override
    public ReferenceCounter getConsumersRefCount() {
       return refCountForConsumers;
@@ -1442,13 +1436,8 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
                if (delayBeforeDispatch >= 0) {
                   dispatchStartTimeUpdater.compareAndSet(this,-1, delayBeforeDispatch + System.currentTimeMillis());
                }
-
-            }
-
-            if (refCountForConsumers != null) {
                refCountForConsumers.increment();
             }
-
          }
       }
    }
@@ -1485,7 +1474,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
             if (consumerRemoved) {
                consumerRemovedTimestampUpdater.set(this, System.currentTimeMillis());
-               if (getConsumerCount() == 0) {
+               if (refCountForConsumers.decrement() == 0) {
                   stopDispatch();
                }
             }
@@ -1495,11 +1484,6 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
             }
 
             groups.removeIf(consumer::equals);
-
-
-            if (refCountForConsumers != null) {
-               refCountForConsumers.decrement();
-            }
 
          }
       }
@@ -1557,7 +1541,7 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
    @Override
    public synchronized void cancelRedistributor() {
       clearRedistributorFuture();
-
+      hasUnMatchedPending = false;
       if (redistributor != null) {
          try {
             redistributor.consumer.stop();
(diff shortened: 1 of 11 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -218,17 +218,17 @@ private void checkIDSupplier(NodeStore<MessageReference> nodeStore) {
 
    protected final ScheduledDeliveryHandler scheduledDeliveryHandler;
 
-   private AtomicLong messagesAdded = new AtomicLong(0);
+   private final AtomicLong messagesAdded = new AtomicLong(0);
```
franz1981, line 221: Sometime we use AtomicLongFieldUpdater and sometime we use AtomicLong. This could be the chance to unify the way we treat counters like these, chosing a single tool for the job...

IMO if these counters are not used for logic, but for telemetry, probably a `LongAdder` (or some of the JCTools counters) would work much better then `AtomicLong/AtomicLongFieldUpdater`. because `LongAdder` scales way better and cost must less for the thread that increment it, especially if is happening in the hot path

franz1981 (changes requested):
Small changes not necessary for this specific PR, could be addressed as a separate refactor too.
Generally LGTM

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
clebertsuconic: QueueImpl is a heavy object.. I agree the FieldUPdater would be better on this case. it would help make it lighter.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -218,17 +218,17 @@ private void checkIDSupplier(NodeStore<MessageReference> nodeStore) {
...
 
-   private AtomicLong messagesAcknowledged = new AtomicLong(0);
+   private final AtomicLong messagesAcknowledged = new AtomicLong(0);
 
-   private AtomicLong ackAttempts = new AtomicLong(0);
+   private final AtomicLong ackAttempts = new AtomicLong(0);
```
clebertsuconic, line 225: hmmm... not part of your change.. but if you could improve it with the FieldUpdater :)

clebertsuconic: the FieldUpdater thing I asked is a wish... not mandatory for your PR... if you can do it.. great... otherwise we can make it another PR for that.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1572,18 +1556,7 @@ public synchronized void cancelRedistributor() {
...
-            final int size = consumers.size();
-            if (size > 0 && redistributor != null) {
-               return size - 1;
-            } else {
-               return size;
-            }
-         }
-      }
-      return consumers.size();
+      return refCountForConsumers.getCount();
```
michaelandrepearce, line 1559: I dont like this tbh, you have a collection with is consumers, should just use its value, else we end up having issues of one size thing being different from another.

gtully: the point is there are two different counts, now that consumers has a redistributor that is internal. The need to keep then in sync is not new and it is confined to add/remove consumers

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3082,9 +3055,7 @@ private boolean deliver() {
                   numNoMatch = 0;
                   numAttempts = 0;
 
-                  if (consumer != redistributor) {
-                     ref = handleMessageGroup(ref, consumer, groupConsumer, groupID);
-                  }
+                  ref = handleMessageGroup(ref, consumer, groupConsumer, groupID);
```
gemmellr, line 3058: If the redistributor now handles groups and didnt before, does it need to do anything extra when it is cancelled that it wasnt doing before (since it didnt need to) that e.g other consumers currently do when removed?

gtully: that != was always false, it was comparing different types in error. has been that way seems since the beginning. the redistributor has logic to reply busy when a message has groups, so it will never handle a group message.  the handle returns the ref if there are no group bits in play so it is safe to always call it.

gtully: it did handle groups before!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -179,7 +179,7 @@
 
    protected final PageSubscription pageSubscription;
 
-   private ReferenceCounter refCountForConsumers;
+   protected final ReferenceCounter refCountForConsumers;
```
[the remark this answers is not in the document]
gtully: true, that is not necessary. will fix. thanks!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gemmellr: Seems the incorrect typed check went in with the previous commit in place of a null check looking to similarly protect it: 
https://github.com/apache/activemq-artemis/commit/224b89810d023e3ee73b014a48585da8f6048237#diff-fc4a9a0dc01f79ecc1435246dc85446fa57d6d04e1dffed53c522d8c145623fdR3085

But yep, didnt know that the redistributor handle check returns NO_MATCH (it only returns 'busy' if not active), so it never got to either of those old checks anyway (though by returning 'NO_MATCH' it will reset the consumers iterator, meaning any consumer after it in the list isnt getting tried until it goes away...).

gtully: the reset on NO_MATCH I am not understanding. Does that make sense to you?

gemmellr: Not particularly no. Neither does the one for 'HANDLED' entirely. Both seem to suggest it will always prefer earlier consumers until they arent 'busy enough' to let it try others. Though it wont then try the others if there is a 'NO_MATCH', it will try the earlier ones again even if still busy.

Looking over it several times now, there is at least the fact that adding a consumer always cancels the redistributor during the process, before adding it to the consumer list, so the redistributor shouldnt ever end up in the middle of the list...so the reset in the redistributor case shouldn't really be different than hitting the end of the list.

The only other thing I do see for the non-distributor case is that you apparently need to call reset to see updates to the consumer list, which can change as the loop is processing due to the fact it goes in and out of the synchronization. Kind of feels like the obvious point for that is at the end of the existing entries though rather on the first no-match.

gtully: ok, the reset() is necessary to have any consumer get a second message or try the next message in its iterator. If there are 6 messages pending and the first does not match, unless we reset, that consumer does not get to check the other 5 till the next delivery attempt. With the reset, it is a candidate for another immediate loop.

gtully: of note: the reset() does not put the cursor back to the start, rather it moves the end such that the current element gets another go. The iterator is circular if reset is called. In this way, reset does not skip the next element.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: in hindsight I should not have touched the finals! but for sure that can be another PR

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gemmellr: Javadoc could probably do with an update to describe the actual behaviour then if thats the case:  reset() "Resets the iterator so you can re-iterate over all elements." 

Last I looked at it it seemed to me like it would reset the cursor (especially if there was a new consumer added before, since it replaced the underlying iterator entirely with a new one), which is what I would expect a 'reset' method to do. There is a separate 'repeat()' I think that I would expect to do what you just described.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: am going to leave that to a separate PR and investigate AtomicLongFieldUpdater

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gtully: I have updated the javadoc.  the reset - will pick up changes, but will continue iteration with a restart from the current point, essentially making the collection circular.
the the repeat just replays the last element.

gemmellr: It didnt seem like it would continue from the current point (especially if there were changes, when it replaces the underlying iterator) since it implements the reset() with a "moveTo(-1)".

But, on closer inspection, it seems to do a mixture of things - it apparently _does_ reset the cursor to the start as I thought due to the above, but only in terms of 'the highest priority' sub-iterator. However it _does not_ reset the cursor within that particular sub-iterator to the start (except perhaps when it has changed, when I guess it will thanks to replacing the iterators). The stuff going on here so far isnt hitting the priority variation, so it isnt seeing the distinction.

In short, I dont think it does exactly what either of us think it does, and I'm thinking that QueueConsumersImplTest is lacking some tests to verify whatever its actually meant to be doing overall.

PR #3743, opened 2021-09-13
RTEMIS-3464 Protecting scheduled tasks counter with a synchronized

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -154,13 +154,17 @@ public void addScanAck(BooleanSupplier retryBeforeScan, ToIntFunction<PagedRefer
...
    @Override
    public void performScanAck() {
-      // we should only have a max of 2 scheduled tasks
-      // one that's might still be currently running, and another one lined up
-      // no need for more than that
-      if (scheduledScanCount.incrementAndGet() <= 2) {
-         executor.execute(this::actualScanAck);
-      } else {
-         scheduledScanCount.decrementAndGet();
+      // notice performScanAck is supposed to be called from a single thread.
```
gemmellr, line 157: Might be useful for the API to say this if its supposed to, and perhaps which thread that is expected to be.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -154,13 +154,17 @@ public void addScanAck(BooleanSupplier retryBeforeScan, ToIntFunction<PagedRefer
...
+      // I am adding a synchronization point here to protected scheduledScanCount
+      synchronized (scanList) {
+         // we should only have a max of 2 scheduled tasks
+         // one that's might still be currently running, and another one lined up
+         // no need for more than that
+         if (scheduledScanCount.incrementAndGet() <= 2) {
+            executor.execute(this::actualScanAck);
+         } else {
+            scheduledScanCount.decrementAndGet();
+         }
```
gemmellr, line 167: Looking at it closer, the only non-test usage always calls this right after the add call, in the same thread, and both methods now synchronize...it would seem simpler to combine them at this point and only have a single synchronized use for the task additions?

In doing that, if moving the 'worker thread decrement' to the start when it collates its local list, under synchronization, the AtomicInteger and related incs/decs could just be replaced entirely with a simple 'scheduled' boolean and simplify things here to just 'if not scheduled, schedule'.

clebertsuconic: good idea..will do that.. thanks

I will mark this conversation as resolved when I have done it

will ammend this PR

clebertsuconic: done...

PR #3744, opened 2021-09-13
ARTEMIS-3475 avoiding possible recursion on toString and improving ov…

clebertsuconic:
…eral PacketImpl toString

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/PacketImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/PacketImpl.java
@@ -418,9 +418,14 @@ public class PacketImpl implements Packet {
       return true;
    }
 
+   /** extensions of this class are supposed to use getPacketString to provide toString functionality */
    @Override
-   public String toString() {
-      return getParentString() + "]";
+   public final String toString() {
+      return getPacketString() + "]";
+   }
+
+   protected String getPacketString() {
+      return this.getClass().getSimpleName() + "(" + this.getClass().getSimpleName() + ")[type=" + type + ", channelID=" + channelID + ", responseAsync=" + isResponseAsync() + ", requiresResponse=" + isRequiresResponse() + ", correlationID=" + getCorrelationID();
    }
 
    @Override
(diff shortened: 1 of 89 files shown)


brusdev:
@clebertsuconic nice work, thanks

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationResponseMessageV2.java
```diff
@@ -62,10 +62,9 @@ public void decodeRest(final ActiveMQBuffer buffer) {
    }
 
    @Override
-   public String toString() {
-      StringBuffer buf = new StringBuffer(getParentString());
+   protected String getPacketString() {
+      StringBuffer buf = new StringBuffer(getPacketString());
```
gemmellr, line 66: This just introduced the very type of recursion the change was looking to prevent. Will send a PR to fix.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/PacketImpl.java
```diff
@@ -418,9 +418,14 @@ public boolean isRequiresConfirmations() {
...
+   /** extensions of this class are supposed to use getPacketString to provide toString functionality */
    @Override
-   public String toString() {
-      return getParentString() + "]";
+   public final String toString() {
+      return getPacketString() + "]";
+   }
+
+   protected String getPacketString() {
+      return this.getClass().getSimpleName() + "(" + this.getClass().getSimpleName() + ")[type=" + type + ", channelID=" + channelID + ", responseAsync=" + isResponseAsync() + ", requiresResponse=" + isRequiresResponse() + ", correlationID=" + getCorrelationID();
```
gemmellr, line 428: This prints the class name twice, in format "\<classname\>(\<classname\>)[etc]" which seems unusual. Previously it was doing "PACKET(\<classname\>)[etc]".

I'm guessing since you dropped the PACKET you meant to do that and make it just the class name without braces, I'll tweak it that way that in my PR, and if not it can be changed to the actual intent.

EDIT: I now see it did used to be included a second time inside the [etc attributes] bit. Still seems odd to have it twice at the start, I'll proceed with just removing one for now.

PR #3747, opened 2021-09-14
ARTEMIS-3476 Upgrade maven-artifact to 3.8.2

brusdev:


gemmellr:
Seems ok, I think you should just push it.

PR #3748, opened 2021-09-14
ARTEMIS-3478 Upgrade netty version to 4.1.68.Final

brusdev:


gemmellr:
Looks good, I think you should just push it.

PR #3749, opened 2021-09-14
ARTEMIS-3477 Upgrade selenium-java to 3.141.59

brusdev:


gemmellr:
Seems ok, I think you should just push it.

(We should really use properties and in the parent pom to govern versions, but theres a wider change needed here, can do it later all together)

PR #3750, opened 2021-09-14
ARTEMIS-3479 Upgrade commons-codec version to 1.15

brusdev:


gemmellr:
Looks good, I think you should just push it.

PR #3752, opened 2021-09-15
[ARTEMIS-2545]: Auto queue creation does not work with MDBs.

ehsavoie:
* Adding support for queue autocreation from the resource adapter.

Issue: https://issues.apache.org/jira/browse/ARTEMIS-2545


jbertram:
The JMS client is doing something _very_ similar in `o.a.a.a.j.c.ActiveMQSession#createConsumer()`. It would be awesome to refactor this code out into the `artemis-commons` module and re-use it between the JMS client and the JCA RA. You could create a new class called `org.apache.activemq.artemis.utils.AutoCreateUtil`, for example.

ehsavoie:
@jbertram I'm afraid this can only live in jms-client as it needs the ClientSession

ehsavoie:
@jbertram I've updated the PR to try to have more code reuse.

PR #3753, opened 2021-09-15
ARTEMIS-3484 flesh out Jakarta Messaging support

jbertram:
Back in version 2.17.0 we began to provide Maven artifacts for Jakarta
Messaging client resources. This commit expands that support in the
following ways:

 - Distribute a Jakarta Messaging 3.0 client with the broker (in the
   'lib/client' directory alongside the JMS client.
 - Update documentation.
 - Add example using the Jakarta Messaging client.
 - Update Artemis CLI to use core instead of JMS as it was causing
   conflicts with the new Jarkarta Messaging client.
 - Add example to build Jarkarta Messaging version of the JCA RA for
   deployment into Jakarta EE 9 application servers.

### DIFF
--- a/artemis-distribution/pom.xml
+++ b/artemis-distribution/pom.xml
@@ -44,6 +44,11 @@
       </dependency>
       <dependency>
          <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-jakarta-client-all</artifactId>
+         <version>${project.version}</version>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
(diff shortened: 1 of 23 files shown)


clebertsuconic:
just for a future reference.. no changes were made to the CLI plugin..


A change was made to the Maven CLI Plugin... on the validate started session... (the one that was creating a session to validate if the server was up.. but that's a different thing)..


not a big deal.. I'm just making this comment for future references if someone ever needs to read this PR

jbertram:
@clebertsuconic, thanks for the clarification. I got that mixed up in my commit message.

PR #3755, opened 2021-09-16
remove duplicate cache-timeout docs

pwright:


pwright:
@brusdev as discussed

PR #3757, opened 2021-09-17
ARTEMIS-3485 Fix web plugin contexts

brusdev:
Support custom web plugin contexts defined with the `url` attribute
of the `app` element in bootstrap.xml.


jbertram:
I tested this, and it works great. Nice job, @brusdev!

PR #3759, opened 2021-09-17
rename protocols page

pwright:
@clebertsuconic  as discussed, but I'm not sure if this makes sense given the content of the page, alternative might be 'Supported Protocols and APIs'?

### DIFF
--- a/docs/user-manual/en/SUMMARY.md
+++ b/docs/user-manual/en/SUMMARY.md
@@ -10,7 +10,7 @@
 * [Using the Server](using-server.md)
 * [Upgrading](upgrading.md)
 * [Address Model](address-model.md)
-* [Protocols and Interoperability](protocols-interoperability.md)
+* [Supported Protocols](supported-protocols.md)
 * [AMQP](amqp.md)
     * [Broker Connections](amqp-broker-connections.md)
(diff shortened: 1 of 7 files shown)


docs/user-manual/en/SUMMARY.md
```diff
@@ -10,7 +10,7 @@
 * [Using the Server](using-server.md)
 * [Upgrading](upgrading.md)
 * [Address Model](address-model.md)
-* [Protocols and Interoperability](protocols-interoperability.md)
```
clebertsuconic, line 13: but you did not rename the md here.. did you?

gtully: a quick search, I see protocols-interoperability used in links in mqtt.md amqp.md etc

pwright: I updated the PR, however I think these links show that the chapter *should not* be renamed.

ie "interoperability"  helps reader understand the scope of the doc, it's not just a list of supported protocols.

Why would i click on 'Supported protocols' to read about 'Configuration of max frame payload length for STOMP web-socket'?

pwright:
closing as 'not a good change'

PR #3763, opened 2021-09-21
2.6.x one client connected to the server ,why there are two sessions created?

alwyngo:
Using the manage notifications to monitor the session created event, but when one client conneted to the server, the server create two sessions. Dose this normal?


clebertsuconic:
I"m closing this . please use the proper way to communicate with us... 
User's list preferably... But opening a Pull Request from 2.6.x into Main is definitely not a way to ask a question on any project on GitHub.

It is not ok to write a question through a Pull Request.

PR #3766, opened 2021-09-22
ARTEMIS-3493 - expose User ID (JMS Message ID) in send tab of console

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-3493

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
@@ -151,6 +151,26 @@ public interface AddressControl {
                       @Parameter(name = "password", desc = "The users password to authenticate with") String password) throws Exception;
 
    /**
+    * @param headers  the message headers and properties to set. Can only
+    *                 container Strings maped to primitive types.
+    * @param body     the text to send
+    * @param durable
+    * @param user
+    * @param password @return
+    * @param createMessageId whether or not to auto generate a Message ID
+    * @throws Exception
+    */
+   @Operation(desc = "Sends a TextMessage to a password-protected address.", impact = MBeanOperationInfo.ACTION)
+   String sendMessage(@Parameter(name = "headers", desc = "The headers to add to the message") Map<String, String> headers,
+                      @Parameter(name = "type", desc = "A type for the message") int type,
+                      @Parameter(name = "body", desc = "The body (byte[]) of the message encoded as a string using Base64") String body,
+                      @Parameter(name = "durable", desc = "Whether the message is durable") boolean durable,
+                      @Parameter(name = "user", desc = "The user to authenticate with") String user,
+                      @Parameter(name = "password", desc = "The users password to authenticate with") String password,
+                      @Parameter(name = "createMessageId", desc = "whether or not to auto generate a Message ID") boolean createMessageId) throws Exception;
+
+
+   /**
     * Pauses all the queues bound to this address.Messages are no longer delivered to all its bounded queues.
     * Newly added queue will be paused too until resume is called.
     * @throws java.lang.Exception
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/QueueControl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/QueueControl.java
@@ -566,6 +566,25 @@ public interface QueueControl {
                       @Parameter(name = "password", desc = "The users password to authenticate with") String password) throws Exception;
 
    /**
+    * @param headers  the message headers and properties to set. Can only
+    *                 container Strings maped to primitive types.
+    * @param body     the text to send
+    * @param durable
+    * @param user
+    * @param password @return
+    * @param createMessageId whether or not to auto generate a Message ID
+    * @throws Exception
+    */
+   @Operation(desc = "Sends a TextMessage to a password-protected destination.", impact = MBeanOperationInfo.ACTION)
+   String sendMessage(@Parameter(name = "headers", desc = "The headers to add to the message") Map<String, String> headers,
+                      @Parameter(name = "type", desc = "A type for the message") int type,
+                      @Parameter(name = "body", desc = "The body (byte[]) of the message encoded as a string using Base64") String body,
+                      @Parameter(name = "durable", desc = "Whether the message is durable") boolean durable,
+                      @Parameter(name = "user", desc = "The user to authenticate with") String user,
+                      @Parameter(name = "password", desc = "The users password to authenticate with") String password,
+                      @Parameter(name = "createMessageId", desc = "whether or not to auto generate a Message ID") boolean createMessageId) throws Exception;
+
+   /**
     * Changes the message's priority corresponding to the specified message ID to the specified priority.
     *
     * @param newPriority between 0 and 9 inclusive.
--- a/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/addressSendMessage.js
+++ b/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/addressSendMessage.js
@@ -42,6 +42,20 @@ var Artemis;
                         <div class="form-group">
                             <label>Durable </label>
                             <input id="durable" type="checkbox" ng-model="$ctrl.message.durable" value="true">
+                            <button type="button" class="btn btn-link jvm-title-popover"
+                                      uib-popover-template="'durable-info.html'" popover-placement="bottom-left"
+                                      popover-title="Durable" popover-trigger="'mouseenter'">
+                                <span class="pficon pficon-info"></span>
+                            </button>
+                        </div>
+                        <div class="form-group">
+                            <label>Create Message ID </label>
+                            <input id="messageID" type="checkbox" ng-model="$ctrl.message.messageID" value="true">
+                            <button type="button" class="btn btn-link jvm-title-popover"
+                                      uib-popover-template="'message-id-info.html'" popover-placement="bottom-left"
+                                      popover-title="Message ID" popover-trigger="'mouseenter'">
+                                <span class="pficon pficon-info"></span>
+                            </button>
                         </div>
                     </form>
                 </div>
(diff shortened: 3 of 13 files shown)


artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/addressSendMessage.js
```diff
@@ -43,6 +43,10 @@ var Artemis;
                             <label>Durable </label>
                             <input id="durable" type="checkbox" ng-model="$ctrl.message.durable" value="true">
                         </div>
+                        <div class="form-group">
+                            <label>Create User ID </label>
```
gtully, line 47: I think, call this message id, it does not identify the user, it is a user supplied id for the message.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
```diff
@@ -150,6 +150,26 @@ String sendMessage(@Parameter(name = "headers", desc = "The headers to add to th
...
+    * @throws Exception
+    */
+   @Operation(desc = "Sends a TextMessage to a password-protected address.", impact = MBeanOperationInfo.ACTION)
+   String sendMessage(@Parameter(name = "headers", desc = "The headers to add to the message") Map<String, String> headers,
+                      @Parameter(name = "type", desc = "A type for the message") int type,
+                      @Parameter(name = "body", desc = "The body (byte[]) of the message encoded as a string using Base64") String body,
+                      @Parameter(name = "durable", desc = "Whether the message is durable") boolean durable,
+                      @Parameter(name = "user", desc = "The user to authenticate with") String user,
+                      @Parameter(name = "password", desc = "The users password to authenticate with") String password,
+                      @Parameter(name = "createUserId", desc = "whether or not to auto generate a userID, used as a Message ID in JMS") boolean createUserId) throws Exception;
```
gemmellr, line 170: Why not just say 'addMessageId'? Contiuning to reference this 'userId' concept, when its absolutely about a message-id, and definitely not a userId, and then having to document that it is a message id, just seems unnecessarily confusing.

andytaylor:
Message ID already exists and this is referred to in the browse table as User ID

artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/addressSendMessage.js
```diff
@@ -87,15 +91,16 @@ var Artemis;
...
             </p>
             <script type="text/ng-template" id="send-message-instructions.html">
             <div>
                 <p>
                     This page allows you to send a message to the chosen address. The message will be of type <code>text</code>
                     message and it will be possible to add headers to the message. The sending of the message will be authenticated
                     using the username and password set ion <code>preferences</code>, if this is not set then these will
-                    be null.
+                    be null. The <code>durable</code> checkbox makes the message persistent and the <code>create user Id</code> will auto generate a
+                    UUID user for the message which for a JMS Message is the JMS Message ID.
```
gemmellr, line 103: _"the \<code\>create user Id\</code\> will"_ seems like it should have _checkbox_ or something else added to complete.

_"a UUID user for"_ seems like it is missing _id_.

gemmellr:
I would suggest some other way to handle that than just perpetuating the existing confusing naming throughout more of the codebase and UI areas that aren't currently affected by it and don't seem like they need to be. Starting to make thing more obvious rather than making it a more widespread issue.

Having a flag called "createUserId" here just seems confusing in general for this, given its specifically about adding what I believe most people would think of when you say Message ID in reference to a sent or received message. More so considering it also isnt 'created by the user' at all in this case (nor for many client cases either, particularly not the JMS case specifically being referenced in the new doc).

The other thing being called MessageID elsewhere in the codebase is I believe essentially internal detail _always_ set by the broker and so not something anyone ever gets to set for themselves during send, so there seems little need to avoid referencing a more intuitive 'addMessageId' here, in something specifically about creating and sending messages including what I expect most users think of as Message IDs. Having the description note the MessageID being created is referenced as the UserID in other UI / Core messages if needed seems more intuitive than the reverse approach currently being taken. 

You said it is referenced as UserID in the browse table. Could said table even adjust the reference to something more obvious and have the description note the other aspect (as with the description here)? That would seem nice too, if separate.

A further larger separate improvement would of course be to actually change that UserID and MessageID naming elsewhere to something more obvious / less confusing, even if it meant duplicating things for a time for compatibility, in e.g management responses etc, until a breaking change could actually remove it. But a good start would just be using the more obvious terminology in cases where it seems simple enough to do, like here.

artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/addressSendMessage.js
gemmellr: It isnt really a user supplied id in this case, or any sent with a JMS client, as its being generated by something else. YEt more reason to reference the message id it is.

andytaylor:
I'm happy to change it if you have a suggestion as to what we could reference it as.

gemmellr:
For the purpose of this change I would just refer to it as the Message ID for the new bits, and have any descriptions/doc note the Message ID in question is 'also referred to as UserID in some areas', as opposed to the reverse approach it takes now. I think it would be more in line with user expectations and be plenty clear from the 'send with message id' context what it all means.

Going beyond that, if you mean for e.g the other table referencing it as User ID, I'd also go with swapping it for some variation of Message ID there that too if possible, e.g change the format in some way, and also updating the other existing 'Message ID' to be referred to as something else (e.g simply ID, or Broker Message ID, or perhaps Message Store ID since thats what it seems to really represent...could also swap ID with Index). Alternatively if changing how that is referenced isnt also possible, then just something long winded like Application Message ID would still seem a lot more obvious/intuitive than User ID.

If you mean the larger change, then I would go with e.g ID / Broker ID / Message Store ID in place of the existing MessageID and deprecate that. Something else to replace UserID (e.g MessageID with different format such as its capitalization, separator etc). Later remove the old stuff entirely.

I think its also worth saying that any management operations etc that do currently take a 'Message ID' currently, i.e the store id, are taking a long. If changing things so the application style 'message id' is stored in and referenced by something named more like 'message id', I dont think this would cause that much confusion since most people are likely used to that being another type e.g String, so I dont think it would become particularly unclear for e.g manager uses that it changed.

andytaylor:
ok, I have changed this to be called 'Message ID' and also added an info popup that explains it's relation to user ID and a JMS Message ID.

gemmellr:
Did you forget to add some changes? Still looks mostly referred to as user id other than a couple bits.

andytaylor:
that should be it sorted now

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/AddressControl.java
```diff
@@ -150,6 +150,26 @@ String sendMessage(@Parameter(name = "headers", desc = "The headers to add to th
...
                       @Parameter(name = "password", desc = "The users password to authenticate with") String password) throws Exception;
 
+   /**
+    * @param headers  the message headers and properties to set. Can only
+    *                 container Strings maped to primitive types.
+    * @param body     the text to send
+    * @param durable
+    * @param user
+    * @param password @return
+    * @param createMessageId whether or not to auto generate a userID, used as a Message ID in JMS
```
gemmellr, line 160: javadoc description needs similarly revised

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/QueueControl.java
```diff
@@ -565,6 +565,25 @@ String sendMessage(@Parameter(name = "headers", desc = "The headers to add to th
...
                       @Parameter(name = "password", desc = "The users password to authenticate with") String password) throws Exception;
 
+   /**
+    * @param headers  the message headers and properties to set. Can only
+    *                 container Strings maped to primitive types.
+    * @param body     the text to send
+    * @param durable
+    * @param user
+    * @param password @return
+    * @param createMessageId whether or not to auto generate a userID, used as a Message ID in JMS
```
gemmellr, line 575: Same

artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/addressSendMessage.js
```diff
@@ -42,6 +42,20 @@ var Artemis;
...
                             <input id="durable" type="checkbox" ng-model="$ctrl.message.durable" value="true">
+                            <button type="button" class="btn btn-link jvm-title-popover"
+                                      uib-popover-template="'durable-info.html'" popover-placement="bottom-left"
+                                      popover-title="Durable" popover-trigger="'mouseenter'">
+                                <span class="pficon pficon-info"></span>
+                            </button>
+                        </div>
+                        <div class="form-group">
+                            <label>Create User ID </label>
+                            <input id="messageID" type="checkbox" ng-model="$ctrl.message.messageID" value="true">
```
gemmellr, line 53: Label needs updated (line 52 if this isnt clear due to the existing comments on it)

artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/addressSendMessage.js
```diff
@@ -98,7 +112,23 @@ var Artemis;
...
+                    The Message ID is an automatically generated UUID that is set on the Message by the broker before it is routed.
+                    If using a JMS client this would be the JMS Message ID on the JMS Message, this typically would not get
+                    set for non JMS clients. Historically and on some other tabs this is also referred to as the User ID.
+                </p>
+            </div>
+            </script>
+            <script type="text/ng-template" id="durable-info.html">
+            <div>
+                <p>
+                    If durable the message will be marked persistent and written to the brokers journal.
```
gemmellr, line 128: [if the destination queue is durable] ?

artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/sendMessage.js
```diff
@@ -98,7 +112,23 @@ var Artemis;
...
+                    The Message ID is an automatically generated UUID that is set on the Message by the broker before it is routed.
+                    If using a JMS client this would be the JMS Message ID on the JMS Message, this typically would not get
+                    set for non JMS clients. Historically and on some other tabs this is also referred to as the User ID.
+                </p>
+            </div>
+            </script>
+            <script type="text/ng-template" id="durable-info.html">
+            <div>
+                <p>
+                    If durable the message will be marked persistent and written to the brokers journal.
```
gemmellr, line 128: Same as other.

gemmellr (commented):
Looks good, couple small niggles remaining.

andytaylor:
updated as requested

gemmellr:
Changes look good. Tried the functionality out, worked well (once you told me how to use that bit of the console, ick).

Only comment would be lack of tests. Seems like the Java bits could/should have some added, perhaps easily as the change seems to updated existing related test classes.

andytaylor:
tests added

PR #3767, opened 2021-09-22
ARTEMIS-3488 Added env variable ARTEMIS_MASK_PASSWORD_CODE

Akvel:


jbertram:
I've reworked this (changed the property name, added docs, etc.) and sent #3780.

PR #3768, opened 2021-09-22
[ARTEMIS-3494]: ActiveMQClientProtocolManagerFactory shouldn't have a…

ehsavoie:
… private constructor.

* Removing the 'private' constructor
* Removing the use of a static getInstance

Issue: https://issues.apache.org/jira/browse/ARTEMIS-3494


jbertram:
@ehsavoie, can you clarify why this change is necessary/beneficial?

ehsavoie:
I was recreating an ActiveMQConnectionFactory from a 'previous' one but when setting the ProtocolManagerFactory it fails because the code underlying is using an empty constructor

brusdev (approved):
@ehsavoie thanks it LGTM +1

gemmellr:
Seems unusual to explicitly render the static method unused in the Artemis codebase and then not modify it in some way (e.g deprecation, or removal). As is anyone could just choose to remove it as 'unused', though it sounds like you may have existing code that was relying on it. Its also now unused and untested for anyone else still relying on it.

PR #3770, opened 2021-09-23
ARTEMIS-3106 - add some doc for SASL SCRAM-SHA

gtully:


docs/user-manual/en/security.md
```diff
@@ -1087,22 +1087,77 @@ amqp-sasl-gssapi {
...
+##### Configuring a user with SCRAM-SHA data on the server
+
+With SCRAM-SHA, the server's users properties file do not contain any passwords, instead they contain derivative data that
+can be used to respond to a challenge.
+The secure encoded form of the password must be generated or registered using the main method of
+org.apache.activemq.artemis.spi.core.security.jaas.SCRAMPropertiesLoginModule from the artemis-server module and inserting
+the resulting lines into your desired user properties file.
+
+````
+java -cp ... org.apache.activemq.artemis.spi.core.security.jaas.SCRAMPropertiesLoginModule <username> <password> [<iterations>]
```
gemmellr, line 1136: I'd suggest just having it use -cp "\<broker-install\>/lib/*" rather than having folks trying to figure out which specific jars are needed.

Also, having an example of the particular format of output to be added to the artemis-users.properties file for SCRAM-SHA-256 might be good (I believe the command prints them all). Or pointing to the related example that has one (I knew that example exists and it still took me a bit to find it, in examples/protocols/amqp/sasl-scram/)

docs/user-manual/en/security.md
```diff
@@ -1087,22 +1087,77 @@ amqp-sasl-gssapi {
...
+};
+```
+
+##### Configuring a user with SCRAM-SHA data on the server
+
+With SCRAM-SHA, the server's users properties file do not contain any passwords, instead they contain derivative data that
+can be used to respond to a challenge.
+The secure encoded form of the password must be generated or registered using the main method of
+org.apache.activemq.artemis.spi.core.security.jaas.SCRAMPropertiesLoginModule from the artemis-server module and inserting
+the resulting lines into your desired user properties file.
```
gemmellr, line 1133:
```suggestion
the resulting lines into your artemis-users.properties file.
```

Since the examples just explicitly defined it as such.

docs/user-manual/en/security.md
gtully: I had the same thought, but -cp lib/* did not work, till it is quoted from the shell! will tidy this up. thanks for all the feedback, much appreciated

gtully:
rebased and squashed

docs/user-manual/en/security.md
```diff
@@ -1087,22 +1087,75 @@ amqp-sasl-gssapi {
 };
 ```
 
+#### SCRAM-SHA SASL Mechanism
```
gemmellr, line 1090: This (and the new sub sections below) seem to be at the wrong level, suggesting they are part of the "### Kerberos Authentication" section, which clearly they arent.

Looking further up, it seems there is a LoginModules section about each of the modules, and then the Kerberos one sits separately with its own section. Then this [almost] does the same.

I think the section depths should be fixed and this should go above the Kerberos section as its likely more widely applicable.

gtully: good catch, will sort

gtully: the two new login modules need their own little section in the login modules section too

PR #3771, opened 2021-09-24
ARTEMIS-3496 Replica connection to its live should fail fast

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3496


franz1981:
@clebertsuconic do you remember any reason why `reconnect attempts` was set to 1 on purpose on the replica connection to its live?

clebertsuconic:
it was set to 1 as in not meant to retry.. .a single connection...

its a bug... I meant for one connection only.


Replication and clustering should have the retries through the cluster bridge and clustered connection. not through the serverLocator.


Is there a test that fails with this? run the whole test suite just in case.. and if you can add a failing test it would be even greater.

franz1981:
> Is there a test that fails with this?

I'm still investigating on multiple test failure *after* applied this change :(

PR #3776, opened 2021-09-28
ARTEMIS-3501 Added exception handling on #handleAddMessage to not sto…

AntonRoskvist:
…p broker from starting with currupted messages

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/PostOfficeJournalLoader.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/PostOfficeJournalLoader.java
@@ -221,23 +221,27 @@ public class PostOfficeJournalLoader implements JournalLoader {
          long currentTime = System.currentTimeMillis();
 
          for (AddMessageRecord record : valueRecords) {
-            long scheduledDeliveryTime = record.getScheduledDeliveryTime();
+            try {
+               long scheduledDeliveryTime = record.getScheduledDeliveryTime();
 
-            if (scheduledDeliveryTime != 0 && scheduledDeliveryTime <= currentTime) {
-               scheduledDeliveryTime = 0;
-               record.getMessage().setScheduledDeliveryTime(0L);
-            }
+               if (scheduledDeliveryTime != 0 && scheduledDeliveryTime <= currentTime) {
+                  scheduledDeliveryTime = 0;
+                  record.getMessage().setScheduledDeliveryTime(0L);
+               }
 
-            if (scheduledDeliveryTime != 0) {
-               record.getMessage().setScheduledDeliveryTime(scheduledDeliveryTime);
-            }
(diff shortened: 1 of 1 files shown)


AntonRoskvist:
I've been unable to write a proper test for this as I cannot simulate a journal broken in the "correct" way. This did however solve the issue I was seeing with an actual broken journal. It will discard and log the corrupted message and proceed with loading the rest of the journal

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/PostOfficeJournalLoader.java
```diff
@@ -221,23 +221,27 @@ public void handleAddMessage(Map<Long, Map<Long, AddMessageRecord>> queueMap) th
...
 
-            ref.setDeliveryCount(record.getDeliveryCount());
+               ref.setDeliveryCount(record.getDeliveryCount());
 
-            if (scheduledDeliveryTime != 0) {
-               record.getMessage().setScheduledDeliveryTime(0L);
+               if (scheduledDeliveryTime != 0) {
+                  record.getMessage().setScheduledDeliveryTime(0L);
+               }
+            } catch (Throwable e) {
```
clebertsuconic, line 243: logger.warn(e.getMessage, e); ?


some context about the message that is causing it?

clebertsuconic:
I don't think it will discard.. the message would still be in the journal until the next restart... right?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/PostOfficeJournalLoader.java
AntonRoskvist: This is preceded by a massive dump about everything related to the message, so it seemed a bit excessive. An example of that can be seen in the attached log in the ticket, but it includes messageID, why it failed and the entire message bytebuffer (pretty printed)

AntonRoskvist:
At least in my case I end up in #handleNoMessageReferences (from AbstractJournalStorageManager) and the message is removed by:
storageManager.deleteMessage(msg.getMessageID());

And the following log:
AMQ221019: Deleting unreferenced message id=$MESSAGEID from the journal

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/PostOfficeJournalLoader.java
AntonRoskvist: (Thrown by org.apache.activemq.artemis.core.message.impl.CoreMessage #onCheckPropertiesError)

AntonRoskvist: Yes that makes sense, sounds good

PR #3780, opened 2021-09-30
ARTEMIS-3488 add system property for setting key in DefaultSensitiveStringCodec

jbertram:


Akvel:
@jbertram Maybe we should also add this param to https://github.com/apache/activemq-artemis/blob/main/artemis-docker/docker-run.sh#L37 ?

jbertram:
@Akvel, the parameter is already in the `artemis` script which is called by `docker-run.sh`. I don't see how adding it to `docker-run.sh` would be helpful. 

Are you suggesting that a place to set the *value* of `ARTEMIS_DEFAULT_SENSITIVE_STRING_CODEC_KEY` should be added to `docker-run.sh`? If so, I would push back on that since one of the goals with adding this environment property in the first place is so that the value of the key *doesn't* appear in any file (script or otherwise). Whoever calls `docker-run.sh` can set `ARTEMIS_DEFAULT_SENSITIVE_STRING_CODEC_KEY` in their environment.

Akvel:
@jbe

> @Akvel, the parameter is already in the `artemis` script which is called by `docker-run.sh`. I don't see how adding it to `docker-run.sh` would be helpful.
> 
> Are you suggesting that a place to set the _value_ of `ARTEMIS_DEFAULT_SENSITIVE_STRING_CODEC_KEY` should be added to `docker-run.sh`? If so, I would push back on that since one of the goals with adding this environment property in the first place is so that the value of the key _doesn't_ appear in any file (script or otherwise). Whoever calls `docker-run.sh` can set `ARTEMIS_DEFAULT_SENSITIVE_STRING_CODEC_KEY` in their environment.

You are rigth. I miss that artemis alredy get this param from env.

PR #3782, opened 2021-10-04
ARTEMIS-3313 routing-type conflict during import/export

jbertram:


AntonRoskvist:
The title got a bit unclear because this is a rework of an old PR made when this issue seemed to only affect actual import/export of messages (and the separate issue of retried messages disappearing if there is no receiving queue defined). Turns out this applies to all operations moving messages from one destination to another where the message "routingType" does not correspond to its destination type. This includes at least Import/Export and scale-down (maybe more).

The problem with the current behavior is that such a move does NOT change the routingType so during the scenarios described above messages get silently dropped because they do not have a receiving destination of the type that the message has.

It primarily affects messages placed on DLA using auto-created resources since that can move anycast messages to the multicast DLA queue.

If it helps I still have the old test covering import/export, but the current ones cover the scenario in the exact same way that test does.

Old test:
https://github.com/AntonRoskvist/activemq-artemis/blob/2bf466be190be21dee037b2fb3aa6fd71cd078c3/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/persistence/XmlImportExportTest.java @ testImportWrongRoutingType

PR #3783, opened 2021-10-04
ARTEMIS-3513 Compacting exception invalidates deletes and updates

clebertsuconic:


artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -710,7 +710,12 @@ private static ActiveMQMessage toAMQMessage(MessageReference reference,
    private static <T> T getObjectProperty(ICoreMessage message, Class<T> type, SimpleString property) {
       if (message.getPropertyNames().contains(property)) {
          try {
-            return type.cast(message.getObjectProperty(property));
+            Object value = message.getObjectProperty(property);
+            if (type == String.class) {
+               return (T)value.toString();
```
clebertsuconic, line 715: @jbertram review this please, since it was a recent change of yours?

jbertram: My recent change was mainly to ensure that any `ClassCastException` would be isolated and logged rather than blowing up the whole message conversion. Your change doesn't alter that semantic so it's fine. Although I think the `else` block should have this instead:
```
return type.cast(value);
```
Since `value` already has the result of `message.getObjectProperty(property)`.

clebertsuconic: thanks @jbertram

PR #3785, opened 2021-10-05
ARTEMIS-3038: unwind effect of defunct changes from ARTEMIS-1264

gemmellr:
Unwinds the effects of the RFC 2712 'Kerberos SSL ciphers' support added in ARTEMIS-1264. The functionality was recommended against use even then, and was removed entirely from JDK11. It has been disabled by default in JDK8 for a while. OpenSSL removed its equivalent support in 2015. It is no longer being tested, with the tests already removed (a3de3d4c75ba1482706e8c42a5c9b0f9811901eb) since no modern JVMs can do it out of the box.

The code has changed and moved around a lot since as the surrounding areas were updated, other functionality added etc, so this was a case of making the related unwinds rather than reverting what was added as such. The docs included a note of the support, including mention it is insecure, though the specific steps needed to configure and use it were seemingly never added.

Once #3696 lands this functionality becomes 100% defunct, but its largely already unusable and shouldn't really be used anyway so I don't think this PR needs to wait for that one.


gemmellr:
@gtully can you take a look

PR #3788, opened 2021-10-06
ARTEMIS-3365 - add simple local-target balancer example with amqp fai…

gtully:
…lover, remove manditory pool and policy config


gtully:
I don't have a use case for a pool with just a local broker in it, does not seem to make any sense even with a custom policy the probe bits would still not be needed. Leaving the schema enforcing a pool has to have some content or config.

PR #3795, opened 2021-10-14
ARTEMIS-2618 Added optional retry handling on critical I/O Error

MrEasy:
This can be used if running in environments where external applications temporarily lock journal files (e.g. virus scanners).
Disabled by default.

### DIFF
--- a/artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/nio/NIOSequentialFile.java
+++ b/artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/nio/NIOSequentialFile.java
@@ -59,6 +59,21 @@ public class NIOSequentialFile extends AbstractSequentialFile {
     */
    private static final int CHUNK_SIZE = 2 * 1024 * 1024;
 
+   /**
+    * Maximum No. of retries that are done on severe I/O errors before server shuts down.
+    * This can be used if running in environments where external applications temporarily lock journal files (e.g. virus scanners).
+    * Default = 0, meaning no retries are done.
+    * @see #RETRIES_TIMEOUT
+    * @see #open(int, boolean)
+    */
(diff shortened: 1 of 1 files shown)


jbertram:
This idea has merit, but there a couple problems:

 - The code isn't formatted properly so CheckStyle is failing.
 - The retry is limited to `org.apache.activemq.artemis.core.io.nio.NIOSequentialFile#open(int, boolean)`, but there are several implementations of `org.apache.activemq.artemis.core.io.SequentialFile#open(int, boolean)` which might benefit from such a retry.
 - The configuration should be available from `broker.xml` rather than just a system property.

MrEasy:
Can't argue with that of course.
As you can see in the issue though, it has its 2nd birthday in 2 days :-) - so would prefer not investing more time if that then would not result in sth. releasable.
Do you think the points you mentioned are the ones that should complete it?

clebertsuconic:
the reason I never merged this, was because I was more afraid of side effects than anything else. 

What if the Open did leave a file handle open somewhere in the OS, and still issued a warning? How to test it (as a matter of fact there are no tests on this PR, and even if we had it.. it would be difficult to validate it).

and the reason I never closed it... it was I wanted to be polite and find a middle term solution.. but I could never accommodate the request in a way I felt comfortable on merging.


@MrEasy I will close this now, and close the JIRA as a won't fix... just because of those reasons.


Sorry it took me 2 years to close it.. but at the time you open I tried to find a solution and I never came to it.


And as anything... 2021 was a weird year to me... it seems 2 years passed like yesterday.

MrEasy:
Thanks for the update @clebertsuconic - 2020 was yesterday from my feeling ;)
I still consider it a valuable change improving robustness (helped several times in our environments), but totally ok if not merged.

PR #3802, opened 2021-10-19
ARTEMIS-3308 - support federation of large messages

gtully:


clebertsuconic:
I am not merging because I can't make any tests now... if you ran tests and feel comfortable with this.. go for it.

PR #3805, opened 2021-10-20
Artemis-2561: Journal timeout should be configurable

nbrendah:
Issue Link:  https://issues.apache.org/jira/browse/ARTEMIS-2561

Hi @clebertsuconic 
This is how i understood the description and comments on ARTEMIS-2561.

If i am on the right truck, however, i cant find-out the best place to write test cases for this enhancement 

Thank you so much.


clebertsuconic:
@nbrendah the timeout should be configured all the way up to the broker.xml

you need to change the broker.xsd, the parser, the initialization of the broker, the journal, to then pass the parameter to where you added here.

There are a couple of layers between this and the initialization...

it's a simple task in terms of algorithm (there's not much algorithm associated, but you need to understand the initialization of the broker and the parsing of the XML).

nbrendah:
Hello @clebertsuconic 

> you need to change the broker.xsd, the parser, the initialization of the broker, the journal, to then pass the parameter to where you added here.

I have read read this couple of times.  The explanations makes sense but implimentation is still in abstract to my minds.  Maybe its because i am not so conversant with this code base.  

At the same time, i dont want to be some other pain, asking questions that seam obvious.

nbrendah:
Hello @clebertsuconic,  I have added some changes.  I kindly request for review.

jbertram:
I'm closing this as it's been a year with no activity. The current changes do not fit the desired out-come since no additional configuration is possible via `broker.xml`. Feel free to re-open this PR or ask additional questions.

PR #3809, opened 2021-10-21
Register listener

somdatta8172:
Issue Link: https://issues.apache.org/jira/browse/ARTEMIS-3085?filter=12350956

Hello @clebertsuconic 
I have added registerIOCriticalErrorListener method to ActiveMQServer Interface,wrote the method's implementation and added IOCriticalErrorListenerTest also,though need suggestions on it.
Thank you
Somdatta

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServer.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServer.java
@@ -200,7 +200,6 @@ public interface ActiveMQServer extends ServiceComponent {
     *
     * @param listener @see org.apache.activemq.artemis.core.io.IOCriticalErrorListener
     */
-
    void registerIOCriticalErrorListener(IOCriticalErrorListener listener);
 
    void replay(Date start, Date end, String address, String target, String filter) throws Exception;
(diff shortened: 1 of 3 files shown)


clebertsuconic:
Please.. rebase against main:

```
git remote add github https://github.com/apache/activemq-artemis.git
git pull --rebase github main
git push origin register-listener -f
```


a merge commit on a pull request is never good.. you should always have your pull request against main. and if you update against main, please rebase and push -f

clebertsuconic:
You should probably squash all your changes Into a single commit.

Also, you are adding the lister to a collection on ActiveMQServer, but you're not doing anything with it. you should register it to the StorageManager when it starts...

ultimately you should just delegate it from the storage manager.


Also: notice the critical error is meant for IO Storage.. not acceptors.. on your test you did an invalid TCP operation, but that wouldn't cause any damage.

We have some tests that are validating the IOCriticalException by simply calling an artificial method, generating a fake failure:

ShutdownOnCriticalIOErrorMoveNextTest was doing that, but that would be a bit convoluted for your test.


As for the change, you will have to change ActiveMQServerImpl a bit further.

Currently the ActiveMQServerImpl is creating a DefaultCriticalErrorListener that will register into the StorageManagers.

I suggest you add this ioCriticalErrorListener to your collection,

and then create a new Listener that will do the following:

ioCriticalErrorListeners.forEach((t) -> t.callitWhateverTheMethodName);

jbertram:
I went ahead and cleaned this up and merged it since it had been sitting idle for so long. Thanks for the contribution, @nbrendah!

PR #3810, opened 2021-10-22
Artemis 3538 Removing // --------- comments from the codebase

Kulex001:


clebertsuconic:
although those probably can go as well (the ones you did).

but you didn't find any of the // Constants? 

try using 'git grep' on your command line for this. it may be helpful.

PR #3819, opened 2021-10-27
ARTEMIS-3541 createQueue ignoring autoDeleteAddress flag

jbertram:
Commit 481b73c8cabf9d7c88bca048b97bd587661b9c52 from ARTEMIS-3502
inadvertently broke this functionality. This commit restores the
original behavior. When autoDeleteAddress is true then the queue's
address should be automatically deleted regardless of whether or not it
is temporary (assuming it meets all the other criteria).


clebertsuconic:
@jbertram my issue was with the remove of the temporary queue from the equation... it will probably break a test.

I will pick your change and test with temporary queues and address if any issues.

clebertsuconic:
the check on the flag is fine.. it will immediately remove the address only if it's from management.

clebertsuconic:
actually... that's still not correct.. as destroyQueue would make it remove it immediately... I will handle this any way.

PR #3826, opened 2021-10-29
ARTEMIS-3140 Extra options in LDAP login module

pahamala:
Adds support for extra configuration options to LDAP login module to
prepare for supporting any future/custom string configuration in LDAP
directory context creation.

Details:
* Changed LDAPLoginModule to pass any string configuration not recognized by the module itself to the InitialDirContext contruction environment.
* Changed the static LDAPLoginModule configuration key fields to an enum to be able to loop through the specified keys (e.g. to filter out the internal LDAPLoginModule configuration keys from the keys passed to InitialDirContext).
* Few fixes for issues reported by static analysis tools.
* Tested that LDAP authentication with TLS+GSSAPI works  against a recent Windows AD server with Java OpenJDK11U-jdk_x64_windows_hotspot_11.0.13_8 by setting the property com.sun.jndi.ldap.tls.cbtype (see ARTEMIS-3140) in JAAS login.conf.
* Moved LDAPLoginModuleTest to the correct package to be able to access LDAPLoginModule package privates from the test code.
* Added a test to LDAPLoginModuleTest for the task changes.
* Updated documentation to reflect the changes.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/LDAPLoginModule.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/LDAPLoginModule.java
@@ -17,7 +17,6 @@
 package org.apache.activemq.artemis.spi.core.security.jaas;
 
 import javax.naming.AuthenticationException;
-import javax.naming.CommunicationException;
 import javax.naming.Context;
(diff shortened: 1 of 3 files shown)


pahamala:
Thanks @jbertram, glad to help.

PR #3832, opened 2021-11-01
ARTEMIS-3461: add some tests and resolve various issues spotted

gemmellr:
ARTEMIS-3461: add some tests and resolve various issues spotted with the prior changes

- Avoid blowing up on string bodies of any size if the valueSizeLimit bits are configured to disable limit.
- Dont NPE if amqp-value + binary body is sent without a content-type, as it always should be.
- Include expected prefix when adding delivery delay and ingress time annotations.
- Use the actual name for ingress time annotation, as with all other annotations.
- Use correct object type when testing equality with content-type value.
- Use consistent case for 'groupId' in different properties.


clebertsuconic:
... and when I started this task I thought it was going to be just an easy 5 min admin console fix :)


thanks @gemmellr

PR #3833, opened 2021-11-02
ARTEMIS-3541 createQueue ignoring autoDeleteAddress flag

clebertsuconic:
Commit 481b73c8cabf9d7c88bca048b97bd587661b9c52 from ARTEMIS-3502
inadvertently broke this functionality. This commit restores the
original behavior.

autoDeleteAddress was renamed to forceAutoDeleteAddress which will ignore the address settings.

delete temporary queues will use forceAutoDeleteAddress=true.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.java
@@ -1090,7 +1090,7 @@ public interface ActiveMQServerControl {
    @Operation(desc = "Destroy a queue", impact = MBeanOperationInfo.ACTION)
    void destroyQueue(@Parameter(name = "name", desc = "Name of the queue to destroy") String name,
                      @Parameter(name = "removeConsumers", desc = "Remove consumers of this queue") boolean removeConsumers,
-                     @Parameter(name = "autoDeleteAddress", desc = "Automatically delete the address if this was the last queue") boolean autoDeleteAddress) throws Exception;
+                     @Parameter(name = "forceAutoDeleteAddress", desc = "Automatically delete the address if this was the last queue") boolean autoDeleteAddress) throws Exception;
 
    /**
     * Enables message counters for this server.
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServer.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServer.java
@@ -672,13 +672,13 @@ public interface ActiveMQServer extends ServiceComponent {
                      SecurityAuth session,
                      boolean checkConsumerCount,
                      boolean removeConsumers,
-                     boolean autoDeleteAddress) throws Exception;
+                     boolean forceAutoDeleteAddress) throws Exception;
 
    void destroyQueue(SimpleString queueName,
                      SecurityAuth session,
                      boolean checkConsumerCount,
                      boolean removeConsumers,
-                     boolean autoDeleteAddress,
+                     boolean forceAutoDeleteAddress,
                      boolean checkMessageCount) throws Exception;
 
    String destroyConnectionWithSessionMetadata(String metaKey, String metaValue) throws Exception;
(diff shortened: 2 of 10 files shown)


clebertsuconic:
@jbertram this is replacing your previous PR on this

jbertram (approved):

jbertram:
@clebertsuconic, LGTM. Will merge ASAP.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueManagerImpl.java
```diff
@@ -72,7 +72,7 @@ public static void performAutoDeleteQueue(ActiveMQServer server, Queue queue) {
       ActiveMQServerLogger.LOGGER.autoRemoveQueue("" + queue.getName(), queue.getID(), "" + queue.getAddress());
 
       try {
-         server.destroyQueue(queueName, null, true, false, settings.isAutoDeleteAddresses(), true);
+         server.destroyQueue(queueName, null, true, false, false, true);
```
gemmellr, line 75: This almost certainly rendered the settings unused, so they should have been removed as well.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.java
```diff
@@ -1090,7 +1090,7 @@ void destroyQueue(@Parameter(name = "name", desc = "Name of the queue to destroy
    @Operation(desc = "Destroy a queue", impact = MBeanOperationInfo.ACTION)
    void destroyQueue(@Parameter(name = "name", desc = "Name of the queue to destroy") String name,
                      @Parameter(name = "removeConsumers", desc = "Remove consumers of this queue") boolean removeConsumers,
-                     @Parameter(name = "autoDeleteAddress", desc = "Automatically delete the address if this was the last queue") boolean autoDeleteAddress) throws Exception;
+                     @Parameter(name = "forceAutoDeleteAddress", desc = "Automatically delete the address if this was the last queue") boolean autoDeleteAddress) throws Exception;
```
gemmellr, line 1093: This name could be fairly misleading with that doc along with it, which isnt precisely what it does. I would consider it reasonable to think think a 'force' auto deletion described that way would always delete and do it there and then, which it doesnt necessarily. It only deletes due to this if the address was itself auto-created, and only if it has an exactly-0 autoDeleteDelay.

This would again give super-surprising behaviour if auto-delete-delay is also disabled in the broker config generally, as it was leading to finding this issue, since the address would then never be removed (by the reaper), even despite this requested 'forcing'.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServer.java
```diff
@@ -672,13 +672,13 @@ void destroyQueue(SimpleString queueName,
...
                      boolean removeConsumers,
-                     boolean autoDeleteAddress) throws Exception;
+                     boolean forceAutoDeleteAddress) throws Exception;
 
    void destroyQueue(SimpleString queueName,
                      SecurityAuth session,
                      boolean checkConsumerCount,
                      boolean removeConsumers,
-                     boolean autoDeleteAddress,
+                     boolean forceAutoDeleteAddress,
```
gemmellr, line 681: Could use some doc here too to explain what it does.

gemmellr:
PR title and commit message are wrong, as with the last ones. It was destroyQueue.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.java
clebertsuconic: we could call it instantAutoDeleteAddress instead of force.

This is just doing the check on the address instantly as opposed to leave it to the reaper (in the new semantics now).

clebertsuconic: @gemmellr lets move the discussion on the  PR following up your notes please?:

https://github.com/apache/activemq-artemis/pull/3835

PR #3834, opened 2021-11-02
ARTEMIS-2293 addPacket in LargeMessageControllerImpl won't notifyAll …

jbertram:
…for exception


MrEasy:
Thanks for the fast fix 👍🏻

PR #3838, opened 2021-11-05
ARTEMIS-3554 Invalid Prepared Transaction could interrupt server reload

clebertsuconic:


clebertsuconic:
I need to add some logging instead of logger.warn directly...

until then I will keep this as draft.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
```diff
@@ -1236,6 +1238,22 @@ public JournalLoadInformation loadMessageJournal(final PostOffice postOffice,
...
    }
 
+   private void failedToPrepareException(PreparedTransactionInfo txInfo, Throwable e) {
+      XidEncoding encodingXid = null;
+      try {
+          encodingXid = new XidEncoding(txInfo.getExtraData());
+      } catch (Throwable ignored) {
+      }
+
+      ActiveMQServerLogger.LOGGER.warn("Failed to load prepared TX " + encodingXid.xid + ", Prepared Transaction will be rolled back", e);
```
clebertsuconic, line 1248: I will make this as a Logger with a code. Hence the PR is a draft still.

PR #3839, opened 2021-11-05
Artemis 3548: Deprecate offensive terms and add new terms

Kulex001:
Issue link: https://issues.apache.org/jira/projects/ARTEMIS/issues/ARTEMIS-3548

I replaced "master/slave" in internal method with "primary/backup"

@clebertsuconic Is this what you meant?


clebertsuconic:
@Asukwo-comfort you need to rebase against main again...

A Pull Request should always be on top of main.. your pull request is showing conflicts against main..


Anyway... I see that you replace these terms in one file.. which is fine

this is a bit more complex than just search and replace. The FileConfigurationParse will have to still be able to read older terms.. and we will need older configuration in the test suite to make sure older XMLs can still be parsed.


So, it is a bit of a search / replace.. but it will be a long task... not very complex.. but it will require some management to make it happen.

PR #3841, opened 2021-11-05
ARTEMIS-3556 show message protocol on message view page

erwindon:
see https://issues.apache.org/jira/browse/ARTEMIS-3556

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/openmbean/MessageOpenTypeFactory.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/openmbean/MessageOpenTypeFactory.java
@@ -124,7 +124,7 @@ public class MessageOpenTypeFactory<M extends Message> {
    public Map<String, Object> getFields(M m, int valueSizeLimit, int deliveryCount) throws OpenDataException {
       Map<String, Object> rc = new HashMap<>();
       rc.put(CompositeDataConstants.MESSAGE_ID, "" + m.getMessageID());
-      rc.put(CompositeDataConstants.PROTOCOL, m.getClass().getSimpleName());
+      rc.put(CompositeDataConstants.PROTOCOL, m.getProtocolName());
       if (m.getUserID() != null) {
          rc.put(CompositeDataConstants.USER_ID, "ID:" + m.getUserID().toString());
       } else {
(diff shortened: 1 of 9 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/openmbean/MessageOpenTypeFactory.java
```diff
@@ -123,6 +124,7 @@ protected void init() throws OpenDataException {
    public Map<String, Object> getFields(M m, int valueSizeLimit, int deliveryCount) throws OpenDataException {
       Map<String, Object> rc = new HashMap<>();
       rc.put(CompositeDataConstants.MESSAGE_ID, "" + m.getMessageID());
+      rc.put(CompositeDataConstants.PROTOCOL, m.getClass().getSimpleName());
```
michaelandrepearce, line 127: This is a bit dangerous as it exposes class name to api, can we make this something more explicit e.g. add interface to message api such as (getProtocol) and then for the protocol implementation classes to implement that returning a constant value.

michaelandrepearce (changes requested):
see inline comment.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/openmbean/MessageOpenTypeFactory.java
erwindon: That was my original thought, until I saw how self-descriptive the class-names were.
I'll follow the original thought and your suggestion.
Still, I'll make the default implementation getClass().getSimpleName() because there are internal classes and client classes for which the name is never shown.

erwindon: please ignore the remark about getClass().getSimpleName(). there is only one internal class that is not a sub of CoreMessage. so that problem is negligable.

clebertsuconic: we could add a method on Message to return a more user friendly word? 

on Message:
   public String getProtocol();
   
   
   then on core you return "CORE" and AMQP you return "AMQP"
   
   ?

erwindon: I already made that change exactly as you and Michael suggest. that code is in the next commit.

erwindon: maybe I should squash to prevent confusion?

PR #3842, opened 2021-11-08
ARTEMIS-3557 OFF_WITH_REDISTRIBUTION - Add Multicast distribution and…

AntonRoskvist:
… add redistribution to old Anycast consumers


AntonRoskvist:
This is my attempt to fix what we discussed earlier @gtully 

This might not be the optimal solution but works for all intents and purposes for my use case it least. I'd be happy to make changes if needed... if nothing else, then at least the test cases should be good to use for a potentially better solution.

PR #3844, opened 2021-11-10
ARTEMIS-3565 Remove mvnw

clebertsuconic:


clebertsuconic:
These are all auto-generated by calling Takari's integration.

gemmellr:
Given the build hasnt worked with anything less <3.5.0 for quite some time now, and this was using 3.3.9, it doesnt seem like this could be getting much use. Do we actually need this at all?

clebertsuconic:
@gemmellr I believe newbies will use it...

however the issue I'm having is that my IDE was using it when I rebuilt the project.. it was the easiest way to fix my problem.


I don't mind if we keep it.. but if we won't update, we should rather remove it.


Takari will be part of maven 4.. so I think it's ok to keep it anyway.

gemmellr:
Personally I would remove it. I've never seen a great need for it, and clearly it isnt been getting used of late. A requirement on Maven 3.5.0 doesnt seem particularly onerous.

We aren't allowed to distribute the jar in the source archive, so I presume its already being excluded there. I do seem to recall some other projects writing their own bootstrap for this to avoid such hoops.

clebertsuconic:
@gemmellr fair enough.. I will remove it instead...

It was nice to see someone contributing it back in 2016.. but I don't see much use for it anyways.

clebertsuconic:
PR and JIRA updated to remove mvnw instead of updating it.

gemmellr:
Looks good.

clebertsuconic:
merging it as I need this on another task I'm working on...

PR #3846, opened 2021-11-11
ARTEMIS-3546 Jakarta javax.json incompability on Jakarta all client

clebertsuconic:


wilkinsona:
Thanks for taking a look at this, @clebertsuconic. While the changes appear to fix the minimal reproducer I provided in ARTEMIS-3546, I don't think they're sufficient to fully address the problem. The `-all` artifact still contains shaded Apache Johnzon classes that refer to `javax.json` and a `/META-INF/services/javax.json.spi.JsonProvider` file as well. Furthermore, switching to the `artemis-jakarta-client` artifact shows that it still has transitive dependencies on `jakarta.json-api-1.1.6` and `johnzon-core-0.9.5` which are EE 8 artifacts.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/JSonUtil.java
```diff
@@ -0,0 +1,33 @@
...
+
+package org.apache.activemq.artemis.core.management.impl;
+
+import javax.json.JsonObject;
+
+import org.apache.activemq.artemis.api.core.JsonUtil;
+import org.apache.activemq.artemis.api.core.TransportConfiguration;
+import org.apache.activemq.artemis.utils.JsonLoader;
+
+public class JSonUtil {
```
wilkinsona, line 26: Typo? Should this be named `JsonUtil` perhaps?

clebertsuconic: You're right.  Thanks.

clebertsuconic:
I thought it would been ok as those are only used with management operations.

wilkinsona:
How would someone use those management operations in a Jakarta EE 9 environment, be that an application server or in a "standalone" app that's using EE 9-level dependencies?

clebertsuconic:
those are server only operations...

I'm going a step further and remove json from the client.

wilkinsona:
Great, thanks. For Spring Boot's purposes, we'll also need the server to be completely EE 9 compatible with no EE 8 dependencies. Would you prefer to tackle that separately?

clebertsuconic:
@wilkinsona even the server's? are you guys embedding the server?

you could just use artemis standalone as is.. and leave the dependency be whatever in the standalone server.


those dependencies leaked to the client for server's management purposes on toJSon and fromJson methods.. these don't need to be on any client.



To be honest I never though on Json as being EE9.. we just need a string exposed in certain places. We used javax just as a way to have the dependency indirect on the API.

wilkinsona:
We support both. A standalone server is the default, but we also support configuring an embedded Artemis broker when the user adds the server dependency to the classpath: https://docs.spring.io/spring-boot/docs/2.5.6/reference/htmlsingle/#messaging.jms.artemis. This won't work with EE 9 dependencies as Spring Boot's dependency management will ensure that the `jakarta.json.*` classes are on the classpath while Artemis will be looking for `javax.json`.

clebertsuconic:
@wilkinsona I will get rid of javax.json everywhere and just json directly... 

and I thought this was a 2 hours job :)....

clebertsuconic:
there's no directly json actually :/

clebertsuconic:
I need help with Karaf here

jbonofre:
@clebertsuconic I will tackle the Karaf part.

artemis-commons/pom.xml
```diff
@@ -53,6 +71,7 @@
       <dependency>
           <groupId>com.google.errorprone</groupId>
           <artifactId>error_prone_core</artifactId>
+         <scope>provided</scope>
```
gemmellr, line 74: Did this make a difference? The dependencyManagement is setting the scope so it should already be at an effective provided. (Mainly just curious why you actually changed it, personally I do prefer each module states it)

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/JsonUtil.java
```diff
@@ -16,13 +16,13 @@
...
-import javax.json.JsonObjectBuilder;
-import javax.json.JsonString;
-import javax.json.JsonValue;
+import org.apache.activemq.artemis.commons.json.JsonArray;
+import org.apache.activemq.artemis.commons.json.JsonArrayBuilder;
+import org.apache.activemq.artemis.commons.json.JsonNumber;
+import org.apache.activemq.artemis.commons.json.JsonObject;
+import org.apache.activemq.artemis.commons.json.JsonObjectBuilder;
+import org.apache.activemq.artemis.commons.json.JsonString;
+import org.apache.activemq.artemis.commons.json.JsonValue;
```
gemmellr, line 25: Just to be sure its intended/realised...noting this is presumably a breaking change for all existing management related usage of these bits. EDIT: As the later test changes show.

artemis-commons/pom.xml
clebertsuconic: I was playing between provided and compile..

The issue with compiled is shade will not pick it up.

Shade will generate another PR, but maven will not look at it internally. It's a bit of a mess I'm not being able to figure out.

artemis-core-client-osgi/pom.xml
```diff
@@ -35,6 +35,17 @@
...
+         <exclusions>
+            <exclusion>
+               <groupId>org.apache.johnzon</groupId>
+               <artifactId>johnzon-core</artifactId>
+            </exclusion>
+            <exclusion>
+               <groupId>jakarta.json</groupId>
+               <artifactId>jakarta.json-api</artifactId>
+            </exclusion>
+         </exclusions>
```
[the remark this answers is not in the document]
clebertsuconic: Yes.. I was trying to play with provided and it was not working.....  shade was ignoring it.
I'm still struggling what to do here.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/JsonUtil.java
clebertsuconic: Json is only used internally... we never (should not) expose through the API.

artemis-core-client-osgi/pom.xml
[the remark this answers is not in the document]
gemmellr: compile scope + optional is another option to not pass deps on transitively.

artemis-commons/pom.xml
gemmellr: So yes, I see a comment in the netty codebase that shading needs compile scope. Plus it does seem that the shading actually removes the dep, so....no idea why you are seeing issues.

artemis-core-client-osgi/pom.xml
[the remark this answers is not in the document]
gemmellr: So yes, I see a comment in the netty codebase that shading needs compile scope. Plus it does seem that the shading actually removes the dep, so....no idea why you are seeing issues.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/JsonUtil.java
gemmellr: Fair enough, just checking that was the position for when anyone comes asking about it.

gemmellr: Another consideration in that case might be to remove (or move to a clearly-internal package) this class if its not considered API (which might seem odd given its in an "api" sub package)

clebertsuconic:
@jbonofre / @gtully the only thing missing here now is the feature failure.

At the moment the module is commented out on the main pom so just I could run some tests.. but if you enable it, you will get failures:


```
[ERROR] 	Feature resolution failed for [artemis-openwire/2.20.0.SNAPSHOT]
[ERROR] Message: Unable to resolve root: missing requirement [root] osgi.identity; osgi.identity=artemis-openwire; type=karaf.feature; version=2.20.0.SNAPSHOT; filter:="(&(osgi.identity=artemis-openwire)(type=karaf.feature)(version>=2.20.0.SNAPSHOT))" [caused by: Unable to resolve artemis-openwire/2.20.0.SNAPSHOT: missing requirement [artemis-openwire/2.20.0.SNAPSHOT] osgi.identity; osgi.identity=org.apache.activemq.artemis-openwire-protocol; type=osgi.bundle; version="[2.20.0.SNAPSHOT,2.20.0.SNAPSHOT]"; resolution:=mandatory [caused by: Unable to resolve org.apache.activemq.artemis-openwire-protocol/2.20.0.SNAPSHOT: missing requirement [org.apache.activemq.artemis-openwire-protocol/2.20.0.SNAPSHOT] osgi.wiring.package; filter:="(osgi.wiring.package=org.apache.activemq.artemis.api.core.client)" [caused by: Unable to resolve org.apache.activemq.artemis-server-osgi/2.20.0.SNAPSHOT: missing requirement [org.apache.activemq.artemis-server-osgi/2.20.0.SNAPSHOT] osgi.wiring.package; filter:="(osgi.wiring.package=org.glassfish.json)"]]]
[ERROR] Repositories: {
[ERROR] 	file:/Users/clebertsuconic/work/apache/activemq-artemis/artemis-features/target/classes/features.xml
[ERROR] 	mvn:org.apache.aries.jpa/jpa-features/2.7.3/xml/features
[ERROR] 	mvn:org.apache.karaf.features/enterprise/4.3.3/xml/features

```

gemmellr:
I tried poking at it and came up with some changes that pass the PR build+tests.

It is failing because the -osgi modules have an entry in their manifest saying they should import the org.glashfish.json package, which the build (in disabled module, now enabled) and karaf tests module cant then resolve. I dont really understand that since nothing seems to reference that package to cause it to be imported, and it doesnt look like the deps which were shaded and removed actually provided it anyway, so its unclear to me why that behaviour arises or would now differ.

I just blindly changed it so that particular package is listed as optional, meaning it will only fail if something needing it is actually used and its not present at the time. Things seem happy with that along with some other cleanup to remove imports/deps that it seems the shading should be removing need of. With all these the build works and the karaf client integration tests pass. Not sure about the broker module, its not clear its tested, but since nothing obviously uses the package and the real change is all in artemis-common it seems like it should be much the same.

I added the changes on a branch in my fork https://github.com/gemmellr/activemq-artemis/tree/json-fixup-playing-ARTEMIS-3546, or you can find the current delta diff below.

<details><summary>Delta diff from prior changes (click to expand)</summary><p>

```
diff --git a/artemis-commons/pom.xml b/artemis-commons/pom.xml
index 1b0fe2442f..f40497c79c 100644
--- a/artemis-commons/pom.xml
+++ b/artemis-commons/pom.xml
@@ -39,11 +39,15 @@
          <groupId>org.apache.johnzon</groupId>
          <artifactId>johnzon-core</artifactId>
          <version>${johnzon.version}</version>
+         <scope>compile</scope>
+         <optional>true</optional>
       </dependency>
       <dependency>
          <groupId>jakarta.json</groupId>
          <artifactId>jakarta.json-api</artifactId>
          <version>${jakarta.json-api.version}</version>
+         <scope>compile</scope>
+         <optional>true</optional>
          <!-- License: EPL 2.0 -->
       </dependency>
       <!-- ^^ these dependencies are shaded on this JAR -->
@@ -69,7 +73,6 @@
       <dependency>
           <groupId>com.google.errorprone</groupId>
           <artifactId>error_prone_core</artifactId>
-         <scope>provided</scope>
       </dependency>
       <dependency>
           <groupId>org.wildfly.common</groupId>
@@ -127,7 +130,6 @@
          <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-shade-plugin</artifactId>
-            <version>3.2.4</version>
             <executions>
                <execution>
                   <phase>package</phase>
diff --git a/artemis-core-client-osgi/pom.xml b/artemis-core-client-osgi/pom.xml
index b84ba94f7b..1fcfd777a4 100644
--- a/artemis-core-client-osgi/pom.xml
+++ b/artemis-core-client-osgi/pom.xml
@@ -98,8 +98,8 @@
                <instructions>
                   <Embed-Dependency>*;scope=compile|runtime;groupId=org.apache.activemq</Embed-Dependency>
                   <Import-Package>
+                     org.glassfish.json*;resolution:=optional,
                      io.netty.buffer;io.netty.*;version="[4.1,5)",
-                     org.apache.johnzon.core,
                      *
                   </Import-Package>
                   <_exportcontents>org.apache.activemq.artemis.*;-noimport:=true</_exportcontents>
diff --git a/artemis-features/src/main/resources/features.xml b/artemis-features/src/main/resources/features.xml
index 5de867530f..d5894b1fc3 100644
--- a/artemis-features/src/main/resources/features.xml
+++ b/artemis-features/src/main/resources/features.xml
@@ -53,9 +53,6 @@
 
 		<bundle dependency="true">mvn:org.jboss.logging/jboss-logging/${jboss.logging.version}</bundle>
 		<bundle dependency="true">mvn:org.jgroups/jgroups/${jgroups.version}</bundle>
-
-		<bundle dependency="true">mvn:org.apache.servicemix.specs/org.apache.servicemix.specs.json-api-1.1/${servicemix.json-1.1.spec.version}</bundle>
-		<bundle>mvn:org.apache.johnzon/johnzon-core/${johnzon.version}</bundle>
 	</feature>
 
 	<feature name="artemis-core" version="${pom.version}" description="ActiveMQ Artemis broker libraries">
diff --git a/artemis-jms-client-osgi/pom.xml b/artemis-jms-client-osgi/pom.xml
index 6a43150ecb..e95708fb3e 100644
--- a/artemis-jms-client-osgi/pom.xml
+++ b/artemis-jms-client-osgi/pom.xml
@@ -107,6 +107,7 @@
                <instructions>
                   <Embed-Dependency>*;scope=compile|runtime;groupId=org.apache.activemq</Embed-Dependency>
                   <Import-Package>
+                     org.glassfish.json*;resolution:=optional,
                      io.netty.buffer;io.netty.*;version="[4.1,5)",
                      *
                   </Import-Package>
diff --git a/artemis-server-osgi/pom.xml b/artemis-server-osgi/pom.xml
index b6dd8e49c8..8a60d7548f 100644
--- a/artemis-server-osgi/pom.xml
+++ b/artemis-server-osgi/pom.xml
@@ -147,6 +147,7 @@
                   <Embed-Dependency>*;scope=compile|runtime;groupId=org.apache.activemq</Embed-Dependency>
                   <Import-Package>
                      org.postgresql*;resolution:=optional,
+                     org.glassfish.json*;resolution:=optional,
                      io.netty.buffer;io.netty.*;version="[4.1,5)",
                      *
                   </Import-Package>
diff --git a/pom.xml b/pom.xml
index 1ee3834ace..8a539d738c 100644
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
       <module>integration/activemq-spring-integration</module>
       <module>artemis-distribution</module>
       <module>tests</module>
-      <!-- <module>artemis-features</module> -->
+      <module>artemis-features</module>
       <module>artemis-quorum-api</module>
       <module>artemis-quorum-ri</module>
    </modules>
@@ -105,6 +105,7 @@
       <errorprone.version>2.9.0</errorprone.version>
       <maven.enforcer.plugin.version>3.0.0-M3</maven.enforcer.plugin.version>
       <maven.bundle.plugin.version>5.1.2</maven.bundle.plugin.version>
+      <maven.shade.plugin.version>3.2.4</maven.shade.plugin.version>
       <maven.checkstyle.plugin.version>3.1.1</maven.checkstyle.plugin.version>
       <sevntu.checks.version>1.39.0</sevntu.checks.version>
       <checkstyle.version>8.43</checkstyle.version>
@@ -1428,7 +1429,7 @@
             <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-shade-plugin</artifactId>
-               <version>3.2.0</version>
+               <version>${maven.shade.plugin.version}</version>
             </plugin>
             <plugin>
                <groupId>org.eclipse.jetty</groupId>
```
</p></details>

clebertsuconic:
@gemmellr wow.. thank you so much for this...


I have pulled your changes in on a separate commit for now...


The only thing I did not do was the optional aspect of the library.

Maven will internally still use the unshaded jar for certain features, and when that happens the dependency still valid, but when using outside of the context of artemis build the shaded version will not have the dependency. I checked that and it's working...

(I needed the dependency for certain tests)


the secret is not not depend on the shaded classes for anything.. I had to pull in @brusdev 's wrapper on JSON, and everything works now.


I will do some reorg on the commits and squash accordingly before I can mark this ready for merge, but all the code changes should be here now.

gemmellr:
> The only thing I did not do was the optional aspect of the library.
> 
> Maven will internally still use the unshaded jar for certain features, and when that happens the dependency still valid, but when using outside of the context of artemis build the shaded version will not have the dependency. I checked that and it's working...
> 
> (I needed the dependency for certain tests)
> 

You dropped the plugin version removal change in the commons pom too, that should be fixed

I would suggest to try and leave them optional, and e.g. add the deps back as test scope deps if still needed in places to run tests using the wrapper, due to maven running them against the original classes rather than the shaded ones in the packaged jar.

As you showed in your original efforts, leaving it this way would mean other modules can still compile with the very changes present that this is aiming to prevent, something that likely wouldnt be picked up until after it was released. At least with them optional they cant compile. If adding them as test deps where needed works then at least they arent hard compile time deps except for the tests which is much less of an issue, and you can be sure no hard usages are added in other modules unless they themselves already depend on the bits.

gemmellr:
Well that was good timing lol.

gemmellr:
Side note that the Java 11 run is the only one running the Karaf based tests (but failed before getting to them this time), so dont use the 17 run as an indication of success in this particular case: https://issues.apache.org/jira/browse/ARTEMIS-3410

clebertsuconic:
@gemmellr internally in maven tests are pulling the unshaded classes. I will have to add the dependency f)r tests at least...

this is not easy !

gemmellr:
Yep, maven tests classes since the package phase is seperate and only comes later, and alas shading is at package time.

I'm just saying given the two choices, if using compile+optional in commons with test scope deps elsewhere as needed does works, then I would probably choose that unless there is some larger downside, otherwise things just remain susceptible to this same compatibility problem arising again by accident again without being noticed.

clebertsuconic:
I actually had to add optional because the unshaded bits were leaking into Artemis-jakarta-client-all and other shaded jars that we distribute.

artemis-commons/pom.xml
```diff
@@ -111,6 +127,46 @@
                </execution>
             </executions>
          </plugin>
+         <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-shade-plugin</artifactId>
+            <version>3.2.4</version>
+            <executions>
```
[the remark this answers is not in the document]
clebertsuconic: thanks

PR #3849, opened 2021-11-12
ARTEMIS-3569 - balancer role_name local target, matches role of authe…

gtully:
…nticated user

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
@@ -211,12 +211,13 @@ public class AMQPSessionCallback implements SessionCallback {
                                                            true, //boolean xa,
                                                            (String) null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain());
       } else {
+         final String validatedUser = manager.getServer().validateUser(user, passcode, protonSPI.getProtonConnectionDelegate(), manager.getSecurityDomain());
          serverSession = manager.getServer().createSession(name, user, passcode, ActiveMQClient.DEFAULT_MIN_LARGE_MESSAGE_SIZE, protonSPI.getProtonConnectionDelegate(), // RemotingConnection remotingConnection,
                                                            false, // boolean autoCommitSends
                                                            false, // boolean autoCommitAcks,
                                                            false, // boolean preAcknowledge,
                                                            true, //boolean xa,
-                                                           (String) null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain());
+                                                           (String) null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain(), validatedUser);
       }
    }
 
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
@@ -58,7 +58,6 @@ import org.apache.activemq.artemis.protocol.amqp.proton.AmqpSupport;
 import org.apache.activemq.artemis.protocol.amqp.proton.ProtonServerReceiverContext;
 import org.apache.activemq.artemis.protocol.amqp.proton.ProtonServerSenderContext;
 import org.apache.activemq.artemis.protocol.amqp.proton.transaction.ProtonTransactionHandler;
-import org.apache.activemq.artemis.protocol.amqp.sasl.PlainSASLResult;
 import org.apache.activemq.artemis.protocol.amqp.sasl.SASLResult;
 import org.apache.activemq.artemis.spi.core.protocol.SessionCallback;
 import org.apache.activemq.artemis.spi.core.remoting.Connection;
@@ -194,30 +193,20 @@ public class AMQPSessionCallback implements SessionCallback {
 
       String name = UUIDGenerator.getInstance().generateStringUUID();
 
-      String user = null;
-      String passcode = null;
-      if (saslResult != null) {
-         user = saslResult.getUser();
-         if (saslResult instanceof PlainSASLResult) {
-            passcode = ((PlainSASLResult) saslResult).getPassword();
-         }
-      }
-
       if (connection.isBridgeConnection())  {
          serverSession = manager.getServer().createInternalSession(name, ActiveMQClient.DEFAULT_MIN_LARGE_MESSAGE_SIZE, protonSPI.getProtonConnectionDelegate(), // RemotingConnection remotingConnection,
                                                            false, // boolean autoCommitSends
                                                            false, // boolean autoCommitAcks,
                                                            false, // boolean preAcknowledge,
                                                            true, //boolean xa,
-                                                           (String) null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain());
+                                                           null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain());
       } else {
-         final String validatedUser = manager.getServer().validateUser(user, passcode, protonSPI.getProtonConnectionDelegate(), manager.getSecurityDomain());
-         serverSession = manager.getServer().createSession(name, user, passcode, ActiveMQClient.DEFAULT_MIN_LARGE_MESSAGE_SIZE, protonSPI.getProtonConnectionDelegate(), // RemotingConnection remotingConnection,
+         serverSession = manager.getServer().createSession(name, connection.getUser(), connection.getPassword(), ActiveMQClient.DEFAULT_MIN_LARGE_MESSAGE_SIZE, protonSPI.getProtonConnectionDelegate(), // RemotingConnection remotingConnection,
                                                            false, // boolean autoCommitSends
                                                            false, // boolean autoCommitAcks,
                                                            false, // boolean preAcknowledge,
                                                            true, //boolean xa,
-                                                           (String) null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain(), validatedUser);
+                                                           null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain(), connection.getValidatedUser());
       }
    }
 
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
@@ -197,6 +197,8 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
 
    private final AtomicBoolean disableTtl = new AtomicBoolean(false);
 
+   private String validatedUser = null;
+
    public OpenWireConnection(Connection connection,
                              ActiveMQServer server,
                              OpenWireProtocolManager openWireProtocolManager,
@@ -210,6 +212,7 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
       this.outWireFormat = wf.copy();
       this.useKeepAlive = openWireProtocolManager.isUseKeepAlive();
       this.maxInactivityDuration = openWireProtocolManager.getMaxInactivityDuration();
+      this.transportConnection.setProtocolConnection(this);
    }
 
    // SecurityAuth implementation
@@ -768,7 +771,7 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
    }
 
    private void createInternalSession(ConnectionInfo info) throws Exception {
-      internalSession = server.createSession(UUIDGenerator.getInstance().generateStringUUID(), context.getUserName(), info.getPassword(), -1, this, true, false, false, false, null, null, true, operationContext, protocolManager.getPrefixes(), protocolManager.getSecurityDomain());
+      internalSession = server.createSession(UUIDGenerator.getInstance().generateStringUUID(), context.getUserName(), info.getPassword(), -1, this, true, false, false, false, null, null, true, operationContext, protocolManager.getPrefixes(), protocolManager.getSecurityDomain(), validatedUser);
    }
 
    //raise the refCount of context
@@ -1010,6 +1013,14 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
       return protocolManager.isSupportAdvisory();
    }
 
+   public String getValidatedUser() {
+      return validatedUser;
+   }
+
+   public void setValidatedUser(String validatedUser) {
+      this.validatedUser = validatedUser;
+   }
+
    class SlowConsumerDetection implements SlowConsumerDetectionListener {
 
       @Override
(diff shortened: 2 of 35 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ActiveMQPacketHandler.java
```diff
@@ -175,6 +171,11 @@ private void handleCreateSession(final CreateSessionMessage request) {
             activeMQPrincipal = connection.getDefaultActiveMQPrincipal();
          }
 
+         if (connection.getTransportConnection().getRedirectTo() != null) {
+            server.validateUser(activeMQPrincipal == null ? request.getUsername() : activeMQPrincipal.getUserName(), activeMQPrincipal == null ? request.getPassword() : activeMQPrincipal.getPassword(), connection, protocolManager.getSecurityDomain());
```
brusdev, line 175: Why not anticipating server.validateUser and passing the validatedUser parameter to createSession to avoid the double execution?

pom.xml
```diff
@@ -206,7 +206,7 @@
 
       <activemq-surefire-argline>-Dorg.apache.commons.logging.Log=org.apache.activemq.artemis.logs.JBossLoggingApacheLoggerBridge -Dorg.apache.activemq.artemis.utils.RetryRule.retry=${retryTests} -Dbrokerconfig.maxDiskUsage=100 -Dorg.apache.activemq.artemis.core.remoting.impl.netty.TransportConstants.DEFAULT_QUIET_PERIOD=0 -Dorg.apache.activemq.artemis.core.remoting.impl.netty.TransportConstants.DEFAULT_SHUTDOWN_TIMEOUT=0 -Djava.util.logging.manager=org.jboss.logmanager.LogManager
          -Dlogging.configuration="file:${activemq.basedir}/tests/config/${logging.config}"
-         -Djava.library.path="${activemq.basedir}/target/bin/lib/linux-x86_64:${activemq.basedir}/target/bin/lib/linux-i686" -Djgroups.bind_addr=localhost -Dorg.apache.activemq.artemis.api.core.UDPBroadcastEndpointFactory.localBindAddress=localhost
+         -Djava.library.path="${activemq.basedir}/target/bin/lib/linux-x86_64:${activemq.basedir}/target/bin/lib/linux-i686" -Djgroups.bind_addr=localhost -Dorg.apache.activemq.artemis.api.core.UDPBroadcastEndpointFactory.localBindAddress=0.0.0.0
```
brusdev, line 209: why do you need this change?

gtully: that is so that the udp tests will run on my machine

gtully: loopback on osx cannot do broadcast

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -1126,10 +1127,12 @@ public AMQConnectionContext getContext() {
       @Override
       public Response processAddConnection(ConnectionInfo info) throws Exception {
          try {
-            if (transportConnection.getRedirectTo() != null && protocolManager.getRedirectHandler()
-               .redirect(OpenWireConnection.this, info)) {
-               shutdown(true);
-               return null;
+            if (transportConnection.getRedirectTo() != null) {
+               protocolManager.validateUser(OpenWireConnection.this, info);
```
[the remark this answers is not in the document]
gtully: let me check this and see if it can be removed. thanks!

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ActiveMQPacketHandler.java
gtully: that makes sense, create session is called in a few places, but preceding those with validate user is probably fine.
I think there is a body of work to deal with the sasl subject and probably replace a validatedUser string with a subject at some stage... but that is for another day.

pom.xml
gtully: Something has changed since a rebase, possibly jgroups. I now need to unset that system property to have the clustered/discovery tests run!

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -768,7 +769,8 @@ public AMQConnectionContext initContext(ConnectionInfo info) throws Exception {
    }
 
    private void createInternalSession(ConnectionInfo info) throws Exception {
-      internalSession = server.createSession(UUIDGenerator.getInstance().generateStringUUID(), context.getUserName(), info.getPassword(), -1, this, true, false, false, false, null, null, true, operationContext, protocolManager.getPrefixes(), protocolManager.getSecurityDomain());
+      final String validatedUser = server.validateUser(context.getUserName(), info.getPassword(), this, protocolManager.getSecurityDomain());
```
brusdev, line 772: if OpenWireProtocolManager.validateUser could set a new field as OpenWireConnection.validatedUser the OpenWireConnection.createInternalSession shouldn't execute server.validateUser again

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
```diff
@@ -133,7 +133,8 @@ public void initialize() {
       // now
 
       try {
-         coreSession = server.createSession(name, username, password, minLargeMessageSize, connection, true, false, false, false, null, this, true, connection.getOperationContext(), protocolManager.getPrefixes(), protocolManager.getSecurityDomain());
+         final String validatedUser = server.validateUser(username, password, connection, protocolManager.getSecurityDomain());
```
brusdev, line 136: if OpenWireProtocolManager.validateUser could set a new field as OpenWireConnection.validatedUser the AMQSession.initialize shouldn't execute server.validateUser again

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -1674,12 +1674,8 @@ public ServerSession createSession(final String name,
...
                                       final OperationContext context,
                                       final Map<SimpleString, RoutingType> prefixes,
-                                      final String securityDomain) throws Exception {
-      String validatedUser = "";
-
-      if (securityStore != null) {
-         validatedUser = securityStore.authenticate(username, password, connection, securityDomain);
-      }
+                                      final String securityDomain,
+                                      String validatedUser) throws Exception {
```
brusdev, line 1678: cool anticipating the authentication, if securityStore != null and validatedUser == null we could call `securityStore.authenticate` or thwroing a security exception

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/BrokerBalancerControlImpl.java
```diff
@@ -59,11 +59,10 @@ public BrokerBalancerControlImpl(final BrokerBalancer balancer, final StorageMan
...
    @Override
    public CompositeData getTarget(String key) throws Exception {
-      Target target = balancer.getTarget(key);
-
-      if (target != null) {
+      Target.Result result = balancer.getTarget(key);
+      if (Target.Status.OK.compareTo(result.status) == 0) {
          CompositeData connectorData = null;
-         TransportConfiguration connector = target.getConnector();
+         TransportConfiguration connector = result.target.getConnector();
```
brusdev, line 65: It looks a bit weird to me `Target.Result result...` and `result.target...`, why an inner class Target.Result and not another class TragetResult that includes Target and Status ?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
```diff
@@ -211,12 +211,13 @@ public void init(AMQPSessionContext protonSession, SASLResult saslResult) throws
                                                            true, //boolean xa,
                                                            (String) null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain());
       } else {
+         String validatedUser = manager.getServer().validateUser(user, passcode, protonSPI.getProtonConnectionDelegate(), manager.getSecurityDomain());
```
gemmellr, line 214: Why did this become necessary? Auth should already have happened by now, seems messy doing it again.

EDIT: so essentially the same comment as Domenicos in other areas.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/BrokerBalancerControlImpl.java
gemmellr: Perhaps similar wondering about a TargetResult class when considering the 'context.getResult()' usage in AMQPRedirectHandler . It would be nice if the status checking added using that was being done against the actual [Target.]Status enum, rather than needing to do object equality check against a full not-really-Target object. You could instead have a TargetResult holder indicate the result with the enum (and/or boolean conversion method checking for success) and not have any actual Target object.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/MQTTRedirectTest.java
```diff
@@ -121,5 +137,38 @@ public void testSimpleRedirect() throws Exception {
...
+
+      MqttConnectOptions connOpts = new MqttConnectOptions();
+      connOpts.setCleanSession(true);
+      connOpts.setUserName("a");
+      connOpts.setPassword("a".toCharArray());
+
+      MqttClient client0 = new MqttClient("tcp://" + TransportConstants.DEFAULT_HOST + ":" + TransportConstants.DEFAULT_PORT, "TEST", new MemoryPersistence());
+      try {
+         client0.connect(connOpts);
+         fail("Expect to be rejected as not in role0");
```
gemmellr, line 158: role b?

(Also, earlier tests used "B" rather than "b", consistency might be nice)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/TargetKeyTest.java
```diff
@@ -174,6 +194,43 @@ public void testUserNameKey() throws Exception {
...
+      startServers(0);
+
+      final int noRetriesSuchThatWeGetAnErrorOnRejection = 0;
+      ConnectionFactory connectionFactory = createFactory(protocol, false, TransportConstants.DEFAULT_HOST,
+                                                          TransportConstants.DEFAULT_PORT + 0, null, "a", "a", noRetriesSuchThatWeGetAnErrorOnRejection);
+
+      // expect disconnect/reject as not role b
+      try (Connection connection = connectionFactory.createConnection()) {
+         connection.start();
+         fail("Expect to be rejected as not in role0");
```
gemmellr, line 220: role b (/B)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/TargetKeyTest.java
```diff
@@ -174,6 +194,43 @@ public void testUserNameKey() throws Exception {
...
+      } catch (Exception rejected) {
+         rejected.printStackTrace();
```
gemmellr, line 222: Maybe log something succinct instead? We already have enough output during tests without stacks that are entirely expected and not used to check anything.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/BrokerBalancerControlImpl.java
gtully: agree, I will replace that inner class with a separate class

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/TargetKeyTest.java
gtully: good catch, thanks.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
gtully: it should not be duplicated, just earlier. I need to investigate the duplication

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
gtully: I think that should go via server.validateUser, it seems securityStore == null only happens on shutdown. But best to utilise the same method as other protocol heads.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
gtully: that is wise. thanks

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
gtully: agree

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/MQTTRedirectTest.java
gtully: I am not seeing the "B" but for sure role0 is incorrect. thanks!

gemmellr: TargetKeyResolverTest and AmqpRedirectTest used role "B"

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
gtully: The same call to validateUser has been remove from manager.getServer().createSession..

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/balancing/targets/TargetResult.java
```diff
@@ -0,0 +1,42 @@
...
+   public Status status;
+   public Target target;
+
+   public TargetResult(Target t) {
+      this.target = t;
+      this.status = Status.OK;
+   }
+
+   private TargetResult(Status s) {
+      this.status = s;
```
gemmellr, line 34: Probably shouldnt allow 'OK' status this way.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/balancing/targets/TargetResult.java
```diff
@@ -0,0 +1,42 @@
...
+
+package org.apache.activemq.artemis.core.server.balancing.targets;
+
+public class TargetResult {
+
+   public static final TargetResult REFUSED_UNAVAILABLE_RESULT = new TargetResult(Status.REFUSED_UNAVAILABLE);
+   public static final TargetResult REFUSED_USE_ANOTHER_RESULT = new TargetResult(Status.REFUSED_USE_ANOTHER);
+
+   public Status status;
+   public Target target;
```
gemmellr, line 26: Should be private, the constants shouldnt be easily mutable.

gtully:
@brusdev thanks for that amqp contribution to address @gemmellr feedback, it is indeed better. You bet me to it!

PR #3851, opened 2021-11-15
ARTEMIS-3573 Support PropertiesLoginModule custom password codecs

brusdev:


docs/user-manual/en/masking-passwords.md
```diff
@@ -257,7 +257,26 @@ codec other than the default one. For example
...
 to be in masked form.
 
-### login.config
+### PropertiesLoginModule
+Artemis supports Properties login module to be configured in JAAS configuration file
+(default name is `login.config`). By default, the passwords of the users are in plain text
+or masked with the DefaultSensitiveStringCodec.
+
+Set the `org.apache.activemq.jaas.properties.password.codec` property to set a custom codec,
+i.e. to use the `org.apache.activemq.artemis.tests.integration.security.MD5SensitiveDataCodec` codec
```
gemmellr, line 266:
```suggestion
To use a custom codec class, set the `org.apache.activemq.jaas.properties.password.codec` property to the class name
e.g. to use the `com.example.MySensitiveDataCodecImpl` codec class:
```
Related to other comment, it definitely seems like we shouldnt be adding docs pointing to a class using MD5. Both since there seems little need using a real class (which is only part of the tests) at all, and since inevitably someone might actually go find it and just use what its doing.

Might also be worth linking to the section on custom codecs, since it wasnt covered yet and comes later.

docs/user-manual/en/masking-passwords.md
```diff
@@ -257,7 +257,26 @@ codec other than the default one. For example
...
+
+```
+PropertiesLoginWithPasswordCodec {
+    org.apache.activemq.artemis.spi.core.security.jaas.PropertiesLoginModule required
+        debug=true
+        org.apache.activemq.jaas.properties.user="users.properties"
+        org.apache.activemq.jaas.properties.role="roles.properties"
+        org.apache.activemq.jaas.properties.password.codec="org.apache.activemq.artemis.tests.integration.security.MD5SensitiveDataCodec";
```
gemmellr, line 274:
```suggestion
        org.apache.activemq.jaas.properties.password.codec="com.example.MySensitiveDataCodecImpl";
```

docs/user-manual/en/masking-passwords.md
```diff
@@ -462,6 +485,12 @@ public class MyCodec implements SensitiveDataCodec<String> {
    public void init(Map<String, String> params) {
       // Initialization done here. It is called right after the codec has been created.
    }
+
+   @Override
+   public boolean compare(Object value, Object encodedValue) {
+      // Return true if the value matches the encodedValue.
+      return true;
```
gemmellr, line 492:
```suggestion
      // Return true if the value matches the encodedValue.
      return checkValueMatchesEncoding(value, encodedValue);
```

Just setting return true seems like a bad example, this at least spells it out.

gemmellr (commented):
The change looks to have broken some other tests, possibly from updating a shared config.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodec.java
```diff
@@ -79,6 +79,21 @@ public void init(Map<String, String> params) throws Exception {
...
+      if (encodedValue instanceof String == false) {
+         throw new IllegalArgumentException("Not supported encodedValue type: " + encodedValue.getClass().getName());
+      }
+
+      if (value instanceof String) {
+         return verify(((String)value).toCharArray(), (String)encodedValue);
+      } else if (value instanceof char[]) {
+         return verify((char[])value, (String)encodedValue);
+      } else {
+         throw new IllegalArgumentException("Not supported value type: " + value.getClass().getName());
```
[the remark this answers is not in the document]
brusdev: I'll replace `Not supported ` with `Unsupported`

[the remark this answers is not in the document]
brusdev: I would not add another condition to check if value is null just to throw NPE

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodec.java
```diff
@@ -79,6 +79,21 @@ public void init(Map<String, String> params) throws Exception {
       }
    }
 
+   @Override
+   public boolean compare(Object value, Object encodedValue) {
+      if (encodedValue instanceof String == false) {
+         throw new IllegalArgumentException("Not supported encodedValue type: " + encodedValue.getClass().getName());
+      }
```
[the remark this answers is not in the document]
brusdev: > doing '== boolean' in an if is pretty ugly. Seems like either using !(encodedValue instanceof String), or changing the if to wrap the overall comparison and throw at the end would be nicer.
Nice suggestion moving it at the end

> This could also NPE generating the exception, on encodedValue.getClass().
I would not add another condition to check if value is null just to throw NPE

> "Unsupported" would seem a better fit rather than "Not supported".
I'll replace Not supported with Unsupported

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/MD5SensitiveDataCodec.java
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.activemq.artemis.tests.integration.security;
+
+import org.apache.activemq.artemis.utils.SensitiveDataCodec;
+import org.apache.commons.codec.digest.DigestUtils;
+
+public class MD5SensitiveDataCodec implements SensitiveDataCodec<String> {
```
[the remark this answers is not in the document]
brusdev: I'lll replace it with an inner class of the `SecurityTest` class.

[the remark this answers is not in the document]
gemmellr: Think you forgot to remove this when adding the other class.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/SecurityTest.java
```diff
@@ -2564,4 +2582,21 @@ private void checkUserSendNoReceive(final String queue, final ClientSession conn
          fail("Invalid Exception type:" + e.getType());
       }
    }
+
+   public static class MD5SensitiveDataCodec implements SensitiveDataCodec<String> {
```
gemmellr, line 2586: Same general comment as before. The DigestUtils class being used in the impl has many options, so I see no reason to use MD5 at this stage, even if it is only for a test impl.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodec.java
```diff
@@ -79,6 +79,21 @@ public void init(Map<String, String> params) throws Exception {
...
+         if (value instanceof String) {
+            return verify(((String)value).toCharArray(), (String)encodedValue);
+         } else if (value instanceof char[]) {
+            return verify((char[])value, (String)encodedValue);
+         } else {
+            throw new IllegalArgumentException("Unsupported value type: " + value.getClass().getName());
+         }
+      } else {
+         throw new IllegalArgumentException("Unsupported encodedValue type: " + encodedValue.getClass().getName());
+      }
```
gemmellr, line 94: This seems like its missing a bunch of unit tests. Its unclear why it should be allowed to pass a char[] given it is implementing "SensitiveDataCodec<String>" and assumes its given and returns String when encoding and decoding. But if its going to then it should be tested explicitly.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodec.java
```diff
@@ -188,6 +208,16 @@ public String encode(String secret) throws Exception {
          BigInteger n = new BigInteger(encoding);
          return n.toString(16);
       }
+
+      @Override
+      public boolean verify(char[] inputValue, String storedValue) {
+         try {
+            return Objects.equals(storedValue, encode(String.valueOf(inputValue)));
```
gemmellr, line 215: The methods here seems somewhat all over the place.

The class implements SensitiveDataCodec<String>, meaning it has to return String (i.e T) for encode and decode methods, but is only taking Object args which it then blindly assumes are actually String and casts to that unchecked.

The new compare method also takes Object args and then for some odd reason requires one of them to be String or char[], but only String in the other. It converts the first to char[] if it wasnt, so it can call this verify method which has fixed char[] and String type args. Which then turns the char[] _[back]_ into a String so it can call encode with a String to get another String to further compare to the given String.

Why allow different types in the various different methods? Why is one methods args fixed type and the others floating Object, that are then assumed or enforced to be a specific types (meaning the broker has to be calling it with those types), and yet always ending up as String ultimately. Almost nothing is using the generic T type to pick up the String generic from the class definition but that appears to be the only type that is used in the broker so nothing else will actually work.

Actually, why add both new verify and compare methods when they do essentially the same thing overall? Verify is probably a better name overall since compare doesnt as much imply the fact one arg has to be manipulated to equal the other (and the not-present javadoc doesnt say what the method actually does at all)

artemis-commons/src/test/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodecTest.java
```diff
@@ -39,20 +40,33 @@ public void testDefaultAlgorithm() throws Exception {
...
-      try {
-         codec.decode(maskedText);
-         fail("one way algorithm can't decode");
-      } catch (IllegalArgumentException expected) {
+      if (algorithm.equals(DefaultSensitiveStringCodec.ONE_WAY)) {
+         //one way can't decode
+         try {
+            codec.decode(maskedText);
+            fail("one way algorithm can't decode");
+         } catch (IllegalArgumentException expected) {
```
gemmellr, line 63: UnsupportedOperation or IllegalState would seem a far better fit, but if this is existing behaviour I guess perhaps retaining it might be the way to go.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SecureHashProcessor.java
```diff
@@ -33,6 +33,6 @@ public String hash(String plainText) throws Exception {
    //storedValue must take form of ENC(...)
    public boolean compare(char[] inputValue, String storedValue) {
       String storedHash = storedValue.substring(4, storedValue.length() - 1);
-      return codec.verify(inputValue, storedHash);
+      return codec.compare(inputValue, storedHash);
```
gemmellr, line 36: Ahhh. I've missed that the verify method isnt new, its an existing method defined in DefaultSensitiveStringCodec that you just implemented in another bit of DefaultSensitiveStringCodec. But added the essentially same compare method to the SensitiveDataCodec interface.

My main thought is still why have both, and why does the private CodecAlgorithm class need to exist within DefaultSensitiveStringCodec at all, just to define methods that are basically already defined on the SensitiveDataCodec interface.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodec.java
brusdev: The compare method has Object args to match other method signatures and to compatible with the existing verify method of DefaultSensitiveStringCodec. See my other comment on SecureHashProcessor for further details.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SecureHashProcessor.java
brusdev: I have changed SecureHashProcessor to relay on  SensitiveDataCodec<String> to compare user passwords but SensitiveDataCodec<String> hadn't a compare/verify method. So I have added the compare method in SensitiveDataCodec and I have implemented it in DefaultSensitiveStringCodec calling the existing verify method.

I see your point on DefaultSensitiveStringCodec we could change its implementation but that could be another PR.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodec.java
gemmellr: Match/compatible in what way though? The existing encode/decode methods args are blindly assumed _unchecked_ to have to be String. The broker only expects and provides String objects for those method, with the overall type used also hard coded to require a SensitiveDataCodec\<String\>.

For SecureHashProcessor compare(), it is [_still_] fixed to char[] + String, as is DefaultSensitiveStringCodec verify (and its internal CodecAlgorithm, which doesnt actually use Object for any of the methods). It seems that adding e.g SensitiveDataCodec.verify(char[], T) would be nicer and/or simpler, easily get rid of a bunch of duplication and unnecessary arg validation. A verify(T, T) could be added compatibly later if actually needed.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SecureHashProcessor.java
gemmellr: The existing stuff is already a little messy and this expands it and makes it more complicated to follow overall. It seems rather unnecessary when it could instead be making it nicer and simpler even while adding/exposing this functionality. The 'could be another PR' changes tends not to happen, just leaving the earlier stuff layering the onion.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodec.java
brusdev: Looking existing encode/decode methods of SensitiveDataCodec<T> interface with Object args, I think they were deigned to encode/decode different sensitive object types to type T. So IMO the type T is to define the outputs not the inputs in the original design. I used the Object args for the compare method to keep the original design of the interface.

While SensitiveDataCodec.verify(char[], T) is fine if T is String, it could not work well when T is not String.

brusdev: I agree on SensitiveDataCodec.verify(T, T), I was against SensitiveDataCodec.verify(char[], T)

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SecureHashProcessor.java
brusdev: Fair enough, I'll move this PR back to draft and I'll revisit DefaultSensitiveStringCodec

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodec.java
gemmellr: I'm fine with that also, though I think passing it a char[] is perfectly reasonable given the use case. Just not defining entirely unnecessary Object args that only muddy the waters given the usage.

brusdev: Used SensitiveDataCodec.verify(T, T)

gemmellr (commented):
Looks good. I'd maybe have stuck with (char[], T) as per earlier comments, which would be a smaller change still. The methods on CodecAlgorithm could perhaps be dropped and implement the interface but thats mostly just cosmetic at this point.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/LazyHashProcessor.java
```diff
@@ -0,0 +1,55 @@
...
+            supplierException = e;
+            throw supplierException;
+         } catch (Exception e) {
+            supplierException = new RuntimeException(e);
+            throw supplierException;
+         }
+      }
+
+      return hashProcessor;
+   }
```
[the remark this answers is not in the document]
brusdev: fixed

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/LazyHashProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+               } catch (RuntimeException e) {
+                  supplierException = e;
+                  throw supplierException;
+               } catch (Exception e) {
+                  supplierException = new RuntimeException(e);
+                  throw supplierException;
+               }
+            }
+         }
+      }
```
gemmellr, line 47: This will still be non-thread-safe because the hashProcessor field isnt volatile (prior to the synchronized addition the other one would have needed to be also). 
https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java

brusdev: fixed

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/LazyHashProcessor.java
```diff
@@ -0,0 +1,63 @@
...
+
+      return hashProcessor;
+   }
```
gemmellr, line 50:
```suggestion
public abstract class LazyHashProcessor implements HashProcessor {
   private volatile HashProcessor hashProcessor;
   private RuntimeException supplierException;

   public HashProcessor getHashProcessor() {
      HashProcessor hp = hashProcessor;
      if (hp == null) {
         synchronized (this) {
            if (supplierException != null) {
               throw supplierException;
            }

            hp = hashProcessor;
            if (hp == null) {
               try {
                  hashProcessor = hp = createHashProcessor();
               } catch (RuntimeException e) {
                  supplierException = e;
                  throw supplierException;
               } catch (Exception e) {
                  supplierException = new RuntimeException(e);
                  throw supplierException;
               }
            }
         }
      }

      return hp;
   }
```
I think the initial if (supplierException != null) bit can just be removed, and that field doesnt then need to be volatile as its then only used under synchronization. The exception will only be non-null if hashProcessor is still null, and whole the idea is to typically not expect it to be null, so it seems unnecessary to pre-check the exception every time through. It can just rely on the check at the start of the synchronized block.

I think it should also use the local variable to copy hashProcessor to as typically expected in these cases. 

Something like the suggestion (didnt check it compiles).

gemmellr (commented):
Seems good other than the last suggestion I just added. I think it would be good for others to review and see what they think, e.g on the char[]->String change on the newly exposed verify/compare methods.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/LazyHashProcessor.java
brusdev: @franz1981 @gemmellr thanks I replaced the getHashProcessor with the MemoizingSupplier.

gemmellr:
I left it saying "I think it would be good for others to review and see what they think, e.g on the char[]->String change on the newly exposed verify/compare methods."

(the context of prior discussion would show I would probably just have used the char[] type of the existing method being exposed here)

jbertram:
@gemmellr, thanks for the clarification.

@brusdev, I agree with @gemmellr here. I think we should keep `char[]`. 

For what it's worth, the reason `char[]` is used for the password in many places is because it is arguably more secure than `String` mainly because `String` is immutable and may stay on the heap for a long time whereas `char[]` can be sanitized after use which means it is not vulnerable to heap inspection. The broker itself should probably do a better job of using `char[]` for passwords in _more_ places, not fewer.

brusdev:
@gemmellr @jbertram thanks for your feedback, I have just reverted to `char[]`

PR #3852, opened 2021-11-15
NO-JIRA Add AllClassesTest

brusdev:


tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/AllClassesTest.java
```diff
@@ -0,0 +1,139 @@
...
+      Arrays.sort(targetConstructors, (c1, c2) -> c2.getParameterCount() - c1.getParameterCount());
+      for (Constructor targetConstructor : targetConstructors) {
+         List<Object> initArgs = new ArrayList<>();
+         Parameter[] constructorParameters = targetConstructor.getParameters();
+
+         for (Parameter constructorParameter : constructorParameters) {
+            Object initArg = null;
+            if (constructorParameter.getType().isAssignableFrom(byte.class)) {
+               initArg = RandomUtil.randomByte();
+            } else if (constructorParameter.getType().isAssignableFrom(byte[].class)) {
```
clebertsuconic, line 104: nice!!!!!!

clebertsuconic: the randoms on the constructors... nice!!!

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/AllClassesTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+   @Test
+   public void testToString() {
+      Object targetInstance = newInstance(targetClass);
+      Assume.assumeTrue("cannot create " + targetClass.getName(), targetInstance != null);
+
+      try {
+         String targetOutput = targetInstance.toString();
+         log.debug("targetOutput: " + targetOutput == null ? "null" : targetOutput);
+      } catch (NullPointerException | UnsupportedOperationException ignore) {
```
clebertsuconic, line 88: Can't you just not call the default constructors?

a toString() from a well defined constructor should not throw any exceptions here.

perhaps we could fix whatever issues come after this?

brusdev: The `newInstance` try to call any constructor of the target classes starting from the one with more parameters to the one with less parameters.

brusdev:
This PR is replaced by #3854 that fixes toString methods throwing exceptions too.

PR #3853, opened 2021-11-15
ARTEMIS-3574 multiple bindings for embedded webserver

MM53:
* Add BindingDTO to allow configuring multiple addresses to listen on
* Start a new ServerConnector for each binding and deploy the corresponding web-applications
* Update documentation and tests

### DIFF
--- a/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/bootstrap-web-settings.txt
+++ b/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/bootstrap-web-settings.txt
@@ -1,6 +1,8 @@
    <!-- The web server is only bound to localhost by default -->
-   <web bind="${web.protocol}://${http.host}:${http.port}" path="web"${extra.web.attributes}>
-       <app url="activemq-branding" war="activemq-branding.war"/>
-       <app url="artemis-plugin" war="artemis-plugin.war"/>
-       <app url="console" war="console.war"/>
+   <web path="web"${extra.web.attributes}>
+       <binding uri="${web.protocol}://${http.host}:${http.port}">
+           <app url="activemq-branding" war="activemq-branding.war"/>
+           <app url="artemis-plugin" war="artemis-plugin.war"/>
+           <app url="console" war="console.war"/>
+       </binding>
    </web>
--- a/artemis-cli/src/test/java/org/apache/activemq/cli/test/ArtemisTest.java
+++ b/artemis-cli/src/test/java/org/apache/activemq/cli/test/ArtemisTest.java
@@ -254,8 +254,9 @@ public class ArtemisTest extends CliTestBase {
       assertTrue(bootstrapFile.exists());
       Document config = parseXml(bootstrapFile);
       Element webElem = (Element) config.getElementsByTagName("web").item(0);
+      Element bindingElem = (Element) webElem.getElementsByTagName("binding").item(0);
 
-      String bindAttr = webElem.getAttribute("bind");
+      String bindAttr = bindingElem.getAttribute("uri");
       String bindStr = "http://" + Create.HTTP_HOST + ":" + Create.HTTP_PORT;
 
       assertEquals(bindAttr, bindStr);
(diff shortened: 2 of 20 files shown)


jbertram:
In general I like this idea since the underlying Jetty server supports this kind of behavior natively. I was concerned at first that supporting this would require creating two instances of Jetty, but that is not the case. This is good work, @MM53!

The problem is that there is no backwards compatibility with the previous configuration. Using the previous configuration results in an NPE, i.e.:
```
java.lang.NullPointerException
        at org.apache.activemq.artemis.dto.WebServerDTO.getBindings(WebServerDTO.java:96)
        at org.apache.activemq.artemis.component.WebServerComponent.configure(WebServerComponent.java:87)
        at org.apache.activemq.artemis.cli.commands.Run.execute(Run.java:125)
        at org.apache.activemq.artemis.cli.Artemis.internalExecute(Artemis.java:157)
        at org.apache.activemq.artemis.cli.Artemis.execute(Artemis.java:105)
        at org.apache.activemq.artemis.cli.Artemis.execute(Artemis.java:132)
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at java.base/java.lang.reflect.Method.invoke(Method.java:566)
        at org.apache.activemq.artemis.boot.Artemis.execute(Artemis.java:134)
        at org.apache.activemq.artemis.boot.Artemis.main(Artemis.java:50)
```
This is a *breaking* change and we cannot introduce it in a minor release. It would need to wait until the next major release.

Alternatively you could provide backwards compatibility and then I think this could be merged and used in the next minor release.

jbertram:
FWIW, there are a handful of checkstyle violations. I recommend you run `mvn compile -Pdev` and inspect the results locally or check out the results of the automated PR builds.

MM53:
@jbertram, thanks for your response. My plan was to keep backwards compatibility but I forgot to check for null instead I only checked if the list contains some bindings. I will fix this issue and test the code with some previous configurations to see if there are more problems.

jbertram:
@MM53, that's good news. It would probably be worth leaving some tests for the old style configuration and creating new tests for the new config. That way the test-suite will exercise both options and problems like this would be detected automatically.

MM53:
I updated my changes to work properly with the old style configuration and added some tests to verify both formats. I also resolved the checkstyle violations. Hope everything is fine now.

jbertram:
@MM53, there's a conflict now. It's a simple change from c502e94ade9e050849a2d7fd2431c28ea77f4f0a. Once you resolve the conflict I'll merge the PR. Thanks!

clebertsuconic:
@jbertram / @MM53 what's the status on this?

I'm going to release today, but this is a big change to include right before the release... I will leave it out for now..

jbertram:
I believe the status on this PR is good. I was going to merge it as soon as the PR builds finished. I'm testing manually now to confirm it doesn't break upgrades for existing installs.

jbertram:
I just verified that these changes don't break existing/upgrading users.

jbertram:
@clebertsuconic, do you mind if I merge this?

clebertsuconic:
@jbertram / @MM53 I am reverting this... there's a failure from org.apache.activemq.artemis.tests.smoke.console.ConsoleMutualSSLTest that looks real to me


I would rather defer the change now than break the release..

Can one of you send a new PR with the fix later please?


Thanks

jbertram:
The failing test-case was already noted - `org.apache.activemq.artemis.tests.smoke.console.ConsoleMutualSSLTest`. It fails for me when I run it in my IDE.

MM53:
I will have a look at it and create a new PR.

clebertsuconic:
@MM53 an easy way to reproduce it if you don't want to try on the IDE is by running the following command from the ./smoke-tests directory:

```
cd activemq-artemis/tests/smoke-tests
../../one-test.sh ConsoleMutualSSLTest
```


if you don't like using the one-test.sh (which is a script I added because I am not good to remember the actual command when I need it fast:

```
cd activemq-artemis/tests/smoke-tests
mvn -Ptests -DfailIfNoTests=false -Ptests-retry -Ptests-CI -Pextra-tests -DskipStyleCheck=true -DskipPerformanceTests=false -Dtest=ConsoleMutualSSLTest test
```

(you will probably understand why I committed one-test.sh)

MM53:
@clebertsuconic thanks, I will try to fix the problem before you build the actual release

MM53:
@clebertsuconic there was a bug while generating the configuration for a new server which mixed up the old and new configuration style. Therefore the keyStore and trustStore parameters were ignored. I fixed this and created a new PR: #3882 containing all changes.

Not sure why the maven build didn't fail because of this but now everything should work as expected.

PR #3856, opened 2021-11-16
ARTEMIS-3581 - allow max-size-bytes=0 configuration force pagingindep…

gtully:
…endent of the page-size-bytes


gtully:
full test looks good, nothing seems to mind that that illegal state relating to page-size < max-size-bytes has gone!

franz1981 (approved):
Honestly surprised it works OOTB, great!!

PR #3857, opened 2021-11-17
ARTEMIS-3522 Implement performance tool for RUL benchmarks

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3522


franz1981:
I still need to clean-up code, improve reporting, add histogram depedencies, expose wall-clock SPI providers (to support microseconds latencies across processes/machines, using JNI/JNA) + add a user guide

It's already possible to play with it (by manually adding hdr histogram dep)

gtully:
I think just with a little user guide this has great value already. nice.

franz1981:
After working on integrating this with CI I'm improving:

- documentation
- error handling (now reported as proper statistics)
- telemetry of load generator itself

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/perf/PerfClientCommand.java
```diff
@@ -0,0 +1,191 @@
...
+import java.util.concurrent.locks.LockSupport;
+
+import io.airlift.airline.Command;
+import io.airlift.airline.Option;
+import io.netty.channel.DefaultEventLoop;
+import io.netty.channel.DefaultEventLoopGroup;
+import io.netty.channel.EventLoop;
+import org.apache.activemq.artemis.cli.commands.ActionContext;
+
+@Command(name = "client", description = "It will produce and consume messages to a broker instance")
```
jbertram, line 34: I wonder if `client` is the best name here. It seems a little too close to the `consumer` and `producer` commands we already have. I realize that it's part of the `perf` suite of commands, but maybe something like `benchmark` or `run` might be better. Naming things is hard. :(

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/perf/PerfConsumerCommand.java
```diff
@@ -0,0 +1,103 @@
...
+import javax.jms.ConnectionFactory;
+import javax.jms.Destination;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.LockSupport;
+
+import io.airlift.airline.Command;
+import io.airlift.airline.Option;
+import org.apache.activemq.artemis.cli.commands.ActionContext;
+
+@Command(name = "consumer", description = "It will consume messages to a broker instance")
```
jbertram, line 28: Should say, "consume messages _from_ a broker instance."

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/perf/PerfClientCommand.java
franz1981: Agree; I didn't called it `benchmark` because `perf benchmark` sounds like a repetition to me, but probably is better then `perf client`.

Consider that the `perf` tools are:
- client
- producer
- consumer

And indeed have duplicates of our existing `producer`/`consumer` tools although I've hoped that the `perf` prefix helped to distinguish them...let's speak online to find some better naming

franz1981:
I plan to move and merge it as soon as I fix the conflicts.
Any additional change will land on new/different commits.

LICENSE
```diff
@@ -208,3 +208,6 @@
 This file is in the public domain.  For more information see:
 
 activemq-commons/src/main/java/org/apache/activemq/utils/Base64.java
+
+This product bundles HdrHistogram, which is available under a
+"2-clause BSD" license. For details, see https://github.com/HdrHistogram/HdrHistogram/blob/master/LICENSE.txt.
```
gemmellr, line 213: This needs changed for a couple reasons.

This is almost certainly the wrong LICENCE file. This is for the source release. You prsumably are depending on a binary and so, if it is distributed, this would go in the binary archives LICENCE file, at _artemis-distribution/src/main/resources/licenses/bin/LICENSE_.

The licence itself needs to be included in the file, or in another file within the repo and referenced to (since the URL contents could change, e.g to a different licence, or just go away, e.g branch change without using the rename tool). Youll see lots of examples of the latter in the other file.

franz1981: I've re-used the same approach of the Apache Pulsar folks here: let me check why/what was the motivation (spoken with @gtully few months ago about it)

franz1981: > You prsumably are depending on a binary and so, if it is distributed, this would go in the binary archives LICENSE file, at artemis-distribution/src/main/resources/licenses/bin/LICENSE.

I see that now; yes good catch!

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/perf/AsyncJms2ProducerFacade.java
```diff
@@ -0,0 +1,256 @@
...
+      pending--;
+      completedMsgInTransaction++;
+      if (transactionCapacity > 0) {
+         if (completedMsgInTransaction == transactionCapacity || (closing && pending == 0)) {
+            completedMsgInTransaction = 0;
+            pendingMsgInTransaction = 0;
+            try {
+               session.commit();
+            } catch (final Throwable t) {
+               // TODO log
```
gemmellr, line 184: Seems fairly important to do this at least. Not immediately clear why it wouldnt just throw. What if it was a TransactionRolledBackException?

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/perf/AsyncJms2ProducerFacade.java
```diff
@@ -0,0 +1,256 @@
...
+   BytesMessage createBytesMessage() throws JMSException {
+      return session.createBytesMessage();
+   }
+
+   private void addedPendingSend() {
+      if (transactionCapacity > 0 && pendingMsgInTransaction == transactionCapacity) {
+         throw new IllegalStateException("cannot add more pending send");
+      }
+      if (maxPending > 0 && pending == maxPending) {
+         throw new IllegalStateException("cannot add more pending");
```
gemmellr, line 117: Should the messages match (or else be more clearly differentiated)?

franz1981:
Going to merge this now and keep on improving it as new feedbacks arrive

PR #3861, opened 2021-11-18
ARTEMIS-3586 - ActiveMQActivationSpec equals method has switched singleConnection and useJNDI null pointer verification

tlbueno:
The equals method on ActiveMQActivationSpec has a switched null pointer verification for singleConnection and useJNDI fields. 

Since both fields are initialized with a value by default it should not be a problem but if one of these fields are set as null at some point and then equals is called it throws a NPE.


tlbueno:
@brusdev

brusdev (approved):
@tlbueno good catch

PR #3862, opened 2021-11-19
ARTEMIS-3593: Be defensive when reading data from `ActiveMQBuffer` and allocating memory.

vkolomeyko:
Or else, an adversary may handcraft the packet causing OOM situation for a running a JVM.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XidCodecSupport.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XidCodecSupport.java
@@ -33,15 +33,20 @@ public class XidCodecSupport {
    }
 
    private static byte[] safeReadBytes(final ActiveMQBuffer in) {
-      int claimedSize = in.readInt();
-      int bufferCapacity = in.capacity();
+      final int claimedSize = in.readInt();
+
+      if (claimedSize < 0) {
+         throw new XidPayloadException("Payload size cannot be negative");
+      }
+
+      final int readableBytes = in.readableBytes();
       // We have to be defensive here and not try to allocate byte buffer straight from information available in the
       // stream. Or else, an adversary may handcraft the packet causing OOM situation for a running JVM.
-      if (claimedSize > bufferCapacity) {
-         throw new IllegalStateException("Buffer size: " + claimedSize +
-                 " exceeds overall buffer size of: " + bufferCapacity);
+      if (claimedSize > readableBytes) {
+         throw new XidPayloadException("Attempted to read: " + claimedSize +
+                 " which exceeds overall readable buffer size of: " + readableBytes);
       }
-      byte[] byteBuffer = new byte[claimedSize];
+      final byte[] byteBuffer = new byte[claimedSize];
       in.readBytes(byteBuffer);
       return byteBuffer;
    }
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XidCodecSupport.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XidCodecSupport.java
@@ -32,12 +32,24 @@ public class XidCodecSupport {
       out.writeBytes(xid.getGlobalTransactionId());
    }
 
+   private static byte[] safeReadBytes(final ActiveMQBuffer in) {
+      int claimedSize = in.readInt();
+      int bufferCapacity = in.capacity();
+      // We have to be defensive here and not try to allocate byte buffer straight from information available in the
+      // stream. Or else, an adversary may handcraft the packet causing OOM situation for a running JVM.
+      if (claimedSize > bufferCapacity) {
+         throw new IllegalStateException("Buffer size: " + claimedSize +
+                 " exceeds overall buffer size of: " + bufferCapacity);
+      }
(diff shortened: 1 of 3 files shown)


vkolomeyko:
FAO: @andytaylor

franz1981 (changes requested):
If exceed readable bytes

franz1981:
Otherwise look good: it's a good change but I would like to see a test

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XidPayloadException.java
```diff
@@ -0,0 +1,29 @@
...
+
+package org.apache.activemq.artemis.utils;
+
+/**
+ * Special type of exception with stacktrace removed which is thrown when binary payload representing
+ * [{@link javax.transaction.xa.Xid}] is invalid.
+ */
+public class XidPayloadException extends RuntimeException {
+
+   public XidPayloadException(String message) {
```
franz1981, line 26: Iirc it should overload fill stack trace and do nothing on it in order to be stackless

franz1981 (changes requested):

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XidPayloadException.java
vkolomeyko: There are few ways to achieve the same effect.
I am creating parent `RuntimeException` with `writableStackTrace = false`, see line below which also ensures stacktrace is not populated.

I have extended the unit test to illustrate my point.

vkolomeyko:
@franz1981 - CI tests have passed. Please let me know if my changes make sense.

vkolomeyko:
@franz1981, @clebertsuconic, @jbertram - I would like to follow-up on this PR.

Could you kindly review my work and let me know what you think?

My interest in this work is not leisured. We have experienced a security vulnerability in our product and I hope that the fix I am making will be included into the next version of Artemis Active MQ.

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XidCodecSupport.java
```diff
@@ -33,15 +33,20 @@ public static void encodeXid(final Xid xid, final ActiveMQBuffer out) {
...
+      }
+
+      final int readableBytes = in.readableBytes();
       // We have to be defensive here and not try to allocate byte buffer straight from information available in the
       // stream. Or else, an adversary may handcraft the packet causing OOM situation for a running JVM.
-      if (claimedSize > bufferCapacity) {
-         throw new IllegalStateException("Buffer size: " + claimedSize +
-                 " exceeds overall buffer size of: " + bufferCapacity);
+      if (claimedSize > readableBytes) {
+         throw new XidPayloadException("Attempted to read: " + claimedSize +
```
clebertsuconic, line 46: Why do you actually need a Custom Exception here?

clebertsuconic:
@vkolomeyko what led you to do this? did you see this particularly happen anywhere?


this needs a JIRA...

and I would not use the Exception you created..I would just use ActiveMQAMQPIllegalStateException here.

this is a way exceptional case.. this should not happen at all.. if this happens it seems a bug to me, hence I'm asking why you did this PR.

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XidCodecSupport.java
vkolomeyko: This was to address input from @franz1981 : https://github.com/apache/activemq-artemis/pull/3862#issuecomment-974047110
To create a custom exception that does not have stacktrace populated.

@clebertsuconic - I note your remark on `ActiveMQAMQPIllegalStateException` and I am flexible to use either approach.

vkolomeyko:
@clebertsuconic - thanks for your input.

I have raised a Jira: https://issues.apache.org/jira/browse/ARTEMIS-3593 explaining the reason for this change.

clebertsuconic:
@vkolomeyko Why you went the extra mile to remove the stack trace from the exception?


I'm raising a new PR based on this one with some refactoring... I will ask your review.

gtully:
It may make sense to whack the connection in this case, it is a protocol violation. 5.x would error out with an IOException and close the connection. It is sensible behaviour if the client is sending rogue or invalid data. A good citizen may even notice and stop.

clebertsuconic:
I agree.. if you actually throw any RuntimeException on the decode that will kill the connection


@vkolomeyko let me handle this PR? I will do some tweaking to the PR, and I will set you as the author on the commit... (with a note of my collaboration)

clebertsuconic:
I am refactoring the safeRead into BytebufferUtil BTW. so we can reuse the method in other packets

vkolomeyko:
> I am refactoring the safeRead into BytebufferUtil BTW. so we can reuse the method in other packets

@clebertsuconic - sure. Please feel free to add me as a reviewer as soon as you have PR ready.

clebertsuconic:
PR #3871 replaces this one

clebertsuconic:
@vkolomeyko if you could review #3871 please?

I am now closing this one.

PR #3865, opened 2021-12-01
ARTEMIS-3576 Fix toString methods throwing exceptions

brusdev:


gemmellr:
Its nice to see the runtime has been reduced from ~30 seconds to 'only' ~5 seconds in this version of the PR. Thats still a significant chunk of time I dont think this is worth spending on each build for this, the classes changed should have specific tests of their behaviour taking milliseconds rather than relying on something like this, however it is at least less egregious with the further 25 seconds knocked off.

If we are going to have this it could at least produce the cleanest output possible to begin with though. Currently the output is still showing stacktraces for a couple of silly cases:

- One of the same general cases many of the changes here are for, of the test causing an NPE on something thats reasonably expected to never to be null. Its just that JournalFileImpl is catching the NPE and logging it in this case, meaning it doesnt fail the test. As with the others it should just be enforced.
- Its also showing up a finalizer log for ClientSessionFactoryImpl...which feels pointless given it doesnt actually seem to have a toString() method impl at all. Given it takes lots of constructor args, I'd guess one of them can be enforced on to shut it up.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
```diff
@@ -238,6 +238,8 @@ public ClientSessionFactoryImpl(final ServerLocatorInternal serverLocator,
       if (connectorConfig.getB() != null) {
          this.backupConnectorConfig = connectorConfig.getB();
       }
+
```
clebertsuconic, line 241: @brusdev I will send a PR removing this right now :)


that way you rebase without this...

clebertsuconic: here https://github.com/apache/activemq-artemis/pull/3872

brusdev: great thanks

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
```diff
@@ -90,7 +90,7 @@
 
    private ConnectorFactory connectorFactory;
 
-   private transient boolean finalizeCheck = true;
```
clebertsuconic, line 93: this will clash with the other PR.. you will have to remove this change after I merge the other.

PR #3867, opened 2021-12-01
ARTEMIS-3594 - add support for a key transformer and in instance of C…

gtully:
…ONSISTENT_HASH_MODULO that can be used to partition in a static cluster


artemis-server/src/test/resources/ConfigurationTest-full-config.xml
```diff
@@ -158,6 +158,14 @@
             <target-key-filter>^[^.]+</target-key-filter>
             <local-target-filter>DEFAULT</local-target-filter>
          </broker-balancer>
+         <broker-balancer name="simple-local-with-transformer">
+            <target-key>CLIENT_ID</target-key>
+            <target-key-filter>^[^.]+</target-key-filter>
+            <local-target-filter>DEFAULT</local-target-filter>
+            <key-transformer name="CONSISTENT_HASH_MODULO">
```
gtully, line 165: @brusdev I dropped the target- prefix on key for key-transformer.

I am thinking a separate PR to maybe remove target- from target-key and target-key-filter, or a general revisit on names. A key, could really be a key-type etc... naming is very hard but I think that is a question of a separate PR.

If you like I could leave the name target-key-transformer - but in that case I may want to call it local-target-key-transformer - as that is more exact.

gtully: maybe is should be local-target-key-transformer, where it is a local-target and not a target-key, ie: removing target from key will still make sense.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/AutoClientIDShardClusterTest.java
```diff
@@ -0,0 +1,343 @@
...
+      assertTrue(sub0.orderShot.get() || sub1.orderShot.get());
+
+      sub0.consumerDone.set(true);
+      sub1.consumerDone.set(true);
+
+      stopServers(0, 1);
+   }
+
+   @Test
+   public void testWithConsistentHashClientIDModTwo() throws Exception {
```
brusdev, line 225: testWithConsistentHashClientIDModTwo is very similar to testWithoutOutSharding, is it possibile reuse the common code?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/AutoClientIDShardClusterTest.java
```diff
@@ -0,0 +1,343 @@
...
+         brokerBalancerConfiguration.setTransformerConfiguration(transformerConfig);
+
+         configuration.setBalancerConfigurations(Collections.singletonList(brokerBalancerConfiguration));
+
+         TransportConfiguration acceptor = getDefaultServerAcceptor(node);
+         acceptor.getParams().put("redirect-to", BROKER_BALANCER_NAME);
+      }
+   }
+
+   protected ConnectionFactory createFactory(String protocol, String clientID, String user, String password) throws Exception {
```
brusdev, line 295: is it possible to reuse the base method to avoid duplicating the common code?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/AutoClientIDShardClusterTest.java
```diff
@@ -0,0 +1,343 @@
...
+      setupClusterConnection("cluster1", "T", MessageLoadBalancingType.ON_DEMAND, 1, true, 1, 0);
+      toSend.set(numMessages);
+   }
+
+   Thread producer = new Thread(new Runnable() {
+      final AtomicInteger producerSeq = new AtomicInteger();
+
+      @Override
+      public void run() {
+         while (toSend.get() > 0) {
```
brusdev, line 94: is it possible to reuse ClusterTestBase.sendInRange?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/AutoClientIDShardClusterTest.java
```diff
@@ -0,0 +1,343 @@
...
+      new Thread(sub0).start();
+      new Thread(sub1).start();
+
+      // waiting for registration before production to give bridges a chance
+      assertTrue(sub0.registered.await(20, TimeUnit.SECONDS));
+      assertTrue(sub1.registered.await(20, TimeUnit.SECONDS));
+
+
+      // wait for bindings!
+      assertTrue(Wait.waitFor(() -> 2 == gatherBindingsOnT(0)[0], 20000, 2000));
```
brusdev, line 194: is it possible to reuse ClusterTestBase.waitForBindings?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/balancing/policies/PolicyFactoryResolver.java
```diff
@@ -56,19 +59,15 @@ private void loadPolicyFactories() {
...
+   public void registerPolicyFactory(String name, PolicyFactory policyFactory) {
+      policyFactories.put(name, policyFactory);
    }
 
-   public void unregisterPolicyFactory(PolicyFactory policyFactory) {
-      for (String policyName : policyFactory.getSupportedPolicies()) {
-         policyFactories.remove(policyName, policyFactory);
-      }
+   String keyFromClassName(String name) {
+      return name.substring(0, name.indexOf("PolicyFactory"));
```
brusdev, line 71: Is it possible to avoid full/partial class names in the broker.xml?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/AutoClientIDShardClusterTest.java
gtully: maybe, but it does not help with understanding. This use case needs failover and I thought it simpler to do here rather than add failover to the existing base. I guess it could go either way.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/balancing/policies/PolicyFactoryResolver.java
gtully: I think it makes sense to use part of the newly provided class to provide a name to identify the feature. That class wil have to implement a specific interface so it will already be tied to a class. But in this case, the name gives the actual intent of the class. I don't see need for any indirection.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/AutoClientIDShardClusterTest.java
gtully: hmm, just tried this, and wait for bindings has a consumer count expectation, that I don't want b/c my consumers are bouncing!

gtully: I guess I can add support for -1 consumerCount

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/AutoClientIDShardClusterTest.java
gtully: I want to keep them separate but similar b/c the testWithoutOutSharding is non deterministic but informative of the potential problem when producers and consumers bounce around a cluster.

PR #3871, opened 2021-12-02
ARTEMIS-3593 Defense against OME on parsing XID

clebertsuconic:
Co-authored-by: Viktor Kolomeyko <Viktor.Kolomeyko@r3.com>


vkolomeyko:
Looks fine to me.

vkolomeyko:
@clebertsuconic - many thanks for integrating this fix.
Do you have an ETA when a new version is likely to be cut with this fix included?

vkolomeyko:
@clebertsuconic - I can see this fix has now been released as part of Artemis 2.20 release and published to Maven Central recently. Thanks for that!

I also note that version 2.20 comes with JRE upgrade from 1.8 to Java 11 as per [ARTEMIS-3420](https://issues.apache.org/jira/projects/ARTEMIS/issues/ARTEMIS-3420).

Unfortunately, for our product we still use Java 1.8 and upgrading to Java 11 is a bigger deal.

Would it be possible to port this fix to version 2.19 and include it into 2.19.1 for example?

PR #3872, opened 2021-12-02
ARTEMIS-3599 Removing finalization calls

clebertsuconic:


clebertsuconic:
unfortunately my CI failed with lots of unopened files.

it's probably just a test issue. but we will need to fix it before we can merge these.

gemmellr:
The 500 failures you saw on your first run look like they would be unrelated to this. That looked to be the same situation I saw once before, essentially a LibaioContext instance was not closed and caused all following test classes [that subclass ActiveMQTestBase] that verify a count during cleanup to be marked as failed after 30 seconds wait. https://issues.apache.org/jira/browse/ARTEMIS-3435 has some more details, I used it to make changes to have it fail the overall run several hours faster if it happened again, and give some pointers to help give an idea where it went pop, which might possibly allow isolating why it happened.

It failed again that way yesterday for the first time since that change, first in a run of mine, and then also in yours on the next run. In my case it appears it happened during ReplicatedPagedFailoverTest, whereas in yours it either happened in ClusteredGroupingTest or in another before that started (need to check the output to see).

Your second run did still have a few failures though in CloseConnectionOnGCTest which I expect are very much related to this change given the name of the test, which probably just needs deleting.

clebertsuconic:
I removed the test that's invalid now.

PR #3874, opened 2021-12-08
ARTEMIS-3054 - fix lock inversion - intermittent failure of PageClean…

gtully:
…upWhileReplicaCatchupTest and hang


franz1981:
I would add a comment there to enforce it

franz1981 (approved):

PR #3875, opened 2021-12-08
fix intermittent failure of org.apache.activemq.artemis.tests.integra…

gtully:
…tion.client.ConsumerTest#testSendAMQPReceiveCore


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -971,6 +971,9 @@ protected synchronized void lazyScanAfterReloadPersistence() {
       scanMessageData();
       messageDataScanned = MessageDataScanningStatus.SCANNED.code;
       modified = false;
+      // reinitialise memory estimate as message will already be on a queue
+      // and lazy decode will want to update
+      getMemoryEstimate();
```
clebertsuconic, line 976: wow.. nice catch!

if this fixes the issue.. go for it! Merge it!

PR #3876, opened 2021-12-09
ARTEMIS-3604 Async sends overflowing server in OpenWire

clebertsuconic:


artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -213,6 +239,7 @@ public OpenWireConnection(Connection connection,
       this.useKeepAlive = openWireProtocolManager.isUseKeepAlive();
       this.maxInactivityDuration = openWireProtocolManager.getMaxInactivityDuration();
       this.transportConnection.setProtocolConnection(this);
+      this.maxActorSize = openWireProtocolManager.getMaxActorSize();
```
franz1981, line 242: I would use an actor which queue is bounded (we have many in JCtools) adding a method boolean tryAct
that return false if the queue has reached maximum capacity.
This would be more performant and easy to manage, given that could be used by other part of code to limit the size of tasks on the actor.

franz1981 (changes requested):

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireProtocolManager.java
```diff
@@ -97,6 +100,8 @@
 
    private boolean prefixPacketSize = true;
 
+   private int maxActorSize = 20;
```
gtully, line 103: the actor is in units of commands, I wonder if it would be possible to limit by bytes size and match to the configured serverSocketBuffer size. I guess that raw bytes are already unmarshalled when we are dealing with commands.
It is just that it seems natural to map the pending data on in the broker to what can be pending on the tcp layer. So from a client perspective there is max 2*socketBufferSize bytes pending. Or maybe it is 3x, with one being on the client side of the tcp buffer.

clebertsuconic: I wouldn't know how to do that TBH

clebertsuconic: I was going with 1000, but my computer is slow to sync due to some crypto thing...


I have not decided on the default yet.. but this is mostly about syncs...


the other thing is I will probably make those writes to the journal to not sync if the client is not waiting for a response.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
clebertsuconic: the thing is, I may have to go beyond the limit some times.. because the parse on the packet already happened. and the messages were already read from the network. I have to read them all.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -285,21 +333,41 @@ public void bufferReceived(Object connectionID, ActiveMQBuffer buffer) {
...
+      }
+   }
+
+   public void disableAutoRead() {
+      autoRead = false;
+      getTransportConnection().setAutoRead(false);
+   }
+
+   protected void actualEnableAutoReadAndTtl() {
+      getTransportConnection().setAutoRead(autoRead);
```
gtully, line 357: should this no go through checkMemory - to sync up with the page store blocking. If the page store is still blocked we don't want to enable auto read till there is some space.

I am thinking, we don't have to be careful about disabling auto read, b/c we will aways get to do it again... but with enabling, having an empty work queue is not sufficient, there also has to be space. so we need to coordinate the enabling.

clebertsuconic: @gtully when blocked, autoRead will be false.

We are still calling the flush o this case though.. 

we will need the paging to send an event to set this call.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireProtocolManager.java
```diff
@@ -236,6 +238,18 @@ public void removeConnection(ConnectionInfo info, Throwable error) throws Invali
       }
    }
 
+   /*** if set, the OpenWire connection will bypass the tcpReadBuferSize and use this value instead.
+    *   This is by default -1, and it should not be used unless in extreme situations like on a slow storage. */
+   public int getMaxActorSize() {
```
gtully, line 243: this does not make sense to me, a config that should not be used. The intent is to bound the queue and it is matched to the receive buffer size. if it needs to be lower or higher then modify the acceptor receive buffer size. The bounding of the actor stops reading, so at a max the receive buffer size will be pending on the socket and in the broker. Then the client will block on a tcp write. When will it make sense to have more than send buffer pending on the broker? I don't know how to reason about setting a maxActorSize on the protocol manager!
in terms of a name, it could be actorPendingQueueMaxSizeBytes but that is an aside. Only add that if we know when it should be set.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -213,6 +230,15 @@ public OpenWireConnection(Connection connection,
       this.useKeepAlive = openWireProtocolManager.isUseKeepAlive();
       this.maxInactivityDuration = openWireProtocolManager.getMaxInactivityDuration();
       this.transportConnection.setProtocolConnection(this);
+      if (openWireProtocolManager.getMaxActorSize() > 0) {
+         this.maxActorSize = openWireProtocolManager.getMaxActorSize();
+      } else {
+         if (acceptorUsed == null) {
+            this.maxActorSize = TransportConstants.DEFAULT_TCP_RECEIVEBUFFER_SIZE;
+         } else {
+            this.maxActorSize = ((NettyAcceptor) acceptorUsed).getTcpReceiveBufferSize();
```
gtully, line 239: this seems the only right option to me, if the actor can be used without an acceptor, then maybe pass in the maxPendingSize rather than the acceptorUsed.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -765,11 +810,19 @@ public AMQConnectionContext initContext(ConnectionInfo info) throws Exception {
...
       // the actor can only be used after the WireFormat has been initialized with versioning
-      this.openWireActor = new Actor<>(executor, this::act);
+      this.openWireActor = new BoundActor<>(executor, this::act, maxActorSize, this::getSize, this::disableAutoRead, this::flushedActor);
 
       return context;
    }
 
+   private int getSize(Command command) {
+      if (command instanceof ActiveMQMessage) {
+         return ((ActiveMQMessage) command).getSize();
```
gtully, line 820: did a quick check, the only way getSize would return a different value after act is if org.apache.activemq.command.ActiveMQMessage#clearProperties was called and from what I can see we don't do that so this is a reasonable estimate. 
The 1k for other commands may be a little heavy, but I guess it is better to error on the side of caution. 
nice!

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
```diff
@@ -0,0 +1,79 @@
...
+   private volatile int size = 0;
+
+   private static final AtomicIntegerFieldUpdater<BoundActor> SCHEDULED_FUSH_UPDATER = AtomicIntegerFieldUpdater.newUpdater(BoundActor.class, "scheduledFlush");
+   private volatile int scheduledFlush = 0;
+
+   private static final Object FLUSH = new Object();
+
+
+   final int maxSize;
+   final ToIntFunction<T> sizeGetter;
```
franz1981, line 36: `private final` on the members

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
```diff
@@ -0,0 +1,79 @@
...
+      this.maxSize = maxSize;
+      this.sizeGetter = sizeGetter;
+      this.overLimit = overLimit;
+      this.clearLimit = clearLimit;
+   }
+
+   @Override
+   protected final void doTask(Object task) {
+      if (task == FLUSH) {
+         this.scheduledFlush = 0;
```
franz1981, line 54: `scheduledFlush` must be `0` after `clearLimit::run` or a racy `act` after this would call `overLimit::run` before it

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
```diff
@@ -0,0 +1,79 @@
...
+      }
+      try {
+         listener.onMessage((T)task);
+      } finally {
+         SIZE_UPDATER.getAndAdd(this, -sizeGetter.applyAsInt((T)task));
+      }
+   }
+
+   public void act(T message) {
+      int size = SIZE_UPDATER.addAndGet(this, (sizeGetter.applyAsInt((T)message)));
```
franz1981, line 66: `sizeGetter.applyAsInt((T)message)` should check if the estimated size is > 0

franz1981 (changes requested):
Generally looking good, but I'm still not sold on the bound actor abstraction because:

- `act` can be called by several threads concurrently while OpenWire does is single-thread, hence the way it behave isn't the same
- I cannot still fit in my mind what `flush` means and why it call `overLimit::run` regardless current actor pending size

I'm looking how the reactive stream abstraction is implemented in other systems because we probably need a backpressure advertizing with an overflow strategy that can accumulate beyond expected capacity and maybe there's already something implemented/tested around

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
clebertsuconic: this will be null on tests only...

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
clebertsuconic: why?

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireProtocolManager.java
```diff
@@ -236,6 +238,18 @@ public void removeConnection(ConnectionInfo info, Throwable error) throws Invali
...
+      return maxActorSize;
+   }
+
+   public OpenWireProtocolManager setMaxActorSize(int maxActorSize) {
+      System.out.println("max actor size " + maxActorSize);
```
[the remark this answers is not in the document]
gemmellr: Also, maxActorSize feels off as a name given it specifically isnt actually a max and can be exceeded, and indeed doesnt cause anything to happen until it has already been exceeded.

Again, not sure on great alternatives though...actorSizeThreshold?

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
franz1981: because if not (a bug/issue on the implementation), it risks to not work as expected

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireProtocolManager.java
clebertsuconic: @gtully if you have a slow device to sync, and you read say 1M worth of data, but take more than the configured TTL to read again your clients will start to fall apart.

This is a backup plan in case this scenario happens. By default it should use the tcp-read-buffer, but when things fall apart you have the option to use a lower read-value. I have seen that happening on my tests and I don't want to take the risk without having a backup plan.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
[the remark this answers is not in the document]
clebertsuconic: You're right.. this was a copy & paste from another place where I had it as an object ... lol

thanks

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
```diff
@@ -0,0 +1,79 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.activemq.artemis.utils.actors;
+
+import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.function.ToIntFunction;
+
+public class BoundActor<T> extends ProcessorBase<Object> {
```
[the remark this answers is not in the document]
clebertsuconic: @gemmellr SizeAwareActor?... I don't have a better name.. if you find one I will use it

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
clebertsuconic: I think it would be ok to have a packet with zero weight.. if the user has a control or something. 
I don't consider this a bug.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
```diff
@@ -713,6 +713,9 @@ public boolean checkMemory(boolean runOnFailure, final Runnable runWhenAvailable
          }
       } else if (pagingManager.isDiskFull() || addressFullMessagePolicy == AddressFullMessagePolicy.BLOCK && (maxSize != -1 || usingGlobalMaxSize)) {
          if (pagingManager.isDiskFull() || maxSize > 0 && sizeInBytes.get() >= maxSize || pagingManager.isGlobalFull()) {
+            if (runWhenBlocking != null) {
+               runWhenBlocking.run();
```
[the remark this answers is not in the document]
clebertsuconic: I don't have time to handle this one now.. I have to finish this PR today.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/paging/FloodServerWithAsyncSendTest.java
```diff
@@ -0,0 +1,174 @@
...
+      consume1.start();
+      Thread consume2 = new Thread(() -> consume(protocol, "queue2"), "ProducerQueue2");
+      consume2.start();
+
+      Thread produce1 = new Thread(() -> produce(protocol, "queue1"), "ConsumerQueue1");
+      produce1.start();
+      Thread produce2 = new Thread(() -> produce(protocol, "queue2"), "ConsumerQueue2");
+      produce2.start();
+
+      Thread.sleep(10_000);
```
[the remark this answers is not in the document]
clebertsuconic: > Generally looking good, but I'm still not sold on the bound actor abstraction because:
> 
> * `act` can be called by several threads concurrently while OpenWire does is single-thread, hence the behaviour in a general purpose context won't be the same
> * I cannot still fit in my mind what `flush` means and why it call `overLimit::run` regardless current actor pending size
> 
> I'm looking how the reactive stream abstraction is implemented in other systems because we probably need a backpressure advertizing mechanism with an overflow strategy that can accumulate beyond expected capacity and maybe there's already something implemented/tested around

this is about stop reading after a certain limit is over. I still need to parse everything that's on the buffer.. I can't just limit the actor here... things would break otherwise.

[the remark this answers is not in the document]
franz1981: Yeah, in this case is a limit aware actor in this case: it better capture what it does

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
franz1981: Yes but negative is a problem

franz1981 (changes requested):

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
franz1981: @clebertsuconic this must be change to save bad things to happen concurrently

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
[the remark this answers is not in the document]
gemmellr: Still not great, but definitely better than 'Bound' given its not actually bound.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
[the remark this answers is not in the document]
gemmellr: Its unclear to me that you can trust this PR without changing it, given the fact it would be running tasks when it shouldnt, and these are now coordinating on important state.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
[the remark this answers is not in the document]
clebertsuconic: I'm naming it ThresholdAwareActor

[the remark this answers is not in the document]
clebertsuconic: ThresholdActor actually, I'm not using the Aware word

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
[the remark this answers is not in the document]
clebertsuconic: I will address this on a next PR if there's an issue here.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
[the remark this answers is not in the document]
gemmellr: Its better than Bound :)

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/paging/FloodServerWithAsyncSendTest.java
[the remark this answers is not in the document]
gemmellr: Your right, it isnt run in the fast-tests profile..I thought it was, but its not...its just burning time building the module and generating things, though it doesnt actually run them. I'd still probably put it elsewhere.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/paging/FloodServerWithAsyncSendTest.java
```diff
@@ -0,0 +1,174 @@
...
+      Connection connection = null;
+      try {
+         connection = factory.createConnection();
+         connection.start();
+         Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+         Queue queue = session.createQueue(queueName);
+         MessageConsumer consumer = session.createConsumer(queue);
+         int rec = 0;
+         while (running) {
+            consumer.receive(5000);
```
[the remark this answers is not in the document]
clebertsuconic: The point is that a third connection should still allowed to perform. The server should not become unresponsive which is what's happening on this situation.

My laptop has some bad cryptographic thing that slows the server down to a point where I can't guarantee the receives and senders will actually happen. I thought it would be safer to just check the unresponsive server than the actual producers and consumers.

the test is reproducing the issue as I expected here.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/BoundActor.java
clebertsuconic: it's a single thread setting this to false.. there's no concurrency here.

franz1981: `act` can set compareAndSet(0, 1) right after this `this.scheduledFlush = 0` meaning that the Runnable will run without the right order

clebertsuconic: Right... the flush should set it to 0 no matter what.

franz1981:
```
tasks = {1, 2, 3, F}
size = 3
max_size = 2
schedule_flush = 1

act(4)				|	actor's doTask) 	
--------------------------------|-------------------------------
size: 3 -> 4			|				
				|	tasks = {1, 2, 3, F} - {F}
				|	schedule_flush: 0
				|	
				|				
schedule_flush: 0 -> 1		|				
				|				
overLimit.run()			|				
				|	clearThreshold.run();
tasks = {} + {F}		|	
tasks = {} + {4}		|	
				|	tasks = {1, 2, 3} - {3}
				|	size: 4 -> 3
				```
				
				this is to help the future me on this :D

clebertsuconic: that's actually why I set it to 0 before calling...

I may eventually have two in flight if I get one packet arriving with more size then the whole actor...


If you think there's a bug in there feel free to send a new PR *after* this one is merged.. I need to merge this asap.

franz1981: No bud this is a bug, because that listeners calls are not supposed to run in random order, but in the right sequence

clebertsuconic: @franz1981 Ok, I am merging this now, If you want to improve this.. or talk to me next week.

franz1981: Ok, but I don't understand why you don't consider it a bug: listeners must not be called concurrently right? The compare and set is there to prevent it

clebertsuconic: I added a test for this condition.. I don't think I would have caused any harm on the openwire flow control... I'm just trying to fix the flow control in openwire now.

franz1981: I know but given that is a racy behaviour it won't happen easily, like any concurrent bugs, but the execution I shown is valid and can still happen so worth to fix it.
At worst it would still fix your issue: run your existing tests and if it pass it means that is still working as you expect, if not, feel free to merge it as is now and I'll find out later what can be wrong (if any).

clebertsuconic: I believe its fixed.

PR #3877, opened 2021-12-11
ARTEMIS-3604 Fixing size adjustment

franz1981:


clebertsuconic:
@franz1981 thanks.. nice catch

PR #3884, opened 2021-12-16
ARTEMIS-3608 - Add distribution for Multicast messages to OFF_WITH_RE…

AntonRoskvist:
…DISTRIBUTION to avoid message loss


jbertram:
Nice work, @AntonRoskvist. Thanks for the PR!

PR #3889, opened 2021-12-22
ARTEMIS-3620 Journal blocking delete/update record with no sync

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3620

### DIFF
--- a/artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
+++ b/artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
@@ -56,6 +56,7 @@ import org.apache.activemq.artemis.api.core.ActiveMQBuffers;
 import org.apache.activemq.artemis.api.core.ActiveMQExceptionType;
 import org.apache.activemq.artemis.api.core.ActiveMQShutdownException;
 import org.apache.activemq.artemis.api.core.Pair;
+import org.apache.activemq.artemis.core.io.DummyCallback;
 import org.apache.activemq.artemis.core.io.IOCallback;
 import org.apache.activemq.artemis.core.io.IOCriticalErrorListener;
 import org.apache.activemq.artemis.core.io.SequentialFile;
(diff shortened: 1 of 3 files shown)


artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -1161,20 +1162,27 @@ public void run() {
...
+      } else {
+         onFoundAddInfo = new SimpleFutureImpl<>();
+      }
+      if (onFoundAddInfo == null) {
+         internalAppendDeleteRecord(id, false, null, callback);
+      } else {
+         internalAppendDeleteRecord(id, sync, (record, result) -> onFoundAddInfo.set(result), callback);
+      }
+      if (onFoundAddInfo != null) {
+         if (!onFoundAddInfo.get()) {
```
clebertsuconic, line 1182: I'm thinking we should just stop doing that.. just always log, like we do on updates...

The only outcome from the IllegalStateException would be log.warn in most cases.. especially on delete.

franz1981:
Some results of this change.
Pre:
```
Benchmark                                          (compactMinFiles)  (compactPercentage)  (fileSize)  (minFiles)  (poolSize)  (recordSize)  (records)  Mode  Cnt      Score     Error  Units
JournalSyncDeletePerfTest.batchAppendUpdateDelete                 10                   30    10485760           4          20            64        100  avgt   20   1476.175 ±  26.177  us/op
JournalSyncDeletePerfTest.batchAppendUpdateDelete                 10                   30    10485760           4          20            64       1000  avgt   20  14994.321 ± 590.080  us/op
```
Post:
```
Benchmark                                    (compactMinFiles)  (compactPercentage)  (fileSize)  (minFiles)  (poolSize)  (recordSize)  (records)  Mode  Cnt     Score     Error  Units
JournalSyncDeletePerfTest.batchAppendDelete                 10                   30    10485760           4          20            64        100  avgt   20   489.463 ±  28.756  us/op
JournalSyncDeletePerfTest.batchAppendDelete                 10                   30    10485760           4          20            64       1000  avgt   20  4476.861 ± 139.642  us/op
```

franz1981:
CI is green with it so I can move it into non-draft ;)

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -1161,20 +1168,25 @@ public void run() {
...
+         onFoundAddInfo = null;
+      } else {
+         onFoundAddInfo = new SimpleFutureImpl<>();
+      }
+      if (onFoundAddInfo == null) {
+         internalAppendDeleteRecord(id, false, null, callback);
+      } else {
+         internalAppendDeleteRecord(id, sync, (record, result) -> onFoundAddInfo.set(result), callback);
+      }
+      if (onFoundAddInfo != null && !onFoundAddInfo.get()) {
```
gemmellr, line 1187: Similar to earlier comment. Did we start being charged for newlines and noone told me? :)

franz1981:
@clebertsuconic in order to help other prj that relies on these fixes I plan to merge it today if none complains (after fixing what @gemmellr suggested)

clebertsuconic:
@franz1981 is this ready to merge? there was some aesthetic suggestions by Robbie.

PR #3890, opened 2021-12-24
ARTEMIS-3621 show 'zero' timestamp as 'none' (instead of 1970...)

erwindon:
see https://issues.apache.org/jira/browse/ARTEMIS-3621


artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/browse.js
```diff
@@ -491,6 +491,9 @@ var Artemis;
              if (isNaN(timestamp) || typeof timestamp !== "number") {
                 return timestamp;
              }
+             if (timestamp === 0) {
+                return "none";
```
brusdev, line 495: thanks for this improvement, what about `N/A`?

erwindon: That would be fine too.
I always avoided N/A because it has 2 meanings (Collins), or 3 relevant meanings (Webster), or even 4 meanings (Wikipedia). But that is more the uncertainty of myself as a non-native speaker :-)
I'll update this PR. Additionally, the only-and-rare occurrence of `n/a` will be replaced by `N/A`.

PR #3891, opened 2021-12-30
ARTEMIS-3623 preserve type of numeric extraProperties

erwindon:
See https://issues.apache.org/jira/browse/ARTEMIS-3623

@clebertsuconic this is a modification on 'your' code. did I overlook a reason why the extraProperties must always be strings?

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
@@ -872,7 +872,12 @@ public abstract class AMQPMessage extends RefCountMessage implements org.apache.
       TypedProperties extraProperties = getExtraProperties();
       if (extraProperties != null) {
          extraProperties.forEach((s, o) -> {
-            map.put(extraPropertiesPrefix + s.toString(), JsonUtil.truncate(o.toString(), valueSizeLimit));
+            if (o instanceof Number) {
+               // keep fields like _AMQ_ACTUAL_EXPIRY in their original type
+               map.put(extraPropertiesPrefix + s.toString(), o);
+            } else {
+               map.put(extraPropertiesPrefix + s.toString(), JsonUtil.truncate(o.toString(), valueSizeLimit));
+            }
          });
       }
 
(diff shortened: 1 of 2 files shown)


clebertsuconic:
I don't recall of a reason...


would be possible to add a test for this?

erwindon:
> I don't recall of a reason...
ok! (I will not cancel this PR)

erwindon:
> would be possible to add a test for this?

What should be tested then? only this function? or the complete context where it is used, i.e. message-expiry?
Each protocol has its own `toPropertyMap` function. Neither of these have dedicated tests. This is a fix/improvement for the one on the amqp side only.
The amount of work for this may quickly become too much for me.
In that case, alternative #2 or #3 (as listed in https://issues.apache.org/jira/browse/ARTEMIS-3623) may be more appropriate.

clebertsuconic:
@erwindon regarding the text... It would be nice to have an unit test perhaps calling the function and showing the bug.


do you consider this a bug or an improvement BTW?

erwindon:
> do you consider this a bug or an improvement BTW?

I see that I filed the issue under "bug".
Without a specification (e.g. https://en.wikipedia.org/wiki/Software_requirements_specification) it is always a thin line between a "bug" and "something-that-could-have-been-done-different".
In this case, information that was previously available (the data-type) was removed. IMHO, that makes it more a bug than an improvement.
There are 2 use cases for this field: (1) applications consuming these messages, and (2) viewing these messages in the console. Previously the difference between a numeric-field and text-field was unnoticeable in the console until I applied formatting functions on it in an earlier PR. So I am to blame for making a small issue bigger.
And I also admit that the proposed change is not formally complete, as it tests for numeric fields only. Any boolean/date/etc that appears in that location is still converted to text. The main reason for this simplification is that I wanted to solve the presentation issue first. And to my knowledge that happens only for one numeric field.

erwindon:
> regarding the text... It would be nice to have an unit test perhaps calling the function and showing the bug.

ok, working on it now...

erwindon:
setup unit-test with 2 extra properties: `{"firstString":"firstValue", "secondLong":1234567}`

unit-test result, pre-test, second value is passed as string (STR)
```
STR firstString: firstValue
STR secondLong: 1234567
```

unit-test result, post-test, second value is now passed as number (NUM)
```
STR firstString: firstValue
NUM secondLong: 1234567
```

(the print statements for the above output have of course not been committed)

erwindon:
@clebertsuconic the unit test was added to this PR

clebertsuconic:
@erwindon why this still a draft? I think it's ready to merge?


just squash both commits?

erwindon:
I forgot to un-draft...

> just squash both commits?
I'll do that now

PR #3895, opened 2022-01-05
ARTEMIS-3686: Adding example showing how to do telemetry support

nbrendah:
**Issue Link:** https://issues.apache.org/jira/browse/ARTEMIS-3686

hello @clebertsuconic, basing on the advise you gave me, to go back a step behind, this is what i have first produced.  
I am looking forward to seeing your comments

Thank you

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
@@ -801,4 +801,10 @@ public interface Message {
    default String getStringBody() {
       return null;
    }
+
+   /** Used for user context data. Useful on interceptors. */
+   Object getUserContext(Object key);
+
+   /** Used for user context data. Useful on interceptors. */
+   void setUserContext(Object key, Object value);
 }
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
@@ -17,6 +17,7 @@
 
 package org.apache.activemq.artemis.api.core;
 
+import java.util.HashMap;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 // import org.apache.activemq.artemis.utils.collections.ConcurrentHashSet; -- #ifdef DEBUG
@@ -26,6 +27,8 @@ public class RefCountMessage {
    private static final AtomicIntegerFieldUpdater<RefCountMessage> DURABLE_REF_COUNT_UPDATER = AtomicIntegerFieldUpdater.newUpdater(RefCountMessage.class, "durableRefCount");
    private static final AtomicIntegerFieldUpdater<RefCountMessage> REF_COUNT_UPDATER = AtomicIntegerFieldUpdater.newUpdater(RefCountMessage.class, "refCount");
    private static final AtomicIntegerFieldUpdater<RefCountMessage> REF_USAGE_UPDATER = AtomicIntegerFieldUpdater.newUpdater(RefCountMessage.class, "usageCount");
+
+   private volatile HashMap userContext;
 
    private volatile int durableRefCount = 0;
 
@@ -168,6 +171,21 @@ public class RefCountMessage {
       int count = REF_COUNT_UPDATER.incrementAndGet(this);
       onUp();
       return count;
+   }
+
+   public Object getUserContext(Object key) {
+      if (userContext == null) {
+         return null;
+      } else {
+         return userContext.get(key);
+      }
+   }
+
+   public void setUserContext(Object key, Object value) {
+      if (userContext == null) {
+         userContext = new HashMap();
+      }
+      userContext.put(key, value);
    }
 
 }
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/MessageInternalImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/MessageInternalImpl.java
@@ -723,4 +723,13 @@ public class MessageInternalImpl implements MessageInternal {
       message.setOwner(object);
    }
 
+   @Override
+   public Object getUserContext(Object key) {
+      return message.getUserContext(key);
+   }
+
+   @Override
+   public void setUserContext(Object key, Object value) {
+      message.setUserContext(key, value);
+   }
 }
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenwireMessage.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenwireMessage.java
@@ -516,4 +516,14 @@ public class OpenwireMessage implements Message {
    public int usageDown() {
       return 0;
    }
+
+   @Override
+   public Object getUserContext(Object key) {
+      return null;
+   }
+
+   @Override
+   public void setUserContext(Object key, Object value) {
+
+   }
 }
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerTest.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerTest.java
@@ -297,6 +297,16 @@ public class ScheduledDeliveryHandlerTest extends Assert {
    class FakeMessage implements Message {
 
       @Override
+      public Object getUserContext(Object key) {
+         return null;
+      }
+
+      @Override
+      public void setUserContext(Object key, Object value) {
+
+      }
+
+      @Override
       public String getProtocolName() {
          // should normally not be visible in GUI
          return getClass().getName();
--- a/examples/features/standard/opentracing/pom.xml
+++ b/examples/features/standard/opentracing/pom.xml
@@ -1,0 +1,201 @@
+<?xml version='1.0'?>
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+   <modelVersion>4.0.0</modelVersion>
+
+   <parent>
+      <groupId>org.apache.activemq.examples.broker</groupId>
+      <artifactId>jms-examples</artifactId>
+      <version>2.21.0-SNAPSHOT</version>
+   </parent>
+
+   <artifactId>opentracing</artifactId>
+   <packaging>jar</packaging>
+   <name>ActiveMQ Artemis Broker Plugin Example</name>
+
+   <properties>
+      <activemq.basedir>${project.basedir}/../../../..</activemq.basedir>
+   </properties>
+
+   <dependencies>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-jms-client-all</artifactId>
+         <version>${project.version}</version>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-server</artifactId>
+         <version>${project.version}</version>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-amqp-protocol</artifactId>
+         <version>${project.version}</version>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.qpid</groupId>
+         <artifactId>qpid-jms-client</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-api</artifactId>
+         <version>1.10.0</version>
+      </dependency>
+      <dependency>
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-sdk</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-semconv</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-sdk-extension-autoconfigure</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-exporter-logging</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-exporter-otlp</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-exporter-jaeger</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-exporter-zipkin</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-sdk-testing</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>junit</groupId>
+         <artifactId>junit</artifactId>
+         <scope>test</scope>
+      </dependency>
+      <!-- https://mvnrepository.com/artifact/org.assertj/assertj-core -->
+      <dependency>
+         <groupId>org.assertj</groupId>
+         <artifactId>assertj-core</artifactId>
+         <version>3.22.0</version>
+         <scope>test</scope>
+      </dependency>
+      <dependency>
+         <groupId>org.mockito</groupId>
+         <artifactId>mockito-core</artifactId>
+         <scope>test</scope>
+      </dependency>
+   </dependencies>
+
+   <build>
+      <plugins>
+         <plugin>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>artemis-maven-plugin</artifactId>
+            <executions>
+               <execution>
+                  <id>create</id>
+                  <phase>verify</phase>
+                  <configuration>
+                     <!-- The broker plugin will install this library on the server's classpath -->
+                     <libList>
+                        <arg>org.apache.activemq.examples.broker:opentracing:${project.version}</arg>
+                     </libList>
+                     <libListWithDeps>
+                        <arg>io.opentelemetry:opentelemetry-api:${opentelemetry.version}</arg>
+                        <arg>io.opentelemetry:opentelemetry-sdk:${opentelemetry.version}</arg>
+                        <arg>io.opentelemetry:opentelemetry-semconv:${opentelemetry-alpha.version}</arg>
+                        <arg>io.opentelemetry:opentelemetry-sdk-extension-autoconfigure:${opentelemetry-alpha.version}
+                        </arg>
+                        <arg>io.opentelemetry:opentelemetry-exporter-logging:${opentelemetry.version}</arg>
+                        <arg>io.opentelemetry:opentelemetry-exporter-otlp:${opentelemetry.version}</arg>
+                        <arg>io.opentelemetry:opentelemetry-exporter-jaeger:${opentelemetry.version}</arg>
+                        <arg>io.opentelemetry:opentelemetry-exporter-zipkin:${opentelemetry.version}</arg>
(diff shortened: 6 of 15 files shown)


nbrendah:
It works very well and each time you build the example, if Zipkin is running at the provided url, then you will see the traces.

image

nbrendah:
I am looking into cause of the build failure and also updating the licensees in my files

examples/features/standard/broker-plugin/src/main/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPlugin.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public void afterSend(Transaction tx, Message message, boolean direct, boolean noAutoCreateQueue,
+                          RoutingStatus result) throws ActiveMQException {
+        span.end();
+    }
+
+    public static Tracer initOpenTracing() {
+        ZipkinSpanExporter zipkinExporter =
```
clebertsuconic, line 61: Is Zipkin the only possibility for open tracing?


Isn't there any API within Open Telemetry that would replace this?


I thought we talked about using a more standard library.

nbrendah: There are very many tracing exporters @clebertsuconic 
actually this takes to using auto-configuration which makes allows the user of the plugin to specify the tracing exporter.

nbrendah: hello @clebertsuconic, i have modified the functionality to be auto-configured other than manually as before.
By default the exports are transported to the `OTLP server` running at `localhost/127.0.0.1:4317` 
You can try running it your self through `docker run otel/opentelemetry-collector` 
For other ways of running this, you can checkout https://opentelemetry.io/docs/collector/getting-started/ and https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md

examples/features/standard/broker-plugin/pom.xml
```diff
@@ -55,6 +55,38 @@ under the License.
...
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-exporter-otlp</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-exporter-jaeger</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>io.opentelemetry</groupId>
+         <artifactId>opentelemetry-exporter-zipkin</artifactId>
```
nbrendah, line 88: i don't what to limit the exporters that can be used.  Am passing all available exporter dependencies so that if a use chooses an exporter he/she wants, the Opentelemetry doesn't complain with `ClassNotFound` Exceptions while trying to access resources.

clebertsuconic: Ohh.. I see.. you're light weight in dependencies...

How do you establish the connection then?

clebertsuconic: @nbrendah I think we are in a good place to start packaging and testing this properly... we should be on a call..


There's one issue I see is that you always set a property called "count"... of course that would not work... we need to establish what property we are going to send.


and I believe the span would need a Thread Local?

nbrendah: > How do you establish the connection then?

this is done the sdk during its auto-configuration.  So its upto opentelemetry to make sure the connection is set.  If any error occurs, then its a bug in auto-configuration and we have to report it back to opentelemetry-java

> we need to establish what property we are going to send.

i use count because i need to identify each span uniquely since the `getMessageId()` was returning null.  I thought using the `count` property set in the `BrokerPligin` would for a moment, create some uniqueness in spans.

examples/features/standard/broker-plugin/src/test/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPluginTest.java
```diff
@@ -0,0 +1,68 @@
...
+import static org.junit.Assert.assertNotNull;
+
+public class OpenTracingBrokerPluginTest {
+
+   private final InMemorySpanExporter exporter = InMemorySpanExporter.create();
+   //private SdkTracerProvider sdkTracerProvider;
+   //private Tracer tracer;
+
+   /**
+    * @Before public void setUp() {
```
nbrendah, line 36: please ignore these commented code for now, i am still trying out things with them

examples/features/standard/broker-plugin/src/main/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPlugin.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.activemq.artemis.core.server.plugin.ActiveMQServerPlugin;
+import org.apache.activemq.artemis.core.transaction.Transaction;
+
+public class OpenTracingBrokerPlugin implements ActiveMQServerPlugin {
+
+   private static Tracer tracer = initOpenTracing();
+   private Span span;
+
+   public static Tracer initOpenTracing() {
+      try {
```
nbrendah, line 38: this is where we can pass the properties.  I think we can create a properties file and comment most configurations.  Leave everything to default.  
A user will comment the properties he/she needs to use.

@clebertsuconic @michaelandrepearce How does activemq-artemis handle its properties configuration.  What i mean is that do you have thing like a universal properties loader or anything like that??

nbrendah:
among the tests that i think need to be added, we need to tests that all exporters are working fine.  if a user changes from the default, it must work fine

examples/features/standard/broker-plugin/src/main/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPlugin.java
```diff
@@ -0,0 +1,79 @@
...
+         InputStream input = OpenTracingBrokerPlugin.class.getClassLoader().getResourceAsStream("tracing.properties");
+         if (input == null) {
+            throw new NullPointerException("Unable to find tracing.properties file");
+         }
+         Properties prop = new Properties(System.getProperties());
+         prop.load(input);
+         System.setProperties(prop);
```
nbrendah, line 48: @clebertsuconic 
This is how i have added the `.properties ` which will contain all opentelemetry configurations and customization
we can specify which exporter to use and its corresponding host and port.

examples/features/standard/broker-plugin/src/test/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPluginTest.java
```diff
@@ -18,51 +18,46 @@
...
-    * .buildAndRegisterGlobal()
-    * .getTracer("org.apache.activemq.artemis.jms.example.OpenTracingBrokerPlugin");
-    * }
-    **/
+   @Mock
+   OpenTracingBrokerPlugin plugin = new OpenTracingBrokerPlugin();
 
    @Test
-   public void testSpans() {
+   public void assertInitOpenTracing() {
```
nbrendah, line 43: @franz1981 running this test shows me and error 
```
Failed to read or configure the org.jboss.logmanager.LogManager
java.lang.NoClassDefFoundError: org/apache/activemq/artemis/shaded/org/jboss/logmanager/ExtHandler
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1017)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:800)
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:698)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:621)
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:579)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	at java.base/java.lang.Class.forName0(Native Method)
	at java.base/java.lang.Class.forName(Class.java:398)
	at org.jboss.logmanager.config.AbstractPropertyConfiguration.<init>(AbstractPropertyConfiguration.java:72)
	at org.jboss.logmanager.config.HandlerConfigurationImpl.<init>(HandlerConfigurationImpl.java:51)
	at org.jboss.logmanager.config.LogContextConfigurationImpl.addHandlerConfiguration(LogContextConfigurationImpl.java:143)
	at org.jboss.logmanager.PropertyConfigurator.configureHandler(PropertyConfigurator.java:673)
	at org.jboss.logmanager.PropertyConfigurator.configureLogger(PropertyConfigurator.java:583)
	at org.jboss.logmanager.PropertyConfigurator.configure(PropertyConfigurator.java:521)
	at org.jboss.logmanager.PropertyConfigurator.configure(PropertyConfigurator.java:97)
	at org.jboss.logmanager.LogManager.readConfiguration(LogManager.java:170)
	at org.jboss.logmanager.LogManager.readConfiguration(LogManager.java:132)
	at java.logging/java.util.logging.LogManager.readPrimordialConfiguration(LogManager.java:445)
	at java.logging/java.util.logging.LogManager$2.run(LogManager.java:394)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.logging/java.util.logging.LogManager.ensureLogManagerInitialized(LogManager.java:382)
	at java.logging/java.util.logging.LogManager.getLogManager(LogManager.java:430)
	at java.logging/java.util.logging.Logger.demandLogger(Logger.java:648)
	at java.logging/java.util.logging.Logger.getLogger(Logger.java:717)
	at java.logging/java.util.logging.Logger.getLogger(Logger.java:701)
	at io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdkBuilder.<clinit>(AutoConfiguredOpenTelemetrySdkBuilder.java:43)
	at io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk.builder(AutoConfiguredOpenTelemetrySdk.java:37)
	at io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk.initialize(AutoConfiguredOpenTelemetrySdk.java:29)
	at org.apache.activemq.artemis.jms.example.OpenTracingBrokerPlugin.initOpenTracing(OpenTracingBrokerPlugin.java:50)
	at org.apache.activemq.artemis.jms.example.OpenTracingBrokerPlugin.<clinit>(OpenTracingBrokerPlugin.java:37)
	at org.apache.activemq.artemis.jms.example.OpenTracingBrokerPluginTest.<init>(OpenTracingBrokerPluginTest.java:39)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
	at org.junit.runners.BlockJUnit4ClassRunner.createTest(BlockJUnit4ClassRunner.java:250)
	at org.junit.runners.BlockJUnit4ClassRunner.createTest(BlockJUnit4ClassRunner.java:260)
	at org.junit.runners.BlockJUnit4ClassRunner$2.runReflectiveCall(BlockJUnit4ClassRunner.java:309)
	at org.junit.runners.BlockJUnit4ClassRunner.methodBlock(BlockJUnit4ClassRunner.java:306)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater$1.execute(IdeaTestRunner.java:38)
	at com.intellij.rt.execution.junit.TestsRepeater.repeat(TestsRepeater.java:11)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:35)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:235)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:54)
Caused by: java.lang.ClassNotFoundException: org.apache.activemq.artemis.shaded.org.jboss.logmanager.ExtHandler
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:581)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:178)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
	... 61 more

Process finished with exit code 0
```

examples/features/standard/broker-plugin/src/main/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPlugin.java
```diff
@@ -0,0 +1,90 @@
...
+                          Transaction tx,
+                          Message message,
+                          boolean direct,
+                          boolean noAutoCreateQueue) throws ActiveMQException {
+      SpanBuilder spanBuilder = GlobalOpenTelemetry.getTracer(OpenTracingBrokerPlugin.class.getName())
+         .spanBuilder(OPERATION_NAME)
+         .setAttribute("message", message.toString())
+         .setSpanKind(SpanKind.SERVER);
+      Span span = spanBuilder.startSpan();
+      threadLocal.set(span);
```
nbrendah, line 71: @franz1981 I thought we can add the ThreadLocal or span instance to the messages itself like `message.putObjectProperty("SpanThread", threadLocal);`  But i get an exception `java.lang.ThreadLocal is not a valid property type: org.apache.activemq.artemis.api.core.ActiveMQPropertyConversionException` showing that `ThreadLocal` is not accepted as an object.

and then in the afterMethod we get it like` (ThreadLocal) message.getObjectProperty();` and finally get the span

nbrendah, line 71: @franz1981 I thought we can add the ThreadLocal or span instance to the messages itself like `message.putObjectProperty("SpanThread", threadLocal);`  But i get an exception `java.lang.ThreadLocal is not a valid property type: org.apache.activemq.artemis.api.core.ActiveMQPropertyConversionException` showing that `ThreadLocal` is not accepted as an object.

and then in the afterMethod we get it like` (ThreadLocal) message.getObjectProperty();` and finally get the span

examples/features/standard/broker-plugin/src/test/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPluginTest.java
nbrendah: i still face the same error and am trying to solve it.

nbrendah:
am also looking into build failure so i think its something we can not too much make a point for now

examples/features/standard/broker-plugin/src/main/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPlugin.java
```diff
@@ -0,0 +1,90 @@
...
+      return sdk;
+   }
+
+   @Override
+   public void beforeSend(ServerSession session,
+                          Transaction tx,
+                          Message message,
+                          boolean direct,
+                          boolean noAutoCreateQueue) throws ActiveMQException {
+      SpanBuilder spanBuilder = GlobalOpenTelemetry.getTracer(OpenTracingBrokerPlugin.class.getName())
```
franz1981, line 66:
```suggestion
      SpanBuilder spanBuilder = TRACER
```
`GlobalOpenTelemetry.getTracer(OpenTracingBrokerPlugin.class.getName())` can be declared in the plugin as 
```java
private static final Tracer TRACER = GlobalOpenTelemetry.getTracer(OpenTracingBrokerPlugin.class.getName());
```

examples/features/standard/broker-plugin/src/main/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPlugin.java
franz1981: This is correct, because messages are not supposed to contain things that cannot be encoded and sent to client(s), and `ThreadLocal` although temporary isn't an object instance that we wish to send to client(s).

examples/features/standard/broker-plugin/src/main/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPlugin.java
```diff
@@ -0,0 +1,90 @@
...
+      threadLocal.set(span);
+
+      /**
+      span = tracer.spanBuilder(String.format("message: %s", message.getStringProperty("count"))).setSpanKind(SpanKind.PRODUCER).startSpan();
+      span.addEvent(String.format("message: %s", message.getStringProperty("count")));
+       **/
+   }
+
+   @Override
+   public void afterSend(Transaction tx,
```
franz1981, line 80: I suggest to override `public void onSendException` as well or we risk to leak `Span` in case of exceptions

examples/features/standard/broker-plugin/src/main/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPlugin.java
```diff
@@ -0,0 +1,90 @@
...
+import io.opentelemetry.sdk.autoconfigure.AutoConfiguredOpenTelemetrySdk;
+import org.apache.activemq.artemis.api.core.ActiveMQException;
+import org.apache.activemq.artemis.api.core.Message;
+import org.apache.activemq.artemis.core.postoffice.RoutingStatus;
+import org.apache.activemq.artemis.core.server.ServerSession;
+import org.apache.activemq.artemis.core.server.plugin.ActiveMQServerPlugin;
+import org.apache.activemq.artemis.core.transaction.Transaction;
+
+public class OpenTracingBrokerPlugin implements ActiveMQServerPlugin {
+   private ThreadLocal threadLocal = new ThreadLocal();
```
franz1981, line 38: I think we have several options here:

- `ThreadLocal` must be declared as `private static final ThreadLocal` to benefit from important optimizations by the JVM
- `ThreadLocal::remove` can be quite expensive and subsequent `ThreadLocal::set`  would create new `ThreadLocalMap::Entry`: saving a leak to happen seems important, but I would search a different way to solve this problem

For the latter @clebertsuconic maybe we can change the plugin API in order to bring context `Object` all around a-la C language (with function pointers)
eg
```java
   default Object beforeSend(ServerSession session, Transaction tx, Message message, boolean direct, boolean noAutoCreateQueue, Object context) throws ActiveMQException {
          this.beforeSend(session, tx, message, direct, noAutoCreateQueue);
          return context;
   }

   default void afterSend(ServerSession session, Transaction tx, Message message, boolean direct, boolean noAutoCreateQueue, RoutingStatus result, Object context) throws ActiveMQException {
      this.afterSend(tx, message, direct, noAutoCreateQueue, result);
   }

   default void onSendException(ServerSession session, Transaction tx, Message message, boolean direct, boolean noAutoCreateQueue, Exception e, Object context) throws ActiveMQException {
       this.onSendException(session, tx, message, direct, noAutoCreateQueue, e);
   }
   ```
This would allows using the `Span` as the `context` (ie state) to be propagated around without needing any extra map/thread local/....

File: examples/features/standard/broker-plugin/src/main/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPlugin.java [hunk 11]
```diff
@@ -0,0 +1,94 @@
...
+
+   @Override
+   public void onSendException(ServerSession session,
+                               Transaction tx,
+                               Message message,
+                               boolean direct,
+                               boolean noAutoCreateQueue,
+                               Exception e) throws ActiveMQException {
+      threadLocal.get().setStatus(StatusCode.ERROR).recordException(e);
+      threadLocal.remove();
```
nbrendah, line 92: @franz1981 does this guarantee that the object is gonna be removed in case of any exception?

examples/features/standard/broker-plugin/src/test/java/org/apache/activemq/artemis/jms/example/OpenTracingBrokerPluginTest.java
nbrendah: @franz1981 I have made some changes in the test logic but the error i shared above is blocking still blocking me.

clebertsuconic:
I am asking this a draft just to avoid merging it by accident.

gls-luis-alves:
@nbrendah is it possible to pass some message header with and existing trace-id? So we have a trace from microservice A ----> artemis -----> microservice B ....etc...

jbertram:
Look at the [OpenTelemetry Example](https://github.com/apache/activemq-artemis-examples/tree/main/examples/features/standard/opentelemetry). You can use a broker plugin (e.g. [here](https://github.com/apache/activemq-artemis-examples/blob/main/examples/features/standard/opentelemetry/src/main/java/org/apache/activemq/artemis/jms/example/OpenTelemetryPlugin.java)) to take a property from the message and use it however you wish.

gls-luis-alves:
Thanks @jbertram. I will give a try when I have time. I was looking at [context-propagation-between-http-requests](https://opentelemetry.io/docs/languages/java/instrumentation/#context-propagation-between-http-requests) so I could understand how this works under the hood. It's great that Artemis is integrating this feature.

svanschalkwyk:
Is there an example of this working? Should a specific plugin be installed for this to work? 
using Docker with image:apache/activemq-artemis:2.33.0
I have `      OTEL_SERVICE_NAME: activemq-service 
      OTEL_TRACES_EXPORTER: otlp
      OTEL_METRICS_EXPORTER: otlp
      OTEL_LOGS_EXPORTER: none
      OTEL_EXPORTER_OTLP_ENDPOINT: http://otel-collector:4317
      OTEL_EXPORTER_OTLP_TRACES_ENDPOINT: http://otel-collector:4317`
No issues in startup (missing jars etc).

PR #3898, opened 2022-01-07
ARTEMIS-3626 fixed simple syntax error

erwindon:
see https://issues.apache.org/jira/browse/ARTEMIS-3626


jbertram:
The `EmptyStatement` checkstyle module doesn't pick this up since it's on an `import`. That seems weird to me. Will merge ASAP. Thanks!

PR #3904, opened 2022-01-07
ARTEMIS-3627 - support broker.properties for augmenting or supplying …

gtully:
…additional configuration - elements with a name attribute can be configured in collections, the type inferred by the add singular fluent api

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
@@ -16,21 +16,27 @@
  */
 package org.apache.activemq.artemis.core.config.impl;
 
+import java.beans.PropertyDescriptor;
+import java.io.BufferedInputStream;
 import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.io.FileInputStream;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.io.PrintWriter;
 import java.io.Serializable;
 import java.io.StringWriter;
 import java.lang.reflect.Array;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.net.URI;
 import java.net.URL;
 import java.security.AccessController;
 import java.security.PrivilegedExceptionAction;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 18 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -429,8 +444,28 @@ public String getSystemPropertyPrefix() {
...
+   }
+
+   public void setBrokerPropertiesSystemPropertyName(String brokerPropertiesSystemPropertyName) {
+      this.brokerPropertiesSystemPropertyName = brokerPropertiesSystemPropertyName;
+   }
+
    @Override
-   public Configuration parseSystemProperties() throws Exception {
+   public Configuration parseProperties() throws Exception {
+      final String propertiesConfigFileName = System.getProperty(getBrokerPropertiesSystemPropertyName());
```
clebertsuconic, line 457: I think we should pass this through the CLI instead.

We could parse a default broker.properties on the instance/etc folder, and the users would be able to pass another file with a ./artemis run --properties <file>

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -599,7 +599,7 @@ private void internalStart() throws Exception {
          return;
       }
 
-      configuration.parseSystemProperties();
+      configuration.parseProperties();
```
clebertsuconic, line 602: the CLI would pass the properties file name into the ActiveMQServerIMPL.. and it would be passed here?

clebertsuconic:
nice one @gtully 


did you think about yaml as well for the properties replacement?

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
gtully: I am not sure of the benefit of multiple files, but the cli could arrange for the system property to be initialised with a value from a --properties argument for sure.

clebertsuconic: @gtully I didn't mean to ask for multiple properties file.. just the cli argument

gtully:
@brusdev I pushed an update to deal with keys with dots, as dots are the nesting delimiter. It is now possible to surround with '"' or any string identified by a property key.surround
thanks!

gtully:
on the yaml, I don't want to reinvent config, just extend the support for augmenting, I think key value pairs, with keys giving the nesting is potentially intuitive enough to be self describing. 
but the more it is used, the more we can add additional accessors to make it prettier where necessary.
I am thinking something like amqpConnections.mirrors.. and amqpConnections.senders.. that sort of thing may be a nice addition. But what is here is sufficient for now I think.

gtully:
made it such that the mapped collection(key) can retrieve an existing entry, but still won't auto create. collection.key will auto create.

PR #3905, opened 2022-01-07
ARTEMIS-3624 fixed missing version numver in pom file (prevented use of mvn eclipse:eclipse)

erwindon:
see https://issues.apache.org/jira/browse/ARTEMIS-3624

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -340,6 +340,7 @@
          <dependency>
             <groupId>org.apache.kerby</groupId>
             <artifactId>token-provider</artifactId>
+            <version>2.0.1</version>
             <scope>test</scope>
             <exclusions>
                <exclusion>


gemmellr:
The version number was never missing, it was specifically omitted to avoid versioning that particular transitive dependency (which the exclusion was being applied to) differently than other related transitive deps, which this change would then begin to do and indeed already starts with a disparate version.

I raised #3909 with a different change which manages the overall exclusion differently, applying it at the base test dep it transitively relates to, which is itself versioned and thus avoids the issue. 

I'd note also Eclipse can important Maven projects itself directly (Import Existing Maven Project), though in complex cases like Artemis you do end up having to jump some hoops to get things fully functional in some cases. I do most of my work via CLI so I've never bothered to jump those hoops, the basic import has been enough for me.

erwindon:
This PR has been superseded by #3909, therefore closing this one.

PR #3907, opened 2022-01-11
ARTEMIS-3638 Support MQTT 5

jbertram:
MQTT 5 is an OASIS standard which debuted in March 2019. It boasts
numerous improvments over its predecessor (i.e. MQTT 3.1.1) which will
benefit users. These improvements are summarized in the specification
at:
https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901293

The specification describes all the behavior necessary for a client or
server to conform. The spec is highlighted with special "normative"
conformance statements which distill the descriptions into concise terms.
The specification provides a helpful summary of all these statements. See:
https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901292

This commit implements all of the mandatory elements from the
specification and provides tests which are identified using the
corresponding normative conformance statement. *All* normative
conformance statements either have an explicit test or are noted in
comments with an explanation of why an explicit test doesn't exist. See
`org.apache.activemq.artemis.tests.integration.mqtt5` for all those
details.

This commit also includes documentation about how to configure
everything related to the new MQTT 5 features.


jbertram:
I apologize for the length of this PR. I realize it is a lot to review. However, I found that I couldn't break it up as changes in one functional area quickly bled over into other areas. I'd love any feedback.

gtully:
looks good, lots of new tests that validate. great work.
I was peeking for some SASL bits, I guess that can layer nicely on this?

jbertram:
@gtully, as noted in the documentation (which you probably saw) there currently are no supported authentication methods for the "Enhanced Authentication" which was introduced in MQTT 5. However, the scaffolding is there so it should be relatively easy to integrate (assuming you're familiar with SASL). See `org.apache.activemq.artemis.core.protocol.mqtt.MQTTProtocolHandler#handleAuth`.

Given your familiarity with SASL I was hoping to get your help here.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTUtil.java
```diff
@@ -107,120 +155,339 @@ public static String convertMQTTAddressFilterToCoreRetain(String filter, Wildcar
...
-                                                    int qos) {
+   private static ICoreMessage createServerMessage(MQTTSession session, SimpleString address, MqttPublishMessage mqttPublishMessage) {
       long id = session.getServer().getStorageManager().generateID();
 
-      CoreMessage message = new CoreMessage(id, DEFAULT_SERVER_MESSAGE_BUFFER_SIZE, session.getCoreMessageObjectPools());
+      CoreMessage message = new CoreMessage(id, mqttPublishMessage.fixedHeader().remainingLength(), session.getCoreMessageObjectPools());
       message.setAddress(address);
-      message.putBooleanProperty(MQTT_MESSAGE_RETAIN_KEY, retain);
-      message.putIntProperty(MQTT_QOS_LEVEL_KEY, qos);
+      message.putBooleanProperty(MQTT_MESSAGE_RETAIN_KEY, mqttPublishMessage.fixedHeader().isRetain());
```
clebertsuconic, line 163: Did you consider the possibility to create a MQTTMessage?

If you're not doing that now, and we plan to eventually do it.. using application properties from CORE wouldn't make it a bit more difficult to make such transition?

jbertram: I didn't seriously consider implementing a native MQTTMessage at this point. My goal in this PR was to correctly implement all the new MQTT 5 features.

We've always used message properties when converting to/from core for MQTT-specific details. This PR isn't changing that. I'm not clear on how this would make the transition to a native MQTTMessage implementation more difficult later. Can you elaborate on that?

jbertram:
I just realized that I used the wrong Jira ID in the commit message and the branch name. I corrected the commit message and the PR title, but I left the branch name as-is to avoid having to create a whole new PR (and lose the comment history).

gtully:
@jbertram looking to resolve the merge conflict via the gh ui, seems not to have the desired effect. sorry! Maybe you do a rebase and resolve against main and force push from your local branch and whack my merge from main. I won't attempt that again any time soon :-)

jbertram:
@gtully, I'll take a look. No worries!

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTPublishManager.java
```diff
@@ -284,11 +358,110 @@ private void sendServerMessage(int messageId, ICoreMessage message, int delivery
...
+         }
+      }
+
+      int remainingLength = MQTTUtil.calculateRemainingLength(address, mqttProperties, payload);
+      MqttFixedHeader header = new MqttFixedHeader(MqttMessageType.PUBLISH, redelivery, MqttQoS.valueOf(qos), isRetain, remainingLength);
+      MqttPublishVariableHeader varHeader = new MqttPublishVariableHeader(address, messageId, mqttProperties);
+      MqttPublishMessage publish = new MqttPublishMessage(header, varHeader, payload);
+
+      if (session.is5()) {
+         int size = MQTTUtil.calculateMessageSize(publish);
```
gtully, line 396: maybe calculate size if maxSize != 0 to avoid the computation if unlimited

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSession.java
```diff
@@ -212,4 +234,42 @@ public CoreMessageObjectPools getCoreMessageObjectPools() {
       return coreMessageObjectPools;
    }
 
+   public boolean is5() {
+      return five;
+   }
+
+   public void set5(boolean five) {
```
gtully, line 241: small point, but I am not sure on five and boolean, it may be better to keep it to int and int > 3, I imagine there will be mqtt 6  etc... and they will build on each other.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolHandler.java
```diff
@@ -159,89 +176,180 @@ public void act(MqttMessage message) {
...
+         /* [MQTT-3.1.2-2] Reject unsupported clients. */
+         int packetVersion = connect.variableHeader().version();
+         if (packetVersion != MqttVersion.MQTT_3_1.protocolLevel() &&
+            packetVersion != MqttVersion.MQTT_3_1_1.protocolLevel() &&
+            packetVersion != MqttVersion.MQTT_5.protocolLevel()) {
+
+            {
+               /*
+                * The return code is different here for 3.1.1 vs 5, but since the broker supports *both* it's not clear
+                * which one should be chosen. Perhaps an acceptor property could be set to decide. Leaving the existing
```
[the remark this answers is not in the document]
jbertram: I think that's fair. I'll push that change.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
```diff
@@ -185,45 +243,83 @@ public void handleBuffer(RemotingConnection connection, ActiveMQBuffer buffer) {
    @Override
    public void addChannelHandlers(ChannelPipeline pipeline) {
       pipeline.addLast(MqttEncoder.INSTANCE);
-      pipeline.addLast(new MqttDecoder(MQTTUtil.MAX_MESSAGE_SIZE));
+      /*
+       * If we use the value from getMaximumPacketSize() here anytime a client sends a packet that's too large it
+       * will receive a DISCONNECT with a reason code of 0x81 instead of 0x95 like it should according to the spec.
```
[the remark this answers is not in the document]
jbertram: My goal with this commit was to be as close to the spec as possible. I wanted to "get it right" first and then later "make it fast" if necessary. In any event, computing the size of the message is pretty cheap since it's just a small bit of simple arithmetic. See `org.apache.activemq.artemis.core.protocol.mqtt.MQTTUtil#calculateMessageSize`.

I'm certainly open to sending a PR to Netty for this, but that it outside the scope of this PR.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTPublishManager.java
jbertram: Good idea. I'll push that change.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTProtocolManager.java
```diff
@@ -185,45 +243,83 @@ public void handleBuffer(RemotingConnection connection, ActiveMQBuffer buffer) {
...
+         readByte(buf) != 81 ||   // Q
+         readByte(buf) != 73 ||   // I
+         readByte(buf) != 115 ||  // s
+         readByte(buf) != 100 ||  // d
+         readByte(buf) != 112)) { // p
+         return false;
+      }
+
+      // Protocol Version
+      b = readByte(buf);
```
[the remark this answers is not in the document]
jbertram: You're right. I didn't catch this because I don't actually have a low-level client implementation where I can pass a bad version and validate the response from the broker. I'll omit the version check here so the spec-compliant response can be returned later.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSession.java
jbertram: That's a fair point. However, I'd prefer to leave it as a `boolean` for now and change it later if/when necessary. It will be just as easy to change it later as it would be to change it now. In my opinion the benefit of not changing it now is that using a `boolean` is more clear.

clebertsuconic:
The only thing I don't like on this PR is the "leakage" of getMqttSessionScanInterval on ConfigurationImpl.

I would prefer it being a bean util property on the MQTT Protocol Manager. Make the scanInterval as part of the default configuration so it's easier for users to see it in there.


Do you really need it on the Configurationimpl / Configuration or would you be able to make it part of MQTTProtocolManager and a beanutil property?

jbertram:
There are a couple of failures on the test-suite that I need to resolve. I'll move this to draft for now.

jbertram:
I fixed the failing tests in the test-suite. This is ready to be merged.

PR #3909, opened 2022-01-11
ARTEMIS-3624: rework dep mangement entry for minikdc dep

gemmellr:
Alternative to #3905, reworks the minikdc dep handling more generally, managing that deps version and applying the previous exclusion via the base dep rather than at the importing transitive dep as in the prior non-version-related management.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -159,6 +159,7 @@
       <!-- used on tests -->
       <groovy.version>2.5.10</groovy.version>
       <vertx.version>3.9.4</vertx.version>
+      <hadoop.minikdc.version>3.3.1</hadoop.minikdc.version>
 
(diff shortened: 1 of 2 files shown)


erwindon:
@gemmellr there was more behind this than I imagined... shall I revoke my PR (#3905)?

gemmellr:
I dont think the other PR should be applied so if this works for you (I forgot to actually try running eclipse:eclipse, I'd need a fresh clone to do so) then I'd say yes.

erwindon:
> I forgot to actually try running eclipse:eclipse, I'd need a fresh clone to do so

I just did that, now `mvn eclipse:eclipse` works ok, thx!

PR #3911, opened 2022-01-11
ARTEMIS-3636 LinkedListImpl leak on mesage consume error

jbertram:


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientConsumerImpl.java
```diff
@@ -710,10 +710,8 @@ public void clear(boolean waitForOnMessage) throws ActiveMQException {
       synchronized (this) {
          // Need to send credits for the messages in the buffer
 
-         Iterator<ClientMessageInternal> iter = buffer.iterator();
-
-         while (iter.hasNext()) {
-            try {
+         try (LinkedListIterator<ClientMessageInternal> iter = buffer.iterator()) {
```
clebertsuconic, line 713: Funny! I had the exact same fix on my laptop ready to send a PR for :) 

if I merge your PR, it won't even show a difference

PR #3913, opened 2022-01-13
ARTEMIS-3186 enable create queue more often

erwindon:
See https://issues.apache.org/jira/browse/ARTEMIS-3186

The PR has 5 commits. For the purpose of the review these are still separate. This can of course be collapsed as needed.
1) remove unused parameter "workspace" from 3 `shouldShow...` functions. only 3 of the 7 functions in this category still had this unused parameter. unrelated, but in the same technical area.
2) The default routing type when creating a queue is often overlooked (in my experience), and one ends up with the wrong type. Reset the default to "no selection". The "Create Queue" button is disabled until the choice is made (next to the existing condition that a name must be provided). see also point (5)
3) All 7 `shouldShow...` functions have been reformatted so that each condition is visible on a separate line. Their logic has not been changed. see point (4) for the exception.
4) Updated the "Create Queue" navigation logic so that "Create Queue" is visible on 3 levels in the "tree". 1st is the traditional address level. 2nd (new) is on the "queues" level. 3rd (new) is on the "anycast" or "multicast" level. Since the "queues"/"anycast"/"multicast" levels in the tree do not represent JMX objects, the selection logic is now based on the "selection/folderNames" property, which is always available. This has a small effect on the callback-function as well, as that now must use that same information.
5) When a queue is created while "anycast"/"multicast" is selected (so in the 3rd case), that is also the default for the routing type that is used for the queue creation.

### DIFF
--- a/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/createAddress.js
+++ b/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/createAddress.js
@@ -57,7 +57,7 @@ var Artemis;
                      <div class="col-sm-offset-2 col-sm-10">
                          <button type="submit" class="btn btn-primary"
                                  ng-click="$ctrl.createAddress($ctrl.addressName, $ctrl.routingType)"
-                                 ng-disabled="!$ctrl.addressName">Create Address
+                                 ng-disabled="!$ctrl.addressName || !$ctrl.routingType">Create Address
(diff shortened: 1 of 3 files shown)


jbertram:
As on #3923 I think these commits should be squashed, and you can add a bulleted list of changes in the commit message.

erwindon:
rebased as needed and squashed as requested

PR #3914, opened 2022-01-17
ARTEMIS-3645 Support broker balancer cache persistence

brusdev:


jbertram:
Nice work, @brusdev. My only concern is with the XML/XSD change. Any configuration using the existing `cache-timeout` element  will fail to parse upon upgrade. I sympathize with your change to group all the `cache` elements together, but we need to maintain backwards compatibility. I see two options:

 1. Keep the existing `cache-timeout`, but deprecate it and add the new `cache` subsection.
 2. Add a new element like `cache-persistence` on the same level as the existing `cache-timeout`.

jbertram:
Given that [the documentation](https://activemq.apache.org/components/artemis/documentation/latest/broker-balancers.html) **boldly** states:

> This feature is still EXPERIMENTAL and not meant to be run in production yet. Furthermore, its configuration can change until declared as officially stable.

I think changing the config is fine. I'll go ahead and merge this.

PR #3915, opened 2022-01-17
NO-JIRA Fix activemq5-unit-tests with SSL

brusdev:


tests/activemq5-unit-tests/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -796,9 +796,19 @@ public ConnectorInfo(URI bindAddress, SpringSslContext context) throws URISyntax
...
          } else {
-            String baseUri = "tcp://" + host + ":" + port + "?" + TransportConstants.SSL_ENABLED_PROP_NAME + "=true&" + TransportConstants.KEYSTORE_PATH_PROP_NAME + "=" + (context == null ? defaultKeyStore : context.getKeyStore()) + "&" + TransportConstants.KEYSTORE_PASSWORD_PROP_NAME + "=" + (context == null ? defaultKeyStorePassword : context.getKeyStorePassword()) + "&" + TransportConstants.KEYSTORE_PROVIDER_PROP_NAME + "=" + (context == null ? defaultKeyStoreType : context.getKeyStoreType());
+            String keyStoreType = context == null ? defaultKeyStoreType : context.getKeyStoreType();
+            if (keyStoreType.equals("jks")) {
+               keyStoreType = "JKS";
+            }
+
+            String baseUri = "tcp://" + host + ":" + port + "?" + TransportConstants.SSL_ENABLED_PROP_NAME + "=true&" + TransportConstants.KEYSTORE_PATH_PROP_NAME + "=" + (context == null ? defaultKeyStore : context.getKeyStore()) + "&" + TransportConstants.KEYSTORE_PASSWORD_PROP_NAME + "=" + (context == null ? defaultKeyStorePassword : context.getKeyStorePassword()) + "&" + TransportConstants.KEYSTORE_PROVIDER_PROP_NAME + "=" + keyStoreType;
```
gemmellr, line 804: This appears to be jumping hoops to manage case for the value of 'key store type', and then using it to set the 'provider type' config property, which is actually for something else entirely. (Presumably the original bit being changed was prior to 186481bbe8e79274eb009f63052b1d0528d0249d)

Would it not be better to change it to just stop setting the provider property at all and use the keystore type config property?

gemmellr (commented):
I presume some other area is referencing the created 'dummy' files and failing as they arent present?

brusdev:
@gemmellr thanks for your review

PR #3917, opened 2022-01-17
NO-JIRA Fix activemq5-unit-tests with byteman

brusdev:


tests/activemq5-unit-tests/pom.xml
```diff
@@ -41,7 +41,7 @@
       <jasypt-version>1.9.3</jasypt-version>
       <directory-version>2.0.0-M6</directory-version>
       <activeio-core-version>3.1.4</activeio-core-version>
-      <byteman.version>2.2.1</byteman.version>
+      <byteman.version>4.0.18</byteman.version>
```
gemmellr, line 44: There is another optional module using byteman, with its own copy of this property, set to a different 4.0.x version. They should probably be made consistent e.g move property up a pom so they share, or have these tests moved to that module and this dropped.

tests/activemq5-unit-tests/src/test/java/org/apache/activemq/transport/failover/FailoverConsumerOutstandingCommitTest.java
```diff
@@ -177,6 +177,7 @@ public void run() {
             targetMethod = "processCommitTransactionOnePhase",
             targetLocation = "ENTRY",
             action = "org.apache.activemq.transport.failover.FailoverConsumerOutstandingCommitTest.stopServerInTransaction();return null")})
+   @Ignore("This test is using Byteman. It's not really accurate.")
```
gemmellr, line 180: If its not really accurate, and is now being permanently disabled, I would just delete it, it can be found in history if needed.

brusdev: It is not so easy searching from history. I would keep it for now. We could remove all ignored tests before the next major release.

gemmellr: Searching from history is fairly trivial. Though based on its prior state and this fix I dont really see anyone actually searching for it.

brusdev: Fair enough, I'll drop them

PR #3918, opened 2022-01-17
ARTEMIS-3649 Fix zero prefetch OpenWire consumers

brusdev:


gtully (approved):
that looks cleaner.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
```diff
@@ -232,11 +232,24 @@ public ConsumerId getId() {
...
-   public void acquireCredit(int n) {
+   public void acquireCredit(int n, boolean delivered) {
       if (messagePullHandler.get() != null) {
          //don't acquire any credits when the pull handler controls it!!
          return;
       }
+
+      if (delivered) {
+         deliveredAcksCreditExtension += n;
+      } else if (deliveredAcksCreditExtension > 0) {
```
gtully, line 243: that is better, now the normal deliveredAcksCreditExtension == 0 branch is clear

PR #3919, opened 2022-01-17
ARTEMIS-3650: rework assembly creation to simplify

gemmellr:


gemmellr:
This produces almost the same assembly lib dir as before except a few small changes. It adds a tiny transitive dependency of guava that was being excluded (I excluded its others, mostly annotation related, but excluding this one causes compilation failure with errorprone), removes the artemis-commons test jar that shouldnt have been there, removes the artemis-rest jar (not usuable on its own anyway as it has many already-omitted deps; there are steps in the docs to emit the needed bits that will be unaffected by the assembly change), and finally adds in another netty transitive dependency that was previously being omitted due to lack of explicit inclusion much like the problem in #3887 (this change would have prevented that from occurring).

gemmellr:
There are further changes that it could be made later, e.g. the change notes the use of jetty-all needed a slight workaround rather than being treated as any regular transitive dep (it brings in the duplicate individual dependencies for everything it contains, or possibly even more), and was actually described as not being for end use by the jetty folks. I think shifting to the needed individual jetty bits should be a separate change though so didnt look to include that. There are possibly also some deps the assembly module is still handling that should be covered in related modules themselves instead.

PR #3921, opened 2022-01-20
NO-JIRA Fix required java version in Using server docs

sergey-vaysman:
From artemis 2.20 java 11 is required. Fixed a note on Using server page.


jbertram:
Thanks for the PR. However, I already updated the related docs. See f7bb4c754a6ab9aa8bdf1b51685f71739a747eda.

PR #3922, opened 2022-01-21
ARTEMIS-3652 - fix test cleanup, allow embedded broker properties to …

gtully:
…be configured, follows up on ARTEMIS-3627

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/ElasticQueueTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/ElasticQueueTest.java
@@ -34,6 +34,7 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
 import org.apache.activemq.artemis.api.core.QueueConfiguration;
@@ -73,6 +74,7 @@ public class ElasticQueueTest extends ActiveMQTestBase {
    static final SimpleString qNameSimple = SimpleString.toSimpleString(qName);
 
    final int base_port = 61616;
+   final Stack<Worker> workers = new Stack<>();
    final Stack<EmbeddedActiveMQ> nodes = new Stack<>();
    private final String balancerConfigName = "role_name_sharder";
 
@@ -86,6 +88,10 @@ public class ElasticQueueTest extends ActiveMQTestBase {
          }
       }
       nodes.clear();
+      for (Worker worker : workers) {
+         worker.done.set(true);
+      }
+      workers.clear();
       executorService.shutdownNow();
    }
 
@@ -98,8 +104,8 @@ public class ElasticQueueTest extends ActiveMQTestBase {
          }
          builder.append("amqp://localhost:").append(port_start++);
       }
-      // fast reconnect, randomize to get to all brokers and timeout sends that block on no credit
-      builder.append(")?failover.randomize=true&failover.maxReconnectAttempts=1&jms.sendTimeout=" + 1000);
+      // fast reconnect, randomize to get to all brokers and timeout sends that block on no credit once connected
+      builder.append(")?failover.randomize=true&failover.maxReconnectAttempts=0&jms.sendTimeout=" + 1000);
       return builder.toString();
    }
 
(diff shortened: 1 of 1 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/embedded/EmbeddedActiveMQ.java
```diff
@@ -144,9 +155,11 @@ protected void initStart() throws Exception {
...
       }
 
-      URL brokerPropertiesFromClasspath = this.getClass().getClassLoader().getResource(ActiveMQDefaultConfiguration.BROKER_PROPERTIES_SYSTEM_PROPERTY_NAME);
-      if (brokerPropertiesFromClasspath != null) {
-         activeMQServer.setProperties(new File(brokerPropertiesFromClasspath.toURI()).getAbsolutePath());
+      if (propertiesResourcePath != null) {
+         URL brokerPropertiesFromClasspath = this.getClass().getClassLoader().getResource(propertiesResourcePath);
+         if (brokerPropertiesFromClasspath != null) {
+            activeMQServer.setProperties(new File(brokerPropertiesFromClasspath.toURI()).getAbsolutePath());
+         }
```
gemmellr, line 162: Feels like this should really be in a separate commit against ARTEMIS-3627 specifically, where the original logic was only just added.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/ElasticQueueTest.java
```diff
@@ -398,7 +407,7 @@ private void prepareNodesAndStartCombinedHeadTail() throws Exception {
          configuration.setName("Node" + nodeId);
          configuration.setBrokerInstance(new File(getTestDirfile(), configuration.getName()));
          configuration.addAcceptorConfiguration("tcp", "tcp://localhost:" + (base_port + (nodeId)) + "?redirect-to=" + balancerConfigName + ";amqpCredits=1000;amqpMinCredits=300");
-         nodes.add(new EmbeddedActiveMQ().setConfiguration(configuration));
+         nodes.add(new EmbeddedActiveMQ().setConfiguration(configuration).setPropertiesResourcePath(null));
```
gemmellr, line 410: This seems odd...is it actually necessary? Is there a default file on the classpath screwing with this test in some way? That would seem like a general issue needing fixed in itself (also against ARTEMIS-3627 perhaps?), it really doesnt feel like unrelated individual tests should need to be randomly setting this null to work.

gtully: It does need a way to turn it off so setting null is good. 
Yes there is a file on the classpath, that changed the jmx name, I will fix that in another commit

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/embedded/EmbeddedActiveMQ.java
gtully: It will get referenced via the commit message

gemmellr: It should be the primary target of the commit, and shouldnt have unrelated test changes for another issue.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/ElasticQueueTest.java
gemmellr: Having a way to do it is fine, but it doesnt seem like this test (or mostly any other) should need to do so.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/embedded/EmbeddedActiveMQ.java
gtully: pushed https://github.com/apache/activemq-artemis/pull/3924 with this logic and sorts the test failure. thanks!

gtully:
seeing intermittent failure of elastic queue test now, need to investigate what is going in. the name property introduced by ARTEMIS-3652  were the cause of failure, so there must be something else in play.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/balancing/ElasticQueueTest.java
gtully: agree. thanks.

gtully:
tracking connection errors via a listener and in onMessage and dealing with a missed send reply for correlation of received messages. cannot reproduce any failure now.

gemmellr:
Seems reasonable.

PR #3923, opened 2022-01-22
ARTEMIS-3653 broker diagram improvement

erwindon:
See https://issues.apache.org/jira/browse/ARTEMIS-3653


jbertram:
IMO all these changes should be squashed into a single commit and the commit message can have a bulleted list of all the individual changes.

clebertsuconic:
+1 on squashing the commits... it would be easier to manage

erwindon:
rebased as needed and squashed as requested

PR #3924, opened 2022-01-24
ARTEMIS-3627 - allow default classpath properties name to be configur…

gtully:
…ed and use non intrusive property for testing, root cause of falure in test identified by ARTEMIS-3652

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/server/config/JMSServerPropertyConfigTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/server/config/JMSServerPropertyConfigTest.java
@@ -45,7 +45,7 @@ public class JMSServerPropertyConfigTest extends ActiveMQTestBase {
 
       try (PrintStream out = new PrintStream(new BufferedOutputStream(new FileOutputStream(propertiesInBindingsDir)))) {
          // use the same name property as from the classpath broker.properties to verify precedence of system prop
-         out.println("name=nameFromCopiedPropertiesRefViaSystemProp");
+         out.println("gracefulShutdownTimeout=-3");
       }
 
       System.setProperty(ActiveMQDefaultConfiguration.BROKER_PROPERTIES_SYSTEM_PROPERTY_NAME, propertiesInBindingsDir.getAbsolutePath());
@@ -54,7 +54,7 @@ public class JMSServerPropertyConfigTest extends ActiveMQTestBase {
          server.setConfiguration(configuration);
          server.start();
 
-         assertEquals("nameFromCopiedPropertiesRefViaSystemProp", server.getActiveMQServer().getConfiguration().getName());
+         assertEquals(-3, server.getActiveMQServer().getConfiguration().getGracefulShutdownTimeout());
 
       } finally {
          System.getProperties().remove(ActiveMQDefaultConfiguration.BROKER_PROPERTIES_SYSTEM_PROPERTY_NAME);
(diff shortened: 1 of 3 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/server/config/JMSServerPropertyConfigTest.java
```diff
@@ -78,7 +78,31 @@ public void testConfigViaBrokerPropertiesFromClasspath() throws Exception {
          server.setConfiguration(configuration);
          server.start();
 
-         assertEquals("ConfiguredViaProperties", server.getActiveMQServer().getConfiguration().getName());
+         assertEquals(-2, server.getActiveMQServer().getConfiguration().getGracefulShutdownTimeout());
```
gemmellr, line 81: Would be good to add a constant with/or comment to explain where the -2 comes from, will be very non-obvious later (isnt even that obvious now, until the diff following later to explain it in a totally different file).

Should probably also assert that ActiveMQDefaultConfiguration.getDefaultGracefulShutdownTimeout() isnt -2, otherwise this test and the next one could be checking the same thing and not really verifying that "server.setPropertiesResourcePath(null);" does anything.

(The tests are so similar they could also share a common impl and pass a boolean to say whether to set the value and govern the expected result..might make them more obvious).

tests/integration-tests/src/test/resources/broker.properties
```diff
@@ -15,4 +15,6 @@
 # limitations under the License.
 #
 
-name=ConfiguredViaProperties
\ No newline at end of file
+# use a very innocuous bit of config b/c it will be picked up by all embedded servers in the integration tests
```
gemmellr, line 18: If unit testing properly verified the file config is being picked up and used as expected I dont see a real need for the integration test using the actual default _broker.properties_ file, and especially not doing so for _every_ integration test even though the broker doesnt provide a default file. That just means the tests are _always_ going on a path that actual use typically wont use. Would seem nicer to use a non-default file or only create the file during the test that needs it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/server/config/JMSServerPropertyConfigTest.java
gtully: the hint is in the name, and most config defaults to -1, -2 is plainly odd just for that reason!

tests/integration-tests/src/test/resources/broker.properties
gtully: it is only the embedded broker server that knows about the default and I think that is the normal path for it, same as for broker.xml.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/server/config/JMSServerPropertyConfigTest.java
gemmellr: Why hint when you can straight up say so. Here where it is being asserted upon, rather than burried away in some other file entirely that you would already need to know about to get there.

tests/integration-tests/src/test/resources/broker.properties
gemmellr: The original change makes it look like the run command knows about it and checksthe default name, passing it on if it exists. The broker having and loading a file for _all_ the tests, that it is _not_ loading by default typically (as it doesnt exist) for normal users, means it isnt the normal path to me. It means the tests dont take the normal path.

gemmellr:
Whilst I'd change this, I recognise it is also a lot better than what is on main and knackering the systests currently (particularly in concert with leading to ARTEMIS 3652 and then taking out much of what follows). So it would probably be worth adding in the meantime if nothign else, along with the remainder from #3922

PR #3925, opened 2022-01-24
ARTEMIS-3654 Fix AllClassesTest leaks

brusdev:


brusdev:
@gemmellr thanks, good catch

gemmellr:
Looks good, but can you put it against https://issues.apache.org/jira/browse/ARTEMIS-3654, I'd like to keep track of these issues.

PR #3927, opened 2022-01-25
ARTEMIS-3647 - OpenWire, remove rolledbackMessageRef on Ack

AntonRoskvist:
I don't really like the test (getting a hold of the current rolledbackMessageRefs) but it does work like it is. If there is some better way of doing that please let me know

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
@@ -113,7 +113,7 @@ public class AMQConsumer {
       return rolledbackMessageRefs;
    }
 
-   private Set<MessageReference> getRolledbackMessageRefs() {
+   protected Set<MessageReference> getRolledbackMessageRefs() {
       return this.rolledbackMessageRefs;
    }
 
@@ -343,6 +343,7 @@ public class AMQConsumer {
             if (ack.isIndividualAck() || ack.isStandardAck()) {
                for (MessageReference ref : ackList) {
                   ref.acknowledge(transaction, serverConsumer);
+                  removeRolledback(ref);
                }
             } else if (ack.isPoisonAck()) {
                for (MessageReference ref : ackList) {
@@ -353,6 +354,7 @@ public class AMQConsumer {
                      ((QueueImpl) ref.getQueue()).incDelivering(ref);
                   }
                   ref.getQueue().sendToDeadLetterAddress(transaction, ref);
+                  removeRolledback(ref);
                }
             }
 
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/amq/RedeliveryPolicyTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/amq/RedeliveryPolicyTest.java
@@ -29,6 +29,10 @@ import java.util.concurrent.TimeUnit;
 
 import org.apache.activemq.ActiveMQConnection;
 import org.apache.activemq.RedeliveryPolicy;
+import org.apache.activemq.artemis.core.protocol.openwire.amq.AMQConsumer;
+import org.apache.activemq.artemis.core.protocol.openwire.amq.AMQConsumerAccessor;
+import org.apache.activemq.artemis.core.server.ServerConsumer;
+import org.apache.activemq.artemis.core.server.ServerSession;
 import org.apache.activemq.artemis.api.core.management.QueueControl;
 import org.apache.activemq.artemis.api.core.management.ResourceNames;
 import org.apache.activemq.artemis.tests.integration.openwire.BasicOpenWireTest;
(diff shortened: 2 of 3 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/amq/RedeliveryPolicyTest.java
```diff
@@ -355,6 +359,53 @@ public void testMaximumRedeliveryDelay() throws Exception {
...
+      }
+      ServerConsumer serverConsumer = null;
+      for (ServerSession session : server.getSessions()) {
+         for (ServerConsumer sessionConsumer : session.getServerConsumers()) {
+            if (sessionConsumer.getQueue().getName().toString() == "TEST") {
+               serverConsumer = sessionConsumer;
+            }
+         }
+      }
+      AMQConsumer amqConsumer = (AMQConsumer) serverConsumer.getProtocolData();
```
franz1981, line 402: At this point it would be better to be able to query `AMQConsumer` re rolledbackMessageRefs directly, exposing an ad-hoc method used just for testing purposes

franz1981 (changes requested):

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/amq/RedeliveryPolicyTest.java
AntonRoskvist: Okay, I'll see what I can do about that. Thanks!

AntonRoskvist: @franz1981 Is this looking better?

franz1981: TBH I'm not a great fan as well to expose public members, but probably there's no easier way to do it

AntonRoskvist: That's actually why I went with the first option to begin with... I've spent 10 times more time on the test than on the actual fix haha.

I'm open to changing it to whatever is most appropriate though so just let me know.

AntonRoskvist:
Squashed the changes made on the test

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
```diff
@@ -306,8 +306,6 @@ public int sendMessage(MessageReference reference,
                           ServerConsumer consumer,
                           int deliveryCount) {
       AMQConsumer theConsumer = (AMQConsumer) consumer.getProtocolData();
-      //clear up possible rolledback ids.
-      theConsumer.removeRolledback(reference);
```
jbertram, line 310: I'm not 100% certain why this was added in the first place, but I think it would be safer to leave it here. As far as I can tell it's not hurting anything.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
```diff
@@ -493,4 +495,10 @@ private boolean isRolledBack(MessageReference messageReference) {
       }
       return rollbackedMessageRefs.contains(messageReference);
    }
+
+   /** For tests only */
+   public Set<MessageReference> listRolledbackMessageRefs() {
```
jbertram, line 500: Why not just make the existing `org.apache.activemq.artemis.core.protocol.openwire.amq.AMQConsumer#getRolledbackMessageRefs` public?

jbertram:
There's a conflict that needs to be resolved before this can be merged.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
gemmellr: Also perhaps rather than make it public at all, make it package-private or protected and add a test-tree class in the same package as it to act as an accessor, e.g Clebert uses this approach, for an example (which just changes such a class) see https://github.com/apache/activemq-artemis/commit/2383aa0125320713b9a753668b203a878c24b2e0#diff-7426cf013ea833e8f1c3cd71cd26e9b584755e13d1fc4ccb4bc95025598795f6

franz1981: Nice method!

AntonRoskvist:
Thanks for your feedback @jbertram  
I made the changes you suggested and also rebased the PR since the conflict came from another recent change.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
jbertram: Robbie's idea is even better.

AntonRoskvist: Okay... I will try to get that working. Do I have to do anything in particular to be able to run the test classes from the integration-tests-suite? Might be an issue with my setup but I can't import: `org.apache.activemq.artemis.core.protocol.openwire.amq.AMQConsumerTest `
straight up?

AntonRoskvist: @jbertram @gemmellr @franz1981 Is this looking better? Have not seen or done that kind of access before but it seems to be working like you described...

gemmellr: Not entirely no. I'll admit I hadnt noticed the test was in a different module when I gave that example where they are in the same one, but I still wouldnt approach it in this way.

Add the new test 'accessor' method in its own new class in the same package, not to the actual test class such as AMQConsumerTest, and place the new class near to the test that needs to use it, i.e in the integration tests module in this case. That will obviously be a split-package, but the integration-tests are not published and arent moduled so that shouldnt cause any issue currently, and theres then no need to add the far uglier test-jar usage.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/amq/RedeliveryPolicyTest.java
```diff
@@ -479,6 +483,54 @@ public void testCountersAreCorrectAfterSendToDLQ() throws Exception {
...
+         for (ServerConsumer sessionConsumer : session.getServerConsumers()) {
+            if (sessionConsumer.getQueue().getName().toString() == "TEST") {
+               serverConsumer = sessionConsumer;
+            }
+         }
+      }
+
+      AMQConsumer amqConsumer = (AMQConsumer) serverConsumer.getProtocolData();
+      assertTrue(AMQConsumerTest.getRolledbackMessageRefs(amqConsumer).isEmpty());
+      //assertTrue(amqConsumer.getRolledbackMessageRefs().isEmpty());
```
gemmellr, line 530: Leftover should be removed.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
AntonRoskvist: @gemmellr Like this?

gemmellr: Yep, thats better

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/amq/RedeliveryPolicyTest.java
clebertsuconic: What's wrong with the accessor? It's easy enough, right?

Or am I missing something?

gemmellr: This is an older comment, discussion moved to another one, and an accessor class was added already.

PR #3928, opened 2022-01-25
ARTEMIS-3627 - fix regression in CriticalCrashTest - statically confi…

gtully:
…gured converters were being lost on new custom bean utils

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/uri/BeanSupport.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/uri/BeanSupport.java
@@ -35,16 +35,25 @@ import org.apache.commons.beanutils.Converter;
 public class BeanSupport {
 
    private static final BeanUtilsBean beanUtils = new BeanUtilsBean();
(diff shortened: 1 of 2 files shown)


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/uri/BeanSupport.java
```diff
@@ -35,18 +35,27 @@
...
       synchronized (beanUtils) {
          beanUtils.getConvertUtils().register(converter, type);
+         customConverters.put(converter, type);
       }
    }
 
+   public static void customise(BeanUtilsBean bean) {
+      for (Map.Entry<Converter, Class> existing : customConverters.entrySet()) {
+         bean.getConvertUtils().register(existing.getKey(), existing.getValue());
+      }
```
gemmellr, line 55: The other customConverters usage is synchronized, so this probably should be also.

Could be nicer with:
```
      customConverters.forEach((key, value) -> {
         bean.getConvertUtils().register(key, value);
      });
```

PR #3929, opened 2022-01-25
NO-JIRA - fix ManifestTest

tlbueno:
This PR contains a fix to the jms jar manifest test:

As the test needs the generated jms jars to be verified I moved it from unit-tests to smoke-tests.
Updated the test to look for the correct jars as the originally specified does not exist.
Update the test to assert against Implementation-Version instead of ActiveMQ-Version in the manifest file as the ActiveMQ-Version property does not exist.

### DIFF
--- a/tests/smoke-tests/pom.xml
+++ b/tests/smoke-tests/pom.xml
@@ -29,6 +29,7 @@
    <properties>
       <activemq.basedir>${project.basedir}/../../</activemq.basedir>
       <sts-surefire-extra-args />
+      <artemis-distribuiton-lib-dir>-Ddistribution.lib="${activemq.basedir}/artemis-distribution/target/apache-artemis-${project.version}-bin/apache-artemis-${project.version}/lib"</artemis-distribuiton-lib-dir>
    </properties>
 
    <dependencies>
@@ -1332,7 +1333,7 @@
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
                <skipTests>${skipSmokeTests}</skipTests>
-               <argLine>${sts-surefire-extra-args} ${activemq-surefire-argline}</argLine>
+               <argLine>${sts-surefire-extra-args} ${activemq-surefire-argline} ${artemis-distribuiton-lib-dir}</argLine>
             </configuration>
          </plugin>
       </plugins>
--- a/tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/jms/ManifestTest.java
+++ b/tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/jms/ManifestTest.java
@@ -14,10 +14,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.activemq.artemis.tests.unit.jms.misc;
+package org.apache.activemq.artemis.tests.smoke.jms;
 
 import javax.jms.ConnectionMetaData;
 import java.io.File;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Properties;
 import java.util.jar.Attributes;
 import java.util.jar.JarFile;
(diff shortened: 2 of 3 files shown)


brusdev:
@tlbueno thanks for this fix, could you add `NO-JIRA` prefix to the commit message and rebase the branch?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/ManifestTest.java
```diff
@@ -31,36 +24,44 @@
...
       ActiveMQServer server = ActiveMQServers.newActiveMQServer(createBasicConfig());
-
       ConnectionMetaData meta = new ActiveMQConnectionMetaData(server.getVersion());
 
-      // Compare the value from ConnectionMetaData and MANIFEST.MF
-      Attributes attrs = manifest.getMainAttributes();
+      for (String jarFile : jarFiles) {
+         // The jar must be there
+         File file = new File(userDir, jarFile + meta.getProviderVersion() + ".jar");
+         Assert.assertTrue(file.exists());
```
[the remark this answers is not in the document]
tlbueno: thanks for the input @gemmellr. That approach make sense and I will work on this and update the PR.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/jms/ManifestTest.java
```diff
@@ -14,53 +14,58 @@
...
-      // Compare the value from ConnectionMetaData and MANIFEST.MF
-      Attributes attrs = manifest.getMainAttributes();
-
-      Assert.assertEquals(meta.getProviderVersion(), attrs.getValue("ActiveMQ-Version"));
+      Version serverVersion = server.getVersion();
+      String serverFullVersion = serverVersion.getFullVersion();
+      ConnectionMetaData meta = new ActiveMQConnectionMetaData(serverVersion);
+
+      String distributionDir = basedir + "/../../artemis-distribution/target/apache-artemis-" +
+         serverFullVersion + "-bin/apache-artemis-" + serverFullVersion;
```
tlbueno, line 53: @gemmellr I made few changes as you suggested. Moved it from integration to smoke. No there is no need for the build.lib paramerter but I didn't find a way to get the artemis-distribution path. If I use the artemis-maven-plugin to generate a server the generated server folder has a profile file that points to the distrubution directory where I can look for the jars inside the lib directory. But I cannot find a way to retrieve this directory in the code. So, for now as you can see it is partially hardcoded. Any suggestions to void this hardcoded part?

gemmellr: Not offhand no, I actually thought the smoke tests bits already had that value, but it seems like its actually only artemis-maven-plugin that does (perhaps from [1]) which in turn puts it in the generated profile file as you noted, with the smoke tests only knowing the path to the configs. So adding a 'distribution lib' style property as you had before may indeed be the simplest for now after all unfortunately.

[1] https://github.com/apache/activemq-artemis/blob/2.20.0/artemis-maven-plugin/src/main/java/org/apache/activemq/artemis/maven/ArtemisCreatePlugin.java#L66

tlbueno: Ok, so I added the property in the smoke test pom. So, @brusdev @gemmellr let me know if any other change will be required.

tests/smoke-tests/pom.xml
```diff
@@ -28,7 +28,7 @@
 
    <properties>
       <activemq.basedir>${project.basedir}/../../</activemq.basedir>
-      <sts-surefire-extra-args />
+      <sts-surefire-extra-args>-Ddistribution.lib="${activemq.basedir}/artemis-distribution/target/apache-artemis-${project.version}-bin/apache-artemis-${project.version}/lib"</sts-surefire-extra-args>
```
gemmellr, line 31: This isnt the right place to declare this, as its used as a profile property to add extra args on JVMs that need them [1], so putting it here actually means those JVMs wont get this config. Add a specific property for this and reference it in the surefire config at: https://github.com/apache/activemq-artemis/blob/main/tests/smoke-tests/pom.xml#L1335

[1] https://github.com/apache/activemq-artemis/blob/main/tests/smoke-tests/pom.xml#L1348

PR #3933, opened 2022-01-31
ARTEMIS-3662 remove deprecated config from default broker.xml

jbertram:
The auto-create-jms-queues, auto-delete-jms-queues,
auto-create-jms-topics, and auto-delete-jms-topics address settings
were deprecated in ARTEMIS-881 way back in 2016. There's no need to keep
them in the default broker.xml at this point.

### DIFF
--- a/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/broker.xml
+++ b/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/broker.xml
@@ -127,8 +127,6 @@ ${cluster-security.settings}${cluster.settings}${replicated.settings}${shared-st
             <address-full-policy>${full-policy}</address-full-policy>
             <auto-create-queues>true</auto-create-queues>
             <auto-create-addresses>true</auto-create-addresses>
-            <auto-create-jms-queues>true</auto-create-jms-queues>
-            <auto-create-jms-topics>true</auto-create-jms-topics>
          </address-setting>
          <!--default for catch all-->
          <address-setting match="#">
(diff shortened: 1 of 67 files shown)


gemmellr:
Seems like a good idea.

There still seems to be _a lot_ of tests remaining with such old 'default config' with these being set though, feels like those also be getting cleaned up so the tests are actually using representative configs, and also just to minimise the presence/use of these deprecated bits to the areas exercising them until eventual removal.

(Perhaps more tangentially, there are also still a lot of references to autoCreateJms* and autoDeleteJms* arguments in non-deprecated api methods that seem to make no mention of the arg [or method] status)

gemmellr:
From thinking on this a bit more though I wonder whether the actual xsd and code default handling should also be getting updated for the 'jms' entries at this point, so that if they werent explicitly configured in the xml config they either default to whatever value the related 'regular' setting is using, or just to false (at least for the auto-delete, since thats the provided default config in xml now...though its still defaults true in the xsd and the code).

Since the -jms settings have been or are here removed from (or never existed in) the default provided xml config, they then arent influenced by the use of these variables set by the cli command flags anymore, but did used to be for the auto-create ones at least. That means that the 'regular' and 'jms' settings effective values will likely differ, e.g the auto-delete bit has xml false config for the regular setting since 2.18.0, yet the -jms ones will still be true from the xsd or code default....then for the auto-create, it has default true set in the xml config but can be configured false with the CLI flags, yet doing so wont affect the -jms one anymore after this change but would before, meaning they would remain true in the xsd and code defaults despite a cli request to disable auto creation.

jbertram:
> There still seems to be a lot of tests remaining with such old 'default config' with these being set though...

I found mostly smoke tests and examples which were still referencing these config elements. I cleaned those up as well.

As for the behavior now that they aren't being set, that shouldn't be a problem. The broker doesn't even use those address setting values anymore. They've simply been ignored since the introduction of their replacements.

The change here is really just about presentation. No semantics should be impacted at all.

gemmellr:
Ok cool. That wasnt really obvious to me from the significant remaining references and mechanics + still-remaining tests checking configuring those values, xsd still defaulting them to true, or the doc not noting that they dont do anything.

PR #3934, opened 2022-01-31
ARTEMIS-2413 upgrade JGroups

jbertram:
JGroups 3.x hasn't been updated in some time now. The last release was
in April 2020 almost 2 years ago. Since then there's been 35 releases
from 4.x and 5.x. Lots of protocols have been updated and added and
users are wanting to use them. Multiple Jiras have been opened to
upgrade:

 - https://issues.apache.org/jira/browse/ARTEMIS-2413
 - https://issues.apache.org/jira/browse/ARTEMIS-2436
 - https://issues.apache.org/jira/browse/ARTEMIS-3066

Notably, there is concern over the security of using older components
triggered mainly by the recent Log4Shell vulnerability discovered in
Log4j 2.x. The upgrade is meant to deal with all these issues.

However, there is a cost associated with upgrading.

The old-style properties configuration is no longer supported. I think
it's unlikely that end-users are leveraging this because it is not
exposed via broker.xml. The JGroups XML configuration has been around
for a long time, is widely adopted, and is still supported. I expect
most (if not all) users are using this. However, a handful of tests
needed to be updated and/or removed to deal with this absence.

Some protocols and/or protocol properties are no longer supported. This
means that users may have to change their JGroups stack configurations
when the upgrade. For example, our own clustered-jgroups example had to
be updated or it wouldn't run properly.


docs/user-manual/en/using-jms.md
```diff
@@ -170,9 +170,8 @@ Lastly, the `jgroups` scheme is supported which provides an alternative to the
 `udp` scheme for server discovery. The URL pattern is either
 `jgroups://channelName?file=jgroups-xml-conf-filename`
 where`jgroups-xml-conf-filename` refers to an XML file on the classpath that
-contains the JGroups configuration or it can be
-`jgroups://channelName?properties=some-jgroups-properties`. In both instance
-the `channelName` is the name given to the jgroups channel created.
+contains the JGroups configuration. The `channelName` is the name given to the
```
gemmellr, line 173: The start of the earlier sentence also needs updated as it is no longer 'either'.

gemmellr:
I dont think the commit message needs as much commentry (thats something JIRAs are good for though). Particularly not about other components or how many jgroups releases there have been. It just makes sense not to be as out of date as this is/was, no real need to explain/justify in the log.

jbertram:
@gemmellr, point taken. I wouldn't normally provide this level of justification in a commit message. However, given that this is a breaking change is likely to be merged and released in a minor version I felt it necessary to provide greater-than-normal detail for the potential users who look back through the log to see why their client broke. In my opinion this break is justified given the nature of the change as noted in the commit message.

gemmellr:
I think its a decent enough JIRA description (though its not actually on the JIRA) but not a great commit log. It would be good to say the version it is upgrading to.

I think the version number jumping from 3 to 5 is clear enough reason in itself. The first half of the log message could go, maybe leaving just the initial 'JGroups 3.x hasn't been updated in some time' + 'and has CVEs', and it would have the same effect. Getting around to the important detail of the change itself having a potential break effect far earlier seems nicer to me. Referencing Log4Shell and Log4J in the commit, in a paragraph ending "The upgrade is meant to deal with all these issues" even seems like an actively bad thing to do when Log4J isnt involved.

I think many users will tend to find the JIRA (e.g via the release notes page), which can and should explain things in detail, before they read the commit log.

jbertram:
@gemmellr, fair enough. I just pushed an updated commit message.

PR #3940, opened 2022-02-04
ARTEMIS-3670 support diverting to multiple addresses

jbertram:
Similar semantics are possible with multicast queues and FQQN, but this feature is different enough and easy enough to implement that it makes sense to offer it.


clebertsuconic:
wouldn't we be able to do the same with multiple diverts instead of one divert with multiple queues?

clebertsuconic:
I guess we had this discussion before, and multiple divert was the reason why we rejected. 

I have nothing against it though.. as long as it's clear the use for users.

jbertram:
Yes, you can do the same thing with 1 divert per address. This is really just a nice-to-have feature for usability. I recently answered a question on Stack Overflow about it so I looked at what it would take to implement it and it was super simple so I did.

clebertsuconic:
@jbertram check style issue on your PR. I would merge it if it wasn't for that.

PR #3943, opened 2022-02-06
ARTEMIS-3670 Fixing Divert Update tests

clebertsuconic:


clebertsuconic:
@jbertram this is a fix for that divert PR, if you could review this one please? (and merge it if you're ok with it).

PR #3948, opened 2022-02-09
ARTEMIS-3677 mitigate NPE when browsing messages

jbertram:


artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/JsonUtil.java
```diff
@@ -338,6 +338,9 @@ public static String truncateString(final String str, final int valueSizeLimit)
    }
 
    public static Object truncate(final Object value, final int valueSizeLimit) {
+      if (value == null) {
+         return "";
```
gemmellr, line 342: Doesnt necessarily follow to me that the truncation of a null is an empty string. Why not return null? Or throw?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -876,7 +875,7 @@ public final void receiveBuffer(ByteBuf buffer) {
                // keep fields like _AMQ_ACTUAL_EXPIRY in their original type
                map.put(extraPropertiesPrefix + s.toString(), o);
             } else {
-               map.put(extraPropertiesPrefix + s.toString(), JsonUtil.truncate(o.toString(), valueSizeLimit));
+               map.put(extraPropertiesPrefix + s.toString(), JsonUtil.truncate(o != null ? o.toString() : o, valueSizeLimit));
```
gemmellr, line 878: With the comment about wanting to keep original types, its not clear to me why it wouldnt just put the null in. Or omit it.

gemmellr (commented):
Missing test(s).

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/JsonUtil.java
jbertram: This method is used to deal with message data which is sent back to the web console for visual inspection. Given that it didn't handle `null` property values properly means that the web console wasn't actually getting any `null` property values either so it wasn't clear to me that it could actually handle them. I didn't want to fix this NPE only to create another one somewhere else.

Also, there was a precedent for using a blank string for a `null` value already in `org.apache.activemq.artemis.core.message.impl.CoreMessage.TextMessageOpenTypeFactory#getFields`. I decided to carry that behavior forward here.

gemmellr:
More comprehensive unit tests of the actual changes would have been nicer than a systest, the related areas already have unit tests that could be easily adapted. The systest also looks to only use Core and so wouldnt seem to cover the original change at all, given it was in AMQPMessage only.

jbertram:
I added a unit test and an AMQP-specific test via 0bc7bf3f7ca2cd1508291f85ba5589430e5bae27.

To be clear, the original problem was observed with core, not AMQP. The original core systest failed without the fix.

PR #3950, opened 2022-02-10
ARTEMIS-3679 Brokers shutdown after daylight saving fall back

franz1981:
https://issues.apache.org/jira/browse/ARTEMIS-3679


artemis-server/src/test/java/org/apache/activemq/artemis/core/server/impl/jdbc/JdbcLeaseLockTest.java
```diff
@@ -116,6 +122,22 @@ public void dropLockTable() throws Exception {
...
    }
 
+   @Test
+   public void shouldStillHeldLockOnTimeZoneChange() {
+      Assume.assumeThat(PropertySQLProvider.Factory.identifyDialect(dbConf.getJdbcDriverClassName()), CoreMatchers.is(PropertySQLProvider.Factory.SQLDialect.DERBY));
+      TimeZone.setDefault(TimeZone.getTimeZone(ZoneId.of("-1")));
+      final long expirationMillis = Duration.ofMinutes(30).toMillis();
+      final LeaseLock lock = lock(expirationMillis);
+      Assert.assertTrue("Must acquire the lock!", lock.tryAcquire());
+      TimeZone.setDefault(TimeZone.getTimeZone(ZoneId.of("+3")));
```
clebertsuconic, line 132: this all look good to me, can't we merge this already?

franz1981: Yep this is not ready to merge yet

franz1981: @fvaleri I still have to decide how to handle testing this :)
But I've decided to use specific UTC get time query per supported DBs

artemis-jdbc-store/src/main/resources/journal-sql.properties
```diff
@@ -75,14 +76,15 @@ create-file-table.mysql=CREATE TABLE %s(ID BIGINT NOT NULL AUTO_INCREMENT, FILEN
...
+current-timestamp.mysql=SELECT UTC_TIMESTAMP(6)
 # 4 GiB
 max-blob-size.mysql=4294967296
 table-names-case.mysql=lower
 
 # Oracle SQL statements
 create-file-table.oracle=CREATE TABLE %s(ID NUMBER(19) GENERATED BY DEFAULT ON NULL AS IDENTITY, FILENAME VARCHAR(255), EXTENSION VARCHAR(10), DATA BLOB, PRIMARY KEY(ID))
 create-journal-table.oracle=CREATE TABLE %s(id NUMBER(19),recordType NUMBER(5),compactCount NUMBER(5),txId NUMBER(19),userRecordType NUMBER(5),variableSize NUMBER(10),record BLOB,txDataSize NUMBER(10),txData BLOB,txCheckNoRecords NUMBER(10),seq NUMBER(19))
-
+current-timestamp.oracle=SELECT SYSTIMESTAMP AT TIME ZONE 'UTC' FROM DUAL
```
fvaleri, line 87: Given the different behaviors and way to express these queries, I would say we should test them on all databases setting a system TZ different from UTC.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/jdbc/JdbcLeaseLock.java
```diff
@@ -114,14 +115,14 @@ private String readableLockStatus() {
          final boolean autoCommit = connection.getAutoCommit();
          connection.setAutoCommit(false);
          try (PreparedStatement preparedStatement = connection.prepareStatement(this.isLocked)) {
+            final long currentTimestamp = dbCurrentTimeMillis(connection);
```
fvaleri, line 118: I guess in dbCurrentTimeMillis() we should select the current UTC timestamp query based on the database type (e.g. `current-timestamp.mssql`) and fall back to `current-timestamp` with an appropriate warning if the database is unknown.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/jdbc/JdbcLeaseLock.java
```diff
@@ -147,14 +148,13 @@ private long dbCurrentTimeMillis(Connection connection) throws SQLException {
...
             resultSet.next();
             final long endTime = stripMilliseconds(System.currentTimeMillis());
-            final Timestamp currentTimestamp = resultSet.getTimestamp(1);
-            final long currentTime = currentTimestamp.getTime();
-            final long currentTimeMillis = stripMilliseconds(currentTime);
-            if (currentTimeMillis < startTime) {
-               LOGGER.warnf("[%s] %s query currentTimestamp = %s on database should happen AFTER %s on broker", lockName, holderId, currentTimestamp, new Timestamp(startTime));
+            final long currentTime = resultSet.getTimestamp(1, Calendar.getInstance(TimeZone.getTimeZone("UTC"))).getTime();
+            final long currentTimeNoMillis = stripMilliseconds(currentTime);
+            if (currentTimeNoMillis < startTime) {
```
fvaleri, line 153: Shouldn't this method be called something like verifyAndGetDatabaseCurrentTimestamp to better reflect the internal logic?

fvaleri (commented):
Overall LGTM, but we need to carefully test the ad-hoc queries.

artemis-jdbc-store/src/main/resources/journal-sql.properties
franz1981: Agree, let's see what tests I can add

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/jdbc/JdbcLeaseLock.java
franz1981: Given that verification here won't imply anything but some output message probably `verify` can be omitted (unless I'll decide to perform a proper validation with thrown exception)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/jdbc/JdbcLeaseLock.java
franz1981: Yep, this is already how internally work the SQLProvider class donated by Jeff Mesnil from WildFly

fvaleri: ok

PR #3951, opened 2022-02-11
NO-JIRA Remove unused tomcat-servlet-api dependency

brusdev:


gemmellr:
Looks good, pushed. I tweaked the log to put it against a recent JIRA cleaning up the assembly since its essentially related.

PR #3952, opened 2022-02-11
Fix for ARTEMIS-3509

BBE78:
Concurrent modifications happened sometimes on non synchronized methods that use "refsToAck" list

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/RefsOperation.java
@@ -83,7 +83,7 @@ public class RefsOperation extends TransactionOperationAbstract {
    }
 
    @Override
-   public void afterRollback(final Transaction tx) {
+   public synchronized void afterRollback(final Transaction tx) {
       Map<QueueImpl, LinkedList<MessageReference>> queueMap = new HashMap<>();
 
       long timeBase = System.currentTimeMillis();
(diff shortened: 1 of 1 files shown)


BBE78:
I don't really understand the build failed relationship with my modifications...

jbertram:
I don't think adding `synchronized` to these methods is a valid fix. This is going to impose a potentially significant performance penalty across the broker for just about any transaction. Also, you have not included a test to validate the fix and mitigate against future regressions.

I recommend tracking down the root cause of the concurrent access in the STOMP implementation and adding protection there.

jbertram:
FWIW, the build failure was spurious so no need to be concerned about that specifically.

clebertsuconic:
a test would be nice

clebertsuconic:
a test would be even nice to validate the synchronize is a the right place..

BBE78:
@jbertram do you prefer my last commit (by the use of a CopyOnWriteArrayList)?
@jbertram and @clebertsuconic I didn't see how I could verify this modification (and also the initial problem) through a simple unit test. Any idea?

jbertram:
Using `CopyOnWriteArrayList` is better than using `synchronized` relatively speaking, but I'm still not in favor of a change that is going to have such a widespread impact for an issue that seems confined to the STOMP implementation. As noted previously, I recommend you track down the root cause of the concurrent access and add protection there.

As far as testing goes. I will typically write a system/integration test that brute forces the issue to occur. For example, if there's a race condition in connection handling then I'll spin up a few hundred threads and create as many connections as possible as quickly as possible. Once I can brute force the issue semi-reliably then I'll gather logging to narrow down the issue until the root cause is clear at which point a proper fix can be applied.

jbertram:
I'm closing this for now. Feel free to resubmit when you deal with the issues listed in the comments. Thanks!

PR #3957, opened 2022-02-15
Bridge work

jbertram:
One feature and one fix related to core bridges.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionBridge.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionBridge.java
@@ -40,6 +40,7 @@ import org.apache.activemq.artemis.api.core.management.ResourceNames;
 import org.apache.activemq.artemis.core.client.impl.ClientSessionFactoryInternal;
 import org.apache.activemq.artemis.core.client.impl.ServerLocatorInternal;
 import org.apache.activemq.artemis.core.client.impl.TopologyMemberImpl;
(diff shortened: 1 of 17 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionBridge.java
```diff
@@ -118,9 +119,20 @@ public ClusterConnectionBridge(final ClusterConnection clusterConnection,
...
-            retryInterval, retryMultiplier, maxRetryInterval, nodeUUID, name, queue, executor, filterString, forwardingAddress, scheduledExecutor, transformer, useDuplicateDetection, user, password, server, ComponentConfigurationRoutingType.valueOf(ActiveMQDefaultConfiguration.getDefaultBridgeRoutingType()));
+                                  final StorageManager storageManager) throws ActiveMQException {
+      super(targetLocator, new BridgeConfiguration()
+         .setInitialConnectAttempts(initialConnectAttempts)
+         .setReconnectAttempts(reconnectAttempts)
+         .setReconnectAttemptsOnSameNode(0) // reconnectAttemptsOnSameNode means nothing on the clustering bridge since we always try the same
+         .setRetryInterval(retryInterval)
+         .setRetryIntervalMultiplier(retryMultiplier)
+         .setMaxRetryInterval(maxRetryInterval)
+         .setFilterString(filterString.toString())
```
gemmellr, line 130: There are numerous test failing now due to NPEs from this area. Presumably the older handling ofthe value in the parent dealt with a null while these new bits dont.

PR #3961, opened 2022-02-18
ARTEMIS-3690 Enhancement of ActiveMQRAManagedConnectionFactory by com…

pkerner:
…pressLargeMessage and initialConnectAttempts

### DIFF
--- a/artemis-commons/src/test/resources/META-INF/MANIFEST.MF
+++ b/artemis-commons/src/test/resources/META-INF/MANIFEST.MF
@@ -1,0 +1,3 @@
+Manifest-Version: 1.0
+Class-Path: 
+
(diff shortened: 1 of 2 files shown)


jbertram:
@pkerner, there are _lots_ of formatting changes that have nothing to do with the new methods you added. Many of these are going to run afoul of the style checker (i.e. `checkstyle`). Please alter your commit to only add the getters and setters for `compressLargeMessage` and `initialConnectAttempts`.

jbertram:
Also, commits like this should go to the `main` branch and then if necessary they can be back-ported to the `2.19.x` branch.

jbertram:
Replaced by #3964.

PR #3975, opened 2022-03-04
ARTEMIS-3701 Do no block libaio on compacting or closing

clebertsuconic:
I am adding a test showing it is safe to not wait pending callbacks before closing a file.
With this I can just close the file and let the kernel to deal with sending the completions.


clebertsuconic:
I am sending more changes.. converting to draft

PR #3977, opened 2022-03-07
Changed artemis group/user id to 1001

pringinacio:
Changed group and user ids to 1001 to allow artemis user to be the owner of /var/lib/artemis-instance and /opt/activemq-artemis. Otherwise the owner will be root (1000). This was already being done in Dockerfile-centos file.


jbertram:
I squashed these commits into one and associated it with a Jira. Thanks for the PR!

PR #3978, opened 2022-03-08
ARTEMIS-3708 Collapse key transformer into policy

brusdev:


gtully (approved):

docs/user-manual/en/connection-routers.md
```diff
@@ -172,15 +167,23 @@ Let's take a look at some connection router examples from broker.xml:
 </connection-routers>
 ```
 
+## Key values
+The key value is retrieved from the incoming client connection.
+If the incoming client connection has no value for the key type used, the key value is set to the special string `NULL`.
+If the incoming client connection has a value for the key type used, the key value retrieved can be sequentially manipulated using a `key-filter` and a `policy`.
+If it is defined a `key-filter` and it matches the key value it is set to the value subsequence matched otherwise it is set to the special string `NULL`.
```
gtully, line 174: If a key-filter is defined....

docs/user-manual/en/connection-routers.md
```diff
@@ -172,15 +167,23 @@ Let's take a look at some connection router examples from broker.xml:
...
+If it is defined a `policy` with a key transformation the key value it is set to the transformed value.
```
gtully, line 175: If  a policy with a key transformation is defined ......

PR #3979, opened 2022-03-08
ARTEMIS-3711 support AMQ_SCHEDULED_DELAY for OpenWire clients

jbertram:


artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -224,6 +225,13 @@ public OpenWireMessageConverter() {
          coreMessage.putStringProperty(AMQ_MSG_ORIG_DESTINATION, origDest.getQualifiedName());
       }
 
+      final Object scheduledDelay = messageSend.getProperties().get(ScheduledMessage.AMQ_SCHEDULED_DELAY);
+      if (scheduledDelay != null && scheduledDelay instanceof Long) {
```
gemmellr, line 229: The null check is redundant

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/OpenWireScheduledDelayTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+   @Override
+   protected void configureAddressSettings(Map<String, AddressSettings> addressSettingsMap) {
+      addressSettingsMap.put("#", new AddressSettings().setDeadLetterAddress(new SimpleString("ActiveMQ.DLQ")));
+   }
+
+   @Test
+   public void testScheduledDelay() throws Exception {
+      final String QUEUE_NAME = RandomUtil.randomString();
+      final long DELAY = 10000;
```
gemmellr, line 46: Burning 10 seconds on this one test seems excessive, especially if only allowing 250ms for the actual delivery to occur before failing the test. Is there another better systest that would take less time? Perhaps send multiple messages, e.g one with a very long delay so it isnt actually delivered, then one with a 'short' delay of e.g <=2sec thats enough to verify delivery is not-immediate for that one and do a quick check the long one doesnt arrive, perhaps check its future delivery time via management? Maybe even throw in a message with no delay?

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -224,6 +225,13 @@ public OpenWireMessageConverter() {
          coreMessage.putStringProperty(AMQ_MSG_ORIG_DESTINATION, origDest.getQualifiedName());
       }
 
+      final Object scheduledDelay = messageSend.getProperties().get(ScheduledMessage.AMQ_SCHEDULED_DELAY);
+      if (scheduledDelay instanceof Long) {
+         coreMessage.putLongProperty(org.apache.activemq.artemis.api.core.Message.HDR_SCHEDULED_DELIVERY_TIME, System.currentTimeMillis() + ((Long) scheduledDelay));
+         // this property may have already been copied, but we don't need it anymore
+         coreMessage.removeProperty(ScheduledMessage.AMQ_SCHEDULED_DELAY);
+      }
+
```
clebertsuconic, line 234: `      final Object scheduledDelay = messageSend.getProperties().get(ScheduledMessage.AMQ_SCHEDULED_DELAY);
       if (scheduledDelay instanceof Number) {
          coreMessage.putLongProperty(org.apache.activemq.artemis.api.core.Message.HDR_SCHEDULED_DELIVERY_TIME, System.currentTimeMillis() + ((Number) scheduledDelay).longValue());
          // this property may have already been copied, but we don't need it anymore
          coreMessage.removeProperty(ScheduledMessage.AMQ_SCHEDULED_DELAY);
       }
`

clebertsuconic: Are you sure about restricting it as long?

What about using Number, and longValue() ?

clebertsuconic: I just checked and the openwire client has a check.. so it's ok

PR #3983, opened 2022-03-11
ARTEMIS-3720 Max Number Of Messages on a deciding policy for paging (global and address-settings)

clebertsuconic:


clebertsuconic:
@gemmellr you mentioned me the other day that Page would run a task twice when releasing max-size on blocking. This would address that I think  (as well)

gemmellr:
Dont think I did? Oh....actually, do you mean https://issues.apache.org/jira/browse/ARTEMIS-3591 from a while back? I had forgotten about that. Skimming the change here, it doesnt seem like it would directly address that.

clebertsuconic:
@gemmellr there is an atomic  flag on a new class called SizeAwareMetric (something like that). I don't have a test for the issue you reported but this could address that.

gemmellr:
> @gemmellr there is an atomic flag on a new class called SizeAwareMetric (something like that). I don't have a test for the issue you reported but this could address that.

I've added a test and the trivial fix in #3985

clebertsuconic:
Oops... I need to add tests for the parsing.. and I forgot to change the XSD for the address-settings

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -213,6 +213,8 @@
 
    private static final String MAX_SIZE_BYTES_NODE_NAME = "max-size-bytes";
 
+   private static final String MAX_MESSAGES_NODE_NAME = "max-messages";
```
jbertram, line 216: I wonder if `max-size-messages` would be better here for consistency with `max-size-bytes`.

clebertsuconic: I thought about that..


I thought that number of messages would not be a size thing... (quantity versus size)..

If you think the term size would be okay here I can definitely include it.

clebertsuconic: @jbertram ok.. I like that actually.. I had thought about max-size-message and I thought it wouldn't be a good term because of English grammar...

although global-max-messages it sounds fine as is I think? 

global-max-size-messages wouldn't sound weird?

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -1479,7 +1500,8 @@ public int getEncodeSize() {
          BufferHelper.sizeOfNullableBoolean(defaultGroupRebalancePauseDispatch) +
          BufferHelper.sizeOfNullableInteger(managementMessageAttributeSizeLimit) +
          BufferHelper.sizeOfNullableInteger(slowConsumerThresholdMeasurementUnit.getValue()) +
-         BufferHelper.sizeOfNullableBoolean(enableIngressTimestamp);
+         BufferHelper.sizeOfNullableBoolean(enableIngressTimestamp) +
```
jbertram, line 1503: Typo here. Should be something like `BufferHelper.sizeOfNullableLong(maxMessages)`.

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
jbertram: Yes, I think `global-max-messages` is the best choice there.

clebertsuconic:
@jbertram I applied the changed

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
```diff
@@ -0,0 +1,304 @@
...
+   public void addOnSizeCallback(IntConsumer onSize, boolean sizeOnly) {
+
+      IntConsumer[] slots = sizeOnly ? onSizeOnlySlots : onSizeSlots;
+
+      for (int i = 0; i < slots.length; i++) {
+         if (slots[i] == null) {
+            slots[i] = onSize;
+            return;
+         }
+      }
```
gemmellr, line 112: Most or all of the non-test uses appears to only have 0 or 1 callback, could this not just be simplified? Anyone needing more can always build upon it later (e.g in their callback).

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
```diff
@@ -189,13 +198,45 @@ public PagingStoreImpl(final SimpleString address,
...
+      if (manager instanceof PagingManagerImpl) {
+         return ((PagingManagerImpl) manager).getSizeAwareMetric();
+      } else {
+         // for unit / mock tests
+         SizeAwareMetric metric = new SizeAwareMetric(-1, -1, -1, -1, -1, 1, 1, 0, 0);
+         metric.addOnSizeCallback(manager::addSize, false);
+         metric.addOnSizeCallback((l) -> manager.addSize(l, true), true);
+         return metric;
+      }
+   }
```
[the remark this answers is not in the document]
clebertsuconic: I did not want to add SizeAwareMetric to the interface... I wanted it to be an internal implementation detail.

if the implementation (PagingManagerImpl) is using it, it could use it on a parent call for the metric. Otherwise the proxy Metric would use the already existing API.

I'm adding it to the interface now with a default implementation returning the proxy. I didn't want to do it.. but I will do it just to settle this.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
```diff
@@ -0,0 +1,304 @@
...
+         if (underSlots[i] == null) {
+            underSlots[i] = under;
+            return;
+         }
+      }
+
+      throw new IllegalStateException("no open slot available");
+   }
+
+   public SizeAwareMetric(long maxSize, long lowerMark, long maxElements, long lowerMarkElements, SizeAwareMetric[] parentSlots, IntConsumer[] onSizeSlots, IntConsumer[] onSizeOnlySlots, Runnable[] overSlots, Runnable[] underSlots) {
```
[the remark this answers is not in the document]
clebertsuconic: I have future uses for this class.. I will keep this constructor private for now.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
clebertsuconic: planned for future usage (e.g. Flow control on Actor).

this is going to always be null and every usage will happen through Parent. (which will be a single instance now after my update.. no more array).

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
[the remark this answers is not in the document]
clebertsuconic: actually, it will look better to just use the SizeAwareMetric.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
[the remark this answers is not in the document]
clebertsuconic: I will just remove it...

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
clebertsuconic: ok.. fair enough, I will simplify it.

gemmellr: Part of the reason I suggested it is, its all inherently non-thread-safe without outside coordination, and then once I realised it wasnt really being used, it just further seemed like it would be better to only have whats required.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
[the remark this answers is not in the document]
gemmellr: Having internal detail is fine. The main point was really the other bit though, its generally a bad idea for non-test code to be just suddenly returning objects only intended for [certain] tests. At some point something will unexpectedly get one of those instead of the real thing, when theres no reason they really should know the behaviour is there. Better for tests that need to to know and apply the test logic they need.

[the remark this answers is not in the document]
clebertsuconic: the other bits is just returning a proxy to the pagingManager API.

anyway I'm adding SizeAwareMetric to the public api with a default implementation to the proxy version now.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
```diff
@@ -0,0 +1,304 @@
...
+   }
+
+   public long addSize(int delta) {
+      return addSize(delta, false);
+   }
+
+   public long addSize(int delta, boolean sizeOnly) {
+
+      if (delta == 0) {
+         logger.debug("SizeAwareMetric ignored with size 0", new Exception("trace"));
```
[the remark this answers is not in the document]
clebertsuconic: impossible to happen. I will add a test to make sure it's validated and add the logger.isDebugEnabled();

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -3934,6 +3943,14 @@
                </xsd:annotation>
             </xsd:element>
 
+            <xsd:element name="max-size-messages" type="xsd:long" default="-1" maxOccurs="1" minOccurs="0">
```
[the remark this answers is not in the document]
clebertsuconic: from the point of view of the config I prefer max-size-messages TBH

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
```diff
@@ -0,0 +1,240 @@
...
+            under();
+         }
+      }
+
+      if (elementsEnabled) {
+         if (currentElements <= lowerMarkElements && changeFlag(OVER_ELEMENTS, FREE)) {
+            under();
+         }
+      }
+   }
```
gemmellr, line 224: TLDR: Seems like it needs to account for the other value as well before simply firing 'under' whenever one of them goes under, or it might fire it incorrectly.

What if both go over but a later change only brings one of them under? Seems like it cant currently track that properly and so could actually fire the callback early and leave itself in the wrong state.

E.g whilst both can be over, the flag state only allows tracking that the size or the elements are over, so will only remember one of those (whichever flag is set first). Just because that remembered one goes under later, doesnt necessarily mean the other one is also under at that point, but currently the flag would just get reset straight to 'free' and the under callback would be fired, both possibly incorrectly.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
```diff
@@ -0,0 +1,240 @@
...
+
+      flagUpdater.compareAndSet(this, NOT_USED, FREE);
+
+      if (parentMetric != null) {
+         if (parentMetric != null) {
+            parentMetric.addSize(delta, sizeOnly);
+         }
+      }
+
+      IntConsumer theFunction = sizeOnly ? onSizeOnlyCallback : onSizeCallback;
```
gemmellr, line 171: Given these only looks to be used to call an equivalent method on anther object, could they be combined into one callback and simply pass the boolean as well? Would need to use a custom callback interface though (or a BiConsumer with boxing). Not really a hardship given it would probably just be supplied with a lambda these days and never actually implemented.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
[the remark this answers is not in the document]
gemmellr: It would be a fine name if it didnt as easily suggest totally different things hehe.

If not trying to contort to 'max-size-<foo>' for alignment to the other config, the most obvious thing for me would be just 'max-message-count'.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
clebertsuconic: once it goes back from size, it can only go back from size... once it goes back from elements it can only go back from elements...  whatever hits it first.

once it goes back from SIZE, the flag will be OVER_SIZE and the other check will be moot. I don't see any need for any extra checks in there.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
clebertsuconic: yes.. it can be done.. 

I wanted to avoid the Boxing on a BiConsumer, and I forgot for an instant I didn't think about the own interface.. thanks. nice idea.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
```diff
@@ -0,0 +1,240 @@
...
+
+   public long addSize(int delta) {
+      return addSize(delta, false);
+   }
+
+   public long addSizeOnly(int delta) {
+      return addSize(delta, true);
+   }
+
+   public long addSize(int delta, boolean sizeOnly) {
```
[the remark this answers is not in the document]
clebertsuconic: I will only have addSize(int), and addSize(int, boolean)

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
```diff
@@ -0,0 +1,240 @@
...
+         if (logger.isDebugEnabled()) {
+            logger.debug("SizeAwareMetric ignored with size 0", new Exception("trace"));
+         }
+         return sizeUpdater.get(this);
+      }
+
+      flagUpdater.compareAndSet(this, NOT_USED, FREE);
+
+      if (parentMetric != null) {
+         if (parentMetric != null) {
```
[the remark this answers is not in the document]
clebertsuconic: thanks.. kind of a typo during the renaming...

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
```diff
@@ -0,0 +1,240 @@
...
+         } else { // (delta < 0)
+            checkUnder(currentElements, currentSize);
+         }
+      } else {
+         if (delta > 0) {
+            checkOver(elementsUpdater.incrementAndGet(this), currentSize);
+         } else { // (delta < 0)
+            checkUnder(elementsUpdater.decrementAndGet(this), currentSize);
+         }
+      }
```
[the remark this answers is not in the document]
clebertsuconic: it is not really saving much, as the delta>0 will need to increment, and delta<0 decrement.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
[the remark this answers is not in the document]
clebertsuconic: I'm also removing setParent and just using the addSize callback as "the parent"

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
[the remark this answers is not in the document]
gemmellr: yeah, I had completely forgot that one was inc and the other was dec so it would obviously need 3 rather than 2 legs

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
clebertsuconic: @gemmellr fixed.. I amended the commit..

This is actually nice as it ensures the encapsulation here was a nice idea. fixing the check on an encapsulated class was better than doing the check on PagingStore and PagingManager.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SizeAwareMetric.java
```diff
@@ -0,0 +1,236 @@
...
+            // checking if we need to switch from OVER_ELEMENTS to OVER_SIZE, to avoid calling under needless
+            if (!(sizeEnabled && currentSize >= maxSize && changeFlag(FREE, OVER_SIZE))) {
+               // this is checking the other side from size (elements).
+               // on this case we are just switching sides and we should not fire under();
+               under();
+            }
+            return; // this return here is moot I know. I am keeping it here for consistence with the size portion
+                    //  and in case eventually further checks are added on this method, this needs to be reviewed.
+         }
+      }
```
gemmellr, line 215: Is it ok for it to fire _over_ repeatedly without ever firing _under_  inbetween, as this would then seem possible now?

clebertsuconic: it's the opposite... it will only fire under if this wasn't achieved...

think of this as 

```java
            if (sizeEnabled && currentSize >= maxSize && changeFlag(FREE, OVER_SIZE)) {
               // nothing to be done as it switch over.
             } else {
               // this is checking the other side from size (elements).
               // on this case we are just switching sides and we should not fire under();
               under();
            }
```


I actually had it like this first and I then added the ! to the check

clebertsuconic: it would still be fine .. although I don't think that's the case.

gemmellr: This stops it firing 'under' here if the _other_ threshold is still over when the first goes down again, yes. However, my point was around that it still sets the state to FREE first, so could something else happen that makes it realise the other threshhold is breached, which will set that as the state, and possibly fire _over_ again having never fired _under_. Though in retrospect I guess it will fire _under_ here though if it fails to _changeFlag(FREE, <other>)_. So it will fire, just maybe not when expected (could be after it fired _over_ in the other check).

PR #3985, opened 2022-03-14
ARTEMIS-3591: stop paging checkMemory task executing twice

gemmellr:
Stop paging checkMemory task executing twice in some cases.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
@@ -717,7 +717,8 @@ public class PagingStoreImpl implements PagingStore {
                runWhenBlocking.run();
             }
 
-            onMemoryFreedRunnables.add(AtomicRunnable.checkAtomic(runWhenAvailable));
+            AtomicRunnable atomicRunWhenAvailable = AtomicRunnable.checkAtomic(runWhenAvailable);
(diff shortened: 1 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
```diff
@@ -717,15 +717,16 @@ public boolean checkMemory(boolean runOnFailure, final Runnable runWhenAvailable
...
+            onMemoryFreedRunnables.add(atomicRunWhenAvailable);
 
             // We check again to avoid a race condition where the size can come down just after the element
             // has been added, but the check to execute was done before the element was added
             // NOTE! We do not fix this race by locking the whole thing, doing this check provides
             // MUCH better performance in a highly concurrent environment
             if (!pagingManager.isGlobalFull() && (sizeInBytes.get() < maxSize || maxSize < 0)) {
                // run it now
-               runWhenAvailable.run();
+               atomicRunWhenAvailable.run();
```
clebertsuconic, line 729: Ohhh... I thought you had a more complex case involved on this.


Thanks for this.

gemmellr: Nope just the obvious cases mentioned on the JIRA, spotted when reading it for some older PR (dont recall which). Theres probably a few different paths it can happen, but the most basic solution used here hits the general issue. Took a lot longer to come up with a reliable test to exercise the problem than to fix.

PR #4003, opened 2022-03-29
ARTEMIS-3749 e2e-tests improvements

tlbueno:
Scripts:
- Fix the preapre-docker.sh to exit with 0 instead of 1 on success

On pom files:
- Change e2e-tests variable names to e2e-tests.xxxxxx for clarity on
  e2e-tests variables
- Add e2e-tests.skipImageBuild variable to control if the docker image
  will be build (defaults to not build)
- Add e2e-tests.dockerfile variable to specify the dockerfile to be
  used (defaults to Dockerfile-centos)
- Bump testcontainers version to 1.16.3
- Add artemis distribution dependency since the docker image build
  depends on it

On ContainerService class:
- Fix exposePorts and exporseFolder to use SELinux shared mode
  otherwise the mount fails on machines with SELinux enabled
- Move the logic to use specific user on container from generic start
  method to broker specific method to avoid affect other images
- Update the broker image name to a more generic name (activemq-artemis
  instead of artemis-centos)
- Update the broker image tag to match with the project version in pom
  file


tests/e2e-tests/pom.xml
```diff
@@ -75,6 +101,50 @@
 
    <build>
       <plugins>
+         <plugin>
```
clebertsuconic, line 104: Can you add a TODO here to replace codehaus for something better.. I think you opened a JIRA on this.

clebertsuconic: I'm confused though.. I thought we needed to replace codehaus.. but that seems someone still maintaining it on GitHub? 


If you could add just a TODO now so we can figure out this later?

tests/e2e-tests/pom.xml
```diff
@@ -75,6 +101,51 @@
...
+            <!-- Review Codehaus usage...ARTEMIS-3750 -->
```
clebertsuconic, line 105: ;)

tests/e2e-tests/pom.xml
gemmellr: Mojo bits from Codehaus were then established as/at Mojohaus, and yet kept the groupId.

tests/e2e-tests/pom.xml
```diff
@@ -28,15 +28,41 @@
...
+               <groupId>org.apache.activemq</groupId>
+               <artifactId>artemis-jakarta-client-all</artifactId>
+            </exclusion>
+         </exclusions>
+      </dependency>
       <dependency>
          <groupId>org.testcontainers</groupId>
          <artifactId>testcontainers</artifactId>
-         <version>1.16.2</version>
+         <version>1.16.3</version>
```
gemmellr, line 65: Should be in a property, and the version controlled in dependencyManagement at the root so it is consistent across the build and can be updated easily.

tests/e2e-tests/pom.xml
```diff
@@ -75,6 +101,51 @@
 
    <build>
       <plugins>
+         <plugin>
+            <!-- TODO: Review Codehaus usage...ARTEMIS-3750 -->
+            <groupId>org.codehaus.mojo</groupId>
+            <artifactId>exec-maven-plugin</artifactId>
+            <version>3.0.0</version>
```
gemmellr, line 108: Should be in a property, and the version controlled in pluginManagement at the root so it is consistent across the build and can be updated easily.

tests/e2e-tests/pom.xml
tlbueno: Ok @gemmellr going to change it. Also, I will do the same for smoke-test project which has a reference for testcontainers selenium module.

tlbueno: PR updated.

tests/e2e-tests/pom.xml
tlbueno: PR updated

PR #4009, opened 2022-04-04
[ARTEMIS-3756]: Enabling extra parameters to be passed by the URI.

ehsavoie:
* passing the full list of parameters when creating the URI.
 * moving the WARN message to DEBUG when unknown parameters are
   received.

Jira: https://issues.apache.org/jira/browse/ARTEMIS-3756

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/ActiveMQClientLogger.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/ActiveMQClientLogger.java
@@ -426,11 +426,6 @@ public interface ActiveMQClientLogger extends BasicLogger {
    void broadcastTimeout(int retry, int maxretry);
 
    @LogMessage(level = Logger.Level.WARN)
-   @Message(id = 212078, value = "Connection factory parameter ignored {0}",
-      format = Message.Format.MESSAGE_FORMAT)
-   void connectionFactoryParameterIgnored(String parameterName);
-
-   @LogMessage(level = Logger.Level.WARN)
    @Message(id = 212079, value = "The upstream connector from the downstream federation will ignore url parameter {0}", format = Message.Format.MESSAGE_FORMAT)
    void ignoredParameterForDownstreamFederation(String name);
 
--- a/artemis-jms-client/src/main/java/org/apache/activemq/artemis/uri/TCPSchema.java
+++ b/artemis-jms-client/src/main/java/org/apache/activemq/artemis/uri/TCPSchema.java
@@ -83,7 +83,7 @@ public class TCPSchema extends AbstractCFSchema {
          if (!key.equals("ha") && !key.equals("type") &&
             !TransportConstants.ALLOWABLE_CONNECTOR_KEYS.contains(key) &&
             !factoryProperties.containsKey(key)) {
-            ActiveMQClientLogger.LOGGER.connectionFactoryParameterIgnored(key);
+            ActiveMQClientLogger.LOGGER.debug(String.format("Connection factory parameter %s is not standard", key));
          }
       }
    }
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/SimpleJNDIClientTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/SimpleJNDIClientTest.java
@@ -36,6 +36,7 @@ import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.logging.Level;
 
 import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
 import org.apache.activemq.artemis.api.core.BroadcastEndpoint;
@@ -59,6 +60,7 @@ import org.apache.activemq.artemis.logs.AssertionLoggerHandler;
 import org.apache.activemq.artemis.spi.core.security.ActiveMQJAASSecurityManager;
 import org.apache.activemq.artemis.tests.util.ActiveMQTestBase;
 import org.apache.activemq.artemis.utils.Wait;
+import org.jboss.logmanager.LogManager;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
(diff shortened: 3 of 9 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/ActiveMQClientLogger.java
```diff
@@ -425,10 +425,10 @@
            format = Message.Format.MESSAGE_FORMAT)
    void broadcastTimeout(int retry, int maxretry);
 
-   @LogMessage(level = Logger.Level.WARN)
-   @Message(id = 212078, value = "Connection factory parameter ignored {0}",
-      format = Message.Format.MESSAGE_FORMAT)
-   void connectionFactoryParameterIgnored(String parameterName);
```
clebertsuconic, line 431: Can you just remove this, and call logger.debug directly where this is being called.

We always use logger.debug at the logger on the class.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/uri/TCPSchema.java
```diff
@@ -83,7 +83,7 @@ private void checkIgnoredQueryFields(ActiveMQConnectionFactory factory, Map<Stri
          if (!key.equals("ha") && !key.equals("type") &&
             !TransportConstants.ALLOWABLE_CONNECTOR_KEYS.contains(key) &&
             !factoryProperties.containsKey(key)) {
-            ActiveMQClientLogger.LOGGER.connectionFactoryParameterIgnored(key);
+            ActiveMQClientLogger.LOGGER.connectionFactoryExtraParameter(key);
```
clebertsuconic, line 86: just define a logger on TCPSchema, can call logger.debug here directly...

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/SimpleJNDIClientTest.java
```diff
@@ -111,22 +115,26 @@ public void testEmptyConnectionFactoryString() throws NamingException, JMSExcept
...
       props.put("java.naming.provider.url", "");
-      new InitialContext(props);//Must not throw an exception
-
+      Context ctx = new InitialContext(props);//Must not throw an exception
+      ctx.close();
    }
 
    @Test
    public void testConnectionFactoryStringWithInvalidParameter() throws Exception {
+      Level initialLevel = LogManager.getLogManager().getLogger("org.apache.activemq.artemis.core.client").getLevel();
```
clebertsuconic, line 124: lets just remove this test?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/ActiveMQClientLogger.java
ehsavoie: Is it ok to remove the id 212078 or should I keep it in the debug trace ? There is a test looking for it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/SimpleJNDIClientTest.java
ehsavoie: ok

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/ActiveMQClientLogger.java
ehsavoie: ok just saw your last comment.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/SimpleJNDIClientTest.java
```diff
@@ -111,22 +115,26 @@ public void testEmptyConnectionFactoryString() throws NamingException, JMSExcept
...
       props.put(Context.INITIAL_CONTEXT_FACTORY, "org.apache.activemq.artemis.jndi.ActiveMQInitialContextFactory");
       props.put("connectionFactory.ConnectionFactory", "tcp://localhost:61616?foo=too");
 
       AssertionLoggerHandler.startCapture();
       try {
-         new InitialContext(props);
-         assertTrue("Expected to find AMQ212078", AssertionLoggerHandler.findText("AMQ212078"));
+         Context ctx = new InitialContext(props);
+         ctx.close();
+         assertTrue("Connection factory parameter foo is not standard", AssertionLoggerHandler.findText("Connection factory parameter foo is not standard"));
```
gemmellr, line 134: This assertion is failing when the test is run (and then afterwards the test sits around waiting for threads to go away).

EDIT: I now see a prior comment about actually removing the test. Not sure if there was other context around that but clearly its still here.

ehsavoie: The trace was moved to debug, could it be some logging misconfiguration ?

ehsavoie: https://github.com/apache/activemq-artemis/pull/4009/files#diff-c86b5f721ed580116525565937bfbc27f4dfcbc64c21e43a55a94780b4d90e73R20

ehsavoie: The code has been changed from  ActiveMQClientLogger.LOGGER.debug(String.format("Connection factory parameter %s is not standard", key)); to logger.debugf("Connection factory parameter %s is not standard", key); private static final Logger logger = Logger.getLogger(TCPSchema.class); which is a different logger category !!! this explains that

gemmellr:
Ah, missed that the code was changed, yep seems likely it could be related hehe.

@clebertsuconic is the fix to change the logging again, or alter the test?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/SimpleJNDIClientTest.java
gemmellr: Ah, missed that the code was changed, yep seems likely it could be related hehe.

@clebertsuconic is the fix to change the logging again, or alter the test?

PR #4011, opened 2022-04-04
ARTEMIS-3757 - allow system and env var substution of properties conf…

gtully:
…ig, respect order of file loaded properties and add generic enum converter

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XMLUtil.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XMLUtil.java
@@ -52,6 +52,12 @@ public final class XMLUtil {
       // Utility class
    }
 
+   public static String CONSIDER_OS_ENV_PROP = "org.apache.activemq.artemis.utils.considerOsEnv";
+   private static final boolean considerOsEnv;
+   static {
+      considerOsEnv = Boolean.parseBoolean(System.getProperty(CONSIDER_OS_ENV_PROP, "true"));
+   }
+
    public static Element streamToElement(InputStream inputStream) throws Exception {
       try (Reader reader = new InputStreamReader(inputStream)) {
          return XMLUtil.readerToElement(reader);
@@ -271,7 +277,14 @@ public final class XMLUtil {
             val = parts[1].trim();
          }
 
-         String sysProp = System.getProperty(prop, val);
+         String sysProp = System.getProperty(prop);
+         if (sysProp == null && considerOsEnv) {
+            sysProp = System.getenv(prop);
+         }
+         if (sysProp == null) {
+            sysProp = val;
+         }
+         // interesting choice to replace with val == "" with no match!
          logger.debug("replacing " + subString + " with " + sysProp);
          xml = xml.replace(subString, sysProp);
       }
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
@@ -17,9 +17,15 @@
 package org.apache.activemq.artemis.core.config.impl;
 
 import java.io.File;
+import java.io.FileOutputStream;
+import java.io.PrintWriter;
+import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.Properties;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 
 import org.apache.activemq.artemis.ArtemisConstants;
 import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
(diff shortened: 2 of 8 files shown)


artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
```diff
@@ -708,6 +720,149 @@ public void testValuePostFixModifier() throws Throwable {
...
+
+      Assert.assertEquals("cc", configuration.getClusterConfigurations().get(0).getName());
+      Assert.assertEquals(MessageLoadBalancingType.OFF_WITH_REDISTRIBUTION, configuration.getClusterConfigurations().get(0).getMessageLoadBalancingType());
+      Assert.assertEquals(CriticalAnalyzerPolicy.SHUTDOWN, configuration.getCriticalAnalyzerPolicy());
+   }
+
+   @Test
+   public void testPropertiesReaderRespectsOrderFromFile() throws Exception {
+
+      File tmpFile = File.createTempFile("ordered-props-test", "");
```
brusdev, line 795: Is it possible remove the tmpFile after the test execution?

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
```diff
@@ -708,6 +720,149 @@ public void testValuePostFixModifier() throws Throwable {
...
+   @Test
+   public void testPropertiesFiles() throws Exception {
+
+      LinkedList<String> files = new LinkedList<>();
+      LinkedList<String> names = new LinkedList<>();
+      names.addLast("one");
+      names.addLast("two");
+
+      for (String suffix : names) {
+         File tmpFile = File.createTempFile("props-test", suffix);
```
brusdev, line 839: Is it possible remove the tmpFile after the test execution?

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XMLUtil.java
```diff
@@ -52,6 +52,12 @@ private XMLUtil() {
       // Utility class
    }
 
+   public static String CONSIDER_OS_ENV_PROP = "org.apache.activemq.artemis.utils.considerOsEnv";
```
brusdev, line 55: An alternative name could be `REPLACE_OS_ENV_PROP`

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
```diff
@@ -708,6 +720,149 @@ public void testValuePostFixModifier() throws Throwable {
...
+   public void testSystemPropValueReplaced() throws Exception {
+      ConfigurationImpl configuration = new ConfigurationImpl();
+      Properties properties = new Properties();
+      final String homeFromDefault = "default-home";
+      final String homeFromEnv = System.getenv("HOME");
+      properties.put("name", "${HOME:" + homeFromDefault + "}");
+      configuration.parsePrefixedProperties(properties, null);
+      if (homeFromEnv != null) {
+         Assert.assertEquals(homeFromEnv, configuration.getName());
+      } else {
```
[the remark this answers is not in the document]
gtully: true, but it also exercise the default value in that case.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
```diff
@@ -708,6 +720,149 @@ public void testValuePostFixModifier() throws Throwable {
...
+      Assert.assertEquals("vv-y", configuration.getName());
+   }
+
+
+   @Test
+   public void testSystemPropKeyReplacement() throws Exception {
+      ConfigurationImpl configuration = new ConfigurationImpl();
+      Properties properties = new Properties();
+
+      final String newKeyName = RandomUtil.randomString();
```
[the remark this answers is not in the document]
gtully: just b/c I don't have full control of the system properties in play

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
gtully: I think there are rules that manage that but I did not check since there were already temp files in play in that test. peeking more, I see there is some base class cleanup, but I need to make use of that. 
Good catch!

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/XMLUtil.java
gtully: it does not replace tho, it only considers if the system prop is absent

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
gtully: pushed a fix for the temp file usage in that test class

brusdev (approved):

PR #4014, opened 2022-04-05
ARTEMIS-3760 Upgrade jackson version to 2.13.2

brusdev:


gemmellr:
Note for clarity: jackson-databind was actually changed to 2.13.2.2 specifically.

PR #4019, opened 2022-04-06
ARTEMIS-3765 Display validated user in Hawtio

ahoffer:
This PR adds a "Validated User" column to the Hawtio plugin. The column is
not visible by default. The "Validated User" column is available in the
Session, Consumer, and Producer tabs of the artemis-navigation notebook.
"Validated User" is also available for sort and filtering in those view.
The "Validated Consumer" in the Producer view is always blank. I debugged
as far as I could, but did not find the cause of the issue.

Screen shots attached.
column_header
column_selector
filter_field
sort_field
producer_view


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/predicate/ConnectionFilterPredicate.java
```diff
@@ -71,6 +66,17 @@ public boolean test(RemotingConnection connection) {
...
+      List<ServerSession> sessions = server.getSessions(connectionId);
+      Set<String> sessionAttributes = new HashSet<>();
+      for (ServerSession session : sessions) {
+         String value = getter.apply(session);
+         String string = value == null ? "" : value;
+         sessionAttributes.add(string);
+      }
+      return sessionAttributes;
+   }
+
```
ahoffer, line 79: i thought I reverted this change. At one point, I was pulling validated users from the sessions associated with the connections and displaying them. I created this method so that I didn't have to copy and paste code for getUsername() and getValidatedUser However, I backed those changes out. I thought I backed this one out too. If you want, I can roll the code back to how it was.

artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/browse.js
```diff
@@ -326,7 +326,7 @@ var Artemis;
                 header: 'User ID'
             },
             {
-                itemField: 'StringProperties',
+                itemField: 'validatedUser',
                 header: 'Validated User',
                 templateFn: function(value) {
                     return value._AMQ_VALIDATED_USER;
```
ahoffer, line 332: The line `header: 'Validated User'` was already in main. I don't know what it does -- I don't see it anywhere on the UI. I change the `itemField` name for consistency.  I'm happy to try and work on this. Let me know how you want it to work.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -2651,7 +2651,7 @@ public void buildSessionJSON(JsonArrayBuilder array, ServerSession sess) {
       JsonObjectBuilder obj = JsonLoader.createObjectBuilder().add("sessionID", sess.getName()).add("creationTime", sess.getCreationTime()).add("consumerCount", sess.getServerConsumers().size());
 
       if (sess.getValidatedUser() != null) {
-         obj.add("principal", sess.getValidatedUser());
+         obj.add("validatedUser", sess.getValidatedUser());
```
ahoffer, line 2654: I'm fine with either validated user or principal. I think it was Justin (?) who said to keep "validated user" to keep things consistent.

jbertram:
@ahoffer, nice work. Thanks for the PR!

PR #4021, opened 2022-04-08
ARTEMIS-3770 refactor MQTT handling of client ID

jbertram:
It would be useful for security manager implementations to be able to
alter the client ID of MQTT connections.

This commit supports this functionality by moving the code which handles
the client ID *ahead* of the authentication code. There it sets the
client ID on the connection and thereafter any component (e.g. security
managers) which needs to inspect or modify it can do so on the
connection.

This commit also refactors the MQTT connection class to extend the
abstract connection class. This greatly simplifies the MQTT connection
class and will make it easier to maintain in the future.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/protocol/AbstractRemotingConnection.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/protocol/AbstractRemotingConnection.java
@@ -161,20 +161,16 @@ public abstract class AbstractRemotingConnection implements RemotingConnection {
 
    @Override
    public List<CloseListener> removeCloseListeners() {
-      List<CloseListener> ret = new ArrayList<>(closeListeners);
-
+      List<CloseListener> deletedCloseListeners = new ArrayList<>(closeListeners);
       closeListeners.clear();
-
-      return ret;
+      return deletedCloseListeners;
    }
 
    @Override
    public List<FailureListener> removeFailureListeners() {
-      List<FailureListener> ret = getFailureListeners();
-
+      List<FailureListener> deletedFailureListeners = getFailureListeners();
       failureListeners.clear();
-
-      return ret;
+      return deletedFailureListeners;
    }
 
    @Override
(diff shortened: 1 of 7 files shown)


jbertram:
@riccardomodanese, does this change work for you?

riccardomodanese:
it works for me.
I tried again with the full Kapua ServerPlugin and SecurityPlugin using the Artemis 2.23.0-SNAPSHOT(updated in the morning) + your commit.
Both the "normal stealing link" and the "multi tenant" are handled correctly (in the second case both the clients are able to stay connected).
With your fix I don't see exceptions thrown by Paho client while, for example, 10 clients are connecting at the same time with the same clientId. The client connects and soon after its status changes to disconnect "softly"
I'll do few more tests in the next days.

riccardomodanese:
About the exception raised by Paho I think it's a good news (but I'm not so deep into MQTT specifications to be sure).
I mean, in one my tests there are n clients connecting more or less at the same time with the same clientId.
The test scope is to check that only one client is still connected after a while and all the clients were able to connect.
Before your fix I noticed a lot of exceptions (n-1) during the connect calls.
Now no client is throwing exception during its connect call.
I think it's right since the MQTT protocol says that the previous client with the same clientId should be disconnected. Not the current one.

Unfortunately I cannot run the Kapua ITs on our ci against Artemis snapshot but the manual test I did today confirmed me the fix is fine.
I added a test to connect n clients (with delay between one connection and the others) with the same clientId but on different accounts (so the Kapua security plugin will change the clientId with accountId:clientId)
All the n clients are able to connect and stay connected together.
From the plugin log I can see
```
### authorizing address: $EDC/stealing-link-7/client-stealing-multi-account/#::e2w9YPXivHY|client-stealing-multi-account.$EDC/stealing-link-7/client-stealing-multi-account/# - check type: CONSUME
### authorizing address: $EDC/stealing-link-7/client-stealing-multi-account/#::e2w9YPXivHY|client-stealing-multi-account.$EDC/stealing-link-7/client-stealing-multi-account/# - check type: CREATE_DURABLE_QUEUE
```
and that's right since the clientId in the durable queue address is (accountId:clientId):
`e2w9YPXivHY|client-stealing-multi-account`

Just one last question about the ETA.
Are you expecting to release this fix with the 2.23?

Thanks a lot!

jbertram:
> Are you expecting to release this fix with the 2.23?

Yes.

riccardomodanese:
Hello, this pr looks like closed without merging.
Did you fix the related issue with another pr?

Thanks!

jbertram:
@riccardomodanese, it was merged. See 446ff61542f47f50c2299d8ef1cae8fe2b98a5ad.

PR #4023, opened 2022-04-08
ARTEMIS-3771 - Rework destination handling for the OpenWire-protocol

AntonRoskvist:
This gives a significant performance increase for my use case but I could not find a good way to demonstrate it through testing, so the added tests only focus on functionality

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
@@ -41,12 +41,12 @@ import org.apache.activemq.artemis.core.protocol.openwire.OpenWireProtocolManage
 import org.apache.activemq.artemis.core.protocol.openwire.util.OpenWireUtil;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
 import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
-import org.apache.activemq.artemis.core.server.BindingQueryResult;
 import org.apache.activemq.artemis.core.server.MessageReference;
 import org.apache.activemq.artemis.core.server.QueueQueryResult;
 import org.apache.activemq.artemis.core.server.ServerConsumer;
 import org.apache.activemq.artemis.core.server.ServerSession;
 import org.apache.activemq.artemis.core.server.SlowConsumerDetectionListener;
+import org.apache.activemq.artemis.core.server.impl.AddressInfo;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
 import org.apache.activemq.artemis.reader.MessageUtil;
 import org.apache.activemq.artemis.spi.core.protocol.SessionCallback;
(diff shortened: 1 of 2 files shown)


artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
```diff
@@ -91,7 +91,7 @@
 
    private final CoreMessageObjectPools coreMessageObjectPools;
 
-   private String[] existingQueuesCache;
+   static String[] existingQueuesCache;
```
franz1981, line 94: This is quite dangerous, making the cache to leak and outlive a broker restart (thinking about activation/deactivation on HA)

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
```diff
@@ -196,26 +195,22 @@ public boolean updateDeliveryCountAfterCancel(ServerConsumer consumer, MessageRe
...
          existingQueuesCache = new String[protocolManager.getOpenWireDestinationCacheSize()];
-         assert (Integer.bitCount(existingQueuesCache.length) == 1) : "openWireDestinationCacheSize must be a power of 2";
-         this.existingQueuesCache = existingQueuesCache;
       }
-      final int hashCode = physicalName.hashCode();
-      //this.existingQueuesCache.length must be power of 2
-      final int mask = existingQueuesCache.length - 1;
-      final int index = hashCode & mask;
+
+      final int index = Math.floorMod(physicalName.hashCode(), existingQueuesCache.length);
```
franz1981, line 203: Modulus is a quite heavyweight operation, that's why the original code was using the mask trick using a power of two size: this would save impacting for the "common" use cases ie not many queues

franz1981 (changes requested):

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
AntonRoskvist: I did consider placing this in the OpenWireProtocolManager instead but opted for this... Would you prefer I place it there and use setter/getter instead?

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
AntonRoskvist: Yes, but since now there will only be one such calculation per new queue, (plus new ones if the cache is full) I figure that might be cheaper overall anyway? I changed it because this method has a lower risk of creating identical hashes for different values (from my testing at least)...

AntonRoskvist:
@franz1981 
Hi,
Do you like these changes better? I opted to add the connection-level cache again since after some more testing I found that I was wrong. Both the bitwise operation and the modulus one produces the same level of cache distribution (and clashing). The distribution "issue" increase with array size... this effect is probably only news to me though.

I should mention that the biggest performance gain here is from removing the "BindingQueryResult", because that becomes a very heavy operation with a large amount of destinations.

clebertsuconic:
Let me try to understand...

you've created an Array on OpenWireprotocolManager that must be sized at the power of 2, you hash the queue and look for it on the ProtcolManager?


it would be better If you're implementing such cache it would be better to create some class... say.. name it as HashStringCache and have the logic encapsulated in there?

AntonRoskvist:
@clebertsuconic thanks for the feedback

Okay... at this point I'm basically just moving the current "session-level" destination cache and making it shared, but the way this cache is implemented now makes collisions happen so often that I'm not really sure it's worth it to be honest. I'm mostly doing this for a use case I have with a very large number of queues and chances of a cache-clash is really high already at "just" some few hundreds of destinations.

If you prefer I change this (and I can do that for sure), do you think it makes sense to just store all destinations in something like a hashmap instead that just holds all destinations, or should I revert this particular part of the change to it's old behavior, i.e Session-level? Either way is fine for me.

clebertsuconic:
this sort off cache should be at session Level I think

AntonRoskvist:
@clebertsuconic @franz1981 
Restored the cache to previous behavior, this PR now primarily just gets rid of the binding-query which was the real performance sink anyway. 

Don't really know how to write a meaningful test to demonstrate the benefit though, but the easiest way to "real-life" demonstrate it is by loading the broker with thousands of queues and have a single producer send messages in individual sessions (such as some frameworks unfortunately tend to do it).

clebertsuconic:
It seems you care about performance.  Why OpenWire though?

AntonRoskvist:
@clebertsuconic I'm running this on a pretty large scale (around 600 client applications or so in some environments, multiple environments interconnected globally) and all of which used to run against ActiveMQ 5. Most are supposed to get migrated over to Core but this is a pretty big undertaking so for some foreseeable future I have to support "backward" compatibility. I figure this is something I can do to make things run as smoothly as possible until such time that they can start to get migrated... and hopefully it might be helpful to someone else too.

clebertsuconic:
@AntonRoskvist I was just curious.. thanks a lot...

probably I should have asked you in private though.. thanks

clebertsuconic:
so, it seems we were doing two checks for the address/. queue, one was useless and not needed,, correct?


on that case can you rename the commit to something more accurate to the current state?


say:

```
ARTEMIS-3771 Avoid not needed lookup for address on OpenWire connections
```


I can of course use some Git-fu and do it myself... but if you do it yourself I guess you wouldn't have my trace as a contributor in your commit.


Also, if you could rebase (git pull --rebase upstream main) and git push -f?


I can then just run my private CI on top of your PR. (I have a parameterized build where I pass in GitHub address and branch name).

AntonRoskvist:
Of course, hope this does it. Otherwise let me know. Thanks @clebertsuconic

clebertsuconic:
full pass on my CI.. merging it.

PR #4029, opened 2022-04-13
ARTEMIS-3778 Streamline Expiration Reaping

clebertsuconic:
Instead of holding a thread and an iterator, we should instead keep moving to next references
without holding any threads. Just with callbacks.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -30,11 +30,9 @@ import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 import java.util.stream.Stream;
 
 import org.apache.activemq.artemis.api.core.ActiveMQAddressDoesNotExistException;
@@ -1846,43 +1844,37 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
          super(scheduledExecutorService, executor, checkPeriod, timeUnit, onDemand);
       }
 
-      volatile CountDownLatch inUseLatch;
+      private Iterator<Queue> iterator;
 
-
-      @Override
-      public void stop() {
-         super.stop();
-         // this will do a best effort to stop the current latch.
-         // no big deal if it failed. this is just to optimize this component stop.
-         CountDownLatch latch = inUseLatch;
-         if (latch != null) {
-            latch.countDown();
-         }
-      }
-
+      private Queue currentQueue;
 
       @Override
       public void run() {
-         // The reaper thread should be finished case the PostOffice is gone
-         // This is to avoid leaks on PostOffice between stops and starts
-         for (Queue queue : iterableOf(getLocalQueues())) {
-            if (!isStarted()) {
-               break;
-            }
-            try {
-               CountDownLatch latch = new CountDownLatch(1);
-               this.inUseLatch = latch;
-               queue.expireReferences(latch::countDown);
-               // the idea is in fact to block the Reaper while the Queue is executing reaping.
-               // This would avoid another eventual expiry to be called if the period for reaping is too small
-               // This should also avoid bursts in CPU consumption because of the expiry reaping
-               if (!latch.await(10, TimeUnit.SECONDS)) {
-                  ActiveMQServerLogger.LOGGER.errorExpiringMessages(new TimeoutException(queue.getName().toString()));
-               }
-            } catch (Exception e) {
-               ActiveMQServerLogger.LOGGER.errorExpiringMessages(e);
-            }
+         if (iterator != null) {
+            logger.debugf("A previous reaping call has not finished yet, and it is currently working on %s", currentQueue);
+            return;
          }
+
+         iterator = iterableOf(getLocalQueues()).iterator();
+
+         moveNext();
+      }
+
+      private void done() {
+         executor.execute(this::moveNext);
+      }
+
+      private void moveNext() {
+         if (!iterator.hasNext() || !this.isStarted()) {
+            iterator = null;
+            currentQueue = null;
+            return;
+         }
+
+         currentQueue = iterator.next();
+
+         // we will expire messages on this queue, once done we move to the next queue
+         currentQueue.expireReferences(this::done);
       }
    }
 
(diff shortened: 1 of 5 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1846,43 +1844,47 @@ private final class ExpiryReaper extends ActiveMQScheduledComponent {
...
-               // This would avoid another eventual expiry to be called if the period for reaping is too small
-               // This should also avoid bursts in CPU consumption because of the expiry reaping
-               if (!latch.await(10, TimeUnit.SECONDS)) {
-                  ActiveMQServerLogger.LOGGER.errorExpiringMessages(new TimeoutException(queue.getName().toString()));
-               }
-            } catch (Exception e) {
-               ActiveMQServerLogger.LOGGER.errorExpiringMessages(e);
-            }
+         if (iterator != null) {
+            logger.debug("iterator is not finished yet");
```
gemmellr, line 1860: The message could make it more obvious what isnt finished to avoid needing to check the source, e.g "Existing expiry reaper iterator is not finished yet, not beginning new sweep."

It would also be good to then add related messages, such as add below that it _is_ starting a new expiry sweep. Perhaps even add another one later that it has finished a sweep, in moveNext(). That way you can easily see when it starts/finishes/skips.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1846,43 +1844,47 @@ private final class ExpiryReaper extends ActiveMQScheduledComponent {
...
+         if (!iterator.hasNext() || !this.isStarted()) {
+            queue = null;
+         } else {
+            queue = iterator.next();
+         }
+
+         if (queue == null) {
+            iterator = null;
+            return;
+         }
```
gemmellr, line 1884: Related to earlier comments...if _iterator_ needs to be volatile due to multiple threads setting it, it would probably be needed to assign it to a new variable and use that (after a null check) to ensure it is consistently used for e.g hasNext() and later next(), and not being be re-read on each use and potentially changing or being nulled between calls (or even before getting into this method).

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2454,10 +2450,10 @@ public void run() {
 
                   if (doneCallback != null) {
                      doneCallback.run();
-                     doneCallback = null;
                   }
                }
 
+
```
gemmellr, line 2456: Empty line addition seems unecessary...but helpful to refer to the existing log message below it.
The _"Scanning for expires on " + QueueImpl.this.getName() + " done"_ message doesnt line up with when it calls the _doneCallback_, as it would be logged every time through, even after the rescheduling. It might be nice if the 'done' messaging did align as it would then work in concert with the other [suggested] logging to be much clearer on what is happening when.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2434,13 +2430,13 @@ public void run() {
                   }
                   if (++elementsIterated >= MAX_DELIVERIES_IN_LOOP) {
                      logger.debug("Breaking loop of expiring");
-                     scannerRunning.incrementAndGet();
+                     rescheduled = true;
```
gemmellr, line 2433: Relating to the line above this one, rather than this line...

Rather than just logging _"Breaking loop of expiring"_ it might be nice to note the queue name as the completion message does, and mention it is rescheduling.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1846,43 +1844,47 @@ private final class ExpiryReaper extends ActiveMQScheduledComponent {
          super(scheduledExecutorService, executor, checkPeriod, timeUnit, onDemand);
       }
 
-      volatile CountDownLatch inUseLatch;
+      volatile Iterator<Queue> iterator;
```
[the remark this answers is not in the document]
clebertsuconic: the very first version of this was using the executor of the queue.... which I later switched to use the same executor on the done method, so it does not need to be volatile..


it's left over.. I forgot to remove it... thanks

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
clebertsuconic: the iterator is always used on the same thread.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1846,43 +1844,47 @@ private final class ExpiryReaper extends ActiveMQScheduledComponent {
...
 
       @Override
       public void stop() {
          super.stop();
          // this will do a best effort to stop the current latch.
          // no big deal if it failed. this is just to optimize this component stop.
-         CountDownLatch latch = inUseLatch;
-         if (latch != null) {
-            latch.countDown();
-         }
```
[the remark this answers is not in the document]
clebertsuconic: removed the left over method

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gemmellr: I wouldnt say this comment still applies. It seems strange to log that it is "done" in the cases it has just been logged that it isnt finished.

gemmellr:
LGTM

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1846,43 +1844,45 @@ private final class ExpiryReaper extends ActiveMQScheduledComponent {
...
-            }
+      private void moveNext() {
+         Queue queue;
+         if (!iterator.hasNext() || !this.isStarted()) {
+            queue = null;
+         } else {
+            queue = iterator.next();
          }
+
+         if (queue == null) {
```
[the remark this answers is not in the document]
clebertsuconic: I added the currentQueue just for debugging. but I guess it could have a double usage now...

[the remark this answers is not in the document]
clebertsuconic: @jbertram good idea.. it does look better.. thanks

PR #4030, opened 2022-04-17
ARTEMIS-3751 Do not override large message size

iliya-gr:
It is possible to receive a compressed message from the client as regular message. Such a message will already contain correct body size, that takes compression into account.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSLargeMessageTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSLargeMessageTest.java
@@ -1,0 +1,192 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.tests.integration.amqp;
+
+import javax.jms.BytesMessage;
+import javax.jms.Connection;
+import javax.jms.Destination;
+import javax.jms.Message;
+import javax.jms.MessageConsumer;
+import javax.jms.MessageProducer;
+import javax.jms.Session;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.Random;
+import java.util.zip.Deflater;
+
+import org.junit.Test;
+
+public class JMSLargeMessageTest extends JMSClientTestSupport {
+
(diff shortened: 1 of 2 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/JMSLargeMessageTest.java
```diff
@@ -0,0 +1,192 @@
...
+      MessageConsumer consumer = session.createConsumer(queue);
+
+      connection.start();
+
+      // What we want is to get a message witch size is larger than client's minLargeMessageSize (200 kibs here) while compressed the size must be lower than that
+      // and at same time it must be greater than server minLargeMessageSize (which is 100 kibs by default)
+      byte[] data = new byte[1024 * 300];
+
+      // We don't want the data to be too random, 42 seems to be a good random seed
+      new DeflateGenerator(new Random(42), 2.0, 2.0).generate(data, 2);
```
clebertsuconic, line 53: I'm trying to understand why you're using Deflate directly? why not use the compressed from the ConnectionFActory directly and why you're compressing it yourself?

clebertsuconic: is this just to generate Random Data?

iliya-gr: It's only for generating random data. Assert is to ensure that it matches criteria: `server large message size < compressed size < client large message size`

iliya-gr: Assert doesn't change data. Deflate in assert is just for checking parameters, if someone decides to change them.

clebertsuconic: >> Deflate in assert is just for checking parameters, if someone decides to change them. <<

I'm not sure what you meant by this statement. Can you speak in java? :)

iliya-gr: assertCompressionSize - doesn't modify input data. It's for checking that DeflateGenerator has generated correct data:

```java
deflate(data).length > storageManager.getMaxRecordSize() && deflate(data).length < serverLocator.minLargeMessageSize
```

iliya-gr: DeflateGenerator - is just compressible data generator 
generate(data, 2) - generate compressible data with compression ratio 2 and write it in data

clebertsuconic: I am merging this with a small adjustment...

the package where you added the test was meant for AMQP tests only... I have moved  your test within LargeMessageCompressTest...


I am keeping you as the author on the commit...

clebertsuconic: as I said on the JIRA the issue is that the message was sent as a regular message, and before writing we check the journal-buffer-size on the server's.. if the buffer-size < message body size, we convert it to large message on the server's side.. this is kind of a protection.

it is a bit of an edge case as you were really supposed to send with minLargeMessageSize = journal-buffer-size... 


I even had users requesting this to be enforced but I could never find a way to do that without breaking the protocol.


I will merge this though, but you should really use minLargeMessageSize == journal-buffer-size.

PR #4031, opened 2022-04-19
ARTEMIS-3780 - OpenWire, UTF8Buffer to String before setting properti…

AntonRoskvist:
…es to avoid exception

Converting to String beforehand to avoid exception.
Shows some performance increase when running a producer/consumer pair sending and receiving 100k messages 5 times, alternating between setting 8 custom headers or not.

Before:
8 user properties: 
Average: 64.236
Each run: 68.30, 60.74, 62.86, 65.39, 63.89

0 user properties: 
Average: 54.862
Each run: 53.55, 52.55, 56.24, 61.14, 50.83

After:
8 user properties: 
Average: 52.944
Each run: 49.48, 57.12, 53.84, 51.66, 52.62

0 user properties: 
Average: 52.1
Each run: 50.45, 52.28, 50.23, 51.69, 55.85

*Values in seconds

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
@@ -447,7 +447,11 @@ public final class OpenWireMessageConverter {
       if (!props.isEmpty()) {
          props.forEach((key, value) -> {
             try {
-               coreMessage.putObjectProperty(key, value);
+               if (value instanceof UTF8Buffer) {
+                  coreMessage.putObjectProperty(key, value.toString());
+               } else {
+                  coreMessage.putObjectProperty(key, value);
+               }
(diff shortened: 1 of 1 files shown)


clebertsuconic:
it seems a no brainer to be merged.. but do you have a test where an exception would happen on such case?

do you know any openwire message that would incur such case? 


if you don't have a test, at least give me some information on how you were reproducing this please?

AntonRoskvist:
Well... it happens on every OpenWire message where there are properties beyond what's defined in the OpenWireMessageConverter... at least that's what I concluded after some basic testing. Just adding any custom header to the message will trigger it.

I noticed this by chance while looking through a CPU profile for something else and the way I confirmed and tested was by adding logging on the thrown exception. (and testing against the unchanged code vs the changed).

But for instance this test will trigger the exception:
org.apache.activemq.artemis.tests.integration.openwire.amq.JmsTransactionTestSupport #testChangeMutableObjectInObjectMessageThenRollback

clebertsuconic:
I did not know an existing test would trigger this already...


I just wanted to have a test.. let me see what happens on the test.. and I will get back here... thanks

PR #4038, opened 2022-04-22
ARTEMIS-3753

clebertsuconic:


clebertsuconic:
@iliya-gr basically instead of looking at the routing context, I just keep the boolean for every queue we add on the context.

PR #4041, opened 2022-04-24
ARTEMIS-3797 Update maven wrapper version

GChuf:
Updates maven wrapper version to 3.8.5 to solve errors when building the project.

### DIFF
--- a/.mvn/wrapper/maven-wrapper.properties
+++ b/.mvn/wrapper/maven-wrapper.properties
@@ -15,4 +15,4 @@
 ## limitations under the License.
 ## ---------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


clebertsuconic:
I will merge this, but although I think this is moot, since mvn wrapper was removed on the latest version.

gemmellr:
I had commented on the JIRA that it would probably make more sense to backport the other change instead. Though it is all rather academic unless another 2.19.x is actually done.

clebertsuconic:
@gemmellr I don't have plans for 2.19.x... not ATM at least.

I was allowing it as I did not really mind about it, and I thought it could be helping him.

PR #4042, opened 2022-04-24
How about add more Forked VM when run all tests in build

CAIWan1998:
Maven will run all tests in a single forked VM by default. This can be problematic if there are a lot of tests or some very memory-hungry ones. We can fork more test VM by setting `<fork>1.5C</fork>`.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/tests/extra-tests/pom.xml
+++ b/tests/extra-tests/pom.xml
@@ -283,7 +283,7 @@
                <!-- make sure maven puts the byteman jar in the classpath rather than in a manifest jar -->
                <useManifestOnlyJar>false</useManifestOnlyJar>
                <!-- when upgrading this plugin from 2.4 to 2.18.1 <forkMode>once</forkMode> was replaced with these: -->
-               <forkCount>1</forkCount>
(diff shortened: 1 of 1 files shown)


gemmellr:
This wont work for many of the Artemis tests without significant reworking of the overall test suite, as they often do things like reuse specific ports and directories for output and other things like that which simply cant work for parallel runs.

The PR run on Travis did pass, though probably only as it doesnt actually run most of the tests in the build, only a small subset. The run was also a few minutes slower than earlier runs without the change, though its admittedly hard to say how much of that was just run to run variability vs e.g. the change itself adding additional overhead in the low resource env (e.g as discussed on your qpid-jms PR).

clebertsuconic:
many tests will use a specific port while accepting. We would need some significant work as Robbie said.

jbertram:
From what I can tell this is some kind of automated PR from @CAIWan1998. Looking at their account details you can see that almost 70 similar PRs were sent yesterday to various projects.

PR #4043, opened 2022-04-24
ARTEMIS-3707 remove use of TransactionManager

waldi5001:
To use the ResourceAdapter in other ApplicationServers, use of TransactionManager is removed. 
From the javadoc:
The TransactionManager is intended for use by the application server, the TransactionSynchronizationRegistry for use by the application server components such as persistence managers, resource adapters...

PR #4044, opened 2022-04-24
ARTEMIS-3707 ResourceAdapter Documentation

waldi5001:
Documentation of the Artemis ResourceAdapter.

### DIFF
--- a/docs/user-manual/en/SUMMARY.md
+++ b/docs/user-manual/en/SUMMARY.md
@@ -88,3 +88,4 @@
 * [Performance Tools](perf-tools.md)
 * [Configuration Reference](configuration-index.md)
 * [Restart Sequence](restart-sequence.md)
+* [Resource Adapter](resource-adapter.md)
--- a/docs/user-manual/en/resource-adapter.md
+++ b/docs/user-manual/en/resource-adapter.md
@@ -1,0 +1,94 @@
+# Resource Adapter
+
+For using Apache ActiveMQ Artemis in a JavaEE/JakartaEE environment, you can use the ResourceAdapter. Before you 
+start, read carefully the [JMS chapter](using-jms.md) 
+
+##In which case do you have to use a RA and in which not?
+The main reason is the requirement of 
+[XA](https://jakarta.ee/specifications/transactions/2.0/jakarta-transactions-spec-2.0.html), distributed 
+transactions over multiple resources. If it is important for you no message get lost and multiple resources (e.g. 
+database and messaging) must be in sync, there is no way around it.
+
+##Versions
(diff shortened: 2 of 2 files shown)


clebertsuconic:
thanks for your contribution

gemmellr:
I dont think this should have been merged with all the links to main in it (see outstanding feedback on the original PR).

I also dont think this one should have been raised at all when considering it is documenting the opposite approach of what the original PR was eventually described as being aimed at achieving (not having to build your own / use the described examples). 

Oh well, I give up.

waldi5001:
@gemmellr It is no problem for me to fix the documentation afterwards and open a PR for that. At the moment no WLP sample is mentioned in the documentation, only available material is used...

Have you found a solution to refer the version in the link?

clebertsuconic:
@gemmellr , @waldi5001 I only made a connection with the previous PR after I merged it. I can revert it and a new PR sent by @waldi5001 ?

waldi5001:
No problem.

clebertsuconic:
@waldi5001 , @gemmellr I reverted this commit as discussed, please send a new PR as we discussed earlier.


thank you

gemmellr:
> @gemmellr It is no problem for me to fix the documentation afterwards and open a PR for that. At the moment no WLP sample is mentioned in the documentation, only available material is used...

Right, and I was saying that seems bizarre, if the original goal was actually to add new rars and get rid of the examples just documented.

> 
> Have you found a solution to refer the version in the link?
> 

I suggested on the original PR you could probably look to include it using some filtering.

I have just raised a PR making it available, #4050

waldi5001:
I opened the new PR #4060 with the version in the links. On my machine with `mvn clean install -Prelease -DskipTests` it works well.

> Right, and I was saying that seems bizarre, if the original goal was actually to add new rars and get rid of the examples just documented.

as I understood you, shipping a artemis.rar is not what you (or the team) want. So, in the documentation is explained what the readers have to do, if they want to use a resource adapter. No example get rid, no artemis.rar is added.

If I misunderstood you, no problem, I will open another PR with artemis.rar files, rid of the examples (or not) and a adjusted documentation. 

But first of all #4043 have to be merged, without that it will become difficult to use the exampleArtemis.rar in the application server of your choice.

gemmellr:
You have definitely misunderstood me. Though I did query at one point why the new rar modules were being added, as they seemed largely duplication of the example and their presence wasnt clear based on the documentation also being added that said you _must_ build your own like the example, we then discussed why on the PR. I reviewed the PR multiple times after without saying 'remove these', and even added stuff to the build myself so you could use it to more easily action some prior feedback I had given, plus when you asked if the PR was dead I said no. You then closed it and raised the other of somewhat different PRs.

PR #4050, opened 2022-04-26
ARTEMIS-3804: make version number available to docs build

gemmellr:
Make the version number available to reference in the user manual, add it to the intro page


clebertsuconic:
isn't that changing how we do release? we need now to build the release out of the scratch doc...

I usually just go to the user/en and type gitbook pdf, gitbook epub, gitbook build


and copy the files over.


now we will need to change the procedure a little bit, right?



perhaps this scratch build should be removed out of the release profile and always be ran?

gemmellr:
I'm fine with it being profiled, it seems unecessary to build it every time...in fact I would like to move the javadoc bit into the profile since its a massive waste of time on each build.

clebertsuconic:
I'm file with using a profile also.. Im just checking what we need to change on the Release... we will need to update the RELEASE.md as well. can you do that?

gemmellr:
@clebertsuconic pushed an update for RELEASE.md ...like that?

clebertsuconic:
thanks

PR #4054, opened 2022-04-26
ARTEMIS-3759 Add mirror controller address filter support

iliya-gr:
Allow replication only certain addresses with mirror controller.
The configuration is similar to cluster address configuration.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
@@ -82,6 +82,7 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
    final boolean acks;
    final boolean addQueues;
    final boolean deleteQueues;
+   final MirrorAddressFilter addressFilter;
    private final AMQPBrokerConnection brokerConnection;
 
    final AMQPMirrorBrokerConnectionElement replicaConfig;
@@ -110,6 +111,7 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
       this.idSupplier = protonProtocolManager.getReferenceIDSupplier();
       this.addQueues = replicaConfig.isQueueCreation();
       this.deleteQueues = replicaConfig.isQueueRemoval();
+      this.addressFilter = new MirrorAddressFilter(replicaConfig.getAddressFilter());
       this.acks = replicaConfig.isMessageAcknowledgements();
       this.brokerConnection = brokerConnection;
    }
@@ -131,6 +133,11 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
       if (getControllerInUse() != null && !addressInfo.isInternal()) {
          return;
       }
+
+      if (ignoreAddress(addressInfo.getName())) {
+         return;
+      }
+
       if (addQueues) {
          Message message = createMessage(addressInfo.getName(), null, ADD_ADDRESS, null, addressInfo.toJSON());
          route(server, message);
@@ -143,6 +150,9 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
          logger.trace(server + " deleteAddress " + addressInfo);
       }
       if (invalidTarget(getControllerInUse()) || addressInfo.isInternal()) {
+         return;
+      }
+      if (ignoreAddress(addressInfo.getName())) {
          return;
       }
       if (deleteQueues) {
@@ -162,6 +172,12 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
          }
          return;
       }
+      if (ignoreAddress(queueConfiguration.getAddress())) {
+         if (logger.isTraceEnabled()) {
+            logger.trace("Skipping create " + queueConfiguration + ", queue address " + queueConfiguration.getAddress() + " doesn't match filter");
+         }
+         return;
+      }
       if (addQueues) {
          Message message = createMessage(queueConfiguration.getAddress(), queueConfiguration.getName(), CREATE_QUEUE, null, queueConfiguration.toJSON());
          route(server, message);
@@ -178,6 +194,10 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
          return;
       }
 
+      if (ignoreAddress(address)) {
+         return;
+      }
+
       if (deleteQueues) {
          Message message = createMessage(address, queue, DELETE_QUEUE, null, queue.toString());
          route(server, message);
@@ -188,12 +208,18 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
       return controller != null && sameNode(getRemoteMirrorId(), controller.getRemoteMirrorId());
    }
 
+   private boolean ignoreAddress(SimpleString address) {
+      return !addressFilter.match(address);
+   }
+
    private boolean sameNode(String remoteID, String sourceID) {
       return (remoteID != null && sourceID != null && remoteID.equals(sourceID));
    }
 
    @Override
    public void sendMessage(Message message, RoutingContext context, List<MessageReference> refs) {
+      SimpleString address = context.getAddress(message);
+
       if (invalidTarget(context.getMirrorSource())) {
          if (logger.isTraceEnabled()) {
             logger.trace("server " + server + " is discarding send to avoid infinite loop (reflection with the mirror)");
@@ -204,6 +230,13 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
       if (context.isInternal()) {
          if (logger.isTraceEnabled()) {
             logger.trace("server " + server + " is discarding send to avoid sending to internal queue");
+         }
+         return;
+      }
+
+      if (ignoreAddress(address)) {
+         if (logger.isTraceEnabled()) {
+            logger.trace("server " + server + " is discarding send to address " + address + ", address doesn't match filter");
          }
          return;
       }
@@ -301,6 +334,13 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
          return;
       }
 
+      if (ignoreAddress(ref.getQueue().getAddress())) {
+         if (logger.isTraceEnabled()) {
+            logger.trace(server + " rejecting postAcknowledge queue=" + ref.getQueue().getName() + ", ref=" + ref + ", queue address is excluded");
+         }
+         return;
+      }
+
       if (logger.isTraceEnabled()) {
          logger.trace(server + " postAcknowledge " + ref);
       }
@@ -337,4 +377,5 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
          return true;
       }
    }
+
 }
--- a/artemis-server/src/main/resources/schema/artemis-configuration.xsd
+++ b/artemis-server/src/main/resources/schema/artemis-configuration.xsd
@@ -2442,6 +2442,14 @@
             </xsd:documentation>
          </xsd:annotation>
       </xsd:attribute>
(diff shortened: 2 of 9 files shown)


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -337,4 +368,52 @@ public boolean isMirrorController() {
          return true;
       }
    }
+
+   public static class AddressFilter {
+
+      private final SimpleString[] whiteList;
```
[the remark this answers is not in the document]
iliya-gr: Thanks for pointing this. We definitely need a maven plugin for this :)

clebertsuconic:
it looks good to me...


I will not merge it now as the release I'm doing today is named 2.21.1... 


I will defer it to 2.22.0 in a couple weeks...


can you squash these commits and push -f?



(git rebase -I, using a fix on the second commit to the first)

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -2442,6 +2442,13 @@
             </xsd:documentation>
          </xsd:annotation>
       </xsd:attribute>
+      <xsd:attribute name="address" type="xsd:string" use="optional">
+         <xsd:annotation>
+            <xsd:documentation>
+               name of the address this mirror connection applies to
```
gemmellr, line 2448: Its not just a simple address name it is taking, so this feels like the wrong attribute name and description. Something conveying the 'filter' nature would seem appropriate. The format of what it does take isnt mentioned , so besides the test usage its unclear how anyone would know. There should be doc updates to e.g https://github.com/apache/activemq-artemis/blob/2.21.0/docs/user-manual/en/amqp-broker-connections.md#mirror-configuration to cover whatever the changes ends up being.

Other 'broker connections' stuff such as the sender+receiver already take "address-match" or "queue-match", as shown in the context of the config change in ConfigurationTest-full-config.xml. It would also seem reasonable to align the terminology between them if possible (and perhaps even the syntax and/or implementation; they are implemented a _completely_ different way currently, which doesnt seem that obvious).

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -337,4 +368,52 @@ public boolean isMirrorController() {
          return true;
       }
    }
+
+   public static class AddressFilter {
```
gemmellr, line 372: This seems like it could be private, or just be its own class, rather than being public so that AMQPMirrorControllerSourceTest can use it for testing, without ever testing anything general of AMQPMirrorControllerSource itself.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPReplicaTest.java
```diff
@@ -473,6 +473,52 @@ public void testNoAddressWithAnnotations() throws Exception {
       }
    }
 
+   @Test
+   public void testAddressFilterMet() throws Exception {
+      addressTest("test", true, 0);
```
gemmellr, line 478: Using a 0 wait seems potentially susceptible to failure, as the mirroring isnt synchronous. I know its only checking for > 0 messages being mirrored, but I'd still make it more robust personally. Perhaps just changing it to wait for them all to be mirrored as we do expect them to quickly be. Actually, no need for sending 200 messages either given it doesnt actually look at any of them, sending a couple is enough, and it should look.

I would also remove the 1 second wait in the 'no match' case and replace it with better verification by other means, e.g monitor the mirroring SnF queue having not had any messages added when the send is done, or just nothing being present for that target even when subsequently-sent matching stuff has already been received from the other, or both

E.g combine this into one test that sends a non-matching and then matching message in that order and then in the reverse order checks the matching consumer does get a message (receive(timeout)) and the non-matching consumer does not (receiveNoWait) from their respective target queues. Faster and more reliable, verification of what actually happened / actual use. (It could perhaps even monitor the mirror SnF queue during and after, verifying nothing got added to the SnF queue for the non-matching case, and is added for the matching case).

artemis-server/src/main/resources/schema/artemis-configuration.xsd
iliya-gr: This is still questionable for me. I don't know what is the better name for this attribute. Because any *-match name will suggest same behavior as in receiver/sender and I suspect that nobody will want to create mirror for each address. I chose this name because it has same meaning as in the cluster configuration and I didn't want to expand terminology.

So I am still in doubt. I don't think we should align mirror and receiver/sender filter behavior because they have different use cases.

gemmellr: I dont think the use case for this configuration is all that different, in this case its about what things you want to mirror from, vs the others being about what you want to send from or recieve to. An address 'match' for a mirror definition certainly doesnt need to mean 'create a seperate mirror for every address' just because it might mean that for e.g a sender. I also presume it couldnt really mean that given each mirror uses a specific queue named based on its parent broker-connections name.

Using *address-match*, *address-pattern(s)*, *address-filter(s)*, or anything to similar effect, all seem equally reasonable and far more appropriate to me than just "address" does for a configuration that generally seems to be expected _not_ to just have a single address in it.

I'm not saying the config has to be the same, but I am saying it seems more obvious, and I dont as yet see any compelling reason it should be entirely different when its basically for doing entirely equivalent things, configuring applicable addresses of interest.

In any case how it is configured and what it does should be documented.

iliya-gr: I agree with all of this, maybe except reusing `address-match`. I think using the same term that has different semantic/behaviour is bad and will lead users to misconfiguration. 

Anyway, I have opted for `address-filter`, but if there are any objections I can change it to anything else.

gemmellr: I agree address-match shouldn't be used if the configuration within it is actually entirely different, as it is now. The comment had the second aspect that its not so obvious to me that it should be entirely different when both things are about governing matched addresses.

docs/user-manual/en/amqp-broker-connections.md
```diff
@@ -100,6 +100,26 @@ Add a `<mirror>` element within the `<amqp-connection>` element to configure mir
 
 The following optional arguments can be utilized:
 
+* `address-filter`: Each amqp connection mirror configuration only applies to addresses (and its queues) that match the specified `address-filter` attribute, that is, mirroring events will only be sent to the target broker for addresses that match this filter. 
+  An address is matched on the mirror amqp connection when it begins with the string specified in this field. The `address-filters` attribute on a mirror amqp connection also supports comma separated lists and an exclude syntax `!`. 
+  To prevent an address from being matched on this mirror amqp connection, prepend an address string with `!`. If this attribute is not specified, then all addresses and their queues will be mirrored.
```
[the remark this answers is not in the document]
gemmellr: I would also possibly suggest moving the `address-filter` option down below the others, as they are relatively tiny by comparison.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPReplicaTest.java
```diff
@@ -473,6 +473,73 @@ public void testNoAddressWithAnnotations() throws Exception {
...
+         Assert.assertEquals("will receive", message.getBody(String.class));
+         connection.close();
+      }
+
+      {
+         ConnectionFactory factory = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + AMQP_PORT);
+         Connection connection = factory.createConnection();
+         Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+         connection.start();
+         MessageConsumer consumer = session.createConsumer(session.createQueue("replicated"));
```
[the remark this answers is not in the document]
iliya-gr: Indeed, it should be `nonReplicated`. I have added double send to prevent such mistype.

clebertsuconic:
I will spend some time reviewing this tomorrow after I finish the release on 2.22.
it seems in good shape.

iliya-gr:
> I will spend some time reviewing this tomorrow after I finish the release on 2.22.

Do I need to squash commits before?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -162,6 +172,12 @@ public void createQueue(QueueConfiguration queueConfiguration) throws Exception
          }
          return;
       }
+      if (ignoreAddress(queueConfiguration.getAddress())) {
+         if (logger.isTraceEnabled()) {
+            logger.trace("Skipping create " + queueConfiguration + ", queue address" + queueConfiguration.getAddress() + " is excluded");
```
gemmellr, line 177: Given the filter does actually have 'exclusions' and it might not actually be excluded, so much as just not-included, I would reword this to avoid saying it is 'excluded' specifically, but instead something more general like 'doesnt match filter'.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -208,6 +232,13 @@ public void sendMessage(Message message, RoutingContext context, List<MessageRef
          return;
       }
 
+      if (ignoreAddress(context.getAddress(message))) {
+         if (logger.isTraceEnabled()) {
+            logger.trace("server " + server + " is discarding send to excluded address");
```
gemmellr, line 237: In addition to the "excluded" comment, for this one I'd also note it should say _which address_ is in play otherwise its not very helpful.

gemmellr:
> > I will spend some time reviewing this tomorrow after I finish the release on 2.22.
> 
> Do I need to squash commits before?

It can be reviewed without squashing them but its never going to be merged like this without squashing, so I'd suggest just doing so.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
iliya-gr: Agree. I have changed both messages.

clebertsuconic:
this looks awesome.. merging it.

clebertsuconic:
thanks @iliya-gr and @gemmellr

gemmellr:
Just to be sure; you didnt have any concerns with the different broker-connections bits having a completely different forms of governing applicable addresses?

gemmellr:
The question was for @clebertsuconic, about the long discussion we already had, where I said it is not obvious to me that two broker-connection elements (which could even be on the same broker-connection at this point) should have completely different ways for doing what is largely the same thing, governing the applicable addresses.

clebertsuconic:
@gemmellr can we have a separate discussion on that? I missed it... and I'm not sure I understand it. are you concerned with users messing up the config and having two connections filtering the same address?

gemmellr:
Not that, although I guess its possible.

Mostly it just seems bizarre for elements of the same thing (a broker connection) to support such entirely distinct ways of doing what amounts to almost exactly the same thing (configuring the applicable addresses). One way supports a list of entries, the other doesnt. One supports wildcarding, the other doesnt. One supports exclusions, the other doesnt.

PR #4066, opened 2022-05-05
ARTEMIS-3815 Await target mirror controller message send completion

iliya-gr:
Await IO completion before retrying message ack, so that pending sendMessage transaction be completed at retry and message would be placed either to intermediate store or to messageReferences or to paged store.
  
Remove duplicate debug messages.


clebertsuconic:
you have some changes on your commit that are not really part of your change.. it makes a bit difficult to Rebase it :)


You should have done the reformat on a separate commit.

clebertsuconic:
not that the changes are not welcomed.. it's just that I'm having a hard time with the rebase.. but leave it with me :)

iliya-gr:
Yes, there are some work should be done here.
I will switch it to draft and rebase myself.

PR #4069, opened 2022-05-08
ARTEMIS-3821 provide readable version of property absoluteExpiryTime (and a few more)

erwindon:
See [ARTEMIS-3821](https://issues.apache.org/jira/browse/ARTEMIS-3821)


brusdev (approved):
LGTM +1

PR #4075, opened 2022-05-12
ARTEMIS-3826 - allow address/queue configuration via properties

gtully:


artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -923,7 +923,6 @@ private Map<String, String> getMapOfChildPropertyElements(Node item) {
       }
       return properties;
    }
-
```
brusdev, line 926: I guess this line was not removed on purpose

gtully: indeed, good catch, thanks.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
```diff
@@ -722,6 +722,23 @@ public void testSetNestedPropertyOnExistingCollectionEntryViaMappedNotation() th
...
    }
 
+
+   @Test
+   public void testAddressViaProperties() throws Throwable {
+      ConfigurationImpl configuration = new ConfigurationImpl();
+
+      Properties properties = new Properties();
+
+      properties.put("addressConfigurations.\"LB.TEST\".queueConfigs.\"LB.TEST\".routingType", "ANYCAST");
```
brusdev, line 732: If you add another queue property configuration.getAddressConfigurations().get(0).getQueueConfigs().size() will be 2 because QueueConfiguration name is a SimpleString
`properties.put("addressConfigurations.\"LB.TEST\".queueConfigs.\"LB.TEST\".durable", "true");`

gtully: that is a problem, let me revisit.... not sure if the lookup needs to be smplestring aware of whether name should always be a string or both!
Probably ensuring the name accessor is a string and hiding the simple string as an implementation detail is best

gtully: the latter is a simpler change too.

gtully: excellent review, thanks!

brusdev (approved):

PR #4079, opened 2022-05-16
ARTEMIS-3834 Include paged messages when running "Send messages to DLA"

AntonRoskvist:


clebertsuconic:
@AntonRoskvist do you mind fixing QueueControlUsingCoreTest.testSendToDLAIncludesPagedMessages ?

AntonRoskvist:
@clebertsuconic Sure no problem, what is the issue you are seeing?

clebertsuconic:
QueueControlUsingCoreTest extends the test you wrote to try management through core messages
there's a proxy, and there's a failure because:

Error Message
class java.lang.Long cannot be cast to class java.lang.Integer (java.lang.Long and java.lang.Integer are in module java.base of loader 'bootstrap')


Basically the proxy method you wrote seems wrong.


if you can't fix it.. I will figure out later...

AntonRoskvist:
Sure no problem, I'll give that a try right away

AntonRoskvist:
@clebertsuconic 
I fixed it so the test passes but I'm not 100% sure this is the correct fix... to be honest I cannot really figure out why it receives a Long as response instead of the expected Integer... stepping through the code with a debugger I can see that my the actual return value from "QueueImpl#sendMessagesToDeadLetterAddress()" is an int just like before the change but I guess it gets cast at some point along the way?

AntonRoskvist:
Travis CI failed at: 
testSoakHornetQ(org.apache.activemq.artemis.tests.compatibility.HornetQSoakTest) 
but only for jdk11... I'm assuming that's unrelated for now

jbertram:
The real way to fix `QueueControlUsingCoreTest` is just to tell `invokeOperation` to cast it to an `Integer.class`, e.g.:
```java
         @Override
         public int sendMessagesToDeadLetterAddress(final String filterStr) throws Exception {
            return (Integer) proxy.invokeOperation(Integer.class,"sendMessagesToDeadLetterAddress", filterStr);
         }
```
That said, I'm concerned that this introduces an inconsistency with `sendMessageToDeadLetterAddress(long)` which *doesn't* include paged messages. I think either both of these methods should work on paged messages or neither of them should.

Thoughts?

AntonRoskvist:
I agree, that makes sense. I will fix both the test and also make sure that the sendMessageToDeadLetterAddress(long) includes paged messages. Thanks

AntonRoskvist:
@jbertram @clebertsuconic Hope this is looking better, otherwise just let med know.

PR #4080, opened 2022-05-16
ARTEMIS-3835 - deprecate addressesSettings in place of addressSetting…

gtully:
…s on the Configuration bean


gtully:
@jbertram what is the story with compatability-tests - i have reverted some of the changes that errorred out, like FQQNConsumerTest but left others, will revert if there are more failures.
It makes sense to leave some references to the old for compatibility, maybe only when a deprecated attribute is removed should the compat tests be modified.
are there rules here?

clebertsuconic:
@gtully the compatibility tests will use older versions. so if you changed configuration and is using an older version and it's failing, that's a good sign the compatibility tests are working.

gtully:
makes sense. I updates the tests to avoid the use of the newly deprecated method but that is of course not compatible with the old version. Sorted now, test as it was. Thanks.

gtully:
I did: https://github.com/apache/activemq-artemis/compare/e427fdda2d8af4cff57fa940667e5652a5c62d09..57e95445db21f546e959707a0c1f54c97492eb31 resolved the failure.
I see the hornetqsoak test failing, but that seems to be fail quite a lot, independent of this change.

brusdev (approved):

PR #4093, opened 2022-05-23
ARTEMIS-3808 Fixing ActiveMQServerControlTest

clebertsuconic:


clebertsuconic:
@jbertram this is fixing a test you recently added. Please check if this fix would still convey what you originally intended?

jbertram:
Neither `testRestartEmbeddedWebServerException` nor `testRestartEmbeddedWebServerTimeout` will work with the proxy as the proxy wraps the exceptions. Therefore, both tests should be ignored when using the proxy.

Also, `testRestartEmbeddedWebServerException` will fail even when run without the proxy due to wrapping the exception.

I've fixed both issues in #4094.

PR #4094, opened 2022-05-24
ARTEMIS-3808 fix tests

jbertram:


gemmellr:
Looks good (though given how the test works it works its probably worth cutting the extra 9ms delay like @clebertsuconic did in his version hehe)

jbertram:
@gemmellr, fixed both.

PR #4098, opened 2022-05-25
ARTEMIS-3848 High CPU usage on ReadWriteLocks

clebertsuconic:
This is caused by too many entires on the HashMap for ThreadLocals.
Also: I'm reviewing some readlock usage on the StorageManager to simplify things a little bit.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
@@ -62,7 +62,7 @@ public class NettyConnection implements Connection {
     * here for when the connection (or Netty Channel) becomes available again.
     */
    private final List<ReadyListener> readyListeners = new ArrayList<>();
-   private final FastThreadLocal<ArrayList<ReadyListener>> localListenersPool = new FastThreadLocal<>();
+   private static final FastThreadLocal<ArrayList<ReadyListener>> readyListenersPool = new FastThreadLocal<>();
 
    private final boolean batchingEnabled;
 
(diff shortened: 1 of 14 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
```diff
@@ -174,12 +174,20 @@ public static JournalContent getType(byte type) {
 
    protected final ScheduledExecutorService scheduledExecutorService;
 
-   protected final ReentrantReadWriteLock storageManagerLock = new ReentrantReadWriteLock(true);
+   protected final ReentrantReadWriteLock storageManagerLock = new ReentrantReadWriteLock(false);
```
franz1981, line 177: Good catch!

clebertsuconic:
@franz1981 it might be an option: however I don't want to break anything else now. I may do that as a major separate task.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
```diff
@@ -138,51 +137,47 @@ public boolean isOpen() {
 
    @Override
    public final void fireReady(final boolean ready) {
-      ArrayList<ReadyListener> readyToCall = localListenersPool.get();
-      if (readyToCall != null) {
-         localListenersPool.set(null);
+
+      // We are reusing a previously created ArrayList for this localArray
+      ArrayList<ReadyListener> localArrayList = readyListenersPool.get();
```
clebertsuconic, line 142: @franz1981 I'm changing this code you wrote here. This was also introducing many entires on the ThreadLocal.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
```diff
@@ -62,7 +61,7 @@ public class NettyConnection implements Connection {
     * here for when the connection (or Netty Channel) becomes available again.
     */
    private final List<ReadyListener> readyListeners = new ArrayList<>();
-   private final FastThreadLocal<ArrayList<ReadyListener>> localListenersPool = new FastThreadLocal<>();
+   private static final ThreadLocal<ArrayList<ReadyListener>> readyListenersPool = new ThreadLocal<>();
```
clebertsuconic, line 64: since we are now fixing whatever was broken with ThreadLocals before. I don't think we need the FastThreadLocal any more.

clebertsuconic: Not really... in previous versions this was just using ThreadLocal. (2.16.0 is using a regular ThreadLocal).

in more recent versions we started using the FastThreadLocal, that's just a ThreadLocal with its own Map away. (there will be a single HashMap entry on the Thread.threadLocals.

this is still an issue.. it should be static.


I could still use the FastThreadlocal here. but I don't see a reason for.

clebertsuconic: @gemmellr even if I used FastThreadLocal here, it should still be a static declaration.

gemmellr: I was commenting on your comment around FastThreadLocal...not anything about the separate change to static. I expect the change to use FastThreadLocal (in https://github.com/apache/activemq-artemis/commit/6126d926ddcdf34b740603e506878059ea663680) had nothing to do with "whatever was broken with ThreadLocals before" but was simply because its optimised for use by Netty threads and the expectation was a Netty thread will likely be calling it.

clebertsuconic: @gemmellr I don't see a real advantage really... I will just make it static though.

PR #4100, opened 2022-06-01
Update ActiveMQServerImpl.java

Yesenkov:
This fixes the situation in ActiveMQServerImpl.createSession () ActiveMQServerImpl.checkSessionLimit () is called with the ServerSessionImpl.validatedUser parameter, but to count the list of sessions, the ActiveMQServerImpl.getSessionCountForUser () method is called, which already iterates over the sessions, ServerSessionImpl.getUsername () is called. In the case of certificate authentication via the TextFileCertificateLoginModule, getUsername() is always null for client connections, while ServerSessionImpl.validatedUser is set to the normal user ID via the call to securityStore.authenticate() at the very beginning of the ActiveMQServerImpl.createSession() method.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
@@ -1729,7 +1729,16 @@ public class ActiveMQServerImpl implements ActiveMQServer {
       int sessionCount = 0;
 
       for (Entry<String, ServerSession> sessionEntry : sessions.entrySet()) {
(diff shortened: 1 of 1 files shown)


jbertram:
This really needs a test to validate the fix and also to ensure there are no regressions in the future.

jbertram:
Basic tests for resource limits are at `org.apache.activemq.artemis.tests.integration.server.ResourceLimitTest`. A test using certificates is at `org.apache.activemq.artemis.tests.integration.security.SecurityTest#testJAASSecurityManagerAuthenticationWithCerts()`. Take bits and pieces from both and create a new class `org.apache.activemq.artemis.tests.integration.server.ResourceLimitTestWithCerts`.

PR #4102, opened 2022-06-02
Update ActiveMQServerImpl.java

Yesenkov:
This fixes the situation in ActiveMQServerImpl.createSession () ActiveMQServerImpl.checkSessionLimit () is called with  the ServerSessionImpl.validatedUser parameter, but to count the list of sessions, the ActiveMQServerImpl.getSessionCountForUser () method is called, which already iterates over the sessions, ServerSessionImpl.getUsername () is called. In the case of certificate authentication via the TextFileCertificateLoginModule, getUsername() is always null for client connections, while ServerSessionImpl.validatedUser is set to the normal user ID via the call to securityStore.authenticate() at the very beginning of the ActiveMQServerImpl.createSession() method.
Test org.apache.activemq.artemis.tests.integration.server.ResourceLimitTestWithCerts.java


Yesenkov:
> ActiveMQServerImpl.getSessionCountForUser

done

gemmellr:
This PR is targetting 2.19.x, it should instead target main (if it were to be backported the change would then be picked from main; however there are no plans to do further 2.19.x releases).

The test should be added in the same commit as the change, i.e in this PR, rather than in a seperate PR as you have created (where you actually did target main). You can update a PR by force pushing to the same PR branch after you have made the needed changes, no need to close and reopen.

Aside, at the very least your test is missing the licence header, which failed the Travis CI build, although it looks like there may be other style issues as well. You can enable the GitHub Actions based CI jobs in your own fork repo (see the Actions tab) which has more granular checks, and use it to run the jobs on your own fork before opening/updating a PR (since you already opened this one, you would need to use a seperate testing branch to pre-test your updates before updating the original branch).

Yesenkov:
> This PR is targetting 2.19.x, it should instead target main (if it were to be backported the change would then be picked from main; however there are no plans to do further 2.19.x releases).
> 
> The test should be added in the same commit as the change, i.e in this PR, rather than in a seperate PR as you have created (where you actually did target main). You can update a PR by force pushing to the same PR branch after you have made the needed changes, no need to close and reopen.
> 
> Aside, at the very least your test is missing the licence header, which failed the Travis CI build, although it looks like there may be other style issues as well. You can enable the GitHub Actions based CI jobs in your own fork repo (see the Actions tab) which has more granular checks, and use it to run the jobs on your own fork before opening/updating a PR (since you already opened this one, you would need to use a seperate testing branch to pre-test your updates before updating the original branch).

I made new PR.
But, main branch contain same problem, possible need same test.

gemmellr:
Im not sure what part of my comment was unclear, but again more succinctly:

- Update *this* PR as requested when feedback is given, dont create new ones.
- Initial PRs should always target *main*, not branches.
- There are no plans to release another 2.19.x so nothing should be targetting it regardless.

Yesenkov:
> Im not sure what part of my comment was unclear, but again more succinctly:
> 
> * Update _this_ PR as requested when feedback is given, dont create new ones.
> * Initial PRs should always target _main_, not branches.
> * There are no plans to release another 2.19.x so nothing should be targetting it regardless.

Ok, it's clear now.
I changed the patch and test branch to main.

PR #4113, opened 2022-06-14
ARTEMIS-3858 fix navigation from consumer/producer columns of a session to those consumers/producers

erwindon:
see https://issues.apache.org/jira/browse/ARTEMIS-3858

### DIFF
--- a/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/consumers.js
+++ b/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/consumers.js
@@ -201,10 +201,10 @@ var Artemis;
         };
 
         if (artemisConsumer.consumer) {
-            Artemis.log.debug("navigating to consumer = " + artemisConsumer.consumer.session);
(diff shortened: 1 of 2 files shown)


artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/consumers.js
```diff
@@ -204,7 +204,7 @@ var Artemis;
             Artemis.log.debug("navigating to consumer = " + artemisConsumer.consumer.session);
```
brusdev, line 204: Should we change the log too?
```suggestion
            Artemis.log.debug("navigating to consumer = " + artemisConsumer.consumer.sessionID);
```

erwindon: <humble-voice>of course!

erwindon: updated the PR (squashed the additional changes)

PR #4116, opened 2022-06-17
ARTEMIS-3863 - allow Role configuration via properties

gtully:


gtully:
Going to leave as draft as I need to address the augment variation - where the xml security-settings config exists and needs to be tweaked. The file config parser needs to create a RoleSet to facilitate this. As is, it works nicely when there are no security settings in the xml config

gtully:
added additional test for preexisting xml - the putSecurityRoles api sorts the creation of the RoleSet, all is good

brusdev (approved):
Nice solution!

PR #4118, opened 2022-06-20
ARTEMIS-3866 Authorize management message sending using context subject

brusdev:


brusdev:
@andytaylor I pushed another commit to move the user preferences to the send page view

use-current-logon-user-selected

use-current-logon-user-unselected

jbertram (approved):

PR #4120, opened 2022-06-23
ARTEMIS-3872 send scheduled msg immediately via mgmnt

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2018,7 +2018,20 @@ public void incrementMesssagesAdded() {
 
    @Override
    public void deliverScheduledMessages() throws ActiveMQException {
-      List<MessageReference> scheduledMessages = scheduledDeliveryHandler.cancel(null);
+      internalDeliverScheduleMessages(scheduledDeliveryHandler.cancel(ref -> true));
+   }
+
+   @Override
+   public void deliverScheduledMessages(String filterString) throws ActiveMQException {
+      internalDeliverScheduleMessages(scheduledDeliveryHandler.cancel(ref -> filterString == null || filterString.length() == 0 ? true : FilterImpl.createFilter(filterString).match(ref.getMessage())));
```
brusdev, line 2026: The performance could be improved creating the filter only one time before creating the comparator
```suggestion
      final Filter filter = filterString == null || filterString.length() == 0 ? null : FilterImpl.createFilter(filterString);
      internalDeliverScheduleMessages(scheduledDeliveryHandler.cancel(ref -> filter == null ? true : filter.match(ref.getMessage())));
```

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerImpl.java
```diff
@@ -135,6 +134,10 @@ public List<MessageReference> cancel(final Filter filter) throws ActiveMQExcepti
       return refs;
    }
 
+   public interface CancelComparator {
+      boolean compare(MessageReference ref) throws ActiveMQException;
```
brusdev, line 138: The ActiveMQException isn't required if the the filter is created before calling the comparator and the CancelComparator could be replaced by java.util.function.Predicate<MessageReference>

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
jbertram: Great suggestion!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerImpl.java
jbertram: This would be a good improvement. Thanks!

PR #4123, opened 2022-06-27
ARTEMIS-3874: reduce use of test-jar and related module interdependencies

gemmellr:
Described at https://issues.apache.org/jira/browse/ARTEMIS-3874


gtully (approved):
nice work, thanks!

PR #4126, opened 2022-06-29
ARTEMIS-3875 - adding consumer and producer metrics

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-3875

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
@@ -103,6 +103,7 @@ import org.apache.activemq.artemis.core.server.ConnectorServiceFactory;
 import org.apache.activemq.artemis.core.server.Consumer;
 import org.apache.activemq.artemis.core.server.Divert;
 import org.apache.activemq.artemis.core.server.JournalType;
+import org.apache.activemq.artemis.core.server.MessageReference;
 import org.apache.activemq.artemis.core.server.Queue;
 import org.apache.activemq.artemis.core.server.ServerConsumer;
 import org.apache.activemq.artemis.core.server.ServerProducer;
@@ -2752,7 +2753,19 @@ public class ActiveMQServerControlImpl extends AbstractControl implements Active
    }
 
    private JsonObject toJSONObject(ServerConsumer consumer) throws Exception {
-      JsonObjectBuilder obj = JsonLoader.createObjectBuilder().add("consumerID", consumer.getID()).add("connectionID", consumer.getConnectionID().toString()).add("sessionID", consumer.getSessionID()).add("queueName", consumer.getQueue().getName().toString()).add("browseOnly", consumer.isBrowseOnly()).add("creationTime", consumer.getCreationTime()).add("deliveringCount", consumer.getDeliveringMessages().size());
+      List<MessageReference> deliveringMessages = consumer.getDeliveringMessages();
+      JsonObjectBuilder obj = JsonLoader.createObjectBuilder()
+            .add("consumerID", consumer.getID())
+            .add("connectionID", consumer.getConnectionID().toString())
+            .add("sessionID", consumer.getSessionID())
+            .add("queueName", consumer.getQueue().getName().toString())
+            .add("browseOnly", consumer.isBrowseOnly())
+            .add("creationTime", consumer.getCreationTime())
+            .add("deliveringCount", deliveringMessages.size())
+            .add("deliveringSize", getDeliveringMessageSize(deliveringMessages))
+            .add("messagesAcknowledged", consumer.getMessagesAcknowledged())
+            .add("lastDeliveredTimeElapsed", getLastDeliveredTimeElapsed(consumer))
+            .add("lastAcknowledgedTimeElapsed", getLastAcknowledgedTimeElapsed(consumer));
       if (consumer.getFilter() != null) {
          obj.add("filter", consumer.getFilter().getFilterString().toString());
       }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
@@ -1770,8 +1770,18 @@ public class QueueControlImpl extends AbstractControl implements QueueControl {
 
             if (consumer instanceof ServerConsumer) {
                ServerConsumer serverConsumer = (ServerConsumer) consumer;
-
-               JsonObjectBuilder obj = JsonLoader.createObjectBuilder().add("consumerID", serverConsumer.getID()).add("connectionID", serverConsumer.getConnectionID().toString()).add("sessionID", serverConsumer.getSessionID()).add("browseOnly", serverConsumer.isBrowseOnly()).add("creationTime", serverConsumer.getCreationTime());
+               List<MessageReference> deliveringMessages = consumer.getDeliveringMessages();
+               JsonObjectBuilder obj = JsonLoader.createObjectBuilder()
+                     .add("consumerID", serverConsumer.getID())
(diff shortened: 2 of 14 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/ConsumerView.java
```diff
@@ -58,6 +60,16 @@ public JsonObjectBuilder toJson(ServerConsumer consumer) {
...
 
+      List<MessageReference> deliveringMessages = consumer.getDeliveringMessages();
+      long deliveringMessageSize = 0;
+      for (int i = 0; i < deliveringMessages.size(); i++) {
+         MessageReference messageReference =  deliveringMessages.get(i);
+         deliveringMessageSize += messageReference.getMessage().getEncodeSize();
+      }
+      long currentTime = System.currentTimeMillis();
+      long lastDeliveredTimeElapsed = consumer.getLastDeliveredTime() == 0 ? 0 : currentTime - consumer.getLastDeliveredTime();
+      long lastAckTimeElapsed = consumer.getLastAcknowledgedTime() == 0 ? 0 : currentTime - consumer.getLastAcknowledgedTime();
```
gemmellr, line 71: Seems familiar..more scope for sharing impl?

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -2752,7 +2753,27 @@ public String listAllConsumersAsJSON() throws Exception {
    }
 
    private JsonObject toJSONObject(ServerConsumer consumer) throws Exception {
-      JsonObjectBuilder obj = JsonLoader.createObjectBuilder().add("consumerID", consumer.getID()).add("connectionID", consumer.getConnectionID().toString()).add("sessionID", consumer.getSessionID()).add("queueName", consumer.getQueue().getName().toString()).add("browseOnly", consumer.isBrowseOnly()).add("creationTime", consumer.getCreationTime()).add("deliveringCount", consumer.getDeliveringMessages().size());
+      List<MessageReference> deliveringMessages = consumer.getDeliveringMessages();
```
clebertsuconic, line 2756: this is going to hold a lock on the consumer while the messages are copied over.


if you have a bad consumer leaking on acks, this could make things worse to the broker.

Besides this will hold a lock on the ServerConsumer.

Perhaps I'm being over zealous here... but I have seen cases where using the console to find what was wrong actually made things worse.

That's my only concern here.

clebertsuconic: it would be better if these metrics were pre-calculated in the Consumer and the console would just get them.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/ConsumerView.java
```diff
@@ -111,6 +128,27 @@ public Object getField(ServerConsumer consumer, String fieldName) {
...
+            return deliveringMessageSize;
+         case MESSAGES_ACKNOWLEDGED:
+            return consumer.getMessagesAcknowledged();
+         case LAST_DELIVERED_TIME_ELAPSED: {
+            long currentTime = System.nanoTime();
+            return consumer.getLastDeliveredTime() == 0 ? 0 : currentTime - consumer.getLastDeliveredTime();
+         }
+         case LAST_ACKNOWLEDGED_TIME_ELAPSED: {
+            long currentTime = System.nanoTime();
+            return consumer.getLastAcknowledgedTime() == 0 ? 0 : currentTime - consumer.getLastAcknowledgedTime();
```
gemmellr, line 149: Ditto

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
```diff
@@ -1770,8 +1770,26 @@ public String listConsumersAsJSON() throws Exception {
 
             if (consumer instanceof ServerConsumer) {
                ServerConsumer serverConsumer = (ServerConsumer) consumer;
-
-               JsonObjectBuilder obj = JsonLoader.createObjectBuilder().add("consumerID", serverConsumer.getID()).add("connectionID", serverConsumer.getConnectionID().toString()).add("sessionID", serverConsumer.getSessionID()).add("browseOnly", serverConsumer.isBrowseOnly()).add("creationTime", serverConsumer.getCreationTime());
+               List<MessageReference> deliveringMessages = consumer.getDeliveringMessages();
+               long deliveringMessageSize = 0;
+               for (int i = 0; i < deliveringMessages.size(); i++) {
+                  MessageReference messageReference =  deliveringMessages.get(i);
+                  deliveringMessageSize += messageReference.getMessage().getEncodeSize();
```
michaelpearce-gain, line 1777: If we want this for metrics would be safer to keep this count as a metric on the consumer adding the size and removing it from the metric as messages are added or ack'd this would also save on the above mentioned npe / concurrent issue

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/predicate/ConsumerFilterPredicate.java
```diff
@@ -59,6 +62,28 @@ public boolean test(ServerConsumer consumer) {
             return matches(server.getSessionByID(consumer.getSessionID()).getRemotingConnection().getTransportConnection().getLocalAddress());
          case REMOTE_ADDRESS:
             return matches(server.getSessionByID(consumer.getSessionID()).getRemotingConnection().getTransportConnection().getRemoteAddress());
+         case DELIVERING_COUNT:
+            return matches(consumer.getDeliveringMessages().size());
+         case DELIVERING_SIZE:
+            List<MessageReference> deliveringMessages = consumer.getDeliveringMessages();
+            long deliveringMessageSize = 0;
```
michaelpearce-gain, line 69: Ditto other comment i would be tracking this at the consumerimpl as messages added or acl'd the size is added and removed so were not later looping, and like wise avoid concurrent issues

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -2752,7 +2753,27 @@ public String listAllConsumersAsJSON() throws Exception {
...
+            .add("connectionID", consumer.getConnectionID().toString())
+            .add("sessionID", consumer.getSessionID())
+            .add("queueName", consumer.getQueue().getName().toString())
+            .add("browseOnly", consumer.isBrowseOnly())
+            .add("creationTime", consumer.getCreationTime())
+            .add("deliveringCount", deliveringMessages.size())
+            .add("deliveringSize", deliveringMessageSize)
+            .add("messagesAcknowledged", consumer.getMessagesAcknowledged())
+            .add("lastDeliveredTimeElapsed", lastDeliveredTimeElapsed)
+            .add("lastAcknowledgedTimeElapsed", lastAckTimeElapsed);
```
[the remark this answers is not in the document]
andytaylor: these already existed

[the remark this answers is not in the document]
gemmellr: Doesnt seem like it, I'm looking at a whole load of new lines being added here.

andytaylor:
I've made some changes but dont have time currently to handle any possible NPE's around messages, I'll close and someone else can pick it up if they feel the need

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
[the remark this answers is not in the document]
andytaylor: they were on a single line but I have replaced anyway

PR #4130, opened 2022-06-30
ARTEMIS-3815 proper retry through IOCompletion when message not found on target queue on Mirror

clebertsuconic:


clebertsuconic:
this supersedes #4066

iliya-gr:
My first intent was to use number in retry. I thought it would make a performAck method too complicated and confusing. But now it seems fine. I am totally ok with that.

PR #4135, opened 2022-07-03
ARTEMIS-3794 System Property Encryption Support

ryan-highley:
Adds support for standard Java TLS and ActiveMQ Artemis-specific override
encrypted system property values for the key store and trust store
passwords

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnector.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnector.java
@@ -123,6 +123,7 @@ import org.apache.activemq.artemis.spi.core.remoting.ssl.SSLContextFactoryProvid
 import org.apache.activemq.artemis.utils.ConfigurationHelper;
 import org.apache.activemq.artemis.utils.FutureLatch;
 import org.apache.activemq.artemis.utils.IPV6Util;
+import org.apache.activemq.artemis.utils.PasswordMaskingUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.lang.invoke.MethodHandles;
@@ -153,6 +154,7 @@ public class NettyConnector extends AbstractConnector {
    public static final String ACTIVEMQ_TRUSTSTORE_TYPE_PROP_NAME = "org.apache.activemq.ssl.trustStoreType";
    public static final String ACTIVEMQ_TRUSTSTORE_PATH_PROP_NAME = "org.apache.activemq.ssl.trustStore";
    public static final String ACTIVEMQ_TRUSTSTORE_PASSWORD_PROP_NAME = "org.apache.activemq.ssl.trustStorePassword";
+   public static final String ACTIVEMQ_SSL_PASSWORD_CODEC_CLASS_PROP_NAME = "org.apache.activemq.ssl.passwordCodec";
 
    // Constants for HTTP upgrade
    // These constants are exposed publicly as they are used on the server-side to fetch
@@ -177,7 +179,6 @@ public class NettyConnector extends AbstractConnector {
       config.put(TransportConstants.PORT_PROP_NAME, TransportConstants.DEFAULT_PORT);
       DEFAULT_CONFIG = Collections.unmodifiableMap(config);
    }
-
 
    private final boolean serverConnection;
 
@@ -229,6 +230,8 @@ public class NettyConnector extends AbstractConnector {
    private String localAddress;
 
    private int localPort;
+
+   private String passwordCodecClass;
 
    private String keyStoreProvider;
 
@@ -324,6 +327,7 @@ public class NettyConnector extends AbstractConnector {
                          final ClientProtocolManager protocolManager) {
       this(configuration, handler, listener, closeExecutor, threadPool, scheduledThreadPool, protocolManager, false);
    }
+
    public NettyConnector(final Map<String, Object> configuration,
                          final BufferHandler handler,
                          final BaseConnectionLifeCycleListener<?> listener,
@@ -395,6 +399,8 @@ public class NettyConnector extends AbstractConnector {
 
       localPort = ConfigurationHelper.getIntProperty(TransportConstants.LOCAL_PORT_PROP_NAME, TransportConstants.DEFAULT_LOCAL_PORT, configuration);
       if (sslEnabled) {
+         passwordCodecClass = ConfigurationHelper.getStringProperty(ActiveMQDefaultConfiguration.getPropPasswordCodec(), TransportConstants.DEFAULT_PASSWORD_CODEC_CLASS, configuration);
+
          keyStoreProvider = ConfigurationHelper.getStringProperty(TransportConstants.KEYSTORE_PROVIDER_PROP_NAME, TransportConstants.DEFAULT_KEYSTORE_PROVIDER, configuration);
 
          keyStoreType = ConfigurationHelper.getStringProperty(TransportConstants.KEYSTORE_TYPE_PROP_NAME, TransportConstants.DEFAULT_KEYSTORE_TYPE, configuration);
@@ -438,6 +444,7 @@ public class NettyConnector extends AbstractConnector {
          keyStorePath = TransportConstants.DEFAULT_KEYSTORE_PATH;
          keyStorePassword = TransportConstants.DEFAULT_KEYSTORE_PASSWORD;
          keyStoreAlias = TransportConstants.DEFAULT_KEYSTORE_ALIAS;
+         passwordCodecClass = TransportConstants.DEFAULT_PASSWORD_CODEC_CLASS;
          trustStoreProvider = TransportConstants.DEFAULT_TRUSTSTORE_PROVIDER;
          trustStoreType = TransportConstants.DEFAULT_TRUSTSTORE_TYPE;
          trustStorePath = TransportConstants.DEFAULT_TRUSTSTORE_PATH;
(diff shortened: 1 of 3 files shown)


tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/remoting/impl/netty/NettyConnectorTest.java
```diff
@@ -171,6 +173,70 @@ public void bufferReceived(final Object connectionID, final ActiveMQBuffer buffe
...
+         public void bufferReceived(final Object connectionID, final ActiveMQBuffer buffer) {
+         }
+      };
+
+      DefaultSensitiveStringCodec codec = new DefaultSensitiveStringCodec();
+
+      System.setProperty(NettyConnector.JAVAX_KEYSTORE_PATH_PROP_NAME, "client-keystore.jks");
+      System.setProperty(NettyConnector.JAVAX_KEYSTORE_PASSWORD_PROP_NAME, PasswordMaskingUtil.wrap(codec.encode("bad password")));
+      System.setProperty(NettyConnector.JAVAX_TRUSTSTORE_PATH_PROP_NAME, "server-ca-truststore.jks");
+      System.setProperty(NettyConnector.JAVAX_TRUSTSTORE_PASSWORD_PROP_NAME, PasswordMaskingUtil.wrap(codec.encode("bad password")));
```
brusdev, line 225: The system properties should be cleared before ending the test because they could affect other tests.

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/remoting/impl/netty/NettyConnectorTest.java
```diff
@@ -334,6 +400,59 @@ public void bufferReceived(final Object connectionID, final ActiveMQBuffer buffe
...
+      params.put(TransportConstants.SSL_ENABLED_PROP_NAME, true);
+
+      NettyConnector connector = new NettyConnector(params, handler, listener, executorService, Executors.newCachedThreadPool(ActiveMQThreadFactory.defaultThreadFactory()), Executors.newScheduledThreadPool(5, ActiveMQThreadFactory.defaultThreadFactory()));
+
+      DefaultSensitiveStringCodec codec = new DefaultSensitiveStringCodec();
+
+      System.setProperty(NettyConnector.ACTIVEMQ_KEYSTORE_PATH_PROP_NAME, "client-keystore.jks");
+      System.setProperty(NettyConnector.ACTIVEMQ_KEYSTORE_PASSWORD_PROP_NAME, PasswordMaskingUtil.wrap(codec.encode("securepass")));
+      System.setProperty(NettyConnector.ACTIVEMQ_TRUSTSTORE_PATH_PROP_NAME, "server-ca-truststore.jks");
+      System.setProperty(NettyConnector.ACTIVEMQ_TRUSTSTORE_PASSWORD_PROP_NAME, PasswordMaskingUtil.wrap(codec.encode("securepass")));
```
brusdev, line 420: The system properties should be cleared before ending the test because they could affect other tests.

brusdev:
I see a test failure due to system properties set by the new test:
https://github.com/apache/activemq-artemis/runs/7171602475?check_suite_focus=true
```
WARN  [io.netty.channel.ChannelInitializer] Failed to initialize a channel. Closing: [id: 0xcf2b8ab7]: java.lang.Exception: Failed to find a store at unknownclient-keystore.jks
	at org.apache.activemq.artemis.core.remoting.impl.ssl.SSLSupport.validateStoreURL(SSLSupport.java:356)
	at org.apache.activemq.artemis.core.remoting.impl.ssl.SSLSupport.loadKeystore(SSLSupport.java:304)
	at org.apache.activemq.artemis.core.remoting.impl.ssl.SSLSupport.loadKeyManagerFactory(SSLSupport.java:332)
	...
```

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnector.java
```diff
@@ -584,15 +585,15 @@ public synchronized void start() {
             realTrustStorePassword = trustStorePassword;
          } else {
             realKeyStorePath = Stream.of(System.getProperty(ACTIVEMQ_KEYSTORE_PATH_PROP_NAME), System.getProperty(JAVAX_KEYSTORE_PATH_PROP_NAME), keyStorePath).map(v -> useDefaultSslContext ? keyStorePath : v).filter(Objects::nonNull).findFirst().orElse(null);
-            realKeyStorePassword = Stream.of(System.getProperty(ACTIVEMQ_KEYSTORE_PASSWORD_PROP_NAME), System.getProperty(JAVAX_KEYSTORE_PASSWORD_PROP_NAME), keyStorePassword).map(v -> useDefaultSslContext ? keyStorePassword : v).filter(Objects::nonNull).findFirst().orElse(null);
+            realKeyStorePassword = processSslPasswordProperty(Stream.of(System.getProperty(ACTIVEMQ_KEYSTORE_PASSWORD_PROP_NAME), System.getProperty(JAVAX_KEYSTORE_PASSWORD_PROP_NAME), keyStorePassword).map(v -> useDefaultSslContext ? keyStorePassword : v).filter(Objects::nonNull).findFirst().orElse(null));
```
brusdev, line 588: The system properties have a grater visibility than any connector properties and multiple connector could rely on the same system properties so it doesn't sound good to me that a connector property affects the use of a system property.

ryan-highley: The stream processing logic is identical to the previous code--I just wrapped the stream result in a method call to handle the possibility of an encrypted system property. However, the encrypted password logic should only be invoked when the value is not the keyStorePassword or trustStorePassword captured with ConfigurationHelper#getPasswordProperty(...) previously as that value has already been decrypted if necessary.

I've tweaked the logic to call processSslPasswordProperty(...) only when the existing stream processing result is one of the system properties, not a query string value.

brusdev: I'm not sure to get the purpose of this PR, why does a user need to set the password-codec in the connector url and the passwords in system properties?

ryan-highley: This PR enables ENC(...) handling for the 'org.apache.activemq.ssl.keyStorePassword', 'o.a.a.ssl.trustStorePassword', 'javax.net.ssl.keyStorePassword', and 'j.n.ssl.trustStorePassword' system property values. The connector URL 'keyStorePassword' and 'trustStorePassword' query string handling is unchanged as these values already are decrypted as needed through the ConfigurationHelper#getPasswordProperty(...) method in lines 401 and 409.

The system properties obviously aren't part of the 'configuration' Map so the ConfigurationHelper#getPasswordProperty(...) method can't be used directly. The processSslPasswordProperty(...) method duplicates the behavior of calling getPasswordProperty(...) using the same ActiveMQDefaultConfiguration.getPropMaskPassword() and .getPropPasswordCodec() used for the connector URL password handling for the 'keyStorePassword' and 'trustStorePassword'.

Hopefully, that makes more sense as to the intent. ARTEMIS-3794 is just a JIRA I saw similar to the AMQP broker connection user and password attribute encryption handling.

brusdev: If the purpose is to override keyStore/trustStore passwords obtained from topology with encrypted system properties maybe we should use system properties to set the mask mode and the password codec too, i.e. 'org.apache.activemq.ssl.passwordCodec', the server and the client could use a different mask mode or password codec.

ryan-highley: That's an interesting thought. I based these updates on the approach handling keyStorePassword and trustStorePassword also applying adequately for their corresponding system property values.

I'm all for providing essential flexibility, but I've also never had to provide an alternate password encoding codec to appease PEN testing or production environment security scans. I'm happy to address that if you see the need for that additional configuration and associated incremental complexity.

Perhaps we can see if anyone files an enhancement request for specifying an alternate mask mode and/or password codec with system properties?

jbertram:
@ryan-highley, can you squash and rebase?

ryan-highley:
@jbertram @brusdev 

Rebased and squashed these updates.

Any thoughts on adding two additional system properties for the password mask and codec configurations? Could always handle those in a different RFE as well.

jbertram:
@ryan-highley, I'm not really in favor of a "useMaskedPassword" property. We've moved away from this style of configuration in recent years and consolidated on the `ENC()` syntax which can be automatically detected.

ryan-highley:
> @ryan-highley, I'm not really in favor of a "useMaskedPassword" property. We've deprecated this style of configuration in recent years and consolidated on the `ENC()` syntax which can be automatically detected.

Agree wholeheartedly!

The parameter-based approach is documented as being deprecated, and I'm perfectly happy to remove its support with the system properties here. I included the system property solely for "principle of least surprise" since elsewhere the "useMaskedPassword" configurations do still work for backwards compatibility.

I'll remove the system property and gladly clean up the associated extraneous unit tests!

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnector.java
```diff
@@ -755,6 +771,23 @@ public void initChannel(Channel channel) throws Exception {
       logger.debug("Started {} Netty Connector version {} to {}:{}", connectorType, TransportConstants.NETTY_VERSION, host, port);
    }
 
+   private String processSslPasswordProperty(String password, String codecClass) {
```
jbertram, line 774: Instead of using `ConfigurationHelper` you could simply use `PasswordMaskingUtil` (which is what `ConfigurationHelper` does eventually), e.g.:
```java
   private String resolveMask(String password, String codecClass) {
      try {
         return PasswordMaskingUtil.resolveMask(password, codecClass);
      } catch (Exception e) {
         throw new RuntimeException(e);
      }
   }
```

jbertram:
I'm ready to merge this except for the comment I've made. Let me know what you think.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnector.java
ryan-highley: Oh, yes. Much better! I wanted to reuse the `ConfigurationHelper` password logic since the extra Boolean logic was already handled there. Without that `useMaskedPassword` that reasoning and need goes right out the window.

Updated and pushed.

jbertram:
@ryan-highley, thanks for the contribution and also thanks for being willing to discuss and accept feedback. It's great working with you!

PR #4143, opened 2022-07-12
ARTEMIS-3890 - rework LVQ implementation to ensure all messages get d…

gtully:
…elivered, replacement of lvq now tied to the deliver loop. Fix issue with duplicates - bug in LinkedListImpl`


jbertram (commented):
LGTM

PR #4144, opened 2022-07-12
ARTEMIS-3767 Fix replication incompatibility between pre 2.18.0 and SNAPSHOT (WIP)

jsmucr:
This PR attempts to solve the issue described in https://issues.apache.org/jira/browse/ARTEMIS-3767.

TL;DR replication between 2.17.0 and newer Artemis versions is broken since 2.18.0.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -175,7 +175,7 @@
       <jakarta.resource-api.version.alt>2.1.0</jakarta.resource-api.version.alt>
 
       <!-- used on tests -->
-      <groovy.version>4.0.0</groovy.version>
+      <groovy.version>4.0.3</groovy.version>
       <vertx.version>3.9.4</vertx.version>
       <hadoop.minikdc.version>3.3.1</hadoop.minikdc.version>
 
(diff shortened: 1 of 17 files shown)


jsmucr:
I've got a problem with this. Changes in #3680 should have introduced a new version of the `REPLICATION_START_FINISH_SYNC` packet as there's suddenly more data in it. This tainted the Artemis releases up to the current snapshot and I don't know how to overcome this unless there's a bugfix release for all recent versions starting with 2.18.0.

clebertsuconic:
I don't think you need the Server compatibility Test.


Please take a look at MultiVersionReplicatest under compatibility-tests

jsmucr:
I didn't notice this test before. Maybe it wasn't on the 2.18.0 I was starting with, and after many failures when trying to make it work with the compatibility-tests base, I created my own (more Groovy-ish way, without shared bindings and the `GroovyRun` helper class). Well, nevermind. I'll throw it away then.
The test does not work with 2.17.0 now as it appears to assume the existence of some newer methods on older implementations, but that's easy to fix, I suppose.

gemmellr:
Also being discussed on https://lists.apache.org/thread/bf8t8gkzg374351d2rjd6pkzr13olclb

clebertsuconic:
lets move the discussion to my new PR?

this PR is being replaced by https://github.com/apache/activemq-artemis/pull/4150

PR #4147, opened 2022-07-13
ARTEMIS-3894 - add conversion from string for list - allow core bridg…

gtully:
…e static connector config via properties

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
@@ -550,6 +550,17 @@ public class ConfigurationImpl implements Configuration, Serializable {
          }
       }, TransformerConfiguration.class);
 
+      beanUtils.getConvertUtils().register(new Converter() {
+         @Override
+         public <T> T convert(Class<T> type, Object value) {
+            List convertedValue = new ArrayList<String>();
+            for (String entry : value.toString().split(",")) {
+               convertedValue.add(entry);
+            }
+            return (T) convertedValue;
+         }
+      }, java.util.List.class);
+
       // support 25K or 25m etc like xml config
       beanUtils.getConvertUtils().register(new Converter() {
          @Override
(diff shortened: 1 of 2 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -550,6 +550,17 @@ public <T> T convert(Class<T> type, Object value) {
          }
       }, TransformerConfiguration.class);
 
+      beanUtils.getConvertUtils().register(new Converter() {
+         @Override
+         public <T> T convert(Class<T> type, Object value) {
+            List convertedValue = new ArrayList<String>();
+            for (String entry : value.toString().split(",")) {
```
clebertsuconic, line 557: I have been following some error prone messages, and this particular line will introduce an error prone warning.

WARNING] /Users/clebertsuconic/work/apache/activemq-artemis/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/uri/URIFactory.java:[112,86] [StringSplitter] String.split(String) has surprising behavior
    (see https://errorprone.info/bugpattern/StringSplitter)
  Did you mean 'List<String> connectorURIS = Splitter.on(',').splitToList(split[0].substring(split[0].indexOf('(') + 1));'?
[

gtully: thanks!

gtully: that warn rule must be quite specific, I am not seeing it with -Derrorprone or -Pjdk11to15-errorprone with openjdk 11

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -2957,7 +2957,7 @@ private Object newNamedInstanceForCollection(String collectionPropertyName, Obje
 
                // create one and initialise with name
                try {
-                  Object instance = candidate.getParameterTypes()[candidate.getParameterCount() - 1].getDeclaredConstructor().newInstance(null);
+                  Object instance = candidate.getParameterTypes()[candidate.getParameterCount() - 1].getDeclaredConstructor().newInstance((java.lang.Object[])null);
```
clebertsuconic, line 2960: couldn't this just be simple as:

```java
Object instance = candidate.getParameterTypes()[candidate.getParameterCount() - 1].getDeclaredConstructor().newInstance();
```

clebertsuconic: that is.. just no arguments passed

gtully: for sure, thanks. I need some H²0

PR #4150, opened 2022-07-15
ARTEMIS-3767 Incompatibility on replication between 2.17 and current version

clebertsuconic:


clebertsuconic:
as Robbie said, this is Also being discussed on https://lists.apache.org/thread/bf8t8gkzg374351d2rjd6pkzr13olclb

PR #4151, opened 2022-07-16
ARTEMIS-3896 clarify logging for transactional ops

jbertram:
Both audit logging and logging from the LoggingActiveMQServerPlugin are
unclear as they relate to transactional sends and acks. Both essentially
ignore the transaction which makes it appear that an operation has taken
place when, in fact, it hasn't (e.g. a transactional ack is rolled back
but the log indicates the ack went through).

This commit fix this with the following changes:

 - Log details when a send or ack is added to a transaction.
 - Log details when the transaction is committed.
 - Log when the transaction is rolled back.
 - Include transaction details in the relevant DEBUG logs.
 - Simplify INFO level logging for sends & acks in
LoggingActiveMQServerPlugin. Ensure details are in the DEBUG logs.

Other changes:

 - Make capitalization more consistent in a handful of audit logs.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/logs/AuditLogger.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/logs/AuditLogger.java
@@ -2420,13 +2420,13 @@ public interface AuditLogger extends BasicLogger {
     *
     * */
    //hot path log using a different logger
-   static void coreSendMessage(Subject user, String remoteAddress, String messageToString, Object context) {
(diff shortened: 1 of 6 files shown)


clebertsuconic:
@jbertram can you send this change to the branch new-logging instead?

jbertram:
@clebertsuconic, will do.

gemmellr:
I think it would be better to just submit this to main, regardless whether before or after the release intended soon...this change isnt related to the SLF4J etc changes on the branch so they should land separately rather than be mashed together, and it would seem nicer to land them first for the existing stuff and rebase the branch as needed.

PR #4158, opened 2022-07-29
in activemq plugin do client validate and it can not send MQTTReasonCodes.CLIENT_IDENTIFIER_NOT_VALID to mqtt client

archlevel:
in activemq plugin do client validate and it can not send MQTTReasonCodes.CLIENT_IDENTIFIER_NOT_VALID to mqtt client


gemmellr:
As with your other PR I dont see any actual useful change here, simply merging existing commits. If you have some changes you should add those to your own branch and raise a PR from it.

You should raise a JIRA for an issue you are looking to address, and then reference it in the commit message (see: most previous commits).

PRs should also target main unless first discussed otherwise, not branches like 2.23.x, as fixes will be picked from main where appropriate. There almost certainly wont be any more 2.23.x releases though as 2.24.0 is almost out, perhaps even late today, having been under release vote for a couple of days already.

PR #4159, opened 2022-07-29
NO-JIRA Making dependency-scan output only visible on maven debug

clebertsuconic:


clebertsuconic:
I will actually change this.. I will write it to the ./target as a file, so it will just work from any IDE integration

gemmellr:
Glad I suggested this, much nicer: ~35 lines of nearly no-op output during build, as opposed to nearly 4000 hehe.

clebertsuconic:
@gemmellr thanks for the suggestion...
I also changed the build now to generate the class path on files... no more need to look at the output... just build it.. and run the compatibility tests from the IDE as nothing changed

PR #4160, opened 2022-08-01
MQTTReasonCodes byte loss of precision，must int type

archlevel:
MQTTReasonCodes byte loss of precision，must int type


jbertram:
This PR is just moving a few commits from the `2.23.x` branch to the `main` branch. There's nothing related to MQTT.

PR #4163, opened 2022-08-02
ARTEMIS-3916 fix OpenWire selector with JMSMessageID

jbertram:


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/SelectorTranslator.java
```diff
@@ -54,6 +56,17 @@ public static String convertToActiveMQFilterString(final String selectorString)
 
    }
 
+   public static String convertOpenWireToActiveMQFilterString(final String selectorString) {
+      if (selectorString == null) {
+         return null;
+      }
+
+      String filterString = SelectorTranslator.convertToActiveMQFilterString(selectorString);
+      filterString = SelectorTranslator.parse(filterString, "AMQUserID", AMQ_MSG_MESSAGE_ID.toString());
```
clebertsuconic, line 65: We should rather make this method public than bringing OpenWire concept into commons.

If it was really needed (very difficult to go around) it's ok.. but this one is easy to fix.

I have the change here and I'm amending as you before merging it.

jbertram: Adding `convertOpenwireToActiveMQFilterString` here seemed reasonable (and consistent) given the existence of `convertHQToActiveMQFilterString`.

clebertsuconic: @jbertram the HQ one was wrong.. but it was small... (I'm moving it now on a separate commit)

PR #4166, opened 2022-08-03
Update address-settings.md

TheCycoONE:
Add a note about match specificity


jbertram:
I added further clarification before I merged this. See dc2b67331dcd3e52254e1567c08424795a11347f.

Thanks for your contribution!

PR #4167, opened 2022-08-03
ARTEMIS-3918 support FQQN + anycast + redistribution

jbertram:
When a message is sent to an anycast queue via FQQN on one node of a
cluster and then a consumer is created on that same anycast queue via
FQQN on another node in the cluster the message is not redistributed to
the node with the consumer.

This commit fixes this use-case primarily by including the FQQN info in
the notification messages sent to other nodes in the cluster.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/BindingsImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/BindingsImpl.java
@@ -192,7 +192,7 @@ public final class BindingsImpl implements Bindings {
          logger.tracef("Redistributing message %s", message);
       }
 
-      final SimpleString routingName = originatingQueue.getName();
+      final SimpleString routingName = CompositeAddress.isFullyQualified(message.getAddress()) && originatingQueue.getRoutingType() == RoutingType.ANYCAST ? CompositeAddress.extractAddressName(message.getAddressSimpleString()) : originatingQueue.getName();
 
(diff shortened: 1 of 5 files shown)


brusdev:
What is the expected behavior with multiple FQQN anycast queues per address?

jbertram:
Messages will routed round-robin to the anycast queues on an address.

rtista:
> Messages will routed round-robin to the anycast queues on an address.

Hey @jbertram, does this mean that a routed message can go from one anycast FQQN queue to another with a different queue name? For example, let's assume an address "test" with anycast queues "test-1" and "test-2", could it occur that a message is placed on "test::test-1" and then routed to "test::test-2"?

jbertram:
> ...does this mean that a routed message can go from one anycast FQQN queue to another with a different queue name?

No. A message sent to an FQQN goes only to that FQQN.

I was answering that question in the context of a message sent to an address with multiple anycast queues (i.e. not using FQQN).

rtista:
> No. A message sent to an FQQN goes only to that FQQN.

I believe I'm currently having this problem. I'm using addresses to group queues from different applications, so the address name is the name of the application and all the application queues are inside it. 

Before 2.25.0, the redistribution in my symmetric  cluster did not work, I've updated it to 2.25.0 and it works, but I'm having messages delivered to a FQQN queue and redistributed to another FQQN queue that is in the same address.

I've looked at the code, and the test case you've created only uses an address with one queue, is it possible that with an address with multiple queues it will fail?

jbertram:
@rtista, this PR is already merged so there's not much else to be done here. If you've got a way to reproduce the problem then please open a [Jira](https://issues.apache.org/jira/browse/ARTEMIS) explaining your use-case and attach your reproducer. Thanks!

PR #4168, opened 2022-08-05
ARTEMIS-3922 Reducing contention on Throwable#getOurStackTrace

MrEasy:
By allowing to pass caller's classname directly to org.apache.activemq.artemis.utils.ActiveMQThreadFactory#defaultThreadFactory instead of calculating it from stack.


jbertram:
I merged this, but I modified the PR to go ahead and remove the default constructor completely.

PR #4177, opened 2022-08-15
ARTEMIS-3938 Adding logger-properties option to Artemis CLI create command

gaohoward:
This feature is to enable creating broker instance with custom logging configurations from CLI.
The command would be like:

artemis create broker0 --logger-properties <logger properties file>

The *logger properties file* is a java properties file containing broker's jboss logging properties intended to update the default logging properties.

example of the properties file:

{noformat}
logger.handlers=CONSOLE
logger.org.apache.activemq.audit.base.handlers=CONSOLE
logger.org.apache.activemq.audit.base.level=WARN
logger.level=WARN
{noformat}

Once applied those properties will be configured in the logging.properties of the created broker instance config dir.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
@@ -45,6 +45,7 @@ import org.apache.activemq.artemis.cli.commands.util.HashUtil;
 import org.apache.activemq.artemis.cli.commands.util.SyncCalculation;
 import org.apache.activemq.artemis.core.server.JournalType;
 import org.apache.activemq.artemis.core.server.cluster.impl.MessageLoadBalancingType;
+import org.apache.activemq.artemis.logs.JBossLoggingConfiguration;
 import org.apache.activemq.artemis.nativo.jlibaio.LibaioContext;
 import org.apache.activemq.artemis.nativo.jlibaio.LibaioFile;
 import org.apache.activemq.artemis.utils.FileUtil;
@@ -345,6 +346,9 @@ public class Create extends InputAbstract {
 
    @Option(name = "--jdbc-lock-expiration", description = "Lock expiration")
    long jdbcLockExpiration = ActiveMQDefaultConfiguration.getDefaultJdbcLockExpirationMillis();
+
+   @Option(name = "--logger-properties", description = "If set, it points to a java properties file containing logging properties")
+   String loggerProperties = null;
 
    private boolean IS_WINDOWS;
    private boolean IS_CYGWIN;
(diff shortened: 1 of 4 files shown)


gemmellr:
This seems quite complicated for just specifying an alternative logging config.

It appears to support updating only a subset of the base config, e.g adding new things, or overriding an existing property value, without necessarily specifying the full configuration, also jumping hoops to make sure other areas are updated appropriately (e.g adding a new logger to the list of loggers). Its not clear it really supports removing any existing stuff (except perhaps by re-definining them an empty value for an existing key, and hoping the impl doesnt barf on such prop values).

It doesnt seem very nice/maintainable that this programatically duplicates the regular logging.properties file contents within JBossLoggingConfiguration as its starting point for creating a modified config. The two definitions could easily get out of sync, the basic props and the programmatic equivalent. In particular, the fact we are also working to remove JBL usage just now also makes it seem like a less than ideal time to introduce something new as complicated as the JBossLoggingConfiguration class which is heavily dependent on that config structure, and would need adapted (if it even could be) to a replacement.

Why not not simply have the CLI option allow providing a _complete_  logging configuration file the user can provide? Then if the option is specified, that file is copied into place as the instance is created, and if it isnt specified then the default one is copied into place the same as it is currently. Much simpler, likely a handful of lines rather than the several hundred used here, with no need for the mechanism to be handling the structure of the file at all.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -826,7 +830,16 @@ public Object run(ActionContext context) throws Exception {
...
 
-      writeEtc(ETC_LOGGING_PROPERTIES, etcFolder, null, false);
+      if (loggerProperties == null) {
+         writeEtc(ETC_LOGGING_PROPERTIES, etcFolder, null, false);
+      } else {
+         context.out.println("applying custom logger.properties from " + loggerProperties);
+         JBossLoggingConfiguration loggingConfig = JBossLoggingConfiguration.createDefaultArtemisLoggingProperties();
+         loggingConfig.updateFrom(new File(loggerProperties));
+         try (PrintWriter writer = new PrintWriter(new File(etcFolder, ETC_LOGGING_PROPERTIES))) {
+            loggingConfig.save(writer);
```
clebertsuconic, line 840: this is extremely fragile... you are mixing a set of properties from the user with existing properties in Jboss Logging...


there's no way to limit what the users can do here....


I would rather have the user providing the whole logging file instead...

but then the whole operation on the CLI gets moot anyways.. as you could just write to the logging.properties yourself.


Besides... we are about to remove Jboss Logging in favor of SLF4J. This PR is a big no for me.

artemis-cli/src/test/java/org/apache/activemq/cli/test/JBossLoggingConfigurationTest.java
```diff
@@ -0,0 +1,178 @@
...
+import java.io.StringWriter;
+import java.net.URL;
+import java.util.Properties;
+
+import org.apache.activemq.artemis.logs.JBossLoggingConfiguration;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+public class JBossLoggingConfigurationTest {
```
clebertsuconic, line 31: We are about to ditch Jboss-logging in favor of SLF4j.

artemis-commons/src/main/java/org/apache/activemq/artemis/logs/JBossLoggingConfiguration.java
```diff
@@ -0,0 +1,724 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.UUID;
+
+/**
+ * This class manages jboss-logging configuration file
+ * in a ordered manner.
+ */
+public class JBossLoggingConfiguration {
```
clebertsuconic, line 37: We are about to remove Jboss logging... I would not add such thing here adding tight coupling with Jboss-logging.


just pass in the whole file.

jbertram:
This change is tightly coupling the broker with JBoss Logging. We plan to move away from JBoss Logging in the not-too-distant future to SLF4J + Log4j 2.x. In order to continue working this code would have to be updated to work with Log4j 2.x (assuming that's even possible). That said, the logging _implementation_ is flexible with SLF4J so this feature would break with any other implementation other than Log4j 2.x. Tightly coupling the broker to a logging implementation is not tenable.

In my opinion, the simplest & best solution for custom configuration is to just copy the custom config file to `etc` _after_ the broker has been created. Having the broker do this adds unnecessary complexity and further expands the CLI's `create` command which was originally designed to be _very_ simple.

PR #4178, opened 2022-08-15
ARTEMIS-3940 Address.maxSizeMessage is ignored on FAIL and Blocking Paging Policy

clebertsuconic:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/paging/MaxMessagesPagingTest.java
```diff
@@ -301,4 +314,120 @@ public void testMaxOnAddressHitGlobal() throws Exception {
...
+
+      AddressSettings set = new AddressSettings();
+      set.setAddressFullMessagePolicy(AddressFullMessagePolicy.FAIL);
+      if (global) {
+         config.setGlobalMaxMessages(10);
+      }
+
+      settings.put( "FAIL_MAX_MESSAGES_CORE", set);
+      settings.put( "FAIL_MAX_MESSAGES_OPENWIRE", set);
+      settings.put( "FAIL_MAX_MESSAGES_AMQP", set);
```
clebertsuconic, line 342: ... I will remove these lines.. forgot those hanging.

jbertram:
The commit message says "ignorecase." It seems like this should just be "ignored."

clebertsuconic:
@jbertram ouch.. thanks....


I typed :set ignorecase in vim, and somehow my mind did that trick to me :)

PR #4183, opened 2022-08-22
ARTEMIS-3875 - adding consumer and producer metrics

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-3875 - new PR for the stats changes, it wouldnt let me reopen that PR

I've refactored incorporating the feedback, all stats are tracked in the consumer now plus more tests and some tidy up

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerImpl.java
@@ -44,6 +44,8 @@ public class ClientProducerImpl implements ClientProducerInternal {
 
    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+   private final int id;
+
    private final SimpleString address;
 
    private final ClientSessionInternal session;
@@ -75,7 +77,10 @@ public class ClientProducerImpl implements ClientProducerInternal {
                              final boolean autoGroup,
                              final SimpleString groupID,
                              final int minLargeMessageSize,
-                             final SessionContext sessionContext) {
+                             final SessionContext sessionContext,
+                             final int producerID) {
+      this.id = producerID;
+
       this.sessionContext = sessionContext;
 
       this.session = session;
@@ -197,12 +202,19 @@ public class ClientProducerImpl implements ClientProducerInternal {
       return producerCredits;
    }
 
+   @Override
+   public int getID() {
+      return id;
+   }
+
    private void doCleanup() {
       if (address != null) {
          session.returnCredits(address);
       }
 
       session.removeProducer(this);
+
+      sessionContext.removeProducer(id);
 
       closed = true;
    }
@@ -290,7 +302,7 @@ public class ClientProducerImpl implements ClientProducerInternal {
 
       theCredits.acquireCredits(creditSize);
 
-      sessionContext.sendFullMessage(msgI, sendBlocking, handler, address);
+      sessionContext.sendFullMessage(msgI, sendBlocking, handler, address, id);
    }
 
    private void checkClosed() throws ActiveMQException {
@@ -381,7 +393,7 @@ public class ClientProducerImpl implements ClientProducerInternal {
             lastChunk = pos >= bodySize;
             SendAcknowledgementHandler messageHandler = lastChunk ? handler : null;
 
-            int creditsUsed = sessionContext.sendServerLargeMessageChunk(msgI, -1, sendBlocking, lastChunk, bodyBuffer.array(), messageHandler);
+            int creditsUsed = sessionContext.sendServerLargeMessageChunk(msgI, -1, sendBlocking, lastChunk, bodyBuffer.array(), id, messageHandler);
 
             credits.acquireCredits(creditsUsed);
          }
@@ -492,7 +504,7 @@ public class ClientProducerImpl implements ClientProducerInternal {
                      headerSent = true;
                      sendInitialLargeMessageHeader(msgI, credits);
                   }
-                  int creditsSent = sessionContext.sendLargeMessageChunk(msgI, messageSize.get(), sendBlocking, true, buff, reconnectID, handler);
+                  int creditsSent = sessionContext.sendLargeMessageChunk(msgI, messageSize.get(), sendBlocking, true, buff, reconnectID, id, handler);
                   credits.acquireCredits(creditsSent);
                }
             } else {
@@ -501,7 +513,7 @@ public class ClientProducerImpl implements ClientProducerInternal {
                   sendInitialLargeMessageHeader(msgI, credits);
                }
 
-               int creditsSent = sessionContext.sendLargeMessageChunk(msgI, messageSize.get(), sendBlocking, false, buff, reconnectID, handler);
+               int creditsSent = sessionContext.sendLargeMessageChunk(msgI, messageSize.get(), sendBlocking, false, buff, reconnectID, id, handler);
                credits.acquireCredits(creditsSent);
             }
          }
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
@@ -161,6 +161,8 @@ public final class ClientSessionImpl implements ClientSessionInternal, FailureLi
 
    private final CoreMessageObjectPools coreMessageObjectPools = new CoreMessageObjectPools();
 
+   private AtomicInteger producerIDs = new AtomicInteger();
+
    ClientSessionImpl(final ClientSessionFactoryInternal sessionFactory,
                      final String name,
                      final String username,
@@ -2031,9 +2033,19 @@ public final class ClientSessionImpl implements ClientSessionInternal, FailureLi
                                                  final int maxRate) throws ActiveMQException {
       checkClosed();
 
-      ClientProducerInternal producer = new ClientProducerImpl(this, address, maxRate == -1 ? null : new TokenBucketLimiterImpl(maxRate, false), autoCommitSends && blockOnNonDurableSend, autoCommitSends && blockOnDurableSend, autoGroup, groupID == null ? null : new SimpleString(groupID), minLargeMessageSize, sessionContext);
+      ClientProducerInternal producer = new ClientProducerImpl(this,
+                                                                      address,
+                                                                      maxRate == -1 ? null : new TokenBucketLimiterImpl(maxRate, false),
+                                                                      autoCommitSends && blockOnNonDurableSend,
+                                                                      autoCommitSends && blockOnDurableSend,
+                                                                      autoGroup, groupID == null ? null : new SimpleString(groupID),
+                                                                      minLargeMessageSize,
+                                                                      sessionContext,
+                                                                      producerIDs.incrementAndGet());
 
       addProducer(producer);
+
+      sessionContext.createProducer(producer);
 
       return producer;
    }
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/CreateProducerMessage.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/CreateProducerMessage.java
@@ -1,0 +1,83 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.core.protocol.core.impl.wireformat;
+
+import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
+import org.apache.activemq.artemis.api.core.SimpleString;
+import org.apache.activemq.artemis.core.protocol.core.impl.PacketImpl;
+
+import java.util.Objects;
+
+public class CreateProducerMessage  extends PacketImpl {
+   protected int id;
+
+   protected SimpleString address;
+
+   public CreateProducerMessage() {
+      super(PacketImpl.CREATE_PRODUCER);
+   }
+
+   public CreateProducerMessage(int id, SimpleString address) {
+      super(PacketImpl.CREATE_PRODUCER);
+      this.id = id;
+      this.address = address;
+   }
+
+   public int getId() {
+      return id;
+   }
+
+   public void setId(int id) {
+      this.id = id;
+   }
+
+   public SimpleString getAddress() {
+      return address;
+   }
+
+   public void setAddress(SimpleString address) {
+      this.address = address;
+   }
+
+
+
+   @Override
+   public void encodeRest(final ActiveMQBuffer buffer) {
+      buffer.writeInt(id);
+      buffer.writeNullableSimpleString(address);
+   }
+
+   @Override
+   public void decodeRest(final ActiveMQBuffer buffer) {
+      id = buffer.readInt();
+      address = buffer.readNullableSimpleString();
+   }
+
+   @Override
+   public boolean equals(Object o) {
+      if (this == o) return true;
+      if (o == null || getClass() != o.getClass()) return false;
+      if (!super.equals(o)) return false;
+      CreateProducerMessage that = (CreateProducerMessage) o;
+      return Objects.equals(id, that.id) && Objects.equals(address, that.address);
+   }
+
+   @Override
+   public int hashCode() {
+      return Objects.hash(super.hashCode(), id, address);
+   }
+}
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
@@ -41,7 +41,6 @@ import org.apache.activemq.artemis.core.server.MessageReference;
 import org.apache.activemq.artemis.core.server.QueueQueryResult;
 import org.apache.activemq.artemis.core.server.RoutingContext;
 import org.apache.activemq.artemis.core.server.ServerConsumer;
-import org.apache.activemq.artemis.core.server.ServerProducer;
 import org.apache.activemq.artemis.core.server.ServerSession;
 import org.apache.activemq.artemis.core.server.impl.AddressInfo;
 import org.apache.activemq.artemis.core.settings.impl.AddressFullMessagePolicy;
@@ -202,14 +201,14 @@ public class AMQPSessionCallback implements SessionCallback {
                                                            false, // boolean autoCommitAcks,
                                                            false, // boolean preAcknowledge,
                                                            true, //boolean xa,
-                                                           null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain());
+                                                           null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain(), false);
       } else {
          serverSession = manager.getServer().createSession(name, connection.getUser(), connection.getPassword(), ActiveMQClient.DEFAULT_MIN_LARGE_MESSAGE_SIZE, protonSPI.getProtonConnectionDelegate(), // RemotingConnection remotingConnection,
                                                            false, // boolean autoCommitSends
                                                            false, // boolean autoCommitAcks,
                                                            false, // boolean preAcknowledge,
                                                            true, //boolean xa,
-                                                           null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain(), connection.getValidatedUser());
+                                                           null, this, true, operationContext, manager.getPrefixes(), manager.getSecurityDomain(), connection.getValidatedUser(), false);
       }
    }
 
@@ -532,7 +531,7 @@ public class AMQPSessionCallback implements SessionCallback {
       OperationContext oldContext = recoverContext();
       try {
          if (invokeIncoming((AMQPMessage) message, (ActiveMQProtonRemotingConnection) transportConnection.getProtocolConnection()) == null) {
-            serverSession.send(transaction, message, directDeliver, false, routingContext);
+            serverSession.send(transaction, message, directDeliver, receiver.getName(), false, routingContext);
 
             afterIO(new IOCallback() {
                @Override
@@ -744,8 +743,8 @@ public class AMQPSessionCallback implements SessionCallback {
       return protonSPI.invokeOutgoingInterceptors(message, connection);
    }
 
-   public void addProducer(ServerProducer serverProducer) {
-      serverSession.addProducer(serverProducer);
+   public void addProducer(String name, String address) {
+      serverSession.addProducer(name, ProtonProtocolManagerFactory.AMQP_PROTOCOL_NAME, address);
    }
 
    public void removeProducer(String name) {
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPSessionContext.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPSessionContext.java
@@ -24,8 +24,6 @@ import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.activemq.artemis.api.core.ActiveMQSecurityException;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
-import org.apache.activemq.artemis.core.server.ServerProducer;
-import org.apache.activemq.artemis.core.server.impl.ServerProducerImpl;
 import org.apache.activemq.artemis.protocol.amqp.connect.mirror.AMQPMirrorControllerSource;
 import org.apache.activemq.artemis.protocol.amqp.connect.mirror.AMQPMirrorControllerTarget;
 import org.apache.activemq.artemis.protocol.amqp.broker.AMQPSessionCallback;
@@ -160,7 +158,6 @@ public class AMQPSessionContext extends ProtonInitializable {
    }
 
    public void removeReceiver(Receiver receiver) {
-      sessionSPI.removeProducer(receiver.getName());
       receivers.remove(receiver);
    }
 
@@ -229,8 +226,7 @@ public class AMQPSessionContext extends ProtonInitializable {
          AMQPMirrorControllerTarget protonReceiver = new AMQPMirrorControllerTarget(sessionSPI, connection, this, receiver, server);
          protonReceiver.initialize();
          receivers.put(receiver, protonReceiver);
-         ServerProducer serverProducer = new ServerProducerImpl(receiver.getName(), "AMQP", receiver.getTarget().getAddress());
-         sessionSPI.addProducer(serverProducer);
+         sessionSPI.addProducer(receiver.getName(), receiver.getTarget().getAddress());
          receiver.setContext(protonReceiver);
          HashMap<Symbol, Object> brokerIDProperties = new HashMap<>();
          brokerIDProperties.put(AMQPMirrorControllerSource.BROKER_ID, server.getNodeID().toString());
@@ -255,8 +251,7 @@ public class AMQPSessionContext extends ProtonInitializable {
          ProtonServerReceiverContext protonReceiver = new ProtonServerReceiverContext(sessionSPI, connection, this, receiver);
          protonReceiver.initialize();
          receivers.put(receiver, protonReceiver);
-         ServerProducer serverProducer = new ServerProducerImpl(receiver.getName(), "AMQP", receiver.getTarget().getAddress());
-         sessionSPI.addProducer(serverProducer);
+         sessionSPI.addProducer(receiver.getName(), receiver.getTarget().getAddress());
          receiver.setContext(protonReceiver);
          connection.runNow(() -> {
             receiver.open();
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
@@ -270,6 +270,7 @@ public class ProtonServerReceiverContext extends ProtonAbstractReceiver {
    @Override
    public void close(boolean remoteLinkClose) throws ActiveMQAMQPException {
       super.close(remoteLinkClose);
+      sessionSPI.removeProducer(receiver.getName());
       org.apache.qpid.proton.amqp.messaging.Target target = (org.apache.qpid.proton.amqp.messaging.Target) receiver.getRemoteTarget();
       if (target != null && target.getDynamic() && (target.getExpiryPolicy() == TerminusExpiryPolicy.LINK_DETACH || target.getExpiryPolicy() == TerminusExpiryPolicy.SESSION_END)) {
          try {
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
@@ -169,7 +169,9 @@ public class MQTTConnectionManager {
                                                          MQTTUtil.SESSION_AUTO_CREATE_QUEUE,
                                                          server.newOperationContext(),
                                                          session.getProtocolManager().getPrefixes(),
-                                                         session.getProtocolManager().getSecurityDomain(), validatedUser);
+                                                         session.getProtocolManager().getSecurityDomain(),
+                                                         validatedUser,
+                                                         false);
       return (ServerSessionImpl) serverSession;
    }
 
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
@@ -132,7 +132,7 @@ public class AMQSession implements SessionCallback {
       // now
 
       try {
-         coreSession = server.createSession(name, username, password, minLargeMessageSize, connection, true, false, false, false, null, this, true, connection.getOperationContext(), protocolManager.getPrefixes(), protocolManager.getSecurityDomain(), connection.getValidatedUser());
+         coreSession = server.createSession(name, username, password, minLargeMessageSize, connection, true, false, false, false, null, this, true, connection.getOperationContext(), protocolManager.getPrefixes(), protocolManager.getSecurityDomain(), connection.getValidatedUser(), false);
       } catch (Exception e) {
          logger.error("error init session", e);
       }
@@ -392,7 +392,6 @@ public class AMQSession implements SessionCallback {
 
       assert clientId.toString().equals(this.connection.getState().getInfo().getClientId()) : "Session cached clientId must be the same of the connection";
       originalCoreMsg.putStringProperty(MessageUtil.CONNECTION_ID_PROPERTY_NAME, clientId);
-
       /* ActiveMQ failover transport will attempt to reconnect after connection failure.  Any sent messages that did
       * not receive acks will be resent.  (ActiveMQ broker handles this by returning a last sequence id received to
       * the client).  To handle this in Artemis we use a duplicate ID cache.  To do this we check to see if the
@@ -438,7 +437,7 @@ public class AMQSession implements SessionCallback {
                restoreAutoRead();
             }
 
-            getCoreSession().send(coreMsg, false, dest.isTemporary());
+            getCoreSession().send(coreMsg, false, producerInfo.getProducerId().toString(), dest.isTemporary());
 
             if (count == null || count.decrementAndGet() == 0) {
                if (sendProducerAck) {
@@ -462,7 +461,7 @@ public class AMQSession implements SessionCallback {
          Exception exceptionToSend = null;
 
          try {
-            getCoreSession().send(coreMsg, false, dest.isTemporary());
+            getCoreSession().send(coreMsg, false, producerInfo.getProducerId().toString(), dest.isTemporary());
          } catch (Exception e) {
             logger.debug("Sending exception to the client", e);
             exceptionToSend = e;
--- a/artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
+++ b/artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
@@ -41,6 +41,7 @@ import org.apache.activemq.artemis.core.server.LargeServerMessage;
 import org.apache.activemq.artemis.core.server.MessageReference;
 import org.apache.activemq.artemis.core.server.Queue;
 import org.apache.activemq.artemis.core.server.ServerConsumer;
+import org.apache.activemq.artemis.core.server.ServerProducer;
 import org.apache.activemq.artemis.core.server.ServerSession;
 import org.apache.activemq.artemis.core.server.impl.ServerSessionImpl;
 import org.apache.activemq.artemis.spi.core.protocol.RemotingConnection;
@@ -55,10 +56,15 @@ import org.slf4j.LoggerFactory;
 import java.lang.invoke.MethodHandles;
 
 import static org.apache.activemq.artemis.core.protocol.stomp.ActiveMQStompProtocolMessageBundle.BUNDLE;
+import static org.apache.activemq.artemis.core.protocol.stomp.StompProtocolManagerFactory.STOMP_PROTOCOL_NAME;
 
 public class StompSession implements SessionCallback {
 
    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+   private final String senderName = UUIDGenerator.getInstance().generateUUID().toString();
+
+   private boolean createProducer = true;
 
    private final StompProtocolManager manager;
 
@@ -217,6 +223,7 @@ public class StompSession implements SessionCallback {
 
    @Override
    public void closed() {
+      session.removeProducer(senderName);
    }
 
    @Override
@@ -379,10 +386,18 @@ public class StompSession implements SessionCallback {
    }
 
    public void sendInternal(Message message, boolean direct) throws Exception {
-      session.send(message, direct);
+      if (createProducer) {
+         session.addProducer(senderName, STOMP_PROTOCOL_NAME, ServerProducer.ANONYMOUS);
+         createProducer = false;
+      }
+      session.send(message, direct, senderName);
    }
 
    public void sendInternalLarge(CoreMessage message, boolean direct) throws Exception {
+      if (createProducer) {
+         session.addProducer(senderName, STOMP_PROTOCOL_NAME, ServerProducer.ANONYMOUS);
+         createProducer = false;
+      }
       int headerSize = message.getHeadersAndPropertiesEncodeSize();
       if (headerSize >= connection.getMinLargeMessageSize()) {
          throw BUNDLE.headerTooBig();
@@ -402,7 +417,7 @@ public class StompSession implements SessionCallback {
 
       largeMessage.toMessage().putLongProperty(Message.HDR_LARGE_BODY_SIZE, bytes.length);
 
-      session.send(largeMessage.toMessage(), direct);
+      session.send(largeMessage.toMessage(), direct, senderName);
    }
 
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AbstractControl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AbstractControl.java
@@ -122,7 +122,7 @@ public abstract class AbstractControl extends StandardMBean {
                                                          Integer.MAX_VALUE, fakeConnection,
                                                          true, true, false,
                                                          false, address.toString(), fakeConnection.callback,
-                                                         false, new DummyOperationContext(), Collections.emptyMap(), null, validatedUser);
+                                                         false, new DummyOperationContext(), Collections.emptyMap(), null, validatedUser, false);
       try {
          CoreMessage message = new CoreMessage(storageManager.generateID(), 50);
          if (headers != null) {
@@ -159,7 +159,7 @@ public abstract class AbstractControl extends StandardMBean {
          }
 
          // There's no point on direct delivery using the management thread, use false here
-         serverSession.send(message, false);
+         serverSession.send(message, false, null);
          return "" + message.getMessageID();
       } finally {
          try {
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
@@ -79,6 +79,7 @@ import org.apache.activemq.artemis.core.config.TransformerConfiguration;
 import org.apache.activemq.artemis.core.filter.Filter;
 import org.apache.activemq.artemis.core.management.impl.view.AddressView;
 import org.apache.activemq.artemis.core.management.impl.view.ConnectionView;
+import org.apache.activemq.artemis.core.management.impl.view.ConsumerField;
 import org.apache.activemq.artemis.core.management.impl.view.ConsumerView;
 import org.apache.activemq.artemis.core.management.impl.view.ProducerView;
 import org.apache.activemq.artemis.core.management.impl.view.QueueView;
@@ -107,6 +108,7 @@ import org.apache.activemq.artemis.core.server.ConnectorServiceFactory;
 import org.apache.activemq.artemis.core.server.Consumer;
 import org.apache.activemq.artemis.core.server.Divert;
 import org.apache.activemq.artemis.core.server.JournalType;
+import org.apache.activemq.artemis.core.server.MessageReference;
 import org.apache.activemq.artemis.core.server.Queue;
 import org.apache.activemq.artemis.core.server.ServerConsumer;
 import org.apache.activemq.artemis.core.server.ServerProducer;
@@ -2609,7 +2611,7 @@ public class ActiveMQServerControlImpl extends AbstractControl implements Active
       try {
          Set<ServerProducer> producers = new HashSet<>();
          for (ServerSession session : server.getSessions()) {
-            producers.addAll(session.getServerProducers().values());
+            producers.addAll(session.getServerProducers());
          }
          ProducerView view = new ProducerView(server);
          view.setCollection(producers);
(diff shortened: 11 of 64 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/AbstractControl.java
```diff
@@ -50,8 +51,15 @@ public AbstractControl(final Class<?> clazz, final StorageManager storageManager
       this.storageManager = storageManager;
    }
 
+   public long getLastDeliveredTimeElapsed(ServerConsumer consumer) {
+      long currentTime = System.currentTimeMillis();
+      return consumer.getLastDeliveredTime() == 0 ? 0 : currentTime - consumer.getLastDeliveredTime();
```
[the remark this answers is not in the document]
andytaylor: Ive reverted to just passing back the current time as per @gtully suggested

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/TargetAddressInfo.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package org.apache.activemq.artemis.core.server.impl;
+
+
+import java.util.concurrent.atomic.AtomicLong;
+
+class TargetAddressInfo {
+   private AtomicLong messagesSent = new AtomicLong(1);
+
+   private AtomicLong messagesSizeSent = new AtomicLong(0);
```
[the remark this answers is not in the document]
andytaylor: fixed

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -2752,7 +2754,24 @@ public String listAllConsumersAsJSON() throws Exception {
...
+      JsonObjectBuilder obj = JsonLoader.createObjectBuilder()
+            .add(ConsumerField.ID.getAlternativeName(), consumer.getID())
+            .add(ConsumerField.CONNECTION.getAlternativeName(), consumer.getConnectionID().toString())
+            .add(ConsumerField.SESSION.getAlternativeName(), consumer.getSessionID())
+            .add(ConsumerField.QUEUE.getAlternativeName(), consumer.getQueue().getName().toString())
+            .add(ConsumerField.BROWSE_ONLY.getName(), consumer.isBrowseOnly())
+            .add(ConsumerField.CREATION_TIME.getName(), consumer.getCreationTime())
+            // deliveringCount is renamed as MESSAGES_IN_TRANSIT but left in json for backward compatibility
+            .add(ConsumerField.MESSAGES_IN_TRANSIT.getAlternativeName(), consumer.getMessagesInTransit())
+            .add(ConsumerField.MESSAGES_IN_TRANSIT.getName(), consumer.getMessagesInTransit())
```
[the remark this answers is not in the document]
andytaylor: test added

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -506,8 +510,10 @@ public void proceedDeliver(MessageReference reference) throws Exception {
             // The deliverer was prepared during handle, as we can't have more than one pending large message
             // as it would return busy if there is anything pending
             largeMessageDeliverer.deliver();
+            //lastDeliveredTime = System.currentTimeMillis();
          } else {
             deliverStandardMessage(reference, message);
+           // lastDeliveredTime = System.currentTimeMillis();
```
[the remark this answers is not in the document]
andytaylor: fixed

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -2631,7 +2633,7 @@ public void testListConsumersAsJSON() throws Exception {
       Assert.assertEquals(queueName.toString(), first.getString("queueName"));
       Assert.assertEquals(false, first.getBoolean("browseOnly"));
       Assert.assertTrue(first.getJsonNumber("creationTime").longValue() > 0);
-      Assert.assertEquals(0, first.getJsonNumber("deliveringCount").longValue());
+      Assert.assertEquals(0, first.getJsonNumber("messagesInTransit").longValue());
```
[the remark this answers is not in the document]
andytaylor: testadded

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -2242,13 +2246,14 @@ public synchronized RoutingStatus doSend(final Transaction tx,
 
          result = postOffice.route(msg, routingContext, direct);
 
-         Pair<Object, AtomicLong> value = targetAddressInfos.get(msg.getAddressSimpleString());
+         Pair<Object, TargetAddressInfo> value = targetAddressInfos.get(msg.getAddressSimpleString());
 
          if (value == null) {
-            targetAddressInfos.put(msg.getAddressSimpleString(), new Pair<>(msg.getUserID(), new AtomicLong(1)));
+            targetAddressInfos.put(msg.getAddressSimpleString(), new Pair<>(msg.getUserID(), new TargetAddressInfo()));
```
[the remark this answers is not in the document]
andytaylor: fixed

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -3586,6 +4115,14 @@ public void testListConsumers() throws Exception {
          Assert.assertNotEquals("localAddress", "", jsonConsumer.getString("localAddress"));
          Assert.assertNotEquals("remoteAddress", "", jsonConsumer.getString("remoteAddress"));
          Assert.assertNotEquals("creationTime", "", jsonConsumer.getString("creationTime"));
+         Assert.assertNotEquals("messagesInTransit", "", jsonConsumer.getString("messagesInTransit"));
+         Assert.assertNotEquals("messagesInTransitSize", "", jsonConsumer.getString("messagesInTransitSize"));
+         Assert.assertNotEquals("messageDelivered", "", jsonConsumer.getString("messageDelivered"));
+         Assert.assertNotEquals("messagesDeliveredSize", "", jsonConsumer.getString("messagesDeliveredSize"));
+         Assert.assertNotEquals("messagesAcknowledged", "", jsonConsumer.getString("messagesAcknowledged"));
+         Assert.assertNotEquals("lastDeliveredTimeElapsed", "", jsonConsumer.getString("lastDeliveredTimeElapsed"));
+         Assert.assertNotEquals("lastAcknowledgedTimeElapsed", "", jsonConsumer.getString("lastAcknowledgedTimeElapsed"));
```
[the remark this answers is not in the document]
andytaylor: fixed

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -704,9 +708,9 @@ public List<MessageReference> cancelRefs(final boolean failed,
          final List<MessageReference> refs = new ArrayList<>(deliveringRefs.size());
          MessageReference ref;
          while ((ref = deliveringRefs.poll()) != null) {
+            metrics.addAcknowledge(ref.getMessage().getEncodeSize());
```
gemmellr, line 711: This one doesnt look to be synchronized on the consumer the way the others are, worth looking at as it could mean thread safety issues corrupting the metrics updates since it does stuff like -= and ++ on the values.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -2769,27 +2775,554 @@ public void testListAllConsumersAsJSON() throws Exception {
...
+      if (legacyCreateQueue) {
+         server.createQueue(queueName, RoutingType.ANYCAST, queueName, null, false, false);
+      } else {
+         server.createQueue(new QueueConfiguration(queueName).setRoutingType(RoutingType.ANYCAST).setDurable(false));
+      }
+
+      ClientConsumer consumer = session.createConsumer(queueName, null, 100, -1, false);
+      addClientConsumer(consumer);
+      session.start();
+      Thread.sleep(200);
```
gemmellr, line 2831: Necessary? Repeat comment for other tests below

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -2769,27 +2775,554 @@ public void testListAllConsumersAsJSON() throws Exception {
...
+      first = (JsonObject) array.get(0);
+      Assert.assertEquals(0, first.getJsonNumber(ConsumerField.MESSAGES_IN_TRANSIT.getName()).longValue());
+      Assert.assertEquals(0, first.getJsonNumber(ConsumerField.MESSAGES_IN_TRANSIT_SIZE.getName()).longValue());
+      Assert.assertEquals(100, first.getJsonNumber(ConsumerField.MESSAGES_DELIVERED.getName()).longValue());
+      Assert.assertEquals(size, first.getJsonNumber(ConsumerField.MESSAGES_DELIVERED_SIZE.getName()).longValue());
+      Assert.assertEquals(100, first.getJsonNumber(ConsumerField.MESSAGES_ACKNOWLEDGED.getName()).longValue());
+      Assert.assertEquals(100, first.getJsonNumber(ConsumerField.MESSAGES_ACKNOWLEDGED_AWAITING_COMMIT.getName()).longValue());
+
+      session.rollback();
+     // Thread.sleep(1000);
```
gemmellr, line 3011: Delete? Same in other instances.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -4023,6 +4565,114 @@ public void testListProducers() throws Exception {
...
+            producer1.send(message);
+            messagesSize += message.getEncodeSize();
+         }
+         //bring back all producers
+         String producersAsJsonString = serverControl.listProducersInfoAsJSON();
+         JsonArray jsonArray = JsonUtil.readJsonArray(producersAsJsonString);
+
+         JsonObject jsonSession = jsonArray.getJsonObject(0);
+
+         System.out.println("jsonSession = " + jsonSession);
```
gemmellr, line 4658: logger.debug("jsonSession = {}",  jsonSession); ?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -1185,7 +1182,7 @@ public void deleteQueue(final SimpleString queueToDelete) throws Exception {
       }
 
       if (server.getAddressInfo(unPrefixedQueueName) == null) {
-         targetAddressInfos.remove(queueToDelete);
+         producers.remove(queueToDelete.toString());
```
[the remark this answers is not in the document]
andytaylor: Actually the only time it needs removing is when it is an FQQN, if a producer is sending just to an address it doesnt matter if any of its queues is deleted as in the subscription case. That being said you might have multiple produceres on different sessions sending to the same FQQN so it would need to use a queue delet callback so all producers were removed

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
andytaylor: Its actually synced on lock like the others

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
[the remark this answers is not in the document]
gemmellr: > Actually the only time it needs removing is when it is an FQQN, if a producer is sending just to an address it doesnt matter if any of its queues is deleted as in the subscription case. 

Ah I guess, so the thinking is it will just try removing something that never actually existed in the sub case.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
gemmellr: The others dont all seem to be synchronized on lock, e.g. the ServerConsumerImpl#acknowledge and ServerConsumerImpl#individualAcknowledge dont look to be when they are calling the same method.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -2342,8 +2331,8 @@ public Pair<SimpleString, EnumSet<RoutingType>> getAddressAndRoutingTypes(Simple
    @Override
    public void addProducer(ServerProducer serverProducer) {
       serverProducer.setSessionID(getName());
-      serverProducer.setConnectionID(getConnectionID().toString());
-      producers.put(serverProducer.getID(), serverProducer);
+      serverProducer.setConnectionID(getConnectionID() != null ? getConnectionID().toString() : null);
+      producers.put(serverProducer.getAddress(), serverProducer);
```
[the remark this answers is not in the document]
andytaylor: Ive updated the addProducer to take the address so its always correct, Ive also removed the removeproducer below and now use a cllback to remove the producer only if its an FQQm and the queue is removed. I have also added tests for FQQN, JMS (core and AMQP) forr queues, topics and anon producers

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -1958,14 +1931,29 @@ public StorageManager getStorageManager() {
...
-         JsonObjectBuilder producerInfo = JsonLoader.createObjectBuilder().add("connectionID", this.getConnectionID().toString()).add("sessionID", this.getName()).add("destination", entry.getKey().toString()).add("lastUUIDSent", uuid, JsonValue.NULL).add("msgSent", entry.getValue().getB().longValue());
+         JsonObjectBuilder producerInfo = JsonLoader.createObjectBuilder()
+               .add(ProducerField.ID.getName(), getName())
+               .add(ProducerField.CONNECTION_ID.getName(), this.getConnectionID().toString())
+               .add(ProducerField.SESSION.getAlternativeName(), this.getName())
+               .add(ProducerField.CLIENT_ID.getName(), sessionClientID != null ? sessionClientID : "")
+               .add(ProducerField.USER.getName(), getUsername() != null ? getUsername() : "")
+               .add(ProducerField.PROTOCOL.getName(), remotingConnection.getProtocolName())
+               .add(ProducerField.LOCAL_ADDRESS.getName(), localAddress != null ? localAddress : "")
+               .add(ProducerField.REMOTE_ADDRESS.getName(), remoteAddress != null ? remoteAddress : "")
```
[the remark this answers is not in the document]
andytaylor: agreed, I have removed

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -4023,6 +4565,114 @@ public void testListProducers() throws Exception {
       }
    }
 
+   @Test
+   public void testListProducersMessageCounts() throws Exception {
```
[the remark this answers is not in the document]
andytaylor: as per previous comment I have added several JMS tests for both core and AMQP

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
andytaylor: fixed

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
andytaylor: fixed

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
andytaylor: removed

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
andytaylor: I think there is a reason why these 2 methods arent synchronised, @clebertsuconic might know? maybe I need to use Atomic Integers?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
gemmellr: git push :)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
gemmellr: Not sure, just seems like it would be unsafe currently as there are 2 different synchronisations at work, but neither seems to cover all the usages of this. If going to atomic updates, might be a candidate for using a FieldUpdater rather than AtomicInteger

andytaylor: Ive added a fiel updaotr, its only 1 field that can be accessed by multiple threads, the rest are threadsafe

andytaylor:
rebased and comments addressed

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -1513,7 +1524,130 @@ public String getConnectionRemoteAddress() {
...
+         messagesDelivered++;
+         lastDeliveredTime = System.currentTimeMillis();
+      }
+
+      public void addAcknowledge(int encodeSize) {
+         messagesInTransitSizeUpdater.addAndGet(this, -encodeSize);
+         messagesAcknowledged++;
+         messagesAcknowledgedAwaitingCommit++;
+         lastAcknowledgedTime = System.currentTimeMillis();
+      }
```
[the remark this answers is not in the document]
gemmellr: I tried this out with a simple modified HelloWorld from Qpid JMS, sending + receiving some messages. The Messages Acknowledged Awaiting Commit value just increases, matching the Messages Acknowledged value, whereas I would expect it to be 0.

The Messages In Transit Size value was also incorrect/negative, being the negation of the Messages Delivered Size, rather than being 0 like the Messages In Transit Size.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -2340,10 +2318,24 @@ public Pair<SimpleString, EnumSet<RoutingType>> getAddressAndRoutingTypes(Simple
...
+   public void addProducer(ServerProducer serverProducer, String address) {
+      if (address == null) {
+         //this is ok as it will be an anonymous producer which we dont track
+         return;
+      }
       serverProducer.setSessionID(getName());
-      serverProducer.setConnectionID(getConnectionID().toString());
-      producers.put(serverProducer.getID(), serverProducer);
+      serverProducer.setConnectionID(getConnectionID() != null ? getConnectionID().toString() : null);
+      producers.put(address, serverProducer);
```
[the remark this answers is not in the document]
andytaylor: Also we will no longer have to track queue deletion and remove producers. For the legacy use case we still have the max producers to stop a DOS attack and they are removed on session close anyway

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
```diff
@@ -490,6 +489,7 @@ public void serverSend(ProtonServerReceiverContext context,
             }
          } else {
             message.setConnectionID(receiver.getSession().getConnection().getRemoteContainer());
+            message.setproducerID(receiver.getName());
```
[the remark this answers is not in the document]
andytaylor: Ok, so I can use a UUID like the other protocols. What do you mean tho by add the names to the producer details?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/CreateProducerMessage.java
```diff
@@ -0,0 +1,83 @@
...
+ * the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
```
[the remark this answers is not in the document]
andytaylor: I cut and paste this from another java cladss, just done a quick search and there are multiple files with licences like this so we probably need to fix them up at some point. Also Im surprised the licence checker didnt catch this

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/ProducerField.java
```diff
@@ -21,15 +21,19 @@
...
    VALIDATED_USER("validatedUser"),
    PROTOCOL("protocol"),
    CLIENT_ID("clientID"),
    LOCAL_ADDRESS("localAddress"),
    REMOTE_ADDRESS("remoteAddress"),
-   CREATION_TIME("creationTime");
+   CREATION_TIME("creationTime"),
+   MESSAGE_SENT("msgSent"),
+   MESSAGE_SENT_SIZE("msgSizeSent"),
+   LAST_PRODUCED_MESSAGE_ID_SENT("lastProducedMessageID");
```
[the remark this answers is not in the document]
andytaylor: Im not sure what you are referring to here?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/CreateProducerMessage.java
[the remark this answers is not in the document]
gemmellr: Fixing them so there arent files for people to cut and paste the 'wrong' thing from is why it needs updating. I have been fixing them generally, e.g see ARTEMIS 4110 / https://github.com/apache/activemq-artemis/commit/f790911c44c4fc830d577debb7163163679670c4

Its still valid in terms of the licence detail actually being there so the licence checker wont complain. Its just off and inconsistent is all.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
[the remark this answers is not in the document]
gemmellr: I was actually more thinking a long, like the broker uses for messages. Unlike those it wouldnt need persisted and should generally increment less often.

I mean, if you want the actual link name in as the 'name' in the describeProducersInfo output you still could do that, but the 'brokers conceptual id' would be the distinguisher (again, as it is for the messages).

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/ProducerField.java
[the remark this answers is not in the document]
gemmellr: It has both 'produced' and 'sent' in the field constant name: LAST_PRODUCED...._MESSAGE_ID_SENT. Seems like 'SENT' can be dropped to match the value.

LAST_PRODUCED_MESSAGE_ID("lastProducedMessageID");

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/CreateProducerMessage.java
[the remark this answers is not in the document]
andytaylor: makes sense, I'll keep an eye in future

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/view/ProducerField.java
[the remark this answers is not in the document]
andytaylor: got it

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ServerProducer.java
```diff
@@ -32,4 +32,16 @@ public interface ServerProducer {
    String getID();
 
    long getCreationTime();
+
+   void setUserID(Object userID);
```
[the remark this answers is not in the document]
andytaylor: fixed

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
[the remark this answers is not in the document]
andytaylor: Ive fixed this and added many tests for every protocol, Ive also added a check for a tx.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -2340,10 +2318,24 @@ public Pair<SimpleString, EnumSet<RoutingType>> getAddressAndRoutingTypes(Simple
...
+      if (CompositeAddress.isFullyQualified(address)) {
+         PostQueueDeletionCallback postQueueDeletionCallback = new PostQueueDeletionCallback() {
+            @Override
+            public void callback(SimpleString address, SimpleString queueName) throws Exception {
+               producers.remove(CompositeAddress.toFullyQualified(address, queueName).toString());
+               server.unregisterPostQueueDeletionCallback(this);
```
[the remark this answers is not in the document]
andytaylor: no longer valid

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -115,6 +114,7 @@ public class ServerSessionImpl implements ServerSession, FailureListener {
 
    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+   private static final int MAX_PRODUCER_METRIC_SIZE = 100;
```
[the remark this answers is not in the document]
andytaylor: fixed

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -2010,14 +1979,21 @@ public StorageManager getStorageManager() {
...
-            uuid = entry.getValue().getA().toString();
+         if (entry.getValue().getLastProducedMessageID() != null) {
+            uuid = entry.getValue().getLastProducedMessageID().toString();
          }
-         JsonObjectBuilder producerInfo = JsonLoader.createObjectBuilder().add("connectionID", this.getConnectionID().toString()).add("sessionID", this.getName()).add("destination", entry.getKey().toString()).add("lastUUIDSent", uuid, JsonValue.NULL).add("msgSent", entry.getValue().getB().longValue());
+         JsonObjectBuilder producerInfo = JsonLoader.createObjectBuilder()
+               .add(ProducerField.ID.getName(), getName())
+               .add(ProducerField.CONNECTION_ID.getName(), this.getConnectionID().toString())
+               .add(ProducerField.SESSION.getAlternativeName(), this.getName())
+               .add(ProducerField.CREATION_TIME.getName(), "" + creationTime)
```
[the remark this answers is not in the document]
andytaylor: fixed and more tests added to check the actual server side oblect is passed back correctly

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/client/TemporaryDestinationTest.java
```diff
@@ -279,7 +279,7 @@ public void testForTempQueueTargetInfosLeak() throws Exception {
          producer.send(s.createMessage());
          temporaryQueue.delete();
          for (ServerSession serverSession : server.getSessions()) {
-            assertFalse(((ServerSessionImpl)serverSession).cloneTargetAddresses().containsKey(SimpleString.toSimpleString(temporaryQueue.getQueueName())));
+            assertFalse(((ServerSessionImpl)serverSession).cloneProducers().containsKey(temporaryQueue.getQueueName()));
```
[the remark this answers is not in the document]
andytaylor: no longer valid as test removed and legacy clients tested elsewhere

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerImpl.java
```diff
@@ -219,6 +233,8 @@ private void doSend(SimpleString sendingAddress,
          // In case we received message from another protocol, we first need to convert it to core as the ClientProducer only understands core
          ICoreMessage msg = msgToSend.toCore();
 
+         msg.putStringProperty(PRODUCER_ID, new SimpleString(uuid.toString()));
```
gemmellr, line 236: The getters always return a String via a toString call, and this always creates a new SimpleString (and calls toString on the UUID) on every send...might be better to store the SimpleString rather than UUID and e.g use the toString as the getter returns and the key to remove from the sessionContext, avoids repeated recreation?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
```diff
@@ -845,6 +845,15 @@ public String getAddress() {
       }
    }
 
+   @Override
+   public String getProducerID() {
+      if (properties.containsProperty(PRODUCER_ID)) {
+         return properties.getSimpleStringProperty(PRODUCER_ID).toString();
+      }
+      //This means the client is pre 2.19.0 and doesn't support creating a server producer for metrics, this will create 1 to track all events
```
gemmellr, line 853: 2.19.0? Typo, or?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPSessionContext.java
```diff
@@ -229,8 +226,7 @@ public void addReplicaTarget(Receiver receiver) throws Exception {
          AMQPMirrorControllerTarget protonReceiver = new AMQPMirrorControllerTarget(sessionSPI, connection, this, receiver, server);
          protonReceiver.initialize();
          receivers.put(receiver, protonReceiver);
-         ServerProducer serverProducer = new ServerProducerImpl(receiver.getName(), "AMQP", receiver.getTarget().getAddress());
-         sessionSPI.addProducer(serverProducer);
+         sessionSPI.addProducer(receiver.getName(), receiver.getTarget().getAddress());
```
gemmellr, line 229: This addProducer call is fed getName(), while the addProducer call changed below uses getUniqueName(), seems inconsistent.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
```diff
@@ -60,6 +61,8 @@ public class ProtonServerReceiverContext extends ProtonAbstractReceiver {
 
    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+   private static final AtomicLong ID_GENERATOR = new AtomicLong();
```
gemmellr, line 64: I was actually suggesting that the broker allocated a differentiating long ID to _all_ producers for consistency, rather than just the AMQP ones, in similar manner as it does for every message received.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
```diff
@@ -169,7 +169,9 @@ ServerSessionImpl createServerSession(String username, String password, String v
                                                          MQTTUtil.SESSION_AUTO_CREATE_QUEUE,
                                                          server.newOperationContext(),
                                                          session.getProtocolManager().getPrefixes(),
-                                                         session.getProtocolManager().getSecurityDomain(), validatedUser);
+                                                         session.getProtocolManager().getSecurityDomain(),
+                                                         validatedUser,
+                                           false);
```
gemmellr, line 174: Tidy up alignment

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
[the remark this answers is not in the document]
andytaylor: Both AMQ and Openwire create this on the client side so tracking it at the broker wouldnt work. They both use UUIDs. This is how the producers are closed by both

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
andytaylor: Both AMQ and Openwire create this on the client side so tracking it at the broker wouldnt work. They both use UUIDs. This is how the producers are created and closed by both.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
andytaylor: fixed

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPSessionContext.java
andytaylor: fixed

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/message/impl/CoreMessage.java
andytaylor: fixed

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerImpl.java
andytaylor: fixed

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -2010,14 +1979,21 @@ public StorageManager getStorageManager() {
...
-         if (entry.getValue().getA() != null) {
-            uuid = entry.getValue().getA().toString();
+         if (entry.getValue().getLastProducedMessageID() != null) {
+            uuid = entry.getValue().getLastProducedMessageID().toString();
          }
-         JsonObjectBuilder producerInfo = JsonLoader.createObjectBuilder().add("connectionID", this.getConnectionID().toString()).add("sessionID", this.getName()).add("destination", entry.getKey().toString()).add("lastUUIDSent", uuid, JsonValue.NULL).add("msgSent", entry.getValue().getB().longValue());
+         JsonObjectBuilder producerInfo = JsonLoader.createObjectBuilder()
+               .add(ProducerField.ID.getName(), getName())
+               .add(ProducerField.CONNECTION_ID.getName(), this.getConnectionID().toString())
+               .add(ProducerField.SESSION.getAlternativeName(), this.getName())
```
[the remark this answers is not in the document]
andytaylor: fixed

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerReceiverContext.java
gemmellr: The brokers own differentiating ID wouldnt need to be the same as the clients string id though. It would just be a broker-side conceptual id. Exactly the same way broker assigns every message a long id for its own purposes, even though messages often already include client-generated MessageID strings, and individual protocols may have their own specific independent way of referencing a particular message for acks etc.

andytaylor: i don't really understand what you mean, the client allocates an ID that it uses to control the creation and deletion of the producer and then adds this to the method so the broker knows what producer is being sent to. I dont see how th ebroker can create its own id

andytaylor: ok, so I understand and I have implemented it as you suggested

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerProducerImpl.java
```diff
@@ -18,23 +18,50 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+
 public class ServerProducerImpl implements ServerProducer {
-   private final String ID;
+
+   private static final AtomicLong PRODUCER_ID_GENERATOR = new AtomicLong();
+
+   private final long ID;
+   private String name;
```
gemmellr, line 29: final like the rest from constructor?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerProducerImpl.java
```diff
@@ -18,23 +18,50 @@
...
    private final String protocol;
    private final long creationTime;
+   private volatile long messagesSent = 0;
+   private volatile long messagesSentSize = 0;
+
+   AtomicLongFieldUpdater<ServerProducerImpl> messagesSentUpdater = AtomicLongFieldUpdater.newUpdater(ServerProducerImpl.class, "messagesSent");
 
+   AtomicLongFieldUpdater<ServerProducerImpl> messagesSentSizeUpdater = AtomicLongFieldUpdater.newUpdater(ServerProducerImpl.class, "messagesSentSize");
```
gemmellr, line 37: The updaters should be static or else it defeats the benefit (memory) of using them vs AtomicLong instances.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenwireMessage.java
```diff
@@ -137,6 +137,11 @@ public String getAddress() {
       return null;
    }
 
+   @Override
+   public String getProducerID() {
+      return null;
```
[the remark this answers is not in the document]
andytaylor: This class already existed, its a wip in progress so not my place to remove it. @clebertsuconic may have input

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
```diff
@@ -392,7 +392,7 @@ public void send(final ProducerInfo producerInfo,
 
       assert clientId.toString().equals(this.connection.getState().getInfo().getClientId()) : "Session cached clientId must be the same of the connection";
       originalCoreMsg.putStringProperty(MessageUtil.CONNECTION_ID_PROPERTY_NAME, clientId);
-
+      originalCoreMsg.putStringProperty(org.apache.activemq.artemis.api.core.Message.PRODUCER_ID, producerInfo.getProducerId().toString());
```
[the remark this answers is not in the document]
andytaylor: you are correct however we already do this with the above connection id, the unused class OpenwireMessage was a wip to fix this. so I think this is the only way to go for now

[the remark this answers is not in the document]
gemmellr: Lets make things better instead of making it worse. Doing the same crap thing 'because someone else did before' only ever leads in the wrong direction.

Field on the message object, e.g the way you did it for the AMQP messages?

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenwireMessage.java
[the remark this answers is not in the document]
gemmellr: Anyone can and should delete unused crap when coming across it, rather than updating it pointlessly. That way it isnt in the way to confuse and/or lead to wasted time for others.

This has been present, seemingly unused from day one, for 6 years now. Theres clearly no actual WIP occurring, but even if there was that is what branches are for.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
```diff
@@ -101,6 +101,8 @@ public final class ClientSessionImpl implements ClientSessionInternal, FailureLi
 
    private final boolean blockOnAcknowledge;
 
+   private final boolean sendProducerID = true;
+
```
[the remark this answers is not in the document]
andytaylor: removed

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
```diff
@@ -408,6 +410,8 @@ default Message setValidatedUserID(String validatedUserID) {
 
    String getAddress();
 
+   String getProducerID();
+
```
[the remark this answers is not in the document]
andytaylor: Ive renamed to getProducerName and added a comment

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerProducerImpl.java
andytaylor: fixed

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerProducerImpl.java
andytaylor: fixed

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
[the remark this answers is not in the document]
andytaylor: This is not part of this commit, implementing OpenwireMessage is an orthogonal piece of work which is not part of  this PR

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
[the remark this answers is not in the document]
andytaylor: So I have removed the use of producerID on the message and now pass it in all protocols as part of the command

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQSession.java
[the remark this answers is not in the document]
andytaylor: Removed as part of previous change, I will talk to @clebertsuconic and send a separate commit to remove OpenWire as not needed

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenwireMessage.java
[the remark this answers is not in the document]
andytaylor: I will remove this separately

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -1513,7 +1524,137 @@ public String getConnectionRemoteAddress() {
...
+      public void addMessage(int encodeSize) {
+         messagesInTransitSizeUpdater.addAndGet(this, encodeSize);
+         messagesDeliveredSize += encodeSize;
+         messagesDelivered++;
+         lastDeliveredTime = System.currentTimeMillis();
+      }
+
+      public void addAcknowledge(int encodeSize, Transaction tx) {
+         messagesInTransitSizeUpdater.addAndGet(this, -encodeSize);
+         messagesAcknowledged++;
```
gemmellr, line 1633: I think this one is unsafe also. Looks like rollback synchronizes 'lock' but not the consumer itself as some other calls of this method from other threads do. The openwire bit using it doesnt synchronize on either.

Given the large number of places this method is used from, some on the queue thread, others not, and the mishmash of different synchronization by those calls...I think it would be simpler/safer to just use an updater for this regardless whether it is currently safe, it is certainly brittle and its likely someone change other bits in a way that breaks this area later if it isnt already (which it looks like it is).

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerProducerImpl.java
```diff
@@ -18,23 +18,50 @@
...
+   private Object lastProducedMessageID;
+
    private String sessionID;
+
    private String connectionID;
 
-   public ServerProducerImpl(String ID, String protocol, String address) {
-      this.ID = ID;
+   public ServerProducerImpl(String name, String protocol, String address) {
+      this.ID = PRODUCER_ID_GENERATOR.getAndIncrement();
```
gemmellr, line 48: Same as earlier incrementAndGet() comment.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerProducerImpl.java
```diff
@@ -18,23 +18,50 @@
...
+   private volatile long messagesSent = 0;
+   private volatile long messagesSentSize = 0;
+
+   private static final AtomicLongFieldUpdater<ServerProducerImpl> messagesSentUpdater = AtomicLongFieldUpdater.newUpdater(ServerProducerImpl.class, "messagesSent");
 
+   private static final AtomicLongFieldUpdater<ServerProducerImpl> messagesSentSizeUpdater = AtomicLongFieldUpdater.newUpdater(ServerProducerImpl.class, "messagesSentSize");
 
    private final String address;
+
+   private Object lastProducedMessageID;
```
gemmellr, line 41: Probably ought to be volatile as well as it will be the session etc thread setting it but the management bits reading it with no clear memory barrier to prod visibility update.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -1513,7 +1524,137 @@ public String getConnectionRemoteAddress() {
...
+
+      private int messagesDeliveredSize = 0;
+
+      private long lastDeliveredTime = 0;
+
+      private long lastAcknowledgedTime = 0;
+
+      private int messagesDelivered = 0;
+
+      private int messagesAcknowledged = 0;
```
gemmellr, line 1593: Probably ought to be volatile as well, even if their writes are safe under synchronization the reader will typically be a different thread that may not synchronize those.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
andytaylor: do you mean the messagesAcknowledged field?

gemmellr: Yes. The method seems like it can still be multi-threaded, so doing ++ isnt safe.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
```diff
@@ -2026,10 +2028,20 @@ private ClientProducer internalCreateProducer(final SimpleString address,
...
+      ClientProducerInternal producer = new ClientProducerImpl(this,
+                                                                      address,
+                                                                      maxRate == -1 ? null : new TokenBucketLimiterImpl(maxRate, false),
+                                                  autoCommitSends && blockOnNonDurableSend,
+                                                     autoCommitSends && blockOnDurableSend,
+                                                                      autoGroup, groupID == null ? null : new SimpleString(groupID),
+                                                                      minLargeMessageSize,
+                                                                      sessionContext,
+                                                                      producerIDs.getAndIncrement());
```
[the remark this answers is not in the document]
clebertsuconic: I also prefer incrementAndGet(). as a matter of fact I never seen the opposite (getAndIncrement)

I know it's just a preference, but if we could use incrementAndGet just to be consistent that would be my choice here.

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
```diff
@@ -174,6 +184,8 @@ public class ServerSessionPacketHandler implements ChannelHandler {
 
    private final Pool<NullResponseMessage_V2> poolNullResponseV2;
 
+   private final Map<Integer, String> producers = new HashMap<>();
```
clebertsuconic, line 187: can you instead use io.netty.util.collection.IntObjectHashMap please? It's a lot better for this case here. It's a simple replacement and it's just like a regular HashMap.

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
```diff
@@ -635,6 +648,27 @@ private void slowPacketHandler(final Packet packet) {
...
                }
+               case CREATE_PRODUCER: {
+                  CreateProducerMessage message = (CreateProducerMessage) packet;
+                  // this is used to create/destroy the producer so needs to be unique
+                  String senderName = null;
+                  if (!producers.containsKey(message.getId())) {
+                     senderName = PRODUCER_ID_PREFIX + UUIDGenerator.getInstance().generateUUID();
+                     producers.put(message.getId(), senderName);
+                  }
+                  session.addProducer(senderName, ActiveMQClient.DEFAULT_CORE_PROTOCOL, message.getAddress() != null ? message.getAddress().toString() : null);
```
gemmellr, line 659: This would still call addProducer even supposing an id was reused, but would call it with null senderName. Should probably not call it, or log or throw?

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
```diff
@@ -635,6 +648,27 @@ private void slowPacketHandler(final Packet packet) {
...
+                  break;
+               }
+               case REMOVE_PRODUCER: {
+                  RemoveProducerMessage message = (RemoveProducerMessage) packet;
+                  String remove = producers.remove(message.getId());
+                  if (remove != null) {
+                     session.removeProducer(remove);
+                  } else {
+                     ActiveMQServerLogger.LOGGER.producerDoesNotExist(session.getName());
```
gemmellr, line 668: Should probably say which producer (id) didnt exist to give some detail beyond the session. Can you tell which connection it is from the session name? If not then perhaps that also to aid tracking down problem client.

artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
```diff
@@ -55,11 +55,14 @@
 import java.lang.invoke.MethodHandles;
 
 import static org.apache.activemq.artemis.core.protocol.stomp.ActiveMQStompProtocolMessageBundle.BUNDLE;
+import static org.apache.activemq.artemis.core.protocol.stomp.StompProtocolManagerFactory.STOMP_PROTOCOL_NAME;
 
 public class StompSession implements SessionCallback {
 
    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+   private static final String senderName = UUIDGenerator.getInstance().generateUUID().toString();
```
gemmellr, line 64: Shouldnt be static or theyll all have same name.

artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
```diff
@@ -109,6 +112,7 @@ public boolean isWritable(ReadyListener callback, Object protocolContext) {
 
    void setServerSession(ServerSession session) {
       this.session = session;
+      session.addProducer(senderName, STOMP_PROTOCOL_NAME, null);
```
gemmellr, line 115: Should this pass ANONYMOUS explicitly like some others?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -1513,7 +1524,139 @@ public String getConnectionRemoteAddress() {
...
+
+      private volatile int messagesAcknowledgedAwaitingCommit = 0;
+
+      private int messagesDeliveredSize = 0;
+
+      private volatile long lastDeliveredTime = 0;
+
+      private volatile long lastAcknowledgedTime = 0;
+
+      private int messagesDelivered = 0;
```
gemmellr, line 1591: If messagesDelivered and messagesDeliveredSize are primarily being read from other threads than the updaters, and arent synchronizing any of the same thing things used when updating the values, they should be volatile also ?

andytaylor: and also FieldUpdaters?

gemmellr: If it is clear that the variable updates are thread-safe (appears so in this case, 2 callers that both use 'lock') and are unlikely to become unsafe, then you dont need a FieldUpdater.

gemmellr:
LGTM now (I ran the PR builds again on my fork, passed) but needs another rebase.

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
```diff
@@ -174,6 +184,8 @@ public class ServerSessionPacketHandler implements ChannelHandler {
 
    private final Pool<NullResponseMessage_V2> poolNullResponseV2;
 
+   private final Map<Integer, String> producers = new IntObjectHashMap<>();
```
clebertsuconic, line 187: This should have been declared as:

private final IntObjectHashMap<String> producers = new IntObjectHashMap<>();

if you keep the Map declaration you will be using boxing every time an integer is used.


not a big deal though.. I'm amending myself during merge.. just a FYI.

PR #4185, opened 2022-08-22
ARTEMIS-3943 Adjusting defalut address settings to avoid OME from paging and flow control

clebertsuconic:


artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/broker.xml
```diff
@@ -144,10 +144,14 @@ ${cluster-security.settings}${cluster.settings}${replicated.settings}${shared-st
...
-            <!-- how many messages are kept in memory from paging. The system will stop reading whenever this or max-read-page-bytes hits the max first. -->
-            <max-read-page-messages>1000</max-read-page-messages>
+            <!-- how many messages are kept in memory from paging. The system will stop reading whenever this or max-read-page-bytes hits the max first.
+                 Change it to a positive value if you want to limit the number of messages read.-->
+            <max-read-page-messages>-1</max-read-page-messages>
             <!-- how many bytes equivalent of messages are kept in memory from paging (based on memory estimate). The system will stop reading whenever this or max-read-page-messages hits the max first. -->
             <max-read-page-bytes>1M</max-read-page-bytes>
+            <!-- The system will not read any more messages if messages are pending acks.
+                 if you set this attribute to false, the broker will allow more reading as soon as messages become on "delivering" state. -->
+            <page-read-flow-control>true</page-read-flow-control>
```
gtully, line 154: I don't see the value in this boolean, if either max-read-page-bytes|messages is set then whether these messages are in the pending or delivering state is irrelevant, they still consume memory/resources on the broker. The only reason to set a max is to protect the broker from memory exhaustion, if speed is the concern then don't limit what can be read at all.

gtully: one other thought, is there a need for a global-max-read? or would global-max-bytes cover that as is the case for the send/producer side. knowing that global-max-size is respected for both senders and receivers - for paging and for limiting de-paging, may make it simpler to reason about from a sizing perspective.

for the page only mode, there needs to be per address max and max-read, but I think the global-max can shared.

clebertsuconic: The reading is always per address.. we don't control a global-max-read...

I need the attribute for the following reason:


if this attribute is false, the max-read-page-bytes still used, however as soon as the message left the queue, the message will open space for reading more messages.

With that semantic in place, as soon as the client sends credits, the messages will be delivered and that will open space for more messages to be read.


Say, the following example:



```java

     Session session = connection.createSession(true, Session.SESSION_TRANSACTED);
    MessageConsumer consumer = session.createConsumer(session.createQueue("myQueue"));
   for (int I = 0; I < 1000; I++) {
         Message message = consumer.receive();
   }
   session.commit();
```


on such example ^^^^^ if I set page-read-flow-control to true, the more messages the client request, the more messages are going to be delivered to the client.


if I set this to false, the receive() might block if you don't have enough memory in the queue... for that reason the users will have to decide on what type of consumers they have.


if you have a well behaved consumer, page-read-flow-control=false still valid and it should work.. however when you have unbehaved clients this value must be true.

gtully: I don't know any broker operator that will trust consumers, I would always want the receive to block if the broker is constrained to the point that paging in more messages would blow a limit that can lead to OOM. If I configure a max-read, then I would want it to include delivered, and behave like the flow-control bool is set.

clebertsuconic: I need to allow the previous semantics from the broker somehow, where we only look at the max-memory and don't count delivering...


What I could do is to allow page-max-read-bytes and page-max-read-messages to be -1, and on that case I use the max-memory settings with the old semantic... I will probably print a big warning if that's the case from the FileconfigurationParser

gtully: that would work and seems sensible.

I would also vote for globalMaxSize to limit total pages in some way, but that can be a future enhancement. 
There is some nice symmetry to having one globalMaxSize that can constrain the broker memory usage for messages, irrespective of whether they come from clients or from the page files.

clebertsuconic:
@gtully , @gemmellr this is now ready for merge / review

PR #4186, opened 2022-08-24
Various clean-ups

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/AdvisoryOpenWireTest.java
```diff
@@ -202,5 +205,24 @@ public void testConnectionAdvisory() throws Exception {
...
+   public void testXAdvisory() throws Exception {
+      final String QUEUE = "foo";
+      server.createQueue(new QueueConfiguration(QUEUE).setRoutingType(RoutingType.ANYCAST));
+
+      try (Connection advisoryConnection = factory.createConnection();
+           Connection connection = factory.createConnection()) {
+
+         advisoryConnection.start();
 
+//         final CountDownLatch numConnectionsCreatedViaAdvisoryNotificationsLatch = new CountDownLatch(19);
```
clebertsuconic, line 218: dead code...


that's about the only thing I'm going to ask here... just remove it

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireProtocolManager.java
```diff
@@ -451,6 +451,7 @@ public void fireAdvisory(AMQConnectionContext context,
                             Command command,
                             ConsumerId targetConsumerId,
                             String originalConnectionId) throws Exception {
+      System.out.println("FireAdvisory: " + command);
```
gemmellr, line 454: Leftover? Logger or remove.

artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/protocol/RemotingConnection.java
```diff
@@ -35,9 +35,11 @@
 /**
  * A RemotingConnection is a connection between a client and a server.
  *
- *
  * Perhaps a better name for this class now would be ProtocolConnection as this
  * represents the link with the used protocol
+ *
+ * A default noop implementation is provided. This, for example, makes test
+ * implementations much simpler.
```
gemmellr, line 42: Though they can also make managing multiple 'real implementations' as the codebase has a little more awkward, and can have an overhead vs concrete method impl. The interface becomes less streamlined also.

There only looks to be 1 test impl this actually changed (which could perhaps have just used mocking instead, maybe using even less lines than the defaultings) and then also 1 'fake' management connection. I'd personally use the abstract one as  proper base for all the real impls like you did and leave the management one closer to as-is than convert all the interface methods to default to largely return false/null just so the fake one doesnt have to.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpOutboundConnectionTest.java
```diff
@@ -132,6 +132,7 @@ public void connectionFailed(ActiveMQException exception, boolean failedOver, St
          }
 
          Wait.assertEquals(0, remote::getConnectionCount);
+         System.out.println(remotingConnection.getClass());
```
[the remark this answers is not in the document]
jbertram: Removed.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireProtocolManager.java
jbertram: Removed.

artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/protocol/AbstractRemotingConnection.java
```diff
@@ -45,6 +45,7 @@ public abstract class AbstractRemotingConnection implements RemotingConnection {
    protected final Connection transportConnection;
    protected final Executor executor;
    protected final long creationTime;
+   protected volatile boolean destroyed = false;
```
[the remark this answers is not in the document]
jbertram: Resolved.

artemis-core-client/src/main/java/org/apache/activemq/artemis/spi/core/protocol/RemotingConnection.java
jbertram: Nice call. I implemented your suggestion.

PR #4192, opened 2022-08-29
ARTEMIS-3960 - Messages received over a bridge do not populate duplic…

AntonRoskvist:
…ateCache

This also affects messages sent over a clusterConnection

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -1797,6 +1797,11 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
          message.usageDown(); // this will cause large message delete
          return DuplicateCheckResult.DuplicateNotStartedTX;
       }
+      final byte[] duplicateIDBytes = message.getDuplicateIDBytes();
+      if (duplicateIDBytes != null) {
+         final DuplicateIDCache cache = getDuplicateIDCache(context.getAddress(message));
+         cache.addToCache(duplicateIDBytes, context.getTransaction(), startedTX);
+      }
       return startedTX ? DuplicateCheckResult.NoDuplicateStartedTX : DuplicateCheckResult.NoDuplicateNotStartedTX;
    }
 
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/DuplicateDetectionTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/DuplicateDetectionTest.java
@@ -20,8 +20,10 @@ import javax.transaction.xa.XAException;
 import javax.transaction.xa.XAResource;
 import javax.transaction.xa.Xid;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.List;
 
 import org.apache.activemq.artemis.api.core.ActiveMQDuplicateIdException;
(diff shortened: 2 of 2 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1797,6 +1797,11 @@ private DuplicateCheckResult checkBridgeDuplicateID(final Message message,
          message.usageDown(); // this will cause large message delete
          return DuplicateCheckResult.DuplicateNotStartedTX;
       }
+      final byte[] duplicateIDBytes = message.getDuplicateIDBytes();
```
clebertsuconic, line 1800: this was "not done" on purpose...

Say the sender is sending messages.... you failed and reconnected... 

the next time you send a message, the operation will then succeed or fail based on the previous cache.


now if you start to fill up the cache in the absence of the producer connected to that node,  you are kind of breaking that concept.

clebertsuconic: I think there's another DuplicateCache just for the Bridge.. .let me double check the code

clebertsuconic: So, there is a duplicate cache just for the bridge... I don't think we should fill the duplicate cache for the producer since there's no producer connected on this case. it would be a waste of resources... 

It wouldn't be wrong.. just it would be doing extra work for no reason, as there's no producer.. 


If you find an use case that you can explain the need, perhaps we can add it, (even make it optional somehow)... but I think we would be better on not having it on this case.


Thanks as usual though, your contributions are always great. Let me know what you decide?

AntonRoskvist: No that makes sense and as I said I don't really have a use case in mind. But... why is the cache populated when running the bridge/cluster with duplicateDetection=false then? Is there more of a use case for it when not using the bridges own duplicate detection?

jbertram: > ...why is the cache populated when running the bridge/cluster with duplicateDetection=false then?

I talked with Anton about this issue in Slack and this was the detail that made me think this was a bug. There's no reason I can see for a discrepancy in the behavior here.

clebertsuconic: if a property to disable duplicate detection is set to false and it still filling it.. then yeah.. that's a bug.

AntonRoskvist: Again, not saying this should be merged, but I guess the confusing thing here is that we are referring to two separate things as the same:
"Regular" duplicateCache - circular record of every message that has passed an address
"Bridge/cluster duplicateCache" - Special cache used on bridged destinations to handle redistribution/redeliveries etc

From what I can tell they serve separate purposes and as such they should not interfere with each other.
If so, it's fair to assume the "useDuplicateDetection" property on a bridge should only use and consider it's own cache.

So... when does it make sense to populate the address-level cache? Is an external producer required or does it get populated by things like import/export, scaling or management operations for instance?

AntonRoskvist: @jbertram @clebertsuconic I'm closing this for now, let me know if you want me to reopen and/or change anything at some point

PR #4193, opened 2022-08-29
ARTEMIS-3961 don't log OperationContext

jbertram:


clebertsuconic:
I amended  a quick change at OperationContextImpl as your commit there. (really easy to do)... just FYI

PR #4194, opened 2022-08-30
Porting a few selector fixes/improvements from "Classic"

jbertram:


clebertsuconic:
I would instead of just referencing the AMQ, add some description on what's the issue in the commit message.

I'm always do git log... and having a descriptive commit would help a lot...


not asking anything big.. just a single statement for each one with a short sentence.

jbertram:
There _is_ a short sentence in each commit message indicating what the commit is for.

clebertsuconic:
@jbertram sorry.. I did not realize the second line on each commit.. thanks

artemis-selector/src/main/java/org/apache/activemq/artemis/selector/filter/ComparisonExpression.java
```diff
@@ -232,6 +232,31 @@ public Object evaluate(Filterable message) throws FilterException {
...
+            if (lv == null ^ rv == null) {
+               return false;
+            }
+            if (lv == rv || lv.equals(rv)) {
+               return true;
+            }
+            if (lv.getClass() == rv.getClass()) {
+               // same class, but 'equals' return false, and they are not the same object
+               // there is no point in doing 'compare'
+               // this case happens often while comparing non equals Strings
```
clebertsuconic, line 253: What about the case where both are the same class and it's actually equals?

clebertsuconic: didn't AMQ5 introduce tests for this?

jbertram: I'm just porting these changes from ActiveMQ "Classic." I can't speak to how certain cases are handled.

If there were any tests for this added they weren't part of the [original Jira](https://issues.apache.org/jira/browse/AMQ-9052).

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionBridge.java
```diff
@@ -251,29 +251,31 @@ private void setupNotificationConsumer() throws Exception {
 
          SimpleString notifQueueName = new SimpleString(qName);
 
-         SimpleString filter = new SimpleString(ManagementHelper.HDR_BINDING_TYPE + "<>" +
-                                                   BindingType.DIVERT.toInt() +
+         SimpleString filter = new SimpleString("(" + ManagementHelper.HDR_BINDING_TYPE + " <> " + BindingType.DIVERT.toInt() +
```
michaelandrepearce, line 254: Why did this existing test have to change? I would expect to not be a break change existing tests are left as was to ensure no break

jbertram: This was from #c05e1d6f26c635e3d4aea60080133f789b5f6d7c which is fixing a bug. I mentioned the semantic change in the commit message:

> Incorrect handling of unknown values in selectors.
> 
> There is a slight semantic change here due to an error in the way we
> were handling null identifiers. This may require a change in selector
> syntax to use "IS NULL" or "IS NOT NULL" when using identifiers which
> may be null in the message being selected.
> 
> This was the case for an internal filter used by the cluster connection
> bridge to select which cluster notification messages to consume.
> 
> See https://issues.apache.org/jira/browse/AMQ-5281 for more details.

PR #4205, opened 2022-09-03
ARTEMIS-3981 Correctly fail docker image preparation if download doesn't exist

ViliusS:
By default curl downloads 404 page even if download doesn't exist. This
could produce empty or incorrect docker images.

Fail flag exits the process early preventing that issue.


brusdev (approved):

michaelandrepearce:
Jira please

jbertram:
I totally missed this PR when I browsed through Jira and found ARTEMIS-3981. It seemed simple so I sent & merged #4216.

PR #4206, opened 2022-09-03
ARTEMIS-3983 Update all docker base images to use Eclipse-Temurin

ViliusS:
JRE is enough to run most Java software, including ActiveMQ Artemis.


brusdev:
I guess we would lose JDK diagnostic tools. What are the benefits to change base image from JDK to JRE?

ViliusS:
Smaller uncompressed image size on disk.

I guess the decision depends on how these docker images are seen. Are they for development or as a base images for production?

michaelandrepearce:
Ill let others discuss merits of jdk vs jre ... but this would require a jira, this isnt a small typo or doc fix.

ViliusS:
The difference is 250MB vs 450MB.

I would agree that running Artemis in development JDK tools could be needed, but I don't agree that including these tools in production (even if needed sometimes) is good idea. Generally production images should be a slim as possible to reduce attack surface, size for fast development, etc. etc.

Maybe it is better to have two Dockerfiles then, one for development and one for production, clearly marked? If you think that would be better, I can adjust PR. Also it's probably wise to forgo Debian image as it is built against deprecated openjdk image which is not updated anymore https://github.com/docker-library/openjdk/issues/505

brusdev:
Save 200MB would be great, +1 on having two Dockerfiles

ViliusS:
OK, I have added JRE option as a separate Dockerfile.

I have also removed Debian image which used now deprecated OpenJDK build, and converted CentOS image to also use Eclipse-Temurin. Previous jboss/base-jdk image was not updated for the past 2 years.

Hopefully this will satisfy everyone.

PR #4207, opened 2022-09-09
ARTEMIS-3978 - allow federation upstream configuration via properties

gtully:


michaelandrepearce:
Generally looks good to me. Re the funny spelling "policie" i don't quite fully get why its not spelt correctly as "policy.

gtully:
it is related to the logic around typed collections. in this example, all of the policies go in the policyMap, but to be able to create or reference a queuePolicy I expose the collection via getQueuePolicies that has a typed addQueuePolicie(FederationQueuePolicy...)
Then:
prop.queuePolicies.x will create a FederationQueuePolicy.name=x and add it to the map.

B/c plural or policy does not just drop the s, it is a little ugly, but largely hidden, only BeanUtil makes use of it.

At some stage it may make sense to throw out beanutil but it is not trivial and it works to date. We could also get the type information from somewhere else, but I don't see the point as that convention of stripping the s from the collection name is mostly sensible. Policy is the exception!

PR #4208, opened 2022-09-09
ARTEMIS-3980 Remove web content from distribution

brusdev:


artemis-distribution/src/main/assembly/dep.xml
```diff
@@ -82,29 +82,6 @@
...
-      <dependencySet>
-         <includes>
-            <include>org.apache.activemq:artemis-website:jar:javadoc</include>
-         </includes>
-         <outputDirectory>web/api</outputDirectory>
-         <unpack>true</unpack>
-         <useProjectArtifact>false</useProjectArtifact>
-         <directoryMode>0755</directoryMode>
-         <fileMode>0644</fileMode>
-      </dependencySet>
```
gemmellr, line 107: These seem like they might mean the module is entirely unused here, in which case it could be dropped from the assembly pom (note the 2 entries), and also the related exclude further up this assembly file (line 51 atm)

clebertsuconic: @brusdev , @gemmellr I did not realize this comment until I merged it...

Can this be addressed on a separate commit please? (If that still to be addressed).

brusdev: @gemmellr thanks for the head up, I create #4221 to fix it

PR #4211, opened 2022-09-12
ARTEMIS-3973 Solving concurrent issue between Subscription.ack and moveNext under heavy CPU usage

clebertsuconic:


clebertsuconic:
@AntonRoskvist  as you found the issue, here is the Pull Request

AntonRoskvist:
Very nice, thanks @clebertsuconic

PR #4214, opened 2022-09-13
ARTEMIS-3990 Ensure that readable buffer string read consumes the bytes

tabish121:
Ensures that when reading a string the readable buffer consumes the bytes by advancing the read index as defined in the interface API docs.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/util/NettyReadable.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/util/NettyReadable.java
@@ -145,7 +145,7 @@ public class NettyReadable implements ReadableBuffer {
 
    @Override
    public String readUTF8() {
-      return buffer.toString(Charset_UTF8);
+      return buffer.readCharSequence(buffer.readableBytes(), Charset_UTF8).toString();
    }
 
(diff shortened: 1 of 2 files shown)


clebertsuconic:
@tabish121 this was copied and paste from somewhere I don't know where it came from now.. 

By looking right now... I see that io.vertx.proton.impl.ProtonReadableBufferImpl is doing the same, and there are other places in the artemis code that are also doing it... (for instance AmqpReadableBuffer from the artemis-amqp-protocol).


How did you find this? and should we fix the other places as well?

tabish121:
The vertx code was patched awhile ago: https://github.com/vert-x3/vertx-proton/commit/fd91d805c8e1dd546662b9b2c70f73b81c0e7f49

As was Qpid JMS but I missed that there was yet another version in Artemis although it apparently isn't getting used like the NettyReadable one is as it doesn't cause tests to fail when running against my proton-j branch where I've removed all the unnecessary buffer slicing and duplicating it was doing just to read an array from a buffer.  I can look at the other usage in Artemis as try and patch that as well

clebertsuconic:
@tabish121 I just looked over AmqpReadableBuffer and that's dead code.. I'm removing it on a separate commit, don't worry about it...


Thanks Tim.

clebertsuconic:
@tabish121 I will include this on 2.25.1 next week by cherry-picking it into 2.25.x

PR #4223, opened 2022-09-19
ARTEMIS-4002 - respect existing env var ARTEMIS_LOGGING_CONF

gtully:


gemmellr:
I didnt notice you raised a PR for this when commenting on the JIRA.

I dont think we should expose the env variable with the current JBL config in use and then transition it to the Log4J 2 config later.

Per the JIRA the new-logging bits already removed this entirely, making the scripts (and service related files, which this change doesnt update to use the env variable similarly but may need to?) have no knowledge of the logging.

jbertram:
@gtully, there's conflicts on this now. Given that and @gemmellr's previous comment I think that this PR probably just needs to be closed.

clebertsuconic:
+1, this should be closed...

@gtully  ?

gtully:
we can close this. the problem still exists, how to easily override an value on the command line for the artemis java command wih an env var.  I noticed this one for logging and it fits the bill in that case. and was a quick win.
There is $DEBUG_ARGS but I think we may need JAVA_OPTS_APPEND, the $JDK_JAVA_ARGS does prepend so it cannot override.

gemmellr:
It doesnt need to override in the logging case anymore, as the scripts dont do anything in that regard.

PR #4225, opened 2022-09-19
ARTEMIS-3264 handle core-to-AMQP conversion failures more gracefully

jbertram:


gemmellr:
'Fix' the logger codes in what sense? Is it that they dont follow the 'number also conveys level' convention? I know @clebertsuconic hates that and doesnt intend doing it going forward (and various things clearly havent followed it historically) so I'm not sure this change actually makes sense at this point if thats the reason.

Regardless it doesnt seem nice to swap codes of distinct messages to reusing codes that other messages were previously using, which this seems to do a couple times, and also leaves their original numbers open for reuse later for a couple more instances. If changing them they should probably use new non-clashing numbers, and try to prevent the other ones being reused; I think I saw mention of deprecating rather than removing methods to prevent that.

It doesnt really seem like a NO-JIRA change whatever way it actually changed.

clebertsuconic:
I really hate that.... I don't think we should be doing that... 


we made a point that these codes are part of the API.. hence I don't think we should change them :)


if they are not part of the API we could just get rid of them :)

clebertsuconic:
a code is a code.. that's it :) right?

jbertram:
@gemmellr & @clebertsuconic, fixed.

PR #4228, opened 2022-09-21
ARTEMIS-4007 - expose status attribute on configuration that can hold…

gtully:
… version or error state and reflect modifications via refresh


artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+ #!/usr/bin/env sh
```
gemmellr, line 1: Should this be here?

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.java
```diff
@@ -521,6 +521,9 @@ public interface ActiveMQServerControl {
    @Attribute(desc = "The runtime size of the authorization cache")
    long getAuthorizationCacheSize();
 
+   @Attribute(desc = "The current status of the server")
+   String getStatus();
+
```
gemmellr, line 526: Having a 'status' attribute ("The current status of the server") which isnt actually a server status but just a configurable config string that defaults to empty, seems off to me. I would call this something else either more specific or more general. Seems more like the intent is to use it as a config update correlation of sorts, perhaps wind that function into the name?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/ConfigurationTest.java
```diff
@@ -122,6 +124,9 @@ public void testPropertiesConfigReload() throws Exception {
             return mytopic_31.getBindings().size() == 3;
          });
 
+         // verify round trip apply
+         Assert.assertTrue(server.getActiveMQServerControl().getStatus().contains("2"));
```
gemmellr, line 128: Is the config update application atomic or can updates to one element be seen mid-application of others? If the latter then this doesnt necessarily seem to reliably achieve what you want as the 'status' value may be visible before other changes in the update are actually applied? I guess you could apply a second status-only change to the config, after the first round is seen, to know the first update process had finished, since you can see the second update in effect.

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service
gtully: nope! good catch! let me revert

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/ConfigurationTest.java
gtully: they are atomic enough via the configuration reload. A new instance is created/modified and swapped. What I want to know in this case is that a reload has happened.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.java
gtully: the other thought, getStatusJson - but I think that is over kill, we have xxJson methods that are better than the xx, so for a new one, we can just return a string of json or not with out the method postfix

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/ConfigurationTest.java
gtully: I peeked a little more, and the reload does not use a new config object, so the status update is visible before the update. If we want better feedback in the future as to what was updated, once the changes were detected, it could be coordinated and communicated via the status. For now, with properties in a config map projection, I am happy that the status can reflect the refresh or the projection and the reaction of the reload task.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.java
gemmellr: I dont think 'status' as an attribute is that general, it implies - and actually outright states for its description here - that it returns the status of the server, which I wouldnt really say it does in this case.

I think if the value is expected to be in a certain format later it should adhere to that from the start...this is described on the JIRA and exposed as being a freeform string set by the config, so 'other things adding keys later' doenst seem to work / is effectively breaking.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -2848,6 +2849,16 @@ public Configuration setSuppressSessionNotifications(boolean suppressSessionNoti
       return this;
    }
 
+   @Override
+   public String getStatus() {
+      return status;
+   }
+
+   @Override
+   public void setStatus(String status) {
```
clebertsuconic, line 2858: can you add a javadoc here:

Some free string that will define the server's status. a Generic mechanism to be set through the configuration?

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.java
gtully: maybe I need to state that it will be json, or empty, but the attributes are up for grabs. At the moment it is only settable by config so there is no coordination, no format.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -2848,6 +2849,19 @@ public Configuration setSuppressSessionNotifications(boolean suppressSessionNoti
       return this;
    }
 
+   @Override
+   public String getStatus() {
+      return status;
+   }
+
+   // This value can reflect a desired state (revision) of config. Useful when configurationFileRefreshPeriod > 0
```
clebertsuconic, line 2857: /** */

PR #4229, opened 2022-09-22
NO-JIRA Adding Release notes about removing rest

clebertsuconic:


brusdev:
@clebertsuconic I created #4230 to add a release not for the removed web content.

PR #4230, opened 2022-09-23
NO-JIRA Adding Release notes about removing web content

brusdev:


gemmellr:
LGTM, can you rebase and make it point '3' (I know the markdown bits likely dont care)? I'd just push it in at that point.

brusdev:
fixed and rebased

PR #4234, opened 2022-09-26
ARTEMIS-4016 Bridges created by management operations are removed on …

AntonRoskvist:
…restart and config reload

@michaelandrepearce 
This is from our discussion on Slack... hope it looks good. 
If anything seems strange then it's probably incorrectly implemented, I should not have the benefit of the doubt on this one


jbertram:
@AntonRoskvist, apologies for letting this sit for so long. Nice work. Thanks for the contribution!

PR #4235, opened 2022-09-28
ARTEMIS-4020: use SLF4J logging API, with Log4J as impl for broker

gemmellr:
Switch to using SLF4J API for logging, use Log4J 2 as impl for the broker distribution and tests.

PR includes work from myself and Clebert Suconic, squashed from the new-logging branch.


gemmellr:
Just pushed a little fixup I spotted...will keep any changes as separate commits for now until its being merged, to make them easier to spot.

tabish121:
Possible future enhancement would be to replace hard coding class names in logger definitions with Methodhandle lookups

`Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());`

This tends to help when cutting and pasting loggers and forgetting to change the class name.

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/AutoCreateUtil.java
```diff
@@ -47,7 +50,7 @@ public static  void autoCreateQueue(ClientSession session, SimpleString destAddr
                      .setAddress(destAddress);
                setRequiredQueueConfigurationIfNotSet(queueConfiguration,response, RoutingType.ANYCAST, selectorString, true);
                session.createQueue(queueConfiguration);
-               ActiveMQClientLogger.LOGGER.debug("The queue " + destAddress + " was created automatically");
+               logger.debug("The queue " + destAddress + " was created automatically");
```
tabish121, line 53: Unnecessary string concatenation could be avoided here.

gemmellr: Yeah there are plenty of concats left..I converted a bunch when I happened to be changing a specific line already, or just had for others in a class, but definitely we didnt try to remove them all (Clebert even likes them :P). A whole other massive PR could be done around that and e.g removing various gating still existing.

clebertsuconic: I would not do that as part of this PR. otherwise it would never get done.

clebertsuconic:
@gemmellr, @tabish121 would the trick apply to static instances? I would prefer to keep the loggers are static final.

tabish121:
> @gemmellr, @tabish121 would the trick apply to static instances? I would prefer to keep the loggers are static final.

Of course, it'd be:
`private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());`

artemis-core-client/src/main/java/org/apache/activemq/artemis/utils/AutoCreateUtil.java
gemmellr: I'll go so far as to say I will work up such an improvement PR, but I'd like to keep it separate and land this one to avoid rebasing the changes yet more as other dev continues

clebertsuconic:
@gemmellr @tabish121 I would merge this PR right away... any further improvements can be done after its merge.

gemmellr:
Given it was only a 2 character change to a placeholder, which was clearly wrong before, I just squashed the fixup commit into the main one.

If you want to merge now I'm fine with that, would be nice not to rebase it again :)

clebertsuconic:
@gemmellr merging it...


I have one change to the log processor.. I will just do it after merged.

clebertsuconic:
YAY!!!!
Merged!!!!!

PR #4237, opened 2022-09-28
ARTEMIS-4020 Using a little trick to create the Loggers

clebertsuconic:
Trick provided by Tim Bish


tests/activemq5-unit-tests/src/test/java/org/apache/activemq/transport/failover/FailoverPriorityTest.java
```diff
@@ -37,6 +37,7 @@
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import java.lang.invoke.MethodHandles;
```
tabish121, line 40: This one doesn't seem like it needs the import, or it should actually use it in the logger declaration below.

tabish121 (commented):
I skimmed through them and found one case that looks like an unused import, or a bad logger declaration.

clebertsuconic:
@tabish121 I did not mean to change the tests under tests/activemq5-unit-tests... I meant to leave these alone...


I'm reverting my changes on that module... thanks for the catch!

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/ActionAbstract.java
```diff
@@ -185,6 +185,7 @@ the ARTEMIS_HOME variable will include back slashes (An invalid file URI charact
 
    @Override
    public Object execute(ActionContext context) throws Exception {
+      System.out.println("I'm here");
```
clebertsuconic, line 188: oops...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/sasl/scram/SCRAMServerSASLFactory.java
```diff
@@ -44,14 +44,16 @@
...
+import java.lang.invoke.MethodHandles;
 
 /**
  * abstract class that implements the SASL-SCRAM authentication scheme, concrete implementations
  * must supply the {@link SCRAM} type to use and be register via SPI
  */
 public abstract class SCRAMServerSASLFactory implements ServerSASLFactory {
 
-   private final Logger logger = LoggerFactory.getLogger(getClass());
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
gemmellr, line 55: This changes the name of the Logger, which is used for different mechanisms and passed around to related objects. The child classes should be changed to have their own static Logger and pass it via the constructor rather than creating here, to retain existing behaviour and allow distinguishing the mechanism the logging is actually for.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -814,7 +815,7 @@ public void run() {
    }
 
    private static void checkCriticalAnalyzerLogging() {
-      Logger criticalLogger = LoggerFactory.getLogger("org.apache.activemq.artemis.utils.critical");
+      Logger criticalLogger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
gemmellr, line 818: This changes the logger name used for the check to the wrong value, the one for ActiveMQServerImpl (and so also duplicates the class-level logger), rather than "org.apache.activemq.artemis.utils.critical" used before, which is what is also referenced in the log4j2.properties config.

Should be changed to a value derived from the declaration of the original logger elsewhere (maybe declared as a constant).

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileBasedNodeManager.java
```diff
@@ -32,14 +32,15 @@
...
+import java.lang.invoke.MethodHandles;
 
 import static java.nio.file.StandardOpenOption.CREATE;
 import static java.nio.file.StandardOpenOption.READ;
 import static java.nio.file.StandardOpenOption.WRITE;
 
 public abstract class FileBasedNodeManager extends NodeManager {
 
-   private static final Logger LOGGER = LoggerFactory.getLogger(FileBasedNodeManager.class);
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
gemmellr, line 43: Should be deleted instead; name changed without any other uses changing, so its unused.

artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/AbstractPrincipalLoginModule.java
```diff
@@ -31,12 +31,13 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import java.lang.invoke.MethodHandles;
 
 /**
  * Abstract login module that uses an external authenticated principal
  */
 public abstract class AbstractPrincipalLoginModule implements AuditLoginModule {
-   private final Logger logger = LoggerFactory.getLogger(getClass());
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
gemmellr, line 40: Similar to earlier comments, this changes the logger used by the subclasses to all be the same name, making it no longer possible to tell which one the logging related to. Should instead make them pass in their own static logger to retain existing logging behaviour.

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -172,17 +173,17 @@
  */
 public abstract class ActiveMQTestBase extends Assert {
 
-   private static final Logger log = LoggerFactory.getLogger(ActiveMQTestBase.class);
+   private static final Logger log = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
gemmellr, line 176: Might be worth looking at why there are 3 identical Logger declarations in this class (more below this one).

tests/artemis-test-support/src/main/java/org/apache/activemq/transport/amqp/client/AmqpProtocolTracer.java
```diff
@@ -37,7 +39,7 @@
  */
 public class AmqpProtocolTracer implements ProtocolTracer, FrameBodyHandler<AmqpFrameValidator> {
 
-   private static final Logger TRACE_FRAMES = LoggerFactory.getLogger(AmqpProtocolTracer.class.getPackage().getName() + ".FRAMES");
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
gemmellr, line 42: Noting it changes the logger name. Maybe not an issue as its only the test client.

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/logging/AssertionLoggerTest.java
```diff
@@ -64,7 +65,7 @@ public void testInfoAMQP() throws Exception {
 
    private void validateLogging(Class<?> clazz) {
       String randomLogging = RandomUtil.randomString();
-      Logger logging = LoggerFactory.getLogger(clazz);
+      Logger logging = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
gemmellr, line 68: This breaks the intent of the method, it should be restored.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/DefaultOpenSSLContextFactory.java
```diff
@@ -16,27 +16,32 @@
...
 import org.apache.activemq.artemis.spi.core.remoting.ssl.SSLContextConfig;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Default {@link OpenSSLContextFactory} for use in {@link NettyConnector} and NettyAcceptor.
  */
 public class DefaultOpenSSLContextFactory implements OpenSSLContextFactory {
 
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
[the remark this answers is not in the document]
clebertsuconic: this is changing the logging name as it should. it was a mistake to have it like it was before.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/DefaultSSLContextFactory.java
```diff
@@ -15,18 +15,23 @@
...
 import org.apache.activemq.artemis.utils.ConfigurationHelper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Simple SSLContextFactory for use in NettyConnector and NettyAcceptor.
  */
 public class DefaultSSLContextFactory implements SSLContextFactory {
 
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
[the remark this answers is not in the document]
clebertsuconic: as it should

artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/AbstractPrincipalLoginModule.java
clebertsuconic: We should make this standard... it was a mistake to do it like this.


Each class should have it's own final static logger.

tests/artemis-test-support/src/main/java/org/apache/activemq/transport/amqp/client/AmqpProtocolTracer.java
clebertsuconic: I think we should keep the change here.. it's better this way.. as you said it's on the test client.

artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/AbstractPrincipalLoginModule.java
gemmellr: Basically all this is doing by using getClass() as the key is sharing a logger name (which various places in the codebase do with the code loggers, by using the package name as the key), so that all the logging for a given LoginModule all comes out the same logger name and can be reasoned about as a unit.

If you make this class have a static logger like this youll also need to change all the log messages themselves to better reflect what module they are logging about or you simply wont be able to tell which the logging is about unless there is only one. Passing a logger for such cases is not that unusual, it is done elsewhere in the codebase.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/sasl/scram/SCRAMServerSASLFactory.java
gemmellr: Its also a mistake to change it in this way.

Not all logging is debug level, the one here is a warning during exceptional failure, and if it ever occurs it wont be possible to distinguish what mechanism it was operating for when it failed after this change, as it will always log for the abstract parent. This is effectively breaking the original/useful aimed use of the logging. Either the logging statement also needs changed to indicate what its operating for, or it should pass in the mechanism-related logger to give the same effect it was doing originally.

gemmellr:
I cant unresolve things but I dont consider the 'can no longer tell which mechanism failed' thing to be resolved. For that and the very similar 'which login module is this logging for', the change significantly reduces the utility of those log messages. They either need the log messages adjusted to actually report what child class they are logging on behalf of, or would be better to retain the existing behaviour, albeit implemented slightly differently...e.g by supplying the class or the logger from the child 'user' class so the parent can log on its behalf (e.g like Netty lets you do for its logging) rather than relying on the getClass() in the abstract parent.

clebertsuconic:
@gemmellr it was definitely a mistake to use an instanceLog on that case.... 

We should never use instanceLog for debugging...


for log.info stuff, we should always use a codified logging...

gemmellr:
The precise implementation may have been a mistake, but there are other ways to acheive the same effect, which isnt unusual. The logging before was more useful than it will be with your change, where it will become less useful becuase you cant actually distinguish which of multiple things it is on behalf of.

You mentioned not being able to find where its logging (which per above is effectively what the change itsealf does). Its relatively trivial to follow a logger name of a specific subclass to a shared-implementation in its immediate parent. There are multiple mechanisms or login module classes that some of those shared-impl log statments could be for, and previously the logs would inherently identified exactly which one it was happening for. With your change they will instead all indicate the abstract parent, making it impossible to tell what one the logging was for if there happens to be more than one in use at the time.

artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/AbstractPrincipalLoginModule.java
clebertsuconic: It's definitely no use to log.debug on an instance logger... that was a mistake... we had no defined workspace when this was done...


if this was doing log.info() with some information for the user I would agree with you.. .but log.debug() it's information for use on this case....  or someone debugging the code. 

There's no point on doing log.debug on a instance logger on this case...  we should remove them all.


We used to have others that I have historically replaced along the way.. it's just a mistake of stuff we used to do and it should be changed.

clebertsuconic:
@gemmellr it's not meant to be that way... it was a mistake.. it needs to be fixed... that's all

gemmellr:
It is almost certainly that way deliberately, and it makes much more sense that way from a 'look at logging, know what happened to what' perspective than it does with your proposed change. Unless you also change all the loggers accordingly to indicate what thing the message is being logged for, it would be far nicer to retain the current 'use logger name representing mechanism/module the logging is for' semantics. That could be better done by passing the class/logger into the parent than the preior getClass()-in-parent though.

clebertsuconic:
I don't really agree, but I will concede :) as I have spent more energy typing than reverting these changes :)

PR #4241, opened 2022-10-03
ARTEMIS-4025 - trap failure to apply properties as errors and report …

gtully:
…via the broker status in a configuration/properties/errors status field


gtully:
I have added in the work for ARTEMIS-4028, which builds on the status and provides verification of update using checksum per file. This makes dynamic updates via periodic reload of configuration deterministic with some polling of the status.

gtully:
Of note, there is also some basic 'server' status, version, up-time and the exiting state.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -538,7 +546,146 @@ public void parsePrefixedProperties(Properties properties, String prefix) throws
...
+                           sb.append(',');
+                        }
+                        sb.append(values[i]);
+                     }
+                     sb.append(']');
+                  } else {
+                     sb.append(value.toString());
+                  }
+                  sb.append(')');
+                  logger.trace(sb.toString());
```
gemmellr, line 597: Would be more readable and simpluer on a single line with placeholders. Can prepare the value related arg in advance where needed though mostly not needed, just the <NULL> and String[] cases...plus for the latter Arrays.toString() seems like it would be simpler.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -538,7 +546,146 @@ public void parsePrefixedProperties(Properties properties, String prefix) throws
...
+                        throw new InvocationTargetException(null, "Resolved nested property for:" + name + ", on: " + bean + " was null");
+                     }
+                     name = resolver.remove(name);
+                  } catch (final NoSuchMethodException e) {
+                     throw new InvocationTargetException(e, "No getter for property:" + name + ", on: " + bean);
+                  }
+               }
+               if (logger.isTraceEnabled()) {
+                  logger.trace("    Target bean = " + target);
+                  logger.trace("    Target name = " + name);
```
gemmellr, line 616: Use formatter string with placeholders, even though its gated...stops people copying the concats later (currently working on removing such concats...).

Also, really need 2 lines? Couldnt remove the gate and just have them on one?

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -599,7 +746,46 @@ public <T> T convert(Class<T> type, Object value) {
 
       BeanSupport.customise(beanUtils);
 
-      beanUtils.populate(this, beanProperties);
+      logger.debug("populate:" + this + ", " + beanProperties);
```
gemmellr, line 768: Use a formatter string with placeholders

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -599,7 +746,46 @@ public <T> T convert(Class<T> type, Object value) {
...
+      HashMap<String, String> errors = new HashMap<>();
+      // Loop through the property name/value pairs to be set
+      for (final Map.Entry<String, ? extends Object> entry : beanProperties.entrySet()) {
+         // Identify the property name and value(s) to be assigned
+         final String name = entry.getKey();
+         try {
+            // Perform the assignment for this property
+            beanUtils.setProperty(this, name, entry.getValue());
+         } catch (InvocationTargetException invocationTargetException) {
+            logger.trace("failed to populate property key: " + name, invocationTargetException);
```
gemmellr, line 779: use placeholder

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -599,7 +746,46 @@ public <T> T convert(Class<T> type, Object value) {
...
+
+         } catch (Exception oops) {
+            trackError(errors, entry, oops);
+         }
+      }
+      updateStatus(errors);
+   }
+
+   private void trackError(HashMap<String, String> errors, Map.Entry<String,?> entry, Throwable oops) {
+      logger.debug("failed to populate property entry(" + entry.toString() + "), reason: " + oops.getLocalizedMessage(), oops);
```
gemmellr, line 794: use placeholders

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -2852,13 +3038,14 @@ public Configuration setSuppressSessionNotifications(boolean suppressSessionNoti
    }
 
    @Override
-   public String getStatus() {
-      return status;
+   public synchronized String getStatus() {
```
gemmellr, line 3065: Why is status volatile if its use is synchronized?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -611,13 +614,24 @@ public void setProperties(String fileUrltoBrokerProperties) {
...
+   }
+
    private void internalStart() throws Exception {
       if (state != SERVER_STATE.STOPPED) {
          logger.debug("Server already started!");
          return;
       }
 
       configuration.parseProperties(propertiesFileUrl);
+      updateStatus("configuration", configuration.getStatus());
```
gemmellr, line 634: status names should probably have a constant somewhere to help consistency, knowing what exists, easily finding related tests..etc

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerStatus.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.activemq.artemis.utils.JsonLoader;
+
+public class ServerStatus {
+
+   private static final ServerStatus instance = new ServerStatus();
+
+   public static synchronized ServerStatus getInstanceFor(ActiveMQServerImpl activeMQServer) {
+      if (instance.server == null) {
+         instance.server = activeMQServer;
+         instance.immutableStateValues.put("version", instance.server.getVersion().getFullVersion());
```
gemmellr, line 36: Ditto

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerStatus.java
```diff
@@ -0,0 +1,81 @@
...
+
+   private synchronized void updateServerStatus() {
+      if (instance.server != null) {
+         HashMap<String, String> snapshotOfServerStatusAttributes = new HashMap<>();
+         snapshotOfServerStatusAttributes.putAll(immutableStateValues);
+         snapshotOfServerStatusAttributes.put("identity", server.getIdentity());
+         SimpleString nodeId = server.getNodeID();
+         snapshotOfServerStatusAttributes.put("nodeId", nodeId == null ? null : nodeId.toString());
+         snapshotOfServerStatusAttributes.put("uptime", server.getUptime());
+         snapshotOfServerStatusAttributes.put("state", server.getState().toString());
```
gemmellr, line 62: Ditto

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
gemmellr: Yep thats what I meant but was too lazy to type ;)

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
gtully: yep, not needed.. thanks

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
gtully: I see there is logging on the brain, but fair... will tidy, thanks.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
gtully: sorted, thanks!

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
gtully: sorted

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
gtully: maybe, not sure where to put them... they mostly should be a direct map from module/component name to status attribute, and within those objects - attribute are some subset of useful named values, using their names.
the "name" could be inferred really.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -523,22 +542,140 @@ public void parsePrefixedProperties(Properties properties, String prefix) throws
...
-   public void populateWithProperties(Map<String, Object> beanProperties) throws InvocationTargetException, IllegalAccessException {
+   public void populateWithProperties(final String propsId, Map<String, Object> beanProperties) throws InvocationTargetException, IllegalAccessException {
       CollectionAutoFillPropertiesUtil autoFillCollections = new CollectionAutoFillPropertiesUtil();
-      BeanUtilsBean beanUtils = new BeanUtilsBean(new ConvertUtilsBean(), autoFillCollections);
+      BeanUtilsBean beanUtils = new BeanUtilsBean(new ConvertUtilsBean(), autoFillCollections) {
+         // override to treat missing properties as errors, not skip as the default impl does
+         @Override
+         public void setProperty(final Object bean, String name, final Object value) throws InvocationTargetException, IllegalAccessException {
+            {
+               logger.trace("setProperty on {}, name: {}, value: {}", bean.getClass(), name, value);
```
gemmellr, line 573: As a 3-arg it should use isTraceEnabled() gate to avoid array overheads when not in use.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -599,7 +736,51 @@ public <T> T convert(Class<T> type, Object value) {
...
+
+         } catch (Exception oops) {
+            trackError(errors, entry, oops);
+         }
+      }
+      updateApplyStatus(propsId, errors);
+   }
+
+   private void trackError(HashMap<String, String> errors, Map.Entry<String,?> entry, Throwable oops) {
+      logger.debug("failed to populate property entry({}), reason: {}", entry.toString(), oops.getLocalizedMessage(), oops);
```
gemmellr, line 765: Ditto

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
gemmellr: It could just be in the impl classes if its tied to the impl+tests, or a specific class for more general ones. Anywhere really, that helps tie the related bits together rather than expecting people will properly follow/associate literals later. This class has 2 related updates for "configuration" but thats not so obvious..whereas following a shared ref to the name would make it much more so.

gtully: agree, thanks.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerStatus.java
gtully: I am going to leave the simple attributes, if we build up this some more, those names could be inferred.

gtully:
thanks for all the feedback, appreciate your time.

PR #4249, opened 2022-10-11
ARTEMIS-4035 all consumers of federated queue drop if only one consum…

jbertram:
…er drops


gtully (approved):
nice test and fix!

clebertsuconic:
@jbertram LGTM: if you could check the whole test suite before merging it though?

jbertram:
@clebertsuconic, I just saw your comment after I merged. I'm running the full test-suite now. FWIW, I ran all the federations tests manually and they all passed.

jbertram:
Full test-suite looks good. No related failures.

PR #4251, opened 2022-10-11
ARTEMIS-4039 Upgrade jackson version to 2.13.4

brusdev:


pom.xml
```diff
@@ -1029,7 +1028,7 @@
          <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>${jackson-databind.version}</version>
+            <version>${jackson.version}</version>
```
gemmellr, line 1031: These 2 versions end up differing often enough (and are likely to again soon..) that I think its worth keeping the separate version property and defaulting it to the others value when they match. I'd undo this change and update the property declarations more along the lines of:

      <jackson.version>2.13.4</jackson.version>
      <jackson-databind.version>${jackson.version}</jackson-databind.version>

PR #4254, opened 2022-10-12
ARTEMIS-4042 - read sensitive string codec env var if system property…

gtully:
… is not set


gtully:
flaky tests? The pass locally
Error:  Errors: 
Error:    SecurityTest.testSendManagementWithRole:1721 » ActiveMQNotConnected AMQ219010:...
Error:    CoreClientOverOneWaySSLTest.testOneWaySSLReloaded:536 » IllegalState AMQ229230...

jbertram:
@gtully they must be. I don't see how your change could be related to those failures. I'll merge this now.

PR #4256, opened 2022-10-13
ARTEMIS-4045 Fixing in Handler ACKRunner on Mirror

clebertsuconic:


TheCycoONE:
In my tests this PR appeared to fix the issue

clebertsuconic:
@TheCycoONE done

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
```diff
@@ -120,6 +120,11 @@ ACKMessageOperation setDelivery(Delivery delivery) {
 
       @Override
       public void run() {
+         if (!connection.isHandler()) {
+            logger.info("Moving execution to proton handler");
```
gemmellr, line 124: Should this be info? Feels more like its either expected to happen, and so e.g trace or debug, or not expected and so perhaps even warn.

clebertsuconic: @gemmellr thanks.. it was meant to be a logger.debug... I will send a commit fixing it.

PR #4258, opened 2022-10-14
ARTEMIS-4025 Fixing testsuite by adding json missing dependencies

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -415,8 +415,27 @@ public class ConfigurationImpl implements Configuration, Serializable {
...
+   // Checksum would be otherwise final
+   // however some Quorum actiations are using Serialization to make a deep copy of ConfigurationImpl and it would become null
+   // for that reason this i making the proper initialization when needed.
+   private Checksum getCheckSun() {
+      if (transientChecksum == null) {
+         transientChecksum = new Adler32();
+      }
+      return transientChecksum;
+   }
+
```
clebertsuconic, line 439: @gtully  please review this?



I just want to get the test suite back running again... so I will probably merge this pretty fast... but if you have a better fix please commit on top of this please?

PR #4259, opened 2022-10-17
ARTEMIS-4049 user add fails w/space in broker path

jbertram:
This problem was originally discovered on Windows, but the problem occurs on *nix as well.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
@@ -26,7 +26,10 @@ import javax.management.NotificationEmitter;
 import javax.management.NotificationFilter;
 import javax.management.NotificationListener;
 import javax.transaction.xa.Xid;
+import java.io.File;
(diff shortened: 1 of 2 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -4535,8 +4537,8 @@ private PropertiesLoginModuleConfigurator getPropertiesLoginModuleConfigurator()
       if (configurationUrl == null) {
          throw ActiveMQMessageBundle.BUNDLE.failedToLocateConfigURL();
       }
-      String path = configurationUrl.getPath();
-      return new PropertiesLoginModuleConfigurator(getSecurityDomain(), path.substring(0, path.lastIndexOf("/")));
+      String path = Paths.get(configurationUrl.toURI()).toString();
```
gemmellr, line 4540: The doc for Paths indicates to use Path.of(..) instead.

jbertram: Fixed.

PR #4260, opened 2022-10-18
ARTEMIS-4056 Paging Management optimizations

clebertsuconic:
- optimize startup time on paging (check-depage on startup)
- otpimize getNextPage() on complete pages
- optimize getFirstMessage() and paging. (avoid iterator usage)

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
@@ -874,7 +874,13 @@ public class QueueControlImpl extends AbstractControl implements QueueControl {
       }
    }
 
-   protected Map<String, Object>[] getFirstMessage() throws Exception {
+   /**
+    * this method returns a Map representing the first message.
+    * or null if there's no first message.
+    * @return
+    * @throws Exception
+    */
+   protected Map<String, Object> getFirstMessage() throws Exception {
       if (AuditLogger.isBaseLoggingEnabled()) {
          AuditLogger.getFirstMessage(queue);
       }
@@ -883,16 +889,12 @@ public class QueueControlImpl extends AbstractControl implements QueueControl {
       clearIO();
       try {
          List<Map<String, Object>> messages = new ArrayList<>();
-         queue.flushExecutor();
          final int attributeSizeLimit = addressSettingsRepository.getMatch(address).getManagementMessageAttributeSizeLimit();
-         try (LinkedListIterator<MessageReference> iterator = queue.browserIterator()) {
-            // returns just the first, as it's the first only
-            if (iterator.hasNext()) {
-               MessageReference ref = iterator.next();
-               Message message = ref.getMessage();
-               messages.add(message.toMap(attributeSizeLimit));
-            }
-            return messages.toArray(new Map[1]);
+         MessageReference firstMessage = queue.peekFirstMessage();
+         if (firstMessage != null) {
+            return firstMessage.getMessage().toMap(attributeSizeLimit);
+         } else {
+            return null;
          }
       } finally {
          blockOnIO();
@@ -905,7 +907,8 @@ public class QueueControlImpl extends AbstractControl implements QueueControl {
       if (AuditLogger.isBaseLoggingEnabled()) {
          AuditLogger.getFirstMessageAsJSON(queue);
       }
-      return toJSON(getFirstMessage());
+      Map<String, Object> message = getFirstMessage();
+      return toJSON(message == null ? new Map[0] : new Map[]{message});
    }
 
    @Override
@@ -914,15 +917,16 @@ public class QueueControlImpl extends AbstractControl implements QueueControl {
          AuditLogger.getFirstMessageTimestamp(queue);
       }
 
-      Map<String, Object>[] _message = getFirstMessage();
-      if (_message == null || _message.length == 0 || _message[0] == null) {
+      Map<String, Object> message = getFirstMessage();
+      if (message == null) {
          return null;
+      } else {
+         if (!message.containsKey("timestamp")) {
+            return null;
+         } else {
+            return (Long) message.get("timestamp");
+         }
       }
-      Map<String, Object> message = _message[0];
-      if (!message.containsKey("timestamp")) {
-         return null;
-      }
-      return (Long) message.get("timestamp");
    }
 
    @Override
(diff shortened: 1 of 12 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
```diff
@@ -883,17 +883,12 @@ protected Map<String, Object>[] getFirstMessage() throws Exception {
...
-               MessageReference ref = iterator.next();
-               Message message = ref.getMessage();
-               messages.add(message.toMap(attributeSizeLimit));
-            }
-            return messages.toArray(new Map[1]);
+         MessageReference firstMessage = queue.peekFirstMessage();
+         if (firstMessage != null) {
+            messages.add(firstMessage.getMessage().toMap(attributeSizeLimit));
          }
+         return messages.toArray(new Map[1]);
```
gemmellr, line 891: If the idea is to optimise this overall getFirstMessage() method, then there still looks to be room for improvement here. It creates an ArrayList of Map, then afterwards figures out whether it will put 0 or 1 things in it, causing creation of a new array of capacity 10 if there is, and then converts the resulting ArrayList of 0/1 items into a new created array of size 1.

Seems like it could just omit the ArrayList and create the final array, then if the map value exists add it, then return it. Or if allowed by the API and its uses, even just only create the array once its known there is something to put in it, and return null otherwise.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -832,13 +832,14 @@ public boolean isPersistent() {
    @Override
    public void processReload() throws Exception {
       if (recoveredACK != null) {
-         logger.trace("********** processing reload!!!!!!!");
+         logger.debug("processing reload queue name={} with id={}", queue != null ? this.queue.getName() : "N/A", this.cursorId);
```
gemmellr, line 835: The cursorId value is a long, so you should now gate the logger to avoid boxing. The "this" qualifiers dont seem needed.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -1278,11 +1286,16 @@ private class CursorIterator implements PageIterator {
       private LinkedListIterator<PagedMessage> currentPageIterator;
 
       private void initPage(long page) {
+         logger.debug("initPage {}", page);
```
gemmellr, line 1289: should be gated

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -1278,11 +1286,16 @@ private class CursorIterator implements PageIterator {
...
          try {
             if (currentPage != null) {
+               if (logger.isTraceEnabled()) {
+                  logger.trace("usage down {} on subscription {}", currentPage.getPageId(), cursorId);
+               }
                currentPage.usageDown();
             }
             if (currentPageIterator != null) {
+               logger.trace("closing pageIterator on {}", cursorId);
```
gemmellr, line 1298: should be gated

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -1460,23 +1473,47 @@ private PagedReference moveNext() {
 
       private PagedReference internalGetNext() {
          for (;;) {
+            if (logger.isDebugEnabled() && currentPageIterator == null) {
```
gemmellr, line 1476: Is the null value unlikely to happen? Seems like it since it will presumably NPE below if so. (EDIT: so do we even need the logging?)

If its the rarer event, I'd put its check first. I'd probably just make it 2 ifs.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -1460,23 +1473,47 @@ private PagedReference moveNext() {
...
 
+      private long getNextPage() {
+         long page = currentPage.getPageId() + 1;
+
+         while (page <= pageStore.getCurrentWritingPage()) {
+            PageCursorInfo info = locatePageInfo(page);
+            if (info == null || info.getCompleteInfo() == null) {
+               return page;
+            }
+            logger.debug("Subscription {} named {}  moving faster from page {} to next", cursorId, queue.getName(), page);
```
gemmellr, line 1511: Triple arg, and long values...add gate

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageSubscriptionImpl.java
```diff
@@ -1460,23 +1473,47 @@ private PagedReference moveNext() {
 
       private PagedReference internalGetNext() {
          for (;;) {
+            if (logger.isDebugEnabled() && currentPageIterator == null) {
+               logger.debug("Page Cursor {} has no pageIterator", cursorId, new Exception("trace"));
```
gemmellr, line 1477: 'tracing exception stacktrace' at debug seems odd.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3164,6 +3174,7 @@ private void checkDepage() {
          return;
       }
       if (pageIterator != null && pageSubscription.isPaging()) {
+         logger.debug("CheckDepage on queue name {}, id={}", this.name, this.id);
```
gemmellr, line 3177: The "this." qualifiers seem unnecessary.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
clebertsuconic: it makes sense... after reading what it does.. 

I was just getting rid of the iterator() which caused an outage on an user I was helping out.

But I will improve that part as well.. thanks

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
clebertsuconic: @gemmellr I actually like the this quite a lot.. makes it explicitly and where it's coming from when you read it.

I guess it will be personal taste here?

PR #4267, opened 2022-10-21
ARTEMIS-4068: speed up release preparation process

gemmellr:
https://issues.apache.org/jira/browse/ARTEMIS-4068


gemmellr:
Discussed with Clebert previously, so just merged.

PR #4268, opened 2022-10-24
NO-JIRA - Fix typo in OpenWireMessageConverter

AntonRoskvist:
I think this is intended to only happen once


jbertram:
That's definitely a typo. Will merge ASAP. Thanks for the PR!

PR #4270, opened 2022-10-25
ARTEMIS-4071 Fix erroneus audit log messages due to console logouts

brusdev:
The HTTP redirection messages (300 – 399) must be ignored they don't mean an authentication failure.

### DIFF
--- a/artemis-web/src/main/java/org/apache/activemq/artemis/component/AuthenticationFilter.java
+++ b/artemis-web/src/main/java/org/apache/activemq/artemis/component/AuthenticationFilter.java
@@ -43,14 +43,18 @@ public class AuthenticationFilter implements Filter {
       filterChain.doFilter(servletRequest, servletResponse);
       if (AuditLogger.isAnyLoggingEnabled()) {
          int status = ((Response) servletResponse).getStatus();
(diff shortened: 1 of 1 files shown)


artemis-web/src/main/java/org/apache/activemq/artemis/component/AuthenticationFilter.java
```diff
@@ -43,14 +43,21 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
       filterChain.doFilter(servletRequest, servletResponse);
       if (AuditLogger.isAnyLoggingEnabled()) {
          int status = ((Response) servletResponse).getStatus();
-         //status 200 means that the user has been authenticated, anything else must be a failure
-         if (status == 200) {
+         if (status < 300) {
```
TheCycoONE, line 46: Does 100 continue need to be handled as well - would that ever reach here?

brusdev: Status ranges updated

clebertsuconic:
@brusdev fine... merge when you feel comfortable then . LGTM

PR #4273, opened 2022-10-27
ARTEMIS-4076: add JDK 19 run to CI jobs

gemmellr:
Fixes a couple of issues causing test failures or hangs, and adds a Java 19 job to the CI runs on push/PR.

### DIFF
--- a/artemis-unit-test-support/src/main/java/org/apache/activemq/artemis/utils/ThreadLeakCheckRule.java
+++ b/artemis-unit-test-support/src/main/java/org/apache/activemq/artemis/utils/ThreadLeakCheckRule.java
@@ -235,7 +235,7 @@ public class ThreadLeakCheckRule extends TestWatcher {
    private boolean isExpectedThread(Thread thread) {
       final String threadName = thread.getName();
       final ThreadGroup group = thread.getThreadGroup();
-      final boolean isSystemThread = group != null && "system".equals(group.getName());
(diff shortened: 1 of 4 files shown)


artemis-unit-test-support/src/main/java/org/apache/activemq/artemis/utils/ThreadLeakCheckRule.java
```diff
@@ -244,7 +244,7 @@ private boolean isExpectedThread(Thread thread) {
          return true;
       } else if (threadName.contains("Attach Listener")) {
          return true;
-      } else if ((javaVendor.contains("IBM") || isSystemThread) && threadName.equals("process reaper")) {
+      } else if ((javaVendor.contains("IBM") || isSystemThread) && (threadName.equals("process reaper") || threadName.startsWith("process reaper (pid"))) {
```
clebertsuconic, line 247: did you mean the (pid) in the string?

clebertsuconic: yes... I think you just do startsWith("process reaper");


As the change on the idk is 

process reaper (pid " + pid + ")");

gemmellr: It isnt always in the name, it comes and goes.

gemmellr: But yes we could use 'starts with' for both if you prefer.

PR #4278, opened 2022-11-04
ARTEMIS-4078 Fix divert configuration reloading

jsmucr:
Reloading has been fixed for divert:
* filter
* address
* exclusive

Source address and exclusivity changes require divert redeployment.

Please review.


jbertram:
@jsmucr, nice work. Thanks for the PR!

PR #571, opened 2016-06-08
[ARTEMIS-561] JMS Bridge should display statistics about messages tha…

bayern39:
https://issues.apache.org/jira/browse/ARTEMIS-561

i have added two metrics about jms-bridge.
the attribute of  'message_count' record the number of messages successfully committed.
the attribute of 'aborted_message_count' record the number of messages aborted/rollback.

PR #4281, opened 2022-11-07
ARTEMIS-4084 Rollbacking massive amounts of messages might crash broker

AntonRoskvist:
I have observed this happening with far fewer messages and multiple consumers attached to the same queue, but have been unable to write a consistent test for it. It also happens with AUTO_ACKNOWLEDGE instead of transacted sessions but this was easier to implement in a test.
I'm not really happy with the runtime of the test so if anyone has suggestions on how to make it run faster just let me know.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
@@ -399,26 +399,19 @@ public class ServerSessionImpl implements ServerSession, FailureListener {
             return;
 
          if (failed) {
-
             Transaction txToRollback = tx;
-            if (txToRollback != null) {
-               if (txToRollback.tryRollback() && txToRollback.getXid() != null) {
-                  resourceManager.removeTransaction(txToRollback.getXid(), remotingConnection);
-               }
+            if (txToRollback != null && txToRollback.getXid() != null && txToRollback.tryRollback()) {
+               resourceManager.removeTransaction(txToRollback.getXid(), remotingConnection);
             }
 
             txToRollback = pendingTX;
-
-            if (txToRollback != null) {
-               if (txToRollback.tryRollback() && txToRollback.getXid() != null) {
-                  resourceManager.removeTransaction(txToRollback.getXid(), remotingConnection);
-               }
+            if (txToRollback != null && txToRollback.getXid() != null && txToRollback.tryRollback()) {
+               resourceManager.removeTransaction(txToRollback.getXid(), remotingConnection);
             }
 
          } else {
             if (tx != null && tx.getXid() == null) {
                // We only rollback local txs on close, not XA tx branches
-
                try {
                   rollback(failed, false);
                } catch (Exception e) {
@@ -426,6 +419,7 @@ public class ServerSessionImpl implements ServerSession, FailureListener {
                }
             }
          }
+
          closed = true;
       }
 
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/clientcrash/ClientCrashMassiveRollback.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/clientcrash/ClientCrashMassiveRollback.java
@@ -1,0 +1,122 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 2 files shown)


jbertram:
I don't think this fix is valid as it's changing the semantics of the transaction rollback when a session fails. Previously _any_ pending transaction was rolled back whether or not it has a xid. Now you're only rolling back transactions with a xid which means in your test the transactions aren't being rolled back at all. It's perfectly valid for a transaction not to have a xid and such transactions should be rolled back when a session fails.

AntonRoskvist:
They are getting rolled back, that's part of the close() call issued just a little bit later. (tx.rollback() on line #571 in my change)
The test also makes an assertion on messageCount right at the end.

The previous (much older) behavior was that only XA transactions was handled here as well:
https://github.com/apache/activemq-artemis/commit/bafefdc8ece23801ac8fb7f5879faa9c3107ae5f

clebertsuconic:
@AntonRoskvist I am confused.. where the rollback is happening for non XA transactions?

line 571 is a a call to server.callBrokerConsumerPlugins(plugin -> plugin.afterCreateConsumer(consumer)); 


I am also having the impression that a rollback is not happening. I'm not doubting you.. I just can't see.. because perhaps there's a better change for this?

clebertsuconic:
@AntonRoskvist I just saw your test. You're not really fixing it.. you're just moving the "issue" in a way it would only fail in the XA scenario.

clebertsuconic:
So, if you did the same test using a XA TX, you wouldn't have fixed it, right?

AntonRoskvist:
@clebertsuconic  @jbertram  Sorry, I see that I was very unclear. The rollback happens here: org.apache.activemq.artemis.core.server.impl.ServerConsumerImpl#571
It's part of the consumer.close(failed)-call made from the "ServerSessionImpl" just after the code change I made here. Looking at the PR I mentioned above I can clearly see I was wrong though, a rollback was made previously as well.

I arrived at this from the stacktrace the CriticalAnalyzer logs when this issue occurs:
https://paste.mozilla.org/HiXUh8yg/raw

Then I set up this test and set things up so I printed a thread dump in:
org.apache.activemq.artemis.core.server.impl.QueueImpl.addSorted(QueueImpl.java:1169)

There I could see that two calls where made during client failure, the later of which seems to lock something up. Thread dumps seen here:
https://paste.mozilla.org/9nipgJbL/raw

Notice the divergence at: org.apache.activemq.artemis.core.server.impl.ServerSessionImpl.doClose
Both thread dumps are from the same thread, the first call inserted maybe 90% of all references and the second the rest. The second call is the one that locks.

This change gets rid of the first call (txToRollback.tryRollback()) but keeps the one in org.apache.activemq.artemis.core.server.impl.ServerConsumerImpl (tx.rollback())

I am very open to the possibility of me being wrong here though

clebertsuconic:
Well, I am not sure the use case is valid though... a huge Transaction like this is certainly bound to a failure.

I'm looking if there is any evidence of a dead lock here though before I say anything.

clebertsuconic:
That second rollback is for the case of a non transactional session. We still use a transaction to return the messages but we would be doing something else underneath it.

clebertsuconic:
@AntonRoskvist but you had to disable consumer flow control to have this happening?

clebertsuconic:
@AntonRoskvist https://github.com/apache/activemq-artemis/pull/4283


I'm still improving the test (to make sure it fails without the fix, as the test is always passing now);

AntonRoskvist:
@clebertsuconic Sorry, I missed your link there, that's looking very good I think! Should I just close this?

clebertsuconic:
Let's keep it open for now ?

I have been coming back and forth the two branches for testing. And it makes me easier if it's open.  

I can close it when I'm done.

clebertsuconic:
@AntonRoskvist the issue is not on the transactions itself. Your rollback ordering here is uncovering some bug on LinkedListImpl::addSorted. For some reason this will only happen if you used the Core API. Some sort of duplicate adds is probably causing it, because if I simply used a regular ConnectionFActory it worked fine.



if you go to LInkedListImpl::addSorted, there is something in a loop after the rollback.


The CriticalAnalyzer actually killed your broker correctly on this case... as the logic was lost.


I will need to work on it tomorrow... but if you could also take a look on that?

clebertsuconic:
hmmm... I think it's just a deadlock.. the rollback is holding a lock on queue, while the rollback on the large tx is getting a new lock.

I mean just because I thought there was a bug in the LinkedList. Leave it with me for another day please?

clebertsuconic:
not a dead lock... it's just the way this rollback happens it is taking a long time to process. relaxing the locks is the way to go here.

I will see if I can do some optimization.


will be fixed tomorrow :)  


it will be after 2.27.0, but we should have a 2.28.0 soon..

AntonRoskvist:
Very interesting!

I think the issues comes down to the transaction getting handled as two consecutive rollbacks for some reason... logging out refs.size() in QueueImpl shows that the first run rolls back some 90% of all messages and this is really fast... then the rest of the messages gets rolled back and I believe these arrive out of order such that the list has to be scanned for each message to find it's proper place...

And yes, I believe the CriticalAnalyzer is absolutely right to stop the broker. In fact I tried to set it to LOG and the broker was never able to recover properly... or at least not for some 30 minutes which is how much time I gave it before manually restarting.

clebertsuconic:
@AntonRoskvist it is about the two consecutive rollbacks. One will add a bunch of data. The second will call the addSorted for a huge list, scanning all the lists for every rollback item.

If you actually rolled back the consumer's pending first, the adds would be in order as the addSorted checks for first and last.


I'm adding an optimization to keep the last added position and also check against the last added, and I will as well have the adds done outside of the locking.

clebertsuconic:
@AntonRoskvist I am closing this PR as this is being replaced by #4283

clebertsuconic:
I got all the way to the bottom of the issue. #4283 will definitely fix the issue here

AntonRoskvist:
Excellent, thanks @clebertsuconic !

PR #4283, opened 2022-11-08
ARTEMIS-4084 Fixing addSorted with large transactions

clebertsuconic:
when cancelling a large number of messages, the addSorted could be holding a lock for too long causing the server to crash under CriticalAnalyzer

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/LinkedListImpl.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/LinkedListImpl.java
@@ -16,11 +16,15 @@
  */
 package org.apache.activemq.artemis.utils.collections;
 
+import java.lang.invoke.MethodHandles;
 import java.lang.reflect.Array;
 import java.util.Comparator;
 import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.function.Consumer;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A linked list implementation which allows multiple iterators to exist at the same time on the queue, and which see any
@@ -29,6 +33,8 @@ import java.util.function.Consumer;
  * This class is not thread safe.
  */
 public class LinkedListImpl<E> implements LinkedList<E> {
+
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
    private static final int INITIAL_ITERATOR_ARRAY_SIZE = 10;
 
@@ -41,6 +47,8 @@ public class LinkedListImpl<E> implements LinkedList<E> {
    private int numIters;
    private int nextIndex;
    private NodeStore<E> nodeStore;
+
+   private volatile Node<E> lastAdd;
 
    public LinkedListImpl() {
       this(null, null);
@@ -155,12 +163,18 @@ public class LinkedListImpl<E> implements LinkedList<E> {
    }
 
    private void itemAdded(Node<E> node, E item) {
+      assert node.val() == item;
+      lastAdd = node;
+      if (logger.isTraceEnabled()) {
+         logger.trace("Setting lastAdd as {}, e={}", lastAdd, lastAdd.val());
+      }
       if (nodeStore != null) {
          putID(item, node);
       }
    }
 
    private void itemRemoved(Node<E> node) {
+      lastAdd = null;
       if (nodeStore != null) {
          nodeStore.removeNode(node.val(), node);
       }
@@ -186,13 +200,22 @@ public class LinkedListImpl<E> implements LinkedList<E> {
    }
 
    public void addSorted(E e) {
+      final Node<E> localLastAdd = lastAdd;
+
+      logger.trace("**** addSorted element {}", e);
+
       if (comparator == null) {
          throw new NullPointerException("comparator=null");
       }
+
       if (size == 0) {
+         logger.trace("adding head as there are no elements {}", e);
          addHead(e);
       } else {
          if (comparator.compare(head.next.val(), e) < 0) {
+            if (logger.isTraceEnabled()) {
+               logger.trace("addHead as e={} and head={}", e, head.next.val());
+            }
             addHead(e);
             return;
          }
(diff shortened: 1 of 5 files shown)


tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/client/ClientCrashMassiveRollbackTest.java
```diff
@@ -0,0 +1,121 @@
...
+      }
+      producer.close();
+
+      while (queueControl.getDeliveringCount() < messageCount) {
+         Thread.sleep(1000);
+      }
+
+      //Ran the consumer in a thread to be able to kill it "uncleanly"
+      //Thread.stop() is deprecated for the reason I am exploiting here
+      thread.stop();
```
gemmellr, line 110: It has been deprecated for 24 years. It is now deprecated _for-removal_ so they will likely remove it at some point soon. EDIT: Actually, they have _already_ deliberately broken it for Java 20, so I expect it might not be there in even 21 (the next LTS): https://bugs.openjdk.org/browse/JDK-8289610. We shouldn't be using this.

tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/client/ClientCrashMassiveRollbackTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+
+
+      Assert.assertEquals(messageCount, queueControl.getMessageCount());
+      Assert.assertEquals(ActiveMQServer.SERVER_STATE.STARTED, server.getState());
+
+      Thread.sleep(5000);
```
gemmellr, line 117: Could it possibly assert something useful after wasting the 5 seconds?

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/LinkedListImpl.java
```diff
@@ -203,20 +223,31 @@ public void addSorted(E e) {
          // This would be an optimization for our usage.
          // avoiding scanning the entire List just to add at the end, so we compare the end first.
          if (comparator.compare(tail.val(), e) >= 0) {
+            logger.trace("addTail as compare is after tail {}", e);
```
gemmellr, line 226: Perhaps log the tail value as well as e? We can already know what e is when reading the logs from the earlier logging by the thread, but we wont necessarily know what tail was when the comparison was made.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/LinkedListImpl.java
```diff
@@ -203,20 +223,31 @@ public void addSorted(E e) {
...
-         while (fetching.next != null) {
-            int compareNext = comparator.compare(fetching.next.val(), e);
-            if (compareNext <= 0) {
-               addAfter(fetching, e);
-               return;
+         if (lastAdd != null) { // as an optimization we check against the last add rather than always scan.
+            if (lastAdd.prev != null && lastAdd.prev.val() != null) {
+               if (comparator.compare(lastAdd.prev.val(), e) > 0 && comparator.compare(lastAdd.val(), e) < 0) {
+                  logger.trace("addABefore last element {}", e);
+                  addAfter(lastAdd.prev, e);
```
gemmellr, line 235: addABefore has a typo, but more importantly the code then executes addAfter which makes it look wrong since the other log messages in this area refer to literal method names, though it isnt wrong since it passes the previous node. Also "last element" sounds a bit like it will be the tail, though clearly isnt here.

Something like "adding before most recently added element" might be clearer?

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/LinkedListImpl.java
```diff
@@ -203,20 +223,31 @@ public void addSorted(E e) {
...
+                  logger.trace("addAfter last element {}", e);
+                  addAfter(lastAdd, e);
+                  return;
+               }
             }
-            fetching = fetching.next;
          }
 
+         if (addSortedScan(e))
+            return;
```
gemmellr, line 249: Can we spend that 1 line on braces? :) More consistent and  less likely to be messed up later.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/LinkedListImpl.java
```diff
@@ -203,20 +223,31 @@ public void addSorted(E e) {
...
             return;
          }
 
-         Node<E> fetching = head.next;
-         while (fetching.next != null) {
-            int compareNext = comparator.compare(fetching.next.val(), e);
-            if (compareNext <= 0) {
-               addAfter(fetching, e);
-               return;
+         if (lastAdd != null) { // as an optimization we check against the last add rather than always scan.
```
gemmellr, line 231: Should probably copy the value rather than accessing it repeatedly given it is volatile, and also makes it less brittle in terms of its not-thread-safe'ty (though obviously the whole list isnt).

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1165,6 +1165,12 @@ public void addHead(final List<MessageReference> refs, boolean scheduling) {
    /* Called when a message is cancelled back into the queue */
    @Override
    public void addSorted(final List<MessageReference> refs, boolean scheduling) {
+
+      if (refs.size() > MAX_DELIVERIES_IN_LOOP) {
+         logger.debug("Switching addSortedCall to addSortedHugeLoad {}", name);
```
gemmellr, line 1170: space before 'call', the addSortedHugeLoad name doesnt match method called

tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/client/ClientCrashMassiveRollbackTest.java
clebertsuconic: this was a mistake... I didn't mean to keep it

tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/client/ClientCrashMassiveRollbackTest.java
clebertsuconic: @gemmellr I will use interrupt... 

I was just keeping the original test done by Anton here. I tried to use an external VM where I could stop but I could not reproduce the issue.


I just used interrupt and it reproduced it... 


thanks

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/LinkedListImpl.java
clebertsuconic: lol

I didn't think about it.. just right click, refactor as a new method.. automatic change :)

PR #4285, opened 2022-11-10
ARTEMIS-4083 ClientLargeMessage Streaming not closing inputStream if compressed

clebertsuconic:


gemmellr:
This change doesnt seem to do anything except add a newline that doesnt seem like it should change behaviour (plus tests)?

gemmellr:
actually, I see there is a third commit already on main...not clear why it isnt on the PR

clebertsuconic:
@gemmellr I had some git-fu errors..I meant to send another NO-JIRA commit on main, and I ended up sending a non tested, not ready commit on main.


At that point I just sent the diff to my latest version after tested and my own review to fix my git-fu mistake. Sorry about that  :)

PR #4288, opened 2022-11-15
ARTEMIS-4091 - Make scaleDown target more deterministic

AntonRoskvist:
This change makes the scaleDown process prefer the selected scaleDown connector as the target for scaleDown, meaning you can more easily predict which node scaleDown will happen to, especially when issued through management

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
@@ -41,6 +41,7 @@ import org.apache.activemq.artemis.spi.core.protocol.RemotingConnection;
 import org.apache.activemq.artemis.spi.core.remoting.BaseConnectionLifeCycleListener;
 import org.apache.activemq.artemis.spi.core.remoting.Connection;
 import org.apache.activemq.artemis.spi.core.remoting.ReadyListener;
+import org.apache.activemq.artemis.utils.ConfigurationHelper;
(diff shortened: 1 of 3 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
```diff
@@ -436,7 +437,9 @@ public boolean isSameTarget(TransportConfiguration... configs) {
             continue;
          }
          if (NettyConnectorFactory.class.getName().equals(cfg.getFactoryClassName())) {
-            if (configuration.get(TransportConstants.PORT_PROP_NAME).equals(cfg.getParams().get(TransportConstants.PORT_PROP_NAME))) {
+            int port1 = ConfigurationHelper.getIntProperty(TransportConstants.PORT_PROP_NAME, TransportConstants.DEFAULT_PORT, configuration);
+            int port2 = ConfigurationHelper.getIntProperty(TransportConstants.PORT_PROP_NAME, TransportConstants.DEFAULT_PORT, cfg.getParams());
+            if (port1 == port2) {
```
jbertram, line 442: Is this change strictly necessary?

AntonRoskvist: Probably not... it comes from the test actually. The way the cluster nodes are set up in ClusterTestBase means that PORT_PROP_NAME will be an int, while a connector read from the config file or one added from management operation would be expressed as a String. I think it was in that order... but regardless, I added this because I though it might be something worth covering for.

PR #4292, opened 2022-11-18
ARTEMIS-4093: Expose properties and support resource adapter in J2EE environments

hawkdshrh:
Fixes to expose properties in environments not recognizing is* accessors.

### DIFF
--- a/artemis-ra/src/main/java/org/apache/activemq/artemis/ra/ActiveMQResourceAdapter.java
+++ b/artemis-ra/src/main/java/org/apache/activemq/artemis/ra/ActiveMQResourceAdapter.java
@@ -821,6 +821,10 @@ public class ActiveMQResourceAdapter implements ResourceAdapter, Serializable {
       raProperties.setUseTopologyForLoadBalancing(useTopologyForLoadBalancing);
    }
 
+   public Boolean getUseTopologyForLoadBalancing() {
+      return raProperties.isUseTopologyForLoadBalancing();
+   }
+
    public Boolean isUseTopologyForLoadBalancing() {
       return raProperties.isUseTopologyForLoadBalancing();
    }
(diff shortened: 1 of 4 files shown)


clebertsuconic:
you should rebase this, and fix the title of your commit. you amend the wrong commit..

clebertsuconic:
I converted your commit to draft until you fix this.

artemis-ra/src/main/java/org/apache/activemq/artemis/ra/ActiveMQResourceAdapter.java
```diff
@@ -821,6 +821,10 @@ public void setUseTopologyForLoadBalancing(Boolean useTopologyForLoadBalancing)
       raProperties.setUseTopologyForLoadBalancing(useTopologyForLoadBalancing);
    }
 
+   public Boolean getUseTopologyForLoadBalancing() {
+      return raProperties.isUseTopologyForLoadBalancing();
```
clebertsuconic, line 825: why?


isUseTopology is the standard for Java Beans.

Why do you need the get method?

artemis-ra/src/main/java/org/apache/activemq/artemis/ra/inflow/ActiveMQActivationSpec.java
```diff
@@ -619,6 +649,10 @@ public void setUseLocalTx(final Boolean localTx) {
       this.localTx = localTx;
    }
 
+   public Boolean getRebalanceConnections() {
```
clebertsuconic, line 652: why? same question as before.

artemis-ra/src/main/java/org/apache/activemq/artemis/ra/ActiveMQResourceAdapter.java
hawkdshrh: Unfortunately, this is not recognized in some environments.  Websphere Application Server 8.x is one of these.

artemis-ra/src/main/java/org/apache/activemq/artemis/ra/inflow/ActiveMQActivationSpec.java
hawkdshrh: Unfortunately, these methods are not recognized in some environments, causing failure to expose the properties (Websphere Application Server is one such).

PR #4295, opened 2022-11-23
ARTEMIS-4098 AMQP messages missing correlation ID in console

jbertram:


artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessageTest.java
```diff
@@ -2285,6 +2287,7 @@ public void testToCompositeDataPropertiesSection() throws Exception {
       assertTrue(properties.contains(PROPERTY_MAP_PROPERTIES_PREFIX + "absoluteExpiryTime" + "=" + testExpiryTime));
       assertTrue(properties.contains(PROPERTY_MAP_PROPERTIES_PREFIX + "to" + "=" + TEST_TO_ADDRESS));
       assertTrue(properties.contains(PROPERTY_MAP_PROPERTIES_PREFIX + "subject" + "=" + testSubject));
+      assertTrue(properties.contains(PROPERTY_MAP_PROPERTIES_PREFIX + "correlationId" + "=" + testCorrelationId));
```
gemmellr, line 2290: This makes the comment below further inaccurate. It already was since the reply-to field was added without any test in https://github.com/apache/activemq-artemis/commit/cbd8df6459b01f816d922efb967800bf020b4e6a

jbertram: Fixed.

PR #4302, opened 2022-12-03
ARTEMIS-4106 do not set prop w/empty key when converting to OpenWire

jbertram:


artemis-protocols/artemis-openwire-protocol/src/test/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverterTest.java
```diff
@@ -248,6 +248,7 @@ public void testBadPropertyConversion() throws Exception {
       coreMessage.putStringProperty(hdrBrokerInTime, "5678");
       coreMessage.putStringProperty(hdrCommandId, "foo");
       coreMessage.putStringProperty(hdrDroppable, "true");
+      coreMessage.putStringProperty("", "");
```
brusdev, line 251: Maybe CORE messages also should reject empty property names as OpenWire messages:
```
java.lang.IllegalArgumentException: Property name cannot be empty or null

	at org.apache.activemq.command.ActiveMQMessage.setObjectProperty(ActiveMQMessage.java:497)
	...
```

This would also avoid the error on converting to OpenWire CORE messages with empty property names.

gemmellr: I'd agree either the test or the impl should be doing something else than it currently is.

The test doesnt look to do anything to check what happens to this empty prop, so it seems like it would just behave the same if the impl behaviour went back to what it was, i.e it doesnt really test the fixed issue/behaviour.

Perhaps the test could do something like verify the count of expected props to ensure no unexpected props are there beyond those its checking values for. Or as Dom suggests the prop setter impl could also throw (the test could still check there isnt anything unexpected after conversion). Though that will obviously break whatever situation allowed the position there was to be an empty prop key in the Core message to begin with.

jbertram: I updated the Jira and the test to be more clear.

brusdev: @jbertram I see your point on backward compatibility for the CORE protocol but from JMS point of view empty property names are not allowed. What would happen if a core msg with an empty property name Is received by a qpid JMS consumer or a core JMS consumer?

brusdev: Fair enough, it seems a good compromise.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -927,6 +932,10 @@ private static void setAMQMsgObjectProperties(final ActiveMQMessage amqMsg,
                                                  final Set<SimpleString> props) throws IOException {
       for (SimpleString s : props) {
          final String keyStr = s.toString();
+         if (keyStr.length() == 0) {
```
clebertsuconic, line 935: shouldn't it check for keyStr.trim().isEmpty() instead?


If I add a property named "   " it will still go through and be converted.


as a matter of fact the test could add such condition?

jbertram: Apparently whitespace is legal. I tested it and it passed without any exception.

PR #4303, opened 2022-12-05
ARTEMIS-3168 - more idomatic usage of mock-server-netty - with hasSta…

gtully:
…rted


gtully:
I still have not found the root cause... need to do more digging.

gtully:
the root cause was running test in a kube cluster, the env was taking precedence over system properties provided in the test. Avoiding the mock server! Fixing precedence for config: system props, then env resolved the issue.

PR #4306, opened 2022-12-06
ARTEMIS-4042 - remove use of codec system property from scripts as en…

gtully:
…v var can now be read directly


gtully:
@gemmellr has this any implications for the new upgrade scripts that I am missing? Functionally there is no behaviour change so I think not.

gemmellr:
This should have its own JIRA for tracking the change, not reusing an existing already released one; the earlier change on the 4042 JIRA has been in 2 previous releases already.

gemmellr:
> @gemmellr has this any implications for the new upgrade scripts that I am missing? Functionally there is no behaviour change so I think not.

I dont believe so, it should just 'remove' this as desired, if present in the old file, by virtue of it no longer being in the current file. Easy enough to check the output of an upgrade (or look at the tests) to confirm though.

gtully:
new jira https://issues.apache.org/jira/browse/ARTEMIS-4112 - thanks for the feedback

PR #1, opened 2015-05-22
This is a test...DO NOT MERGE

jbertram:

PR #4307, opened 2022-12-06
ARTEMIS 4045 - Added multistage docker build for an alpine image.

SamTV12345:
Added multistage docker image build

This adds the possibility to create an artemis image with just the command:

 ```docker build -f Docker-alpine -t artemis_alpine  .```

Furthermore it has the smallest attack surface compared to the other images as it is based on alpine and only contains the jre, bash and libaio for running the application.


jbertram:
How is this related to https://issues.apache.org/jira/browse/ARTEMIS-4045?

SamTV12345:
Sorry my bad. Didn't get that the number refers to a jira ticket.

SamTV12345:
Thanks for the hint. I added a message according to your guidelines. This pr solves exactly that issue.

jbertram:
It's not clear to me how your changes are directly related to https://issues.apache.org/jira/browse/ARTEMIS-3042 aside from just the wording. Your commit is just adding a new Dockerfile which is similar to the three other Dockerfiles which are already present. These three Dockerfiles were also already present when ARTEMIS-3042 was opened originally. The Jira description even says, "The current docker image build is not really user friendly or convenient at all." Therefore, adding another version of what already exists doesn't seem to help resolve the issue. Can you shed any light on this?

jbertram:
To be clear, I am happy to merge your commit eventually. I just want the Jira/commit association to be clear for posterity's sake.

artemis-docker/docker-run.sh
```diff
@@ -44,6 +44,4 @@ else
     echo "broker already created, ignoring creation"
 fi
 
-exec ./bin/artemis "$@"
-
-
+exec ./bin/artemis "$@"
```
jbertram, line 47: This change isn't necessary.

artemis-docker/prepare-docker.sh
```diff
@@ -187,4 +187,4 @@ cp ./docker-run.sh "$ARTEMIS_DIST_DIR/docker"
 echo "Docker file support files at:"
 tree "$ARTEMIS_DIST_DIR/docker"
 
-next_step
+next_step
```
jbertram, line 190: This change also isn't necessary.

jbertram (changes requested):
Please remove the unnecessary changes. Thanks!

SamTV12345:
> It's not clear to me how your changes are directly related to https://issues.apache.org/jira/browse/ARTEMIS-3042 aside from just the wording. Your commit is just adding a new Dockerfile which is similar to the three other Dockerfiles which are already present. These three Dockerfiles were also already present when ARTEMIS-3042 was opened originally. The Jira description even says, "The current docker image build is not really user friendly or convenient at all." Therefore, adding another version of what already exists doesn't seem to help resolve the issue. Can you shed any light on this?

A docker multistage build contains one or multiple steps of building an image and one final image that is used during runtime. A multistage build has the advantage that some components are only needed during build time leading to a smaller footprint. It also has the charm that I don't need to know anything about building this image. Everything is handled during build process.

Before my pr you had to follow the guides, run the script, cd into the _TMP/artemis/version folder and start the build process. This can be teadious and result in user errors: At first I forgot to change to that directory I had to read through the proprietary instructions to this project. 

Now you can start directly with building. Just use the Dockerfile as you would with any other Dockerfile and execute ```docker build -t myartemis -f Dockerfile-alpine .``` You don't need to go through any instructions and have a standardized process of building images. 

If you agree I would remove the other Dockerfiles as they are not needed anymore and may confuse the end user.

SamTV12345:
The other two files are solved. Seems Windows/IntelliJ doesn't add a newline to the file.

jbertram (approved):

SamTV12345:
Seems like the failing test is not the cause of my commit. The other pr is also failing.

SamTV12345:
When will this be merged? Is there something to do from my side?

jbertram:
At this point two main things are not clear in your commit:
1. How to use your new Docker file.
2. How your new Docker file fits with the other Docker files already available.

To resolve #1 you should probably just add some documentation. You could do this, for example, in the `artemis-docker/readme.md`.

To resolve #2 you might also just add documentation. However, if the approach you've taken in this new Docker file is superior to the approach offered by the existing Docker files then it may be worth updating those to use your new approach so that users don't need to run `prepare-docker.sh` manually but they still have the option of using CentOS & Ubuntu with JDK or JRE.

SamTV12345:
I'm done. I adapted the README.md to reflect the changes in the build process and added the multistage capability to every Dockerfile.

SamTV12345:
I guess this pr can be merged.

jbertram:
@clebertsuconic, can you take a look at this since I believe you originally created the Docker files.

SamTV12345:
Thanks for reporting back. I'd like to get this pr merged as soon as possible so I can move on to the next ticket. I guess it doesn't make much sense that I create a lot of pr's that don't get merged at the end.
But I can understand that you have other priorities.  I'll wait till I hear back from you.

jbertram:
> I'd like to get this pr merged as soon as possible so I can move on to the next ticket.

Is this PR a dependency for another task you're working on?

I think most folks work on tasks in parallel. Once they send a PR they move on to their next task knowing that the PR will be reviewed and merged asynchronously.

> I guess it doesn't make much sense that I create a lot of pr's that don't get merged at the end.

Have you created other PRs that haven't been merged? I looked through the list of open and closed PRs and I don't see any others from you. Perhaps you used a different account when you created those?

artemis-docker/Dockerfile-centos7-11
```diff
@@ -17,12 +17,30 @@
...
+FROM eclipse-temurin:11-jdk as builder
+ARG CURRENT_VERSION
+
+ENV VERSION=$CURRENT_VERSION
+
+RUN apt update -y && apt upgrade -y && apt install tree curl -y
+
+ADD ./prepare-docker.sh /bin/prepareDocker
+WORKDIR /root/artemis-build
+COPY ./Dockerfile-alpine .
```
jbertram, line 31: Shouldn't this be `COPY ./Dockerfile-centos7-11 .`?

artemis-docker/Dockerfile-ubuntu-11
```diff
@@ -17,8 +17,28 @@
...
+FROM eclipse-temurin:11-jdk as builder
+ARG CURRENT_VERSION
+
+ENV VERSION=$CURRENT_VERSION
+
+RUN apt update -y && apt upgrade -y && apt install tree curl -y
+
+ADD ./prepare-docker.sh /bin/prepareDocker
+WORKDIR /root/artemis-build
+COPY ./Dockerfile-alpine .
```
jbertram, line 31: Shouldn't this be `COPY ./Dockerfile-ubuntu-11 .`?

artemis-docker/Dockerfile-ubuntu-11-jre
```diff
@@ -17,7 +17,30 @@
...
+FROM eclipse-temurin:11-jdk as builder
+ARG CURRENT_VERSION
+
+ENV VERSION=$CURRENT_VERSION
+
+RUN apt update -y && apt upgrade -y && apt install tree curl -y
+
+ADD ./prepare-docker.sh /bin/prepareDocker
+WORKDIR /root/artemis-build
+COPY ./Dockerfile-alpine .
```
jbertram, line 31: Shouldn't this be `COPY ./Dockerfile-ubuntu-11-jre .`?

artemis-docker/readme.md
```diff
@@ -1,140 +1,62 @@
...
-Use the script ./prepare-docker.sh as it will copy the docker files under the
-binary distribution.
+This folder contains four different images for hosting your ActiveMQ Artemis instance in docker
 
-Below is shown the command to prepare the build of the Docker Image starting
-from the local distribution (from the source codes of ActiveMQ Artemis)
+- Alpine
+- CentOS7
+- Ubuntu 11 with JDK
+- Ubuntu 11 with JRE
```
jbertram, line 13: `11` isn't the version of Ubuntu. It is the version of _Java_. Therefore, the list should probably be:
 - Alpine (latest) with JRE 11
 - CentOS 7 with JDK 11
 - Ubuntu (latest LTS) with JDK 11
 - Ubuntu (latest LTS) with JRE 11

jbertram (changes requested):

artemis-docker/readme.md
SamTV12345: You're right.

artemis-docker/Dockerfile-ubuntu-11-jre
SamTV12345: It doesn't matter as I don't need the Dockerfile in my multistage build. So I removed the line in the prepareDocker script that checks if a Dockerfile* is available. It still works as expected.

jbertram (approved):

clebertsuconic:
This looks nice, although I was not able to build the distribution I am building itself. and I'm a bit confused on how I am supposed to use this.


I just built it. Went to artemis-distribution/target/apache-artemis-2.28.0-SNAPSHOT-bin/apache-artemis-2.28.0-SNAPSHOT


and typed:

podman build -f ../../../../artemis-docker/Dockerfile-alpine -t clebert-alpine .



I was not able to build it. got an error


If I removed the . I would always build a 2.17 image


I'm puzzled on how I'm supposed to use your change.


I will defer this PR to next version (that shouldn't be long)

SamTV12345:
This error seems to be correct. I currently download the jar from the apache downloads and copy it into the alpine/centos/ubuntu image. But if you use it for development I have two options:
1. Create a new Dockerfile for building the image. 
2. Modify the existing Dockerfile.

Which one do you prefer?

jbertram:
@SamTV12345, FWIW instructions on how to build are in the ["Building" chapter of the Hacking Guide](https://github.com/apache/activemq-artemis/blob/main/docs/hacking-guide/en/building.md):

> To build the full release with documentation, Javadocs, and the full web site:
> 
> `$ mvn -Prelease package`
>
> To install it to your local maven repo:
>
> `$ mvn -Prelease install`

I would add `-DskipTests` to speed up the build significantly.

jbertram:
@SamTV12345, I'm not sure where that's coming from. Searching the code-base for `os.detected.classifier` turns up 0 results. What exact command did you execute and what branch are you on?

SamTV12345:
I executed `mvn dependency:resolve-plugins dependency:go-offline` on my branch to cache dependencies.

jbertram:
I reproduced the error by executing `mvn dependency:go-offline`. It's not clear to me whether the problem is in the ActiveMQ Artemis `pom.xml` or in one of Netty's. What happens if you just run `mvn -Prelease install -DskipTests`?

SamTV12345:
That should work without a problem. But you need to be aware that this will always download every dependency when you make a change in any file of the project.

jbertram:
Looks like this can be fixed by adding this to the `<build>` of the main `pom.xml`:
```xml
      <extensions>
         <!-- Generate os.detected.classifier property -->
         <extension>
            <groupId>kr.motd.maven</groupId>
            <artifactId>os-maven-plugin</artifactId>
            <version>1.7.1</version>
         </extension>
      </extensions>
```

jbertram:
I rarely (if ever) use Maven's offline mode as downloading dependencies isn't a problem for me, but I sent #4355 to support those who need it. Hope that helps!

SamTV12345:
Thanks for creating the pr. I will checkout your commit hopefully this week. It seems like the release profile really takes a long time. 4GB weren't enough for testing. So the question is how many times would somebody do that when it takes more than 10 minutes to compile.

jbertram:
Those numbers seem excessive to me. When I eliminate the time required to download artifacts by first using `mvn dependency:go-offline` and then run:
```
mvn -o clean install -Prelease -DskipTests
```
It finishes in half the time you're reporting:
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  05:18 min
[INFO] Finished at: 2023-01-31T17:32:56-06:00
[INFO] ------------------------------------------------------------------------
```
Also the maximum heap usage is _barely_ over 2G:
mavenHeapUsage
I'm running this on a laptop that's not particularly fancy.

SamTV12345:
Weird. Last time I waited 20 minutes and nothing changed. I am stuck at this screen 
image



If you have time could you start this Dockerfile

````yaml
ARG CURRENT_VERSION=2.17.0

FROM maven:3.8.7-eclipse-temurin-11  as builder
ARG CURRENT_VERSION

COPY settings.xml /.m2/settings.xml
ENV MAVEN_SETTINGS_PATH=/.m2/settings.xml

COPY . .

RUN mvn -ntp clean  install -Prelease -DskipTests -Dmaven.repo.local=/.m2/repository -s /.m2/settings.xml

FROM alpine:latest

ARG CURRENT_VERSION

ENV VERSION=$CURRENT_VERSION

RUN apk --no-cache add openjdk17-jre-headless bash libaio\
        --repository=http://dl-cdn.alpinelinux.org/alpine/edge/community


LABEL maintainer="Apache ActiveMQ Team"
# Make sure pipes are considered to determine success, see: https://github.com/hadolint/hadolint/wiki/DL4006
SHELL ["/bin/bash", "-o", "pipefail", "-c"]
WORKDIR /opt

ENV ARTEMIS_USER artemis
ENV ARTEMIS_PASSWORD artemis
ENV ANONYMOUS_LOGIN false
ENV EXTRA_ARGS --http-host 0.0.0.0 --relax-jolokia

# add user and group for artemis
RUN addgroup  -g 1001 artemis && adduser -u 1002 --ingroup artemis --disabled-password  artemis

USER artemis

COPY --from=builder /root/artemis-build/_TMP_/artemis/${VERSION}/ /opt/activemq-artemis

# Web Server
EXPOSE 8161 \
# JMX Exporter
    9404 \
# Port for CORE,MQTT,AMQP,HORNETQ,STOMP,OPENWIRE
    61616 \
# Port for HORNETQ,STOMP
    5445 \
# Port for AMQP
    5672 \
# Port for MQTT
    1883 \
#Port for STOMP
    61613

USER root

RUN mkdir /var/lib/artemis-instance && chown -R artemis.artemis /var/lib/artemis-instance

COPY --from=builder /root/artemis-build/_TMP_/artemis/${VERSION}/docker/docker-run.sh /

USER artemis

# Expose some outstanding folders
VOLUME ["/var/lib/artemis-instance"]
WORKDIR /var/lib/artemis-instance

ENTRYPOINT ["/docker-run.sh"]
CMD ["run"]
```

If you don't have a settings.xml defined you need to remove the instructions COPY settings.xml /.m2/settings.xml, ENV MAVEN_SETTINGS_PATH=/.m2/settings.xml and the part in the mvn clean install script starting with -Dnaveb.repo.local...

After copying the Dockerfile in artemis-docker, execute from the root of the project execute `docker build -t test -f ./artemis-docker/Dockerfile-build-alpine .`

It is currently not working I just want to make sure that the build in the Dockerfile is working

The specs of my proxmox test machine are:
i5 4570T and 16 Gigabyte RAM. 

The alpine Linux LXC for building has all cores and 8 Gigabyte assigned.

SamTV12345:
I had the problem that it said Java heap space error because heap space was full. I then increased it with ENV MAVEN_OPTS="-Xmx3000m -XX:MaxPermSize=500m" to 3GB of heap space. Now it just states killed and an error code of 137. I still fail at the step mentioned in the screenshot.

SamTV12345:
I am done including your feedback @clebertsuconic . You can now build images with Dockerfiles ending with build while the others can be used to build images regularly with the latest underlying image. cc @jbertram

clebertsuconic:
@SamTV12345 thanks.. will merge before the release next week for sure

ViliusS:
I'm sorry, but I think this is a step back for CI workflows.

Previously what you had to do to integrate Artemis docker builds into CI pipeline is just download provided Docker file + prepare/run.sh files in the pipeline from Artemis repository and run a few commands. This allowed us to add custom configuration or additional files to docker image itself. Now the preparation step is "hidden" inside docker file. So in order to customize the image we will need to customize provided Docker file which destroys the purpose of that file all together.

Multistage docker builds are great for building the the software from source _for final deployment_, but they are not so great  for processes which require customization.

SamTV12345:
I think Docker multistage builds are great. I don't need to configure anything on the host machine and can simply press "build" in Jenkins. Other than that you can still always customize the Dockerfile by cloning your own Dockerfile before docker build and replace the standard build with your own.

ViliusS:
> I think Docker multistage builds are great. I don't need to configure anything on the host machine and can simply press "build" in Jenkins.

I don't argue that, but as I said, such images are great only for final deployment or local development.

> Other than that you can still always customize the Dockerfile by cloning your own Dockerfile before docker build and replace the standard build with your own.

I know that, however this PR changes the purpose of these Dockerfiles. Previously they contained only configuration needed for the runtime and served more as a tool to build your own custom images. Now they contain old the build flow inside, hence they now serve just as easier way to run Artemis.

ViliusS:
I gave this more thought today. In theory, if the requirement is to just conveniently allow users to build docker image and run it locally there is a docker-compose exactly for that purpose. All prepare/run steps can be automated via Docker Compose YAML file. 
It also allows for better build and runtime separation.

SamTV12345:
The question is how minimal is the produced image? Could you create a docker-compose with a few steps that would satisfy the requirement for different base images? If we talk about a docker-compose yaml you also need suitable tooling in form of a docker compose cli plugin or v1 docker-compose. If you could easily switch between base images and select them by a build arg that would indeed be interesting. I am unfortunately currently pretty involved in other FOSS projects so if you could please go through creating said yaml I should be able to adapt the scripts and add the alpine image which was one of the reasons this pr exists.

amarkevich:
Minor suggestion:
> apt install tree curl -y
can be replaced with existing tools:
- tree -> ls -l (easy one)
- curl -> wget (more complex)

SamTV12345:
> Minor suggestion:
> 
> > apt install tree curl -y
> 
> can be replaced with existing tools:
> 
> * tree -> ls -l (easy one)
> * curl -> wget (more complex)

This is performed in the builder section. This section is executed one time at the beginning of the pipeline. Any subsequent pipeline runs in the future will have this command cached and it will never be rerun as it is done just at the beginning. I tuned the Dockerfile to cache the things that are static like apt packages and add the dynamic parts of the app e.g. artemis at the end.

amarkevich:
> command cached

anyway 'tree' can be safely removed https://github.com/apache/activemq-artemis/pull/4453

SamTV12345:
> > command cached
> 
> anyway 'tree' can be safely removed #4453

Removed.

artemis-docker/prepare-docker.sh
```diff
@@ -181,10 +163,9 @@ if [ ! -d "${ARTEMIS_DIST_DIR}/docker" ]; then
   mkdir "${ARTEMIS_DIST_DIR}/docker"
 fi
 
-cp ./Dockerfile-* "$ARTEMIS_DIST_DIR/docker"
 cp ./docker-run.sh "$ARTEMIS_DIST_DIR/docker"
 
 echo "Docker file support files at:"
 tree "$ARTEMIS_DIST_DIR/docker"
```
amarkevich, line 169: its still there

SamTV12345: What exactly is still there? tree has been removed.

ViliusS:
> The question is how minimal is the produced image? Could you create a docker-compose with a few steps that would satisfy the requirement for different base images? If we talk about a docker-compose yaml you also need suitable tooling in form of a docker compose cli plugin or v1 docker-compose. If you could easily switch between base images and select them by a build arg that would indeed be interesting. I am unfortunately currently pretty involved in other FOSS projects so if you could please go through creating said yaml I should be able to adapt the scripts and add the alpine image which was one of the reasons this pr exists.

I did had much time lately, but I will try to come up with YAML file which does the same as this PR this week.

Produced image won't change because this doesn't depend on a tooling. Compose plugin is installed by default with Docker Desktop, so it's not a problem either.

artemis-docker/prepare-docker.sh
amarkevich: 'tree' command in L188 will cause build error since not available in build stage. I propose to remove it like done in https://github.com/apache/activemq-artemis/pull/4453/files#diff-09d0fd7b380274547af0cdaf1cb41c22b3990dfed9848b560e9d9f77b97fb204L188
or replace with 'ls -l'

SamTV12345:
I removed the leftover tree command. I maybe can ditch all these sh files and build it with docker-compose. @ViliusS  I'll try that today. Maybe I find a clever way to build that with docker-compose

SamTV12345:
> > The question is how minimal is the produced image? Could you create a docker-compose with a few steps that would satisfy the requirement for different base images? If we talk about a docker-compose yaml you also need suitable tooling in form of a docker compose cli plugin or v1 docker-compose. If you could easily switch between base images and select them by a build arg that would indeed be interesting. I am unfortunately currently pretty involved in other FOSS projects so if you could please go through creating said yaml I should be able to adapt the scripts and add the alpine image which was one of the reasons this pr exists.
> 
> I didn't had much time lately, but I will try to come up with YAML file which does the same as this PR this week.
> 
> Produced image won't change because this doesn't depend on a tooling. Compose plugin is installed by default with Docker Desktop, so it's not a problem either.

It's implemented. It is now save to review and hopefully merge.

ViliusS:
I've went through couple of different YAML and bash solution iterations today. I have also tested the solution provided in this PR. I still have to conclude that at least for us multistage builds give more headache than it solves.

My main concern, as mentioned previously, is that Artemis preparation process will be integrated into Docker image which doesn't allow for easy customization. For example previously we had such Jenkins CI pipeline:

```
      stage('prepare configuration') {
        steps {
          sh "./prepare-docker.sh --from-release --artemis-version $ARTEMIS_VERSION"
          sh "........ our modifications and custom configuration goes here........."
        }
      }
      stage('build image') {
        steps {
          script {
            docker.withRegistry(env.DOCKER__REPOSITORY, env.DOCKER_LOGIN) {
              sh '''
                cd _TMP_/artemis/$ARTEMIS_VERSION
                docker build -f ./docker/Dockerfile-ubuntu-17-jre -t $BUILD_IMAGE $BUILD_DIR
                docker push $BUILD_IMAGE
                rm -rf _TMP_
              '''
            }
          }
        }
      }
```

If this PR is ought to be accepted we won't be able to modify Artemis image as easily anymore. We would need to download official docker file, modify it by using "patch" or similar tools and only then build our image. This means we would need to maintain our own Dockerfile which looses the purpose of having official Docker file altogether. It also integrates preparation process into build process which lowers visibility in Jenkins (or any other CI system) UI.

The final word will be on Artemis maintainers, by my proposal is to leave current Docker files as their are. If building Artemis images for fast local development is really a problem (which I don't think it is to run 2 commands instead of 1), then have separate Docker files. Or better yet, have official Docker image uploaded to Docker Hub for easy run with default settings. That way Docker files can stay in the repo for those wanting to modify the image.

On a side note, what I would really want to see happening is proper customization via environment variables or Docker entrypoint as mentioned in [ARTEMIS-3042](https://issues.apache.org/jira/browse/ARTEMIS-3042).

SamTV12345:
What exactly is this step sh "........ our modifications and custom configuration goes here........." and can't you simply do this before the jar is put into the container?

ViliusS:
> What exactly is this step sh "........ our modifications and custom configuration goes here........." and can't you simply do this before the jar is put into the container? The final build step is simply take the compiled jar with modifications if you like and put it into the various docker images.

Modifications to web console for example.

In the past this was also the place where we applied our own patches before building Artemis from source.

SamTV12345:
Well that should be possible without a problem. You do your changes in the  jenkins pipeline first, then compile it on the machine with `mvn clean install` or a similar command and then simply copy it to ./artemis-docker/target/apache-artemis*-bin/*SNAPSHOT . After that just `docker-compose -f artemis-docker/docker-compose-final.yml up -d` and you are done.

ViliusS:
> Well that should be possible without a problem. You do your changes in the jenkins pipeline first, then compile it on the machine with `mvn clean install` or a similar command and then simply copy it to ./artemis-docker/target/apache-artemis*-bin/*SNAPSHOT . After that just `docker-compose -f artemis-docker/docker-compose-final.yml up -d` and you are done.

As I said multiple times before, we can't without modifying provided Dockerfile. For modification to happen tarball needs to be downloaded and extracted first.

clebertsuconic:
@SamTV12345 @jbertram what is really the advantage of using this? just to avoid the prepare phase?


This is doing a build all over again?


The ./prepare.sh and build seemed a lot simpler to me.


if you really wanted to have this committed, we could create a new folder ./artemis-distribution-multistage ... or move all the docker possibilities under a single directory.


although it gets complicated with competing things.. we already have the ./artemis-image being a different OCI image by @gtully

clebertsuconic:
@ViliusS do you understand the advantage of multistage builds?

@SamTV12345 I'm trying to understand the difference here? what's the issue with ./prepare.sh?


again.. this seems moot by ./artemis-image ... did you look at it?

clebertsuconic:
it would be nice if we could chat somewhere? Slack at Apache? 

at the end of build it just hangs as it's running the brokers

SamTV12345:
> is this making a docker official build any easier? is that the reason behind this?

This turbocharges new builds. You can build all platforms in next to no time and push it to Dockerhub. It streamlines the build process. Before that you had to go through the steps in the ./prepare.sh. Then go to the sub directory, then start building dockerfiles one after the other. If I want to start the whole build that would mean 1 x prepare.sh, 5-6 times Dockerfiles with the right tags. Non parallel builds. With this approach I can build all 6 images in around 5 minutes with minimum resource usage of the maven repository. 
This docker compose also allows for non contributors/people new to the project build images with their known tooling docker and docker-compose. 
Last but not least you can simple select an image with docker composer service <my-desired-image> up.

SamTV12345:
> it would be nice if we could chat somewhere? Slack at Apache?
> 
> at the end of build it just hangs as it's running the brokers

You can add my developer email samelus1998@outlook.de and we can go through the build process. Without any screenshots/output its hard to evaluate.

ViliusS:
> @ViliusS do you understand the advantage of multistage builds?

If you are asking in general, then yes. It's just a mechanism to separate build stage from the runtime.

In this particular case I see no advantage, sorry. These dockerfiles previously were used just for runtime, by packing downloaded already built releases. So the separation was already there.

clebertsuconic:
@SamTV12345 I had everything ready on my build locally, building the image for me would been a couple seconds... and this made everything from scratch, including downloading the entire maven artifacts somewhere.


I wouldn't have a problem if this was an additional option to build the image, meaning if I want to use the "regular" way we could just do what we have been doing so far. in that sense if you could reuse the docker files as they are now, and just call the "multi-stage" build as an option.. then I would be ok... as long as it is well documented.

SamTV12345:
@clebertsuconic I rebased the commits. What should be done for this pr to be merged?

jbertram:
No movement in over 2 years on this. Closing.

SamTV12345:
Yes because no one cared for this merge request. I won't commit in the future to this repository. Thank you.

jbertram:
@SamTV12345, I appreciate the engagement on this, but, as I understand it, the changes weren't compelling enough to be merged. Additional PRs are always welcome!

jbertram:
@SamTV12345, to be clear, there were also conflicts which prevented merging anyway.

PR #4308, opened 2022-12-07
Update ActiveMQSession.java

Yesenkov:
When calling Session.recover(), unacknowledged messages must be returned to the queue, that is, ClientSession.rollback (false). If you call ClientSession.rollback (true), the ActiveMQServerMessagePlugin.messageAcknowledged () is called. Which is illogical and misleading. However, this introduces the problem that deliveryCount does not increase on rereads. ut this needs to be fixed elsewhere.

### DIFF
--- a/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
+++ b/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
@@ -338,7 +338,13 @@ public class ActiveMQSession implements QueueSession, TopicSession {
       }
 
       try {
-         session.rollback(true);
+         /*
+         When calling Session.recover(), unacknowledged messages must be returned to the queue, that is, ClientSession.rollback (false). 
+         If you call ClientSession.rollback (true), the ActiveMQServerMessagePlugin.messageAcknowledged () is called. Which is illogical 
+         and misleading.
+         However, this introduces the problem that deliveryCount does not increase on rereads. ut this needs to be fixed elsewhere.
+         */
+         session.rollback (false);
       } catch (ActiveMQException e) {
          throw JMSExceptionHelper.convertFromActiveMQException(e);
       }
--- a/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
+++ b/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
@@ -66,6 +66,7 @@ import org.apache.activemq.artemis.api.core.client.ClientSession.AddressQuery;
 import org.apache.activemq.artemis.api.core.client.ClientSession.QueueQuery;
 import org.apache.activemq.artemis.api.core.RoutingType;
 import org.apache.activemq.artemis.core.protocol.core.impl.PacketImpl;
+import org.apache.activemq.artemis.api.jms.ActiveMQJMSConstants;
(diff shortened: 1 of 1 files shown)


jbertram:
At the very least this needs both a Jira and an integration test to confirm the fix is valid and mitigate against future regressions.

jbertram:
Also, you'll need to squash all your commits and reformat your commit message as detailed in the [hacking guide](https://github.com/apache/activemq-artemis/blob/main/docs/hacking-guide/en/code.md#commitMessageDetails).

Yesenkov:
OK!
Please point me where to required test made.

jbertram:
You could probably add your test to `org.apache.activemq.artemis.tests.integration.openwire.amq.JmsTopicRedeliverTest` as it already has a test for `Session.recover()`.

Yesenkov:
Okay, I'll do that.

jbertram:
You could create a new test called `RecoverTest` in `org.apache.activemq.artemis.tests.integration.jms.client` (i.e. at `tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/client`). There are other JMS tests in that package which will give you an idea of how to create your connection, etc.

jbertram:
To be clear, this would not qualify as a "NO-JIRA" change. It definitely needs a Jiras as noted previously.

jbertram:
I can create one for you if you provide:
- email address
- preferred username (N.B. hyphens not allowed)
- alternate username (in case the preferred one is already in use)
- display name, if it is different from the username

Yesenkov:
Andrey.Yesenkov@gmail.com
Andrey.Yesenkov@gmail.com - preferred username (N.B. hyphens not allowed)
Andrey.Yesenkov - alternate username (in case the preferred one is already in use)
Andrey Yesenkov

jbertram:
Done! I used `andreyyesenkov` for the username as no capital letters and no punctuation is allowed.

clebertsuconic:
- you need to squash all your commits into a single one, git push -f on your branch
- you need to provide a test

I'm not convinced there's anything broken. you need to provide a test as part of your commit.

or if you are convinced there's an issue, you need to at least provide a way to reproduce and someone would take care of this issue and make a new PR with the test and the proper fix.

Yesenkov:
> I had to squash all your commits, and all you're doing is to only call true at the rollback(false) in case of individualAck
> 
> your concern here is on the call to the plugin itself (based on your texts)?
Well I am writing a test showing an error. If it is not necessary, then it is not necessary, we will not call Session.recover (), we will call Session.getCoreSession().rollback(), which works correctly.
In the case of Session.Recover(), all unacknowledged messages should be returned to the queue, this is not only for the individual acknowledge mode, I just don't care about other modes. That is, in any case, rollback (false) from CoreSession should be called
For unacknowledged messages, ActiveMQServerMessagePlugin.messageAcknowledged (Normal) should not be called, when acknowledged - must - this is important. The same situation in Session.rollback () - ActiveMQServerMessagePlugin.messageAcknowledged (Normal) should not be called.
Current implementation callActiveMQServerMessagePlugin.messageAcknowledged (Normal) on Session.recover () & Session.rollback (). If calling from Session.recover () & Session.rollback ()  - Session.getCoreSession().rollback(false) - all works corectly, all messages on his right places, and all plugin calls maded correctly.

clebertsuconic:
>>. Well I am writing a test showing an error.


Where? I did not see your test in any of your commits.

jbertram:
@Yesenkov, the commit needs to include a test in order to validate the fix is correct and to mitigate against future regressions. Without a test this PR will be closed. Thanks!

PR #4310, opened 2022-12-12
ARTEMIS-4113 Fix NPE for backup brokers with connection routers

brusdev:
The nodeID on backup brokers is available only after they become live.


jbertram:
@brusdev, any news on this one? Do you want this in 2.28.0?

brusdev:
@jbertram the fix is completed but I would like to add other tests, I'll try on Monday.

PR #4312, opened 2022-12-13
ARTEMIS-3993 fixed some MB style uses to use MiB; added support for MiB/GiB/etc in config

erwindon:
See https://issues.apache.org/jira/browse/ARTEMIS-3993

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java
@@ -39,9 +39,9 @@ public class ByteUtil {
    private static final String prefix = "^\\s*(\\d+)\\s*";
    private static final String suffix = "(b)?\\s*$";
    private static final Pattern ONE = Pattern.compile(prefix + suffix, Pattern.CASE_INSENSITIVE);
-   private static final Pattern KILO = Pattern.compile(prefix + "k" + suffix, Pattern.CASE_INSENSITIVE);
-   private static final Pattern MEGA = Pattern.compile(prefix + "m" + suffix, Pattern.CASE_INSENSITIVE);
-   private static final Pattern GIGA = Pattern.compile(prefix + "g" + suffix, Pattern.CASE_INSENSITIVE);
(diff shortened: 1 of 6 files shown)


clebertsuconic:
can you fix FileConfigurationParserTest.testNotations, and amend your PR with the fix?

other than that LGTM

erwindon:
ow! thanks, I will take a look.
while reviewing my code I realized that the proposed code would also allow adding "i" in trivial sizes, e.g. "12345i". that is now a separate commit. but I'll squash all before un-draft-ing it.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ByteUtil.java
```diff
@@ -39,9 +39,9 @@ public class ByteUtil {
    private static final String prefix = "^\\s*(\\d+)\\s*";
    private static final String suffix = "(b)?\\s*$";
    private static final Pattern ONE = Pattern.compile(prefix + suffix, Pattern.CASE_INSENSITIVE);
-   private static final Pattern KILO = Pattern.compile(prefix + "k" + suffix, Pattern.CASE_INSENSITIVE);
-   private static final Pattern MEGA = Pattern.compile(prefix + "m" + suffix, Pattern.CASE_INSENSITIVE);
-   private static final Pattern GIGA = Pattern.compile(prefix + "g" + suffix, Pattern.CASE_INSENSITIVE);
+   private static final Pattern KILO = Pattern.compile(prefix + "ki?" + suffix, Pattern.CASE_INSENSITIVE);
```
michaelandrepearce, line 42: This will break anyone using existing styles. Add new style support but don't break existing

erwindon: Michael, that was strictly my intention. the unit test that I added verifies that. how does this break anything?

clebertsuconic: @michaelandrepearce I didn't see him changing the style.. just adding  a new one... unless I missed something.

PR #4313, opened 2022-12-15
ARTEMIS-3609 Do not use netty thread for thread completion listener

clebertsuconic:


clebertsuconic:
I am keeping the page-counter commit on this PR as I am combining tests on the CI.

PR #4314, opened 2022-12-15
ARTEMIS-4115 ArrayIndexOutOfBoundsException when dup cache size is 0

jbertram:


clebertsuconic:
@jbertram test failures...


if you could also run a whole CI in your env before we merge this please?

jbertram:
Testsuite was clean. Merging...

PR #4315, opened 2022-12-15
This PR has a few minor changes that I am testing altogether before merging.

clebertsuconic:


clebertsuconic:
@jbertram I will still expand ARTEMIS-4116

clebertsuconic:
I made this a draft since I will add more operations

clebertsuconic:
I am adding more calls to managementLock.


BTW: I thought about renaming ActiveMQServer::managementLock as managementSemop, however using a Semaphore here is like an implementation detail. I will keep it as managementLock unless someone help me with an argument to rename it or find me a better name :)

PR #4318, opened 2022-12-19
ARTEMIS-4120 show description for mqtt-qos header

erwindon:
see https://issues.apache.org/jira/browse/ARTEMIS-4120

### DIFF
--- a/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/browse.js
+++ b/artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/browse.js
@@ -280,7 +280,7 @@ var Artemis;
                 itemField: 'type',
                 header: 'Type',
                 templateFn: function(value) {
-                    return formatType(value);
+                    return formatWithList(value, typeLabels);
                 }
             },
             {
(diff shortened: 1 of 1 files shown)


artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/browse.js
```diff
@@ -528,6 +528,8 @@ var Artemis;
 
         var routingTypes = ["multicast", "anycast"];
 
+        var mqttQosTypes = ["at most once", "at least once", "exactly once"];
```
jbertram, line 531: I think the numeral should be retained here along with the description, e.g.:
```java
var mqttQosTypes = ["0 - at most once", "1 - at least once", "2 - exactly once"];
```
This is because lots of MQTT documentation, tutorials, and even the specification refer to QoS values directly by their numeral without any description so lots of folks more readily recognize the numeral. Also, lots of APIs use an integer for the QoS value so many developers will be accustomed to this representation as well.

erwindon: @jbertram the formatting function does that always, see the screen print in https://issues.apache.org/jira/browse/ARTEMIS-4120

jbertram: I just saw your screenshot. Thanks for the clarification. :+1:

PR #4320, opened 2022-12-21
ARTEMIS-3790 Support masked JMS credentials

ryan-highley:
Adds standard username and password value unmasking for JMS ActiveMQConnection instances and its subclasses

### DIFF
--- a/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQConnectionFactory.java
+++ b/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQConnectionFactory.java
@@ -59,6 +59,7 @@ import org.apache.activemq.artemis.spi.core.remoting.ClientProtocolManagerFactor
 import org.apache.activemq.artemis.uri.ConnectionFactoryParser;
 import org.apache.activemq.artemis.uri.ServerLocatorParser;
 import org.apache.activemq.artemis.utils.ClassloadingUtil;
+import org.apache.activemq.artemis.utils.PasswordMaskingUtil;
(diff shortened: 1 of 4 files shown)


jbertram:
My only concern here is that there's no way to set the codec. It is hard-coded to use the default coded. I think the simplest way to implement codec configuration is via a new property on `org.apache.activemq.artemis.jms.client.ActiveMQConnectionFactory` much like `deserializationWhiteList` so that the value could be set programmatically or via the connection URL. For example, if you created a `String` named `passwordCodec` then this could be set using `ActiveMQConnectionFactory.setPasswordCodec` or, e.g., something like `tcp://myHost:61616?passwordCodec=com.foo.Codec`. Then any encrypted credentials used on connections created from this connection factory would use that codec.

jbertram:
@ryan-highley, any feedback here?

ryan-highley:
@jbertram Sorry for the delay--day job and all....

That makes sense. I was attempting to move the unmasking with as little invasiveness as possible, hence the additional properties and unmask calls in the ActiveMQConnection constructor.

Looking at the `o.a.a.a.jms.client.ConnectionFactoryOptions` for how `deserializationWhiteList` is implemented led me to the equivalent unmasking approach in the resource adapter, which handles the `passwordCodec` property in its ConnectionFactory as you described.

I'll update this code to enable setting the codec aligning with the precedent in the RA code.

jbertram:
@ryan-highley, no worries. I just didn't want this to fall completely off our radar. Your next steps sounds good. I'll review again once you push. Thanks!

jbertram:
@ryan-highley, nice work! Thanks a lot.

PR #4328, opened 2023-01-12
ARTEMIS-4127 refactor multiprotocol JMS tests in AMQP package

jbertram:
Over time org.apache.activemq.artemis.tests.integration.amqp has become home to many multi-protocol JMS tests even though the package is really for AMQP-specific tests. This commit splits those tests out into their own package.

This is a preliminary step to clarify these tests before I add another one for a different issue.


michaelandrepearce (approved):

michaelandrepearce:
This change looks good, there is test failure, but don't believe its related to this change. But if in another cycle anyone has a chance to look at would be good, bit of a worry.......


Build / Test (17) (pull_request) Failing after 25m 

- [INFO] 
Error:  Failures: 
Error:    AcknowledgementTest.testDupsOKAcknowledgeQueue:555->ActiveMQServerTestCase.assertRemainingMessages:364 expected:<20> but was:<21>

PR #4329, opened 2023-01-12
ARTEMIS-4125 Fixing AutoCreate Topics remove of the address

clebertsuconic:


jbertram:
@clebertsuconic, LGTM

PR #4338, opened 2023-01-18
ARTEMIS-4136 Mirrored sync replica

clebertsuconic:
I am adding an option sync=true or false on mirror. if sync, any client blocking operation will wait a roundtrip to the mirror acting like a sync replica.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
@@ -16,6 +16,7 @@
  */
 package org.apache.activemq.artemis.protocol.amqp.connect.mirror;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -24,6 +25,9 @@ import org.apache.activemq.artemis.api.core.Message;
 import org.apache.activemq.artemis.api.core.QueueConfiguration;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.config.amqpBrokerConnectivity.AMQPMirrorBrokerConnectionElement;
+import org.apache.activemq.artemis.core.io.IOCallback;
+import org.apache.activemq.artemis.core.persistence.OperationContext;
+import org.apache.activemq.artemis.core.persistence.impl.journal.OperationContextImpl;
 import org.apache.activemq.artemis.core.postoffice.impl.PostOfficeImpl;
 import org.apache.activemq.artemis.core.server.ActiveMQComponent;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
@@ -34,6 +38,9 @@ import org.apache.activemq.artemis.core.server.impl.AckReason;
 import org.apache.activemq.artemis.core.server.impl.AddressInfo;
 import org.apache.activemq.artemis.core.server.impl.RoutingContextImpl;
 import org.apache.activemq.artemis.core.server.mirror.MirrorController;
+import org.apache.activemq.artemis.core.transaction.Transaction;
+import org.apache.activemq.artemis.core.transaction.TransactionOperationAbstract;
+import org.apache.activemq.artemis.core.transaction.TransactionPropertyIndexes;
 import org.apache.activemq.artemis.protocol.amqp.broker.AMQPMessage;
 import org.apache.activemq.artemis.protocol.amqp.broker.AMQPMessageBrokerAccessor;
 import org.apache.activemq.artemis.protocol.amqp.broker.ProtonProtocolManager;
@@ -86,6 +93,7 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
    final boolean deleteQueues;
    final MirrorAddressFilter addressFilter;
    private final AMQPBrokerConnection brokerConnection;
+   private final boolean sync;
 
    final AMQPMirrorBrokerConnectionElement replicaConfig;
 
@@ -116,6 +124,7 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
       this.addressFilter = new MirrorAddressFilter(replicaConfig.getAddressFilter());
       this.acks = replicaConfig.isMessageAcknowledgements();
       this.brokerConnection = brokerConnection;
+      this.sync = replicaConfig.isSync();
    }
 
    public Queue getSnfQueue() {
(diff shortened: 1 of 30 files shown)


clebertsuconic:
hmmm... I need to change the docs..... I will do this on Friday... (It could be a separate PR fro this one).

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPSyncMirrorTest.java
```diff
@@ -0,0 +1,481 @@
...
+                  try {
+                     session.commit();
+                     sendPending.countDown();
+                  } catch (Throwable e) {
+                     e.printStackTrace();
+                  }
+               });
+            }
+
+            Assert.assertFalse("Send epending not supposed to succeed", sendPending.await(10, TimeUnit.MILLISECONDS));
```
tabish121, line 292: probably meant: ""Send pending not supposed to succeed"

clebertsuconic: thanks!

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -107,6 +115,7 @@ public boolean isStarted() {
    public AMQPMirrorControllerSource(ProtonProtocolManager protonProtocolManager, Queue snfQueue, ActiveMQServer server, AMQPMirrorBrokerConnectionElement replicaConfig,
                                      AMQPBrokerConnection brokerConnection) {
       super(server);
+      assert server != null;
```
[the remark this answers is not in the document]
clebertsuconic: I will remove this.. I added this just to debug and forgot it.. thanks

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -239,37 +249,107 @@ public void sendMessage(Message message, RoutingContext context, List<MessageRef
...
+         List<MessageReference> refs;
+
+         if (tx != null) {
+            if (logger.isDebugEnabled()) {
+               logger.debug("Mirroring Message " + message + " with TX");
+            }
+            MirrorSendOperation mirrorSendOperation = getSendOperation(tx);
+            refs = mirrorSendOperation.mirroredRefs;
+         } else {
+            refs = new LinkedList<>();
```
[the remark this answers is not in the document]
clebertsuconic: actually, I will just use the ref directly.. thanks a lot!

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
```diff
@@ -128,11 +133,17 @@ public void storeLineUp() {
    @Override
    public void replicationLineUp() {
       REPLICATION_LINEUP_UPDATER.incrementAndGet(this);
+      if (logger.isTraceEnabled()) {
```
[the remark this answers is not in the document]
clebertsuconic: I usually always use a gate when there's a native type (integer, long) to avoid a conversion to an Object. Do you think that's an overkill?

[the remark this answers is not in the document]
tabish121: It probably will never be an actual performance issue but if you feel it adds value then that's fine.  I tend to find the gate just make the code harder to read.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -318,28 +395,181 @@ public void postAcknowledge(MessageReference ref, AckReason reason) throws Excep
...
+      }
+
+      return sendOperation;
+   }
+
+   private static class MirrorACKOperation extends TransactionOperationAbstract {
+
+      final ActiveMQServer server;
+
+      final LinkedList<MessageReference> refs = new LinkedList<>();
```
tabish121, line 479: I cringe any time I see a LinkedList used as they are almost always the wrong choice as opposed to the simpler ArrayList, what is driving the use of these over the simpler option?

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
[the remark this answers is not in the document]
gemmellr: I decide based on how hot the path is. If its 'relatively occasional' I'll just let it box if it leaves things a lot nicer to read. If its very hot, I'll gate it. I have no idea in this case :)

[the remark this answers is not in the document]
clebertsuconic: OperationContext is burning hot actually. I will place the gate back in this case.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
clebertsuconic: the number of elements is variable. It could be from one element to a large transaction.

tabish121: I'm not sure how that's an argument for using a LinkedList....

clebertsuconic: actually... all the other TransactionOperations that are using a similar pattern are using ArrayList. 


I just will just keep the same here.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -2447,6 +2447,14 @@
             </xsd:documentation>
          </xsd:annotation>
       </xsd:attribute>
+      <xsd:attribute name="sync" type="xsd:boolean" use="optional" default="true">
+         <xsd:annotation>
+            <xsd:documentation>
+               If this is true, producers will be waiting a response from the mirror before the sync is finished.
+               This is false by default.
```
gemmellr, line 2454: Seems conflicted:

default="true"

"This is false by default."

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPReplicaTest.java
```diff
@@ -698,7 +700,7 @@ private void replicaTest(boolean largeMessage,
       Connection connection = factory.createConnection();
       Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
       MessageProducer producer = session.createProducer(session.createQueue(getQueueName()));
-      producer.setDeliveryMode(DeliveryMode.NON_PERSISTENT);
+      producer.setDeliveryMode(DeliveryMode.PERSISTENT);
```
gemmellr, line 703: Could also just remove the line.

clebertsuconic:
I just added the documentation. I will keep it as a separate commit.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
clebertsuconic: you're right.. thanks

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -318,28 +397,178 @@ public void postAcknowledge(MessageReference ref, AckReason reason) throws Excep
...
+
+      return sendOperation;
+   }
+
+   private static class MirrorACKOperation extends TransactionOperationAbstract {
+
+      final ActiveMQServer server;
+
+      // This pair contains the Message used to generate the command towards the target, the reference being acked
+      final List<Pair<Message, MessageReference>> acks = new ArrayList<>();
```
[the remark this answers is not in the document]
clebertsuconic: There's no remove... The add is always single threaded.

Each Message has its own Reference. I could just use an user property on the message.. but it would have the same effect honestly.

clebertsuconic:
I just squashed the docs as part of the first commit. I made some mistakes on the amends, it would be just easier to keep a single commit :)

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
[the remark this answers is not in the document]
clebertsuconic: I don't think it will make much difference, but I will use setUseContext on the message command stored in the Operation.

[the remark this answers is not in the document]
gemmellr: So, no removals  as the entire thing is discarded along with the parent tx object, ok. (Oops, I hadnt got to looking at the rest when commenting that, didnt go back hehe),

If the 'UserContext' from the message is already by something else, then there likely isnt much difference as you say, though it will need to do a lookup (I was thinking a map here, it could foreach). If it isnt already in use, then starting to use it would be worse than the original Pair's.

So yeah...maybe just put back the annoying Pair. It only applies in the Sync case, right?

artemis-server/src/main/resources/schema/artemis-configuration.xsd
gemmellr: Is there a simple test for this type of default that could be updated to check this value? I notice no test needed to change when you updated it, which probably explains how it came to be wrong without notice to start with.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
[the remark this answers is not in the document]
clebertsuconic: I had made a comment last week and I pressed start a review by accident (weird), and all my comments were lost.. 

I actually mentioned I didn't think it would make much difference... and I just realized what you meant on the Map.

I'm using a Map on a separate commit. If you could take a look and I will squash it after you look.

I don't think it will make much difference on the runtime... although the code is cleaner this way.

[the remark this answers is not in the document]
clebertsuconic: @gemmellr I will just squash the separate commit. It looks better actually.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
clebertsuconic: the default on parsing was actually correct... just the default value on the XSD was wrong.

clebertsuconic: there was another test validating the default on broker properties.. and I just added a single assert on the next mirror option that wasn't touched on the XML

I think we're good

clebertsuconic: I am marking it resolved.. please unresolved it if you still see anything else.

PR #4340, opened 2023-01-19
ARTEMIS-4137 MQTT sub-queue clean-up can fail due to auth

jbertram:


artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSubscriptionManager.java
```diff
@@ -369,7 +373,7 @@ Map<Long, Integer> getConsumerQoSLevels() {
 
    void clean() {
       for (MqttTopicSubscription mqttTopicSubscription : session.getState().getSubscriptions()) {
-         removeSubscription(mqttTopicSubscription.topicName());
+         removeSubscription(mqttTopicSubscription.topicName(), true);
```
brusdev, line 376: The `testSubscriptionQueueRemoved` test is failing, should enforceSecurity=false in this case?
```suggestion
         removeSubscription(mqttTopicSubscription.topicName(), false);
```

jbertram: Yes! I changed that for a quick test and forgot to change it back.

brusdev:
I was wondering if this fix could open a security hole. Could this fix be used to delete queues without the required permissions? Should we check the permission when the subscription is created?

jbertram:
@brusdev, there may be one potential use-case that's problematic. I'll push a fix for that.

We already check permissions when the subscription is created.

brusdev (approved):

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -1178,7 +1183,7 @@ public void deleteQueue(final SimpleString queueToDelete) throws Exception {
          throw new ActiveMQNonExistentQueueException();
       }
 
-      server.destroyQueue(unPrefixedQueueName, this, true, false, true);
+      server.destroyQueue(unPrefixedQueueName, enforceSecurity ? this : null, true, false, true);
```
alizard0, line 1186: What does `null` do here? It is not very clear if you enforcing something that previously was enforce always. @jbertram

jbertram: If `enforceSecurity` is `true` then `this` is passed and security will be enforced. Otherwise `null` is passed and security is not enforced. Previously security was always enforced which would prevent the broker from cleaning up the subscription queue in certain situations.

PR #4342, opened 2023-01-20
ARTEMIS-4132 AMQP Receiver default to ANYCAST when creating an address

tabish121:
When an AMQP client subscribes to a new address (non-existing) with a receiver link, the address is created with routing type ANYCAST regardles of the default address creation configuration of the broker, and ignores even the broker wide default of MULTICAST.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
@@ -25,6 +25,8 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.PooledByteBufAllocator;
+
+import org.apache.activemq.artemis.api.config.ActiveMQDefaultConfiguration;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
 import org.apache.activemq.artemis.api.core.ActiveMQExceptionType;
 import org.apache.activemq.artemis.api.core.ActiveMQQueueMaxConsumerLimitReached;
@@ -960,7 +962,6 @@ public class ProtonServerSenderContext extends ProtonInitializable implements Pr
       SimpleString tempQueueName;
       String selector;
 
-      private final RoutingType defaultRoutingType = RoutingType.ANYCAST;
       private RoutingType routingTypeToUse = RoutingType.ANYCAST;
 
       private boolean isVolatile = false;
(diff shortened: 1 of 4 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AutoCreateWithDefaultRoutingTypesTest.java
```diff
@@ -0,0 +1,205 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
gemmellr, line 10: Incorrect formatting, tags and indent. Presumably copied, original to be fixed also. I recently fixed all the ones that were javadoc headers, but guess there are more that were already actually comment headers.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AutoCreateWithDefaultRoutingTypesTest.java
```diff
@@ -0,0 +1,205 @@
...
+         map.put("#", as);
+      }
+      Map.Entry<String, AddressSettings> entry = map.entrySet().iterator().next();
+      AddressSettings settings = entry.getValue();
+      settings.setAutoCreateQueues(true);
+      settings.setDefaultAddressRoutingType(routingType);
+      settings.setDefaultQueueRoutingType(routingType);
+      logger.info("server cofg, isauto? {}", entry.getValue().isAutoCreateQueues());
+      logger.info("server cofg, default queue routing type? {}", entry.getValue().getDefaultQueueRoutingType());
+      logger.info("server cofg, default address routing type? {}", entry.getValue().getDefaultAddressRoutingType());
```
gemmellr, line 89: cofg = config?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
```diff
@@ -1115,8 +1116,15 @@ public Consumer init(ProtonServerSenderContext senderContext) throws Exception {
             } else {
                // if not we look up the address
                AddressQueryResult addressQueryResult = null;
+
+               // Set this to the broker configured default for the address prior to the lookup so that
+               // an auto create will actually use the configured defaults.  The actual query result will
+               // contain the true answer on what routing type the address actually has though.
+               routingTypeToUse = sessionSPI.getDefaultRoutingType(addressToUse);
+               routingTypeToUse = routingTypeToUse == null ? ActiveMQDefaultConfiguration.getDefaultRoutingType() : routingTypeToUse;
```
gemmellr, line 1124: I think this one might be more obvious with a simple if doing a potential update, rather than using the ternary and setting itself to itself in many/most cases.

gemmellr (commented):
Though I'd really prefer things actually defaulted to anycast as this effectively does for receivers, the described 'different autocreate behaviour if creating sender first' is clearly broken, and the expectation was the address/routing config controls it so making it use that does seem correct.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
tabish121: Agree to disagree

gemmellr:
LGTM...started a full test run on it.

PR #4344, opened 2023-01-26
ARTEMIS-4143 improve mitigation against split-brain with shared-storage

jbertram:
Configurations employing shared-storage with NFS are susceptible to split-brain in certain scenarios. For example:

  1) Primary loses network connection to NFS.
  2) Backup activates.
  3) Primary reconnects to NFS.
  4) Split-brain.

In reality this situation is pretty unlikely due to the timing involved, but the possibility still exists. Currently the file lock held by the primary broker on the NFS share is essentially worthless in this situation. This commit adds logic by which the timestamp of the lock file is updated during activation and then routinely checked during runtime to ensure consistency. This effectively mitigates split-brain in this situation (and likely others). Here's how it works now:

  1) Primary loses network connection to NFS.
  2) Backup activates.
  3) Primary reconnects to NFS.
  4) Primary detects that the lock file's timestamp has been updated and
     shuts itself down.

When the primary shuts down in step #4 the Topology on the backup can be damaged. Protections were added for this via ARTEMIS-2868 but only for the replicated use-case. This commit applies the protection for removeMember() so that the Topology remains intact.

There are no tests for these changes as I cannot determine how to properly simulate this use-case. However, there have never been robust, automated tests for these kinds of NFS use-cases so this is not a departure from the norm.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
@@ -22,6 +22,7 @@ import java.io.RandomAccessFile;
 import java.nio.ByteBuffer;
 import java.nio.channels.FileChannel;
 import java.nio.channels.FileLock;
+import java.util.Date;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
@@ -45,8 +46,6 @@ public class FileLockNodeManager extends FileBasedNodeManager {
 
    private static final int BACKUP_LOCK_POS = 2;
 
-   private static final long LOCK_LENGTH = 1;
-
    private static final byte LIVE = 'L';
 
    private static final byte FAILINGBACK = 'F';
@@ -64,6 +63,8 @@ public class FileLockNodeManager extends FileBasedNodeManager {
    private FileLock backupLock;
 
    private final FileChannel[] lockChannels = new FileChannel[3];
+
+   private long serverLockLastModified;
 
    private final long lockAcquisitionTimeoutNanos;
 
@@ -110,6 +111,7 @@ public class FileLockNodeManager extends FileBasedNodeManager {
       super.setUpServerLockFile();
 
       lockChannels[0] = channel;
+      serverLockLastModified = serverLockFile.lastModified();
 
       for (int i = 1; i < 3; i++) {
          if (lockChannels[i] != null && lockChannels[i].isOpen()) {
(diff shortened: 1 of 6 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -213,9 +217,9 @@ public void awaitLiveNode() throws NodeManagerException, InterruptedException {
                logger.debug("awaiting live node failing back");
                Thread.sleep(2000);
             } else if (state == FileLockNodeManager.LIVE) {
-               if (logger.isDebugEnabled()) {
-                  logger.debug("acquired live node lock state = {}", (char) state);
-               }
+               logger.debug("acquired live node lock state = {}", (char) state);
+               serverLockFile.setLastModified(System.currentTimeMillis());
+               logger.debug("touched " + serverLockFile.getAbsoluteFile() + "; new time: " + serverLockFile.lastModified());
```
tabish121, line 222: Logger should be using formatting as to string concatenations.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -345,6 +349,8 @@ private void writeFileLockStatus(byte status) throws NodeManagerException {
                lock.release();
             }
          }
+         serverLockLastModified = serverLockFile.lastModified();
+         logger.debug("Modified " + serverLockFile.getName() + " at " + serverLockLastModified);
```
tabish121, line 353: Fix logger formatting as above

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -429,16 +435,19 @@ protected FileLock lock(final int lockPosition) throws ActiveMQLockAcquisitionTi
             FileLock lock = tryLock(lockPosition);
             isRecurringFailure = false;
 
+            logger.debug("lock: " + lock);
```
tabish121, line 438: Fix logger formatting as above

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -554,9 +572,26 @@ public void run() {
...
+      private boolean hasBeenModified() {
+         boolean modified = false;
+
+         // Create a new instance of the File object so we can get the most up-to-date information on the file.
+         File freshServerLockFile = new File(serverLockFile.getAbsolutePath());
 
+         if (freshServerLockFile.exists()) {
+            logger.debug(freshServerLockFile.lastModified() + " > " + serverLockLastModified + "?");
+            if (freshServerLockFile.lastModified() > serverLockLastModified) {
+               logger.debug("Lock file " + serverLockFile.getAbsolutePath() + " originally locked at " + new Date(serverLockLastModified) + " was modified at " + new Date(freshServerLockFile.lastModified()));
```
tabish121, line 586: Fix logger formatting as above

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/FileLockNodeManager.java
```diff
@@ -554,9 +572,26 @@ public void run() {
...
+               modified = true;
+            }
+         } else {
+            logger.debug("Lock file " + serverLockFile.getAbsolutePath() + " does not exist");
```
tabish121, line 590: Fix logger formatting as above

clebertsuconic:
@jbertram since I have seen the failures associated with this PR, I am making it a draft. Please make it ready to review after fixed?

PR #4346, opened 2023-01-27
ARTEMIS-4145 MQTT shared sub queue may be inadvertently removed

jbertram:
o.a.a.a.c.p.m.MQTTSubscriptionManager#removeSubscription() had a chunk of code from 971f673c602f859f342e22afe988c71687f754b6 removed. I couldn't determine why that bit of code was actually added in the first place, but the tests from that commit all still pass even with it removed now (as well as all the other MQTT tests) so I think it's safe.


gemmellr:
The last point of the [original changes description](https://github.com/apache/activemq-artemis/pull/2466) seems to be about why that code was added:
>3.  For the MQTT protocol, there is one and only one consumer connection per queue, which is a good choice for closing the old MQTT consumer before the new MQTT consumer connects.
The original code could not effectively clean up the 'old consumer' in the queue when the 'new MQTT connection' was connected to the Artemis broker. Modify ‘MQTTSubscriptionManager.removeSubscription’ to get the queue consumer collection from the ‘QueueImpl’ instance and close them.

jbertram:
@gemmellr, thanks for the clarification. I'll update the commit message based on this.

PR #4349, opened 2023-01-27
ARTEMIS-3178 Page Limitting (max messages and max bytes)

clebertsuconic:
I am adding three attributes to Address-settings:

* page-limit-bytes: Number of bytes. We will convert this metric into max number of pages internally by dividing max-bytes / page-size. It will allow a max based on an estimate.
* page-limit-messages: Number of messages
* page-full-message-policy: fail or drop

We will now allow paging, until these max values and then fail or drop messages.

Once these values are retracted, the address will remain full until a period where cleanup is kicked in by paging. So these values may have a certain delay on being applied, but they should always be cleared once cleanup happened.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/PagingStore.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/PagingStore.java
@@ -23,6 +23,7 @@ import org.apache.activemq.artemis.api.core.Message;
 import org.apache.activemq.artemis.api.core.RefCountMessageListener;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.paging.cursor.PageCursorProvider;
+import org.apache.activemq.artemis.core.paging.cursor.PageSubscription;
 import org.apache.activemq.artemis.core.paging.impl.Page;
 import org.apache.activemq.artemis.core.persistence.StorageManager;
 import org.apache.activemq.artemis.core.replication.ReplicationManager;
@@ -30,6 +31,7 @@ import org.apache.activemq.artemis.core.server.ActiveMQComponent;
 import org.apache.activemq.artemis.core.server.RouteContextList;
 import org.apache.activemq.artemis.core.settings.impl.AddressFullMessagePolicy;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
+import org.apache.activemq.artemis.core.settings.impl.PageFullMessagePolicy;
 import org.apache.activemq.artemis.core.transaction.Transaction;
 import org.apache.activemq.artemis.utils.actors.ArtemisExecutor;
 
@@ -59,6 +61,19 @@ public interface PagingStore extends ActiveMQComponent, RefCountMessageListener 
    File getFolder();
 
    AddressFullMessagePolicy getAddressFullMessagePolicy();
+
+   PageFullMessagePolicy getPageFullMessagePolicy();
+
+   Long getPageLimitMessages();
+
+   Long getPageLimitBytes();
+
+   /** Callback to be used by a counter when the Page is full for that counter */
+   void pageFull(PageSubscription subscription);
+
+   boolean isPageFull();
+
+   void checkPageLimit(long numberOfMessages);
 
    long getFirstPage();
 
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
@@ -19,6 +19,7 @@ package org.apache.activemq.artemis.core.paging.cursor.impl;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
@@ -37,12 +38,14 @@ import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
 import org.apache.activemq.artemis.core.transaction.Transaction;
 import org.apache.activemq.artemis.core.transaction.impl.TransactionImpl;
 import org.apache.activemq.artemis.utils.ArtemisCloseable;
+import org.apache.activemq.artemis.utils.SimpleFutureImpl;
 import org.apache.activemq.artemis.utils.collections.ConcurrentLongHashMap;
 import org.apache.activemq.artemis.utils.collections.LinkedList;
 import org.apache.activemq.artemis.utils.collections.LongHashSet;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.lang.invoke.MethodHandles;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Consumer;
 
 public class PageCursorProviderImpl implements PageCursorProvider {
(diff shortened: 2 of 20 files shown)


clebertsuconic:
I need to finish docs, and add testing for dropping.. which I should finish Monday morning.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
```diff
@@ -225,13 +237,100 @@ public void applySetting(final AddressSettings addressSettings) {
       addressFullMessagePolicy = addressSettings.getAddressFullMessagePolicy();
 
       rejectThreshold = addressSettings.getMaxSizeBytesRejectThreshold();
+
+      pageFullMessagePolicy = addressSettings.getPageFullMessagePolicy();
+
+      pageLimitBytes = addressSettings.getPageLimitBytes();
+
+      pageLimitMessages = addressSettings.getPageLimitMessages();
```
gemmellr, line 245: Should it validate you have actually set a policy if setting one or both of the others to enforcing values? That way you cant accidentally think you have a limit when you dont.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
```diff
@@ -179,11 +182,13 @@ public void close(PageSubscription cursor) {
...
 
    @Override
-   public void scheduleCleanup() {
+   public Future<Boolean> scheduleCleanup() {
+      final SimpleFutureImpl<Boolean> future = new SimpleFutureImpl<>();
       if (!cleanupEnabled || scheduledCleanup.intValue() > 2) {
          // Scheduled cleanup was already scheduled before.. never mind!
          // or we have cleanup disabled
-         return;
+         pagingStore.execute(() -> future.set(true));
```
tabish121, line 190: Is there a reason why you need to fire off a task into the executor just to return a completed future?  Seems simpler to just return CompletableFuture.completedFuture or just complete this one and return it directly.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
```diff
@@ -1029,6 +1131,25 @@ public boolean page(Message message,
          lock.readLock().unlock();
       }
 
+      if (pageFull && pageFullMessagePolicy != null) {
```
gemmellr, line 1134: Related to earlier comment, if it validates you have a policy, can this then just simply check "if (pageFull)" ?

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -4062,6 +4079,20 @@
                </xsd:simpleType>
             </xsd:element>
 
+            <xsd:element name="page-full-policy" maxOccurs="1" minOccurs="0">
+               <xsd:annotation>
+                  <xsd:documentation>
+                     After entering page mode, a second limit will be set by page-limit-bytes, page-limit-messages. page-full-policy will configure what to do when that limit is reach.
```
gemmellr, line 4085: page-limit-bytes and/or page-limit-messages. The page-full-policy...

...when that limit is reach**ed.**

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCursorProviderImpl.java
clebertsuconic: Yes... I wanted to make sure any previous call to cleanup was finished. Say you called scheduleCleanup a couple of times.. then you call scheduleCleanup again...


 I don't need to schedule another one as I already have a call scheduled, but I want to return a Future that will be released after these calls have finished.

clebertsuconic: I am amending the comment before future.set call as:

         // Scheduled cleanup was already scheduled before.
         // On that case just flush the executor returning the future.set(true) 
         // after any previous scheduled cleanup is finished.

clebertsuconic: marking it as resolved.. please unresolved it if you find any issues with my comments please?

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/PagingStore.java
```diff
@@ -60,6 +62,19 @@ public interface PagingStore extends ActiveMQComponent, RefCountMessageListener
 
    AddressFullMessagePolicy getAddressFullMessagePolicy();
 
+   PageFullMessagePolicy getPageFullMessagePolicy();
+
+   Long getPageLimitMessages();
```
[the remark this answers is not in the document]
clebertsuconic: I was going against the -1 usage here. and just keep it as null.

I can edit the default on the XSD to state null

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -4000,6 +4000,23 @@
                </xsd:annotation>
             </xsd:element>
 
+            <xsd:element name="page-limit-bytes" type="xsd:string" default="-1" maxOccurs="1" minOccurs="0">
+               <xsd:annotation>
+                  <xsd:documentation>
+                     After the address entering into page mode, this attribute will configure how many pages can be written into page before activating the page-full-policy.
```
tabish121, line 4006: This reads a bit funny, should it be "After the address enters into page mode...."  same for the below comment.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
```diff
@@ -225,13 +237,100 @@ public void applySetting(final AddressSettings addressSettings) {
...
+      pageFullMessagePolicy = addressSettings.getPageFullMessagePolicy();
+
+      pageLimitBytes = addressSettings.getPageLimitBytes();
+
+      pageLimitMessages = addressSettings.getPageLimitMessages();
+
+      if (pageLimitBytes != null && pageSize > 0) {
+         estimatedMaxPages = pageLimitBytes / pageSize;
+         logger.debug("Address {} should not allow more than {} pages", storeName, estimatedMaxPages);
+      }
```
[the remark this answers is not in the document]
clebertsuconic: I'm changing the constructor to make any -1 as null.


I won't just change the File parser as the same could be set through properties.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
clebertsuconic: added validation

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
clebertsuconic: Probably. It feels wrong not doing the redundant check though! :)

clebertsuconic: My OCD wants to keep it, but I will remove it :)

clebertsuconic:
all the comments are resolved.. tests are 100%... I'm merging this...


if anyone sees anything else wrong I will send a new PR.. (or feel free to send any adjustments yourself of course).

clebertsuconic:
thanks for the reviews!

PR #4354, opened 2023-01-31
ARTEMIS-4151 tighten default MBean access

jbertram:
This is a preventative measure to limit what users can do with any MBean other than those in the hawtio or org.apache.activemq.artemis domains.


clebertsuconic:
@jbertram can you rebase and run the whole test suite please?

jbertram:
Full test-suite passed.

PR #4359, opened 2023-02-03
ARTEMIS-4155 Fixing deadlock on LargeMessage conversion and retention

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -878,7 +878,7 @@ public final void addBytesToLargeMessage(final SequentialFile file,
    private void historyBody(long messageId, EncodingSupport partialBuffer) {
 
       try {
-         messageJournal.appendAddEvent(messageId, JournalRecordIds.ADD_MESSAGE_BODY, EncoderPersister.getInstance(), partialBuffer, true, null);
+         messageJournal.appendAddEvent(messageId, JournalRecordIds.ADD_MESSAGE_BODY, EncoderPersister.getInstance(), partialBuffer, false, null);
```
clebertsuconic, line 881: lol... this is "the fix"

clebertsuconic:
I am adding actual replay to the tests. To make sure the large message conversion works through retention. 


Will do that later between now and Monday ...

clebertsuconic:
I added the replay test as I mentioned. and I also moved the test into soak-tests (from smoke tests where I originally placed it, which is a better place).

PR #4360, opened 2023-02-07
ARTEMIS-4157: Fix error setting broker properties for AddressSettings

DreamPearl:
Added the test for all the possible AddressSettings and fixed the exception throw because of the type mismatch between the read and write methods.


jbertram:
@gtully, I'm not 100% sure, but my guess is that the Objects were used to detect whether or not the value had been set which is important since address settings are hierarchical and explicit settings take precedence over default ones.

gtully (approved):

PR #4362, opened 2023-02-08
ARTEMIS-4160 Handle hostnames with invalid char sequences.

davidlanouette:
The hostname gets put into an xml comment in the jolokia-access.xml file.  If the hostname has invalid char sequences for an xml comment, jolokia will fail at load time because the xml file is invalid.

Signed-off-by: David Lanouette <David.Lanouette@RedHat.com>

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
@@ -395,8 +395,20 @@ public class Create extends InstallAbstract {
       this.etc = etc;
    }
 
+   protected void setNoAutoTune(boolean autTune) {
+      this.noAutoTune = autTune;
+   }
+
    public boolean isAutoDelete() {
       return autoDelete;
+   }
+
+   protected void setHttpHost(String httpHost) {
+      this.httpHost = httpHost;
+   }
+
+   protected void setRelaxJolokia(boolean relaxJolokia) {
+      this.relaxJolokia = relaxJolokia;
    }
 
    public Create setAutoDelete(boolean autoDelete) {
--- a/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/jolokia-access.xml
+++ b/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/jolokia-access.xml
@@ -23,7 +23,7 @@ under the License.
 <restrict>
 
     <cors>
-        <!-- Allow cross origin access from ${http.host} ... -->
+        <!-- Allow cross-origin access from the origins that match the following pattern ... -->
         <allow-origin>*://${http.host}*</allow-origin>
 
 
--- a/artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
+++ b/artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.cli.commands;
+
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.ParserConfigurationException;
+import java.io.File;
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collection;
+
+import org.apache.activemq.artemis.utils.XmlProvider;
+import org.apache.activemq.cli.test.CliTestBase;
+import org.apache.activemq.cli.test.TestActionContext;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.xml.sax.SAXException;
+
+@RunWith(Parameterized.class)
+public class CreateTest extends CliTestBase {
+
+   private final String testName;
+   private String httpHost;
+   private boolean relaxJolokia;
+
+   public CreateTest(String testName, String httpHost, boolean relaxJolokia) throws IOException {
+      this.testName = testName;
+      this.httpHost = httpHost;
+      this.relaxJolokia = relaxJolokia;
+   }
+
+   @Parameterized.Parameters(name = "{0}")
+   public static Collection<Object[]> testData() {
+      return Arrays.asList(new Object[][]{
+         {"Happy path + relaxJolokia", "sampledomain.com", true},
+         {"Happy path - relaxJolokia", "sampledomain.net", false},
+         {"Domain with dash + relaxJolokia", "sample-domain.co", true},
+         {"Domain with dash - relaxJolokia", "sample-domain.co.uk", false},
+         {"Domain with double dashes + relaxJolokia", "sample--domain.name", true},
+         {"Domain with double dashes - relaxJolokia", "sample--domain.biz", false},
+         {"Domain with leading dashes + relaxJolokia", "--sampledomain.company", true},
+         {"Domain with leading dashes - relaxJolokia", "--sampledomain.email", false},
+         {"Domain with trailing dashes + relaxJolokia", "sampledomain--.shop", true},
+         {"Domain with trailing dashes - relaxJolokia", "sampledomain--.java", false},
+      });
+   }
+
+   @Test
+   public void testWriteJolokiaAccessXmlCreatesValidXml() throws Exception {
+      TestActionContext context = new TestActionContext();
+      File testInstance = new File(temporaryFolder.getRoot(), "test-instance");
+
+      Create c = new Create();
+      c.setNoAutoTune(true);
+      c.setInstance(testInstance);
+      c.setHttpHost(httpHost);
+      c.setRelaxJolokia(relaxJolokia);
+      c.execute(context);
+
+      Assert.assertTrue(isXmlValid(new File(testInstance, "etc/" + Create.ETC_JOLOKIA_ACCESS_XML)));
+   }
+
+   /**
+    * IsXmlValid will check if a given xml file is valid by parsing the xml to create a Document.
+    * <p>
+    * If it parses, the xml is assumed to be valid. If any exceptions occur, the xml is not valid.
+    *
+    * @param xml The xml file to check for validity.
+    * @return whether the xml file represents a valid xml document.
+    */
+   private boolean isXmlValid(File xml) {
+      try {
+         DocumentBuilder dbuilder = XmlProvider.newDocumentBuilder();
+         dbuilder.parse(xml);
+
+      } catch (ParserConfigurationException e) {
+         return false;
+      } catch (IOException e) {
+         return false;
+      } catch (SAXException e) {
+         return false;
+      }
+      return true;
+   }
+}


davidlanouette:
Hi all.  This is my first PR for this project, and I didn't find anything that talked about needing a Bugzilla bug before submitting.

---

We ran into this issue.  We have hosts with "--" in the name, and that ended up in the comment in the jolokia-access.xml file as a comment.  That's an invalid char sequence for an xml comment.

I've tried to handle other illegal char sequences too.

I hope this PR is useful.

jbertram:
@davidlanouette, thanks for the contribution! Most changes will, in fact, need an associated Jira (we don't use Bugzilla). See more details in our [Hacking Guide](https://github.com/apache/activemq-artemis/blob/main/docs/hacking-guide/en/code.md#typical-development-cycle). I went ahead and created [one for you](https://issues.apache.org/jira/browse/ARTEMIS-4160).

davidlanouette:
@jbertram Do I need to get a jira login and update the bug before it gets reviewed?

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -1115,6 +1116,26 @@ private void writeEtc(String source, File etcFolder, HashMap<String, String> fil
...
+    *
+    * XML comments cannot have "--" inside them, and they cannot end with "-"
+    * ref: https://www.w3.org/TR/2008/REC-xml-20081126/#sec-comments
+    *
+    * @param text The text to format
+    * @return text that is safe to put into an xml comment.
+    */
+   protected String safeForXmlComment(String text) {
+
+      String newText = text.replaceAll("---", "-").replaceAll("--", "-");
```
brusdev, line 1132: safeForXmlComment doesn't work with all texts i.e. `safeForXmlComment("-------")` should return `--`

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/jolokia-access.xml
```diff
@@ -23,7 +23,7 @@ under the License.
 <restrict>
 
     <cors>
-        <!-- Allow cross origin access from ${http.host} ... -->
+        <!-- Allow cross origin access from ${http.host-for-comment} ... -->
```
brusdev, line 26: ncluding in a comment a hostname that doesn't match the real hostname is confusing, what about just removing `${http.host}` from the comment?
```suggestion
        <!-- Allow cross origin access from the origins that match the following pattern -->
```
I

davidlanouette: That is a much simpler solution if nobody is tied to having the hostname in that comment.

I'll adjust (remove the other changes, and update this comment, then update the PR.

jbertram: Removing the comment altogether is the best solution in my opinion.

davidlanouette: I've reverted my previous commit, and updated the comment in `jolokia-access.xml` so that it has no substitution.  See what you all think.

brusdev: A test would be good to avoid regressions.
Can you squash your commits and prepend the commit message with `ARTEMIS-4160`, i.e. 
```
ARTEMIS-4160 Handle hostnames with double dashes
```

brusdev:
@gtully this is the same issue we have seen ArtemisCloud.io test suite

davidlanouette:
> this is the same issue we have seen ArtemisCloud.io test suite
That is where we found this originally.  The pod name had "--" in the middle of the name.  We ended up needing to add an init container with a simple script that modified the `jolokia-access.xml` file.

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/jolokia-access.xml
davidlanouette: Squished and updated the commit message!

brusdev:
@davidlanouette your previous test would be good to avoid future regressions, could you add it back?

davidlanouette:
> @davidlanouette your previous test would be good to avoid future regressions, could you add it back?

Yes, I'll add it back, but with some tweaks to test the whole file.

davidlanouette:
@jbertram @brusdev 

Can I get someone to take a look at this PR again?

I've added a test that will create the full jolokia-access.xml file, then test that it's valid xml.

I'll likely do the same for the other xml files, but in another PR.

artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
```diff
@@ -0,0 +1,123 @@
...
+   *
+   * @param xml The xml string to check for validity.
+   * @return whether the xml string represents a valid xml document.
+   */
+  private boolean isXmlValid(String xml) {
+    try {
+      var xmlStream = new ByteArrayInputStream(xml.getBytes());
+
+      DocumentBuilder dbuilder = dbFactory.newDocumentBuilder();
+      Document doc = dbuilder.parse(xmlStream);
```
brusdev, line 112:
```suggestion
      DocumentBuilder dbuilder = XmlProvider.newDocumentBuilder();
      dbuilder.parse(xmlStream);
```

artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
```diff
@@ -0,0 +1,123 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import javax.xml.parsers.DocumentBuilder;
+import javax.xml.parsers.DocumentBuilderFactory;
```
brusdev, line 28:
```suggestion
import org.apache.activemq.artemis.utils.XmlProvider;
```

artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
```diff
@@ -0,0 +1,123 @@
...
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+import org.w3c.dom.Document;
+import org.xml.sax.SAXException;
+
+@RunWith(Parameterized.class)
+public class CreateTest {
+
+  private final String testName;
+  private final DocumentBuilderFactory dbFactory = DocumentBuilderFactory.newInstance();
```
brusdev, line 42:
```suggestion
```

brusdev (commented):
The test shoud use org.apache.activemq.artemis.utils.XmlProvider to create a  DocumentBuilder

brusdev:
There are `57 Checkstyle violations`, you can check violations executing `mvn clean install -Pdev -DskipTests`

davidlanouette:
@brusdev Thanks for the feedback.  I'll address the xml DocumentBuilder and the checkstyle violations.

davidlanouette:
@brusdev all issues you pointed out have been resolved.  

I think the checks are waiting for someone to approve them.  The checks ui shows "This workflow is awaiting approval from a maintainer in https://github.com/apache/activemq-artemis/pull/4362"

artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
```diff
@@ -0,0 +1,119 @@
...
+   private String httpHost;
+   private boolean relaxJolokia;
+
+   public CreateTest(String testName, String httpHost, boolean relaxJolokia) {
+      this.testName = testName;
+      this.httpHost = httpHost;
+      this.relaxJolokia = relaxJolokia;
+   }
+
+   @Parameters
```
brusdev, line 52:
```suggestion
   @Parameterized.Parameters(name = "{0}")
```
The test name could be used to improve test description

artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
```diff
@@ -0,0 +1,119 @@
...
+      // TODO: Refactor that code to it's own method.
+      if (this.relaxJolokia) {
+         filters.put("${jolokia.options}", "<!-- option relax-jolokia used, so strict-checking will be removed here -->");
+      } else {
+         filters.put("${jolokia.options}", "<!-- Check for the proper origin on the server side, too -->\n" + "        <strict-checking/>");
+      }
+
+      Path temp = Files.createTempFile("jolokia-access-", ".xml");
+      try {
+         c.write("etc/" + source, temp.toFile(), filters, false, true);
```
brusdev, line 86: The test could avoid to duplicate embedded Create class stuff exposing setNoAutoTune, setHttpHost and setRelaxJolokia:
```suggestion
      TestActionContext context = new TestActionContext();
      File testInstance = new File(temporaryFolder.getRoot(), "test-instance");

      Create c = new Create();
      c.setNoAutoTune(true);
      c.setInstance(testInstance);
      c.setHttpHost(httpHost);
      c.setRelaxJolokia(relaxJolokia);
      c.execute(context);
```

artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+         String xml = Files.readString(temp);
+         Assert.assertTrue(testName + " - should be valid, but isn't", isXmlValid(xml));
+      } finally {
+         Files.delete(temp);
```
brusdev, line 91: If CreateTest inherits `temporaryFolder` from CliTestBase doesn't need to delete temp

artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
```diff
@@ -0,0 +1,119 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+import org.w3c.dom.Document;
+import org.xml.sax.SAXException;
+
+@RunWith(Parameterized.class)
+public class CreateTest {
```
brusdev, line 40: CreateTest extend CliTestBase to inherit `temporaryFolder`
```suggestion
public class CreateTest extends CliTestBase {
```

artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
```diff
@@ -0,0 +1,119 @@
...
+      } else {
+         filters.put("${jolokia.options}", "<!-- Check for the proper origin on the server side, too -->\n" + "        <strict-checking/>");
+      }
+
+      Path temp = Files.createTempFile("jolokia-access-", ".xml");
+      try {
+         c.write("etc/" + source, temp.toFile(), filters, false, true);
+
+         String xml = Files.readString(temp);
+         Assert.assertTrue(testName + " - should be valid, but isn't", isXmlValid(xml));
```
brusdev, line 89: The xml check should not hide the failure cause
```suggestion
      try {
         XmlProvider.newDocumentBuilder().parse(new File(testInstance, "etc/" + Create.ETC_JOLOKIA_ACCESS_XML));
      } catch (Exception e) {
         Assert.fail(testName + " should be valid, but " + e.toString());
      }
```

brusdev (commented):
Some suggestions to remove the `TODO: Refactor that code to it's own method.` comment

artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
davidlanouette: This has been removed.

artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+   /**
+    * IsXmlValid will check if a given xml file is valid by parsing the xml to create a Document.
+    * <p>
+    * If it parses, the xml is assumed to be valid. If any exceptions occur, the xml is not valid.
+    *
+    * @param xml The xml file to check for validity.
+    * @return whether the xml file represents a valid xml document.
+    */
+   private boolean isXmlValid(File xml) {
```
davidlanouette, line 91: Note: I left this as a separate method to make it more explicit what was being done.

davidlanouette:
@brusdev thanks for the feedback.  I've incorporated it with the latest commit.

artemis-cli/src/test/java/org/apache/activemq/artemis/cli/commands/CreateTest.java
```diff
@@ -0,0 +1,105 @@
...
+      c.setNoAutoTune(true);
+      c.setInstance(testInstance);
+      c.setHttpHost(httpHost);
+      c.setRelaxJolokia(relaxJolokia);
+      c.execute(context);
+
+      try {
+         isXmlValid(new File(testInstance, "etc/" + Create.ETC_JOLOKIA_ACCESS_XML));
+      } catch (Exception e) {
+         Assert.fail(testName + " should be valid, but " + e.toString());
```
brusdev, line 79: The `isXmlValid` method is catching the exceptions and converting them in a boolean but the returned boolean isn't checked.

You could check the returned boolean or remove the try/catch from the `isXmlValid` method.

I would remove the try/catch from the `isXmlValid` method to print the cause exception when it fails.

davidlanouette: You are correct.  I missed that when adding your suggestions.

I'll change this to `Assert.true(isXmlValid(new File(testInstance, "etc/" + Create.ETC_JOLOKIA_ACCESS_XML)));`, and remove the try/catch.

davidlanouette:
@brusdev Thanks for all the help getting my first PR for this project complete.  @jbertram Thanks for your help too.

brusdev:
@davidlanouette thanks for your contribution.

PR #4363, opened 2023-02-09
ARTEMIS-4161 Implementing Memory Leak tests and fixing a few initial small leaks found

clebertsuconic:


clebertsuconic:
There are test failures...

tests/leak-tests/pom.xml
```diff
@@ -0,0 +1,488 @@
...
+            <filtering>true</filtering>
+         </testResource>
+      </testResources>
+      <plugins>
+         <plugin>
+            <!-- exposed through jitpack.io.
+                 It will install the proper libdleak.dll required for the current platform. -->
+            <groupId>com.github.d-leak.d-leak</groupId>
+            <artifactId>maven-plugin</artifactId>
+            <version>0.2</version>
```
gemmellr, line 413: Would be nicer for version to be in property...if not plugin-managed and not defined here at all.

tests/leak-tests/pom.xml
```diff
@@ -0,0 +1,488 @@
...
+         <id>jitpack.io</id>
+         <url>https://jitpack.io</url>
+      </repository>
+   </repositories>
+   <pluginRepositories>
+      <pluginRepository>
+         <id>jitpack.io</id>
+         <url>https://jitpack.io</url>
+      </pluginRepository>
+   </pluginRepositories>
```
gemmellr, line 486: Would much prefer to see this deployed to central than relying on an 'automagically built code' third party repo.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -4218,7 +4219,7 @@ public QueueConfiguration getQueueConfiguration() {
          .setAutoCreated(autoCreated);
    }
 
-   protected static class ConsumerHolder<T extends Consumer> implements PriorityAware {
+   protected class ConsumerHolder<T extends Consumer> implements PriorityAware {
```
gemmellr, line 4222: Rather than make it non-static and add a new reference to the parent queue object, it seems like all the Consumer impls passsed to addConsumer already have getQueue(). Could that be promoted to the interface and used instead to get the name for the toString?

tests/leak-tests/pom.xml
```diff
@@ -0,0 +1,488 @@
+<!--
```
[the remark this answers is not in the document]
clebertsuconic: I will remove the test-jar


My intention was to be equivalent to integration-tests.. that's why I copied the dependencies here.

tests/leak-tests/pom.xml
clebertsuconic: agreed.. I will try to do it.

tests/leak-tests/pom.xml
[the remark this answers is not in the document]
gemmellr: It seems unlikely folks will keep the two places in line going forward, so I dont see much reason for them to start that way when most of the dependencies arent even used here. We should just add dependencies as they are actually needed and keep the tests closer to the way things are actually used in practice in e.g the broker distribution.

tests/leak-tests/pom.xml
clebertsuconic: this is done...

tests/leak-tests/pom.xml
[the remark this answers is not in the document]
gemmellr: Unresolving comment since the test jar is still being created. Plus the 'entire pom is mostly unused stuff' aspect also still stands.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
```diff
@@ -86,11 +86,21 @@ public boolean add(T t) {
...
    @Override
    public boolean remove(T t) {
+      iterator.removed(t);
       boolean result = consumers.remove(t);
       if (result) {
          iterator.update(consumers.resettableIterator());
+         if (consumers.isEmpty()) {
+            // TODO: can I always reset?
+            reset();
+         }
```
gemmellr, line 103: Did you decide no?

Calling the 'external' reset() method instead of iterator.reset() when everything else in the method uses the iterator and consumers objects directly seems a bit odd.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -4220,11 +4222,13 @@ public QueueConfiguration getQueueConfiguration() {
 
    protected static class ConsumerHolder<T extends Consumer> implements PriorityAware {
 
-      ConsumerHolder(final T consumer) {
+      ConsumerHolder(final T consumer, QueueImpl queue) {
```
gemmellr, line 4225: Odd to have one listed final and the other not.

tests/leak-tests/pom.xml
```diff
@@ -0,0 +1,369 @@
...
+   <build>
+      <testResources>
+         <testResource>
+            <directory>src/test/resources</directory>
+            <filtering>true</filtering>
+         </testResource>
+      </testResources>
+      <plugins>
+         <plugin>
+            <!-- exposed through jitpack.io.
```
gemmellr, line 301: Not anymore so comment also needs updated

gemmellr:
More generally, there are plenty of non-test code changes, they seem like they should be the focus of the JIRA now rather than the tests.

tests/leak-tests/pom.xml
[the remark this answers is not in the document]
clebertsuconic: I am only keeping MQTT Paho client for now...other than that... I think it's all clear...

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueConsumersImpl.java
clebertsuconic: I will keep it.. 
and remove it

tests/leak-tests/pom.xml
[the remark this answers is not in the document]
gemmellr: There still appears to be rather a lot kept around that are either unused or superfluous, even besides the MQTT 3 and 5 clients (which if they arent used as it seems, can be dropped added whenever any tests are added that use them).

Most of the dependencies still seem unused or superfluous repetition of transitive deps...e.g various individual netty deps? They seem like they should all be implicit transitively unless the tests are explicitly using netty themslves, which they dont appear to be. Is anything using artemis-ra, artemis-spring-integration, transaction and management specs? Etc.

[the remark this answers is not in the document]
clebertsuconic: I removed all I could.


I could still remove stomp-server and MQTT server, however I don't want to do that.. as I want to keep these libraries in the broker... not just for future usage, but also to keep the broker's protocols closer to the runtime.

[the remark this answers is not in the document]
clebertsuconic: I will have to add a few back (Probably JSON only.. but I'm reopening this until I fix it)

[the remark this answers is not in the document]
clebertsuconic: @gemmellr fixed JSON. At this point I don't know what else to take out.

tests/leak-tests/pom.xml
```diff
@@ -0,0 +1,250 @@
...
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-jms-client</artifactId>
+         <version>${project.version}</version>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-commons</artifactId>
+         <version>${project.version}</version>
+      </dependency>
```
gemmellr, line 61: Feels like it could just be a transitive dep, I broke the test-chains using it in specific test code. Can it be?

tests/leak-tests/pom.xml
```diff
@@ -0,0 +1,250 @@
...
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-openwire-protocol</artifactId>
+         <version>${project.version}</version>
+         <exclusions>
+            <exclusion>
+               <groupId>org.apache.geronimo.specs</groupId>
+               <artifactId>geronimo-jms_1.1_spec</artifactId>
+            </exclusion>
+         </exclusions>
```
gemmellr, line 86: We shouldnt really need to be doing this here, is it actually required/doing anything in this case? (Since artemis-openwire-protocol should be excluding this itself from any of its deps if its e.g relying on the JMS 2 spec instead, and so should not actually be passing it on to us here).

tests/leak-tests/pom.xml
```diff
@@ -0,0 +1,250 @@
...
+      </dependency>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-core-client</artifactId>
+         <version>${project.version}</version>
+      </dependency>
```
gemmellr, line 92: Seems like it should be transitive unless we are using it directly instead of via the JMS client.

tests/leak-tests/pom.xml
```diff
@@ -0,0 +1,250 @@
...
+               <artifactId>geronimo-j2ee-management_1.1_spec</artifactId>
+            </exclusion>
+         </exclusions>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-mqtt-protocol</artifactId>
+         <version>${project.version}</version>
+      </dependency>
+      <!-- this is required for openwire clients since we remove the management dependency on the exclusions for activemq-client -->
```
gemmellr, line 134: Lets move/group it right beside the openwire client dep then, instead of having the mqtt-protocol dep inbetween them.

Maybe also group the broker protocol deps together with each other...and along with the broker dep.

Maybe group the client deps together.

tests/leak-tests/pom.xml
```diff
@@ -0,0 +1,250 @@
...
+         <artifactId>junit</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.qpid</groupId>
+         <artifactId>qpid-jms-client</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.qpid</groupId>
+         <artifactId>proton-j</artifactId>
+      </dependency>
```
gemmellr, line 150: Should just be transitive unless using proton-j classes directly.

tests/leak-tests/pom.xml
```diff
@@ -0,0 +1,250 @@
...
+   <artifactId>leak-tests</artifactId>
+   <packaging>jar</packaging>
+   <name>ActiveMQ Artemis Leak Tests</name>
+
+   <properties>
+      <activemq.basedir>${project.basedir}/../..</activemq.basedir>
+      <check-leak-version>0.7</check-leak-version>
+      <!-- whatever name you choose it has to match between the check-leak-maven-plugin install call and the JVM argument -->
+      <check-leak-deploy-name>checkleak.lib</check-leak-deploy-name>
+      <its-surefire-extra-args />
```
gemmellr, line 35: This and its doing-nothing usage at the surefire config can be removed as isnt really doing anything (plus the original use was meant to have module-specific names, this one is copied)

tests/leak-tests/pom.xml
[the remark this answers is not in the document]
gemmellr: Commented on some more that seem like they could go. I can live with the broker protocol dependencies since that is a representative-test thing.

Not being able to easily look at whats there and see something is unused or unnecessary to declare (i.e purely transitive) as youve found, is precisely why it should start with the minimal appropriate and then only have things added/removed as we go.

tests/leak-tests/pom.xml
clebertsuconic: this is for the server's.

tests/leak-tests/pom.xml
clebertsuconic: I am using it on the assertions.

tests/leak-tests/pom.xml
clebertsuconic: the server needs it

tests/leak-tests/pom.xml
clebertsuconic: the ActiveMQTestBase will need it

tests/leak-tests/pom.xml
gemmellr: and will bring it in transitively

tests/leak-tests/pom.xml
clebertsuconic: actually.. it can go.. I thought I was

tests/leak-tests/pom.xml
gemmellr: I'd expect it to be brought in transitively then since that class isnt actually in the commons module.

tests/leak-tests/pom.xml
gemmellr: There is no clash, the artemis-openwire-protocol module is not passing on a geronimo-jms_1.1_spec dependency. This is doing nothing.

tests/leak-tests/pom.xml
gemmellr: what I meant was, maybe have artemis-server, thenthe related dependencies like artemis-amqp-protocol, artemis-mqtt-protocol etc...and then have the client deps like artemis-jms-client, qpid-jms-client, activemq-client etc...then maybe any other misc dependencies like the test ones...just so its simple to look at and see which bits are included as they are right alongside the other similar or highly related bits.

tests/leak-tests/pom.xml
```diff
@@ -0,0 +1,238 @@
...
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-openwire-protocol</artifactId>
+         <version>${project.version}</version>
+         <exclusions>
+            <!-- we are replacing this dependency by jakarta.management.j2ee-api, otherwise it would clash with Core JMS client -->
+            <exclusion>
+               <groupId>org.apache.geronimo.specs</groupId>
+               <artifactId>geronimo-jms_1.1_spec</artifactId>
+            </exclusion>
+         </exclusions>
```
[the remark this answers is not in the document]
clebertsuconic: it's weird... it's defined on the integration-tests 
removed it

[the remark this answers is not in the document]
gemmellr: it possibly was 'needed' at some point, because artemis-openwire-protocol was actually doing the wrong thing...the fix should always have been to ensure the dep wasnt passed on rather than exclude it in the tests though.

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -251,6 +251,9 @@ public void runAfter() {
    }
 
 
+   protected void clearServers() {
```
jbertram, line 254: This doesn't appear to be used anywhere. Was it added by mistake?

gemmellr: Believe its used in the new test which subclasses this (I also thought it was unused initially)

clebertsuconic: although I will make the other call to servers.clear to use this method explicitly ...

PR #4368, opened 2023-02-13
ARTEMIS-4167 Enhanced deserialization filter

swerner0:
Now that Artemis is Java 11+ compatible, there is now the ability to set an ObjectInputFilter on an ObjectInputStream. This allows us to write more advanced filter patterns or plug-in our own object filter implementation.

### DIFF
--- a/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQConnection.java
+++ b/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQConnection.java
@@ -689,6 +689,13 @@ public class ActiveMQConnection extends ActiveMQConnectionForContextImpl impleme
       return this.factoryReference.getDeserializationWhiteList();
    }
 
+   public String getSerialFilter() {
+      return this.factoryReference.getSerialFilter();
+   }
+
+   public String getSerialFilterClassName() {
+      return this.factoryReference.getSerialFilterClassName();
+   }
 
    private static class JMSFailureListener implements SessionFailureListener {
 
--- a/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQConnectionFactory.java
+++ b/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQConnectionFactory.java
@@ -95,6 +95,10 @@ public class ActiveMQConnectionFactory extends JNDIStorable implements Connectio
 
    private String deserializationWhiteList;
 
+   private String serialFilter;
+
+   private String serialFilterClassName;
+
    private boolean cacheDestinations;
 
    // keeping this field for serialization compatibility only. do not use it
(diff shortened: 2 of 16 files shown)


jbertram:
The name "SerialFilter" is used throughout the changes. However, I think this should be changed to "DeserializationFilter" in order to be more accurate about what it is and to be consistent with the existing "DeserializationWhileList" and "DeserializationBlackList" as well.

swerner0:
Makes sense, I will rename

gemmellr:
Would actually passing in the actual filter object make sense vs setting the option with a class name? It wouldnt work with JNDI but passing the object (or class name) to begin with seems like much more of more of niche case vs just setting the string based filter as in the other option.

The precedence between the 2 different factory options, and indeed further between those options and the related system properties, seems like it needs documented as its not immediately clear.

swerner0:
This was my original ask on the mailing list: https://lists.apache.org/thread/0zktnksnxvnpdqwp1wp2tojw3rtq2wz6

Setting the object is much more convenient but breaks everything else as the properties are serialized in some form, this would be the first outlier to that.

Agreed on enhancing the docs.

swerner0:
To add, the existing DeserializationWhileList and DeserializationBlackList will become redundant after DeserializationFilter is introduced, as it handles both cases, with just a different syntax which is maintained by Oracle. So we can state the whitelist and blacklist will be removed by X.Y.Z release in favor for the filter or mention the filter will be the preferred way going forward, tbd on the deprecation / removal.

jbertram:
That's a nice win because those lists and their associated methods need to be deprecated anyway since the black/white terms have been deemed non-inclusive (see more at [ARTEMIS-3474](https://issues.apache.org/jira/browse/ARTEMIS-3474)). With the new method you're adding we won't need to create any other new methods. :+1:

swerner0:
Ok, how would you like to proceed? After the variable name change above, I can leave as is, or tackle the deprecation/removal as part of this jira as well.

jbertram:
Go ahead and add `@Deprecated` where necessary, but don't remove anything yet. We'll have to wait for a major version before we can do that. The new documentation will need to be clear on the deprecation as well as how to use the new method to accomplish the same tasks as before.

gemmellr:
> This was my original ask on the mailing list: https://lists.apache.org/thread/0zktnksnxvnpdqwp1wp2tojw3rtq2wz6
> 
> Setting the object is much more convenient but breaks everything else as the properties are serialized in some form, this would be the first outlier to that.
> 
> Agreed on enhancing the docs.

I would think there should be a way to stop serializing certain stuff. Doing so in a case like this wouldnt seem odd to me. If setting the object is more convenient, and most folks should use the filter string instead anyway, it seems further odd to do add the second class option.

Related somewhat...in Qpid JMS we added an 'extensions' mechanism to the factory so we could define specific BiFunction extension points to get custom config etc objects at the appropriate time/place, but only need a single setter on the ConnectionFactory itself, helping helps keep the API (and impl) simpler whilst adding such things that few might use, and also help in not needing to refer to custom types in the factory API. The method just takes a string extension name key and a provided function (to match whatever is defined for that extension name). Applications set their function impl on the factory (perhaps just using a lambda), and when the relevant client code is hit it uses the function if present to get returned what is needed in that case...e.g here such a function could get returned the filter object for that connection as it is created.

https://github.com/apache/qpid-jms/blob/2.2.0/qpid-jms-client/src/main/java/org/apache/qpid/jms/JmsConnectionExtensions.java

For example there are a couple of simple extensions for providing the username and password, which can be seen in use here:
https://github.com/apache/qpid-jms/blob/2.2.0/qpid-jms-client/src/test/java/org/apache/qpid/jms/integration/SaslIntegrationTest.java#L514-L519

Those clearly aren't very custom things, but the main point for those ones was more that it enables them to be given specifically when needed as the connection is being created, and more importantly given again when it is later re-connecting (thats covered elsewhere in a failover test) as it is then fired again and updated valid credentials can be provided if needed, useful for certain use case with expiring credentials.

jbertram:
At this point it appears this PR has been abandoned. Furthermore, it has numerous conflicts. Therefore, I will go ahead and close this PR. @swerner0, please re-open it with updates if you want it to be considered again. Thanks!

swerner0:
No worries, I saw the renames under the latest release and figured would cause conflicts. I'll revisit in the future and open a new PR. Thanks

jbertram:
FWIW, you can update your branch, push -f, and then re-open this PR when you're ready.

PR #4370, opened 2023-02-15
ARTEMIS-4169 Auto detect dead-letter and expiry queues in the console

brusdev:


jbertram:
I pulled your branch and built a release in order to test it. However, when I create an address named `foo` with an anycast queue named `foo` and send a message to it nothing is showing up in the browse tab. If I send a message directly to `DLQ` I can browse messages there.

Am I doing something wrong or is the PR broken somehow?

brusdev:
> I pulled your branch and built a release in order to test it. However, when I create an address named `foo` with an anycast queue named `foo` and send a message to it nothing is showing up in the browse tab. If I send a message directly to `DLQ` I can browse messages there.

@jbertram when I send messages to the foo address `./broker/bin/artemis producer --verbose --destination foo --user admin --password admin`, I see:
image

What client are you using to send messages to the foo address?

jbertram:
I was using the "Send Message" tab on the console itself. I just tested it again to make sure it wasn't a fluke, and I reproduced it. Here's exactly what I did:

1. Checkout your branch and build with `mvn clean package -Prelease -DskipTests`.
2. Create a new instance and start it.
3. Open the console.
4. Create an address `foo` which supports anycast.
5. Create an anycast queue `foo` on the address `foo`.
6. Use the "Send Message" tab to send a message to `foo` with a body of `xxx`.
7. Browse the contents of the queue `foo`. Here's what it shows:
foo

brusdev:
@jbertram good catch!!! I fixed it and added a test.

PR #4371, opened 2023-02-15
ARTEMIS-4170 remove redundant queue creation for OpenWire

jbertram:
o.a.a.a.c.p.o.a.AMQConsumer#init will *always* try to create a core queue when creating a consumer for a JMS queue. However, this is already done in o.a.a.a.c.p.o.a.AMQSession#createConsumer.


clebertsuconic:
@jbertram had you ran the test suite on this. I think the test suite is broken because of this (on -Ptests) or perhaps the examples are broken.

jbertram:
I believe I ran it, but it's hard to remember at this point.

jbertram:
The full test-suite passed on this branch.

clebertsuconic:
@jbertram thanks for checking... the tests frozen twice after I merged this... and passed once without it.. but it was just coincidence... 

Thanks for checking.. I just got a full pass on main now.

PR #4374, opened 2023-02-17
ARTEMIS-4174 Listen only to provided connector-host for JMX RMI sockets

mblauth:
With this change applied, artemis now only listens to localhost when passing connector-host as localhost (or leaving it empty, as it is the default) in management.xml.

**Testing done**

I created an instanced and uncommented the "<connector…" line in etc/management.xml:

`  <connector connector-port="1099"/>`

With this set, I started the instance and confirmed it was only listening to localhost:

`$ netstat -tan | grep 1099 | grep LISTEN
tcp6       0      0 127.0.0.1:1099          :::*                    LISTEN `

I was still able to connect via jconsole when connecting to localhost:1099.

Screenshot from 2023-02-17 17-44-50

Connection attempts from another system were refused as expected.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/ManagementConnector.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/ManagementConnector.java
@@ -55,6 +55,7 @@ public class ManagementConnector implements ActiveMQComponent {
    public void start() throws Exception {
       rmiRegistryFactory = new RmiRegistryFactory();
       rmiRegistryFactory.setPort(configuration.getConnectorPort());
+      rmiRegistryFactory.setHost(configuration.getConnectorHost());
       rmiRegistryFactory.init();
 
       mbeanServerFactory = new MBeanServerFactory();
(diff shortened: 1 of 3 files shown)


jbertram:
This change looks good, but I'm concerned that there's no automatic test to validate the fix and mitigate against future regressions. I can't think of a good way to test this off the top of my head. Do you have any potential ideas?

jbertram:
@mblauth, I think tracking the socket factories makes sense.

mblauth:
I just had a bit of time of writing a little unit test for it. Again, it is quite limited in what it achieves. It just ensures that the server sockets created by the factory are indeed limited to the host specified when creating the registry. To implement the test I had to expose a little package-local helper method to create a test socket, less than ideal but maybe it does what you had in mind. Let me know what you think.

mblauth:
To me it seems the errors above are unrelated to my changes. Let me know if that is not correct.

jbertram:
FYI - I squashed the commits before I merged them. Thanks for the contribution!

PR #4375, opened 2023-02-18
some tasks around Memory Leaks in the broker

clebertsuconic:
I am now removing the negatives once the file is removed or reused

### DIFF
--- a/tests/leak-tests/src/test/java/org/apache/activemq/artemis/tests/leak/ClientLeakTest.java
+++ b/tests/leak-tests/src/test/java/org/apache/activemq/artemis/tests/leak/ClientLeakTest.java
@@ -1,0 +1,138 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.tests.leak;
+
+import javax.jms.Connection;
+import javax.jms.ConnectionFactory;
+import javax.jms.Message;
+import javax.jms.MessageConsumer;
+import javax.jms.MessageProducer;
+import javax.jms.Session;
+import java.lang.invoke.MethodHandles;
+import java.util.HashMap;
+
+import io.github.checkleak.core.CheckLeak;
+import org.apache.activemq.artemis.api.core.TransportConfiguration;
+import org.apache.activemq.artemis.core.config.impl.ConfigurationImpl;
+import org.apache.activemq.artemis.core.server.ActiveMQServer;
+import org.apache.activemq.artemis.core.server.ActiveMQServers;
+import org.apache.activemq.artemis.tests.util.ActiveMQTestBase;
+import org.apache.activemq.artemis.tests.util.CFUtil;
+import org.apache.activemq.artemis.utils.SpawnedVMSupport;
+import org.apache.qpid.proton.engine.impl.ReceiverImpl;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Assume;
+import org.junit.Before;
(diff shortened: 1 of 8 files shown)


clebertsuconic:
another win for https://github.com/check-leak/check-leak 

:)

tests/leak-tests/src/test/java/org/apache/activemq/artemis/tests/leak/ClientLeakTest.java
```diff
@@ -0,0 +1,214 @@
...
+                     message.setIntProperty("i", msg);
+                     sourceProducer.send(message);
+                  }
+                  sourceProducer.close();
+                  producerSession.commit();
+               }
+            }
+
+            // it shouldn't have any sender connects still. The retry loop is leaving a few behind.:w
+            assertMemory(checkLeak, 5, SenderImpl.class.getName());
```
clebertsuconic, line 155: @gemmellr SenderImpl keeps piling up on this test... Even though the producers and consumers are  closed... 

I think this is similar to what I see on the server's.

clebertsuconic: false alarm... I had a producer I forgot to close. duh

tests/leak-tests/src/test/java/org/apache/activemq/artemis/tests/leak/ClientLeakTest.java
```diff
@@ -0,0 +1,137 @@
...
+         }
+
+         for (int j = 0; j < 10; j++) {
+            Session session = connection.createSession(true, Session.SESSION_TRANSACTED);
+            MessageConsumer consumer = session.createConsumer(session.createQueue("test"));
+            connection.start();
+            Message message = consumer.receive(1000);
+            Assert.assertNotNull(message);
+            session.commit();
+            // consumer.close(); // uncomment this and the test will pass.
```
clebertsuconic, line 129: @gemmellr I am not sure if this has any relation to anything I'm seeing now... but if I remove this comment, the test will pass. closing the consumer will throw some log.warns on the server that I'm investigating and are expected to happen, but this will make the test to not keep any references on  any client Receivers.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPSessionCallback.java
```diff
@@ -418,6 +418,17 @@ public void cancel(Object brokerConsumer, Message message, boolean updateCounts)
       }
    }
 
+   /** to be used when credits were drained between handle and actual delivery */
+   public void returnToQueue(Object brokerConsumer, MessageReference reference) {
```
michaelandrepearce, line 422: Is this issue re proton credits possibly occuring in earlier versions say 2.21? Asking as had odd issue with consumers just pausing like they run out of credits.

clebertsuconic: The issue here is really about a memory leak. 

I had credits when handleDelivery was called. later on I called deliver, the AMQPStandardMessage would be referenced on a DeliveryImpl that would stay forever in the Link.workingList until the session or the connection is closed.


I had other issues that I fixed as part of this PR regarding QueueConsumer. where the iterator would hung in the Iterators. Perhaps that one would cause some issues.

https://github.com/apache/activemq-artemis/pull/4375/commits/a1aea9a79aa844673ef705d2a60fadb4435accfc#diff-fc4a9a0dc01f79ecc1435246dc85446fa57d6d04e1dffed53c522d8c145623fdR3098-R3105

clebertsuconic: This is being tested on ConnectionLeakTest::testCheckIterators*, which is using https://github.com/check-leak/check-leak

clebertsuconic: @michaelandrepearce those issues we had were mostly related to closing consumers with a lot of pre-fetch in the queue, which is an anti-pattern. One user reported issues here, and even though we will make sure he changes his usage,I  still wanted to fix these.

Although the Iterator and Journal were issues independent of prefetch or AMQP.

clebertsuconic:
@gemmellr I have organized the PR here. I have the "bad fix" and the test separate here now. If you deal with that leak in Proton we would just drop the "proposed fix to deal with cancellations"

clebertsuconic:
@gemmellr this PR will sit here until a proton-j release is done

clebertsuconic:
@gemmellr I will update this PR with the upgrade on Proton, and make these tests faster, and remove it from the fast-tests

PR #4378, opened 2023-02-23
ARTEMIS-4179 Upgrade keycloak version to 21.0.0

brusdev:


gtully (commented):
does keycloak now just die on an interrupt signal after mvn verify?

gtully:
it does!

PR #4382, opened 2023-02-27
ARTEMIS-4184 Bidges with concurrency not cleared properly on config r…

AntonRoskvist:
…eload

This is causes bridges to leak, every reload adding more bridges when concurrency > 1

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
@@ -4619,34 +4619,49 @@ public class ActiveMQServerImpl implements ActiveMQServer {
          deployQueuesFromListQueueConfiguration(configuration.getQueueConfigs());
 
          ActiveMQServerLogger.LOGGER.reloadingConfiguration("bridges");
+
          for (BridgeConfiguration newBridgeConfig : configuration.getBridgeConfigurations()) {
-            newBridgeConfig.setParentName(newBridgeConfig.getName());
-            Bridge existingBridge = clusterManager.getBridges().get(newBridgeConfig.getParentName());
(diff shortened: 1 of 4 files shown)


jbertram:
There's a conflict that needs to be resolved. Please rebase and resolve it. Thanks!

clebertsuconic:
I got this one as well..

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -4586,9 +4586,12 @@ private void deployReloadableConfigFromConfiguration() throws Exception {
          for (BridgeConfiguration newBridgeConfig : configuration.getBridgeConfigurations()) {
             newBridgeConfig.setParentName(newBridgeConfig.getName());
             Bridge existingBridge = clusterManager.getBridges().get(newBridgeConfig.getParentName());
+            if (existingBridge == null) {
+               existingBridge = clusterManager.getBridges().get(newBridgeConfig.getParentName() + "-0");
```
clebertsuconic, line 4590: I don't understand what is this concatenation here. is this really necessary?

clebertsuconic: It doesn't seem very "product" to me.. it would work somewhere individually.. but it does not seem right to me.

AntonRoskvist: @clebertsuconic 
I agree, that really wasn't nice. I was using the concatenation there because concurrent bridges get named uniquely to be manageable by appending -0, -1 and so on to their name. At this point that has yet to happen to the newBridge.

I made some changes here and also rebased against main, hope it's looking better.

clebertsuconic:
LGTM... I will run the full test suite on my server, and merge this if it's good.

PR #4383, opened 2023-02-27
ARTEMIS-4185 Resending compressed message uncompressed throws excepti…

AntonRoskvist:
…on in consumer

Don't know how common this pattern is but I encountered this when routing messages using camel-jms. The message is send compressed over a WAN link and as step two it is supposed to get published to a broker cluster on a LAN link. Running with compressLargeMessages=true in step one and =false in step two causes a nasty looking exception for the upcoming consumer.


jbertram:
There's a conflict that needs to be resolved. Please rebase and resolve it. Thanks!

clebertsuconic:
I got this one...

PR #4385, opened 2023-03-01
ARTEMIS-4189 NettyConnection.isSameTarget() should compare host names by IPs

gaohoward:


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/NetworkUtil.java
```diff
@@ -0,0 +1,84 @@
...
+import java.net.UnknownHostException;
+
+/**
+ * This is a wrapper class of some InetAddress methods
+ * so we can mock them with static values in test environment
+ */
+public class NetworkUtil {
+
+   public static final InetAddressWrapper[] EMPTY_ADDRESSES = {};
+   public static TestMode testMode = null;
```
clebertsuconic, line 30: This TestMode is a big concern for me.

I would play with interfaces, say make an interface called NetowrkResolver,


Then you would make a NetowrkUtil implement NetworkResolver, that would be a singleton.

NetworkConnection would get NetworkUtil.getInstance() as the NetworkResolver by default.


you make a method package private in NetworkConnection to replace the resolver by your test Instance.


You create an accessor that leaves on a test only.


There are a couple examples that are using Accessor at Package level exposing a public method.

clebertsuconic: That was just a suggestion. But Key here is the package level method to set the resolver on the NetworkConnection. and Test Accessor that live in the same package but only accessible through test code. (That is the accessor lives in the test classes).


Let me know if you understand this?

clebertsuconic: well... another concern is that test is not actually testing the new code. How can you guarantee the codebase of NetworkUtil itself?

gemmellr:
I havent looked at the code yet, but have a more general wonder about this change.

Not all that long ago we effectively made an opposite kind of change in the failover bits of qpid-jms to resolve an issue where its use of resolved IPs for hostnames actually broke things. We made it stop resolving failover hosts to their IPs for comparison and just use the hostname string values as given, because resolving them like it previously did meant it would often be de-duping away entirely distinct hostname values that all just happened to resolve to the same IP, which is a lot more common these days than when the original code was written, due to use of TLS SNI in general and specifically in cloud environments (e.g connections hitting some form of gateway that uses SNI to determine which 'host' is actually being connected to).

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/NetworkUtil.java
gaohoward: > I havent looked at the code yet, but have a more general wonder about this change.
> 
> Not all that long ago we effectively made an opposite kind of change in the failover bits of qpid-jms to resolve an issue where its use of resolved IPs for hostnames actually broke things. We made it stop resolving failover hosts to their IPs for comparison and just use the hostname string values as given, because resolving them like it previously did meant it would often be de-duping away entirely distinct hostname values that all just happened to resolve to the same IP, which is a lot more common these days than when the original code was written, due to use of TLS SNI in general and specifically in cloud environments (e.g connections hitting some form of gateway that uses SNI to determine which 'host' is actually being connected to).

It probably is not that complex for this use case. By the time this kind of comparison is called the hostnames have been compared already. I'm not sure if broker can be setup to distinguish 2 hostnames as different hosts even if they are of same ip? I mean can 2 broker in a cluster binding to same ip and same port?

gemmellr: If its only comparing their IPs if the two hostnames were already decided to be different like it appears, then it seems like its just reduced to essentially the same situation we changed qpid-jms to stop getting into anymore. Thats where it was comparing the resolved IPs of two actually-different hostname values, found them to resolve to the same IP, and so then decided those two different things were really goign to the same place and so it de-duped them...but in reality the different hostnames values really meant it would actually lead to connection to two different things, hosted behind the same gateway IP and port thanks to TLS SNI.

gaohoward: > well... another concern is that test is not actually testing the new code. How can you guarantee the codebase of NetworkUtil itself?

Well kinda testing the logics of the new code. I've difficulty in testing the new code itself mainly because I couldn't find a good way to simulate InetAddress methods. So I do a wrapper just to mock those methods but the logic is real.

gaohoward: > That was just a suggestion. But Key here is the package level method to set the resolver on the NetworkConnection. and Test Accessor that live in the same package but only accessible through test code. (That is the accessor lives in the test classes).
> 
> Let me know if you understand this?

Not sure I understand it. Can you point me some example?

clebertsuconic:
I also think it shouldn't be resolving IPs.... a hostname is a hostname...


what's the point of having two hostnames pointing to the same IP and considering it the same thing?

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/NetworkUtil.java
clebertsuconic: Notice we have two discussions going on here: 1st. do we need to resolve the IPs?

2nd: if you still need the Util class, should you allow the test thing....


I'm not sure about I yet, but if you need to change behavior in test, what you're doing is definitely wrong.


Look at https://github.com/apache/activemq-artemis/blob/main/artemis-server/src/test/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeTestAccessor.java


PostOfficeTestAccessor plays with a method that's package private, while calling it from a public static method.


And PostOfficeTestAccesssor lives on the test-jar. Still the same java package, but it's only available through tests.


That way that test method can only be accessed by the test.

You should then implement ConnectorAccessor (whatever the name for. your class) implemen setResolvert

In your test, you, and call the ConnectoraAccessor.setResolver(connector, new TestResolver());

gaohoward:
@clebertsuconic @gemmellr I think I understand the concerns raised by this PR. I have a fix on the image side so that it always uses hostnames to do scaledown -- it will not land in comparison between a hostname and it's IP. I'm closing this PR now. Thanks for the feedback.

PR #4387, opened 2023-03-01
Artemis 4187

Maverick19841972:


jbertram:
You defined `logDebug` which takes a `Supplier<String>`, and you use it like this, e.g.:
```java
logDebug(() -> String.format("Switch from OVER_SIZE to OVER_ELEMENTS [currentSize=%s, currentElements=%s, lowerMarkSize=%s, maxElements=%s]",
        currentSize, currentElements, lowerMarkSize, maxElements));
```
However, I don't see how that is preferable to just following the same logging pattern that's used in the rest of the code-base, e.g.:
```java
logger.debug("Switch from OVER_SIZE to OVER_ELEMENTS [currentSize={}, currentElements={}, lowerMarkSize={}, maxElements={}]", currentSize, currentElements, lowerMarkSize, maxElements);
```

Couple of other issues:

- All commits need to be squashed.
- Your commit message should conform to the pattern outlined in our [Hacking Guide](https://github.com/apache/activemq-artemis/blob/main/docs/hacking-guide/en/code.md#commitMessageDetails).

Maverick19841972:
The logDebug method confirms to the already present logging construct in the SizeAwareMetrics-class. e.g. in de addSize-method which uses the logger.isDebugEnable-method: 

      if (delta == 0) {
            logger.debug("SizeAwareMetric ignored with size 0", new Exception("trace"));
         }
         return sizeUpdater.get(this);
      }

It is a separate method because op the DRY-principle and the supplier is there to "lazy init" the string for performance reasons.

But I'm more than happy to boyscout out the entire logging contruct ??

Futhermore, I will squash and have a look at the commit-message(s).

jbertram:
Since moving to SLF4J and adopting a new logging pattern we've eliminated almost all of the invocations of `isDebugEnabled`, `isTraceEnabled`, etc. due to the fact that the new pattern uses invocations like `debug(String format, Object... arguments)` and this _already_ implements lazy init. As noted in [the JavaDoc for SLF4J](https://www.slf4j.org/apidocs/org/slf4j/Logger.html#debug-java.lang.String-java.lang.Object...-), e.g.:

> This form avoids superfluous string concatenation when the logger is disabled for the DEBUG level.

The reason `isDebugEnabled` is still used in the instance you cited is that it doesn't follow this normal pattern and furthermore uses `new Exception("trace")` which is especially costly in terms of performance. If you follow the pattern outlined in my previous comment I think the code will be simpler, more readable, more consistent with the rest of the code-base, and essentially equivalent in terms of performance.

I'm not sure what you mean by "boyscout out".

Maverick19841972:
I change the logging etc.

Maverick19841972:
@jbertram Do you expect anymore work on my part ?

rvais:
Legit from my POV! Generally good piece of code and improvement.

jbertram:
@Maverick19841972, sorry it took so long to get this merged. Thanks for the contribution!

I added a small test called `testConsistency` to `org.apache.activemq.artemis.utils.SizeAwareMetricTest` because I noticed that none of the modified tests actually failed without your fix. This test validates your fix directly.

PR #4388, opened 2023-03-02
NO-JIRA Adding CompareUpgradeTest to fast-tests

clebertsuconic:


clebertsuconic:
- first commit will fail, as I'm only adding it to fast tests (to test the test is actually picked up and failing since it's supposed to fail)

I will add the fix after the failure.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/upgradeTest/CompareUpgradeTest.java
```diff
@@ -149,7 +149,12 @@ private void compareDirectories(String expectedFolder, String upgradeFolder, Str
 
                   String expectedString = expectedIterator.next().replace("Expected", "").trim();
                   String upgradeString = upgradeIterator.next().trim();
-                  Assert.assertEquals("error on line " + line + " at " + upgradeFile, expectedString, upgradeString);
+
+                  if (f.getName().endsWith("xml") && expectedString.trim().startsWith("<!--") && upgradeString.trim().startsWith("<!--")) {
+                     logger.debug("Both {} and {} are comments, which I don't really mind what's going on inside the comment", expectedString, upgradeString);
```
clebertsuconic, line 154: does anyone have a strong feeling about doing this?

brusdev: I think it's okay to ignore the comments. Should we check if a line includes other content after the comment?
```
<!-- comment text --> <otherContent>...</otherContent>
```

gemmellr: I think the comments of files we are not actually upgrading will change infrequently enough and trivially enough (given they are comments), that it would be better just to update the original thing being compared in this case so it matches the post-update expectations, and leave the test handling as it is rather than trying to make it clever enough to ignore comments but not break other necessary checks. We have each all spent longer already than doing that would have taken.

(We could presumably instead even make the upgrader upgrade this comment-only-changed file so it matched the newly generated expectations file.)

tests/smoke-tests/pom.xml
```diff
@@ -1339,6 +1339,30 @@
...
+                           <include>**/smoke/upgradeTest/*Test.java</include>
+                        </includes>
+                     </configuration>
+                  </plugin>
+               </plugins>
+            </pluginManagement>
+         </build>
+         <properties>
+            <skipSmokeTests>false</skipSmokeTests>
+         </properties>
```
gemmellr, line 1363: I think this should be done in the root pom like it is for other modules participating in the fast-tests profile. Makes it easier to see at the root which modules tests are going to be enabled. It can reference that module additionally restricts which tests are run. 

E.g:
https://github.com/apache/activemq-artemis/blob/8078dd098ccc38cd57eefa3e94ba1dfe50c6865b/pom.xml#L1296-L1299

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/upgradeTest/CompareUpgradeTest.java
clebertsuconic: @gemmellr it's not that simple to generate the jolokia-access.xml

we would have to parse the host-name and if we are relaxing jolokia or not.

At this point we should just ignore the comment on this file. Going beyond that would take a long time.


I will just update the test to ignore comments on this file only. 


We could deal with that in the future if we have to.

clebertsuconic:
I will merge this for now. If anyone has a better solution please follow up with a new commit.

clebertsuconic:
@gemmellr / @brusdev the OCD is strong on this one, I will have to fix it properly.. sending a fix by the end of the day :)


I will do the proper upgrade and parsing in jolokia-acces.xml

gemmellr:
@clebertsuconic  I didnt see this comment until I came here to reference a commit...and I'm not sure youve seen my comment elsewhere, lol.

I've already pushed a change to address it in the way I was suggesting yesterday: https://github.com/apache/activemq-artemis/commit/1f3745c87d77252d81fa22486d4cc2a06420d66b

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/upgradeTest/CompareUpgradeTest.java
gemmellr: I wasnt suggesting generating anything. I was saying, we should change the initial pre-upgrade file, which is also the post-upgrade reference in this case, to just match what the test is already causing to be generated by the new broker...since we dont actually upgrade that file currently, just copy it.

PR #4389, opened 2023-03-02
introduce ClientIDCallback to allow JAAS login module to retrieve Client ID

amarkevich:
Currently there is no way to get client ID in custom login module. Its required for some MQTT cases when username is not used.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/ClientIDCallback.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/ClientIDCallback.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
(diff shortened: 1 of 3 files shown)


jbertram:
Couple of things:

- You need a Jira for this change.
- You need to squash all your commits.
- Your commit message should conform to the pattern outlined in the [Hacking Guide](https://github.com/apache/activemq-artemis/blob/main/docs/hacking-guide/en/code.md#commitMessageDetails).

brusdev:
What is the JAAS login module that requires the ClientIDCallback?

amarkevich:
> What is the JAAS login module that requires the ClientIDCallback?

In my particular case custom JAAS login module contains implementation specific logic but Client ID is required to handle the case

gtully (commented):
looks good. nice tidy up in there. thanks for the contribution. I would include a test, there is no existing unit test but it would benefit from a JaasCallbackHandlerTest to protect this addition since there is on login module that exercises it.

amarkevich:
@gtully added JaasCallbackHandlerTest

PR #4394, opened 2023-03-06
ARTEMIS-3640 Use client connectors for HA

brusdev:


gtully (commented):
this looks great. the initial urls can be reused and tagged to allow a match with the topology. if the topology url info is invalid, fallbck to the initial list, the name attribute can be used to select an entry to reconnect to. In this way, a ha pair can be respected and the invalid urls ignored.
this is a neat trick. 
is there some place for a little bit of doc?

gtully:
w.r.t to doc, HA=true is really Pair Affinity = true. This feature allows a client url to respect the affinity but to override the url.

gtully (approved):

brusdev:
@gtully thanks for your feedback, after your review I only added a section in the doc to describe how to use client connectors to fail over

PR #4395, opened 2023-03-07
ARTEMIS-4193 Large Message Files orphaned after server killed

clebertsuconic:
This fix is scanning journal and paging for existing large messages. We will remove any large messages that do not have a corresponding record in journals or paging.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/PagingManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/PagingManager.java
@@ -17,6 +17,7 @@
 package org.apache.activemq.artemis.core.paging;
 
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.Future;
 import java.util.function.BiConsumer;
 
@@ -84,6 +85,11 @@ public interface PagingManager extends ActiveMQComponent, HierarchicalRepository
    void addBlockedStore(PagingStore store);
 
    void injectMonitor(FileStoreMonitor monitor) throws Exception;
+
+   /** Execute a runnable inside the PagingManager's executor */
+   default void execute(Runnable runnable) {
+      throw new UnsupportedOperationException("not implemented");
+   }
 
    /**
     * Lock the manager. This method should not be called during normal PagingManager usage.
@@ -164,7 +170,7 @@ public interface PagingManager extends ActiveMQComponent, HierarchicalRepository
    /**
     * Rebuilds all page counters for destinations that are paging in the background.
     */
-   default Future<Object> rebuildCounters() {
+   default Future<Object> rebuildCounters(Set<Long> storedLargeMessages) {
       return null;
    }
 
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCounterRebuildManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCounterRebuildManager.java
@@ -34,6 +34,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.lang.invoke.MethodHandles;
+import java.util.Set;
 import java.util.function.BiConsumer;
 
 /** this class will copy current data from the Subscriptions, count messages while the server is already active
@@ -49,14 +50,16 @@ public class PageCounterRebuildManager implements Runnable {
    private long limitPageId;
    private int limitMessageNr;
    private LongObjectHashMap<CopiedSubscription> copiedSubscriptionMap = new LongObjectHashMap<>();
+   private final Set<Long> storedLargeMessages;
 
 
-   public PageCounterRebuildManager(PagingStore store, LongHashSet transactions) {
+   public PageCounterRebuildManager(PagingStore store, LongHashSet transactions, Set<Long> storedLargeMessages) {
       // we make a copy of the data because we are allowing data to influx. We will consolidate the values at the end
       initialize(store);
       this.pgStore = store;
       this.sm = store.getStorageManager();
       this.transactions = transactions;
+      this.storedLargeMessages = storedLargeMessages;
    }
    /** this method will perform the copy from Acked recorded from the subscription into a separate data structure.
     * So we can count data while we consolidate at the end */
@@ -216,6 +219,12 @@ public class PageCounterRebuildManager implements Runnable {
          try (LinkedListIterator<PagedMessage> iter = msgs.iterator()) {
             while (iter.hasNext()) {
                PagedMessage msg = iter.next();
+               if (storedLargeMessages != null && msg.getMessage().isLargeMessage()) {
+                  if (logger.isDebugEnabled()) {
+                     logger.debug("removing storedLargeMessage {}", msg.getMessage().getMessageID());
+                  }
+                  storedLargeMessages.remove(msg.getMessage().getMessageID());
+               }
                if (limitPageId == pgid) {
                   if (msg.getMessageNumber() >= limitMessageNr) {
                      if (logger.isDebugEnabled()) {
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
@@ -45,7 +45,6 @@ import javax.transaction.xa.Xid;
 import io.netty.buffer.Unpooled;
 import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
 import org.apache.activemq.artemis.api.core.ActiveMQBuffers;
-import org.apache.activemq.artemis.api.core.ActiveMQShutdownException;
 import org.apache.activemq.artemis.api.core.Message;
 import org.apache.activemq.artemis.api.core.Pair;
 import org.apache.activemq.artemis.api.core.SimpleString;
@@ -925,6 +924,7 @@ public abstract class AbstractJournalStorageManager extends CriticalComponentImp
                                                     Map<Long, QueueBindingInfo> queueInfos,
                                                     final Map<SimpleString, List<Pair<byte[], Long>>> duplicateIDMap,
                                                     final Set<Pair<Long, Long>> pendingLargeMessages,
+                                                    final Set<Long> storedLargeMessages,
                                                     List<PageCountPending> pendingNonTXPageCounter,
                                                     final JournalLoader journalLoader) throws Exception {
       SparseArrayLinkedList<RecordInfo> records = new SparseArrayLinkedList<>();
@@ -995,6 +995,10 @@ public abstract class AbstractJournalStorageManager extends CriticalComponentImp
 
                      messages.put(record.id, largeMessage.toMessage());
 
+                     if (storedLargeMessages != null) {
+                        storedLargeMessages.remove(largeMessage.getMessageID());
+                     }
+
                      largeMessages.add(largeMessage);
 
                      break;
@@ -1006,6 +1010,10 @@ public abstract class AbstractJournalStorageManager extends CriticalComponentImp
                   case JournalRecordIds.ADD_MESSAGE_PROTOCOL: {
 
                      Message message = decodeMessage(pools, buff);
+
+                     if (message.isLargeMessage() && storedLargeMessages != null) {
+                        storedLargeMessages.remove(message.getMessageID());
+                     }
 
                      messages.put(record.id, message);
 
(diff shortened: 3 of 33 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/PagingManager.java
```diff
@@ -85,6 +86,11 @@ public interface PagingManager extends ActiveMQComponent, HierarchicalRepository
 
    void injectMonitor(FileStoreMonitor monitor) throws Exception;
 
+   /** Execute a runnable inside the PagingManager's executor */
```
tabish121, line 89: Having the default method that violates the intended contract of the method as documented seems less than ideal.  If the intention is to prevent having to add this to other PagingManager implementations that won't ever have this method invoked then throwing an UnsupportedOperationException would be better because at least then you might notice that the method is being used and not doing something on a thread that was unintended.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCounterRebuildManager.java
```diff
@@ -186,6 +189,7 @@ public void run() {
    }
 
    public void rebuild() throws Exception {
+      logger.info("Rebuilding " + pgStore.getStoreName());
```
tabish121, line 192: Should be using logger formatting as opposed to string concats.

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCounterRebuildManager.java
```diff
@@ -216,6 +220,11 @@ public void rebuild() throws Exception {
          try (LinkedListIterator<PagedMessage> iter = msgs.iterator()) {
             while (iter.hasNext()) {
                PagedMessage msg = iter.next();
+               logger.info("Reading " + msg);
```
tabish121, line 223: Ditto

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCounterRebuildManager.java
clebertsuconic: I didn't mean to keep this method.. thank you

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCounterRebuildManager.java
clebertsuconic: It's debug.. not meant to be kept.. thank you

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCounterRebuildManager.java
```diff
@@ -216,6 +219,10 @@ public void rebuild() throws Exception {
          try (LinkedListIterator<PagedMessage> iter = msgs.iterator()) {
             while (iter.hasNext()) {
                PagedMessage msg = iter.next();
+               if (storedLargeMessages != null && msg.getMessage().isLargeMessage()) {
+                  logger.debug("removing storedLargeMessage {}", msg.getMessage().getMessageID());
```
gemmellr, line 223: As getMessageID returns a long it will be autoboxing, may be worth a gate since all the similar logging before and after this uses them already.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
```diff
@@ -1272,13 +1280,16 @@ final class MutableLong {
...
-         loadPreparedTransactions(postOffice, pagingManager, resourceManager, queueInfos, preparedTransactions, this::failedToPrepareException, pageSubscriptions, pendingLargeMessages, journalLoader);
+         loadPreparedTransactions(postOffice, pagingManager, resourceManager, queueInfos, preparedTransactions, this::failedToPrepareException, pageSubscriptions, pendingLargeMessages, storedLargeMessages, journalLoader);
 
          for (PageSubscription sub : pageSubscriptions.values()) {
             sub.getCounter().processReload();
          }
 
          for (LargeServerMessage msg : largeMessages) {
+            if (storedLargeMessages != null && storedLargeMessages.remove(msg.getMessageID())) {
+               logger.debug("Large message in folder removed on {}", msg.getMessageID());
```
gemmellr, line 1291: Ditto re: gate.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
```diff
@@ -1829,6 +1825,9 @@ private void loadSinglePreparedTransaction(PostOffice postOffice,
 
          switch (recordType) {
             case JournalRecordIds.ADD_LARGE_MESSAGE: {
+               if (storedLargeMessages != null && storedLargeMessages.remove(record.id)) {
+                  logger.debug("PreparedTX load removing stored large message {}", record.id);
```
gemmellr, line 1829: guessing the same here

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/AbstractJournalStorageManager.java
```diff
@@ -1839,6 +1838,9 @@ private void loadSinglePreparedTransaction(PostOffice postOffice,
             }
             case JournalRecordIds.ADD_MESSAGE_PROTOCOL: {
                Message message = decodeMessage(pools, buff);
+               if (storedLargeMessages != null && message.isLargeMessage() && storedLargeMessages.remove(record.id)) {
+                  logger.debug("PreparedTX load removing stored large message {}", record.id);
```
gemmellr, line 1842: might be useful to differentiate the message from the currently-identical log message above to help discern which occurred without resorting to line-number output comparisons (Aside: I'm just assuming there is some 'known'/'obvious' reason for the 2 different record types doing such similar thing to extent of having such simialr log messages).

also, guessing same about gate.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -797,6 +764,14 @@ private Map<Long, Pair<String, Long>> recoverPendingLargeMessages() throws Excep
       return largeMessages;
    }
 
+   @Override
+   public void recoverLargeMessagesOnFolder(Set<Long> files) throws Exception {
+      List<String> filenames = largeMessagesFactory.listFiles("msg");
+      filenames.forEach(f -> {
+         files.add(getLargeMessageIdFromFilename(f));
+      });
+   }
```
gemmellr, line 773: Might be clearer to call the Set 'messageIDs' or something, given it doesnt really contain files/names in the end.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -3381,10 +3382,23 @@ synchronized void initialisePart2(boolean scalingDown) throws Exception {
...
+         pagingManager.rebuildCounters(storedLargeMessages);
+
+         pagingManager.execute(() -> {
+            storedLargeMessages.forEach(id -> {
+               try {
+                  SequentialFile file = storageManager.createFileForLargeMessage(id, true);
+                  logger.debug("Removing pending large message for file={}", file);
+                  file.delete();
+               } catch (Exception e) {
+                  logger.warn(e.getMessage(), e);
```
gemmellr, line 3398: Might be good to give some context in the log message about what was being done when the exception occurred. May not be so obvious to a user from the exception message itself.

tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/interruptlm/LargeMessageInterruptTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+      Assert.assertTrue(killAt.await(60, TimeUnit.SECONDS));
+      serverProcess.destroyForcibly();
+      serverProcess = startServer(SERVER_NAME_0, 0, 0);
+      QueueControl queueControl = getQueueControl(liveURI, liveNameBuilder, queueName, queueName, RoutingType.ANYCAST, 5000);
+
+      Assert.assertTrue(done.await(60, TimeUnit.SECONDS));
+      Assert.assertEquals(0, errors.get());
+
+      long numberOfMessages = queueControl.getMessageCount();
```
gemmellr, line 223: Its not clear to me why this seems to go to the effort of starting the broker without waiting, spinning in a loop trying to connect to management to get a QueueControl (and potentially waiting 500ms at a time, and spewing tracktrace, while doing it), only to not use the QueueControl until after it is done waiting for sending to complete?

tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/interruptlm/LargeMessageInterruptTest.java
```diff
@@ -0,0 +1,244 @@
...
+
+      try (Connection connection = factory.createConnection()) {
+         Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+         MessageConsumer consumer = session.createConsumer(session.createQueue(queueName));
+         connection.start();
+         for (int i = 0; i < numberOfMessages; i++) {
+            TextMessage message = (TextMessage) consumer.receive(5000);
+            Assert.assertNotNull(message);
+            Assert.assertTrue(message.getText().equals("forcePage") || message.getText().equals(largebody));
+            Assert.assertNotNull(message);
```
gemmellr, line 234: Redundant, already checked and used above

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/cursor/impl/PageCounterRebuildManager.java
clebertsuconic: I didn't think this one was a big deal.. it's only happening on startup.. but I will add the gate.

tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/interruptlm/LargeMessageInterruptTest.java
clebertsuconic: I am actually using it to get the numberOfMessages.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
clebertsuconic: At this leve I don't have the messages yet.. just the large messages that are still on the message-folder. At this point I really have a File Name that I would need to parse to recover the messageID,.

clebertsuconic:
I have test failures.. marking it as a draft until I fix them..

clebertsuconic:
I fixed the tests.. I think this is ready to merge

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
gemmellr: The List of filenames contains filenames, but the Set does not, it contains the results added to it which are messageID, so it seems weird to call the Set 'files'.

The caller calls the same Set "storedLargeMessages" which seems nicer.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -3381,10 +3382,24 @@ synchronized void initialisePart2(boolean scalingDown) throws Exception {
...
+
+         pagingManager.execute(() -> {
+            storedLargeMessages.forEach(id -> {
+               try {
+                  SequentialFile file = storageManager.createFileForLargeMessage(id, true);
+                  logger.debug("Removing pending large message for file={}", file);
+                  file.delete();
+               } catch (Exception e) {
+                  // this shouldn't really happen, unless something is off with the storage
+                  logger.warn("It was not possible to remove previously stored large message on folder::It will be retried on next startup::" + e.getMessage(), e);
```
gemmellr, line 3399: Dont think you need to add the exception message, when adding the exception itself, just goign to end up with the message twice. (But if doing so, should use a placeholder rather than concat).

It would however perhaps be good to add at least the message ID that the failure related to though otherwise it may just end up printing a bunch of similar or identical entries that dont make clear which message/files it relates to.

tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/interruptlm/LargeMessageInterruptTest.java
gemmellr: Sure, I didnt say it was unused, only that it goes to the effort of spinning and potentially spewing stacktraces by connecting before the broker is ready...only to wait for the sending to complete before then using it. Why not create it after the wait for send completion when we know the broker is ready?

clebertsuconic: @gemmellr OH.. I see what you mean...


I was going to place either that loop, or a connection loop over JMS... to wait the server to start.
Since the threads already have a retry, I didn't consider to use them as the retry mechanism, which works perfectly. Kind of obvious but I was too stuck in my line of though.. thanks.


I'm moving down the queueControl creation by 2 lines.

PR #4396, opened 2023-03-08
One fix and one feature for MQTT dealing with link stealing

jbertram:


eanbiso:
Hi Justin, could I ask you if it is possible to add some specific logs for tracing the stealing link issue.
I mean, I think it would be very useful to have some logs to identify both the use cases, i.e.:
- a client trying to open a connection with a clientId already used is able to steal the connection to another client already connected with the same clientId (default use case)
- a client trying to open a connection with a clientId already used is not able to do that (allowLinkStealing=false use case)
since it would be very useful to be able to explain a customer the reason why a client cannot connect, or why a client already connected looses its connection (depending on the configuration).
A proposal (if possible obviously) could be to have some specific Audit Logs for tracing this specific use cases since they could be relevant.
Many thanks

jbertram:
FTI - full test-suite passed.

PR #4399, opened 2023-03-10
ARTEMIS-4203 Replace aether-api with maven-resolver-api

brusdev:


gemmellr:
Looks good to me (note for later/others: org.apache.maven.resolver:aether-api actually contains the old org.eclipse.aether package, the code moved)

PR #4401, opened 2023-03-15
NO-JIRA Cleanup MBeanServer usage in the testsuite as it is leaking

clebertsuconic:
Basically I started the testsuite and attached check leak with "java -jar check-leak.jar --pid <pid> --report testsuite-report --sleep 1000" and saw the allocations of this were pretty high.

### DIFF
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
@@ -16,6 +16,8 @@
  */
 package org.apache.activemq.artemis.tests.util;
 
+import javax.management.MBeanServer;
+import javax.management.MBeanServerFactory;
 import javax.naming.Context;
 import javax.transaction.xa.XAException;
 import javax.transaction.xa.Xid;
@@ -175,6 +177,12 @@ import org.junit.runner.Description;
  * Base class with basic utilities on starting up a basic server
  */
 public abstract class ActiveMQTestBase extends Assert {
+
+   public MBeanServer getMBeanServer() {
+      MBeanServer mBeanServer = MBeanServerFactory.createMBeanServer();
+      runAfter(() -> MBeanServerFactory.releaseMBeanServer(mBeanServer));
+      return mBeanServer;
+   }
 
    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/BridgeControlTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/BridgeControlTest.java
@@ -17,7 +17,6 @@
 package org.apache.activemq.artemis.tests.integration.management;
 
 import javax.management.MBeanServer;
-import javax.management.MBeanServerFactory;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -139,7 +138,7 @@ public class BridgeControlTest extends ManagementTestBase {
 
       Configuration conf_0 = createBasicConfig().addAcceptorConfiguration(new TransportConfiguration(InVMAcceptorFactory.class.getName())).addConnectorConfiguration(connectorConfig.getName(), connectorConfig).addQueueConfiguration(sourceQueueConfig).addBridgeConfiguration(bridgeConfig);
 
-      server_1 = addServer(ActiveMQServers.newActiveMQServer(conf_1, MBeanServerFactory.createMBeanServer(), false));
+      server_1 = addServer(ActiveMQServers.newActiveMQServer(conf_1, getMBeanServer(), false));
       addServer(server_1);
       server_1.start();
 
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/BridgeControlUsingCoreTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/BridgeControlUsingCoreTest.java
@@ -16,7 +16,6 @@
  */
 package org.apache.activemq.artemis.tests.integration.management;
 
-import javax.management.MBeanServerFactory;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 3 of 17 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/BridgeControlTest.java
```diff
@@ -139,7 +138,7 @@ public void setUp() throws Exception {
 
       Configuration conf_0 = createBasicConfig().addAcceptorConfiguration(new TransportConfiguration(InVMAcceptorFactory.class.getName())).addConnectorConfiguration(connectorConfig.getName(), connectorConfig).addQueueConfiguration(sourceQueueConfig).addBridgeConfiguration(bridgeConfig);
 
-      server_1 = addServer(ActiveMQServers.newActiveMQServer(conf_1, MBeanServerFactory.createMBeanServer(), false));
+      server_1 = addServer(ActiveMQServers.newActiveMQServer(conf_1, getMBeanServer(), false));
```
gemmellr, line 141: This would now be using the same MBeanServer as the other call below using the variable, mbeanServer, coming from the parent, rather than using two differnet MBeanServers as the different variables suggests...so you might as well make the two consistently use 1 if its acceptable for them to actually be using the same MBeanServer.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/BridgeControlUsingCoreTest.java
```diff
@@ -116,7 +115,7 @@ public void setUp() throws Exception {
 
       Configuration conf_0 = createBasicConfig().addAcceptorConfiguration(new TransportConfiguration(INVM_ACCEPTOR_FACTORY)).addConnectorConfiguration(connectorConfig.getName(), connectorConfig).addQueueConfiguration(sourceQueueConfig).addBridgeConfiguration(bridgeConfig);
 
-      server_1 = addServer(ActiveMQServers.newActiveMQServer(conf_1, MBeanServerFactory.createMBeanServer(), false));
+      server_1 = addServer(ActiveMQServers.newActiveMQServer(conf_1, getMBeanServer(), false));
```
gemmellr, line 118: Ditto

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ClusterConnectionControl2Test.java
```diff
@@ -118,7 +117,7 @@ public void setUp() throws Exception {
 
       Configuration conf_0 = createBasicConfig(1).addClusterConfiguration(clusterConnectionConfig_0).addAcceptorConfiguration(acceptorConfig_0).addConnectorConfiguration("netty", connectorConfig_0).addDiscoveryGroupConfiguration(discoveryName, discoveryGroupConfig).addBroadcastGroupConfiguration(broadcastGroupConfig);
 
-      mbeanServer_1 = MBeanServerFactory.createMBeanServer();
+      mbeanServer_1 = getMBeanServer();
```
gemmellr, line 120: Basically the same

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ClusterConnectionControlTest.java
```diff
@@ -206,7 +205,7 @@ public void setUp() throws Exception {
 
       Configuration conf_0 = createBasicConfig().addAcceptorConfiguration(new TransportConfiguration(InVMAcceptorFactory.class.getName())).addConnectorConfiguration(connectorConfig.getName(), connectorConfig).addClusterConfiguration(clusterConnectionConfig1).addClusterConfiguration(clusterConnectionConfig2).addDiscoveryGroupConfiguration(discoveryGroupName, discoveryGroupConfig);
 
-      mbeanServer_1 = MBeanServerFactory.createMBeanServer();
+      mbeanServer_1 = getMBeanServer();
```
gemmellr, line 208: ..and again..

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -176,6 +178,15 @@
  */
 public abstract class ActiveMQTestBase extends Assert {
 
+   // MBeanServer is an expensive object.
+   // Even if you called unregisterMBeanServer it will still be allocated somewhere in the VM.
+   // this will make the testsuite a lot lighter
+   static MBeanServer mBeanServer = MBeanServerFactory.createMBeanServer();
```
gemmellr, line 184: If all the tests are largely going to use the same MBeanServer, for all their brokers, I'd wonder if they couldn't just use the probably-already-existing _PlatformMBeanServer_ and remove need for this?
https://docs.oracle.com/javase/8/docs/api/java/lang/management/ManagementFactory.html#getPlatformMBeanServer--

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ManagementTestBase.java
```diff
@@ -71,17 +70,12 @@ protected static void consumeMessages(final int expected,
    public void setUp() throws Exception {
       super.setUp();
 
-      createMBeanServer();
-   }
-
-   protected void createMBeanServer() {
-      mbeanServer = MBeanServerFactory.createMBeanServer();
+      getMBeanServer();
```
gemmellr, line 73: This seems like a redundant duplicate.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/util/JMSClusteredTestBase.java
```diff
@@ -119,7 +118,7 @@ private void setupServer2() throws Exception {
 
       JMSConfigurationImpl jmsconfig = new JMSConfigurationImpl();
 
-      mBeanServer2 = MBeanServerFactory.createMBeanServer();
+      mBeanServer2 = getMBeanServer();
```
gemmellr, line 121: As in other cases, misleading having 2 variables both using the same actual MBeanServer

gemmellr (commented):
I guess the only thing about the change is youve made various stuff that was using distinct MBeanServers now use the same one...so at some point it might crop up with e.g collisions (or alternatively thinking you are using one MBean when really using an identically-named different-MBean) that might need handled.

The other thing I'd wonder is as noted in one of the comments, if using a single MBeanServer works then could it just be the Platform MBeanServer so you dont need to create another one at all?

clebertsuconic:
@gemmellr I will just close this for now...


it will be a known issue for us.. but this needs to be handled better.

clebertsuconic:
@gemmellr I'm handling this differently now...

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -176,6 +178,12 @@
  */
 public abstract class ActiveMQTestBase extends Assert {
 
+   public MBeanServer getMBeanServer() {
+      MBeanServer mBeanServer = MBeanServerFactory.createMBeanServer();
+      runAfter(() -> MBeanServerFactory.releaseMBeanServer(mBeanServer));
+      return mBeanServer;
+   }
```
gemmellr, line 185: Renaming this something like createMBeanServer() would seem appropriate now, given its no longer doing the simple get it started out as in the previously reviewed version, and this method is only called in place of prior MBeanServerFactory.createMBeanServer() uses.

clebertsuconic: It's getting you one from somewhere, right? :) 


but I can rename it .. sure

PR #4403, opened 2023-03-15
ARTEMIS-4205 Database Primitive Manager

funkyjive:
Using locking capabilities of the database, implement the a primitive manager to allow quorum voting to occur. This implementation has support for Oracle, MSSQL and Postgres -- with H2 being used for unit tests.

The locking features used are not locks against a tables, but instead the idea of an arbitrarily named "application" lock. There is no SQL standard for this behavior and each database implementation is different. I have accounted for this by creating the concept of a database adapter -- i.e. adapting the primitive manager to work for the particular database. I have attempted to reduce duplication by declaring a BaseDatabaseAdapter which I think should work generally for all databases.

Additionally, I introduced the concept of a DatabaseConnectionProvider. This was done for any users that are utilizing Artemis as an embedded feature or that otherwise need to declare their own mechanism for getting credentials. For instance, they may very well want to utilize some encryption for storing a database password in the configuration. The DefaultDatabaseConnectionProvider just gets url, username and password from the properties. It is likely that this implementation is enough for most uses, but we may want to see if we can make it optional to store the password encrypted. I am wondering if there is stuff already in existence for this kind of thing?

I tried to follow the same locking approach in the zookeeper curator implementation to keep the threading/calling equivalently safe, but review of this is welcome.

In the end it was only necessary to create a database record when storing a mutable long value. These records are held in an ARTEMIS_LOCKS table. A lock that doesn't use and store a long value can come and go without any lasting signature in the database signature. The default value for a lock is 0 and doesn't require a record in the database until the long value is updated. The longs must be stored in the database for a different node to be able to acquire a lock and have the same long value present there, of course.

I would welcome logical review of my approach to maintenance and cleanup by someone more familiar with all aspects of using these locks:
I store a "LAST_ACCESS" value with each of those LONG_VALUE records. Periodically I enumerate the records that have LAST_ACCESS over two hours ago, then grab the lock. The assumption is this: if the maintenance system can grab the lock and it hasn't been accessed in over two hours, it is probably defunct and should be cleaned up.

I didn't know what exactly was culture regarding the database drivers. Oracle and MSSQL drivers are both proprietary and Postgres is open source. Ultimately in order for anyone to use this feature, they need to provide the right driver for the database they are using. But to make the code implementation testable via unit tests, I used the open source H2 database and wrote a little piece that extended the database to add locking functions similar to Oracle, MSSQL and Postgres. I include instructions in the comments at the top of the DatabaseDistributedLockTest on how someone can setup to run the unit tests in Oracle, MSSQL or Postgres modes. I am ultimately leaving this running as much of the code as possible via unit tests, but inherently to run all of it would require 3 database servers with logins.

My rationale for creating this feature is that I believe if someone is using the broker in an application environment where there is a database already present, it would be nice for them to be able to simply use that database and not have to have yet another piece of infrastructure to provide for the quorum based ha replication that Artemis supports. And I also think that it should be possible to do this kind of thing with just 2 broker nodes without requiring a shared file system -- if someone wanted this.

### DIFF
--- a/artemis-quorum-ri/pom.xml
+++ b/artemis-quorum-ri/pom.xml
@@ -76,8 +76,19 @@
          <artifactId>artemis-commons</artifactId>
          <version>${project.version}</version>
       </dependency>
+      <dependency>
+         <groupId>com.sangupta</groupId>
+         <artifactId>murmur</artifactId>
+         <version>1.0.0</version>
+      </dependency>
 
       <!-- tests -->
+      <dependency>
+         <groupId>com.h2database</groupId>
+         <artifactId>h2</artifactId>
+         <version>2.1.214</version>
+         <scope>test</scope>
+      </dependency>
       <dependency>
          <groupId>junit</groupId>
          <artifactId>junit</artifactId>
--- a/artemis-quorum-ri/src/main/java/org/apache/activemq/artemis/quorum/db/BaseDatabaseAdapter.java
+++ b/artemis-quorum-ri/src/main/java/org/apache/activemq/artemis/quorum/db/BaseDatabaseAdapter.java
@@ -1,0 +1,196 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.quorum.db;
+
+import org.apache.activemq.artemis.quorum.UnavailableStateException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.lang.invoke.MethodHandles;
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+
+public abstract class BaseDatabaseAdapter {
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+   protected DatabaseConnectionProvider provider;
+
+   static DatabaseConnectionProvider getProvider(Map<String, String> properties) throws Exception {
+      // a DatabaseConnectionProvider implementation class can be implemented so that the mechanism to get
+      // database connections can be injected by an implementation where artemis is embedded or other circumstances
+      if (properties.containsKey("database-connection-provider-class")) {
+         return (DatabaseConnectionProvider) Class.forName(properties.get("database-connection-provider-class"))
+            .getDeclaredConstructor(Map.class).newInstance(properties);
+      } else {
+         return new DefaultDatabaseConnectionProvider(properties);
+      }
+   }
+
+   public BaseDatabaseAdapter(Map<String, String> properties) throws Exception {
+      this.provider = getProvider(properties);
+   }
+
+   /**
+    * this shouldn't generally be called by the adapter implementation as it gets a new connection.
+    * Connections are retrieved when the DatabaseDistributedLock is created and live for the life of that lock.
+    * From within the adapter call  DatabaseDistributedLock.getAssociatedConnection() instead to get the connection
+    * associated with the lock context.
+    */
+   Connection getConnection() throws SQLException {
+      return provider.getConnection();
+   }
+
+   /**
+    * Implementations must return true if a lock is successfully acquired for the given connection,
+    * false if the lock is unable to be acquired but the attempt was otherwise functional,
+    * exceptions in case things didn't work
+    */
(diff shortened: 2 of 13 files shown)


gtully:
great idea, this should have a lot in common with the jdbc lease lock, have you seen that impl?
from what I recall it has db abstractions etc... it may be worth some investigation to see if there is duplication.

https://github.com/apache/activemq-artemis/blob/main/artemis-server/src/test/java/org/apache/activemq/artemis/core/server/impl/jdbc/JdbcLeaseLockTest.java

funkyjive:
> 

I took a look through that code for the artemis-jdbc-store.  I think the problem being solved, the reasoning and the approach around the locking structure is very different from my approach.  The approach I am taking is to use a uniquely implemented database feature on a per-database basis which provides the logical locking required without much of the accompanying overhead in the form of tables and records.

funkyjive:
Status, I am working on validating all of this implementation against integration tests the quorum stuff

rvais:
I like the approach, but it somewhat duplicates already existing locking mechanism. There is for example https://github.com/apache/activemq-artemis/blob/main/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/jdbc/JdbcSharedStateManager.java. It would be good to somewhat unify the approach. The time I have spent on reading the code and my understanding of implemented locking mechanism is limited so it would be worth to spend some more time on this. I will check this out later again.

clebertsuconic:
@funkyjive how would this different from JdbcSharedStateManager?

funkyjive:
Full disclosure, I don't know each detail about the inner functioning of the JdbcSharedStateManager itself.  So there may be things that I am missing.  

But high level differences between the approach:  
The JdbcSharedStateManager is written to utilize a locking mechanism more closely to the JDBC Spec -- i.e. implement locking oriented around transactions and stuff in the database server that can be done in a more common way between database systems and providers.  This involves record insertions and update locks and stuff.  There is infrastructure built up around this.  My purpose is not to detract from this implementation at all.  There was thought and complexity put into this implementation.

The implementation in this PR can be thought of more as a simple "pass through" to database features offered by specific databases in a per-database way.  There is no JDBC specification correlation for the database features that are used here, it is just that each of these databases happens to support an "application locking" concept in its own specific way.  

There is no table or row lock involved, instead each individual database implementation provides a mechanism for acquiring a logical lock.  Here is the heart of it from each database:  

Oracle: { ? = call dbms_lock.request( lockhandle => ?, lockmode => DBMS_LOCK.X_MODE, timeout => 0)}
MSSQL: {? = call sp_getapplock(@Resource=?, @LockMode='Exclusive', @LockOwner='Session', @LockTimeout=0)}
MySQL: {? = call GET_LOCK(?, 0)}
Postgres: {? = call pg_try_advisory_lock(?) }

As you can see, these are all individual stored procedure concepts that interact with special implementation on the database system itself.  No standardization or specification via JDBC except for that they are called through the JDBC Callable interface for calling functions.

As you look and compare with the journal-sql.properties file from the the JdbcSharedStateManager, you can see that while there does exist database specialization of configuration via queries -- you can also see that there is a logical construct created and maintained by the JdbcSharedStateManager and then each database needs to be able to conform to that logical construct.  And that logical construct relies upon database artifact creation and manipulation (i.e. tables and rows, etc).  

If I was to make arguments for or against any particular approach, I could argue that the implementation of JdbcSharedStateManager is more generic and has the potential to be brought to more JDBC database types.  It is based on concepts that should generally be available within database supporting JDBC drivers and should be potentially able to be ported to more things. I am fairly certain that was front and center in mind during its design.

I can also argue that my implementation has a certain simplicity in that it is essentially a pass-through mechanism to allow Artemis in Java to utilize a clusterable mutex provided by the database -- which is exactly the concept sought after in the pluggable quorum voting concept with the DistributedPrimitiveManager base class.

I suspect that someone who knows JdbcSharedStateManager better than me could utilize it to implement a DatabasePrimitiveManager that only interacted with the database via JDBC specification calls and didn't use specific implementation features.

rvais:
@funkyjive would you agree that your implementation could be changed/updated to work as an implementation of potential locking mechanism for the jdbcSharedStateManager to leverage? What I, personally would like to see is simplified and unified approach. What came to my mind when reading your PR was that you could use the existing API for the locking mechanism, and just fill in your approach to locks themselves.

funkyjive:
@rvais Truth is I don't know if I really do agree with that.  I coded an implementation oriented around of the  DistributedPrimitiveManager specification using specialized database features.

I don't know that attempting to utilize those features and approaches inside JdbcSharedStateManager ultimately does create something "simplified and unified". Those underlying mechanisms are actually quite different from anything in JdbcSharedStateManager.  For instance, go look at any method in JdbcSharedStateManager -- for instance "writeState" or "readState".  The entirety of the implementation is oriented around managing a transaction and its state with the database.  But none of that correlates with the database specific features in my implementation -- there aren't any transactions, just remote locks held in the database itself.

So to use JdbcSharedStateManager and utilize this other form of database based locking it would have to be refactored and generalized.  But to what end?  What would do we get out of it? If JdbcSharedStateManager already supports the databases (Oracle,MSSQL, Postgres, MySQL), why even do it?

So if the consensus thinking is:  
"We like the idea of a DatabasePrimitveManager but we think that the implementation should utilize the existing JdbcSharedStateManager to achieve that end",  I can accept that.  Fundamentally, I believe that Artemis *should* be capable of having two brokers do high availability w/ replication (no need for shared SAN) and still be able to avoid "split-brain" through utilization of a database that is already available from the application.  So however that can be achieved.

I'm not going to do an implementation with JdbcSharedStateManager because my needs have changed.  I am not attempting to use replication anymore. I just wanted to offer this concept I researched before I moved on.  

From what I can tell, JdbcSharedStateManager is not exactly what is called for in the DatabasePrimitiveManager spec because there is all kinds of other state information in there ("Live", "Falling Back", "Paused").  I think it was written with a different state and purpose in mind.  You will likely have to do some additional things to "fit it in" anyway.  I could be wrong, but it doesn't seem like it will be simpler to me -- but I don't have mastery on all the thinking and background that went into the creation of JdbcSharedStateManager.

PR #4407, opened 2023-03-17
NO-JIRA Cleaning up test suite

clebertsuconic:


clebertsuconic:
Derby was leaking.
all tests were issuing derby shutdown even if not using them.
a test was not using derby at all (JMSStorageTest) since we don't have a JMSSTorage implementation (just regular StorageManager).

clebertsuconic:
image

clebertsuconic:
> image

this was before the updates..

clebertsuconic:
This is after these updates. 

image

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -235,29 +222,40 @@ public MBeanServer getMBeanServer() {
...
+         runAfter = Collections.synchronizedList(new ArrayList<>());
       }
       runAfter.add(run);
    }
 
    @After
    public void runAfter() {
-      if (runAfter != null) {
-         runAfter.forEach((r) -> {
+      List<RunnableEx> theRunafterList = runAfter;
```
gemmellr, line 237: Perhaps localRunAfter[List] ?

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -235,29 +222,40 @@ public MBeanServer getMBeanServer() {
...
+      if (theRunafterList != null) {
+         // I do this replacement as eventually a runAfter lambda could be adding another runAfter.
+         // this shouldn't happen very often but I saw it happening once.
+         // it's better to be protected against
```
gemmellr, line 241: What did you see doing it? Is it really always invalid? If it is, then rather than silently drop it like this does, it might be better to be aware of it happening...e.g set a boolean noting the 'runAfter' things are starting to execute and then it can at least throw or log if anything calls runAfter(Runnable) to add a new task, and not add the task, rather than just allowing it to create a new collection and then ignore its contents before nulling it again as this change seems to.

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -235,29 +222,40 @@ public MBeanServer getMBeanServer() {
...
+         runAfter = null;
```
gemmellr, line 242: If doing the above suggestion, this could just be moved to straight after the initial local-assignment and done once regardless.

The runAfter() method should perhaps be synchronized, as the runAfter(Runnable) method is and yet still isnt safe despite it due to this nulling and the variable being made volatile.

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
clebertsuconic: I have a case where destroyDerby is being called as runAfter();
destroyDerby is calling the getURI, which will then turn runAfter back on.


On my case I had explicitly set runAFter() to be called even though JDBC was not used.


So, it was an artificial case, but I thought It would be better to protect it since I saw it happening.


all the calls to runAfter are from single threaded. I thought when I added the volatile that it was multi-threaded.. but it's not the case.

gemmellr: AH! So I resisted the temptation to comment when passing that getter that I really didnt like the method resulting in scheduling stuff.  But I totally overloooked that was this issue. lol.

That seals it, I really dont think that URI getter should be doing that, something else should be specifically opening the DB, which calls the URI getter.

artemis-server/src/test/java/org/apache/activemq/artemis/core/server/impl/jdbc/JdbcNodeManagerTest.java
```diff
@@ -56,29 +57,34 @@ public void configure() {
...
+      } else {
+         try {
+            DriverManager.getConnection("jdbc:derby:" + getEmbeddedDataBaseName() + ";drop=true");
+         } catch (Exception e) {
+            logger.info(e.getMessage());
+         }
+         try {
+            DriverManager.getConnection("jdbc:derby:;shutdown=true;deregister=false");
+         } catch (Exception e) {
+            logger.info(e.getMessage());
```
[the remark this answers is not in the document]
clebertsuconic: I actually added code on the upper method for this...

let me check if I'm doing it right

PR #4410, opened 2023-03-22
NO-JIRA doc update added some Broker Properties reference and added s…

andytaylor:
…ome docs to the schema


brusdev:
@andytaylor could you create a JIRA to track this change?

docs/user-manual/en/configuration-index.md
```diff
@@ -118,6 +118,9 @@ Enable debug logging for `org.apache.activemq.artemis.core.config.impl.Configura
 
 There are a growing number of examples of what can be explicitly configured in this way in the [unit test](https://github.com/apache/activemq-artemis/blob/065bfe14f532858f2c2a20b0afb1a226b08ce013/artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java#L675).
 
+For an incomplete list of some of the properties that are available and have been tested see [Broker Properties Reference](#broker-properties-reference)
```
brusdev, line 121:
```suggestion
For a partial list of some of the supported properties that have been tested see [Broker Properties Reference](#broker-properties-reference)
```

docs/user-manual/en/configuration-index.md
```diff
@@ -481,3 +484,285 @@ Name | Description | Default
 [password](amqp-broker-connections.md#amqp-server-connections) | Broker authentication password (optional) | n/a
 [reconnect-attempts](amqp-broker-connections.md#amqp-server-connections) | How many attempts should be made to reconnect after failure. | -1 (infinite)
 [auto-start](amqp-broker-connections.md#amqp-server-connections) | Broker connection starts automatically with broker | true
+
+
+## Broker Properties Reference
```
brusdev, line 489: This list could become stale, could we generate it from the sources?

andytaylor: however that turned out to be a huge bit of work
which is going to take a long time to get to  a point where it works and is usable
hence why I updated the schema
as this is where I drive it from

docs/user-manual/en/configuration-index.md
andytaylor: updated as suggested

docs/user-manual/en/configuration-index.md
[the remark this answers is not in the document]
rvais: As there is no specific pattern to these (as far as I recall from conversations with Gary), coming up with the explicit naming convention could prove difficult. It's one of the things I've tried to push for ...

Disclaimer about precedence / priority for application of configuration would be definitely beneficial.

jbertram:
Please re-open with conflicts resolved if this is still something worth contributing.  Thanks!

PR #4413, opened 2023-03-24
ARTEMIS-4215 JournalFlush might never happen when journal-sync-* is f…

AntonRoskvist:
…alse

I'm very open to the idea that this might not be the right solution but if nothing else the test works to demonstrate the issue.
Can also very easily be demonstrated by:

configuring <journal-sync-transactional>false</journal-sync-transactional>
                   <journal-sync-non-transactional>false</journal-sync-non-transactional>
Starting the broker
Sending some messages i.e:
bin/artemis producer --destination queue://TEST.1 --message-count 100
Wait _any_ amount of time
kill -9 artemisPID
Start broker
Messages are gone

This is new behavior since broker version 2.18.0

I have run some local performance tests between this change and an unchanged broker and can see no meaningful difference in terms of resource utilization or overall throughput.

I have tested with sync=true and false, 
paging 50% of the messages
Running all message production in transacted batches of 10 messages each
Running all message production with AUTO_ACK
Results of the tests:

```
1M messages
max-messages=500k
  sync=false
    batch
      PR:  231995 ms
      ORG: 234791 ms
    AUTO_ACK
      PR:  33627 ms
      ORG: 31928 ms
  sync=true
    batch
      PR:  254064 ms
      ORG: 247203 ms
    AUTO_ACK
      PR:  2438805 ms
      ORG: 2439546 ms
```

### DIFF
--- a/artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/buffer/TimedBuffer.java
+++ b/artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/buffer/TimedBuffer.java
@@ -271,9 +271,10 @@ public final class TimedBuffer extends CriticalComponentImpl {
 
             if (sync) {
                pendingSync = true;
-
-               startSpin();
             }
+
+            startSpin();
+
          }
       }
    }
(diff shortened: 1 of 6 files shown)


clebertsuconic:
@AntonRoskvist I was going to merge this today.. however I can't figure out yet why InfiniteRedeliverySmokeTest is failing. It's creating too many files for some reason.

clebertsuconic:
never mind.. I figured out

PR #4415, opened 2023-03-27
ARTEMIS-4218 Support console smoke tests from remote servers

brusdev:


brusdev:
@gemmellr thanks for your feedback, I figured out an alternative solution to support connections from remote servers preserving the default restrictions

PR #4418, opened 2023-03-28
ARTEMIS-4206 Adding debug support for large messages not being removed

clebertsuconic:
you add the property -DARTEMIS_DEBUG_REF=true in your java arguments, And you should see Debug output when the messages are not being removed.


artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
```diff
@@ -49,6 +76,28 @@ public int getDurableCount() {
...
+   public RefCountMessage() {
+      if (DEBUG_REF_COUNT) {
+         AtomicBoolean refFired = this.firedAtomic;
+         List<Exception> crumbs = debugCrumbs;
+         crumbs.add(new Exception("new Instance"));
+         String clazz = this.getClass().getName();
+         // notice you can't have a direct reference on the LargeMessage Itself,
+         // Otherwise you would get a leak from itself
+         // RefCountMessageLeakTest is validating this scenario
+         CLEANER.register(this, () -> debug(clazz, refFired, crumbs));
```
tabish121, line 97: Even with the comment I'm not a huge fan of the lambda being used here because it is super easy for someone to later change this to capture 'this' inadvertently.  I'd recommend passing the state data to a static function that returns a Runnable just as an additional safety mechanism as that prevents a capture of 'this' in the Runnable.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
```diff
@@ -58,54 +107,42 @@ public int getDurableCount() {
...
-   }
 
+   private static void debug(String clazz, AtomicBoolean firedAtomic, List<Exception> crumbs) {
+      if (!firedAtomic.get()) {
+         StringWriter writer = new StringWriter();
+         PrintWriter out = new PrintWriter(writer);
+         crumbs.forEach(e -> e.printStackTrace(out));
+         ActiveMQClientLogger.LOGGER.debugMessageNotReleased(clazz, writer.toString());
+      }
+   }
```
tabish121, line 137: Extra newline to remove.

tests/leak-tests/src/test/java/org/apache/activemq/artemis/tests/leak/RefCountMessageLeakTest.java
```diff
@@ -0,0 +1,97 @@
...
+
+      @Override
+      public String toString() {
+         return "debugMessage(" + string + ")";
+      }
+
+      public void fired() {
+         this.markFired();
+      }
+
```
tabish121, line 52: Unnecessary extra newlines to remove.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Test.java
tabish121, line 1: This seems like an odd thing to be adding to the main jar for the core client, surely you don't need to add test classes into the client ?

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
```diff
@@ -16,13 +16,40 @@
...
 
-// import org.apache.activemq.artemis.utils.collections.ConcurrentHashSet; -- #ifdef DEBUG
+import org.apache.activemq.artemis.core.client.ActiveMQClientLogger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class RefCountMessage {
 
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
```
tabish121, line 36: Extra newline to remove.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
clebertsuconic: I actually amended this with a State Object.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Test.java
clebertsuconic: removed... it was a mistake.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
```diff
@@ -49,6 +93,28 @@ public int getDurableCount() {
       return DURABLE_REF_COUNT_UPDATER.get(this);
    }
 
+   /** Sub classes may mark fired=true when they were explicitly called.
+    * E.g large message removed the file upon cancellation */
+   protected void markFired() {
+      fired = true;
```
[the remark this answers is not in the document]
clebertsuconic: if the system called the fire method directly, say, a cancellation from a bad producer, there's no need to call the fire event again on countDown... (which might not actually happen since the message is going to be GCed).


I need the variable in the status as well as I can't have access to the message from the status.

tests/soak-tests/pom.xml
```diff
@@ -260,7 +260,7 @@
                      <configuration>${basedir}/target/classes/servers/interruptlm</configuration>
                      <args>
                         <arg>--java-options</arg>
-                        <arg>-Djava.rmi.server.hostname=localhost</arg>
+                        <arg>-Djava.rmi.server.hostname=localhost -DARTEMIS_DEBUG_REF=true</arg>
```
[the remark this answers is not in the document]
clebertsuconic: I actually wanted to have the soak tests to provide information about leaked messages if the test failed.

tests/leak-tests/pom.xml
```diff
@@ -246,7 +246,8 @@
             <artifactId>maven-surefire-plugin</artifactId>
             <configuration>
                <skipTests>${skipLeakTests}</skipTests>
-               <argLine>-agentpath:${project.basedir}/target/lib/${check-leak-deploy-name} -Djgroups.bind_addr=::1 ${activemq-surefire-argline} -Dorg.apache.activemq.SERIALIZABLE_PACKAGES="java.lang,javax.security,java.util,org.apache.activemq,org.fusesource.hawtbuf"</argLine>
+               <!-- -DARTEMIS_DEBUG_REF is needed for RefCountMessageLeakTest -->
+               <argLine>-agentpath:${project.basedir}/target/lib/${check-leak-deploy-name} -DARTEMIS_DEBUG_REF=true -Djgroups.bind_addr=::1 ${activemq-surefire-argline} -Dorg.apache.activemq.SERIALIZABLE_PACKAGES="java.lang,javax.security,java.util,org.apache.activemq,org.fusesource.hawtbuf"</argLine>
```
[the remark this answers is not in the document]
clebertsuconic: @gemmellr separate config is an overkill on this... there should be only one test validating for large message.


I didn't want to have a volatile configuration (meaning I could enable it just for the test), but I may have to do it

tests/soak-tests/pom.xml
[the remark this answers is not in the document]
gemmellr: Its no longer soak testing the normal use case at all as a result though, that just seems wrong.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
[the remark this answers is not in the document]
gemmellr: So is the idea that during the 'normal' countDown it marks the 'outer fired' variable, reaches out somewhere else in firing releaseComplete(), then that other places [maybe] reaches back in here eventually and calls 'markFired()' and doubly-sets the 'outer fired' variable and ALSO sets the separate 'debugStatus fired' variable to _really_ mark the message as cleaned up? The cleaner only seems to look at the latter variable when deciding whether the message was cleaned up or not, but only the markFired() 'direct call' is setting that. I looked at/thought about this code for 30mins earlier and am asking questions as I still dont understand what its expected to do normally. Its pretty impenetrable as-is. If I wasnt seeing the diff I probably wouldnt even have noticed the 2 different 'fired' variables, with one being used in 2 different ways by the two routes through the code.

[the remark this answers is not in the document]
clebertsuconic: The first variable is used to control the releaseResources...

The second variable is used to control if the resource leaked or not.


I'm renaming it to leaked.




and I can't read anything from RefCountMessage coming from RefCountDebugState (I just renamed it BTW).

Having a reference back to RefCountMessage would make it impossible to release the reference. and I don't want to complicate things with a WeakReference... I would rather duplicate the variable... in this case I renamed it to leaked to make a distinction.

tests/soak-tests/pom.xml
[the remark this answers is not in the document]
clebertsuconic: This was for the server running only. not the actual test.. 

But I have removed it anyway.

tests/leak-tests/pom.xml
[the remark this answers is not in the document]
clebertsuconic: I'm using a variable to set this to true now. The debug was actually wrong.. as ReferenceCount is also used for CoreMessage.

clebertsuconic:
I have been debugging with these messages, and I am making a few adjustments (Missed debug statements... etc)... this is being useful in a scenario where the Cpu Is overloaded and everything is failing on the delivering, leaving messages behind.

it is working as I expected to tell me what is happening, but I'm making a few adjustments.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
```diff
@@ -58,55 +108,35 @@ public int getDurableCount() {
...
+      if (debugStatus != null) {
+         debugStatus.up(debugString());
+      }
+   }
 
-   private void onUp() {
-      // #ifdef DEBUG -- comment out anything before endif if you want to debug REFERENCE COUNTS
-      // upSet.add(new Exception("upEvent(" + debugString() + ")"));
-      // #endif
+   protected void onFired() {
```
gemmellr, line 118: Perhaps onDelete[d]? onRelease?

Calling it onFired() seems weird since its so often called when the 'releaseComplete()' being gated by the 'fired' variable apparently _has not_ been fired.

gemmellr (commented):
I know its still draft but as its been a couple days and you keep updating it..some WIP comments.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/ActiveMQClientLogger.java
```diff
@@ -344,4 +344,7 @@ public interface ActiveMQClientLogger {
 
    @LogMessage(id = 214033, value = "Cannot resolve host ", level = LogMessage.Level.ERROR)
    void unableToResolveHost(UnknownHostException e);
+
+   @LogMessage(id = 214034, value = "{} is being garbage collected without removing its resources.:\nlocations:\n{}", level = LogMessage.Level.INFO)
+   void debugMessageNotReleased(String clazz, String debugString);
```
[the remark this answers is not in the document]
clebertsuconic: RefCountMessage is at Client.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
clebertsuconic: I am going to rename this as accountedFor...


although I'm improving this... I'm only going to report an issue if there was at least on onUp, so I will not need this as often... probably it will be called by onFired only.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/ActiveMQClientLogger.java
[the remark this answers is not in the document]
clebertsuconic: I tried moving CoreMessage to the server, but that would mean having to send the Server jar to the clients...


I have been there before. this was the best compromise some time ago.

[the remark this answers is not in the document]
gemmellr: I hadnt actually noticed that those methods are defined in the Message interface, since RefCountMessage doesnt actually indicate it implements anything itself...but things subclass it to effectively implement the Message methods, which makes it rather hard to see whats using it and what the original impl is when the Message method is being overridden in various places.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/LargeServerMessage.java
```diff
@@ -27,6 +27,8 @@
 
 public interface LargeServerMessage extends ReplicatedLargeMessage {
 
+   boolean DEBUG = Boolean.parseBoolean(System.getProperty("ARTEMIS_LARGE_MESSAGE_DEBUG", "false"));
```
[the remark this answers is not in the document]
clebertsuconic: I will just use logger.isDebugEnabled();

if RefCountMessage.logger is debug, it will be using debug


Maybe I will also expose the logger from RefCountMessage with RefCountMessage::isDebugEnabled()

As I will be adding a method refCountMessage.addDebug(String );

(Not sure I will keep this method in the final version after I figure out the issue I'm dealing with now...
  apparently a message could be duplicated after a failed AMQP Consumer in some weird race, and if I add logger.debug() the error is gone... the only way I have been able to capture this was by adding a new event to the debugState, and checking if I'm trying to remove the message twice.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
clebertsuconic: I have pushed my latest changes now.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/ObjectCleaner.java
```diff
@@ -0,0 +1,32 @@
...
+import java.lang.ref.Cleaner;
+
+public class ObjectCleaner {
+
+   private static final Cleaner cleaner;
+   static {
+      cleaner = Cleaner.create();
+   }
+
+   public static void register(Object obj, Runnable callback) {
```
tabish121, line 29: An issue here with this Cleaner wrapper is that it doesn't really expose a cleaner API that is compatible with the JDK Cleaner appropriate usage.  You are meant to take the returned Cleanable from the Cleaner API and actually call clean on it in most cases leaving the Cleaner to handle those rare or missed cases where a cleanable resource wasn't actively closed and "cleaned" up by the user code.  Just putting every resource into a Cleaner and waiting on it to actually manage the cleanup will likely result in untimely reclaim of resources and bloat in the tracking structures.

tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/interruptlm/LargeMessageFrozenTest.java
```diff
@@ -0,0 +1,269 @@
...
+         consumer.receive(100);
+         connection.createSession(false, Session.AUTO_ACKNOWLEDGE); // just to force an exception
+      } catch (Exception expected) {
+         logger.info(expected.getMessage(), expected);
+         failed = true;
+      }
+
+      Wait.assertEquals(0, () -> server.getActiveMQServerControl().getConnectionCount());
+
+      Thread.sleep(500);
```
clebertsuconic, line 358: This sleep is not needed, I'm removing it.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
```diff
@@ -49,6 +172,13 @@ public int getDurableCount() {
       return DURABLE_REF_COUNT_UPDATER.get(this);
    }
 
+   protected void registerDebug() {
+      if (debugStatus == null) {
+         debugStatus = new DebugState(this.toString());
+         ObjectCleaner.register(this, debugStatus);
+      }
+   }
```
gemmellr, line 180: As this is basically a separated-out part of the constructor, to be used by specific subclasses, it should probably go next to it, and maybe note that.

tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/interruptlm/LargeMessageFrozenTest.java
clebertsuconic: ... removed

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1130,6 +1136,10 @@ public void addHead(final MessageReference ref, boolean scheduling) {
    /* Called when a message is cancelled back into the queue */
    @Override
    public void addSorted(final MessageReference ref, boolean scheduling) {
+
+      if (RefCountMessage.isTraceEnabled()) {
+         RefCountMessage.deferredDebug(ref.getMessage(), "returning to queue {}", this.getAddress());
+      }
```
gemmellr, line 1142: The newline should be between the two entirely separate if's rather than before them.

artemis-unit-test-support/src/main/java/org/apache/activemq/artemis/utils/ThreadLeakCheckRule.java
```diff
@@ -284,8 +284,10 @@ private boolean isExpectedThread(Thread thread) {
       } else if (threadName.contains("GC Daemon")) {
          return true;
       } else {
+         // validating for known stack traces
          for (StackTraceElement element : thread.getStackTrace()) {
-            if (element.getClassName().contains("org.jboss.byteman.agent.TransformListener")) {
+            if (element.getClassName().contains("org.jboss.byteman.agent.TransformListener") ||
+                element.getClassName().contains("dk.internal.ref.CleanerImpl")) {
```
gemmellr, line 290: jdk?

gemmellr (commented):
You were right, I didnt like that there were two sets of changes in here, intertwined no less :)

I'm not a huge fan of using the log levels to decide to randomly start changing behaviour and stashing stacktraces, even if the intent is sometimes log them...or of using that mechanism throughout the codebase rather than just inside the message classes as it was originally.

clebertsuconic:
I can squash 2 commits in one for two different and related issues.  I thought it was better to keep separated.  

Regarding the trace. I think it's better this way.  Say an user starts having issues if we just set the logger we would get what we need.

clebertsuconic:
This mechanism is how I found these issues.  I would like to keep this door open for future issues.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
clebertsuconic: There's no constructor defined on this class. I'm moving it to the top of the class with a note about it.

clebertsuconic:
The idea of this feature is to be a hidden feature...

So, I am going to use the following logger to enable this type of debugging:

   private static final Logger refLogger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass().getName() + ".REF_DEBUG");

gemmellr:
> I can squash 2 commits in one for two different and related issues. I thought it was better to keep separated.

No, lets not squash 'different changes' into one commit (yes I saw, its too late), lets keep different changes separate. In that spirit, the original comment was around lets also try not to make different changes entirely reliant on each other anyway by intertwining and overlapping their multiple commits, such that you cant very simply pick or see the effect of one of the different changes in isolation because the commits were mixed together rather than sequential while adding.

Aside, on the squashed commit log message, it starts telling you how to enable the ref counting debug, and then also ends telling you how to enable the ref counting debug but in a less clear way. Seems like one of them could go.

gemmellr:
> The idea of this feature is to be a hidden feature...
> 
> So, I am going to use the following logger to enable this type of debugging:
> 
> private static final Logger refLogger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass().getName() + ".REF_DEBUG");

It will certainly be hidden, people will enable it without knowing they are since it is a descendent in the regular class name logger hierarchy. Anyone simply enabling debug logging for the class (or more likely, many classes / further up the hierarchy) will get the change in actual system behaviour of the ref counting debug and saving lots of stacktraces. They would have to know about the change in behaviour and reconfigure the specific ref count logger back off in order to restore the prior behaviour and only add debug output for the normal system behaviour.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
```diff
@@ -815,7 +815,7 @@ void slowConsumerDetected(String sessionID,
    @LogMessage(id = 222168, value = "The '" + TransportConstants.PROTOCOL_PROP_NAME + "' property is deprecated. If you want this Acceptor to support multiple protocols, use the '" + TransportConstants.PROTOCOLS_PROP_NAME + "' property, e.g. with value 'CORE,AMQP,STOMP'", level = LogMessage.Level.WARN)
    void warnDeprecatedProtocol();
 
-   @LogMessage(id = 222169, value = "You have old legacy clients connected to the queue {} and we can't disconnect them, these clients may just hang", level = LogMessage.Level.WARN)
+   @LogMessage(id = 222169, value = "You have old legacy clients connected::{} and we can't disconnect them, these clients may just hang", level = LogMessage.Level.WARN)
```
gemmellr, line 818: the "::" seems a bit odd?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Consumer.java
```diff
@@ -93,6 +93,8 @@ default Binding getBinding() {
     */
    void disconnect();
 
+   void failed(Throwable message);
```
gemmellr, line 96: 'message seems odd...'t' or 'cause' or something would seem a better fit

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/Redistributor.java
```diff
@@ -135,6 +140,12 @@ public void proceedDeliver(MessageReference ref) {
       // no op
    }
 
+
+   @Override
+   public void failed(Throwable message) {
```
gemmellr, line 145: Ditto

michaelandrepearce:
> The idea of this feature is to be a hidden feature...
> 

Hmmm this worries me, I think any feature should be something we're open about have the capability documented etc. We shouldn't be hiding things so only select few know about it.

Fair enough call it like advanced troubleshooting or something but do not add something hidden

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1108,6 +1109,11 @@ public void addHead(final MessageReference ref, boolean scheduling) {
       if (logger.isTraceEnabled()) {
          logger.trace("AddHead, size = {}, intermediate size = {}, references size = {}\nreference={}", queueMemorySize, intermediateMessageReferences.size(), messageReferences.size(), ref);
       }
+
+      if (RefCountMessage.isTraceEnabled()) {
+         RefCountMessage.deferredDebug(ref.getMessage(), "AddHead queue {}", this.getName());
```
gemmellr, line 1114: The others seem to be using more of a sentence structure, i.e "add head"

Edit...following the next comment, could alternatively go the other way, and make them all method-name structured.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1130,9 +1136,14 @@ public void addHead(final MessageReference ref, boolean scheduling) {
    /* Called when a message is cancelled back into the queue */
    @Override
    public void addSorted(final MessageReference ref, boolean scheduling) {
+      if (RefCountMessage.isTraceEnabled()) {
+         RefCountMessage.deferredDebug(ref.getMessage(), "returning to queue {}", this.getAddress());
```
gemmellr, line 1140: There are a few "returning to queue" events. Although they can be separated by the stacktrace detail it would probably be nicer to make the text more distinct to begin with, and help out the poor sod digging through streams of stacktraces later.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -1078,21 +1088,26 @@ public synchronized MessageReference removeReferenceByID(final long messageID) t
          // This is an optimization, if the reference is the first one, we just poll it.
          // But first we need to make sure deliveringRefs isn't empty
          if (deliveringRefs.isEmpty()) {
+            if (logger.isTraceEnabled()) {
+               logger.trace("removeReferenceByID {} return null", messageID);
+            }
```
gemmellr, line 1093: Seems odd to gate this one, and not gate the few similar ones added in the next 20 lines below.

gemmellr:
> > The idea of this feature is to be a hidden feature...
> 
> Hmmm this worries me, I think any feature should be something we're open about have the capability documented etc. We shouldn't be hiding things so only select few know about it.
> 
> This is ethos of open source Apache in my mind.
> 
> Fair enough if it's something a general user may not need to know, call it like advanced troubleshooting or something but do not add something hidden

Yeah, its not a 'feature-feature' in any way, and not something I think a user would normally care about so I dont think I'd ever document it personally. Its 'additional broker developer leak debug/trace details via saved stacktraces'.

clebertsuconic:
@gemmellr I will keep the squash.. they are all related anyways.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
```diff
@@ -16,13 +16,109 @@
...
+
+   public static boolean isTraceEnabled() {
+      return refLogger.isTraceEnabled();
+   }
```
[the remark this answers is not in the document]
clebertsuconic: funny, I actually just renamed these methods as isRefDebugEnabled and isRefTraceEnabled without reading this comment.

clebertsuconic:
@michaelandrepearce I didn't mean as an actual hidden property. I will have a JIRA open for this that will be part of the release notes, and this is actually open source.

Just that I won't add it to the documentation... as most users won't care about it.. and hopefully won't need to.


Over the years I have been uncommenting code when I needed to debug things... I think it's time to make that thing as part of the codebase now.

PR #4423, opened 2023-03-30
ARTEMIS-4210 refactor connection audit logging

jbertram:
This commit fixes the following things:

 - Moves connection audit logging to the resource audit logger instead of using a dedicated logger as that would adversely impact upgrading users, and arguably didn't make sense in the first place.
 - Mitigates an potential NPE w.r.t. connection ID.
 - Updates the "dummy" management connection to return a valid connection ID.


artemis-commons/src/main/java/org/apache/activemq/artemis/logs/AuditLogger.java
```diff
@@ -2646,14 +2641,14 @@ static void isAutoDelete(Object source) {
    void isAutoDelete(String user, Object source);
 
    static void createdConnection(String protocol, Object connectionID, String remoteAddress) {
-      CONNECTION_LOGGER.createdConnection(protocol, connectionID.toString(), String.format("unknown%s", formatRemoteAddress(remoteAddress)));
+      RESOURCE_LOGGER.createdConnection(protocol, connectionID == null ? null : connectionID.toString(), String.format("unknown%s", formatRemoteAddress(remoteAddress)));
```
clebertsuconic, line 2644: String.valueOf(connection) would been easier

jbertram: Done.

PR #4426, opened 2023-04-04
ARTEMIS-4235 fix map msg conversion from OpenWire->core

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageConsumerTest.java
```diff
@@ -193,4 +195,66 @@ private void testDurableSubscriptionWithConfigurationManagedQueue(ConnectionSupp
...
+   public void testMapMessageConversionBetweenCoreAndOpenWire() throws Exception {
+      testMapMessageConversion(createCoreConnection(), createOpenWireConnection());
+   }
+
+   @Test(timeout = 30000)
+   public void testMapMessageConversionBetweenOpenWireAndCore() throws Exception {
+      testMapMessageConversion(createOpenWireConnection(), createCoreConnection());
+   }
+
+   private void testMapMessageConversion(Connection sender, Connection consumer) throws Exception {
```
gemmellr, line 229: Something like senderConnection and consumerConnection would seem more descriptive...you could then follow through with 'sender[Foo]' and 'consumer[Foo]' convention for related objects and wouldnt need to call the sole consumer "consumer2".

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageConsumerTest.java
```diff
@@ -193,4 +195,66 @@ private void testDurableSubscriptionWithConfigurationManagedQueue(ConnectionSupp
...
+      final String NAME = "myPropertyName";
+      final String VALUE = RandomUtil.randomString();
+
+      try {
+         Session session1 = sender.createSession(false, Session.AUTO_ACKNOWLEDGE);
+         Session session2 = consumer.createSession(false, Session.AUTO_ACKNOWLEDGE);
+
+         javax.jms.Queue queue1 = session1.createQueue(getQueueName());
+         javax.jms.Queue queue2 = session2.createQueue(getQueueName());
```
gemmellr, line 238: Should be able to use the same destination object for both, but if wanting to avoid it then just inlining the creations might be neater given the variables arent reused, avoiding them giving impression there are 'two queues' when its just one.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageConsumerTest.java
```diff
@@ -193,4 +195,66 @@ private void testDurableSubscriptionWithConfigurationManagedQueue(ConnectionSupp
...
+
+         MessageProducer producer = session1.createProducer(queue1);
+         producer.setDeliveryMode(DeliveryMode.PERSISTENT);
+         sender.start();
+
+         MapMessage message = session1.createMapMessage();
+         message.setStringProperty(NAME, VALUE);
+         producer.send(message);
+
+         Message received = consumer2.receive(100);
```
gemmellr, line 250: 100 seems a little low for a message we do expect to receive, could be unreliable in slow CI envs.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -111,9 +111,13 @@ public static org.apache.activemq.artemis.api.core.Message inbound(final Message
...
       final ByteSequence contents = messageSend.getContent();
-      if (contents == null && coreType == org.apache.activemq.artemis.api.core.Message.TEXT_TYPE) {
-         body.writeNullableString(null);
-      } else if (contents != null) {
+      if (contents == null) {
+         if (coreType == org.apache.activemq.artemis.api.core.Message.TEXT_TYPE) {
+            body.writeNullableString(null);
+         } else if (coreType == org.apache.activemq.artemis.api.core.Message.MAP_TYPE) {
+            body.writeByte(DataConstants.NULL);
+         }
```
gemmellr, line 119: The new tests dont appear to cover this case, the main addition of the PR and how the issue was found?

Also feels like this could be unit tested rather than system tested.

jbertram: Not all of the tests cover this specific issue. Most of them, in fact, cover other uses cases involving the same JMS code with different protocols (i.e. core and AMQP). However, I felt it was worth confirming those work as expected.

That said, both `testMapMessageConversionBetweenOpenWireAndAMQP` and `testMapMessageConversionBetweenOpenWireAndCore` fail without the fix.

I'll add a unit test for good measure.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageConsumerTest.java
jbertram: Will do.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageConsumerTest.java
jbertram: That's fair.

gemmellr:
> Not all of the tests cover this specific issue. Most of them, in fact, cover other uses cases involving the same JMS code with different protocols (i.e. core and AMQP). However, I felt it was worth confirming those work as expected.
> 

Yep, wasnt arguing against testing other things, just noting that it didnt look like it tested the reported issue and main thrust of the change.

> That said, both `testMapMessageConversionBetweenOpenWireAndAMQP` and `testMapMessageConversionBetweenOpenWireAndCore` fail without the fix.
> 
> I'll add a unit test for good measure.

Interesting, I'm actually surprised to hear those failed in the reported manner given the description and earlier discussion on the JIRA. It sounded like it was the handling of an empty/null map message bodies at issue, and the code change looked to be along lines of 'treat null map message contents as being empty', yet as the test then wasnt using an empty map message at all it just didnt seem like it was exercising the reported issue / main change at all.

So, does that actually mean that no Openwire map messages worked if being converted?

artemis-protocols/artemis-openwire-protocol/src/test/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverterTest.java
```diff
@@ -135,6 +138,15 @@ public void testProperties() throws Exception {
       }
    }
 
+   @Test
+   public void testEmptyMapMessage() throws Exception {
+      CoreMessage artemisMessage = (CoreMessage) OpenWireMessageConverter.inbound(new ActiveMQMapMessage().getMessage(), openWireFormat, null);
+      ActiveMQBuffer buffer = artemisMessage.getDataBuffer();
+      TypedProperties map = new TypedProperties();
+      buffer.resetReaderIndex();
+      map.decode(buffer.byteBuf());
```
gemmellr, line 147: Might be nice to check the CoreMessage actually represents a MapMessage, and maybe what comes out of the decode matches expectations of what went in, i.e verify some behaviours of the conversion and not just simply that it didnt throw. Essentially most of what the systests is really doing around verifying the area/behaviour changed, but much faster.

gemmellr:
> .. as the test then wasnt using an empty map message at all it just didnt seem like it was exercising the reported issue / main change at all.
> 

Yeah, so...its been pointed out to me that it wasnt adding anything to the map...just the properties...which somehow I was completely oblivious to. Oopsie :)

tabish121:
> > .. as the test then wasnt using an empty map message at all it just didnt seem like it was exercising the reported issue / main change at all.
> 
> Yeah, so...its been pointed out to me that it wasnt adding anything to the map...just the properties...which somehow I was completely oblivious to. Oopsie :)

It might be good to check that an totally empty message (no properties or body) was handled as well, or check that it was tested someplace else, possible a check for each type although I think the Openwire marshaler handles the "content" portion the same in each message type.

artemis-protocols/artemis-openwire-protocol/src/test/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverterTest.java
jbertram: Done.

jbertram:
> So, does that actually mean that no Openwire map messages worked if being converted?

@gemmellr, no. I added a handful of tests exercising a non-empty `MapMessage` and all those tests succeed even without the fix.

jbertram:
> It might be good to check that an totally empty message (no properties or body) was handled as well...

@tabish121, I removed the property from the "empty" test-cases to make it less confusing and mimic the original condition more closely. I also added some "non-empty" tests for good measure.

PR #4429, opened 2023-04-07
ARTEMIS-4234 Fix some issue with Embedded Resource send and receive

tabish121:
The JUnit resource has a couple bugs in both the send and receive path that result in only one message being received if the receive method is called repeatedly and some send drop the provided properties. Cleaned up some tests to the point of showing the errors and ensuring that at least basic functionality is tested.

### DIFF
--- a/artemis-junit/artemis-junit-commons/src/main/java/org/apache/activemq/artemis/junit/EmbeddedActiveMQDelegate.java
+++ b/artemis-junit/artemis-junit-commons/src/main/java/org/apache/activemq/artemis/junit/EmbeddedActiveMQDelegate.java
@@ -471,7 +471,6 @@ public class EmbeddedActiveMQDelegate implements EmbeddedActiveMQOperations {
 
    @Override
    public ClientMessage sendMessageWithProperties(SimpleString address, String body, Map<String, Object> properties) {
-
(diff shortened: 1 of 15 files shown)


artemis-junit/artemis-junit-commons/src/main/java/org/apache/activemq/artemis/junit/EmbeddedActiveMQDelegate.java
```diff
@@ -626,43 +625,51 @@ public void sendMessage(SimpleString address, ClientMessage message) {
...
-         } else {
-            try {
-               message = consumer.receive();
-            } catch (ActiveMQException amqEx) {
-               throw new EmbeddedActiveMQResourceException(String.format("ClientConsumer.receive() for %s failed",
-                                                                         address.toString()),
-                                                           amqEx);
+
+            return message;
+         } catch (ActiveMQException amqEx) {
```
gemmellr, line 665: It only seems to look at and throws 'failureCause' if it catches ActiveMQException in the outer catch wrapper. All the inner catch wrappers swallow ActiveMQException while setting failureCause, meaning that failureCause then wont be looked at since if they failed 'message' will just be null and the overall method will have successfully returned null just above this.

PR #4430, opened 2023-04-07
ARTEMIS-4171/ARTEMIS-4233 Large Message Send

clebertsuconic:
This is not really supposed to happen but I'm dealing with it just in case. In case it happened I would like to see the message removed and an exception thrown to the client

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/RefCountMessage.java
@@ -139,6 +139,8 @@ public class RefCountMessage {
 
    private volatile boolean released = false;
 
+   private volatile boolean errorCheck = true;
+
    /** has the refCount fired the action already? */
    public boolean isReleased() {
       return released;
@@ -203,7 +205,7 @@ public class RefCountMessage {
 
    protected void released() {
       released = true;
-      accountedFor();
+      disableErrorCheck();
    }
 
    void runOnLeak(Runnable run) {
@@ -212,7 +214,8 @@ public class RefCountMessage {
       }
    }
 
-   public void accountedFor() {
+   public void disableErrorCheck() {
+      this.errorCheck = false;
       if (debugStatus != null) {
          debugStatus.accountedFor();
       }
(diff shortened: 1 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
```diff
@@ -1130,14 +1130,16 @@ private void sendContinuations(final int packetSize,
...
                session.send(session.getCurrentTransaction(), EmbedMessageUtil.extractEmbedded((ICoreMessage) message.toMessage(), storageManager), false, producers.get(senderID), false);
-            } catch (Exception e) {
-               message.deleteFile();
-               throw e;
             } catch (Throwable e) {
-               logger.warn("********************************************************************************");
-               logger.warn("Throwable on currentLargeMessage {}", message.getMessageID(), e);
-               logger.warn("********************************************************************************");
-
+               logger.warn(e.getMessage(), e);
```
gemmellr, line 1134: Putting this here and removing the 'catch(Exception e)' means it will begin printing warnings with stacktraces for _every_ exception thrown in send now, vs none before, even though its still throwing the exception onward, to someplace that may also print it if not handling it specifically. Its not clear to me thats desirable now if it was specifically avoided before, I expect it to generate a bunch of stacktraces people will then report as 'new errors' as they arent used to it.

Previously it only logged warning for the 'unhandled' Error's it then swallowed. You've changed it to rethrow those too. Perhaps just make a debug log for the Exceptions if wanting to add potential visibility, and retain the warning level for the Throwables?

Using 't' is more common for throwables than 'e'.

gemmellr: The code this method is calling looks to throw exceptions in fairly 'typical' situations, e.g security checks, disk usage checks. I think starting to log a full stacktrace for every one of them, when it specifically didnt before (presumably as the code calling this method should either handle, or itself print, them), is likely asking for a lot of new questions. Along the lines of the more typical 'I suddenly got this new error \<provides non-error info log message and/or stacktrace\>'.

PR #4431, opened 2023-04-08
ARTEMIS-4239 use StandardCharsets.UTF_8

amarkevich:
avoid UnsupportedEncodingException handling


jbertram:
This needs a [Jira](https://issues.apache.org/jira/browse/ARTEMIS).

tabish121:
The change seems to be more expansive than the short title might imply perhaps either limit the changes or provide more context into what and why this is being done

amarkevich:
> limit the changes

done

PR #4432, opened 2023-04-08
ARTEMIS-4207 Improving redistribution fix over large messages

clebertsuconic:


clebertsuconic:
This is just avoiding creating a TransactionImpl object if no route is going to happen.

PR #4433, opened 2023-04-11
[ARTEMIS-4238]: transaction-timeout ActivationConfigProperty is no lo…

ehsavoie:
…nger working.

* Re-adding the missing code

Jira: https://issues.apache.org/jira/browse/ARTEMIS-4238

### DIFF
--- a/artemis-ra/src/main/java/org/apache/activemq/artemis/ra/inflow/ActiveMQMessageHandler.java
+++ b/artemis-ra/src/main/java/org/apache/activemq/artemis/ra/inflow/ActiveMQMessageHandler.java
@@ -55,6 +55,7 @@ import org.apache.activemq.artemis.utils.VersionLoader;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.lang.invoke.MethodHandles;
+import javax.transaction.TransactionManager;
 
 /**
  * The message handler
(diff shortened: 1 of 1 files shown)


artemis-ra/src/main/java/org/apache/activemq/artemis/ra/ActiveMQRAManagedConnection.java
```diff
@@ -336,7 +336,7 @@ public void associateConnection(final Object obj) throws ResourceException {
 
    public void checkTransactionActive() throws JMSException {
       // don't bother looking at the transaction if there's an active XID
-      if (!inManagedTx && tsr != null) {
+      if (!inManagedTx && tsr != null && tsr.getTransactionKey() != null) {
```
jbertram, line 339: This change doesn't appear related to the `transactionTimeout` change. Why is this necessary?

jbertram:
It would be ideal to have some tests for this to catch future regressions.

jbertram:
The commit message doesn't follow the [recommended format](https://github.com/apache/activemq-artemis/blob/main/docs/hacking-guide/en/code.md#commitMessageDetails). There's no need for square brackets and a colon around the Jira reference.

artemis-ra/src/main/java/org/apache/activemq/artemis/ra/ActiveMQRAManagedConnection.java
ehsavoie: a leftover

jbertram:
@ehsavoie, this looks good. I transitioned it from draft and will merge shortly.

PR #4435, opened 2023-04-13
ARTEMIS-4238 deprecate transactionTimeout activation config prop

jbertram:


artemis-ra/src/main/java/org/apache/activemq/artemis/ra/inflow/ActiveMQActivationSpec.java
```diff
@@ -120,6 +120,7 @@ public class ActiveMQActivationSpec extends ConnectionFactoryProperties implemen
    /**
     * Transaction timeout
     */
+   @Deprecated
```
gemmellr, line 123: If it is deprecated for future removal as indicated in the documentation, should the Deprecated tag indicate that with the forRemoval flag?

artemis-ra/src/main/java/org/apache/activemq/artemis/ra/inflow/ActiveMQActivationSpec.java
```diff
@@ -631,6 +632,7 @@ public Integer getTransactionTimeout() {
     *
     * @param value The value
     */
+   @Deprecated
```
gemmellr, line 635: Ditto

artemis-ra/src/main/java/org/apache/activemq/artemis/ra/inflow/ActiveMQActivationSpec.java
jbertram: Done.

artemis-ra/src/main/java/org/apache/activemq/artemis/ra/inflow/ActiveMQActivationSpec.java
jbertram: Ditto

PR #4440, opened 2023-04-19
ARTEMIS-4244 Set web config using system properties

brusdev:
It is possible to use system properties to add or update web configuration items.
If you define a system property starting with "webconfig." it will be parsed at the startup
to update the web configuration.

To enable the client authentication for an existing binding with the name `artemis`,
set the system property `webconfig.bindings.artemis.clientAuth` to `true`, i.e.

```
java -Dwebconfig.bindings.artemis.clientAuth=true
```

To add a new binding or app set the new binding or app attributes using their new names, i.e.

```
java -Dwebconfig.bindings.my-binding.uri=http://localhost:8162
java -Dwebconfig.bindings.my-binding.apps.my-app.uri=my-app
java -Dwebconfig.bindings.my-binding.apps.my-app.war=my-app.war
```

To update a binding without a name use its uri and to update an app without a name use its url , i.e.

```xml
<web path="web">
  <binding uri="http://localhost:8161">
    <app url="activemq-branding" war="activemq-branding.war"/>
...
```

```
java -Dwebconfig.bindings."http://localhost:8161".clientAuth=true
```

```
java -Dwebconfig.bindings."http://localhost:8161".apps."activemq-branding".war=my-branding.war
```

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Run.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Run.java
@@ -133,6 +133,10 @@ public class Run extends LockAbstract {
             broker.components.add(broker.web);
          }
 
+         String systemWebPropertyPrefix = ActiveMQDefaultConfiguration.getDefaultSystemWebPropertyPrefix();
+         server.getServer().getConfiguration().parsePrefixedProperties(broker.web,
+            "system-" + systemWebPropertyPrefix, System.getProperties(), systemWebPropertyPrefix);
+
          for (ComponentDTO componentDTO : broker.components) {
             Class clazz = this.getClass().getClassLoader().loadClass(componentDTO.componentClassName);
             ExternalComponent component = (ExternalComponent) clazz.getDeclaredConstructor(null).newInstance();
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/web/WebServerDTOConfigTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/web/WebServerDTOConfigTest.java
@@ -1,0 +1,183 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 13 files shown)


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Run.java
```diff
@@ -133,6 +133,8 @@ public void deActivate() {
             broker.components.add(broker.web);
          }
 
+         server.getServer().getConfiguration().parsePrefixedProperties(broker.web, "web", System.getProperties(), ActiveMQDefaultConfiguration.getDefaultWebPropertyPrefix());
```
gtully, line 136: think maybe system-web, or even "system-" + ActiveMQDefaultConfiguration.getDefaultWebPropertyPrefix()

this will only show up in the config.getStatus, to identify checksum and to capture any errors specific to that set of properties.

gtully: that would make it "system-webconfig." am wondering if the dot should be removed, but if the prefix is configured by a user, it is good that the actual configured value appears in the status. So i would keep the dot.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/web/WebServerDTOConfigTest.java
```diff
@@ -0,0 +1,93 @@
...
+      properties.put(ActiveMQDefaultConfiguration.getDefaultWebPropertyPrefix() + "bindings.test-binding.apps.test-app.war", "test-war");
+      Configuration configuration = new ConfigurationImpl();
+      configuration.parsePrefixedProperties(webServer, "web", properties, ActiveMQDefaultConfiguration.getDefaultWebPropertyPrefix());
+
+      BindingDTO testBinding = webServer.getAllBindings().stream().filter(binding -> "test-binding".equals(binding.getName())).findFirst().get();
+      AppDTO testApp = testBinding.getApps().stream().filter(app -> "test-app".equals(app.getName())).findFirst().get();
+      Assert.assertEquals("test-url", testApp.getUrl());
+      Assert.assertEquals("test-war", testApp.getWar());
+   }
+}
```
gtully, line 93: maybe one more test, that verifies feedback when an attribute is not found or misspelled. but otherwise this is great :-)

gtully (approved):

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Run.java
gtully: I was sort of combining them for the name that identifies where the properties come from. system-<prefix> may get us there. It only shows up in the status.

brusdev: @gtully thanks for your suggestion, I applied it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/web/WebServerDTOConfigTest.java
brusdev: I added other tests to improve the coverage.

brusdev:
I added the support for binding and apps without names and a section in the `Embedded Web Server` doc.

gtully (approved):

PR #4441, opened 2023-04-19
ARTEMIS-4245 Expose web SNI settings

brusdev:
This PR depends on #4440


artemis-web/src/test/java/org/apache/activemq/cli/test/WebServerComponentTest.java
```diff
@@ -215,11 +221,11 @@ public void testComponentStopStartBehavior() throws Exception {
       Assert.assertFalse(webServerComponent.isStarted());
    }
 
-   @Test
-   public void simpleSecureServer() throws Exception {
+   private WebServerComponent startSimpleSecureServer(String keyStorePath, String keyStorePassword, Boolean sniHostCheck) throws Exception {
       BindingDTO bindingDTO = new BindingDTO();
       bindingDTO.uri = "https://localhost:0";
       bindingDTO.keyStorePath = "./src/test/resources/server.keystore";
+      bindingDTO.setSniHostCheck(sniHostCheck);
```
gemmellr, line 228: might be nice to only set it if non-null?

artemis-web/src/test/java/org/apache/activemq/cli/test/WebServerComponentTest.java
```diff
@@ -283,6 +298,40 @@ public void simpleSecureServer() throws Exception {
...
+
+   @Test
+   public void testSimpleSecureServerWithSniHostCheckEnabled() throws Exception {
+      testSimpleSecureServerWithSniHostCheck(true);
+   }
+
+   @Test
+   public void testSimpleSecureServerWithSniHostCheckDisabled() throws Exception {
+      testSimpleSecureServerWithSniHostCheck(false);
+   }
```
gemmellr, line 310: Do we want to test the default behaviour is 'enabled' (i.e change the boolean to Boolean and pass null as well)?

Currently nothing looks to be checking the default, since these 2 tests only verify it succeeds/fails using the IP address when the setting was _explicitly_ disabled or enabled.

artemis-web/src/test/java/org/apache/activemq/cli/test/WebServerComponentTest.java
brusdev: If the test doesn't set bindingDTO.sniHostCheck the WebServerComponent behaviour depends on org.eclipse.jetty.server.SecureRequestCustomizer._sniHostCheck default value and it already changed from 9.x to 10.x

gemmellr: Sure, but the comment was just about 'getting the default behaviour' rather than explicitly setting the null (whichone testdoes), and instead relying on the bonafiide default behaviour so it could be verified to match what we expect.

artemis-web/src/test/java/org/apache/activemq/cli/test/WebServerComponentTest.java
gemmellr: We could do that instead yes.

artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
```diff
@@ -245,7 +245,10 @@ private ServerConnector createServerConnector(HttpConfiguration httpConfiguratio
 
          SslConnectionFactory sslConnectionFactory = new SslConnectionFactory(sslFactory, "HTTP/1.1");
 
-         httpConfiguration.addCustomizer(new SecureRequestCustomizer());
+         SecureRequestCustomizer secureRequestCustomizer = new SecureRequestCustomizer();
+         secureRequestCustomizer.setSniHostCheck(binding.getSniHostCheck() != null ? binding.getSniHostCheck() : true);
+         secureRequestCustomizer.setSniRequired(binding.getSniRequired() != null ? binding.getSniRequired() : false);
```
gemmellr, line 250: Perhaps constants for the defaults?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/web/WebServerDTOConfigTest.java
```diff
@@ -97,6 +97,8 @@ private void testSetWebBindingProperties(WebServerDTO webServer, String bindingN
       properties.put(ActiveMQDefaultConfiguration.getDefaultSystemWebPropertyPrefix() + "bindings." + bindingName + ".excludedCipherSuites", "test-excludedCipherSuites,3");
       properties.put(ActiveMQDefaultConfiguration.getDefaultSystemWebPropertyPrefix() + "bindings." + bindingName + ".keyStorePassword", "test-keyStorePassword");
       properties.put(ActiveMQDefaultConfiguration.getDefaultSystemWebPropertyPrefix() + "bindings." + bindingName + ".trustStorePassword", "test-trustStorePassword");
+      properties.put(ActiveMQDefaultConfiguration.getDefaultSystemWebPropertyPrefix() + "bindings." + bindingName + ".sniHostCheck", "true");
+      properties.put(ActiveMQDefaultConfiguration.getDefaultSystemWebPropertyPrefix() + "bindings." + bindingName + ".sniRequired", "true");
```
gemmellr, line 101: These are both being set true...meaning one is being set to its already-expected default and the other isnt. Perhaps have both set the non-default value?

gemmellr: oops, for some reason I thought it was asserting the final value on WebServerComponent, not the DTO.

PR #4447, opened 2023-04-24
ARTEMIS-4251 Support CORE client failover to other live servers

brusdev:
Improve the CORE client failover connecting to other live servers when all reconnect attempts fails, i.e. in a cluster composed of 2 live servers, when the server to which the CORE client is connected goes down the CORE client should reconnect its sessions to the other liver broker.


AntonRoskvist:
Really nice addition.

Haven't really looked into how it can be implemented, but there is a feature in "ActiveMQ Classic" that I think would be a great addition on top of this change. It's called "priorityBackup" ( https://activemq.apache.org/failover-transport-reference.html#priority-backup )

It has the client perform this type of failover to carry on its work, but then afterwards tries to continuously connect back to the original broker, releasing the new connection if/when it succeeds. This is good for handling client distribution across a broker cluster that sort of "persists" through maintenance, crashes, network outages etc on the broker side.

Either way, this is a really nice addition as is!

docs/user-manual/en/client-reconnection.md
```diff
@@ -89,9 +89,20 @@ Client reconnection is configured using the following parameters:
...
 
+- `ha`. This optional parameter determines weather the client will try to
+  reconnect to the backup node when the live node is not reachable.
+  The default value is `false`.
+  For more information on HA, please see [High Availability and Failover](ha.md).
+
 - `reconnectAttempts`. This optional parameter determines the total number of
-  reconnect attempts to make before giving up and shutting down. A value of
-  `-1` signifies an unlimited number of attempts. The default value is `0`.
+  reconnect attempts to make to the current live/backup pair before giving up.
```
gtully, line 98: I am confused :-)

I thought failover was between live and backup

reconnect should first try to failover, then fall back to any connection to eventually reconnect. these seems the opposite.

gemmellr: Yeah, could probably do with some docs to clarify what specifically is a 'reconnect attempt' vs a 'failover attempt' and thus help show when you would use one/other/both, for those that dont already know the previous/existing behaviour and thus also wont be able tell later which one of these is the new option/behaviour (/me raises hand).

(E.g, in other clients I know, there is only one notion of 'attempts', where I believe generally they retry the server originally connected to again and then try all the other servers currently known about until succeeding, then repeat [after delay] if still not connected...with that one-shot-at-each-server process collectively being '1 attempt')

brusdev: @gtully @gemmellr thanks for your feedback, I'll review the doc and I'll move the pause after trying all available connectors.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/client/ServerLocator.java
```diff
@@ -653,6 +653,21 @@ ClientSessionFactory createSessionFactory(TransportConfiguration transportConfig
     */
    int getInitialConnectAttempts();
 
+   /**
+    * Sets the maximum number of attempts to establish a connection after a failed reconnection.
```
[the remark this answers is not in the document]
brusdev: done, thanks for the suggestion

docs/user-manual/en/client-reconnection.md
brusdev: done

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
```diff
@@ -664,33 +669,94 @@ private void failoverOrReconnect(final Object connectionID,
...
+
+                        oldConnection = connection;
+                        connection = null;
+                     }
+                  }
+
+                  if (connectorsCount >= serverLocator.getConnectorsSize()) {
+                     connectorsCount = 0;
+                     failoverReties++;
+                     if (failoverRetryPredicate.test(false, failoverReties)) {
```
gemmellr, line 750: Since this overall 'failover connect block' doesnt use *getConnectionWithRetry* like the earlier 'reconnect connect block' does, that looks to mean it also doesnt use the retryInterval scaling/max within *getConnectionWithRetry*, instead only using the base retryInterval config value during its waits (on the line below this). I didnt see anything obvious skimming the doc...should probably either document it doesnt use the scaling/max, or make it do so.

brusdev: Use the exponential back-off algorithm also for `failover connect block` sounds good to me

gemmellr:
Seems good to me, would be good to get a review from folks more familiar with the core client.

brusdev:
@gemmellr thanks for your review and suggestion

docs/user-manual/en/client-reconnection-failover.md
```diff
@@ -1,10 +1,32 @@
-# Client Reconnection and Session Reattachment
+# Client Reconnection and Failover
```
clebertsuconic, line 1: Can we simplify the doc here?


I still haven't figured out how you would configure the client to failover to another server... 

I thought about just having two topics would make it simpler.




- Failover to a backup Server


(don't worry too much about mentioning Apache ActiveMQ bblablablabla all the time)

just say, if you want to reconnect to a backup server, configure this, and that.





- Failover to non backup servers

Mention what you have to do if you want to failover to non backup servers...





There are so many options here that's getting confusing to understand.


that's my only concern here.

brusdev: @clebertsuconic thanks for your feedback, I'll rewrite the doc

brusdev: @clebertsuconic I partially rewrote the doc following your suggestions, could you take a look?

clebertsuconic:
@brusdev can you squash / rebase it? (there are a few small conflicts)

PR #4448, opened 2023-04-24
ARTEMIS-4252 Fix flaky QuorumFailOverTest tests

brusdev:


artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -1328,6 +1330,17 @@ protected static final void waitForComponent(final ActiveMQComponent component,
       }
    }
 
+   protected static final void waitForBridges(final ActiveMQServer server, int connectedBridges) throws Exception {
+      waitForBridges(server.getClusterManager().getDefaultConnection(null), connectedBridges);
+   }
+
+   protected static final void waitForBridges(ClusterConnection clusterConnection, int connectedBridges) throws Exception {
+      Bridge[] bridges = clusterConnection.getBridges();
```
gemmellr, line 1338: This doesnt look used. Is it needed? Was a quick-exit check intended but missed?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionImpl.java
```diff
@@ -630,6 +630,11 @@ public long getCallTimeout() {
       return callTimeout;
    }
 
+   @Override
+   public Bridge[] getBridges() {
+      return records.values().stream().map(MessageFlowRecord::getBridge).toArray(Bridge[]::new);
```
gemmellr, line 635: The 'records' field has 'recordsGuard' to protect its use in other cases. Is that needed here?

brusdev: It makes sense, I'll fix it. I'll create another PR to fix it.

artemis-server/src/test/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
brusdev: You are right I missed to remove that line, I'll create another PR to fix it.

PR #4453, opened 2023-04-26
NO-JIRA Remove dependency on tree utility

ViliusS:
Tree utility is not present by default in a lot of smaller Linux installations, like for example CentOS or Rocky Linux minimal image. As it serves almost no purpose other than visual, I suggest removing this dependency from Docker build requirements.

Alternatively, I can prepare a PR with `ls -la` alternative if you think it is still needed to list those files.


jbertram:
I think this is going to be moot once #4307 is merged.

ViliusS:
Since we have have official Artemis Docker images now and the state of https://github.com/apache/activemq-artemis/pull/4307 hangs in a balance, can we have this PR merged?

PR #4460, opened 2023-04-28
ARTEMIS-4265 make more web console tabs conditional on permission

jbertram:
Many of the tabs on the web console show up even though the user doesn't have permission to execute the command corresponding to the tab. For example the "Connections" tab shows up even though the user can't execute the `listConnections` management operation.


brusdev:
The change LGTM, are you planning to add any console smoke test, i.e. [QueuesTest](https://github.com/apache/activemq-artemis/blob/main/tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/console/QueuesTest.java)?

jbertram:
@brusdev, I wasn't aware of that class. Based on your comment I took a look, but I don't see how to customize the following files used by the broker:
 - `management.xml`
 - `artemis.profile`
 - `artemis-users.properties`
 - `artemis-roles.properties`

I need to set up a handful of different users in unique roles with access to the console. Is this possible?

brusdev:
@jbertram you could create the `console` folder with those files in the [servers](https://github.com/apache/activemq-artemis/tree/main/tests/smoke-tests/src/main/resources/servers) folder and add the `configuration` element to the `create-create-console` execution, i.e.
https://github.com/apache/activemq-artemis/blob/main/tests/smoke-tests/pom.xml#L258
```
 <execution>
    <phase>test-compile</phase>
    <id>create-create-console</id>
    <goals>
       <goal>create</goal>
    </goals>
    <configuration>
       <configuration>${basedir}/target/classes/servers/console</configuration>
       ...
```

jbertram:
@brusdev, thanks for the tip! Tests added.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/console/QueuesTest.java
```diff
@@ -60,6 +62,89 @@ public void testDefaultQueues() throws Exception {
       assertEquals(0, queuesPage.getMessagesCount("ExpiryQueue"));
    }
 
+   @Test
+   public void testConnectionsTab() {
```
brusdev, line 66: Nice tests but they don't seem related to queues, could you move them to a new test class, i.e. `TabsTest` or any other names?

brusdev, line 66: Nice tests but they don't seem related to queues, could you move them to a new test class, i.e. `TabsTest` or any other names?

jbertram: Done.

brusdev (approved):

PR #4462, opened 2023-05-01
ARTEMIS-4268 AMQPMessage copy constructor shouldn't copy all message annotations

clebertsuconic:
During redistribution, we should not copy all message annotations.

In particular we should not copy any of the x-opt-ORIG annotations used on DLQ and other copies.

this was broken after f632e8104bbdae1fbf3658fec47e180784e957da (ARTEMIS-3833 Preserve JMSCorrelationID of distributed AMQP large messages)

The change preserved too much, and as a result of that AmqpLargeMessageRedistributionTest::testSendMessageToBroker0GetFromBroker2 is intermittently failing.

There is no test in this commit as this is fixing AmqpLargeMessageRedistributionTest

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
@@ -252,13 +252,27 @@ public abstract class AMQPMessage extends RefCountMessage implements org.apache.
       this.encodedDeliveryAnnotationsSize = copy.encodedDeliveryAnnotationsSize;
       this.deliveryAnnotations = copy.deliveryAnnotations == null ? null : new DeliveryAnnotations(copy.deliveryAnnotations.getValue());
       this.messageAnnotationsPosition = copy.messageAnnotationsPosition;
-      this.messageAnnotations = copy.messageAnnotations == null ? null : new MessageAnnotations(copy.messageAnnotations.getValue());
+      this.messageAnnotations = copyAnnotations(copy.messageAnnotations);
       this.propertiesPosition = copy.propertiesPosition;
       this.properties = copy.properties == null ? null : new Properties(copy.properties);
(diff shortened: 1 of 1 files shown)


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
```diff
@@ -130,6 +130,10 @@ private AMQPLargeMessage(final AMQPLargeMessage copy,
                                   final SequentialFile fileCopy,
                                   final long newID) {
       super(copy);
+      if (this.messageAnnotations != null) {
+         this.messageAnnotations.getValue().forEach((a, b) -> logger.info("{}={}", a, b));
```
tabish121, line 134: Logging every annotation on every copy seems excessive especially at INFO, should this be lowered to TRACE to avoid spamming the logs ?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
```diff
@@ -130,6 +130,10 @@ private AMQPLargeMessage(final AMQPLargeMessage copy,
...
+      }
+      this.messageAnnotations = null;
```
tabish121, line 136: I'm confused....why is this nulling the messageAnnotations that the super constructor just copied using the newly added filtering bit?

clebertsuconic: I'm an idiot and I forgot to remove it.
although the copy is not fixing it... :/

clebertsuconic: fixed it... thanks a lot @tabish121

PR #4465, opened 2023-05-03
ARTEMIS-4244 Fix testSetWebBindingProperties

brusdev:


gemmellr:
Fixup for #4441 where I missed that during the updates, oopsie :)

PR #4466, opened 2023-05-04
ARTEMIS-4256 - support removal of configuration via properties

gtully:


artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
```diff
@@ -569,8 +569,12 @@ public static String getDefaultHapolicyBackupStrategy() {
 
    public static final String BROKER_PROPERTIES_KEY_SURROUND = "\"";
 
+   public static final String BROKER_PROPERTIES_REMOVE_VALUE = "-";
```
brusdev, line 572: Other markup launguages as YAML use `~` for the null value, https://yaml.org/type/null.html

brusdev (commented):
This improvement will bring the feature of broker properties to a new level, well done!!!

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
gtully: interesting, thanks for the link. This is exactly the sort of thing we should be aware of and follow, convention over configuration etc.
However, in this case, I think the semantic of that is a little different, it is a null value, where the key still exists. x=null
x=- means remove x. 
I am not sure if there is any case in artemis config where we want to have a collection with a null entry.

brusdev: I see, good point!

brusdev (approved):

rvais (commented):
Tests look fine. Thumbs up!

PR #4473, opened 2023-05-12
ARTEMIS-4279 Shutdown critical analyzer thread if error on broker ini…

sjhiggs:
…tialization

The broker process fails to exit if an error is encountered starting the NodeManager.  The issue is resolved by converting the critical analyzer thread to a daemon thread.  As added protection, the thread is manually stopped when this error is encountered.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/critical/CriticalAnalyzerImpl.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/critical/CriticalAnalyzerImpl.java
@@ -16,12 +16,15 @@
  */
 package org.apache.activemq.artemis.utils.critical;
 
+import java.security.AccessController;
+import java.security.PrivilegedAction;
 import java.util.ConcurrentModificationException;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.apache.activemq.artemis.core.server.ActiveMQScheduledComponent;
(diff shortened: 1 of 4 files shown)


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/critical/CriticalAnalyzerImpl.java
```diff
@@ -48,11 +51,28 @@ public CriticalAnalyzerImpl() {
...
+
          @Override
          public void run() {
             logger.trace("Checking critical analyzer");
             check();
          }
+
+         @Override
+         protected ActiveMQThreadFactory getThreadFactory() {
+            return new ActiveMQThreadFactory("CriticalAnalyzer", "Critical-Analyzer-", true, getThisClassLoader());
```
jbertram, line 63: How come the thread needs to be a daemon thread now?

sjhiggs: The thought was that the critical analyzer should never prevent shutdown.  Catching the exception fixes this one case; a daemon thread is additional protection.

clebertsuconic: @jbertram I was actually talking to @sjhiggs .. I think these thread should always been daemon. It was a mistake not to.

jbertram: I was mixed up on how daemon threads related to the JVM exiting. Sorry for the confusion. This looks good.

PR #4479, opened 2023-05-16
ARTEMIS-4191 Adding debug statements around reclaiming and compacting

clebertsuconic:


artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -3565,6 +3580,8 @@ protected void moveNextFile(final boolean scheduleReclaim, boolean blockOnClose)
 
       if (scheduleReclaim) {
          scheduleReclaim();
+      } else {
+         logger.trace("JournalImpl::moveNextFile scheduleReclaim is false ( {} ), not calling scheduleReclaim", scheduleReclaim);
```
gemmellr, line 3584: Seems odd to have a text statement it is false, and then also a placeholder printing the same value as well (also the arg for that will be autoboxing the boolean...not necessarily an issue if it isnt especially hot or typically does scheduleReclaim)

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -2625,12 +2631,13 @@ private void scheduleCompact() {
          return;
       }
 
+      logger.debug("JournalImpll::scheduleCompact");
```
gemmellr, line 2634: Just checking its intended...this one is debug when all the others added are trace?

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
clebertsuconic: When I wrote this, I was concerned about the value changing between the check and the if value. so I was just reprinting just in case.. but I guess it can go away.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
clebertsuconic: I will keep it as debug, and moved others to debug

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
gemmellr: The value looks to be coming from a final method parameter and so cant change.

artemis-journal/src/main/java/org/apache/activemq/artemis/core/journal/impl/JournalImpl.java
```diff
@@ -3267,20 +3279,30 @@ void scheduleReclaim() {
...
                } catch (Exception e) {
                   ActiveMQJournalLogger.LOGGER.errorSchedulingCompacting(e);
+               } finally {
+                  logger.debug("JournalImpl::scheduleReclaim finished");
                }
             }
          });
+      } else {
+         if (logger.isTraceEnabled()) {
+            logger.trace("Ignoring compactorExecutor call because of autoReclaim={} and compactorRunning={}", isAutoReclaim(), compactorRunning.get());
```
gemmellr, line 3304: These values on the other hand, seem like they could actually change between first inspection in the if and later being logged, when the values are retrieved again.

If getting accurate info on these is important, would need to assigning the values so the logging reflects the values used for the decision.

If not, would maybe be worth tweaking the log message (e.g dropping 'because of', add some brackets, say 'current') to make it clearer the values logged are current and actually separate from the earlier decision.

clebertsuconic: I think it is ok

PR #4483, opened 2023-05-18
ARTEMIS-4284 - sync operwire remove consumer with the operation conte…

gtully:
…xt to ensure prefetched messages are available to the next consumer in order


gtully:
@clebertsuconic @jbertram I need some more eyes on this to validate that I am using the most effective way to ensure that the storage manager has completed. Is it ever possible that the storage manager does not run?

the crux of the issue is that the addSorted on the queue is called from the operation context and if we don't wait on that, we can accept a new consumer that will miss these prefetched messages, and get them out of order. That whacks the redelivery accounting and messages go to the DLQ!

PR #4485, opened 2023-05-22
ARTEMIS-4286 sometimes federated consumer won't stop

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
```diff
@@ -144,6 +152,8 @@ private void connect() throws Exception {
 
    @Override
    public void close() {
+      started.set(false);
```
brusdev, line 155: The connect could be already scheduled and not yet executed while a disconnect could be scheduled with zero delay. In this case, the disconnect would be executed before a connect.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/federation/FederatedQueueConsumerTest.java
```diff
@@ -0,0 +1,43 @@
...
+   @Test
+   public void testClose() throws Exception {
+      ActiveMQServer server = createServer(false, createDefaultInVMConfig());
+      server.start();
+      Federation federation = new Federation(server, new FederationConfiguration().setName(RandomUtil.randomString()));
+      federation.start();
+      FederatedQueueConsumerImpl consumer = new FederatedQueueConsumerImpl(federation, server, null, null, null, null);
+      consumer.start();
+      Wait.waitFor(() -> consumer.getConnectionAttemptTimestamp() > 0);
+      consumer.close();
```
brusdev, line 39: Is this test checking a close while the consumer is trying to connect?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
```diff
@@ -129,21 +134,23 @@ private void connect() throws Exception {
...
+            try {
+               if (clientSessionFactory != null) {
+                  clientSessionFactory.cleanup();
+               }
+               disconnect();
+            } catch (ActiveMQException ignored) {
             }
-            disconnect();
-         } catch (ActiveMQException ignored) {
+            throw e;
```
gemmellr, line 145: Not necessarily the spot to do it (perhaps in the task where it retries), but the easiest to comment in this PR. It seems like it might be good to have at least a trace log to indicate it has failed and is going to try connecting again.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
```diff
@@ -109,41 +113,46 @@ private void scheduleConnect(int delay) {
...
          }
-         throw e;
       }
    }
 
    @Override
-   public void close() {
+   public synchronized void close() {
+      started = false;
```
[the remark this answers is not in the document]
jbertram: I added checks to both `start` & `close`.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/federation/FederatedQueueConsumerTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+   @Test
+   public void testClose() throws Exception {
+      ActiveMQServer server = createServer(false, createDefaultInVMConfig());
+      server.start();
+      Federation federation = new Federation(server, new FederationConfiguration().setName(RandomUtil.randomString()));
+      federation.start();
+      FederatedQueueConsumerImpl consumer = new FederatedQueueConsumerImpl(federation, server, null, null, null, null);
+      consumer.start();
+      Wait.waitFor(() -> consumer.getConnectionAttemptTimestamp() > 0);
```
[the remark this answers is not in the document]
jbertram: Fixed.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/federation/FederatedQueueConsumerTest.java
```diff
@@ -0,0 +1,43 @@
...
+      consumer.close();
+      long closed = System.currentTimeMillis();
+      assertFalse(Wait.waitFor(() -> consumer.getConnectionAttemptTimestamp() > closed, 5000, 100));
```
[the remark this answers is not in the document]
jbertram: I lowered the timeout and the interval. I couldn't think of a better way to make the check faster or more reliable. If you have any ideas I'm open.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
jbertram: Logging added in `scheduleConnect`.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/federation/FederatedQueueConsumerTest.java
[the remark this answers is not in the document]
gemmellr: My basic suggestion yesterday was just inc/dec a counter, e.g instead of setting/resetting a timestamp as you did, keep a simple count of outstanding connect tasks. If there are none outstanding there will be no connect without calling start. Since the initial scheduling is all synchronized now, and the connect task reschedules itself on failure, its either going to be 0 (nothing scheduled, no connect will happen), or 1 (connect scheduled, or task still actually in progress), or 2 (a task in progress and failing, plus 1 retry it has just scheduled). You could then just wait for it to be 0 again.

A better alternative would probably be keeping the currently-thrown-away ScheduledFuture for the latest scheduled connect task. You could then tell whether there is another attempt scheduled. You could also use it in close() to cancel any future connect attempt that hasnt been started yet, rather than just relying on it to no-op potentially much later. This could also make the test even faster than the ~1sec typical of the first retry period, since then it wouldnt need to wait for the retry to happen, it could then tell it wont happen.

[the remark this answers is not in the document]
jbertram: I just did a `push -f`. Let me know what you think.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
```diff
@@ -49,6 +55,8 @@ public class FederatedQueueConsumerImpl implements FederatedQueueConsumer, Sessi
    private final int intialConnectDelayMultiplier = 2;
    private final int intialConnectDelayMax = 30;
    private final ClientSessionCallback clientSessionCallback;
+   private boolean started = false;
+   private ScheduledFuture currentConnectTask;
```
gemmellr, line 59: its set/used from different threads so should probably be volatile

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/federation/FederatedQueueConsumerTest.java
```diff
@@ -0,0 +1,42 @@
...
+   public void testClose() throws Exception {
+      ActiveMQServer server = createServer(false, createDefaultInVMConfig());
+      server.start();
+      Federation federation = new Federation(server, new FederationConfiguration().setName(RandomUtil.randomString()));
+      federation.start();
+      FederatedQueueConsumerImpl consumer = new FederatedQueueConsumerImpl(federation, server, null, null, null, null);
+      consumer.start();
+      Wait.assertTrue(() -> consumer.getCurrentConnectTask() != null, 2000, 20);
+      consumer.close();
+      Wait.assertTrue(() -> consumer.getCurrentConnectTask().isDone() || consumer.getCurrentConnectTask().isCancelled(), 2000, 50);
```
gemmellr, line 40: To cover the potential re-scheduling race, may be worth adding a check that calling consumer.getCurrentConnectTask() again returns the same future. So, its done and you get the same one, or its cancelled and you get the same one.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
```diff
@@ -95,24 +103,31 @@ public int decrementCount() {
...
    private void scheduleConnect(int delay) {
-      scheduledExecutorService.schedule(() -> {
+      currentConnectTask = scheduledExecutorService.schedule(() -> {
          try {
             connect();
          } catch (Exception e) {
-            scheduleConnect(FederatedQueueConsumer.getNextDelay(delay, intialConnectDelayMultiplier, intialConnectDelayMax));
+            int nextDelay = FederatedQueueConsumer.getNextDelay(delay, intialConnectDelayMultiplier, intialConnectDelayMax);
+            if (logger.isTraceEnabled()) {
+               logger.trace(this + " failed to connect. Scheduling reconnect in " + nextDelay + " seconds.", e);
```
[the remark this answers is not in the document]
jbertram: The problem with using place-holders is that there's no method which takes place-holders which also takes the cause and will log the whole stack-trace.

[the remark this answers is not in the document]
gemmellr: All the methods take the last vararg to be an exception if it is one.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
```diff
@@ -129,22 +144,26 @@ private void connect() throws Exception {
...
-      } catch (Exception e) {
-         try {
-            if (clientSessionFactory != null) {
-               clientSessionFactory.cleanup();
+         } catch (Exception e) {
+            try {
+               if (clientSessionFactory != null) {
+                  clientSessionFactory.cleanup();
+               }
+               disconnect();
```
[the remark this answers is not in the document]
jbertram: There's certainly no _checked_ exceptions. I don't see anything that would throw an unchecked exception either.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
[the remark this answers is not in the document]
jbertram: That's very cool, if true. Out of curiosity, how do you know that? I looked in the JavaDoc but I didn't see any mention of that semantic.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
```diff
@@ -129,22 +144,26 @@ private void connect() throws Exception {
...
       }
    }
 
    @Override
-   public void close() {
-      scheduleDisconnect(0);
+   public synchronized void close() {
+      if (started) {
+         started = false;
+         currentConnectTask.cancel(true);
```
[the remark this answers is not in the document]
jbertram: Fair enough.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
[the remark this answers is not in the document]
gemmellr: Its worked that way for as long as I've used SLF4J really. https://www.slf4j.org/faq.html#paramException

I'd agree the javadoc isnt clear on this, which I'd never noticed as...I knew it worked that way and so have never really needed to look :)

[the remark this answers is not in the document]
gemmellr: That is linked from the main Logger doc also: https://www.slf4j.org/apidocs/org/slf4j/Logger.html

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/federation/FederatedQueueConsumerTest.java
jbertram: Done.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
jbertram: Done.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
```diff
@@ -49,6 +55,8 @@ public class FederatedQueueConsumerImpl implements FederatedQueueConsumer, Sessi
    private final int intialConnectDelayMultiplier = 2;
    private final int intialConnectDelayMax = 30;
    private final ClientSessionCallback clientSessionCallback;
+   private volatile boolean started = false;
```
gemmellr, line 58: The use of _started_ all looks synchronized, if so it doesnt need to be volatile (the earlier comment was only about _currentConnectTask_ which isnt all synced)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/federation/FederatedQueueConsumerTest.java
```diff
@@ -0,0 +1,51 @@
...
+      assertNull(consumer.getCurrentConnectTask());
+      consumer.start();
+      assertNotNull(consumer.getCurrentConnectTask());
+      consumer.close();
+      AtomicReference<ScheduledFuture> task = new AtomicReference<>();
+      Wait.assertTrue(() -> {
+         task.set(consumer.getCurrentConnectTask());
+         return task.get().isDone() || task.get().isCancelled();
+      }, 2000, 50);
+      assertTrue(task.get() == consumer.getCurrentConnectTask());
```
gemmellr, line 49: The check needs to be inside the Wait condition, to check the finished task is still the current one, i.e there will be no more rather than that complete task having scheduled another after the future was retrieved. (It could still assert on it after the wait too though).

jbertram: I see what you mean.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/federation/FederatedQueueConsumerTest.java
```diff
@@ -0,0 +1,48 @@
...
+      Federation federation = new Federation(server, new FederationConfiguration().setName(RandomUtil.randomString()));
+      federation.start();
+      FederatedQueueConsumerImpl consumer = new FederatedQueueConsumerImpl(federation, server, null, null, null, null);
+      assertNull(consumer.getCurrentConnectTask());
+      consumer.start();
+      assertNotNull(consumer.getCurrentConnectTask());
+      consumer.close();
+      Wait.assertTrue(() -> {
+         ScheduledFuture task = consumer.getCurrentConnectTask();
+         return task.isDone() || task.isCancelled() && task == consumer.getCurrentConnectTask();
```
gemmellr, line 45: This only checks the future when when it is cancelled, but needs to check it is the same when it is 'done', since it may reschedule while running, after we got it initially.

jbertram: Ah yes. I forgot my parentheses.

PR #4486, opened 2023-05-23
ARTEMIS-4290 Separating some integration tests into legacy-integration-tests

clebertsuconic:
legacy-integration-tests is being created to hold LDAP Tests (or any other tests that won't play well with keeping threads clean) it will have fork-mode=always on the maven-surefire-plugin


clebertsuconic:
the test suite behaved incredibly well with this change.


If anyone does not like any dependencies or anything we can send more commits later one... 

I'm very happy with the result of this split, so I want to start using it

PR #4487, opened 2023-05-25
ARTEMIS-4291 Add the "broker" tag in a consistent manner

naude-r:
ARTEMIS-4291 Add the "broker" tag in a consistent manner to all exported metrics.
    
Some scrapers, e.g. prometheus, add an "instance" tag. This value may not be the same as
the broker name, which results in these metrics becoming more difficult to match up with
the corresponding broker.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/MetricsManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/MetricsManager.java
@@ -57,20 +57,23 @@ public class MetricsManager {
                          MetricsConfiguration metricsConfiguration,
                          HierarchicalRepository<AddressSettings> addressSettingsRepository) {
       this.brokerName = brokerName;
-      meterRegistry = metricsConfiguration.getPlugin().getRegistry();
-      Metrics.globalRegistry.add(meterRegistry);
+      this.meterRegistry = metricsConfiguration.getPlugin().getRegistry();
       this.addressSettingsRepository = addressSettingsRepository;
-      if (metricsConfiguration.isJvmMemory()) {
-         new JvmMemoryMetrics().bindTo(meterRegistry);
-      }
-      if (metricsConfiguration.isJvmGc()) {
-         new JvmGcMetrics().bindTo(meterRegistry);
-      }
-      if (metricsConfiguration.isJvmThread()) {
-         new JvmThreadMetrics().bindTo(meterRegistry);
-      }
-      if (metricsConfiguration.isNettyPool()) {
-         new NettyPooledAllocatorMetrics(PooledByteBufAllocator.DEFAULT.metric()).bindTo(meterRegistry);
+      if (meterRegistry != null) {
+         Metrics.globalRegistry.add(meterRegistry);
+         meterRegistry.config().commonTags("broker", brokerName);
+         if (metricsConfiguration.isJvmMemory()) {
+            new JvmMemoryMetrics().bindTo(meterRegistry);
+         }
+         if (metricsConfiguration.isJvmGc()) {
+            new JvmGcMetrics().bindTo(meterRegistry);
+         }
+         if (metricsConfiguration.isJvmThread()) {
+            new JvmThreadMetrics().bindTo(meterRegistry);
+         }
+         if (metricsConfiguration.isNettyPool()) {
+            new NettyPooledAllocatorMetrics(PooledByteBufAllocator.DEFAULT.metric()).bindTo(meterRegistry);
+         }
       }
    }
 
@@ -92,7 +95,6 @@ public class MetricsManager {
       builder.accept((metricName, state, f, description) -> {
          Gauge.Builder meter = Gauge
             .builder("artemis." + metricName, state, f)
-            .tag("broker", brokerName)
             .tag("address", address)
             .tag("queue", queue)
             .description(description);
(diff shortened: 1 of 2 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/ActiveMQMetricsPlugin.java
```diff
@@ -26,4 +26,10 @@ public interface ActiveMQMetricsPlugin extends Serializable {
    ActiveMQMetricsPlugin init(Map<String, String> options);
 
    MeterRegistry getRegistry();
+
+   /**
+    * Allow additional binders to be bound at the right time
+    */
+   default void bind() {
```
jbertram, line 33: It's not clear why this new method is needed. Can you elaborate?

naude-r: sure. we have a timing issue in that ActiveMQMetricsPlugin::init is called before MetricsManager have had the chance to add the common tags.

any metrics created / registered before then will not have the "broker" tag. 

not sure there is an easy way to do that without breaking existing plugins?

jbertram: Are you adding your _own_ metrics in your plugin implementation?

jbertram: In my opinion the broker should simply support the processor and uptime metrics just like it does the JVM and Netty metrics. That way _every_ user could benefit from them and the plugin interface wouldn't need to change. I created https://issues.apache.org/jira/browse/ARTEMIS-4292 for this work.

jbertram: Back this change out and I'll merge it (assuming it passes the PR tests). Then I'll send a PR for ARTEMIS-4292.

jbertram:
Please reformat your commit message to match [our standards](https://github.com/apache/activemq-artemis/blob/main/docs/hacking-guide/en/code.md#commitMessageDetails). Thanks!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/ActiveMQMetricsPlugin.java
naude-r: thanks. done.

jbertram:
I'm concerned this may break existing users since the name of the JVM and Netty metrics are doing to change now. Can you speak to this?

naude-r:
consider the following two metrics as scraped by prometheus:
```
artemis_address_size{address="ExpiryQueue",broker="$FQDN",endpoint="metrics-port",instance="$IP:8161",job="artemis",namespace="infra",service="artemis-exporters"}
jvm_info{endpoint="metrics-port",instance="$IP:8161",job="artemis",namespace="infra",runtime="OpenJDK Runtime Environment",service="artemis-exporters",vendor="Red Hat, Inc.",version="17.0.5+8-LTS"}
```

the only way to link the two metrics are the "instance" tag, which in our case is a raw ip (broker running external to k8s and k8s endpoints only supports ip addresses). the "broker" contains a user friendly name which make the dashboard easier to use.

in my opinion there is no way that adding a "broker" tag can break any existing dashboards and/or expectation. the cardinality will also not increase since the same broker tag/value pair is already available from the address metrics.

hope that make sense?

jbertram:
Yes, that makes sense. Before I merge it can you add a test to ensure that the {{broker}} tag is set on the JVM & Netty metrics as expected? You could modify {{org.apache.activemq.artemis.tests.integration.plugin.JvmMetricsTest}} for the JVM metrics. However, we don't have an existing test for the Netty metrics so you could copy the aforementioned {{JvmMetricsTest}} and modify it for the Netty use-case.

naude-r:
> Yes, that makes sense. Before I merge it can you add a test to ensure that the {{broker}} tag is set on the JVM & Netty metrics as expected? You could modify {{org.apache.activemq.artemis.tests.integration.plugin.JvmMetricsTest}} for the JVM metrics. However, we don't have an existing test for the Netty metrics so you could copy the aforementioned {{JvmMetricsTest}} and modify it for the Netty use-case.
> 
> To be clear, just about every commit needs a test to validate the change is working as expected and to mitigate against future regressions.

certainly. will get to it either today or tomorrow.

naude-r:
> > Yes, that makes sense. Before I merge it can you add a test to ensure that the {{broker}} tag is set on the JVM & Netty metrics as expected? You could modify {{org.apache.activemq.artemis.tests.integration.plugin.JvmMetricsTest}} for the JVM metrics. However, we don't have an existing test for the Netty metrics so you could copy the aforementioned {{JvmMetricsTest}} and modify it for the Netty use-case.
> > To be clear, just about every commit needs a test to validate the change is working as expected and to mitigate against future regressions.
> 
> certainly. will get to it either today or tomorrow.

test cases added as requested.

jbertram:
I just sent #4497 to deal with the metrics you were adding manually before. Once merged & released you can use `broker.xml` to configure those.

jbertram:
BTW, thanks for the contribution!

naude-r:
Thank you for the additional metrics!

fogninid:
@naude-r @jbertram I am having troubles with this change adding the `broker` tag/label to all metrics of my application.

Can you have a look at the draft PR I linked to this? Or suggest an alternative approach how to avoid artemis leaking these upwards to all meters of my MeterRegistry?

PR #4488, opened 2023-05-25
ARTEMIS-4293 add mngmnt ops to clear authn/z caches

jbertram:


brusdev (approved):

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -257,6 +257,32 @@ public void testSecurityCacheSizes() throws Exception {
...
+      m.putStringProperty("hello", "world");
+      producer.send(m);
+
+      Assert.assertTrue(serverControl.getAuthenticationCacheSize() > 0);
+      Wait.assertTrue(() -> serverControl.getAuthorizationCacheSize() > 0);
+
+      serverControl.clearAuthenticationCache();
+      serverControl.clearAuthorizationCache();
+
+      Assert.assertEquals(usingCore() ? 1 : 0, serverControl.getAuthenticationCacheSize());
```
rvais, line 282: I am unsure what this assertion actually does...?!  Test doesn't seem to be parametrized and doesn't seem to contain anything that would influence the return value of "usingCore" method. If the return value doesn't change, why the ternal operator is used in the first place? The same applies for the condition below. What these actually assert?

rvais (commented):
@jbertram I am unsure about assertions in the added test. See the comment in the code.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
jbertram: The `usingCore()` method is implemented by:

 - `org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlTest` (returns `false`)
 - `org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlUsingCoreTest` (returns `true`)

This method is meant to indicate when core _messages_ are being used for management rather than JMX. Using core requires an extra connection so the management messages can be sent. This extra connection (and everything that goes with it like authentications and authorizations) need to be ignored so that the same tests work with or without core.

PR #4490, opened 2023-05-29
ARTEMIS-4295 artemis-docker: replace 'tree' command with 'ls -l'

amarkevich:
'tree' command is not available in base image, so cloud build failed:

`./prepare-docker.sh: 188: tree: not found`


jbertram:
A PR with this change has already been submitted. See #4453.

PR #4492, opened 2023-05-30
Add test to verify no MQTT messages are lost (when using response topics)

inad9300:
In the following test, a broker client sends 1000 messages which are expected to be received by a second client, which in turn replies to every message with new ones (thus being labeled "server"). _It fails every time_ – the broker just seems to get stuck and stop sending messages at a given moment.

In case it is useful, when lowering the number of messages sent to 100, you should see that the test sometimes passes.

I have been running the tests with:

```sh
$ cd tests/integration-tests
$ mvn verify -DskipIntegrationTests=false -Dtest="MQTT5Test#testRequestResponseMessages*"
```

I created this test to reproduce a slightly more complicated case I am dealing with at the moment where, in a production setting, messages are getting lost left and right and I haven't got a clue why.


jbertram:
I added a comment to https://issues.apache.org/jira/browse/ARTEMIS-4296. Let's continue the discussion there.

PR #4493, opened 2023-05-31
ARTEMIS-4294 support text-encoded WebSocket frames

jbertram:


jbertram:
Something weird happened on `WebSocketFrameEncoder.java` with the EOLs so the whole file got changed. I think it originally used CRLF instead of just LF.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/protocol/websocket/WebSocketFrameEncoderType.java
```diff
@@ -0,0 +1,33 @@
...
+   BINARY, TEXT;
+
+   public static WebSocketFrameEncoderType valueOfType(String type) {
+      switch (type) {
+         case "binary":
+            return BINARY;
+         case "text":
+            return TEXT;
+         default:
+            return null;
```
tabish121, line 30: Instead of returning null here why not just throw an exception that indicates that this is an invalid argument here to avoid any chance of NPE later

artemis-server/src/test/java/org/apache/activemq/artemis/core/server/protocol/websocket/WebSocketFrameEncoderTest.java
```diff
@@ -60,11 +61,22 @@ public class WebSocketFrameEncoderTest {
...
+      textSpy = spy(new WebSocketFrameEncoder(maxFramePayloadLength, WebSocketFrameEncoderType.TEXT));
    }
 
    @Test
-   public void testWriteNonByteBuf() throws Exception {
+   public void testWriteNonByteBufBinary() throws Exception {
+      testWriteNonByteBuf(binarySpy);
+   }
+
+
```
tabish121, line 73: Extra newline

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/ProtocolHandler.java
```diff
@@ -158,7 +160,12 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
                }
                stompMaxFramePayloadLength = stompMaxFramePayloadLength != -1 ? stompMaxFramePayloadLength : TransportConstants.DEFAULT_WEB_SOCKET_MAX_FRAME_PAYLOAD_LENGTH;
                int webSocketMaxFramePayloadLength = ConfigurationHelper.getIntProperty(TransportConstants.WEB_SOCKET_MAX_FRAME_PAYLOAD_LENGTH, -1, nettyAcceptor.getConfiguration());
-               ctx.pipeline().addLast("websocket-handler", new WebSocketServerHandler(websocketSubprotocolIds, webSocketMaxFramePayloadLength != -1 ? webSocketMaxFramePayloadLength : stompMaxFramePayloadLength));
+               String encoderConfigType = ConfigurationHelper.getStringProperty(TransportConstants.WEB_SOCKET_ENCODER_TYPE, TransportConstants.DEFAULT_WEB_SOCKET_ENCODER_TYPE, nettyAcceptor.getConfiguration());
+               WebSocketFrameEncoderType encoderType = WebSocketFrameEncoderType.valueOfType(encoderConfigType);
+               if (encoderType == null) {
```
tabish121, line 165: As noted below I'd just have the 'valuefType' method validate and throw instead of needing to check here

PR #4500, opened 2023-06-05
ARTEMIS-4303 Fix windows service

five-iron:
When installing artemis on windows as a service, if your chosen installation path contained a space, then an extra escape character (%) was being inserted in the generated
artemis-service.xml, causing the service to fail to start. This fix changes the value of ARTEMIS_INSTANCE_ETC_URI in that file from "artemis.instance.etc.uri.windows" to "artemis.instance.etc.uri" to get rid of that unnecessary escape character.


clebertsuconic:
I have no way to test this.. but the changes are ok... merging it

PR #4502, opened 2023-06-06
ARTEMIS-4259 JMS consumer + FQQN + selector not working

clebertsuconic:
co-authored with Justin Bertram

### DIFF
--- a/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
+++ b/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
@@ -840,26 +840,15 @@ public class ActiveMQSession implements QueueSession, TopicSession {
                   throw new RuntimeException("Subscription name cannot be null for durable topic consumer");
                // Non durable sub
 
-               queueName = new SimpleString(UUID.randomUUID().toString());
 
-               if (!CompositeAddress.isFullyQualified(dest.getAddress())) {
-                  createTemporaryQueue(dest, RoutingType.MULTICAST, queueName, coreFilterString, response);
+               if (CompositeAddress.isFullyQualified(dest.getAddress())) {
+                  queueName = createFQQNSubscription(dest, coreFilterString, response);
                } else {
-                  if (!response.isExists() || !response.getQueueNames().contains(AutoCreateUtil.getCoreQueueName(session, dest.getSimpleAddress()))) {
-                     if (response.isAutoCreateQueues()) {
-                        try {
-                           createQueue(dest, RoutingType.MULTICAST, dest.getSimpleAddress(), null, true, true, response);
-                        } catch (ActiveMQQueueExistsException e) {
-                           // The queue was created by another client/admin between the query check and send create queue packet
-                        }
-                     } else {
-                        throw new InvalidDestinationException("Destination " + dest.getName() + " does not exist");
-                     }
-                  }
-                  queueName = CompositeAddress.extractQueueName(dest.getSimpleAddress());
+                  queueName = new SimpleString(UUID.randomUUID().toString());
+                  createTemporaryQueue(dest, RoutingType.MULTICAST, queueName, coreFilterString, response);
                }
 
-               consumer = createClientConsumer(dest, queueName, null);
+               consumer = createClientConsumer(dest, queueName, coreFilterString);
                autoDeleteQueueName = queueName;
             } else {
                // Durable sub
@@ -926,6 +915,38 @@ public class ActiveMQSession implements QueueSession, TopicSession {
       } catch (ActiveMQException e) {
          throw JMSExceptionHelper.convertFromActiveMQException(e);
       }
+   }
+
+   // This method is for the actual queue creation on the Multicast queue / subscription
+   private SimpleString createFQQNSubscription(ActiveMQDestination dest,
+                                        SimpleString coreFilterString,
+                                        AddressQuery response) throws ActiveMQException, JMSException {
+      SimpleString queueName;
+      queueName = CompositeAddress.extractQueueName(dest.getSimpleAddress());
+      if (!response.isExists() || !response.getQueueNames().contains(AutoCreateUtil.getCoreQueueName(session, dest.getSimpleAddress()))) {
+         if (response.isAutoCreateQueues()) {
+            try {
+               createQueue(dest, RoutingType.MULTICAST, dest.getSimpleAddress(), coreFilterString, true, true, response);
+               return queueName;
+            } catch (ActiveMQQueueExistsException e) {
+               // The queue was created by another client/admin between the query check and send create queue packet
+               // on this case we will switch to the regular verification to validate the coreFilterString
+            }
+         } else {
+            throw new InvalidDestinationException("Destination " + dest.getName() + " does not exist");
+         }
+      }
+
+      QueueQuery queueQuery = session.queueQuery(queueName);
+
+      if (!queueQuery.isExists()) {
+         throw new InvalidDestinationException("Destination " + queueName + " does not exist");
+      }
+
+      if (coreFilterString != null && queueQuery.getFilterString() != null && !coreFilterString.equals(queueQuery.getFilterString())) {
+         throw new JMSException(queueName + " filter mismatch [" + coreFilterString + "] is different than existing filter [" + queueQuery.getFilterString() + "]");
+      }
+      return queueName;
    }
 
    private ClientConsumer createClientConsumer(ActiveMQDestination destination, SimpleString queueName, SimpleString coreFilterString) throws ActiveMQException {
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSFQQNConsumerTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSFQQNConsumerTest.java
@@ -1,0 +1,456 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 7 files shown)


clebertsuconic:
this supersedes https://github.com/apache/activemq-artemis/pull/4452

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSFQQNConsumerTest.java
```diff
@@ -0,0 +1,209 @@
...
+      ConnectionFactory factory = CFUtil.createConnectionFactory(protocol, "tcp://localhost:5672");
+      final String queue = "queue";
+      final String address = "address";
+      final String filter = "prop='value'";
+      try (Connection c = factory.createConnection()) {
+         c.start();
+         Session s = c.createSession(false, Session.AUTO_ACKNOWLEDGE);
+         Topic t = s.createTopic(CompositeAddress.toFullyQualified(address, queue));
+         MessageConsumer mc = s.createConsumer(t, filter);
+         Wait.assertTrue(() -> server.locateQueue(queue) != null, 2000, 100);
```
rvais, line 65: I think these two requests are redundant. I don't think it's a problem, but it seems a bit strange.

rvais (commented):
Looks good. I just have a question. Would it make sense to use more consumers and more messages?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSFQQNConsumerTest.java
```diff
@@ -0,0 +1,209 @@
...
+   @Test
+   public void testFQQNTopicConsumerWithSelectorCore() throws Exception {
+      testFQQNTopicConsumerWithSelector("CORE", true);
+   }
+
+   private void testFQQNTopicConsumerWithSelector(String protocol, boolean validateFilterChange) throws Exception {
+      ConnectionFactory factory = CFUtil.createConnectionFactory(protocol, "tcp://localhost:5672");
+      final String queue = "queue";
+      final String address = "address";
+      final String filter = "prop='value'";
```
gemmellr, line 59: Using e.g 'match' and 'mismatch' would be more readable than having value of 'value' and 'novalue'.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSFQQNConsumerTest.java
```diff
@@ -0,0 +1,209 @@
...
+         message.setStringProperty("prop", "value");
+         producer.send(message);
+         Assert.assertNotNull(mc.receive(5000));
+         message = s.createTextMessage("hello");
+         message.setStringProperty("prop", "novalue");
+         producer.send(message);
+         Assert.assertNull(mc.receiveNoWait());
+      }
+
+      if (validateFilterChange) {
```
gemmellr, line 83: This only validates the case there was an existing filter. Should also validate 'changing' (adding) filter when there wasnt one (see earlier comments looking like it could NPE).

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSFQQNConsumerTest.java
```diff
@@ -0,0 +1,209 @@
...
+         Assert.assertEquals(RoutingType.ANYCAST, serverQueue.getRoutingType());
+         Assert.assertNull(serverQueue.getFilter());
+         MessageProducer p = s.createProducer(q);
+         Message m = s.createMessage();
+         m.setStringProperty("prop", "inside");
+         p.send(m);
+         assertNotNull(mc.receive(1000));
+         m = s.createMessage();
+         m.setStringProperty("prop", RandomUtil.randomString());
+         assertNull(mc.receiveNoWait());
```
gemmellr, line 202: Some newlines to group separate 'steps' would make this more readable.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSFQQNConsumerTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+         serverQueue.getConsumers().forEach(queueConsumer -> {
+            Assert.assertNotNull(queueConsumer.getFilter());
```
gemmellr, line 205: Can it match the actual filter?

To Roman's points, having more than 1 consumer with _different_ filters would seem useful in this test to verify they are handled correctly. Probably also sending a message that doesnt match any of them...check it remains, awaiting some later consumer on the queue.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSFQQNConsumerTest.java
clebertsuconic: This one is just to make sure the queue exists. The creation of the queue is asynchronous, so it should keep trying with the Wait doing a loop for 2 seconds, waiting 100 milliseconds in between.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
```diff
@@ -928,6 +917,35 @@ private ActiveMQMessageConsumer createConsumer(final ActiveMQDestination dest,
...
+         }
+      } else {
+         QueueQuery queueQuery = session.queueQuery(queueName);
+
+         if (!queueQuery.isExists()) {
+            throw new InvalidDestinationException("Destination " + queueName + " does not exist");
+         }
+
+         if (coreFilterString != null && !queueQuery.getFilterString().equals(coreFilterString)) {
+            throw new JMSException(queueName + " filter mismatch " + coreFilterString + " is different than " + queueQuery.getFilterString());
```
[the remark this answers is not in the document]
clebertsuconic: @gemmellr I made a few mistake with my git local repo and I lost a change where this was supposed to be 

!coreFilterString.equals(queueuQuery.getFilterString()).


I will add a test.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
```diff
@@ -928,6 +917,35 @@ private ActiveMQMessageConsumer createConsumer(final ActiveMQDestination dest,
...
+            throw new InvalidDestinationException("Destination " + dest.getName() + " does not exist");
+         }
+      } else {
+         QueueQuery queueQuery = session.queueQuery(queueName);
+
+         if (!queueQuery.isExists()) {
+            throw new InvalidDestinationException("Destination " + queueName + " does not exist");
+         }
+
+         if (coreFilterString != null && !queueQuery.getFilterString().equals(coreFilterString)) {
```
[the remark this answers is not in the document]
clebertsuconic: @[clebertsuconic](https://github.com/clebertsuconic) clebertsuconic [now](https://github.com/apache/activemq-artemis/pull/4502#discussion_r1221592602)
@gemmellr I made a few mistake with my git local repo and I lost a change where this was supposed to be

!coreFilterString.equals(queueuQuery.getFilterString()).

I will add a test.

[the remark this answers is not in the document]
gemmellr: You changed it further by actually adding the reverse-check, so that this seems to be saying that if you have a queue with a filter defined on it, you can attach a 'normal consumer' without a filter to that queue...but you cant connect an 'FQQN consumer' to that same queue unless it has a filter defined? If thats true it seems a bit weird.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSFQQNConsumerTest.java
gemmellr: It should double check that the forEach was actually entered, e.g count the consumers....which again there remains only one, there should be at least 2 with different filters to verify they are handled correctly.

clebertsuconic: this test will be only one.. there's another test I added that is validating filters.

It is validating the filters by matching the messages.

clebertsuconic: > It should double check that the forEach was actually entered, e.g count the consumers....which again there remains only one, there should be at least 2 with different filters to verify they are handled correctly.

testFQQNTopicMultiConsumerWithSelector was added to validate multiple consumers.

gemmellr: those are for topic usage...this one is queue usage

clebertsuconic: and I'm not interested in validating the routing itself here.. I am closing that consumer, and opening a new one now.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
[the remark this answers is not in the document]
clebertsuconic: I am going to allow to not define the selector on the consumer or on the queue. and I added tests for that.

it will be more permissive and it will still validate the case of a different filter.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSFQQNConsumerTest.java
rvais: I mean I would do the wait, but when you get the queue from that loop, it is unnecessary to repeat the request imho. I would just write that differently that's all.

PR #4504, opened 2023-06-12
ARTEMIS-4310 Add Alpine/Java17 based Docker image

thorstenhirsch:
This new Dockerfile uses Alpine Linux as base image, so the resulting image is smaller. Also, it makes use of Java 17.

It's not multi-stage. It's basically a copy of Dockerfile-ubuntu-11, but with Alpine-specific commands.

### DIFF
--- a/artemis-docker/Dockerfile-alpine-17
+++ b/artemis-docker/Dockerfile-alpine-17
@@ -1,0 +1,71 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# ActiveMQ Artemis
+
+FROM eclipse-temurin:17-jdk-alpine
+LABEL maintainer="Apache ActiveMQ Team"
+
+# add user and group for artemis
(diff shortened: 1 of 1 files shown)


thorstenhirsch:
The only testing I did so far was sending some AMQP messages from a QPID ProtonJ2 client.

clebertsuconic:
You should take a look on ./artemis-image that was added by @gtully 


The idea is to not need the ./artemis create before running a server. There will be a new EmbeddedServer that's being used by that image. 


it's an effort to simplify the images. Can you take a look?

thorstenhirsch:
./artemis-image looks useful, but I want to use "my" Docker image for an Artemis cluster with Java 17

guptatapan:
@thorstenhirsch Does AIO work for you with Alpine?
I tried it sometime back and it defaulted to use NIO at startup.
libaio was installed, but I could not get much time to look at it

thorstenhirsch:
Yes, I just checked it: `Journal Type: ASYNCIO` (on a Linux host). I didn't really know if this really means libaio, so I also checked what Artemis says on my Mac: `NIO`.

So, yes, it looks like Artemis works successfully with libaio in the Alpine Linux image.

artemis-docker/Dockerfile-alpine-17
```diff
@@ -0,0 +1,71 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# ActiveMQ Artemis
+
+FROM eclipse-temurin:17-jdk-alpine
+LABEL maintainer="T. Hirsch"
```
jbertram, line 21: This should be "Apache ActiveMQ Team" since you're contributing it to the project.

thorstenhirsch: fixed

jbertram:
@thorstenhirsch, can you squash the commits down into just one?

thorstenhirsch:
Like so...?

jbertram:
@thorstenhirsch, yes exactly. Thanks!

jbertram:
I just noticed that the `Dockerfile-alpine-17` file you added is in the root directory rather than in the `artemis-docker` directory with the other Docker files. Can you move the `Dockerfile-alpine-17` file to the `artemis-docker` directory?

thorstenhirsch:
Happened when squashing the commits. :-(
Again... fixed, this time by amending the last commit.

PR #4506, opened 2023-06-13
ARTEMIS-4312 dupes w/redistribution and multicast

jbertram:
Multiple multicast queues on the same address can lead to duplicate messages during redistribution in a cluster.


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/BindingsImpl.java
```diff
@@ -270,6 +270,11 @@ public Message redistribute(final Message message,
          logger.debug("Message {} being copied as {}", message.getMessageID(), copyRedistribute.getMessageID());
       }
       copyRedistribute.setAddress(message.getAddress());
+      for (SimpleString property : copyRedistribute.getPropertyNames()) {
```
clebertsuconic, line 273: @jbertram I think you should rather use coopyRedistribute.clearInternalProperties() here. and if that does not work, there's a predicate thing that Michael Pierce introduce that should have to be initialized.

clebertsuconic: It does work actually....

PR #4509, opened 2023-06-15
ARTEMIS-4314 support queue federation batchOnEmpty via consumerWindow…

gtully:
…Size=0

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Queue.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Queue.java
@@ -389,6 +389,8 @@ public interface Queue extends Bindable,CriticalComponent {
 
    boolean hasMatchingConsumer(Message message);
 
+   long getPendingMessageCount();
+
    Collection<Consumer> getConsumers();
 
    Map<SimpleString, Consumer> getGroups();
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
@@ -17,6 +17,7 @@
 package org.apache.activemq.artemis.core.server.federation;
 
 import java.lang.invoke.MethodHandles;
+import java.util.concurrent.Executor;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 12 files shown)


gtully:
updated to batch on excess capacity, as determined by the pending metric, we can have messages in delivery which we will now ignore for the purpose of pulling a batch. this means that if we have some spare capacity to process, messages on the upstream won't be stranded.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Queue.java
```diff
@@ -389,6 +389,8 @@ default int retryMessages(Filter filter, Integer expectedHits) throws Exception
 
    boolean hasMatchingConsumer(Message message);
```
clebertsuconic, line 391: If you're using this outside of ring, I would rename this to something else...


getRawMessageCount (terrible name I know.. but anything removing the "ring" term)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
```diff
@@ -155,6 +167,67 @@ private synchronized void connect() throws Exception {
...
+
+         @Override
+         public int getCreditWindow() {
+            return finalCreditWindow;
+         }
+      };
+   }
+
+   private void scheduleCreditOnEmpty(final int delay, final QueueHandle handle) {
+      if (handle != null) {
```
tabish121, line 203: Personally I'd treat a null handle being passed here as a terminal event and throw since it's probably gonna break something if you try and schedule for credit and don't actually schedule anything to replenish credit.

clebertsuconic:
@gtully for some reason I am not being able to make a PR to your repo. Github is not recognizing your user on the filters and giving me a hard time.

So, I created a branch you can pull the commit:

https://github.com/clebertsuconic/activemq-artemis/tree/_4314


Basically you will endup in this calling flowControl method. which needs to be protected and using the same thread used to deliver. it's not just that state.. any call to flowControl has to be done through the same thread used to deliver... otherwise you will have races on the number of credits.

clebertsuconic:
another option would be to use a sycnhronize block everywhere to make sure the flow control is correct.. so it's better to just use the same executor guaranteeing the proper threading I think.

clebertsuconic:
and BTW I wasn't able to use the ActiveMQSCheduledComponent as I suggested because of some of the state needed in the calls.

clebertsuconic:
disclaimer.. I didn't test my changes.. those are just suggestions.

gtully:
thanks @clebertsuconic
I have used the queue executor to sync on the metrics, great input thanks. That will make it more responsive, missing a trigger would incur an unnecessary delay.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
gtully: agree, that check was just defensive but unnecessary, the consumer can go stale with a pending check but that is already covered and any failure will result in the session/consumer getting recreated, so all that in necessary is that we don't reschedule in that case.
thanks for the feedback.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/Queue.java
gtully: agree, I was leaving it unchanged but calling it what it is PendingMessageCount is best, sorted. thanks.

docs/user-manual/en/federation-queue.md
```diff
@@ -49,6 +49,10 @@ e.g. as many messages as possible are consumed from the same broker as they were
...
 
 
+* Dual Federation - potential for messages to flip-flop between clusters.
+  If the backlog on your queues exceeds the available local credit across consumers, any lower priority federation consumer becomes a candidate for dispatch and messages will be federated. Eventually all messages may migrate and the scenario can repeat on the other cluster. Applying a rate limit to the connector url can help mitigate but this could have an adverse effect on migration when there are no local consumers.
+  To better support this use case, it is possible to configure the consumerWindowSize to zero on the referenced connector URI: ```tcp://<host>:<port>?consumerWindowSize=0```. This will cause the federation consumer to pull messages in batches only when the local queue has excess capacity. This means that federation won't ever drain more messaces than it can handle, such that messages would flip-flop. The batch size is derived from the relevant address settings defaultConsumerWindowSize.
```
rvais, line 54: Typo in following text: This means that federation won't ever drain more _**messaces**_ than it can hand ...
I suppose the word "messages" should be there.

PR #4512, opened 2023-06-19
ARTEMIS-4319 mitigate NPE in paging log statement

jbertram:
Also improves the wording of some several related statements.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
@@ -1573,13 +1573,13 @@ public interface ActiveMQServerLogger {
    @LogMessage(id = 224122, value = "Address {} number of messages is under page limit again, and it should be allowed to page again.", level = LogMessage.Level.INFO)
    void pageFree(SimpleString address);
 
-   @LogMessage(id = 224123, value = "Address {} has more pages than allowed. System currently has {} pages, while the estimated max number of pages is {}, based on the limitPageBytes ({}) / page-size ({})", level = LogMessage.Level.WARN)
-   void pageFullMaxBytes(SimpleString address, long pages, long maxPages, long limitBytes, long bytes);
+   @LogMessage(id = 224123, value = "Address {} has more pages than allowed. System currently has {} pages, while the estimated max number of pages is {} based on the page-limit-bytes ({}) / page-size ({})", level = LogMessage.Level.WARN)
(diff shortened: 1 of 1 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
```diff
@@ -1573,13 +1573,13 @@ void slowConsumerDetected(String sessionID,
    @LogMessage(id = 224122, value = "Address {} number of messages is under page limit again, and it should be allowed to page again.", level = LogMessage.Level.INFO)
    void pageFree(SimpleString address);
 
-   @LogMessage(id = 224123, value = "Address {} has more pages than allowed. System currently has {} pages, while the estimated max number of pages is {}, based on the limitPageBytes ({}) / page-size ({})", level = LogMessage.Level.WARN)
-   void pageFullMaxBytes(SimpleString address, long pages, long maxPages, long limitBytes, long bytes);
+   @LogMessage(id = 224123, value = "Address {} has more pages than allowed. System currently has {} pages, while the estimated max number of pages is {} based on the page-limit-bytes ({}) / page-size ({})", level = LogMessage.Level.WARN)
+   void pageFullMaxBytes(SimpleString address, long pages, Long maxPages, Long limitBytes, int bytes);
```
gemmellr, line 1577: If the value can be null then having the logger method allow for that seems reasonable...however...

If the limit value was null, why was it even logging that a non-existent limit was exceeded? Does this only happen when the value is removed/changed while already running, or something? Or is this pointing to an actual behavioural problem somewhere leading to a mis-logging?

jbertram: I'm not entirely sure how this gets logged when the limit is `null`. This is what was reported by a user on the [mailing list](https://lists.apache.org/thread/j9v7j6w0hff0sbmhn7qyzy79cj54ohhv):

> My hunch is that once we set a limit, we cannot remove the section completely, so set a very low `<page-limit-bytes>` and start the broker, let the address start paging and then remove the `<page-limit-bytes>` completely from address setting.

It looks like there may be other, larger issues present, but I wanted to mitigate the NPE before going deeper.

gemmellr: Fair enough..the change here is certainly an improvement, so I've merged it...does seem like it woudl be worth looking at why it was logging at all though.

PR #4516, opened 2023-06-20
ARTEMIS-4318 migrate to Airline 2

jbertram:
The "Airline" library we're currently using is deprecated according the GitHub project - https://github.com/airlift/airline. It recommends using either Airline 2 or Picocli. The former offers the simplest migration path as it's almost completely compatible with the current code. This commit implements that migration.

### DIFF
--- a/artemis-cli/pom.xml
+++ b/artemis-cli/pom.xml
@@ -100,7 +100,7 @@
       </dependency>
 
       <dependency>
-         <groupId>io.airlift</groupId>
+         <groupId>com.github.rvesse</groupId>
          <artifactId>airline</artifactId>
       </dependency>
       <dependency>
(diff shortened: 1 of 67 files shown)


jbertram:
BTW, the full test-suite passed with this change.

brusdev:
@jbertram I don't see any activities on https://github.com/rvesse/airline in the last 10 months, is that project  still active?

jbertram:
I'm trying to get in touch with the repo's owner Rob Vesse to see about this. He's actually a committer and PMC member of Apache Jena so that's encouraging.

clebertsuconic:
@jbertram @brusdev I am thinking we should merge this. Airlift has been pretty stable for us. at the worst case we can eventually fork it if it needed.

jbertram:
FWIW, I was never able to get in touch with Rob.

PR #4524, opened 2023-06-23
ARTEMIS-4331 bumped jgroups version

erwindon:
see https://issues.apache.org/jira/browse/ARTEMIS-4331

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -102,7 +102,7 @@
       <jsr305.version>3.0.2</jsr305.version>
       <jetty.version>10.0.15</jetty.version>
       <tomcat.servlet-api.version>8.5.78</tomcat.servlet-api.version>
-      <jgroups.version>5.2.0.Final</jgroups.version>
+      <jgroups.version>5.2.16.Final</jgroups.version>
       <errorprone.version>2.20.0</errorprone.version>
       <maven.bundle.plugin.version>5.1.9</maven.bundle.plugin.version>
       <maven.checkstyle.plugin.version>3.2.2</maven.checkstyle.plugin.version>


jbertram:
I'm setting this to draft until JGroups 5.2.15 is actually available.

erwindon:
I've updated the version designation.
Sorry, I was a bit confused about the release status as the story from jgroups was already marked with that release number.
(and you answered the question on how to proceed, thx)

erwindon:
Jgroups 5.2.15.Final has been published.
But also its successor 5.2.16.Final is already there.
I've updated the PR to use 5.2.16.Final.

erwindon:
checks failing with "Unable to resolve wrap_file__home_runner_.m2_repository_org_jgroups_jgroups_5.2.16.Final_jgroups-5.2.16.Final.jar/0.0.0: missing requirement[...]"
so seems related
back to local development...

erwindon:
Artemis build procedure works fine for current jgroups version 5.2.0.
jgroups 5.2.1 does not exist.
Artemis build procedure works fine for jgroups version 5.2.2.
jgroups 5.2.3 does not exist.
Artemis build procedure with jgroups 5.2.4+ fails in 'artemis-features'.

rebase did not change anything about that.

@jbertram 
unfortunately, I have to give up here as this is unknown territory...

jbertram:
I took a look at this, and I couldn't figure it out either. I _suspect_ there's something going awry with the inclusion of Byteman as a `compile` dependency rather than a `test` dependency. See https://github.com/belaban/JGroups/commit/69af029ee35f259ea3323552455ed411545e3c3e.

Generally speaking I dislike the `artemis-features` module because this isn't the first time it's tripped me up. I know there are folks out there who use the Karaf integration, but it seems like a rather small set of users to deal with this head-ache although I readily admit the head-ache is probably most due to my ignorance of how it works.

gemmellr:
I put together a new PR: https://github.com/apache/activemq-artemis/pull/4550

PR #4525, opened 2023-06-25
NO-JIRA fix typos in key-differences.md and authentication.md

iamgd67:
change
 `Another big difference that's good to cover early on is the difference is how message addressing and routing is done` 
to
 `Another big difference that's good to cover early on is the difference **of** how message addressing and routing is done`

and

`If your deployment is more complicated that this and requires some advanced JAAS configuration`
to
`If your deployment is more complicated **than** this and requires some advanced JAAS configuration`


gemmellr:
Squashed and pushed in https://github.com/apache/activemq-artemis/commit/e15bc1b35fef5553239d28b353c62dc488b3aedf, thanks.

PR #4526, opened 2023-06-26
ARTEMIS-4332 Add management method to close stuck server sessions

brusdev:
ARTEMIS-4332 Add management method to close stuck server sessions 
In rare cases a store operation could silently fails or starves, blocking the related server session and all delivering messages. Those server sessions can be closed adding a management method that cleans their operation context before closing them.

This PR also includes an optional commit to track store operations that would help to analyze heap dumps with stuck server sessions: ARTEMIS-4332 Add store operation trackers

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
@@ -28,6 +28,9 @@ import org.apache.activemq.artemis.core.journal.impl.SimpleWaitIOCallback;
 import org.apache.activemq.artemis.core.persistence.OperationContext;
 import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
 import org.apache.activemq.artemis.utils.ExecutorFactory;
+import org.apache.commons.collections.Buffer;
+import org.apache.commons.collections.BufferUtils;
+import org.apache.commons.collections.buffer.CircularFifoBuffer;
 
 /**
  * Each instance of OperationContextImpl is associated with an executor (usually an ordered Executor).
@@ -38,6 +41,9 @@ import org.apache.activemq.artemis.utils.ExecutorFactory;
  * If there are no pending IO operations, the tasks are just executed at the callers thread without any context switch.
  *
  * So, if you are doing operations that are not dependent on IO (e.g NonPersistentMessages) you wouldn't have any context switch.
+ *
+ * If you need to track store operations you can set the system property "ARTEMIS_OPCONTEXT_MAX_DEBUG_TRACKERS"
+ * with the max number of trackers that you want to keep in memory.
  */
 public class OperationContextImpl implements OperationContext {
 
@@ -104,6 +110,24 @@ public class OperationContextImpl implements OperationContext {
 
    private final Executor executor;
 
+   private static int maxDebugTrackers = Integer.parseInt(
+      System.getProperty("ARTEMIS_OPCONTEXT_MAX_DEBUG_TRACKERS", "0"));
+
+   private Buffer debugTrackers = OperationContextImpl.maxDebugTrackers > 0 ?
+      BufferUtils.synchronizedBuffer(new CircularFifoBuffer(OperationContextImpl.maxDebugTrackers)) : null;
+
+   protected static int getMaxDebugTrackers() {
+      return OperationContextImpl.maxDebugTrackers;
+   }
+
+   protected static void setMaxDebugTrackers(int maxDebugTrackers) {
+      OperationContextImpl.maxDebugTrackers = maxDebugTrackers;
+   }
+
+   protected Buffer getDebugTrackers() {
+      return debugTrackers;
+   }
+
    public OperationContextImpl(final Executor executor) {
       super();
       this.executor = executor;
@@ -122,7 +146,10 @@ public class OperationContextImpl implements OperationContext {
 
    @Override
    public void storeLineUp() {
-      STORE_LINEUP_UPDATER.incrementAndGet(this);
+      long storeLineUpValue = STORE_LINEUP_UPDATER.incrementAndGet(this);
+      if (debugTrackers != null) {
+         debugTrackers.add(new Exception(">" + storeLineUpValue));
+      }
    }
 
    @Override
@@ -216,7 +243,12 @@ public class OperationContextImpl implements OperationContext {
 
    @Override
    public synchronized void done() {
-      stored++;
+      this.stored++;
+
+      if (debugTrackers != null) {
+         debugTrackers.add(new Exception("<" + stored));
+      }
+
       checkTasks();
    }
 
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
@@ -447,4 +447,27 @@ public class OperationContextImpl implements OperationContext {
          executorsPendingField +
          "]";
    }
+
+   @Override
+   public synchronized void clear() {
+      stored = 0;
+      storeLineUpField = 0;
+      minimalReplicated = 0;
+      replicated = 0;
+      replicationLineUpField = 0;
+      paged = 0;
+      minimalPage = 0;
+      pageLineUpField = 0;
+      errorCode = -1;
+      errorMessage = null;
+      executorsPendingField = 0;
+
+      if (tasks != null) {
+         tasks.clear();
+      }
+
+      if (storeOnlyTasks != null) {
+         storeOnlyTasks.clear();
+      }
+   }
 }
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextUnitTest.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextUnitTest.java
@@ -422,4 +422,45 @@ public class OperationContextUnitTest extends ActiveMQTestBase {
       Assert.assertEquals(0, operations.get());
    }
 
+   @Test
+   public void testContextWithoutDebugTrackers() {
+      ExecutorService executor = Executors.newSingleThreadExecutor(ActiveMQThreadFactory.defaultThreadFactory(getClass().getName()));
+      runAfter(executor::shutdownNow);
+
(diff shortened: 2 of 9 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
```diff
@@ -415,4 +449,27 @@ public String toString() {
...
+      errorMessage = null;
+      executorsPendingField = 0;
+
+      if (tasks != null) {
+         tasks.clear();
+      }
+
+      if (storeOnlyTasks != null) {
+         storeOnlyTasks.clear();
+      }
```
gemmellr, line 473: These lists are not thread-safe, and this could potentially be a concurrent operation (people can tend to use 'force' toggles even if they dont actually need to). Was this method meant to be synchronized, or did you not want to synchronize it (although setting the other fields also wont really be safe without doing so..) ? If the latter do they need changed to a concurrent structure, or is it ok they may get corrupted in this scenario?

artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextUnitTest.java
```diff
@@ -422,4 +422,37 @@ public void onError(final int errorCode, final String errorMessage) {
...
+
+      context.done();
+      Assert.assertNull(context.getStoreOperationTrackers());
+   }
+
+   @Test
+   public void testContextWithStoreOperationTrackers() {
+      ExecutorService executor = Executors.newSingleThreadExecutor(ActiveMQThreadFactory.defaultThreadFactory(getClass().getName()));
+
+      OperationContextImpl.setMaxStoreOperationTrackers(1);
```
gemmellr, line 444: This sets a static, but it doesnt seem like it is reset later, so does it remain set for all subsequent tests?

artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextUnitTest.java
```diff
@@ -422,4 +422,37 @@ public void onError(final int errorCode, final String errorMessage) {
       Assert.assertEquals(0, operations.get());
    }
 
+   @Test
+   public void testContextWithoutStoreOperationTrackers() {
+      ExecutorService executor = Executors.newSingleThreadExecutor(ActiveMQThreadFactory.defaultThreadFactory(getClass().getName()));
+
+      OperationContextImpl.setMaxStoreOperationTrackers(0);
```
gemmellr, line 429: Rather than set it 0, could it just verify it is 0, i.e test the default?

gemmellr (commented):
Adding JIRA ref in PR title would be good.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
brusdev: @gemmellr yes this could potentially be a concurrent operation and it could cause inconsistent state. I could change it to be thread-safe but I'm not sure this is necessary. I mean this method should be the last chance to close a stuck session and it is supposed to be manually called by a user. If the call is successful the context/session will be closed otherwise users could retry.

gemmellr: So I guess the question is really just, what are the potential consequences of corrupting the list and is it important?

Closing 'stuck' sessions is what the method is being added for, but its not necessarily when it will actually be called. I expect some folks will use the 'force' just because it exists at all, to avoid having to think whether they actually need it or not. E.g the same way folks use rm -rf a lot of the time.

brusdev: The potential consequences of corrupting should only be related to the closing session itself and its context but I see your point I'll change it to be thread-safe.

artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextUnitTest.java
```diff
@@ -422,4 +422,43 @@ public void onError(final int errorCode, final String errorMessage) {
       Assert.assertEquals(0, operations.get());
    }
 
+   @Test
+   public void testContextWithoutStoreOperationTrackers() {
+      ExecutorService executor = Executors.newSingleThreadExecutor(ActiveMQThreadFactory.defaultThreadFactory(getClass().getName()));
+
+      OperationContextImpl.resetMaxStoreOperationTrackers();
```
gemmellr, line 429: Slight nitpick but I think it would be better if it didnt do this reset first, and so verified the default behaviour was as expected?

gemmellr (commented):
LGTM other than the slight nitpick. I'd like someone else to review it too though.

brusdev:
@gemmellr thanks for your review

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
```diff
@@ -104,6 +107,29 @@ public static void setContext(final OperationContext context) {
 
    private final Executor executor;
 
+   public static final String MAX_STORE_OPERATION_TRACKERS_PROP = "artemis.maxStoreOperationTrackers";
+
+   private static int maxStoreOperationTrackers = -1;
```
clebertsuconic, line 112: - I think this should be final 
and it should be initialized with a static initializer, or simply use the System.getProperty straight away.

{
}

- Also the property name.  and the pattern used here.

RefCountMessage used a different pattern. I think we should stick to the same pattern used in there.


ARTEMIS_


and maxStoreOperationsTrackers.. it doesn't not really translate to anything unless I read the code and understand what is doing.

I'm not sure what better name we could use (I'm terrible with names). but what about:

ARTEMIS_OPCONTEXT_TRACKER_NUMBER?

And some javadoc of what it does?

brusdev: What about `ARTEMIS_OPCONTEXT_MAX_DEBUG_TRACKERS`/`maxDebugTrackers` ?

artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextUnitTest.java
```diff
@@ -422,4 +422,43 @@ public void onError(final int errorCode, final String errorMessage) {
       Assert.assertEquals(0, operations.get());
    }
 
+   @Test
+   public void testContextWithoutDebugTrackers() {
+      ExecutorService executor = Executors.newSingleThreadExecutor(ActiveMQThreadFactory.defaultThreadFactory(getClass().getName()));
```
clebertsuconic, line 427: You are not shutting down this executor. it's leaking a thread.

clebertsuconic: add a runAfter(executor::shutdownNow);

artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextUnitTest.java
```diff
@@ -422,4 +422,43 @@ public void onError(final int errorCode, final String errorMessage) {
...
+      Assert.assertNull(context.getDebugTrackers());
+
+      context.done();
+      Assert.assertNull(context.getDebugTrackers());
+   }
+
+   @Test
+   public void testContextWithDebugTrackers() {
+      int maxStoreOperationTrackers = OperationContextImpl.getMaxDebugTrackers();
+      ExecutorService executor = Executors.newSingleThreadExecutor(ActiveMQThreadFactory.defaultThreadFactory(getClass().getName()));
```
clebertsuconic, line 443: runAfter(executor::shutdownDown);

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
```diff
@@ -104,6 +110,24 @@ public static void setContext(final OperationContext context) {
...
+      System.getProperty("ARTEMIS_OPCONTEXT_MAX_DEBUG_TRACKERS", "0"));
+
+   private Buffer debugTrackers = OperationContextImpl.maxDebugTrackers > 0 ?
+      BufferUtils.synchronizedBuffer(new CircularFifoBuffer(OperationContextImpl.maxDebugTrackers)) : null;
+
+   public static int getMaxDebugTrackers() {
+      return OperationContextImpl.maxDebugTrackers;
+   }
+
+   public static void setMaxDebugTrackers(int maxDebugTrackers) {
```
clebertsuconic, line 123: no need to be public, the test is the same package I think

artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextUnitTest.java
brusdev: done

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
brusdev: done

artemis-server/src/test/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextUnitTest.java
brusdev: done

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java
brusdev: done

clebertsuconic:
@brusdev only thing I could ask now is to verify if the new test over Core is failing in a loop or not? We had a few overCore tests failing in management because of noise from the Core itself (the testing framework itself is creating a connection and a session in the server's).


Other than that I think we should merge this.

PR #4529, opened 2023-06-29
ARTEMIS-4095: fix delivering message size accounting

artyom-tarasenko:
https://issues.apache.org/jira/browse/ARTEMIS-4095

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
@@ -288,9 +288,7 @@ public class AMQConsumer {
             return 0;
          }
 
-         if (session.getConnection().isNoLocal() || session.isInternal()) {
-            //internal session always delivers messages to noLocal advisory consumers
-            //so we need to remove this property too.
+         if (session.getConnection().isNoLocal() || (session.isInternal() && AdvisorySupport.isAdvisoryTopic(openwireDestination))) {
             message.removeProperty(MessageUtil.CONNECTION_ID_PROPERTY_NAME);
          }
          //handleDeliver is performed by an executor (see JBPAPP-6030): any AMQConsumer can share the session.wireFormat()
--- a/artemis-protocols/artemis-openwire-protocol/src/test/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumerTest.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/test/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumerTest.java
@@ -16,29 +16,55 @@
  */
 package org.apache.activemq.artemis.core.protocol.openwire.amq;
 
+import static org.junit.Assert.assertEquals;
+
 import java.util.concurrent.ScheduledExecutorService;
 
 import org.apache.activemq.artemis.api.core.ICoreMessage;
+import org.apache.activemq.artemis.api.core.Message;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.message.impl.CoreMessage;
 import org.apache.activemq.artemis.core.protocol.openwire.OpenWireConnection;
+import org.apache.activemq.artemis.core.protocol.openwire.OpenWireMessageConverter;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
 import org.apache.activemq.artemis.core.server.MessageReference;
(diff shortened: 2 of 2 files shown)


jbertram:
This fix looks good as far as I can see, but I still don't understand why your session is being considered "internal" in the first place. Can you shed any light on this?

jbertram:
Checkstyle issues:
```
Error:  src/test/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumerTest.java:[52,5] (indentation) Indentation: 'member def modifier' has incorrect indentation level 4, expected level should be 3.
Error:  src/test/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumerTest.java:[53,5] (indentation) Indentation: 'member def modifier' has incorrect indentation level 4, expected level should be 3.
Error:  src/test/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumerTest.java:[55,5] (indentation) Indentation: 'method def modifier' has incorrect indentation level 4, expected level should be 3.
Error:  src/test/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumerTest.java:[66] (regexp) RegexpSingleline: Line has trailing spaces.
Error:  src/test/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumerTest.java:[67,5] (indentation) Indentation: 'method def rcurly' has incorrect indentation level 4, expected level should be 3.
```

artyom-tarasenko:
> This fix looks good as far as I can see, but I still don't understand why your session is being considered "internal" in the first place. Can you shed any light on this?

Basically I think it's explained in the comment here:
https://github.com/apache/activemq-artemis/blob/main/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java#L189

InternalSessions are used for XA-Operations, so the sessions from a resource-adapter ra are internal:

https://github.com/apache/activemq-artemis/blob/main/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java#L1464
https://github.com/apache/activemq-artemis/blob/main/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java#L1471

jbertram:
I'm still confused because the `internalSession` from `org.apache.activemq.artemis.core.protocol.openwire.OpenWireConnection` is a `org.apache.activemq.artemis.core.server.ServerSession` whereas the `session` variable being checked in `org.apache.activemq.artemis.core.protocol.openwire.amq.AMQConsumer#handleDeliver` (which you changed) is a `org.apache.activemq.artemis.core.protocol.openwire.amq.AMQSession`. I don't see the overlap here. This value appears to be coming from the OpenWire _client_.

artyom-tarasenko:
> I'm still confused because the `internalSession` from `org.apache.activemq.artemis.core.protocol.openwire.OpenWireConnection` is a `org.apache.activemq.artemis.core.server.ServerSession` whereas the `session` variable being checked in `org.apache.activemq.artemis.core.protocol.openwire.amq.AMQConsumer#handleDeliver` (which you changed) is a `org.apache.activemq.artemis.core.protocol.openwire.amq.AMQSession`. I don't see the overlap here. This value appears to be coming from the OpenWire _client_.

The bug is triggered when the message is delivered to a client. The way from the client to server accounts clientid, but when the message is consumed by ra client the CONNECTION_ID_PROPERTY is removed. If I understand the flow correctly, the server opens a transaction, creates an internal session and uses it to transfer the data to a subscriber.

jbertram:
> If I understand the flow correctly, the server creates an internal session, opens a transaction and uses it to transfer the data to a subscriber.

As noted previously, the `internalSession` you pointed out in `o.a.a.a.c.p.o.OpenWireConnection` is an `o.a.a.a.c.s.ServerSession`. This is a _core_ session, and it can't be used to transfer data to an OpenWire client. Furthermore, it doesn't have anything to do with the `o.a.a.a.c.p.o.a.AMQSession` for which `isInternal()` is returning `true` and triggering this problem.

I looked through the source code of the OpenWire JMS and I see a handful of places where the session ID is set to `-1` which would cause `o.a.a.a.c.p.o.a.AMQSession#isInternal()` to return `true`. I still don't understand _why_ this is done, but it helps to understand where it's coming from.

Please take care of the checkstyle issues with your PR and then I'll merge it. Thanks!

jbertram:
Please squash your commits and ensure the commit message follows the [Hacking Guide](https://github.com/apache/activemq-artemis/blob/main/docs/hacking-guide/en/code.md#commitMessageDetails). Thanks!

artyom-tarasenko:
Squashed

jbertram:
@artyom-tarasenko, thanks for the contribution!

PR #4530, opened 2023-06-30
ARTEMIS-4338 + a couple of related fixes

jbertram:


jbertram:
I need some help with the implementation of `org.apache.activemq.artemis.core.protocol.ProtocolHandler.ProtocolDecoder#exceptionCaught` so anybody with Netty know-how please speak up!

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/ProtocolHandler.java
```diff
@@ -243,6 +243,18 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
          ctx.flush();
       }
 
+      @Override
+      public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+         // We definitely want to log the failure
+         ActiveMQServerLogger.LOGGER.failureDuringProtocolHandshake(ctx.channel().remoteAddress().toString(), cause);
+
+         // However, what else should be done here? I've laid out a few options...
+         ctx.fireChannelInactive();
```
tabish121, line 252: You shouldn't need to do this directly unless there's something in there you really need to run right now.

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/ProtocolHandler.java
```diff
@@ -243,6 +243,18 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
+         ctx.close(); // this is done frequently in other implementations of exceptionCaught()
+         ctx.fireExceptionCaught(cause);
```
tabish121, line 254: If you are fully handling the exception caught case you don't need to forward this, but if there's another handler that is in the chain that you want to see the event then you could forward it.

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/ProtocolHandler.java
```diff
@@ -243,6 +243,18 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
...
+         ctx.pipeline().remove(this); // this is done at the end of decode(); should we do this as well?
```
tabish121, line 255: If the outcome is terminal then you shouldn't need to remove this, depends on the handler but in general the pipeline is going to get torn down if you close the context.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/logging/AuditLoggerTest.java
```diff
@@ -117,6 +121,56 @@ public void testAuditLog() throws Exception {
...
+
+   @Test
+   public void testAMQPConnectionAuditLog() throws Exception {
+      testConnectionAuditLog("AMQP");
+   }
+
+   @Test
+   public void testOpenWireConnectionAuditLog() throws Exception {
+      testConnectionAuditLog("OPENWIRE");
+   }
```
gemmellr, line 137: Since these are considered 'resource' logs, they might be better in AuditLoggerResourceTest

tests/smoke-tests/src/main/resources/servers/audit-logging2/log4j2.properties
```diff
@@ -40,7 +40,7 @@ logger.audit_base = INFO, audit_log_file
 logger.audit_base.name = org.apache.activemq.audit.base
 logger.audit_base.additivity = false
 
-logger.audit_resource = OFF, audit_log_file
+logger.audit_resource = INFO, audit_log_file
```
gemmellr, line 43: Related to other comment, the resource logger is enabled in the AuditLoggerResourceTest test config already.

tests/smoke-tests/pom.xml
```diff
@@ -188,6 +188,13 @@
          <version>${project.version}</version>
          <scope>test</scope>
       </dependency>
+      <dependency>
+         <groupId>org.apache.activemq.tests</groupId>
+         <artifactId>integration-tests</artifactId>
+         <version>${project.version}</version>
+         <type>test-jar</type>
+         <scope>test</scope>
+      </dependency>
```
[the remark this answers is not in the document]
gemmellr: PR raised: #4536. Were that merged you should be able to just remove this dep, as the module already depends on artemis-test-support.

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/ProtocolHandler.java
```diff
@@ -243,6 +243,12 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out) t
          ctx.flush();
       }
 
+      @Override
+      public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+         ActiveMQServerLogger.LOGGER.failureDuringProtocolHandshake(ctx.channel().remoteAddress().toString(), cause);
+         ctx.close();
```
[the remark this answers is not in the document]
jbertram: You mean something like this?
```java
      @Override
      public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
         try {
            ActiveMQServerLogger.LOGGER.failureDuringProtocolHandshake(ctx.channel().remoteAddress().toString(), cause);
         } finally {
            ctx.close();
         }
      }
```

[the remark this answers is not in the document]
gemmellr: Yep

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
```diff
@@ -582,7 +582,7 @@ public void connectionCreated(final ActiveMQComponent component,
    public void addConnectionEntry(Connection connection, ConnectionEntry entry) {
       connections.put(connection.getID(), entry);
       if (AuditLogger.isResourceLoggingEnabled()) {
-         AuditLogger.createdConnection(connection.getProtocolConnection().getProtocolName(), connection.getID(), connection.getRemoteAddress());
+         AuditLogger.createdConnection(connection.getProtocolConnection() == null ? null : connection.getProtocolConnection().getProtocolName(), connection.getID(), connection.getRemoteAddress());
```
gemmellr, line 585: Having the local-address in the actual audit log also seems like it would be [more] useful here, as in the handshake failure case. Probably a separate change though.

tests/smoke-tests/src/main/resources/servers/audit-logging2/log4j2.properties
```diff
@@ -30,6 +30,9 @@ logger.artemis_journal.level=INFO
 logger.artemis_utils.name=org.apache.activemq.artemis.utils
 logger.artemis_utils.level=INFO
 
+logger.StompConnection.name=org.apache.activemq.artemis.core.protocol.stomp.StompConnection
+logger.StompConnection.level=TRACE
+
```
gemmellr, line 35: Is this still needed if the new tests moved to AuditLoggerResourceTest? (looks to use different server config)

jbertram: I left this in by mistake.

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
jbertram: I agree on both counts.

PR #4533, opened 2023-07-03
ARTEMIS-4340: fix ThreadLocalByteBufferPool.borrow zeroing

gemmellr:
The report for ARTEMIS-4340 was raised about ByteUtil.uncheckedZeros throwing IndexOutOfBoundsException. Though some of ByteUtil's behaviour can perhaps be questioned around its consistency with different buffer types, in this case the problem ultimately starts earlier in the call tree with org.apache.activemq.artemis.core.io.util.ThreadLocalByteBufferPool.borrow(int, boolean) passing it invalid arguments. It passes in a buffer for zeroing, with its limit still set below the requested size to be zero'd, leading to the IndexOutOfBoundsException in this case when tryign to write beyond the limit. The reason this seemingly long-standing bug has perhaps not been hit so far is that it occurs in a less typical case where the buffer either isnt direct and able to be updated with Unsafe, or also doesnt have an array.

ThreadLocalByteBufferPool should clear position+limit before zeroing the buffer.

### DIFF
--- a/artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/util/ThreadLocalByteBufferPool.java
+++ b/artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/util/ThreadLocalByteBufferPool.java
@@ -44,6 +44,7 @@ final class ThreadLocalByteBufferPool implements ByteBufferPool {
       } else {
          bytesPool.set(null);
          if (zeroed) {
+            byteBuffer.clear();
             ByteUtil.zeros(byteBuffer, 0, size);
          }
          byteBuffer.clear();
--- a/artemis-journal/src/test/java/org/apache/activemq/artemis/core/io/util/ThreadLocalByteBufferPoolTest.java
+++ b/artemis-journal/src/test/java/org/apache/activemq/artemis/core/io/util/ThreadLocalByteBufferPoolTest.java
@@ -16,6 +16,12 @@
  */
 package org.apache.activemq.artemis.core.io.util;
 
+import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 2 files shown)


artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/util/ThreadLocalByteBufferPool.java
```diff
@@ -44,6 +44,7 @@ public ByteBuffer borrow(final int size, boolean zeroed) {
       } else {
          bytesPool.set(null);
          if (zeroed) {
+            byteBuffer.clear();
```
gemmellr, line 47: An alternative could be moving this before the if, and replacing the call present after the if already..but its probably safer just to leave both for all it actually does (sets a few ints).

dhoffer:
What version of Atremis is this planned for?  Since it affects all versions, would this be a service pack fix on several/all versions?  Or is this just going to go into the baseline for the next minor release?  Or preference would be service packs on the last few versions so we have more flexibility on which versions we could use.

gemmellr:
Whatever the next release is, most likely 2.30.0.

jbertram:
As far as I'm aware ActiveMQ has never done a "service pack." We've done micro releases, especially on the "Classic" broker. However, in this case we'll just do a normal minor release. If there's an older version where you really need the fix immediately while you prepare for an upgrade you can cherry-pick this fix on top of that version's tag, rebuild it, and deploy the new `artemis-journal` jar.

tabish121:
LGTM

clebertsuconic:
we are doing a release early next week.

PR #4536, opened 2023-07-05
ARTEMIS-4346: move stomp test client to artemis-test-support module

gemmellr:
Move the stomp test client to artemis-test-support module, where its transport innards already live, to easily share it amongst different test modules.


tests/artemis-test-support/src/main/java/org/apache/activemq/artemis/tests/integration/stomp/util/AbstractStompClientConnection.java
```diff
@@ -91,8 +90,6 @@ public AbstractStompClientConnection(URI uri) throws Exception {
       transport.setTransportListener(new StompTransportListener());
       transport.connect();
 
-      Wait.waitFor(() -> transport.isConnected(), 1000);
```
gemmellr, line 94: I could have replaced this with the [parent] Wait class from artemis-unit-test-support, but since the transport connect is actually synchronous it didnt seem needed and I removed it instead.

PR #4539, opened 2023-07-05
ARTEMIS-4348 Parse connectors on broker.xml for default URI

clebertsuconic:


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/ActionAbstract.java
```diff
@@ -113,7 +115,23 @@ public String getBrokerURLInstance(String acceptor) {
...
+
+                  if (InetAddress.getByName(host).isAnyLocalAddress()) {
+                     host = "localhost";
+                  }
+
+                  return new URI(scheme, null, host, port, null, null, null).toString();
+               }
+            }
          } catch (Exception e) {
+            e.printStackTrace();
```
brusdev, line 134: Why is not the stack trace printed using `getActionContext().out` ?

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/Transfer.java
```diff
@@ -43,7 +43,7 @@ public class Transfer extends InputAbstract {
    protected String sourceURL = DEFAULT_BROKER_URL;
 
    @Option(name = "--source-acceptor", description = "Acceptor used to build URL towards the broker. Default: 'artemis'.")
-   protected String sourceAcceptor;
+   protected String sourceAcceptor = "artemis";
```
brusdev, line 46: The `DEFAULT_BROKER_ACCEPTOR` could be reused here:
```suggestion
   protected String sourceAcceptor = DEFAULT_BROKER_ACCEPTOR;
```

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/ActionAbstract.java
clebertsuconic: I didn't mean to commit this line.  Will remove it.

clebertsuconic: I will actually keep the e.printStackTrace with your suggestion.

brusdev (approved):

PR #4540, opened 2023-07-05
ARTEMIS-4349 Replace Guava cache with Caffeine

amarkevich:
based on benchmark https://github.com/ben-manes/caffeine/wiki/Benchmarks

### DIFF
--- a/artemis-hawtio/artemis-console/pom.xml
+++ b/artemis-hawtio/artemis-console/pom.xml
@@ -55,18 +55,6 @@
       <version>${hawtio.version}</version>
       <scope>provided</scope>
     </dependency>
-    <dependency>
-      <groupId>com.google.guava</groupId>
-      <artifactId>guava</artifactId>
-      <!-- License: Apache 2.0 -->
-      <!-- see https://github.com/google/guava/issues/3006: checker-compat-qual could be excluded -->
-      <exclusions>
-        <exclusion>
-          <groupId>org.checkerframework</groupId>
-          <artifactId>checker-compat-qual</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
   </dependencies>
 
   <build>
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTUtil.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTUtil.java
@@ -20,7 +20,6 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 
-import com.google.common.base.CaseFormat;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufUtil;
 import io.netty.handler.codec.EncoderException;
@@ -49,6 +48,7 @@ import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.config.WildcardConfiguration;
 import org.apache.activemq.artemis.core.message.impl.CoreMessage;
 import org.apache.activemq.artemis.reader.MessageUtil;
+import org.apache.commons.text.CaseUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.lang.invoke.MethodHandles;
@@ -380,7 +380,7 @@ public class MQTTUtil {
    }
 
    private static String formatCase(String string) {
-      return CaseFormat.UPPER_UNDERSCORE.to(CaseFormat.LOWER_CAMEL, string);
+      return CaseUtils.toCamelCase(string, false, '_');
    }
 
    private static String getPayloadForLogging(MqttPublishMessage message, int maxPayloadLogSize) {
--- a/pom.xml
+++ b/pom.xml
@@ -97,7 +97,7 @@
       <commons.io.version>2.10.0</commons.io.version>
       <commons.codec.version>1.15</commons.codec.version>
       <fuse.mqtt.client.version>1.16</fuse.mqtt.client.version>
-      <guava.version>32.0.1-jre</guava.version>
+      <caffeine.version>3.1.6</caffeine.version>
       <hawtio.version>2.15.0</hawtio.version>
       <jsr305.version>3.0.2</jsr305.version>
       <jetty.version>10.0.15</jetty.version>
@@ -110,6 +110,7 @@
       <sevntu.checks.version>1.39.0</sevntu.checks.version>
       <checkstyle.version>10.11.0</checkstyle.version>
       <mockito.version>5.4.0</mockito.version>
+      <classgraph.version>4.8.161</classgraph.version>
       <jctools.version>2.1.2</jctools.version>
       <netty.version>4.1.94.Final</netty.version>
       <hdrhistogram.version>2.1.12</hdrhistogram.version>
@@ -382,6 +383,14 @@
          </dependency>
 
          <dependency>
+            <groupId>io.github.classgraph</groupId>
+            <artifactId>classgraph</artifactId>
+            <version>${classgraph.version}</version>
+            <scope>test</scope>
+            <!-- License: MIT -->
+         </dependency>
+
+         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-minikdc</artifactId>
             <version>${hadoop.minikdc.version}</version>
@@ -410,12 +419,6 @@
          </dependency>
 
          <!-- ## End Test Dependencies ## -->
-         <dependency>
-             <groupId>com.google.errorprone</groupId>
-             <artifactId>error_prone_core</artifactId>
-             <version>${errorprone.version}</version>
-             <scope>provided</scope>
-         </dependency>
          <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-core</artifactId>
@@ -915,9 +918,9 @@
             <!-- License: Apache 2.0 -->
          </dependency>
          <dependency>
-            <groupId>com.google.guava</groupId>
-            <artifactId>guava</artifactId>
-            <version>${guava.version}</version>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <version>${caffeine.version}</version>
             <!-- License: Apache 2.0 -->
          </dependency>
          <dependency>
@@ -1078,12 +1081,6 @@
                   <name>errorprone</name>
               </property>
           </activation>
-          <dependencies>
-              <dependency>
-                  <groupId>com.google.errorprone</groupId>
-                  <artifactId>error_prone_core</artifactId>
-              </dependency>
-          </dependencies>
           <build>
               <plugins>
                   <plugin>
@@ -1097,6 +1094,13 @@
                               <arg>-XDcompilePolicy=simple</arg>
                               <arg>-Xplugin:ErrorProne -Xep:ThreadLocalUsage:ERROR -Xep:MissingOverride:ERROR -Xep:NonAtomicVolatileUpdate:ERROR -Xep:SynchronizeOnNonFinalField:ERROR -Xep:StaticQualifiedUsingExpression:ERROR -Xep:WaitNotInLoop:ERROR -Xep:BanJNDI:OFF -XepExcludedPaths:.*/generated-sources/.*</arg>
                           </compilerArgs>
+                          <annotationProcessorPaths>
+                              <path>
+                                  <groupId>com.google.errorprone</groupId>
+                                  <artifactId>error_prone_core</artifactId>
+                                  <version>${errorprone.version}</version>
+                              </path>
+                          </annotationProcessorPaths>
                       </configuration>
                   </plugin>
               </plugins>
@@ -1111,12 +1115,6 @@
                   <name>errorprone</name>
               </property>
           </activation>
-          <dependencies>
-              <dependency>
-                  <groupId>com.google.errorprone</groupId>
-                  <artifactId>error_prone_core</artifactId>
-              </dependency>
-          </dependencies>
           <build>
               <plugins>
                   <plugin>
@@ -1141,6 +1139,13 @@
                               <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED</arg>
                               <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED</arg>
                           </compilerArgs>
+                          <annotationProcessorPaths>
+                              <path>
+                                  <groupId>com.google.errorprone</groupId>
+                                  <artifactId>error_prone_core</artifactId>
+                                  <version>${errorprone.version}</version>
+                              </path>
+                          </annotationProcessorPaths>
                       </configuration>
                   </plugin>
               </plugins>
--- a/pom.xml
+++ b/pom.xml
@@ -420,6 +420,12 @@
 
          <!-- ## End Test Dependencies ## -->
          <dependency>
+             <groupId>com.google.errorprone</groupId>
+             <artifactId>error_prone_core</artifactId>
+             <version>${errorprone.version}</version>
+             <scope>provided</scope>
+         </dependency>
+         <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-core</artifactId>
             <version>${spring.version}</version>
@@ -1081,6 +1087,12 @@
                   <name>errorprone</name>
               </property>
           </activation>
+          <dependencies>
+              <dependency>
+                  <groupId>com.google.errorprone</groupId>
+                  <artifactId>error_prone_core</artifactId>
+              </dependency>
+          </dependencies>
           <build>
               <plugins>
                   <plugin>
@@ -1094,13 +1106,6 @@
                               <arg>-XDcompilePolicy=simple</arg>
                               <arg>-Xplugin:ErrorProne -Xep:ThreadLocalUsage:ERROR -Xep:MissingOverride:ERROR -Xep:NonAtomicVolatileUpdate:ERROR -Xep:SynchronizeOnNonFinalField:ERROR -Xep:StaticQualifiedUsingExpression:ERROR -Xep:WaitNotInLoop:ERROR -Xep:BanJNDI:OFF -XepExcludedPaths:.*/generated-sources/.*</arg>
                           </compilerArgs>
-                          <annotationProcessorPaths>
-                              <path>
-                                  <groupId>com.google.errorprone</groupId>
-                                  <artifactId>error_prone_core</artifactId>
-                                  <version>${errorprone.version}</version>
-                              </path>
-                          </annotationProcessorPaths>
                       </configuration>
                   </plugin>
               </plugins>
@@ -1115,6 +1120,12 @@
                   <name>errorprone</name>
               </property>
           </activation>
+          <dependencies>
+              <dependency>
+                  <groupId>com.google.errorprone</groupId>
+                  <artifactId>error_prone_core</artifactId>
+              </dependency>
+          </dependencies>
           <build>
               <plugins>
                   <plugin>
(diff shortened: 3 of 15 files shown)


clebertsuconic:
any idea how active is Caffeine? more specifically how good are them with fixing CVEs?


there was a couple of CVEs in the past with guava.  https://www.cvedetails.com/product/52274/Google-Guava.html?vendor_id=1224 and we had to update the version before.

my concern with the replacement is how good would be fix things like these?

clebertsuconic:
well.. at least some initial googling shows that they are active on fixing this:

https://security.snyk.io/package/maven/com.github.ben-manes.caffeine:caffeine


any opinions?

tabish121:
I'm no fan of guava but my first question on any change like this is what benefit if any does it actually impart? Is there some measurable performance improvement worth changing it now vs at some later stage like a major version bump where these types of changes are more expected? (e.g. Artemis 3.0)

jbertram:
Caffeine is essentially the successor to Google's Guava Cache. It's used by a lot of projects. Spring, for example, started moving [away from Guava to Caffeine in 2015](https://github.com/spring-projects/spring-framework/issues/18370). I considered using Caffeine when I first started using Guava Cache on [ARTEMIS-2886](https://issues.apache.org/jira/browse/ARTEMIS-2886), but I decided on Guava mainly because we already had a dependency on it. This PR completely removes Guava from the code-base in lieu of Caffeine which is fine by me.

My only real concern is essentially the same as Tim's which I put on [ARTEMIS-4349](https://issues.apache.org/jira/browse/ARTEMIS-4349?focusedCommentId=17740337&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17740337). Changing Caffeine for the sake of performance doesn't make much sense without clear data that changing will actually make any meaningful difference. At this point we don't have that data.

clebertsuconic:
@jbertram based on what you said it sounds like caffeine has a bigger community than guava? that would be a good reason for the replacement by itself.

clebertsuconic:
I'm running the whole test suite with these changes (on a separate CI).

If caffeine has a bigger community, the jar is smaller, I think we should bring this right away (if tests are ok of course).

my initial concern  was about the community support and CVEs.. but from what I later gathered this is actually better.

performance wise I think the differences would be impossible to measure. We currently use some login changes from Security as Justin mentioned... but you would have to use the cache quite a lot to have any significant result.

I think we should merge this, but not much for the performance reasons.


anyone has anything against that?

jbertram:
Guava itself is much more than just the cache, but if you're comparing Guava Cache vs. Caffeine then I think more folks are using Caffeine.

Also, since Guava is a lot more than just a cache that means Caffeine is much smaller and will likely reduce risk of CVEs.

clebertsuconic:
apparently there are a lot of test failures if running this. leave it with me for another day please so I can check if it was a real failure or a build issue.

ben-manes:
Hi,

If you do not have a performance reason to switch then staying with Guava's is perfectly fine. I would not use their `asMap().compute` as a wonky implementation given it was added so late (and you don't use it, so that's fine). Guava's is robust but not actively maintained (CacheBuilder's JavaDoc recommends Caffeine).

The most common cause for tests to fail during a migration is that Caffeine switches eviction and listeners to be executed async by default, whereas Guava piggybacks on the calling thread. You can emulate that using `Caffeine.executor(Runnable::run)` if desired. The problem is that often tests assumed no concurrency (immediate eviction or callbacks invoked) and that is no longer a valid assumption. In both cache's their internal logic is very inexpensive (e.g. LRU reordering), but we do not know the cost of the user's callback. Since the JVM now offers shared threads and Caffeine offers an AsyncCache, defaulting to async to minimize user-facing latencies seemed like a reasonable choice. There are other subtle differences, such as if assuming lru eviction order, so you might want to review this [migration page](https://github.com/ben-manes/caffeine/wiki/Guava).

Whatever you decide to do is fine with me. Feel welcome to reach out if you have any questions or concerns.

artemis-hawtio/artemis-console/pom.xml
```diff
@@ -55,18 +55,6 @@
...
-      <artifactId>guava</artifactId>
-      <!-- License: Apache 2.0 -->
-      <!-- see https://github.com/google/guava/issues/3006: checker-compat-qual could be excluded -->
-      <exclusions>
-        <exclusion>
-          <groupId>org.checkerframework</groupId>
-          <artifactId>checker-compat-qual</artifactId>
-        </exclusion>
-      </exclusions>
-    </dependency>
```
gemmellr, line 69: This was here to override the version of guava in the repackaged hawtio war for the console. See the exclusion further down that is still present. Think we will need to check what happens to the management console here, its possible this may have broken it, but at least removed the version override.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTUtil.java
```diff
@@ -380,7 +380,7 @@ private static String formatByte(byte bite) {
    }
 
    private static String formatCase(String string) {
-      return CaseFormat.UPPER_UNDERSCORE.to(CaseFormat.LOWER_CAMEL, string);
+      return CaseUtils.toCamelCase(string, false, ' ');
```
gemmellr, line 383: This seems off? If its going from _UPPER_UNDERSCORE_ its unlikely to have space delimiters..seems more like it will end up with _lower_underscore_ after this? Or does the original also not do what it seems like its trying?

pom.xml
```diff
@@ -1097,6 +1094,13 @@
                               <arg>-XDcompilePolicy=simple</arg>
                               <arg>-Xplugin:ErrorProne -Xep:ThreadLocalUsage:ERROR -Xep:MissingOverride:ERROR -Xep:NonAtomicVolatileUpdate:ERROR -Xep:SynchronizeOnNonFinalField:ERROR -Xep:StaticQualifiedUsingExpression:ERROR -Xep:WaitNotInLoop:ERROR -Xep:BanJNDI:OFF -XepExcludedPaths:.*/generated-sources/.*</arg>
                           </compilerArgs>
+                          <annotationProcessorPaths>
+                              <path>
+                                  <groupId>com.google.errorprone</groupId>
+                                  <artifactId>error_prone_core</artifactId>
+                                  <version>${errorprone.version}</version>
+                              </path>
+                          </annotationProcessorPaths>
```
gemmellr, line 1103: I know that Emmanuel deliberately avoided doing this when overhauling the ErrorProne usage a while back for JDK > 9, as it restricts use of annotation processors from the classpath and requires every processor be specified here in the XML. We do have another annotation processor being used via the classpath (artemis-log-annotation-processor).

I would guess that is what caused the 'full test run' CI failures @clebertsuconic  noted since I know that build enables ErrorProne during compile, meaning it probably then failed to run the other annotation processor as it isnt also specified here (or rather outwith these profiles), and so failed to generate the logger classes and then tripped over their absence.

Since ErrorProne is already isolated to a profile I think I would leave this as it was. At least I dont think it really belongs in the same commit as a swap of caching to use Caffeine.

artemis-hawtio/artemis-console/pom.xml
amarkevich: Content of artemis-console-2.30.0-SNAPSHOT.war\WEB-INF\lib :
checker-qual-3.5.0.jar
commons-codec-1.15.jar
commons-io-2.11.0.jar
commons-logging-1.2.jar
error_prone_annotations-2.3.4.jar
failureaccess-1.0.1.jar
hawtio-core-2.15.0.jar
hawtio-system-2.15.0.jar
hawtio-util-2.15.0.jar
httpclient-4.5.13.jar
httpcore-4.4.13.jar
j2objc-annotations-1.3.jar
jolokia-core-1.7.0.jar
json-20171018.jar
json-simple-1.1.1.jar
jsr305-3.0.2.jar
listenablefuture-9999.0-empty-to-avoid-conflict-with-guava.jar

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTUtil.java
amarkevich: fixed

pom.xml
amarkevich: ErrorProne configuration changed because produce `java.lang.NoClassDefFoundError: com/google/common/collect/ImmutableMap` during compilation

artemis-hawtio/artemis-console/pom.xml
gemmellr: Right...Guava isnt there, but is is meant to be there. Note its dependencies still being present.

The exclusion I mentioned earlier, in the repackaging the module build does, removed the version the original war included. You have removed the dependency that lead to its replacement in the repackaged war.

amarkevich: `checker-compat-qual` is not there
BTW artemis-console-2.29.0.war\WEB-INF\lib\ contains
checker-qual-3.12.0.jar
checker-qual-3.5.0.jar

Should be `checker-qual` excluded then?

pom.xml
amarkevich: https://github.com/apache/activemq-artemis/actions/runs/5487090252/jobs/9998059886?pr=4540

amarkevich:
```
activemq-artemis\artemis-quorum-ri>mvn dependency:tree -Derrorprone

[INFO] --- dependency:3.6.0:tree (default-cli) @ artemis-quorum-ri ---
[INFO] org.apache.activemq:artemis-quorum-ri:jar:2.30.0-SNAPSHOT
[INFO] +- org.apache.curator:curator-recipes:jar:5.2.0:compile
[INFO] |  \- org.apache.curator:curator-framework:jar:5.2.0:compile
[INFO] +- org.apache.curator:curator-client:jar:5.2.0:compile
[INFO] |  \- com.google.guava:guava:jar:27.0.1-jre:compile
[INFO] |     +- com.google.guava:failureaccess:jar:1.0.1:compile
[INFO] |     +- com.google.guava:listenablefuture:jar:9999.0-empty-to-avoid-conflict-with-guava:compile
[INFO] |     +- org.checkerframework:checker-qual:jar:2.5.2:compile
[INFO] |     +- com.google.j2objc:j2objc-annotations:jar:1.1:compile
[INFO] |     \- org.codehaus.mojo:animal-sniffer-annotations:jar:1.17:compile
[INFO] +- org.apache.zookeeper:zookeeper:jar:3.6.3:compile
[INFO] |  +- org.apache.zookeeper:zookeeper-jute:jar:3.6.3:compile
[INFO] |  \- io.netty:netty-handler:jar:4.1.94.Final:compile
[INFO] |     +- io.netty:netty-common:jar:4.1.94.Final:compile
[INFO] |     +- io.netty:netty-resolver:jar:4.1.94.Final:compile
[INFO] |     +- io.netty:netty-transport-native-unix-common:jar:4.1.94.Final:compile
[INFO] |     \- io.netty:netty-codec:jar:4.1.94.Final:compile
[INFO] +- org.apache.curator:curator-test:jar:5.2.0:test
[INFO] |  +- io.dropwizard.metrics:metrics-core:jar:3.2.5:test
[INFO] |  \- org.xerial.snappy:snappy-java:jar:1.1.7:test
[INFO] +- org.apache.activemq:artemis-quorum-api:jar:2.30.0-SNAPSHOT:compile
[INFO] +- org.slf4j:slf4j-api:jar:1.7.36:compile
[INFO] +- org.apache.logging.log4j:log4j-slf4j-impl:jar:2.20.0:test
[INFO] |  +- org.apache.logging.log4j:log4j-api:jar:2.20.0:test
[INFO] |  \- org.apache.logging.log4j:log4j-core:jar:2.20.0:test
[INFO] +- org.apache.activemq:artemis-commons:jar:2.30.0-SNAPSHOT:compile
[INFO] |  +- io.netty:netty-buffer:jar:4.1.94.Final:compile
[INFO] |  +- io.netty:netty-transport:jar:4.1.94.Final:compile
[INFO] |  \- commons-beanutils:commons-beanutils:jar:1.9.4:compile
[INFO] |     +- commons-logging:commons-logging:jar:1.2:compile
[INFO] |     \- commons-collections:commons-collections:jar:3.2.2:compile
[INFO] +- junit:junit:jar:4.13.2:test
[INFO] |  \- org.hamcrest:hamcrest-core:jar:1.3:test
[INFO] +- org.apache.activemq:artemis-unit-test-support:jar:2.30.0-SNAPSHOT:test
[INFO] +- org.hamcrest:hamcrest:jar:2.1:test
[INFO] \- com.google.errorprone:error_prone_core:jar:2.20.0:provided
[INFO]    +- com.google.errorprone:error_prone_annotation:jar:2.20.0:provided
[INFO]    +- com.google.errorprone:error_prone_type_annotations:jar:2.20.0:provided
[INFO]    +- com.google.errorprone:error_prone_check_api:jar:2.20.0:provided
[INFO]    |  +- io.github.java-diff-utils:java-diff-utils:jar:4.0:provided
[INFO]    |  |  \- org.eclipse.jgit:org.eclipse.jgit:jar:4.4.1.201607150455-r:provided
[INFO]    |  +- com.github.kevinstern:software-and-algorithms:jar:1.0:provided
[INFO]    |  +- com.github.ben-manes.caffeine:caffeine:jar:3.1.6:provided
[INFO]    |  \- com.google.inject:guice:jar:5.1.0:provided
[INFO]    |     \- aopalliance:aopalliance:jar:1.0:provided
[INFO]    +- org.pcollections:pcollections:jar:3.1.4:provided
[INFO]    +- com.google.auto:auto-common:jar:1.2.1:provided
[INFO]    +- com.google.code.findbugs:jsr305:jar:3.0.2:compile
[INFO]    +- io.github.eisop:dataflow-errorprone:jar:3.34.0-eisop1:provided
[INFO]    +- com.google.auto.value:auto-value-annotations:jar:1.9:provided
[INFO]    +- com.google.errorprone:error_prone_annotations:jar:2.20.0:compile
[INFO]    +- com.google.protobuf:protobuf-java:jar:3.19.6:provided
[INFO]    +- com.google.auto.service:auto-service-annotations:jar:1.0.1:provided
[INFO]    \- javax.inject:javax.inject:jar:1:provided
```

clebertsuconic:
All the arguments seems to indicate we should do the replacement.

However this is still failing to build. I have a pipeline where I execute the whole test suite: 

"[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project artemis-quorum-ri: Compilation failure
[ERROR] An unknown compilation problem occurred
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :artemis-quorum-ri
[Pipeline] }"


This is also translated as errors on the build here.

to test this locally you could do:


```
mvn clean; mvn install -DskipTests=true -Pdev -Derrorprone

```

amarkevich:
> still failing to build

without the latest commit build is successful

artemis-hawtio/artemis-console/pom.xml
gemmellr: The checker-compat-qual is presumably not there because it was excluded, per the comment in your removal. Having 2 checker-qual versions would seem like an oversight though, not noticed during an update.

gemmellr:
It may completes but as per earlier comments it isn't really 'successfull'; the compiled code then doesnt actually work because all the code wasnt generated, since the change stopped the logging annotation processor from running at all, meaning there are missing classes that stop various modules (client, broker, etc) from working.

pom.xml
gemmellr: So that module is now using Guava 27, which presumably doesn't have the thing needed by ErrorProne, which normally uses a newer version.

Maven is picking the older version during resolution, probably because it is earlier on the dependency set than the ErrorProne dep is (it will be last due to being in a profile). Its doing that instead of just using the specified dependencyManagement guava version that was removed from the build. Restoring that may fix things. Alternative perhaps there is a newer curator-client version that uses a recent guava.

artemis-hawtio/artemis-console/pom.xml
amarkevich: same for now - no `checker-compat-qual` in war

pom.xml
amarkevich: still I can't reproduce error using `annotationProcessorPaths` approach

amarkevich:
> stopped the logging annotation processor

let me check this regression

clebertsuconic:
still not building:

[INFO] Compiling 6 source files with javac [debug target 11] to target/classes
compiler message file broken: key=compiler.misc.msg.bug arguments=11.0.8, {1}, {2}, {3}, {4}, {5}, {6}, {7}
java.lang.NoSuchMethodError: 'com.google.common.collect.ImmutableMap com.google.common.collect.ImmutableMap$Builder.buildOrThrow()'
	at com.google.errorprone.scanner.ScannerSupplier.defaultSeverities(ScannerSupplier.java:67)
	at com.google.errorprone.scanner.ScannerSupplier.fromBugCheckerInfos(ScannerSupplier.java:88)
	at com.google.errorprone.scanner.BuiltInCheckerSuppliers.allChecks(BuiltInCheckerSuppliers.java:600)
	at com.google.errorprone.scanner.BuiltInCheckerSuppliers.defaultChecks(BuiltInCheckerSuppliers.java:608)
	at com.google.errorprone.ErrorProneJavacPlugin.init(ErrorProneJavacPlugin.java:35)
	at jdk.compiler/com.sun.tools.javac.api.BasicJavacTask.initPlugins(BasicJavacTask.java:215)
	at jdk.compiler/com.sun.tools.javac.api.JavacTaskImpl.prepareCompiler(JavacTaskImpl.java:199)

amarkevich:
> still not building

yep, going to do https://github.com/apache/activemq-artemis/pull/4545 first

pom.xml
gemmellr: I'd guess because it is doing a completely different thing there, with the ErrorProne deps defined earlier off-classpath, likely meaning its transitive guava dep version gets picked up, before starting to look at the compile deps on the classpath...rather than the reverse in the profile case. Which is fine if you dont remove the dependencyManagement entries ensuring the desired versions are chosen.

amarkevich:
> want the guava dependencyManagement restored

for the moment - yes, keep `annotationProcessorPaths` for further change

gemmellr:
> > want the guava dependencyManagement restored
> 
> for the moment - yes, keep `annotationProcessorPaths` for further change

Regardless of the annotationProcessorPaths situation. Guava will still be getting used in the project after this change, no matter which way the ErrorProne etc annotation processors are being configured (or even if we removed ErrorProne entirely).

amarkevich:
> Guava will still be getting used in the project

directly or as transitive dependency?

gemmellr:
Both (once you restore the console bit you have removed)

artemis-hawtio/artemis-console/pom.xml
gemmellr: Right, the original war doesnt have it (but does have guava), so it was also excluded from the repackaged war to match. The original war has checker-qual-3.5.0.jar, which should presumably have been excluded when unpacking it, due to the newer guava bringing in checker-qual-3.12.0.jar

gemmellr:
FWIW I dont see #4545 as a pre-requisite to this PR. It may get rid of the build error, in a different way, but it doesnt change the fixes/restorations still needing made here, which would also get rid of the build problem even without the other PR.

amarkevich:
> pre-requisite to this PR

its not a blocker but allow to implement current changes easy way

gemmellr:
I raised a new PR, #4584, to address the various trivial commented build etc issues on this PR which were preventing this progressing. Unfortunately it turns out this PR was still failing a test (MQTTSecurityManagerTest.testSecurityManagerModifyClientIDAndStealConnection) in its current state, and so the new PR is also and also cant be merged without more work first to identify+fix that issue.

PR #4541, opened 2023-07-06
ARTEMIS-4350 use consistent delimiters when constructing mqtt subscription queues

erwindon:
see https://issues.apache.org/jira/browse/ARTEMIS-4350

### DIFF
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSubscriptionManager.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSubscriptionManager.java
@@ -66,11 +66,14 @@ public class MQTTSubscriptionManager {
 
    private final char anyWords;
 
+   private final char delimiter;
+
    public MQTTSubscriptionManager(MQTTSession session) {
       this.session = session;
 
       singleWord = session.getServer().getConfiguration().getWildcardConfiguration().getSingleWord();
       anyWords = session.getServer().getConfiguration().getWildcardConfiguration().getAnyWords();
+      delimiter = session.getServer().getConfiguration().getWildcardConfiguration().getDelimiter();
 
       consumers = new ConcurrentHashMap<>();
       consumerQoSLevels = new ConcurrentHashMap<>();
@@ -301,9 +304,9 @@ public class MQTTSubscriptionManager {
          int slashIndex = topic.indexOf(SLASH) + 1;
          String sharedSubscriptionName = topic.substring(slashIndex, topic.indexOf(SLASH, slashIndex));
          String parsedTopicName = topic.substring(topic.indexOf(SLASH, slashIndex) + 1);
-         return new SimpleString(sharedSubscriptionName).concat(".").concat(parsedTopicName);
+         return new SimpleString(sharedSubscriptionName).concat(delimiter).concat(parsedTopicName);
       } else {
-         return new SimpleString(session.getState().getClientId()).concat(".").concat(topic);
+         return new SimpleString(session.getState().getClientId()).concat(delimiter).concat(topic);
       }
    }
 


clebertsuconic:
@jbertram can you look into this? 

@jbertram , @erwindon what about compatibility with previous versions? usually compatibility is a deterrent for things like this.

clebertsuconic:
@erwindon checkstyle issues ^^

but we should spend some time on the compatibility

jbertram:
I do believe this will break compatibility for users who upgrade and also have existing MQTT subscriptions and a non-default wildcard configuration (which is fairly common among folks who use the broker mainly for MQTT).

Furthermore, I'm not entirely sure this is actually more consistent overall. Using the configured delimiter instead of `.` is certainly more consistent with MQTT syntax, but the name of the queue is an internal implementation detail and `.` characters are used in other places for queue names so now this name is inconsistent with that.

I think this should probably just stay the way it is.

gemmellr:
I wouldn't change it in this way either. It is a breaking change as Justin says, very questionable for whats only marked as a 'minor improvement', and as the ClientID and address are ultimately distinct things and the wildcard deliminator is about the address its also not clear to me either that this would really be 'more consistent'.

erwindon:
> checkstyle issues
ah, sorry, my local build did not croak on that. adjusted+pushed.

And the rest of my comments have become obsolete after the recommendations from @jbertram and @gemmellr.

clebertsuconic:
@erwindon what's the point of using the same subscription between MQTT and AMQP anyways? each client should have its own subscription.


if you really do for any reason I can't understand, you should use FQQN and bypass the subscription altogether.

erwindon:
@clebertsuconic
> what's the point of using the same subscription between MQTT and AMQP anyways[...]

but that was not the intention...
I was investigating a scenario where MQTT messages were published using a MQTT connection and then consumed as AMQP messages using an AMQP connection. Obviously with Artemis in the middle. But I goofed on the configuration of the delimiter, so the defaults still applied. The mistake was that I had placed the whole {{wildcard-addresses}} block just after the {{core}} section in the config.
the MQTT connection published on topic {{aaa/bbb/ccc}} but Artemis converted that to address {{aaa.bbb.ccc}} inside the broker.
but for AMQP, no conversion takes place and Artemis uses the address {{aaa/bbb/ccc}} as given by the consumer.
this mismatch prevented messages from reaching their destination application.
this problem was quickly solved by your team's comments ("no problem here!") and myself verifying the config one more time, and then fixing that.

the cosmetic issue that I tried to solve in _this_ PR was just something that I found while investigating how addresses were handled by the MQTT protocol handler to solve the above problem. This PR was about the generated queue names for the MQTT subscriptions only that I found to be slightly inconsistent.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSubscriptionManager.java
```diff
@@ -301,9 +304,9 @@ private SimpleString getQueueNameForTopic(String topic) {
          int slashIndex = topic.indexOf(SLASH) + 1;
          String sharedSubscriptionName = topic.substring(slashIndex, topic.indexOf(SLASH, slashIndex));
          String parsedTopicName = topic.substring(topic.indexOf(SLASH, slashIndex) + 1);
-         return new SimpleString(sharedSubscriptionName).concat(".").concat(parsedTopicName);
+         return new SimpleString(sharedSubscriptionName).concat(delimiter).concat(parsedTopicName);
       } else {
-         return new SimpleString(session.getState().getClientId()).concat(".").concat(topic);
+         return new SimpleString(session.getState().getClientId()).concat(delimiter).concat(topic);
```
clebertsuconic, line 309: if you could add some testing? a simple one is fine?

Also.. I would update versions.md, in case someone is updating the delimiter, with the fact that MQTT subscription now are using the configured delimiter as it was missed before because of "ARTEMIS-4350".

if you don't want to update the versions.md that's fine and I would do it myself.. but if you could at least add a test?

clebertsuconic:
I don't understand.. was this closed?

clebertsuconic:
after I spoke with Justin, it's a common change to mess with wildcard configuration.. that will really create issues with previous versions.

PR #4544, opened 2023-07-07
ARTEMIS-4353 clean up Maven dependencies

jbertram:
This commit contains the following changes:
 - eliminate used, undeclared dependencies
 - eliminate unused, declared dependencies
 - fix scope for test dependencies
 - eliminate org.hamcrest completely as its use involved deprecated code as well as dependencies from multiple versions

### DIFF
--- a/artemis-junit/artemis-junit-5/pom.xml
+++ b/artemis-junit/artemis-junit-5/pom.xml
@@ -57,7 +57,7 @@
       </dependency>
       <dependency>
          <groupId>org.apache.activemq</groupId>
-         <artifactId>artemis-jms-server</artifactId>
+         <artifactId>artemis-core-client</artifactId>
          <version>${project.version}</version>
       </dependency>
       <dependency>
@@ -86,11 +86,6 @@
          <artifactId>artemis-unit-test-support</artifactId>
          <version>${project.version}</version>
          <scope>test</scope>
-      </dependency>
-
-      <dependency>
-         <groupId>jakarta.jms</groupId>
-         <artifactId>jakarta.jms-api</artifactId>
       </dependency>
 
       <dependency>
--- a/artemis-quorum-ri/pom.xml
+++ b/artemis-quorum-ri/pom.xml
@@ -34,6 +34,10 @@
    <dependencies>
       <dependency>
          <groupId>org.apache.curator</groupId>
+         <artifactId>curator-framework</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.curator</groupId>
          <artifactId>curator-recipes</artifactId>
       </dependency>
       <dependency>
@@ -43,6 +47,11 @@
       <dependency>
          <groupId>org.apache.zookeeper</groupId>
          <artifactId>zookeeper</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>org.apache.zookeeper</groupId>
+         <artifactId>zookeeper-jute</artifactId>
+         <scope>test</scope>
       </dependency>
       <dependency>
          <groupId>org.apache.curator</groupId>
@@ -71,11 +80,6 @@
          <artifactId>log4j-slf4j-impl</artifactId>
          <scope>test</scope>
       </dependency>
-      <dependency>
-         <groupId>org.apache.activemq</groupId>
-         <artifactId>artemis-commons</artifactId>
-         <version>${project.version}</version>
-      </dependency>
 
       <!-- tests -->
       <dependency>
@@ -87,12 +91,6 @@
          <groupId>org.apache.activemq</groupId>
          <artifactId>artemis-unit-test-support</artifactId>
          <version>${project.version}</version>
-         <scope>test</scope>
-      </dependency>
-      <dependency>
-         <groupId>org.hamcrest</groupId>
-         <artifactId>hamcrest</artifactId>
-         <version>${hamcrest.version}</version>
          <scope>test</scope>
       </dependency>
    </dependencies>
--- a/artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/DistributedLockTest.java
+++ b/artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/DistributedLockTest.java
@@ -31,9 +31,6 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
-import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.greaterThanOrEqualTo;
-
 public abstract class DistributedLockTest {
 
    private final ArrayList<AutoCloseable> closeables = new ArrayList<>();
@@ -222,7 +219,7 @@ public abstract class DistributedLockTest {
       final long timeoutSec = 1;
       Assert.assertFalse(manager.getDistributedLock("a").tryLock(timeoutSec, TimeUnit.SECONDS));
       final long elapsed = TimeUnit.NANOSECONDS.toSeconds(System.nanoTime() - start);
-      assertThat(elapsed, greaterThanOrEqualTo(timeoutSec));
+      Assert.assertTrue(elapsed + " less than timeout of " + timeoutSec + " seconds", elapsed >= timeoutSec);
    }
 
    @Test
--- a/pom.xml
+++ b/pom.xml
@@ -115,6 +115,7 @@
       <hdrhistogram.version>2.1.12</hdrhistogram.version>
       <curator.version>5.2.0</curator.version>
       <zookeeper.version>3.6.3</zookeeper.version>
+      <woodstox.version>4.4.0</woodstox.version>
 
       <!-- tracing -->
       <opentelemetry.version>1.24.0</opentelemetry.version>
@@ -128,6 +129,7 @@
       <qpid.jms.version>1.10.0</qpid.jms.version>
       <johnzon.version>1.2.16</johnzon.version>
       <hawtbuff.version>1.11</hawtbuff.version>
+      <hawtdispatch.version>1.22</hawtdispatch.version>
       <airline.version>2.9.0</airline.version>
       <jakarta.activation-api.version>1.2.2</jakarta.activation-api.version>
       <jakarta.annotation-api.version>1.3.5</jakarta.annotation-api.version>
@@ -158,6 +160,7 @@
       <paho.client.mqtt.version>1.2.5</paho.client.mqtt.version>
       <postgresql.version>42.4.3</postgresql.version>
       <testcontainers.version>1.17.6</testcontainers.version>
+      <selenium.version>4.8.3</selenium.version>
       <exec-maven-plugin.version>3.0.0</exec-maven-plugin.version>
       <apache.httpcore.version>4.4.13</apache.httpcore.version>
       <apache.httpclient.version>4.5.13</apache.httpclient.version>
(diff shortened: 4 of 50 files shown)


jbertram:
@gemmellr, I know you've worked on the `artemis-server` `test-jar` in the past. I added that dependency to a few modules since `mvn dependency:analyze` was complaining about it. If you think those should be left out let me know.

More generally, I wonder if we shouldn't move all the test support classes (e.g. `org.apache.activemq.artemis.tests.util.ActiveMQTestBase`) into `artemis-test-support` and then move all the necessary tests out of `artemis-server` and into `integration-tests` so we don't have a circular dependency. As it stands at the moment it's kind of a mess.

gemmellr:
I havent actually yet looked at this, but a few comments before I get to it:

- The commons-configuration2 upgrade mentioned should be probably done separately.
- Yep, I elected not to have the artemis-server test-jar dep declared everywhere, and isolated it to artemis-test-support to give one place to break it later. It sticks out transitively anyway, and its mainly used within the tests tree so it didnt seem like it was really any nicer listing it.
- Moving what are really more like integration test base classes such as ActiveMQTestBase into artemis-test-support (or some in artemis-unit-test-support as well) was partof my long term plan yes. Ideally 'actual unit tests' should be in the modules, but these current tests are often wiring so much together I'd question if they they really are even though thats where they live now (not sure 'kind of a mess' really does it justice hehe). 

One thing to consider is, moving any tests from e.g artemis-server into e.g integration-tests will reduce the set of tests run in the PR check. They could be retained in the PR run tests (if still seeming suitable) by including them in the integration-tests subset run by surefure when enabling the fast-tests profile (see the pom).

jbertram:
I backed out the commons-configuration2 change and sent #4548 instead. Also, I removed any additions of the `artemis-server` `test-jar`.

tests/integration-tests-isolated/pom.xml
```diff
@@ -302,93 +133,59 @@
          </exclusions>
       </dependency>
       <dependency>
-         <groupId>org.apache.directory.server</groupId>
-         <artifactId>apacheds-server-annotations</artifactId>
-         <version>${directory-version}</version>
+         <groupId>org.apache.directory.api</groupId>
+         <artifactId>api-ldap-model</artifactId>
+         <version>2.0.0.AM1</version>
```
gemmellr, line 138: We didnt have a direct dep on this before, so presumably it was transitive? I'd leave it that way rather than hard code a version; add a version prop or remove?

tests/integration-tests/pom.xml
```diff
@@ -50,56 +56,61 @@
...
       </dependency>
       <dependency>
          <groupId>org.apache.activemq</groupId>
          <artifactId>artemis-commons</artifactId>
          <version>${project.version}</version>
-      </dependency>
-      <dependency>
-         <groupId>org.apache.activemq</groupId>
-         <artifactId>artemis-spring-integration</artifactId>
-         <version>${project.version}</version>
```
gemmellr, line 77: If it isnt even being used here, and says in the module it isnt really needed...is time to nuke it?

tests/integration-tests/pom.xml
```diff
@@ -340,6 +402,19 @@
          </exclusions>
       </dependency>
 
+      <dependency>
+         <groupId>org.codehaus.woodstox</groupId>
+         <artifactId>woodstox-core-asl</artifactId>
+         <version>4.4.0</version>
```
gemmellr, line 408: While we are 'cleaning up deps' and moving the dep declaration, lets at add a prop for the version as done for some others.

gemmellr (commented):
There are a number of warnings from duplicate definitions that need fixed:

https://github.com/apache/activemq-artemis/actions/runs/5511789477/jobs/10047790267?pr=4544#step:5:9

> Warning:  
Warning:  Some problems were encountered while building the effective model for org.apache.activemq:artemis-log-annotation-processor:jar:2.30.0-SNAPSHOT
Warning:  'dependencyManagement.dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.seleniumhq.selenium:selenium-api:jar -> duplicate declaration of version ${selenium.version} @ org.apache.activemq:artemis-pom:2.30.0-SNAPSHOT, /home/runner/work/activemq-artemis/activemq-artemis/pom.xml, line 1057, column 22
Warning:  
Warning:  Some problems were encountered while building the effective model for org.apache.activemq.tests:soak-tests:jar:2.30.0-SNAPSHOT
Warning:  'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.apache.activemq.tests:artemis-test-support:jar -> duplicate declaration of version ${project.version} @ line 140, column 19
Warning:  
Warning:  Some problems were encountered while building the effective model for org.apache.activemq.tests:stress-tests:jar:2.30.0-SNAPSHOT
Warning:  'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.apache.activemq.tests:artemis-test-support:jar -> duplicate declaration of version ${project.version} @ line 126, column 19
Warning:  
Warning:  Some problems were encountered while building the effective model for org.apache.activemq.tests:smoke-tests:jar:2.30.0-SNAPSHOT
Warning:  'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.seleniumhq.selenium:selenium-api:jar -> duplicate declaration of version (?) @ line 171, column 19
Warning:  'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.apache.activemq.tests:artemis-test-support:jar -> duplicate declaration of version ${project.version} @ line 261, column 19
Warning:  
Warning:  Some problems were encountered while building the effective model for org.apache.activemq:artemis-pom:pom:2.30.0-SNAPSHOT
Warning:  'dependencyManagement.dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.seleniumhq.selenium:selenium-api:jar -> duplicate declaration of version ${selenium.version} @ line 1057, column 22
Warning:  
Warning:  It is highly recommended to fix these problems because they threaten the stability of your build.
Warning:  
Warning:  For this reason, future Maven versions might no longer support building such malformed projects.

artemis-quorum-ri/pom.xml
```diff
@@ -43,6 +47,12 @@
       <dependency>
          <groupId>org.apache.zookeeper</groupId>
          <artifactId>zookeeper</artifactId>
+         <scope>test</scope>
```
gemmellr, line 50: Also not sure about this one changing to test. The main purpose of this module is currently to use ZK. That it does it via the curator stuff doesnt really mean this is a test-only dep, whilst listing it like this really makes it seem like it is...however anything using this module will still get ZK since it isnt really. Maven doesnt have a way of saying "runtime + test" which is what this really is. I think I'd leave it at compile for clarity.

tests/integration-tests/pom.xml
jbertram: Yes. I'll nuke it from orbit.

artemis-junit/artemis-junit-5/pom.xml
```diff
@@ -57,12 +57,7 @@
       </dependency>
       <dependency>
          <groupId>org.apache.activemq</groupId>
-         <artifactId>artemis-jms-server</artifactId>
-         <version>${project.version}</version>
-      </dependency>
-      <dependency>
-         <groupId>org.apache.activemq</groupId>
-         <artifactId>artemis-jms-client</artifactId>
```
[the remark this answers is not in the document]
jbertram: The JUnit infrastructure should be using the core API exclusively so I don't think there's any reason to depend on the core JMS client.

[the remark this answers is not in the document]
gemmellr: I was 'ok' with this removal since the junit 5 stuff is actually 'core centric' (which again, I dont actually think makes any sense, since we would recommend the JMS client to most, and most I've seen using this want it to test with the JMS client, so they typically shouldn't be using the core API directly for anything).

Thats unlike the original junit 4 aimed stuff though, which has some very JMS-specific bits as well, that do use the JMS client, so I dont think it should be removed from the junit 4 stuff at this point since that has always supplied the JMS client, and some will expect it to.

I have however been forgetting about junit-commons module also existing, which is perhaps still supplying the jms client to _both_ the junit 4 and 5 modules (the latter should arguably be excluding it if so) ?

pom.xml
```diff
@@ -85,7 +85,7 @@
       <activemq-artemis-native-version>2.0.0</activemq-artemis-native-version>
       <karaf.version>4.3.3</karaf.version>
       <pax.exam.version>4.13.4</pax.exam.version>
-      <commons.config.version>2.8.0</commons.config.version>
+      <commons.config.version>2.9.0</commons.config.version>
```
gemmellr, line 88: Shouldnt be here, PR is needing a rebase.

tests/smoke-tests/pom.xml
```diff
@@ -97,30 +139,58 @@
...
+         <groupId>org.testcontainers</groupId>
+         <artifactId>testcontainers</artifactId>
+         <scope>test</scope>
+      </dependency>
+
+      <dependency>
+         <groupId>com.github.docker-java</groupId>
+         <artifactId>docker-java-transport-zerodep</artifactId>
+         <version>3.2.13</version>
+         <scope>test</scope>
```
gemmellr, line 193: Hard coded version. Doesnt look present before so was presumably transitive, lets either leave it that way or add a property.

tests/soak-tests/pom.xml
```diff
@@ -63,8 +116,20 @@
          <type>test-jar</type>
       </dependency>
       <dependency>
-         <groupId>junit</groupId>
-         <artifactId>junit</artifactId>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>activemq-client</artifactId>
+         <version>${activemq5-version}</version>
```
gemmellr, line 121: This is a managed dependency, should be able to drop the version element. (The exclusions are managed also, but can be nice for clarity)

artemis-junit/artemis-junit-5/pom.xml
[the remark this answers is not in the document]
jbertram: I see your point about the module providing the client to the depender to make testing easier. I think that's fair.

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/DistributedLockTest.java
```diff
@@ -222,7 +219,7 @@ public void timedTryLockFailAfterTimeout() throws ExecutionException, Interrupte
       final long timeoutSec = 1;
       Assert.assertFalse(manager.getDistributedLock("a").tryLock(timeoutSec, TimeUnit.SECONDS));
       final long elapsed = TimeUnit.NANOSECONDS.toSeconds(System.nanoTime() - start);
-      assertThat(elapsed, greaterThanOrEqualTo(timeoutSec));
+      Assert.assertTrue(elapsed >= timeoutSec);
```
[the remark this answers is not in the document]
jbertram: I agree that Hamcrest provides better messages when assertions fail, although there are huge number of vague uses of `assertTrue` throughout the code-base. I've added messages where I thought it made sense or changed `assertTrue` to `assertEquals` to get an automated message which is more useful.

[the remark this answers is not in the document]
gemmellr: Yeah, I wasnt saying you needed to do them all (though it is certainly nicer to so you dont need to open the code to read the assertion when a test run fails).

Its more the things where if it failed, you would only know partial detail...e.g if an assert was checking for a specific class and failed due to an unexpected class (but which one was it actually?), or a value wasnt between expected X and Y (but what was it?), etc. The ones where the assertion failure doesnt implicitly give you enough context to know what did happen, even though that might greatly help you figure out why what you expected didnt happen.

pom.xml
```diff
@@ -85,7 +85,7 @@
       <activemq-artemis-native-version>2.0.0</activemq-artemis-native-version>
       <karaf.version>4.3.3</karaf.version>
       <pax.exam.version>4.13.4</pax.exam.version>
-      <commons.config.version>2.9.0</commons.config.version>
+      <commons.config.version>2.8.0</commons.config.version>
```
gemmellr, line 88: Think the rebase went awry, this is now downgrading rather than not touching the version.

PR #4545, opened 2023-07-10
ARTEMIS-4355 Update Curator to 5.5.0; Zookeeper 3.8.2

amarkevich:


jbertram:
Why were the test changes necessary?

amarkevich:
There were compilation errors because `getQuorumPeer` removed

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/zookeeper/CuratorDistributedLockTest.java
```diff
@@ -84,8 +82,6 @@ public void setupEnv() throws Throwable {
       }
       testingServer = new TestingCluster(clusterSpecs);
       testingServer.start();
-      // start waits for quorumPeer!=null but not that it has started...
-      Wait.waitFor(this::ensembleHasLeader);
```
gemmellr, line 88: Did this actually change on the Curator side, or did you just remove the 'wait for start' as the getQuorumPeer method went away?

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/zookeeper/CuratorDistributedLockTest.java
```diff
@@ -337,21 +333,8 @@ public void beNotifiedOfAlreadyUnavailableManagerAfterAddingListener() throws Ex
...
-         return true;
-      }
-      long leaderId = server.getQuorumPeer().getLeaderId();
-      long id = server.getQuorumPeer().getId();
-      return id == leaderId;
-   }
-
    private void stopMajorityNotLeaderNodes(boolean fromLast) throws Exception {
-      List<TestingZooKeeperServer> followers = testingServer.getServers().stream().filter(Predicates.not(CuratorDistributedLockTest::isLeader)).collect(Collectors.toList());
+      List<TestingZooKeeperServer> followers = testingServer.getServers();
```
gemmellr, line 337: Is this now just stopping 'a majority of nodes' rather than the original 'majority of non-leader nodes'?

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/quorum/ZookeeperPluggableQuorumPeerTest.java
```diff
@@ -48,15 +47,7 @@ public ZookeeperPluggableQuorumPeerTest() {
...
       backup = new BrokerControl("primary-peer-b", JMX_PORT_BACKUP, "zkReplicationPrimaryPeerB", BACKUP_PORT_OFFSET);
-      brokers = new LinkedList(Arrays.asList(primary, backup));
-   }
-
-   @Test
-   @Override
-   public void testBackupFailoverAndPrimaryFailback() throws Exception {
-      // peers don't request fail back by default
-      // just wait for setup to avoid partial stop of zk via fast tear down with async setup
-      Wait.waitFor(this::ensembleHasLeader);
```
gemmellr, line 59: Again, did something change on the Curator side, or is this now potentially doing the very 'partial stop of zk via fast tear down with async setup' this was specifically there to avoid?

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/zookeeper/CuratorDistributedLockTest.java
amarkevich: `ensembleHasLeader` used 'getQuorumPeer' inside

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/zookeeper/CuratorDistributedLockTest.java
amarkevich: renamed

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/zookeeper/CuratorDistributedLockTest.java
gemmellr: I realise, which is why I mentioned it in my question. What I wondered is whether the reason it did so, to wait for _actual startup_, has changed on the curator side. If it didn't then you have just removed a step that I expect was there for a reason.

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/zookeeper/CuratorDistributedLockTest.java
gemmellr: I was not really pointing out the name wasnt descriptive, but that you appeared to have changed _what the method is trying to do_. As that could be a difference that alters expected behaviour for tests using it, it isnt immediately obvious as being desirable.

amarkevich:
`[INFO] Tests run: 14, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 188.484 s - in org.apache.activemq.artemis.tests.smoke.quorum.ZookeeperPluggableQuorumPeerTest
`

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/ssl/CoreClientOverOneWaySSLTest.java
```diff
@@ -85,9 +85,9 @@ public static Collection getParameters() {
          {"SunJCE", "JCEKS", false, false},
          {"SUN", "JKS", false, false},
          {"SunJSSE", "PKCS12", false, false},
-         {"JCEKS", null, true, false}, // for compatibility with old keyStoreProvider
-         {"JKS", null, true, false},   // for compatibility with old keyStoreProvider
-         {"PKCS12", null, true, false} // for compatibility with old keyStoreProvider
+         {"JCEKS", null, false, false}, // for compatibility with old keyStoreProvider
+         {"JKS", null, false, false},   // for compatibility with old keyStoreProvider
+         {"PKCS12", null, false, false} // for compatibility with old keyStoreProvider
```
jbertram, line 90: I don't understand why these changes were necessary. This eliminates the tests for the warning log message.

amarkevich: To pass tests for this PR - otherwise quorum tests are not running. Expected changes in first commit only

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/zookeeper/CuratorDistributedLockTest.java
amarkevich: reverted

artemis-quorum-ri/src/test/java/org/apache/activemq/artemis/quorum/zookeeper/CuratorDistributedLockTest.java
amarkevich: Since leader api was removed there is no significant choice which instance stopped

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/quorum/ZookeeperPluggableQuorumPeerTest.java
amarkevich: reverted with different check. In this PR quorum tests are enabled to show that its passed

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/common/SmokeTestBase.java
```diff
@@ -158,24 +156,6 @@ protected static final void recreateBrokerDirectory(final String homeInstance) {
...
-            System.out.println("awaiting server " + uri + " start at ");
-            Thread.sleep(500);
-            continue;
-         }
-         return true;
-      }
-
-      return false;
-   }
-
```
jbertram, line 178: Is this change actually related to the ZooKeep & Curator upgrade? It looks like an unrelated removal of an unused method.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/ssl/CoreClientOverOneWaySSLTest.java
amarkevich: this test started to fail because Zookeeper 3.8.x uses logback. rollback to 3.7.1

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/common/SmokeTestBase.java
amarkevich: looks like actual logic is moved to ServerUtil and this method confuses when looking for corresponding check

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/ssl/CoreClientOverOneWaySSLTest.java
amarkevich: I'm not sure that Zookeeper will work after such changes

jbertram: I took a look at the ZooKeeper code. As far as I can tell it's using SLF4J as the API (just like the broker), and Logback is the implementation. I think you'd be fine to exclude Logback via the `pom.xml`.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/quorum/ZookeeperPluggableQuorumPeerTest.java
```diff
@@ -48,15 +47,15 @@ public ZookeeperPluggableQuorumPeerTest() {
...
+      brokers = Arrays.asList(primary, backup);
    }
 
    @Test
    @Override
    public void testBackupFailoverAndPrimaryFailback() throws Exception {
       // peers don't request fail back by default
       // just wait for setup to avoid partial stop of zk via fast tear down with async setup
-      Wait.waitFor(this::ensembleHasLeader);
+      Assert.assertTrue(awaitAsyncSetupCompleted(TimeUnit.SECONDS.toMillis(30L), TimeUnit.MILLISECONDS));
```
gemmellr, line 58: Being curious how you were able to remove the methods in the other class without affecting anything else, I had a look, and it appears like this different check will in fact do nothing really, as the called method just returns true. So its asserting true on something that just returns true.

The fact awaitAsyncSetupCompleted simply returns true, is interesting...though the effect of this change is still to no wait for startup, which it did before.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/common/SmokeTestBase.java
gemmellr: Thats fine, but Justin's correct point is it shouldn't be in the same commit as it isnt related.

amarkevich: moved to additional PR

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/ssl/CoreClientOverOneWaySSLTest.java
amarkevich: done

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/quorum/ZookeeperPluggableQuorumPeerTest.java
amarkevich: Should I remove this dummy check everywhere?

gemmellr:
FWIW I ran the earlier version of this (cd5887b0e8) through the full test suite in another CI env and it rather fell over in a heap, failing various tests (even though in various cases they seem to have done what was expected), and timing out and being terminated overall. I didnt bother to re-run it (in case it was a sporadic env issue) since the PR had changed singnificantly by that point, and has now again.

I ran c64e2e31a6 through the full suite today though and it seemed ok, even if I do still question some of the actual behaviour changes being made.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/quorum/ZookeeperPluggableQuorumPeerTest.java
amarkevich: just @Ignore'd

gemmellr:
Still not that sure of some of the changes, but they dont seem to break the tests, and it is past time these were updated, so...in they go.

PR #4546, opened 2023-07-10
ARTEMIS-4354 Allow the recovery XAResource to be osing and reconnceting

ehsavoie:
the underlying CLientSession.

### DIFF
--- a/artemis-service-extensions/src/main/java/org/apache/activemq/artemis/service/extensions/xa/recovery/ActiveMQXAResourceWrapper.java
+++ b/artemis-service-extensions/src/main/java/org/apache/activemq/artemis/service/extensions/xa/recovery/ActiveMQXAResourceWrapper.java
@@ -67,6 +67,13 @@ public class ActiveMQXAResourceWrapper implements XAResource, SessionFailureList
       }
    }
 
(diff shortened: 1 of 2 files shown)


jbertram:
Couple of typos in the commit message.

artemis-service-extensions/src/main/java/org/apache/activemq/artemis/service/extensions/xa/recovery/ActiveMQXAResourceWrapper.java
```diff
@@ -67,6 +67,15 @@ public ActiveMQXAResourceWrapper(XARecoveryConfig... xaRecoveryConfigs) {
       }
    }
 
+   public void updateRecoveryConfig(XARecoveryConfig... xaRecoveryConfigs) {
+      synchronized (ActiveMQXAResourceWrapper.lock) {
+         if (delegate != null) {
+            delegate = null;
```
clebertsuconic, line 73: shouldn't close be called here instead? the previous resource will probably leak?


isn't there a way to write a test for this?

clebertsuconic: of course you could have the close called with a try .. catch (Exception ignored) {}

PR #4550, opened 2023-07-11
ARTEMIS-4331: update to jgroups 5.2.16

gemmellr:


jbertram:
Where did you find this secret sauce?
```
$Import-Package=org.jboss.byteman.*;resolution:=&quot;optional&quot;
```
I figured something like this must be possible, but I couldn't find it.

gemmellr:
Looked at the documentation: https://karaf.apache.org/manual/latest/#_wrap_deployer

erwindon:
after @jbertram 's attempt, I spent at least an hour investigating the same thing. I saw that page too, but I did not spot any clues there.
thx @gemmellr + @jbertram for solving the riddle!

PR #4551, opened 2023-07-11
ARTEMIS-4360 Make compatibility-tests opt in with -Pcompatibility-tests

clebertsuconic:


tests/leak-tests/pom.xml
```diff
@@ -29,7 +29,7 @@
 
    <properties>
       <activemq.basedir>${project.basedir}/../..</activemq.basedir>
-      <check-leak-version>0.8</check-leak-version>
+      <check-leak-version>0.11</check-leak-version>
```
gemmellr, line 32: unrelated change, separate commit

tests/compatibility-tests/pom.xml
```diff
@@ -259,425 +259,431 @@
...
-         </plugin>
-      </plugins>
    </build>
+   <profiles>
+      <profile>
+         <id>compatibility-tests</id>
+         <build>
+            <plugins>
+            <plugin>
+               <groupId>org.apache.activemq</groupId>
```
gemmellr, line 269: Some alignment issues here.

tests/compatibility-tests/pom.xml
```diff
@@ -259,425 +259,431 @@
...
+                     <id>generate-depends-file</id>
+                     <goals>
+                        <goal>generate-depends-file</goal>
+                     </goals>
+                  </execution>
+               </executions>
+            </plugin>
+           </plugins>
+         </build>
+      </profile>
```
gemmellr, line 685: Also [different?) alignment issues here.

tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/base/ClasspathBase.java
```diff
@@ -181,7 +179,14 @@ protected ClassLoader getClasspath(String name, boolean forceNew) throws Excepti
          classPathValue = buffer.toString();
       }
 
-      Assert.assertTrue("Cannot run compatibility tests, no classpath found on ./target/" + name + ".cp", classPathValue != null && !classPathValue.trim().equals(""));
+      boolean foundClassPath = classPathValue != null && !classPathValue.trim().equals("");
+
+      if (!foundClassPath) {
+         logger.info("Rebuild project with -Pcompatibility-tests. Cannot find classpath definition for {}", name);
+         Assume.fail("Rebuild project with -Pcompatibility-tests.");
```
gemmellr, line 186: The build is failing here, fail isnt an Assume method but an Assert method.

clebertsuconic:
thanks @gemmellr

tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/base/ClasspathBase.java
```diff
@@ -181,7 +179,15 @@ protected ClassLoader getClasspath(String name, boolean forceNew) throws Excepti
...
-      Assert.assertTrue("Cannot run compatibility tests, no classpath found on ./target/" + name + ".cp", classPathValue != null && !classPathValue.trim().equals(""));
+      boolean foundClassPath = classPathValue != null && !classPathValue.trim().equals("");
+
+      if (!foundClassPath) {
+         logger.info("Rebuild project with -Pcompatibility-tests. Cannot find classpath definition for {}", name);
+      }
+
+      Assume.assumeTrue("Rebuild project with -Pcompatibility-tests.", foundClassPath);
+
+      Assume.assumeTrue("Rebuild project with -Pcompatibility-tests to be able to run this test. No classpath found on ./target/" + name + ".cp", foundClassPath);
```
clebertsuconic, line 190: please @clebertsuconic  remove this line!

gemmellr: I was going to ask earlier about the Assume's but didnt as its Draft :)

clebertsuconic: @gemmellr I made it a draft to send fail commits and make sure tests would fail.

gemmellr: Ok. Then I'll actually ask. I was confused why you changed the existing assertion check to an Assume, which will skip rather than fail the test? Originally it used an assertTrue that would fail if tripped, and then you had tried to add an explicit fail here at first, but then switched it to the [semi-duplicated] assumeTrue instead.

clebertsuconic: @gemmellr I used to use assume before I switched to files. The idea was that if the variable wasn't set, the test for that particular class path would just be ignored.

I thought it would still be valid to have the tests running at least on the snapshots as that would at least validate the test itself.. but I can just disable it altogether if the profile is not run and keep as before. which I will do.

clebertsuconic: @gemmellr PR has what I intend to merge now (apart from having to remove the DO-NOT-MERGE commit as I'm validating if the PR can handle the failure).

clebertsuconic: tests are not running if I just do:

mvn -Pcompatibility-tests from ./tests/compatibility-tests


I'm fixing it.

pom.xml
```diff
@@ -1287,6 +1286,12 @@
             <skipLeakTests>false</skipLeakTests>
          </properties>
       </profile>
+      <profile>
+         <id>compatibility-tests</id>
```
gemmellr, line 1290: Would prefer this profile was in the compatibility-tests module. 

Well, it is already there...would prefer the property was redefined in the module.

pom.xml
```diff
@@ -1276,7 +1276,6 @@
             <skipJoramTests>false</skipJoramTests>
             <skipIntegrationTests>false</skipIntegrationTests>
             <skipIsolatedIntegrationTests>false</skipIsolatedIntegrationTests>
-            <skipCompatibilityTests>false</skipCompatibilityTests>
```
gemmellr, line 1279: You left it set not-skipped/enabled in the fast-tests profile.

clebertsuconic: ouch! thanks!

pom.xml
clebertsuconic: @gemmellr  I thought about doing it the way you mentioned.. but it is how the other properties are defined.. so I kept the way it's currently done. I thought it would be easier to manage the other skips.

gemmellr: Those ones at least have the case of applying to lots of modules, and setting the prop is often the only effect of the profile.

This one is entirely specific to compatibility-tests, which already has the profile itself, which does more than just setting the prop, so it would seem nicer if it was all together in the module. (You also just had to make 2 fixup commits that were basically the result of splitting such prop updates).

PR #4552, opened 2023-07-12
ARTEMIS-4362 Produce log.warn when the system cannot depage because of pending acks

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3298,8 +3306,28 @@ private boolean needsDepage() {
...
             (maxReadMessages <= 0 || (queueMemorySize.getElements() + deliveringMetrics.getMessageCount()) < maxReadMessages);
+
+         if (!needsDepageResult) {
+            if (!pageFlowControlled && (maxReadBytes > 0 && deliveringMetrics.getPersistentSize() >= maxReadBytes || maxReadMessages > 0 && deliveringMetrics.getMessageCount() >= maxReadMessages)) {
+               if (System.currentTimeMillis() - pageFlowControlledLastLog > PAGE_FLOW_CONTROL_PRINT_INTERVAL) {
+                  pageFlowControlledLastLog = System.currentTimeMillis();
+                  ActiveMQServerLogger.LOGGER.warnPageFlowControl(String.valueOf(name), String.valueOf(address), deliveringMetrics.getMessageCount(), deliveringMetrics.getPersistentSize(), maxReadMessages, maxReadBytes);
+               }
+               if (logger.isDebugEnabled()) {
+                  logger.warn("Queue {}/Address {} is currently delivering {} messages that are holding {} bytes, with max sizes configured at addressSettings.maxPageReadMessages={} and addressSettings.maxPageReadBytes={}. System will not read more data from paging until acks are made. " +
```
clebertsuconic, line 3319: logger.debug!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
```diff
@@ -1584,4 +1584,7 @@ void slowConsumerDetected(String sessionID,
 
    @LogMessage(id = 224126, value = "Failure during protocol handshake on connection to {} from {}", level = LogMessage.Level.ERROR)
    void failureDuringProtocolHandshake(SocketAddress localAddress, SocketAddress remoteAddress, Throwable e);
+
+   @LogMessage(id = 224127, value = "Queue {}/Address {} is currently delivering {} messages that are holding {} bytes, with max reading sizes configured as addressSettings.maxPageReadMessages={} and addressSettings.maxPageReadBytes={}. System will not read more data from paging until acks are made. You can increase the address settings page-max-read attributes or make sure your consumers use lower batch sizes.", level = LogMessage.Level.WARN)
```
gtully, line 1588: thinking of the event:
dispatch is blocked, then the reason and what to do about it. so reorder the message a little.

possibly:
dispatch blocked. Address|Queue will not read more data from paging until dispatched|inflight|delivered messages are acked. X is currently delivering..... You can ...... etc

clebertsuconic: What's blocked here is reading from paging.  If you close a consumers those messages will be delivered.  But it won't read any more messages until these are a mesma.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -179,6 +179,14 @@ public class QueueImpl extends CriticalComponentImpl implements Queue {
 
    private volatile boolean queueDestroyed = false;
 
+   // Variable to control if we should print a flow controlled message or not.
+   // Once it was flow controlled, we will stop warning until it's cleared once again
+   private volatile boolean pageFlowControlled = false;
+
+   private volatile long pageFlowControlledLastLog = 0;
+
+   private static final long PAGE_FLOW_CONTROL_PRINT_INTERVAL = 60_000L;
```
gtully, line 188: I think this will need to be configurable.... possibly combining with what we do for flow control of a producer, when an address is blocked, and a new producer tries to send... it is analogous to a consumer on a blocked queue. Both are reaching configured limits that we expect to be met in normal operation, however for a new producer or consumer - the limits will influence expected behaviour, they cannot send or consume... and may revert to the logs to see why.

we could have flowControllLoggingPeriod - that can be tweaked to configure the interval and disable. it could be applicable to both producers and consumers.
Periodic will work, b/c there will be some correlation between new consumer being added and the period. Is it true that will always attempt a dispatch on adding a new consumer?

gtully: we did the same in classic a while back: https://issues.apache.org/jira/browse/AMQ-6854

but I don't think it is a warn, more like some info.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
gtully: it is only reading to dispatch, from a user perspective, dispatch is blocked. it is transient, but for a new consumer on a queue with pending messages, that is the disconnect that we want to fill with the logging.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
clebertsuconic: I was trying to avoid another configuration attribute in broker.xml.

the system will always try to depage when a new consumer is added.. yeah.

Reason I made it WARN was : first I was in doubt which level I should use.... as I talked to a few users they usually ignore anything bellow WARN and will only get notifications when WARN is used.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
clebertsuconic: What about this one:

"Message dispatch from paging is blocked. Address {}/Queue{} will not read any more messages from paging until pending messages are acknowledged. There are currently {} messages pending ({} bytes) with max reads at maxPageReadMessages({}) and maxPageReadBytes({}). Either increase reading attributes at the address-settings or change your consumers to acknowledge more often."

PR #4555, opened 2023-07-17
ARTEMIS-4366 Missing Mirrored ACKs with MULTICAST and subscriptions

clebertsuconic:


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/ProtonProtocolManager.java
```diff
@@ -125,11 +125,11 @@ public ProtonProtocolManager(ProtonProtocolManagerFactory factory, ActiveMQServe
       routingHandler = new AMQPRoutingHandler(server);
    }
 
-   public synchronized ReferenceNodeStore getReferenceIDSupplier() {
+   public synchronized ReferenceNodeStoreFactory getReferenceIDSupplier() {
```
gemmellr, line 128: The method name also seems off now.

gemmellr:
Its odd/confusing that the ReferenceNodeStoreFactory is still named to in most parameters and fields as 'referenceIDSupplier' or 'nodeStore', especially the latter with there being an actual seperate ReferenceNodeStore. Its mixing roles and field/param names up in a confusing manner. Feels like the behaviours should be split into separate concerns, as they largely were originally.

Alternatively many of the ID methods now on ReferenceNodeStoreFactory could perhaps just be made static helpers rather than being 'passed around on the side of a factory', since they only pull details from the message[reference] rather than the factory itself.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/ProtonProtocolManager.java
gemmellr: Think the only one that wouldnt go static is getDefaultNodeID() , which didnt originally exist as the caller already had the value (serverId).

Renaming would be good...the current naming just doesnt fit, makes it confusing to read.

PR #4556, opened 2023-07-17
ARTEMIS-4365 MQTT retain flag not set correctly

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/MQTTTest.java
```diff
@@ -2219,4 +2223,71 @@ public void testAutoDeleteRetainedQueue() throws Exception {
...
+      subscriber.setCallback(new DefaultMqtt3Callback() {
+         @Override
+         public void messageArrived(String topic, MqttMessage message) throws Exception {
+            if (message.isRetained()) {
+               latch.countDown();
+            }
+         }
+      });
+
+      assertTrue(latch.await(1000, TimeUnit.MILLISECONDS));
```
gemmellr, line 2288: I saw this test fail here on a CI run.

Adding a failure message would be good. Its annoying having to go and match up line numbers in remote repositories to see what the failure actually meant.

jbertram: I fixed a race condition between the `subsribe` and `setCallback`. I can't get it to fail anymore.

jbertram: I also added messages to the relevant assertions.

jbertram:
I fixed the failure coming from `MQTTInterceptorPropertiesTest.checkMessageProperties`. The failure was due to the new (correct) semantics for the `retain` flag.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTPublishManager.java
```diff
@@ -404,12 +406,15 @@ private boolean publishToClient(int messageId, ICoreMessage message, int deliver
...
       boolean redelivery = qos == 0 ? false : (deliveryCount > 1);
 
-      boolean isRetain = message.getBooleanProperty(MQTT_MESSAGE_RETAIN_KEY);
+      boolean isRetain = message.containsProperty(MQTT_MESSAGE_RETAIN_INITIAL_DISTRIBUTION_KEY);
       MqttProperties mqttProperties = getPublishProperties(message);
 
       if (session.getVersion() == MQTTVersion.MQTT_5) {
-         if (session.getState().getSubscription(message.getAddress()) != null && !session.getState().getSubscription(message.getAddress()).option().isRetainAsPublished()) {
-            isRetain = false;
+         if (message.getBooleanProperty(MQTT_MESSAGE_RETAIN_KEY) && !message.containsProperty(MQTT_MESSAGE_RETAIN_INITIAL_DISTRIBUTION_KEY)) {
```
gemmellr, line 413: Since we already know the presence result from the earlier check, and only need to do the rest if that was false, what about:

```suggestion
         if (!isRetain && message.getBooleanProperty(MQTT_MESSAGE_RETAIN_KEY)) {
```

jbertram: I considered doing that, but it didn't seem as clear as I wanted it to be at the time. Taking another look I think it's fine.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/MQTTInterceptorPropertiesTest.java
```diff
@@ -89,25 +83,25 @@ public void testCheckInterceptedMQTTMessageProperties() throws Exception {
...
+            e.printStackTrace();
          }
       });
       thread.start();
 
-      final MQTTClientProvider publishProvider = getMQTTClientProvider();
-      initializeConnection(publishProvider);
-      publishProvider.publish(addressQueue, msgText.getBytes(), AT_MOST_ONCE, retained);
-
       latch.await(10, TimeUnit.SECONDS);
```
clebertsuconic, line 105: I know it's not part of this change.. but you should Assert.assertTrue(latch.await...) here.

otherwise a failure from the Thread wouldn't be translated as an assertion on the main process.

PR #4560, opened 2023-07-19
NO-JIRA remove unused SmokeTestBase#waitForServerToStart

amarkevich:


clebertsuconic:
@amarkevich Thanks... I don't know when I removed the usage for that method.

PR #4561, opened 2023-07-20
NO-JIRA - ensure predictable order for accurate logging

gtully:


jbertram (approved):
It would be ideal to have a test to verify the fix. Otherwise the fix looks good to me.

PR #4565, opened 2023-07-24
ARTEMIS-4372 Implement Pico-cli and script auto-complete

clebertsuconic:


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/AutoCompletion.java
```diff
@@ -0,0 +1,49 @@
...
+@CommandLine.Command(name = "auto-complete", description = "Generates the auto complete helper file")
+public class AutoCompletion implements Runnable {
+
+   public AutoCompletion(CommandLine parent) {
+      this.parent = parent;
+   }
+
+   CommandLine parent;
+
+   @CommandLine.Parameters (description = "The script that will be used to invoke check-leak")
```
jbertram, line 34: Reference to "check-leak" here seems wrong.

clebertsuconic:
https://github.com/apache/activemq-artemis/assets/750514/2e094597-8172-44e6-9965-e88766c396e2

docs/user-manual/en/using-server.md
```diff
@@ -50,6 +50,30 @@ The following highlights some important folders on the distribution:
...
 
+## Artemis Shell
+
+You can activate the artemis shell by typing:
+
+```shell
+./artemis shell
+```
+
+This will start a JLine3 terminal with auto-completion:
```
gemmellr, line 61: Sentence fragment, seems to end mid:

---
Also, this drop-in about the shell usage also doesnt really fit with the rest of the existing information on the page below that all assumes you are **not** using it. It could do with some elaboration / description linking and contrasting the different usages so the two routes are clear, and some new user actually following this page linearly isnt met with a sequence that essentially says 'do foo' followed by 'do this incompatible thing that requires you didnt do the earlier thing we already said'.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/Artemis.java
```diff
@@ -194,62 +186,102 @@ public static Object execute(boolean inputEnabled, boolean useSystemOut, File ar
...
 
-      if (action.isVerbose()) {
-         context.out.print("Executing " + action.getClass().getName() + " ");
-         for (String arg : args) {
-            context.out.print(arg + " ");
+         return action.execute(context);
+      } else {
+         if (userObject instanceof Runnable) {
+            ((Runnable) userObject).run();
+         }
```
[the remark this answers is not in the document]
clebertsuconic: we shouldn't really add anything to do the CommandLine that's not a Runnable or an Action. But I will add a check.

That should prevent future devs from making a mistake.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -36,11 +34,12 @@
 import org.apache.activemq.artemis.nativo.jlibaio.LibaioContext;
 import org.apache.activemq.artemis.nativo.jlibaio.LibaioFile;
 import org.apache.activemq.artemis.utils.FileUtil;
+import picocli.CommandLine;
 
 /**
  * CLI action that creates a broker instance directory.
  */
-@Command(name = "create", description = "Create a new broker instance.")
+@CommandLine.Command(name = "create", description = "Create a new broker instance.")
```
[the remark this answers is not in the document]
clebertsuconic: we could use import static.. which I don't like much.

But I can make an exception here

docs/user-manual/en/using-server.md
gemmellr: A note on how to exit the shell may also be in order, some people wont actually know.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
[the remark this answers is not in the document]
gemmellr: Dont think it should need to be static, its just a nested interface/class, regular import of the full name should work.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/user/UserGroup.java
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.activemq.artemis.cli.commands.user;
+
+import org.apache.activemq.artemis.cli.commands.HelpAction;
+import picocli.CommandLine;
+
+@CommandLine.Command(name = "user", description = "file-based user management (example ./artemis user list)", subcommands = {ListUser.class, AddUser.class, RemoveUser.class, ResetUser.class})
```
gemmellr, line 23: These group examples like "(example ./artemis user list)" show up in the shell help output, where it is actually _not_ what should be done. Not sure if it could cater to shell vs not-shell usage mode, but might be nice if it was possible.

clebertsuconic: I think I can specify the description on the API and use the isShell that I implemented. will take a look.

clebertsuconic: My options here are limited. I could create a different DataGroup for ShellDataGroup, or I can make a more generic text.

The other option would be some horrible hacking into the internal API.. which would be even worse. So I will go for the more generic text.

clebertsuconic: I think I have solved this..

@gemmellr  but I will let you decide if this one could be resolved or not

clebertsuconic:
I have added one more review from myself. When using user/password the system would ask me for the user/password all the time. I'm now saving the last successful user/password between executions, adding a connect and disconnect that's only available through the shell (what makes the shell more useful since it won't ask for user/password every time when using security)

clebertsuconic:
I will finish the other 2 items tomorrow. there's one item I didn't close (resolve) from our conversation here because I'm not sure it's the best choice yet.. will look at it again tomorrow.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/DestAbstract.java
```diff
@@ -21,29 +21,29 @@
...
-   @Option(name = "--message-count", description = "Number of messages to act on. Default: 1000.")
+   @Option(names = "--message-count", description = "Number of messages to act on. Default: 1000.")
    int messageCount = 1000;
 
-   @Option(name = "--sleep", description = "Time wait between each message.")
+   @Option(names = "--sleep", description = "Time wait between each message.")
    int sleep = 0;
 
-   @Option(name = "--txt-size", description = "Transaction batch size.")
+   @Option(names = {"--txt-size", "--tx-batch-size"}, description = "Transaction batch size. (--txt-size is deprecated)")
```
gemmellr, line 40: The non-deprecated option name being first would read better (unless the cli is reverse-ordering the options)

docs/user-manual/en/using-server.md
gemmellr: unresolved as the docs didnt seem to change

gemmellr:
> I have added one more review from myself. When using user/password the system would ask me for the user/password all the time. I'm now saving the last successful user/password between executions, adding a connect and disconnect that's only available through the shell (what makes the shell more useful since it won't ask for user/password every time when using security)

I think the ability to set and clear the URI + credentials used for subsequent commands in the shell is good. I'm less clear on it doing it silently in the background based on every previous command.

At least one of the additions you made was around it handling the URIs crossing between the different clients, presumably as the URI is saved but the protocol value is not, probably how you hit that, the client used will change between commands in some cases, whilst the URI + credentials etc do not even though the URIs are client-specific.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
```diff
@@ -159,11 +176,19 @@ private ConnectionFactory createAMQPConnectionFactory(String brokerURL,
       try {
          Connection connection = cf.createConnection();
          connection.close();
+         connectionSucces(brokerURL, user, password);
```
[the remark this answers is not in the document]
clebertsuconic: the connection is already called on line 177

docs/user-manual/en/using-server.md
clebertsuconic: I removed the line.. and I'm adding a chapter for the CLI.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/ActionAbstract.java
```diff
@@ -205,6 +205,20 @@ the ARTEMIS_HOME variable will include back slashes (An invalid file URI charact
       return brokerHome;
    }
 
+   @Override
+   public void run() {
+      try {
+         execute(getActionContext());
+      } catch (Throwable e) {
+         e.printStackTrace();
```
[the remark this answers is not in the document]
clebertsuconic: the Shell will only call the Runnable. So this to interface back to the execute method.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/Artemis.java
```diff
@@ -87,9 +68,29 @@
...
+
+   @Override
+   public void run() {
+      // We are running the shell by default.
+      // if you type ./artemis we will go straight to the shell
+      Shell.runShell();
+   }
 
    public static void main(String... args) throws Exception {
+      System.out.println("Line::" + getNameFromBanner());
```
gemmellr, line 93: Is this a leftover?

artemis-distribution/src/main/resources/licenses/bin/LICENSE
```diff
@@ -286,6 +286,12 @@ For HdrHistogram:
 This product bundles HdrHistogram, which is available under the
 "2-clause BSD" license. For details, see licences/LICENSE-hdrhistogram.txt.
 
+==============================================================================
+For JLine:
+==============================================================================
+This product bundles JLine, which is available under the
+"3-clause BSD" license. For details, see licences/LICENSE-jline.txt.
```
gemmellr, line 293: Referenced file still needs added.

docs/user-manual/en/using-server.md
```diff
@@ -53,6 +53,8 @@ The following highlights some important folders on the distribution:
 
 ## Creating a Broker Instance
 
+Before the broker is used, an broker *instance* must be created.
```
gemmellr, line 56: an -> a

I would put the sentence at the start of the existing paragraph below about broker instances.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/Artemis.java
clebertsuconic: yeah!! lol thank you

artemis-distribution/src/main/resources/licenses/bin/LICENSE
clebertsuconic: It was added here but apparently .txt is on the gitignore and I missed adding it. thank you.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/DestAbstract.java
clebertsuconic: I'm using hidden now...


and I opened a PR and bug on pico to fix the auto-completion on the shell:

https://github.com/remkop/picocli/pull/2075

clebertsuconic:
for my future reference. I opened an issue and Pull Request on Pico: https://github.com/remkop/picocli/pull/2075

the jshell auto completion is showing hidden values. the PR sent would fix it but I'm waiting for a best way to merge this upstream in Pico. hidden is the way to go to "deprecate" stuff.

clebertsuconic:
Only thing left on my todo is the JDK parameters and testing with a newer version. will do it before the end of today.

artemis-distribution/src/main/resources/bin/artemis
```diff
@@ -91,6 +91,9 @@ if $cygwin ; then
 fi
 
 exec "$JAVACMD" $JAVA_ARGS $ARTEMIS_CLUSTER_PROPS \
+    --add-opens java.base/java.lang=ALL-UNNAMED \
+    --add-opens java.base/java.io=ALL-UNNAMED \
+    --add-opens java.base/java.util=ALL-UNNAMED \
```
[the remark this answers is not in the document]
clebertsuconic: hmmmmmmm... I don't need these with JDK17.

@gnodet it would be so nice to get your hep here? :)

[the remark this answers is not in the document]
clebertsuconic: @gnodet I"m adding jline3 into artemis, integrated with pico-CLI. This PR is taking an example from the jline3 along with some examples I saw on pico-cli .. and I had to include these with JDK-11, while these are not needed in 17.


I'm confused what I need to do to integrate this. can you help us here?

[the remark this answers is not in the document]
gemmellr: Its not actually surprising that behaviour here might differ on different JDKs, thats exactly why I wondered if you tried it on the newer ones, especially with JDK21 almost done (RC1 due in 10 days, later RC2 if needed; JDK17 didnt get an RC2)...though I was more concerned it may need other config, than not need any, we will need to use what works for them all.

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis
```diff
@@ -102,6 +102,9 @@ fi
 
 exec "$JAVACMD" \
     $JAVA_ARGS \
+    --add-opens java.base/java.lang=ALL-UNNAMED \
+    --add-opens java.base/java.io=ALL-UNNAMED \
+    --add-opens java.base/java.util=ALL-UNNAMED \
```
gnodet, line 107: Adding the jansi library will remove the need for those additional JVM options.  Those are only required when using the exec provider, which won't be used if you add jansi to support windows.

artemis-distribution/src/main/resources/bin/artemis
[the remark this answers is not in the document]
gnodet: See above, adding jansi will remove the need for those options.

gnodet (changes requested):
You need to add jansi library, which is an optional dependency of JLine for better support.

artemis-cli/pom.xml
```diff
@@ -124,10 +124,17 @@
...
+         <groupId>info.picocli</groupId>
+         <artifactId>picocli</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>info.picocli</groupId>
+         <artifactId>picocli-shell-jline3</artifactId>
+      </dependency>
+      <dependency>
+         <groupId>org.jline</groupId>
+         <artifactId>jline</artifactId>
```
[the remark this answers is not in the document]
clebertsuconic: @gnodet thank you so much.. that worked like a charm!


@gemmellr  ^^^ C/C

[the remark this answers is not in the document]
clebertsuconic: @gnodet thank you so much for your help here.

clebertsuconic:
@gemmellr I am merging this... further changes can be sent into main after this.


Thanks for your help on this

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/DestAbstract.java
```diff
@@ -137,4 +141,16 @@ public DestAbstract setSerializer(String serializer) {
...
       return this;
    }
+
+   @Override
+   public Object execute(ActionContext context) throws Exception {
+      super.execute(context);
+
+      if (oldBatchSize > 0) {
+         context.out.println("--txt-size is deprecated, please use --commit-interval");
+         txBatchSize = oldBatchSize;
```
gemmellr, line 151: If both were somehow set, I would expect the new one to always win...here the old one will win.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/perf/PerfCommand.java
```diff
@@ -26,46 +26,56 @@
...
    private final CountDownLatch completed = new CountDownLatch(1);
 
    @Override
    public Object execute(ActionContext context) throws Exception {
       super.execute(context);
-      final ConnectionFactory factory = createConnectionFactory(brokerURL, user, password, null, getProtocol());
+      if (txSize > 0) {
+         System.out.println("--tx-size is deprecated, please use --commit-interval");
+         commitInterval = txSize;
+      }
```
gemmellr, line 77: If both were somehow set, I would expect the new one to always win...here the old one will win.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/Artemis.java
```diff
@@ -194,68 +191,140 @@ public static Object execute(boolean inputEnabled, boolean useSystemOut, File ar
...
-      Action action = builder(artemisInstance).build().parse(args);
-      action.setHomeValues(artemisHome, artemisInstance, etcFolder);
+      boolean isInstance = artemisInstance != null || System.getProperty("artemis.instance") != null;
+      CommandLine commandLine = buildCommand(isInstance, !isInstance);
+
+      Object userObject = parseAction(commandLine, args);
+
+      // Pico shouldn't allow generating a commandLine without an userObject.
+      // the following assert "should" never happen
+      assert userObject != null;
```
gemmellr, line 205: Good chance the assert also wont happen unless people turn on assertions. Seems like a null check would have been simpler. Though it will at least still NPE later, when failing to throw the IAE.

clebertsuconic: this is never supposed to happen. 
Only chance userObject is null is if you set the userObject wrongly during development. and picocli actually validate that already.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ConnectionAbstract.java
```diff
@@ -157,68 +174,132 @@ private ConnectionFactory createAMQPConnectionFactory(String brokerURL,
...
+         user = inputUser(user);
+         password = inputPassword(password);
+         cf = new ActiveMQConnectionFactory(brokerURL, user, password);
          if (clientID != null) {
             cf.setClientID(clientID);
          }
+         try {
+            tryConnect(brokerURL, user, password, cf);
+         } catch (Exception e2) {
+         }
```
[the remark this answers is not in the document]
clebertsuconic: not really in both cases.. This is the exception of the exception. IN case the input user/password does not succeed, if this does not work this will simply not save the password here.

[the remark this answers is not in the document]
clebertsuconic: I see what you mean now.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/Artemis.java
gemmellr: Fair enough. You didnt note the last bit before, just that 'we shouldnt add things that arent Runnable or Action', then you added the fallback check which would NPE if it was null, which I then noted.  I'd still have just added a simple Objects.nonNull("oh-dear developer-error", userObject) rather than the multi line assert+comment that quite posibly still wont do anything and allow the raw NPE to occur.

PR #4567, opened 2023-07-24
NO-JIRA move Error Prone to annotationProcessorPaths

amarkevich:


clebertsuconic:
you are making error prone to always run on this case, aren't you?


We currently only run error prone with -Derrorprone defined in the build. This seems to be running it always?

amarkevich:
explicit annotationProcessorPaths configuration added only to errorprone profiles

gemmellr:
It is still only running it in the profile, just a different way.

However, as I said a few weeks ago on the original PR, I would leave ErrorProne configured the way it is, it is simpler than this alternative. Emmanuel deliberately didnt use annotationProcessorPath config when he initially reworked the build for JDK > 9 a while back because of the 'must then specify every processor' hassle that comes with it. On seeing it, I definitely agree with that decision. More so now that its isolated to its own profile and never enabled by default, which wasnt true back then.

Adding explicit annotationProcessorPath config that complicates the build (and more so if we e.g add other annotation processors later, or need any additional ErrorProne profiles to handle JDK disparities), for something which is already isolated to a profile and only activated on request,doesnt really seem like an improvement.

It only ever caused any problem on the original PR initially because you had removed dependencyManagement that we will want to be there, regardless of whether we used annotationProcessorPath for this or not, or even used ErrorProne at all. I'm tired of repeating the same feedback.

amarkevich:
configuring annotation processor via dependencyManagement cause changing transitive dependencies versions. thats why annotationProcessorPaths should be used to isolate this functionality. from architectural point of view annotation processor can be splitted to optional/provided dependency with custom annotations and processor itself which activated via annotationProcessorPaths

gemmellr:
Yes this way can be better in some regards, but at a cost of increased complexity, and in this case really only in a specific on-demand development case that is already isolated by profile, and doesnt apply at all in the regular builds. I would say the added complexity outweighs the benefit in the current case. As did another developer previously when expressly avoiding doing this (even though they instead selected adding the processor as a constantly present provided dep in every module in the tree...I later removed all that and profiled it off as as present to isolate it all from normal builds, as yes the normal build should not have ErrorProne dependencies in it).

We are are only setting the version of the artifact via dependencyManagement so its consistent and defined in one place. The configuration of actually using the processor is being done on demand via the dependency within the profile, which is already isolating it to be optionally enabled-on-request, typically only in CI checks, adding it as the last dependency entry. The version being managed for interesting transitive dependencies such as guava means there is already minimal scope for change even there (except, say, if you also removed that dependencyManagement against feedback).

This is simpler as it is and for the vast majority of the time, i.e all regular build ones, still the same as it would be after this change. Lets jsut leave it as it is.

amarkevich:
> added complexity outweighs the benefit in the current case

from my point of view - do things right way and avoid annotation processor magic from compile dependencies

pom.xml
```diff
@@ -1191,6 +1185,18 @@
                               <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED</arg>
                               <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED</arg>
                           </compilerArgs>
+                          <annotationProcessorPaths>
+                              <path>
+                                  <groupId>org.apache.activemq</groupId>
+                                  <artifactId>artemis-log-annotation-processor</artifactId>
+                                  <version>${project.version}</version>
+                              </path>
```
clebertsuconic, line 1193: hmmm... it will be difficult to maintain this down the road.

you don't specify log-annotation-processor if not using errorprone, but if using, then you do...


I won't know what's going in a few weeks from now.... a few months or an year for now and I have to debug anything this gets unpredictable, right?

gemmellr:
I hadn't even thought of that problem until now, re: the new build failures. By just putting the log annotation processor in the processorPath for every module it then also applies to those modules built before the processor module, such as the unit test support module, which the processor module itself depends on for tests, thus creating a cycle. The build then only partly works at all due to it having downloaded stale snapshot output from a _previous_ build and seemingly satisfy the cycle that exists. Even that ugliness will presumably fail for release builds or the first post-release builds though, when no prior deployed release or snapshot will exist yet for it to download.

Even if its possible to work around that with yet more configuration/complexity, again just leaving this as it was seems preferable to me.

PR #4569, opened 2023-07-25
[ARTEMIS-4159] Support duplicate cache size configuration per address

fartzy:
### **Why were these changes made?**
This pull request introduces support for configuring a specific id-cache-size per address - [ARTEMIS-4159](https://issues.apache.org/jira/browse/ARTEMIS-4159). The changes include modifications to various files:
* artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
* artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
* artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
* artemis-server/src/main/resources/schema/artemis-configuration.xsd
* artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
* artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationTest.java
* artemis-server/src/test/java/org/apache/activemq/artemis/core/settings/AddressSettingsTest.java
* artemis-server/src/test/resources/ConfigurationTest-full-config.xml
* artemis-server/src/test/resources/ConfigurationTest-xinclude-config-address-settings.xml
* artemis-server/src/test/resources/ConfigurationTest-xinclude-schema-config-address-settings.xml
* docs/user-manual/en/address-settings.md
* docs/user-manual/en/configuration-index.md
* docs/user-manual/en/duplicate-detection.md
* tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/DuplicateDetectionTest.java

**Changes Made:**
* FileConfigurationParser.java
Added a constant ID_CACHE_SIZE to represent the id-cache-size attribute in the XML configuration.
Updated the FileConfigurationParser to parse the id-cache-size attribute from the XML and set it on the AddressSettings.
* PostOfficeImpl.java
Introduced a new method resolveIdCacheSize to determine the effective id-cache-size for an address.
Updated getDuplicateIDCache method to use the resolveIdCacheSize method to get the proper id-cache-size for the address.
* AddressSettings.java
Added a new attribute idCacheSize to store the id-cache-size configuration for an address.
Implemented getter and setter methods for idCacheSize.
* artemis-configuration.xsd
Updated the XSD schema to include the id-cache-size attribute for the address settings.
* ConfigurationImplTest.java
Added a new test method to verify parsing of id-cache-size configuration from prefixed properties.
* FileConfigurationTest.java
Added a new test method to verify the parsing of the id-cache-size configuration from the XML configuration file.
* ConfigurationTest-full-config.xml, ConfigurationTest-xinclude-config-address-settings.xml, ConfigurationTest-xinclude-schema-config-address-settings.xml
Added <id-cache-size> elements to the address settings to test the configuration.
* DuplicateDetectionTest.java
Added a new test method to verify the functionality of the address-specific id-cache-size setting.
* AddressSettingsTest.java 
Added to the testSingleMerge to make sure idCacheSize merges correctly.

* docs/user-manual/en/address-settings.md, docs/user-manual/en/configuration-index.md, docs/user-manual/en/duplicate-detection.md
Added documentation for the above for address specific information about `id-cache-size`.

**User Facing Changes:**
New Configuration Option: id-cache-size for Address Settings
With this pull request, administrators now have the ability to configure a specific id-cache-size for each address, allowing better customization and optimization of resources for different addresses.

**Motivation:**
This enhancement provides the ability to configure a specific id-cache-size for each address, allowing better customization and optimization of resources for different addresses. Administrators can now specify the cache size for individual addresses, which will default to the global setting if not explicitly set for an address. This improvement eliminates the need for deploying multiple instances of Artemis with different global cache sizes based on address requirements. The pull request aims to simplify configuration and reduce resource consumption in scenarios where different addresses have different caching needs.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
@@ -342,6 +342,8 @@ public final class FileConfigurationParser extends XMLConfigurationUtil {
 
    private static final String ENABLE_INGRESS_TIMESTAMP = "enable-ingress-timestamp";
 
+   private static final String ID_CACHE_SIZE = "id-cache-size";
+
    private boolean validateAIO = false;
 
    private boolean printPageMaxSizeUsed = false;
@@ -1456,6 +1458,10 @@ public final class FileConfigurationParser extends XMLConfigurationUtil {
             addressSettings.setEnableMetrics(XMLUtil.parseBoolean(child));
          } else if (ENABLE_INGRESS_TIMESTAMP.equalsIgnoreCase(name)) {
             addressSettings.setEnableIngressTimestamp(XMLUtil.parseBoolean(child));
+         } else if (ID_CACHE_SIZE.equalsIgnoreCase(name)) {
+            int idCacheSize = XMLUtil.parseInt(child);
+            Validators.GE_ZERO.validate(ID_CACHE_SIZE, idCacheSize);
+            addressSettings.setIDCacheSize(XMLUtil.parseInt(child));
          }
       }
       return setting;
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -1407,9 +1407,15 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
       return null;
    }
 
+   private int resolveIdCacheSize(SimpleString address) {
+      final AddressSettings addressSettings = addressSettingsRepository.getMatch(address.toString());
+      return addressSettings.getIDCacheSize() == null ? idCacheSize : addressSettings.getIDCacheSize();
+   }
+
    @Override
    public DuplicateIDCache getDuplicateIDCache(final SimpleString address) {
-      return getDuplicateIDCache(address, idCacheSize);
+      int resolvedIdCacheSize = resolveIdCacheSize(address);
+      return getDuplicateIDCache(address, resolvedIdCacheSize);
    }
 
    @Override
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
@@ -293,6 +293,8 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
 
    private Boolean enableIngressTimestamp = null;
 
+   private Integer idCacheSize = null;
+
    //from amq5
    //make it transient
    private transient Integer queuePrefetch = null;
@@ -370,6 +372,7 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
       this.managementMessageAttributeSizeLimit = other.managementMessageAttributeSizeLimit;
       this.slowConsumerThresholdMeasurementUnit = other.slowConsumerThresholdMeasurementUnit;
       this.enableIngressTimestamp = other.enableIngressTimestamp;
+      this.idCacheSize = other.idCacheSize;
    }
 
    public AddressSettings() {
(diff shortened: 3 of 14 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -1797,6 +1817,8 @@ public void encode(ActiveMQBuffer buffer) {
 
       BufferHelper.writeNullableLong(buffer, pageLimitMessages);
 
+      BufferHelper.writeNullableInteger(buffer, idCacheSize);
```
jbertram, line 1820: The writing and reading of the nullable `Integer` is not in the same order so decoding will fail.

michaelandrepearce:
Nice contribution, i see justins spotted already one issue to be fixed.
I would ask if we can also update documentation for this configuration and feature.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
fartzy: I moved the read up to the 4th ordinal position from the bottom so they are both same now. Thank you for catching.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1407,9 +1407,16 @@ public Pair<RoutingContext, Message> redistribute(final Message message,
       return null;
    }
 
+   // ARTEMIS-4159
```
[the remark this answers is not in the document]
fartzy: I took out this comment

fartzy:
> 

I added some doc to three of the `.md` files. Wherever `id-cache-size` is in the doc already then I added an address specific setting to that.  And also I added to the general `address-settings.md`.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
jbertram: Both the read and the write for the `idCacheSize` should be at the *end* otherwise the broker may choke on existing persisted address settings when upgrading.

jbertram:
All commits should be squashed and the commit message should follow the pattern outlined in the [Hacking Guide](https://github.com/apache/activemq-artemis/blob/main/docs/hacking-guide/en/code.md#commitMessageDetails).

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
fartzy: oh yeah it will definitely get mixed up with currently persisted messages - thanks. I put both the read and write at end now

artemis-server/src/main/java/org/apache/activemq/artemis/core/deployers/impl/FileConfigurationParser.java
```diff
@@ -1456,6 +1458,10 @@ protected Pair<String, AddressSettings> parseAddressSettings(final Node node) {
             addressSettings.setEnableMetrics(XMLUtil.parseBoolean(child));
          } else if (ENABLE_INGRESS_TIMESTAMP.equalsIgnoreCase(name)) {
             addressSettings.setEnableIngressTimestamp(XMLUtil.parseBoolean(child));
+         } else if (ID_CACHE_SIZE.equalsIgnoreCase(name)) {
+            int idCacheSize = XMLUtil.parseInt(child);
+            Validators.GT_ZERO.validate(ID_CACHE_SIZE, idCacheSize);
```
jbertram, line 1463: I wonder if this should use `Validators.GE_ZERO` so that users can disable the cache for specific addresses even if a "parent" address-setting uses a value > 0.

fartzy: Yeah makes sense! changed to `Validators.GE_ZERO`

jbertram:
Great work here, @fartzy. Thanks for the contribution!

PR #4571, opened 2023-07-27
ARTEMIS-4377 add script for pushing docker images for a release

jbertram:


jbertram:
This technically depends on #4570.

RELEASING.md
```diff
@@ -335,6 +335,15 @@ Note: Generating PDFs, etc. with gitbook requires the installation of [Calibre](
 You can install this manually, but it is recommended you use your platform's package management to install
 (e.g. `sudo apt-get install calibre`).
 
+## Upload Docker Images
+
+1. If you don't have an account on https://hub.docker.com/ then create one.
+2. [Install `docker`](https://docs.docker.com/engine/install/) in your environment.
+3. Ensure you have access to push images to `apache/activemq-artemis`. If you don't have access you can request it by
+   creating an INFRA Jira ticket (e.g. https://issues.apache.org/jira/browse/INFRA-24831).
+4. Go to the `scripts` directory, open `release-docker.sh`, modify the `VERSION` variable for the version you're
```
clebertsuconic, line 344: Can't you pass that by parameter? $1 ?

scripts/release-docker.sh
```diff
@@ -0,0 +1,37 @@
...
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+COMMAND="docker"
+VERSION="2.30.0"
+REPO="apache"
```
clebertsuconic, line 21: I think you should parameterize at least the version. 

parameterizing command and repo would be too much but it would be nice... but if you could parameterize at least the version?

clebertsuconic:
this is still not parameterized. and it needs rebase against the new RELEASING

scripts/release-docker.sh
jbertram: Done.

jbertram:
@clebertsuconic, it should all be fixed now.

PR #4572, opened 2023-08-01
ARTEMIS-4382 Long Time to process export / import

clebertsuconic:


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/DBOption.java
```diff
@@ -172,7 +183,7 @@ public Object execute(ActionContext context) throws Exception {
       super.execute(context);
 
       if (output != null) {
-         fileOutputStream = new FileOutputStream(output);
+         fileOutputStream = new BufferedOutputStream(new FileOutputStream(output));
```
clebertsuconic, line 186: this is the fix for the export ... it would take forever without this..

clebertsuconic:
I'm inducing errors on the CI to validate the changes. hence a draft.

PR #4573, opened 2023-08-01
ARTEMIS-4383 migrate user docs to AsciiDoc

jbertram:
Markdown lacks features which would make the documentation easier to read, easier to navigate, and just look better overall.

The current tool-chain uses honkit and a tool called Calibre. Honkit is written in TypeScript and is installed via NPM. Calibre is a native tool so it must be installed via an OS-specific package manager. All this complexity makes building, releasing, uploading, etc. a pain.

AsciiDoc is relatively simple like Markdown, but it has more features for presentation and navigation not to mention Java-based Maven tooling to generate both HTML and PDF. Migrating will improve both the appearance of the documentation as well as the processes to generate and upload it.

This commit contains the following changes:
 - Convert all the Markdown for the User Manual, Migration Guide, and
   Hacking guide to AsciiDoc via kramdown [1].
 - Update the `artemis-website` build to use AsciiDoctor Maven tooling.
 - Update `RELEASING.md` with simplified instructions.
 - Update Hacking Guide with simplified instructions.
 - Use AsciiDoc link syntax in Artemis Maven doc plugin.
 - Drop EPUB & MOBI docs for User Manual as well as PDF for the Hacking
   Guide. All docs will be HTML only except for the User Manual which
   will have PDF.
 - Move all docs up out of their respective "en" directory. This was a
   hold-over from when we had docs in different languages.
 - Migration & Hacking Guides are now single-page HTML since they are
   relatively short.

Benefits of the change:
 - Much simplified tooling. No more NPM packages or native tools.
 - Auto-generated table of contents for every chapter.
 - Auto-generated anchor links for every sub-section.
 - Overall more appealing presentation.
 - All docs will use the ActiveMQ favicon.
 - No more manual line-wrapping! AsciiDoc recommends one sentence per
   line and paragraphs are separated by a blank line.
 - AsciiDoctor plugins for IDEA are quite good.
 - Resulting HTML is less than *half* of the previous size.

All previous links/bookmarks should continue to work.

[1] https://github.com/asciidoctor/kramdown-asciidoc


jbertram:
If you want to peruse the new docs just pull down the branch and do the following:
```sh
cd artemis-website
mvn clean package -Prelease -DskipWebsitePdfGeneration
```
The new User Manual will be in `target/user-manual`. Open `index.html` in your browser.

docs/user-manual/_diagrams/client-classpath-jakarta.adoc
```diff
@@ -1 +1 @@
-This file will be generated/replaced in compile time by artemis-website
\ No newline at end of file
+This file will be generated/replaced in compile time by artemis-website
```
gemmellr, line 1: Is this meant to be in the diagrams dir now?

docs/user-manual/_diagrams/client-classpath-jms.adoc
```diff
@@ -1 +1 @@
-This file will be generated/replaced in compile time by artemis-website
\ No newline at end of file
+This file will be generated/replaced in compile time by artemis-website
```
gemmellr, line 1: Ditto

docs/user-manual/index.adoc
```diff
@@ -0,0 +1,136 @@
+:idprefix:
+:idseparator: -
+
```
gemmellr, line 3: Should probably have a note comment in here, indicating people need to update the separate *_book.adoc* file as well for the PDF, mirroring the comment in that file.

gemmellr (commented):
Looks really good. I had actually just thought to myself on Monday 'I wonder if Justin started on that asciidoc docs idea he noted several weeks back' hehe, guess so :) 

Beyond the minor niggles in the comments, the generated output also has a couple things it might be nice to change.

The CSS is embedded in every file it seems. I'd expect theres a way to request otherwise, or else add the same content in a file provided as being 'ours'.

The pages have a last updated footer, which seems unnecessarily large and also means it isnt reproducible and every update will update every page. Would be nice to either remove it or find a way to have it be set so it is reproducible.

docs/user-manual/_diagrams/client-classpath-jakarta.adoc
jbertram: No. I nuked it.

jbertram:
Thanks for the feedback, @gemmellr. I believe I've address all your inline suggestions as well as the shared CSS and footer issue.

After doing the 2.30.0 release I decided enough was enough and pulled the trigger on this change. :smile:

gemmellr:
Looks good re: previous comments, but I noticed something thing needing fixed while reviewing that (possibly didnt see it amongst the inline CSS earlier hehe).

The generated pages are still linking out for fonts, while we are meant to host all site resources like that locally these days. One at least seems to have a precise option for controlling it (https://docs.asciidoctor.org/asciidoctor/latest/html-backend/local-font-awesome/) , the other isnt clear.

clebertsuconic:
this LGTM.

clebertsuconic:
@jbertram please change README... ##Documentation still showing SUMMARY.md

and a note to how to build it would be great on README.

clebertsuconic:
I think we should merge this right away... and further changes can be made on main


really nice work @jbertram

gemmellr:
Generally nicer to let in-progress reviews complete before merging, especially when changes have been made based on previous review feedback. Then hopefully stuff doesnt end up on main before it is ready to, and doesnt e.g break the builds as this has. It is also missing a src release LICENCE update for the font files added. Its also still using remote resources.

clebertsuconic:
@gemmellr I didn't think this was breaking the build. I thought it would be ok to keep changing on main and collaborate on main after the merge.  Sorry if I caused an issue there.

PR #4576, opened 2023-08-09
Fix a memory leak in HandlerBase.

cardamon:
The `ThreadLocal`'s value in `HandlerBase` is never cleaned up.

Found by log statements issued by Tomcat during the undeployment of a web application that uses an embedded broker instance.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/HandlerBase.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/HandlerBase.java
@@ -37,11 +37,18 @@ public abstract class HandlerBase {
    }
 
    public static boolean inHandler() {
-      return inHandler.get().count > 0;
+      int count = inHandler.get().count;
+      if (count == 0) {
+         inHandler.remove();
+      }
+      return count > 0;
    }
 
    protected static void leave() {
-      inHandler.get().count--;
+      int count = --inHandler.get().count;
+      if (count == 0) {
+         inHandler.remove();
+      }
    }
 
 }


jbertram:
This looks good. Thanks for the contribution! 

However, before merging this needs a couple of things...

First, it needs an associated Jira. If you don't already have an Apache Jira account you can [request one](https://selfserve.apache.org/jira-account.html). I'll watch for your request and approve it ASAP. Once you create a Jira you should update the commit message to reference it. See more details about commit messages in our [Hacking Guide](https://github.com/apache/activemq-artemis/blob/main/docs/hacking-guide/_code.adoc#commitMessageDetails).

Second, it needs a test to both validate the fix and also prevent regressions in the future.

cardamon:
Thanks, will do!

(may be a few days before I get to it)

jbertram:
@cardamon, do you plan on following up on this PR with the requested changes?

cardamon:
Actually, I'm not sure how this could be easily & meaningfully tested in a unit test. `ThreadLocal` instances are really quite annoying. The don't allow their value to be queried without actually setting it.

One direction would be for the `ThreadLocal` to be mocked with Mockito, and then verify that the correct calls are made. But the unit test would just be a replication of the logic of the code that's being tested. Although it would I suppose prevent regressions by accidental modification of code.

Would that be enough?

clebertsuconic:
a check-leak test?

https://github.com/apache/activemq-artemis/tree/main/tests/leak-tests

cardamon:
> @cardamon : a check-leak test?
> 
> https://github.com/apache/activemq-artemis/tree/main/tests/leak-tests

Didn't knew that existed as I'm not that familiar with the project but I'll have a look at that, thx.

clebertsuconic:
This is being replaced by https://github.com/apache/activemq-artemis/pull/4880

PR #4577, opened 2023-08-14
ARTEMIS-4387 Fix empty consumer filter string leak

jsmucr:
Repeated creation and disposal of queue consumers with empty string (`""`) filters could cause these strings to build up within a `QueueInfo` instance belonging to that queue.


jsmucr:
> @jsmucr please check the extra commit I added for your fix. it's been merged.

Thank you.
Yes, that leak detect tool of yours looks handy. :-)

clebertsuconic:
@jsmucr BTW: thank you so much for the PR and contribution. the fix was spot on.. very nice!

(I had written this somewhere before but I guess I missed clicking into something.)

PR #4578, opened 2023-08-15
ARTEMIS-4390 Fix upgrade-linux smoke test on Windows

jsmucr:
The test cannot work on Windows unless I can make the `upgrade` CLI command respect my choice to upgrade a Linux distribution. This commit therefore adds a new `--linux` option for the `upgrade` command, and leverages it in the `upgrade-linux` smoke test.

* The `--cygwin` option has been preserved for backwards compatibility.
* The `IS_CYGWIN` attribute has been renamed to `IS_NIX` to reflect the change.
* The OS "recognition" method (in `InstallAbstract::run`) has been updated to reflect the need for enforcing *nix behavior, which is now the default if all other methods fail.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
@@ -756,7 +756,7 @@ public class Create extends InstallAbstract {
          writeEtc(ETC_ARTEMIS_PROFILE_CMD, etcFolder, filters, false);
       }
 
-      if (!IS_WINDOWS || IS_CYGWIN) {
+      if (IS_NIX) {
          write(BIN_ARTEMIS, filters, true);
          makeExec(BIN_ARTEMIS);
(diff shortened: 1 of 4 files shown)


jsmucr:
@clebertsuconic This is more of a suggestion. I was unable to build Artemis on Windows.

clebertsuconic:
@jsmucr can you create a JIRA, and amend your commit with the jira ID in the prefix?

jsmucr:
> @jsmucr can you create a JIRA, and amend your commit with the jira ID in the prefix?

Done.

jbertram:
I think a much simpler fix would be to just ignore the test if the OS is not Linux, e.g.:
```
Assume.assumeTrue(System.getProperty("os.name").toUpperCase().contains("LINUX");
```

jsmucr:
That's not how I usually do things. 🙂

clebertsuconic:
I'm running the whole CI on this. if it's all good I will merge it tomorrow.


thanks again.

PR #4581, opened 2023-08-17
ARTEMIS-4392 Upgrade Netty to 4.1.96.Final

ehsavoie:


jbertram:
The version number is wrong in the commit message. It should be `4.1.96.Final`. Once you fix that I'll merge it.

PR #4582, opened 2023-08-21
ARTEMIS-4393 Explicit logger implementation instantiation

amarkevich:
- compilation error instead of runtime exception in case of logger annotation processor issue;
- get rid of java.lang.reflect.* & java.security.* usage for logger case


jbertram:
What's the use-case for this? I've never had a problem with compile-time vs. runtime issues here.

clebertsuconic:
I prefer to leave the Bundle as it was there before and not depend on the generated implementation from the interface. I prefer to keep the interface not knowing about the implementations.. which that was the real use case. (Say we could even expand the logger for new functionality).

clebertsuconic:
@amarkevich can we keep as it was before here?

amarkevich:
> a problem with compile-time vs. runtime issues here

in case artemis-log-annotation-processor broken @LogBundle consumers will be built successfully but raise CNFE in runtime

jbertram:
> in case artemis-log-annotation-processor broken @LogBundle consumers will be built successfully but raise CNFE in runtime

Surely this is an uncommon event, and even if it does happen then any test or release validation will fail, right?

amarkevich:
> any test

tests running during PR are green anyway

amarkevich:
> keep the interface not knowing about the implementations

its better to hide logger implementation - done

clebertsuconic:
-1000 on this.. sorry!!!


The whole point of the BundleFactory is not require knowing the implementation from the Interface, and having the Factory figuring out what to do.


I don't mind about this reflection here as it's only happening once.. it's more trouble really.


the framework where this came from is Jboss Logging, where we are doing exactly what it was done before, except that we did that to fix some licensing and to be able to use SLF4J. I hate to block a PR from someone but I strongly disagree with this one man! sorry about that!

amarkevich:
Log Annotation Processor looks like any other regular annotation processors such as Lombok, Mapstruct or jpa-codegen, when generated code can be used directly in the code. There is no any reasonable points why its prohibited in this case.

clebertsuconic:
It's not prohibited. just not desirable.

The dependency between the interface and its implementation is supposed to be determined at runtime by that class. If we keep the dependency explicit by just a new instance, perhaps we could just get rid of the whole processing.. but it's not something I'm looking to do at this point.

amarkevich:
I see this interface as place where annotation can be generally placed. And log processor logic is well-known (_impl class suffix), like jpa modelgen: `The name of the metamodel class is derived from the name of the managed class by appending "_" to the name of the managed class.` Having this rule documented - its official way to instantiate logger implementation.

clebertsuconic:
We don't see it that way. Interfaces shouldn't have the implementation defined in there. The factory provides somewhat an abstraction. 

We can agree to disagree here.

amarkevich:
> Interfaces shouldn't have the implementation defined in there.

So in case abstract class with annotations will be used instead of current interface - it can be approved?

amarkevich:
> _concrete use-case_

I listed them in jira:

- compilation error instead of runtime exception in case of logger annotation processor issue;
- get rid of java.lang.reflect.* & java.security.* usage for logger case
- hide logger implementation

And still no any valuable cons

clebertsuconic:
Man, this is painting the bike shed.. you're not hiding the annotation.. 
Look how Jboss Logging used to work.. we don't want to diverge here.

If I was up to do new LoggingImplementation. I would just get rid of the annotation and use a concrete implementation directly.

and please, don't do that.

clebertsuconic:
the annotation generates the implementation, and the interface must not have knowledge about the implementation, done thought the little framework.

if I could instantiate directly, we would just get rid of the processor and go straight to the implementation always.

amarkevich:
> you're not hiding the annotation

yep - logger implementation details are hidden

> Jboss Logging

is not a part of this project

> use a concrete implementation directly

current log processor generates message prefix general way and protect log messages from being corrupted by mistake

amarkevich:
> interface must not have knowledge

its not mandatory. I see this interface like a centralized place where all log annotations collected

amarkevich:
> than this

Aren't listed pros enough?

clebertsuconic:
you're doing compile time validation.. sure.but that's a small gain compared to other things. as I said I would rather keep it this way.

jbertram:
> I listed them in jira...

The bullet points you provided in the Jira and here in the PR are _not_ a use-case. They are simply a summary of what you changed. There is no real explanation of _why_ the existing code is problematic or _how_ your change benefits the project. Perhaps those things are obvious to you, but they are not obvious to me and likely not obvious to many other developers and users.

> And still no any valuable cons

Generally speaking, the project prefers stability and continuity as that typically mitigates risk for our users - something which is greatly valued. Therefore, anybody who is proposing a change must make an understandable case for _why_ the change should be accepted.

> Aren't listed pros enough?

At this point the pros you've listed are not enough.

amarkevich:
> prefers stability

in one of previous PR I made a mistake and broke artemis-log-annotation-processor. And PR check passed. This is my own expirience of using this project as developer. Feel free to keep such behaviour.

> why the change should be accepted

get rid of reflection is the main point

PR #4583, opened 2023-08-21
ARTEMIS-966 MQTT subscription state isn't durable

jbertram:
Durable subscrption state is part of the MQTT specification which has not been supported until now. This functionality is implemented via an internal last-value queue. When an MQTT client creates, updates, or adds a subscription a message using the client-ID as the last-value is sent to the internal queue. When the broker restarts this data is read from the queue and populates the in-memory MQTT data-structures. Therefore subscribers can reconnect and resume their session's subscriptions without have to manually resubscribe.

The subscriptions are serialized as raw bytes with a "version" byte for potential future use, but I intentionally avoided adding complex scaffolding to support multiple versions. We can add that complexity later if necessary.

### DIFF
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionState.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionState.java
@@ -30,9 +30,14 @@ import java.util.regex.Pattern;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.handler.codec.mqtt.MqttProperties;
+import io.netty.handler.codec.mqtt.MqttQoS;
+import io.netty.handler.codec.mqtt.MqttSubscriptionOption;
 import io.netty.handler.codec.mqtt.MqttTopicSubscription;
+import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
+import org.apache.activemq.artemis.api.core.Message;
 import org.apache.activemq.artemis.api.core.Pair;
 import org.apache.activemq.artemis.core.config.WildcardConfiguration;
+import org.apache.activemq.artemis.core.message.impl.CoreMessage;
 import org.apache.activemq.artemis.core.settings.impl.Match;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -42,11 +47,13 @@ public class MQTTSessionState {
 
    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
-   public static final MQTTSessionState DEFAULT = new MQTTSessionState(null);
+   public static final MQTTSessionState DEFAULT = new MQTTSessionState((String) null, null);
 
    private MQTTSession session;
 
-   private String clientId;
+   private final String clientId;
+
+   private final MQTTStateManager stateManager;
 
    private final ConcurrentMap<String, Pair<MqttTopicSubscription, Integer>> subscriptions = new ConcurrentHashMap<>();
 
@@ -91,8 +98,50 @@ public class MQTTSessionState {
 
    private Map<String, Integer> serverTopicAliases;
 
-   public MQTTSessionState(String clientId) {
+   public MQTTSessionState(String clientId, MQTTStateManager stateManager) {
       this.clientId = clientId;
+      this.stateManager = stateManager;
+   }
+
+   /**
+    * This constructor deserializes session data from a message. The format is as follows.
+    *
+    *  - byte: version
+    *  - int: subscription count
+    *
+    *  There may be 0 or more subscriptions. The subscription format is as follows.
+    *
+    *  - String: topic name
+    *  - int: QoS
+    *  - boolean: no-local
+    *  - boolean: retain as published
+    *  - int: retain handling
+    *  - int (nullable): subscription identifier
+    *
+    * @param message the message holding the MQTT session data
+    * @param stateManager the manager used to add and remove sessions from storage
+    */
+   public MQTTSessionState(CoreMessage message, MQTTStateManager stateManager) {
+      logger.debug("Deserializing MQTT session state from {}", message);
+      this.clientId = message.getStringProperty(Message.HDR_LAST_VALUE_NAME);
+      this.stateManager = stateManager;
+      ActiveMQBuffer buf = message.getDataBuffer();
+
+      // no need to use the version at this point
+      byte version = buf.readByte();
+
+      int subscriptionCount = buf.readInt();
+      logger.debug("Deserializing {} subscriptions", subscriptionCount);
+      for (int i = 0; i < subscriptionCount; i++) {
+         String topicName = buf.readString();
+         MqttQoS qos = MqttQoS.valueOf(buf.readInt());
+         boolean nolocal = buf.readBoolean();
+         boolean retainAsPublished = buf.readBoolean();
+         MqttSubscriptionOption.RetainedHandlingPolicy retainedHandlingPolicy = MqttSubscriptionOption.RetainedHandlingPolicy.valueOf(buf.readInt());
+         Integer subscriptionId = buf.readNullableInt();
+
+         subscriptions.put(topicName, new Pair<>(new MqttTopicSubscription(topicName, new MqttSubscriptionOption(qos, nolocal, retainAsPublished, retainedHandlingPolicy)), subscriptionId));
+      }
    }
 
    public MQTTSession getSession() {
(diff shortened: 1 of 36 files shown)


artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionState.java
```diff
@@ -91,8 +98,50 @@ public class MQTTSessionState {
...
+    *  - int: subscription count
+    *
+    *  There may be 0 or more subscriptions. The subscription format is as follows.
+    *
+    *  - String: topic name
+    *  - int: QoS
+    *  - boolean: no-local
+    *  - boolean: retain as published
+    *  - int: retain handling
+    *  - int (nullable): subscription identifier
```
gemmellr, line 119: Curious, what is the case when a durable subscription wouldnt have an id?

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTUtil.java
```diff
@@ -372,7 +374,7 @@ public static void logMessage(MQTTSessionState state, MqttMessage message, boole
                   break;
             }
 
-            logger.trace(log.toString());
+            logger.info(log.toString());
```
gemmellr, line 377: Seems a bit spammy, leftover debug change? EDIT Also the whole method is wrapped in an 'isTraceEnabled' gate.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/MQTT5Test.java
```diff
@@ -82,6 +83,69 @@ public void testTimestamp() throws Exception {
       context.close();
    }
 
+   @Test(timeout = DEFAULT_TIMEOUT)
+   public void testResumeSubscriptionsAfterRestart() throws Exception {
+      final int SUBSCRIPTION_COUNT = 100;
+      List<String> topicNames = new ArrayList<>(SUBSCRIPTION_COUNT);
+      for (int i = 0; i < SUBSCRIPTION_COUNT; i++) {
+         topicNames.add(RandomUtil.randomString());
```
gemmellr, line 91: trace logs etc might be easier to read if just using simple 'getTestName + i' names.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/MQTT5TestSupport.java
```diff
@@ -92,7 +92,7 @@ public class MQTT5TestSupport extends ActiveMQTestBase {
    public static Collection<Object[]> getParams() {
       return Arrays.asList(new Object[][] {
          {TCP},
-         {WS}
+//         {WS}
```
gemmellr, line 95: Leftover?

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionStateManager.java
```diff
@@ -0,0 +1,181 @@
...
+
+      for (String key : toRemove) {
+         try {
+            MQTTSessionState state = removeSessionState(key);
+            if (state != null && state.isWill() && !state.isAttached() && state.isFailed()) {
+               state.getSession().sendWillMessage();
+            }
+         } catch (Exception e) {
+            // TODO: make this a real error message
+            e.printStackTrace();
```
gemmellr, line 112: still TODO?

gemmellr:
The changes also look to have broken a couple of the compatibility tests:
>2023-08-21T19:22:09.1037015Z [ERROR] Errors: 
2023-08-21T19:22:09.1038656Z [ERROR]   JournalCompatibilityTest.testSendReceiveQueueMetrics:149->ClasspathBase.evaluate:196->ClasspathBase.tclCall:246->ClasspathBase.lambda$evaluate$2:199 » InvocationTarget
2023-08-21T19:22:09.1040561Z [ERROR]   JournalCompatibilityTest.testSendReceiveSizeQueueMetricsPaging:172->ClasspathBase.evaluate:196->ClasspathBase.tclCall:246->ClasspathBase.lambda$evaluate$2:199 » InvocationTarget

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionStateManager.java
jbertram: That `TODO` was copied from `MQTTProtocolManager`. It is still to-do, but not related to this PR.

jbertram:
Strangely enough, I need #4586 to be merged before the compatibility tests will be fixed.

artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionStateManager.java
gemmellr: There isnt a TODO in MQTTProtocolManager that I can see, and none was removed in the diff seemingly moving this code, so it appears to be getting introduced here along with the try-catch it is in.

jbertram: My apologies. I totally misremembered what happened here. I've resolved the issue. Thanks for verifying.

artemis-protocols/artemis-mqtt-protocol/src/test/java/org/apache/activemq/artemis/protocol/mqtt/SessionStateManagerTest.java
```diff
@@ -0,0 +1,151 @@
...
+               return false;
+         } else if (a.option().retainHandling().value() != b.option().retainHandling().value()) {
+            return false;
+         }
+         if (a.option().isRetainAsPublished() != b.option().isRetainAsPublished()) {
+            return false;
+         }
+         if (a.option().isNoLocal() != b.option().isNoLocal()) {
+            return false;
+         }
```
gemmellr, line 146: Its a bit weird mixing single-line-ifs with-braces and without-braces in the same method multiple times. With braces is nicer and safer. Some separating newlines would make it a lot more readable too.

(I find it bizarre the build enforces there is a space before the brace but also allows it not to be there hehe)

jbertram:
This should be ready to go.

artemis-protocols/artemis-mqtt-protocol/pom.xml
```diff
@@ -111,6 +111,22 @@
          <artifactId>log4j-slf4j-impl</artifactId>
          <scope>test</scope>
       </dependency>
+      <dependency>
+         <groupId>org.apache.activemq.tests</groupId>
+         <artifactId>artemis-test-support</artifactId>
+         <version>${project.version}</version>
+         <scope>test</scope>
+      </dependency>
```
gemmellr, line 119: I missed it previously but this dep is problematic. That module is aimed at for use within the tests/ subtree only currently and like all those module it isnt deployed, so we shouldnt be using it in modules that are being deployed.

I guess perhaps you used it as a proxy for artemis-server:test-jar due to prior discussion on that and eventually getting rid of the test-jar, but it doesnt really work in this case unfortunately.

Of course, it only seems to be getting used to have ActiveMQTestBase as a base class for 1 test, which it then uses to stand up a full VM server, so arguably the test should be in the tests/integration-tests (or stretching maybe tests/unit-tests) anyway rather than the mqtt-protocol module since its not really a classical unit test to start with. Then it could indeed depend on artemis-test-support as those bits already do.

That or change to the actual test-jar dep and add it to the cleanup list hehe.

gemmellr:
There also look to be some repeatable test failures in full suite runs (I ran one with the changes as-is, and another with the changes rebased on main to pick up some other test fixes from main).

artemis-protocols/artemis-mqtt-protocol/pom.xml
jbertram: This should be fixed now. I went in a different direction with this whole test.

jbertram:
I've converted this to draft while I investigate some failures on the test-suite.

jbertram:
IMO this is good to go. My latest test-suite run had a few failures, but I couldn't find any connection between those failures and my changes. Furthermore, all the tests that failed succeeded when run locally.

gemmellr:
Recent updates LGTM, just needs another test run due to the other big changes newly landed on main.

jbertram:
My latest run of the test-suite was clean. :+1:

PR #4584, opened 2023-08-21
ARTEMIS-4349: use Caffeine cache rather than Guava cache

gemmellr:
As it has been several weeks of the overall Caffeine related changes not progressing due to largely unrelated issues and despite some simple fixups that would let them, I decided to update PR #4540 myself and raise a new PR.

I have kept the original changes and my followups in separate commits here to make the alterations easier to see. We can squash all the updates upon PR completion before the push.

This now concentrates more on the simple 'adding Caffeine for caching' side the JIRA is actually for, and unwinds some of the earlier changes so as to fix the console build issue they introduced and also not breaking the overall [ErrorProne-enabled] build as they do (without other/additional changes like #4545 or #4567 first to partially-avoid the issues being introduced). In the end I also decided to skip introducing the classgraph dep or adjusting the related test, instead just having the test continue to use the Guava dep since that is still/already being used elsewhere in the build anyway.

The PR builds pass on these changes, and I ran a full test suite run on them. (EDIT: Turns out [there is a test the changes break](https://github.com/apache/activemq-artemis/pull/4584#issuecomment-1687832880)).

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
@@ -22,8 +22,6 @@ import java.security.AccessController;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
 import org.apache.activemq.artemis.api.core.Pair;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.api.core.management.CoreNotificationType;
@@ -54,6 +52,10 @@ import org.apache.activemq.artemis.utils.collections.ConcurrentHashSet;
 import org.apache.activemq.artemis.utils.collections.TypedProperties;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import com.github.benmanes.caffeine.cache.Cache;
+import com.github.benmanes.caffeine.cache.Caffeine;
+
 import java.lang.invoke.MethodHandles;
 
 /**
@@ -101,7 +103,7 @@ public class SecurityStoreImpl implements SecurityStore, HierarchicalRepositoryC
       if (authenticationCacheSize == 0) {
          authenticationCache = null;
       } else {
-         authenticationCache = CacheBuilder.newBuilder()
+         authenticationCache = Caffeine.newBuilder()
                                            .maximumSize(authenticationCacheSize)
                                            .expireAfterWrite(invalidationInterval, TimeUnit.MILLISECONDS)
                                            .build();
@@ -109,7 +111,7 @@ public class SecurityStoreImpl implements SecurityStore, HierarchicalRepositoryC
       if (authorizationCacheSize == 0) {
          authorizationCache = null;
       } else {
-         authorizationCache = CacheBuilder.newBuilder()
+         authorizationCache = Caffeine.newBuilder()
                                           .maximumSize(authorizationCacheSize)
                                           .expireAfterWrite(invalidationInterval, TimeUnit.MILLISECONDS)
                                           .build();
@@ -481,7 +483,7 @@ public class SecurityStoreImpl implements SecurityStore, HierarchicalRepositoryC
       if (authenticationCache == null) {
          return 0;
       } else {
-         return authenticationCache.size();
+         return authenticationCache.estimatedSize();
       }
    }
 
@@ -489,7 +491,7 @@ public class SecurityStoreImpl implements SecurityStore, HierarchicalRepositoryC
       if (authorizationCache == null) {
          return 0;
       } else {
-         return authorizationCache.size();
+         return authorizationCache.estimatedSize();
       }
    }
 
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
@@ -104,17 +104,17 @@ public class SecurityStoreImpl implements SecurityStore, HierarchicalRepositoryC
          authenticationCache = null;
       } else {
          authenticationCache = Caffeine.newBuilder()
-                                           .maximumSize(authenticationCacheSize)
-                                           .expireAfterWrite(invalidationInterval, TimeUnit.MILLISECONDS)
-                                           .build();
+                                       .maximumSize(authenticationCacheSize)
+                                       .expireAfterWrite(invalidationInterval, TimeUnit.MILLISECONDS)
+                                       .build();
       }
       if (authorizationCacheSize == 0) {
          authorizationCache = null;
       } else {
          authorizationCache = Caffeine.newBuilder()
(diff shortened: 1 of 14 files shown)


gemmellr:
This still isnt necesarily complete yet however, there is one other thing that I just noticed about the changes that is worth looking into first.

The removal of Guava from artemis-server has seemingly made these changes alter the resulting assembly more than simply adding the additional Caffeine dep, instead adding multiple other jars that were not present before. I expect this is as the Guava dependency in artemis-server previously also excluded various transitve deps from Guava, and the artemis-server dependency then dominated in most (but perhaps not all) of the other modules in the build, so the exclusions then applied there also. With removal of that dep from artemis-server the transitive Guava dependency from artemis-quorum-ri is instead surfacing as the new route-to-Guava (via curator-client) and as that module didnt have the same excludes it is presumably now bringing the further dependencies into artemis-distribution unlike before.

An additional change still to try out is thus whether making the same excludes (build/checker annotation related deps) for artemis-quorum-ri works ok, and if so making them to return the assembly to its previous state + just adding Caffeine as it seemed this would.

gemmellr:
I actually missed that test MQTTSecurityManagerTest.testSecurityManagerModifyClientIDAndStealConnection is failing consistently with these changes, and passing consistently without them. Need to investigate why, and whether that was the case originally or not on the other PR, unclear there was a complete run on the previous changes in their latest state.

gemmellr:
The latest push now addresses the [comment from yesterday](https://github.com/apache/activemq-artemis/pull/4584#issuecomment-1686688047) around the assembly ending up with various new additions besides just Caffeine. It now only adds Caffeine.

It does not address the consistent MQTTSecurityManagerTest.testSecurityManagerModifyClientIDAndStealConnection test failure noted above. That is still failing and needs investigation.

gemmellr:
I simply ran all the tests in our CI env, if you look you'll find 3 successive runs from the PR branches where it failed, and many other runs off main etc where it passed. Will look into it more tomorrow.

gemmellr:
I wasnt able to make MQTTSecurityManagerTest.testSecurityManagerModifyClientIDAndStealConnection fail locally. Justin was able to get it to fail once while putting his system under outside stress during the run. I have seen 8 consecutive failures of the test during full test suite runs of either this PR branch or the original PR branch in a CI env I use. IT has been passing on all the other test runs being done from main etc.

I made a further change to have Caffeine use an inline executor for its cleanup etc, and the test then passed on the first run. I'm going to do some more runs on it, but at the moment it does seem like the async cleanup is the likely reason for the test failure.

jbertram:
I wonder if we should detect when size = 0, set the cache to `null`, and check for `null` before writing to or reading from the cache. Then we wouldn't be wasting CPU on caching entries that are just going to be removed or attempting to get items from the cache that we know aren't there.

Thoughts?

gemmellr:
I had similar thoughts when making the latest update. I think it probably would be worth doing that, separately from switching to Caffeine (anyone wanting it for an old version could then pick it out later).

jbertram:
> I think it probably would be worth doing that, separately from switching to Caffeine (anyone wanting it for an old version could then pick it out later).

I'll do that via [ARTEMIS-4399](https://issues.apache.org/jira/browse/ARTEMIS-4399) which was just opened by coincidence.

gemmellr:
Haha, brilliant timing.

jbertram:
See #4589.

gemmellr:
> See #4589.

With this PR rebased after those changes went in, I retested these changes with and without the final commit making the eviction processing behave more like Guava's. As expected, both ways now pass the previously-failing test, now that the likely underlying issue of 'temporarily caching when max size is 0' issue was resolved by the other PR omitting the cache in that situation.

So I guess its just a case of deciding whether we want to retain the guava-style inline execution behaviour, or change to the Caffeine async execution default.

artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
```diff
@@ -101,15 +103,15 @@ public SecurityStoreImpl(final HierarchicalRepository<Set<Role>> securityReposit
       if (authenticationCacheSize == 0) {
          authenticationCache = null;
       } else {
-         authenticationCache = CacheBuilder.newBuilder()
+         authenticationCache = Caffeine.newBuilder()
                                            .maximumSize(authenticationCacheSize)
                                            .expireAfterWrite(invalidationInterval, TimeUnit.MILLISECONDS)
                                            .build();
```
jbertram: This is a bit of a nit-pick, but the lines should be re-aligned so all the `.` characters are in line vertically.

jbertram:
> So I guess its just a case of deciding whether we want to retain the guava-style inline execution behaviour, or change to the Caffeine async execution default.

Regarding the cache usage in `org.apache.activemq.artemis.core.security.impl.SecurityStoreImpl` my thoughts are:

1. The only use-case where it really matters is with size = 0 and that's been fixed that already via #4589.
2. It makes sense to optimize security since it is used basically every time a client connects.

Therefore, I'm in favor of using the default from Caffeine.

artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
gemmellr: Fair. That was probably result of rebasing the changes again on top of your earlier PR, and those lines not having changes on them.

gemmellr:
> > So I guess its just a case of deciding whether we want to retain the guava-style inline execution behaviour, or change to the Caffeine async execution default.
> 
> Regarding the cache usage in `org.apache.activemq.artemis.core.security.impl.SecurityStoreImpl` my thoughts are:
> 
>     1. The only use-case where it really matters is with size = 0 and that's been fixed that already via [ARTEMIS-4399 fix disabled authn/z cache #4589](https://github.com/apache/activemq-artemis/pull/4589).
> 
>     2. It makes sense to optimize security since it is used basically every time a client connects.
> 
> 
> Therefore, I'm in favor of using the default from Caffeine.

Fair enough, since the SecurityStoreImpl doesnt look to use any of the listener stuff so it probably isn't really much/any different. The other usage does use it though, so I want to look into whether its prepared to handle the new concurrency that could result from offloading to the FJ pool (though if not, it may also not be prepared to handle any existing concurrency that could occur already.)

gemmellr:
> Fair enough, since the SecurityStoreImpl doesnt look to use any of the listener stuff so it probably isn't really much/any different.

Backed out the change to SecurityStoreImpl to leave it with the Caffeine default async behaviour now.

> The other usage does use it though, so I want to look into whether its prepared to handle the new concurrency that could result from offloading to the FJ pool (though if not, it may also not be prepared to handle any existing concurrency that could occur already.)

Confirmed the other usage would not currently be safe for that (or the existing behaviour..we should address that separately before proceeding) so leaving that as is.

gemmellr:
I think these changes are now ready for review. Again, the commits are kept separate for now just to show what changed, we can combine them when pushing.

I prepared an initial combined commit as referenced above (at https://github.com/apache/activemq-artemis/compare/main...gemmellr:activemq-artemis:caffeine-B) that could be pushed to close the PRs.

jbertram:
LGTM! Nice work!

PR #4587, opened 2023-08-23
ARTEMIS-4398 - support configuring database using Broker Properties

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-4398

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
@@ -84,6 +84,7 @@ import org.apache.activemq.artemis.core.remoting.impl.invm.InVMConnectorFactory;
 import org.apache.activemq.artemis.core.remoting.impl.netty.NettyConnectorFactory;
 import org.apache.activemq.artemis.core.security.Role;
 import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
+import org.apache.activemq.artemis.core.server.ActiveMQMessageBundle;
 import org.apache.activemq.artemis.core.server.JournalType;
 import org.apache.activemq.artemis.core.server.NetworkHealthCheck;
 import org.apache.activemq.artemis.core.server.SecuritySettingPlugin;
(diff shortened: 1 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -773,6 +774,17 @@ public <T> T convert(Class<T> type, Object value) {
...
       }, TransformerConfiguration.class);
 
+      beanUtils.getConvertUtils().register(new Converter() {
+         @Override
+         public <T> T convert(Class<T> type, Object value) {
+            //we only care about DATABASE type as it is the only one used
+            if (StoreConfiguration.StoreType.DATABASE.toString().equals(value)) {
+               return (T) new DatabaseStorageConfiguration();
+            }
+            return (T) null;
```
gtully, line 784: I think throw some runtimeexception here - value must be DATABASE, otherwise the error on the next set will be hard to debug

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
```diff
@@ -1554,6 +1555,48 @@ public void testSystemPropKeyReplacement() throws Exception {
...
+      Assert.assertEquals(dsc.getJdbcLockExpirationMillis(), 321);
+      Assert.assertEquals(dsc.getJdbcLockRenewPeriodMillis(), 654);
+      Assert.assertEquals(dsc.getJdbcNetworkTimeout(), 987);
+      Assert.assertEquals(dsc.getJdbcPassword(), "pass");
+      Assert.assertEquals(dsc.getJdbcUser(), "user");
+      Assert.assertEquals(dsc.getNodeManagerStoreTableName(), "nmtn");
+      Assert.assertEquals(dsc.getPageStoreTableName(), "pstn");
+   }
+
    @Test
```
gtully, line 1600: a negative test where the value is wrong would be great. to see what error comes out. the error should be in the Status.

gtully (changes requested):
looks good, think all that is needed is a better error when incorrectly configured, eg:
storeConfiguration=FILE

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
andytaylor: Ive changed this

gtully (approved):

PR #4588, opened 2023-08-23
ARTEMIS-4400 artemis-cdi-client: artemis-unit-test-support in test scope

amarkevich:


gemmellr:
Oops. Seems like it was changed incorrectly in https://github.com/apache/activemq-artemis/commit/cce565e78f990d1969cadbf809ec89e1386f66ee, surprised I missed it.

This change should have a JIRA, please raise one and update the PR

PR #4589, opened 2023-08-23
ARTEMIS-4399 fix disabled authn/z cache

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
```diff
@@ -424,26 +432,71 @@ private void handleNoCacheLoginException(NoCacheLoginException e) {
...
+                                                              String password,
+                                                              RemotingConnection connection) {
+      if (authenticationCache == null) {
+         return null;
+      } else {
+         return authenticationCache.getIfPresent(createAuthenticationCacheKey(user, password, connection));
+      }
+   }
+
+   private void getAuthorizationCacheEntry(ConcurrentHashSet<SimpleString> set, String key) {
```
tabish121, line 454: Think this should be putAuthorizationCacheEntry

jbertram: Indeed!

PR #4590, opened 2023-08-24
ARTEMIS-4378 ignore address federation config if connection is config…

gtully:
…ured as pull, consumerWindowSize=0

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/address/FederatedAddress.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/address/FederatedAddress.java
@@ -17,6 +17,7 @@
 package org.apache.activemq.artemis.core.server.federation.address;
 
 import java.io.Serializable;
+import java.lang.invoke.MethodHandles;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/address/FederatedAddress.java
```diff
@@ -310,10 +312,15 @@ private boolean match(AddressInfo addressInfo) {
...
 
    private boolean match(SimpleString address, RoutingType routingType) {
-      //Currently only supporting Multicast currently.
+      // only supporting Multicast currently.
       if (RoutingType.ANYCAST.equals(routingType)) {
          return false;
       }
+      if (hasPullConnectionConfig) {
+         // multicast address federation has no local queue to trigger batch pull requests, a regular fast consumer with credit window is necessary
+         // otherwise the upstream would fill up and block.
```
tabish121, line 321: It is probably worth logging something here (debug level probably) that indicates that a match was rejected because pull consumer was configured and the core federation doesn't allow a way to have both address and queue federation when in pull mode.  At least then you have a breadcrumb to debug it when someone complains their address federation is broken.

gtully: thanks Tim, sure it won't hurt and could help. Can do the same for both config error clauses.

PR #4592, opened 2023-08-24
POC demonstrates successful PR checks even artemis-log-annotation-processor disabled

amarkevich:
Prepared as a proof of value https://github.com/apache/activemq-artemis/pull/4582 changes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1622,6 +1622,7 @@
                   <compilerArgs>
                      <arg>--add-exports=java.base/jdk.internal.misc=ALL-UNNAMED</arg>
                   </compilerArgs>
+                  <annotationProcessorPaths/>
                </configuration>
             </plugin>
             <plugin>


clebertsuconic:
everything is working. Log Processor worked in all the modules...

gemmellr:
As Clebert said..this actually didn't disable the log annotation processor, if you look you will find it still generating the classes. I'd guess there is some no-op checking in there somewhere (maven or the JDK) for the config such that your empty setting didnt count as 'specifying a processor' and so also didnt disable the 'if none specified, use found processors' behaviour.

amarkevich:
> everything is working. Log Processor worked in all the modules

Codegeneration broken In case profile activated via `-Derrorprone`

gemmellr:
The log bundle generation still occurs with your change in place even with -Derrorprone in use, I actually tried that earlier. Maybe you mixed things up with your other PRs which updated the processor config and did actually break the generation?

For example, give this change a rebase on main and run something like "mvn clean install -Derrorprone -Dtest=*LogBundle*Test -DfailIfNoTests=false" and you will see it run and pass just the new tests I added earlier which all rely on the generated code being there.

PR #4594, opened 2023-08-29
ARTEMIS-4405 wrong user logged for authz audits

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/StompAuditLoggingTest.java
```diff
@@ -0,0 +1,107 @@
...
+   public static void clearLogger() throws Exception {
+      try {
+         loggerHandler.close();
+      } finally {
+         AssertionLoggerHandler.setLevel(BASE_AUDIT_LOGGER_NAME, previousLevel);
+      }
+   }
+
+   @Test
+   public void testConcurrentLogging() throws Exception {
```
gemmellr, line 90: Not really clear from the test which bit is concurrent / with what, i.e what is the interesting bit of what the test is actually doing? Perhaps add a comment?

jbertram: My bad. The test name was copied from elsewhere. I'll rename it to actually reflect what it's doing.

PR #4595, opened 2023-08-29
ARTEMIS-4406: update LocalCache to use a ConcurrentHashMap

gemmellr:
Address thread safety issue noticed during #4584


rvais (commented):
methods "_onRemoval_" and "_put_" seem to access the original StorageManager implementation. Shouldn't that be protected against concurrency conflicts as well?

rvais (commented):
Misunderstanding resolved ...

gemmellr:
Actuallly...I had a further look at this after our discussion earlier, and though it appeared safe, it turns out there is indeed a related issue within AbstractJournalStorageManager class. https://github.com/apache/activemq-artemis/pull/4603 raised for that.

PR #4597, opened 2023-08-30
ARTEMIS-4410 - process delivery cancellations before removing [exclusive] consumer

gtully:
…deliveries

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
@@ -573,8 +573,6 @@ public class ServerConsumerImpl implements ServerConsumer, ReadyListener {
          del.finish();
       }
 
-      removeItself();
-
       List<MessageReference> refs = cancelRefs(failed, false, null);
 
       Transaction tx = new TransactionImpl(storageManager);
@@ -586,6 +584,9 @@ public class ServerConsumerImpl implements ServerConsumer, ReadyListener {
       });
 
       tx.rollback();
+
+      // started is false, leaving remove till after cancel ensures order for a single exclusive consumer
+      removeItself();
 
       addLingerRefs();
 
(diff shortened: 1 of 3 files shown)


gtully:
want to see what the test suite says and need some validation that I have not missed any countDown! will remove draft status once validate a full test suite run. but any eyes in the mean time are welcome :-)

gtully:
full test suite looks good.

gtully:
I think I see a logic error, investigating - back to draft

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -1434,6 +1436,10 @@ private boolean canDispatch() {
             return false;
          }
 
+         if (isExclusive() && hasExclusivePendingDeliveries()) {
+            return false;
+         }
```
gemmellr, line 1441: Doesnt this mean it will only send 1 outstanding message at a time? That doesnt seem desirable?

gemmellr (commented):
I'm not sure tracking every outstanding message is he way to go. Seems potentially brittle over time. The issue described seems to occur when closing one consumer and letting another begin sending before the cleanup for the old one is really done. There is already some logic in there to treat the exclusive consumer as a group consumer. If so, isnt there a way that the existing consumer can be made inactive but left treated as the group consumer until its cleanup is actually complete, thus preventing any others from dispatching until then?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gemmellr: (Sometimes maybe, but overall no due to the earlier-and-later 'remember can dispatch' logic I had overlooked)

gtully:
the canDispatch gate logic looks brittle, a simpler solution as suggested by @gemmellr is to leave the exclusive consumer in place as BUSY till it remove context has completed. Will investigate a fix with that approach.

gtully:
good observation.
there are only two calls to org.apache.activemq.artemis.core.server.impl.ServerConsumerImpl#cancelRefs one in a session where the started flags is flipped for this same ordering issue.
the other is the close (in the connection error case), where i see started is already flipped.
I think it may be a case of just deferring the remove till after the cancel in the close.
when !started a server consumer reports BUSY!

clebertsuconic:
the fix I see now seems very simple. if test suite is good, we should merge it.

PR #4599, opened 2023-08-31
ARTEMIS-3057 Add min-disk-free feature

jbertram:
This is a follow-up to #3812 with the following changes:

 - Resolved conflicts.
 - Eliminated a lot of code duplication.
 - Simplified `org.apache.activemq.artemis.core.server.files.FileStoreMonitor.Callback` and its support for multiple settings.
 - I didn't bother with deprecating anything since these are internal classes.
 - Added documentation.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
@@ -3510,12 +3510,25 @@ public class ActiveMQServerImpl implements ActiveMQServer {
          recoverStoredBridges();
       }
 
-      if (configuration.getMaxDiskUsage() != -1) {
-         try {
-            injectMonitor(new FileStoreMonitor(getScheduledPool(), executorFactory.getExecutor(), configuration.getDiskScanPeriod(), TimeUnit.MILLISECONDS, configuration.getMaxDiskUsage() / 100f, ioCriticalErrorListener));
-         } catch (Exception e) {
-            ActiveMQServerLogger.LOGGER.unableToInjectMonitor(e);
+      deployFileStoreMonitor();
+   }
+
+   private void deployFileStoreMonitor() throws Exception {
+      FileStoreMonitor.FileStoreMonitorType fileStoreMonitorType = null;
+      Number referenceValue = null;
+      if (configuration.getMinDiskFree() != -1) {
+         if (configuration.getMaxDiskUsage() != -1) {
+            ActiveMQServerLogger.LOGGER.configParamOverride(FileConfigurationParser.MIN_DISK_FREE, configuration.getMinDiskFree(), FileConfigurationParser.MAX_DISK_USAGE, configuration.getMaxDiskUsage());
          }
+         fileStoreMonitorType = FileStoreMonitor.FileStoreMonitorType.MinDiskFree;
+         referenceValue = configuration.getMinDiskFree();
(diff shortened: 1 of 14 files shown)


artemis-server/src/test/java/org/apache/activemq/artemis/core/server/files/FileStoreMonitorTest.java
```diff
@@ -72,71 +72,117 @@ public void testSimpleTick() throws Exception {
...
+      BufferedOutputStream bufferedOutputStream = new BufferedOutputStream(garbage);
+      PrintStream out = new PrintStream(bufferedOutputStream);
+
+      // This is just to make sure there is at least something on the device.
+      // If the testsuite is running with an empty tempFS, it would return 0 and the assertion would fail.
+      for (int i = 0; i < 100; i++) {
+         out.println("Garbage " + i);
+      }
+
+      bufferedOutputStream.close();
```
gemmellr, line 120: Could this use try-with-resources to eliminate the need to explicitly close, and make it safer?

artemis-server/src/test/java/org/apache/activemq/artemis/core/server/files/FileStoreMonitorTest.java
```diff
@@ -72,71 +72,117 @@ public void testSimpleTick() throws Exception {
...
+      FileStoreMonitor.Callback callback = (usableSpace, totalSpace, ok, type) -> {
+         tick.incrementAndGet();
+         if (ok) {
+            underMaxUsage.incrementAndGet();
+         } else {
+            overMaxUsage.incrementAndGet();
          }
+         logger.debug("tick:: usableSpace: " + usableSpace + ", totalSpace:" + totalSpace);
+      };
+      FileStoreMonitor storeMonitor = new FileStoreMonitor(scheduledExecutorService, executorService, 100L, TimeUnit.MILLISECONDS, 0.999, null, FileStoreMonitor.FileStoreMonitorType.MaxDiskUsage);
```
gemmellr, line 88: Importing FileStoreMonitor.FileStoreMonitorType could make things a little less verbose.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -3493,12 +3493,25 @@ public void run() {
...
-            ActiveMQServerLogger.LOGGER.unableToInjectMonitor(e);
+      deployFileStoreMonitor();
+   }
+
+   private void deployFileStoreMonitor() throws Exception {
+      FileStoreMonitor.FileStoreMonitorType fileStoreMonitorType = null;
+      Number referenceValue = null;
+      if (configuration.getMinDiskFree() != -1) {
+         if (configuration.getMaxDiskUsage() != -1) {
+            ActiveMQServerLogger.LOGGER.configParamOverride(FileConfigurationParser.MIN_DISK_FREE, configuration.getMinDiskFree(), FileConfigurationParser.MAX_DISK_USAGE, configuration.getMaxDiskUsage());
```
gemmellr, line 3504: This will log a warning, however its clearly documented that minDiskFree takes precedence, and it is already disabled by default in both code and xml config, whereas in contrast maxDiskUsage is enabled by default in both code and xml config. So to avoid a log warning users will *need* to explicitly configure both of them. I'm not sure thats necessary. At the very least it seems more like info or lower level than a warning if we keep it logging.

jbertram:
This should be good to go now. Thanks for the review @gemmellr.

PR #4600, opened 2023-08-31
ARTEMIS-4413 Improve reliability of CheckTest.testNodeCheckTopology

brusdev:

PR #4601, opened 2023-08-31
ARTEMIS-4413 Improve reliability of CheckTest.testNodeCheckTopology

clebertsuconic:


clebertsuconic:
@brusdev , @gemmellr ... as we talked about this ^^^

clebertsuconic:
this supersedes #4600

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/check/QueueCheck.java
```diff
@@ -179,7 +179,7 @@ private void checkQueueConsume(final CheckContext context) throws Exception {
          connection.start();
 
          if (consume == -1) {
-            queueConsumer.receiveNoWait();
+            System.out.println("REceived " + queueConsumer.receiveNoWait());
```
gemmellr, line 182: Typo.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/checkTest/CheckTest.java
```diff
@@ -0,0 +1,350 @@
...
+      queueCheck.setPassword("admin");
+      queueCheck.setName(queueName);
+      queueCheck.setConsume(null);
+      Assert.assertEquals(1, queueCheck.execute(ACTION_CONTEXT));
+
+
+      try (Connection connection = cf.createConnection(); Session session = connection.createSession(true, Session.SESSION_TRANSACTED)) {
+         MessageProducer producer = session.createProducer(session.createQueue(queueName));
+         for (int i = 0; i < messages; i++) {
+            System.out.println("Sending " + i);
```
gemmellr, line 281: Logger...debug?

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/checkTest/CheckTest.java
```diff
@@ -0,0 +1,350 @@
...
+      cleanupData(SERVER_NAME_1);
+      cleanupData(SERVER_NAME_2);
+
+      File location = new File(getServerLocation(SERVER_NAME_1));
+      deleteDirectory(new File(location, "../data"));
+
+      disableCheckThread();
+   }
+
+   @Test
```
gemmellr, line 74: Might be good to add a timeout to the tests.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/checkTest/CheckTest.java
```diff
@@ -0,0 +1,350 @@
...
+      try (Connection connection = cf.createConnection(); Session session = connection.createSession(true, Session.SESSION_TRANSACTED)) {
+         MessageProducer producer = session.createProducer(session.createQueue(queueName));
+         for (int i = 0; i < messages; i++) {
+            System.out.println("Sending " + i);
+            TextMessage message = session.createTextMessage("hello " + i);
+            message.setStringProperty("local", String.valueOf(i));
+            producer.send(message);
+         }
+         session.commit();
+         Wait.assertEquals(messages, () -> getMessageCount(simpleManagement, queueName));
```
gemmellr, line 287: Using the default 30 seconds of wait with 100ms periods seems excessive here....and everywhere below.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -272,7 +272,7 @@ public class Create extends InstallAbstract {
    @Option(names = "--jdbc", description = "Store message data in JDBC instead of local files.")
    boolean jdbc;
 
-   @Option(names = "--staticCluster", description = "Cluster node connectors list separated by comma, e.g. \"tcp://server:61616,tcp://server2:61616,tcp://server3:61616\".")
+   @Option(names = {"--staticCluster", "--static-cluster"}, description = "Cluster node connectors list separated by comma, e.g. \"tcp://server:61616,tcp://server2:61616,tcp://server3:61616\".")
```
gemmellr, line 275: Feels like it should be a seperate change, has little directly to do with moving+improving this test, and folks are less likely to know it was added if buried in this commit.

tests/smoke-tests/pom.xml
```diff
@@ -1199,7 +1199,7 @@
                         <arg>--name</arg>
                         <arg>cluster1</arg>
                         <arg>--clustered</arg>
-                        <arg>--staticCluster</arg>
+                        <arg>--static-cluster</arg>
```
gemmellr, line 1202: Similarly seems like its unrelated - not even for the test in question. Would also likely be clearer if separated out, why this one is being changed and the other server for the test isnt, that they are synonym options.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/check/QueueCheck.java
clebertsuconic: it is a debug leftover.. thanks.. removing it.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/checkTest/CheckTest.java
clebertsuconic: It's never happening though. I like the large timeout for debugging. 

I've tried to change the default in the past.. but I don't think we need to individualize things here.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
gemmellr: It could still have been in the same PR if needed but its clearly a seperate change that should have a commit+JIRA of its own rather than going in with the otherwise unrelated test-only changes.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/checkTest/CheckTest.java
gemmellr: Its never happening until it does. Thats the point of adding a timeout. Checks should have action-appropriate timeouts so things fail in a timely manner when issues do arise and not taking e.g 30sec for no good reason. Even one 100ms delay if it isnt ready on the very first check is likely wasting time really.

(E.g, remember that AIO check thing? Until I added the fail-fast bailou , whenever that issue arose it used to take hours for the test job to actually fail, as it sat waiting unnecessarily for a full 30sec on every check).

clebertsuconic: These are all simple checks with message counts.. I don't think we need to do much more about this.

clebertsuconic: @gemmellr the long default on Wait is a general concern. we should rather change that default. I don't think we should make it different here.

I have tried before but it was a large task that I postponed. I would merge this as is and tackle the large timeout eventually. 


I think the default timeout should be 1 second (5 seconds, the very largest). The main idea of Wait was async acts and sends on queue.. so a second should be enough.. anything other than that should have a separate method call. We should do the opposite here.

gemmellr: Precisely, they are just simple checks, expected to pass immediately or almost...so giving them 30 seconds does not seem like an action-appropriate timeout.

Similarly, if somehow they didnt pass on the very first check, burning 100ms at a time until they do is potentially a significant waste of time sat doing nothing. The overall test suite is damn slow, we really should be paying attention to optimising such things.

gemmellr: I would personally remove the default wait entirely and have checks set an appropriate timeout and period for whatever they are doing.

clebertsuconic: This is not part of this change though... These sleeps were already part of the previous test.

I will add a timeout, but it shouldn't be subject of this task now.

clebertsuconic: The default timeout makes it very simple to convert an Assert.assertEquals into an Wait.assertEquals.

However the 30seconds wait timeout is a mistake that we should fix.


that's a separate change.. I added the timeout now.

gemmellr: > This is not part of this change though... These sleeps were already part of the previous test.

It seems like you added the waits here, in which case yes its subject to this task.

> The default timeout makes it very simple to convert an Assert.assertEquals into an Wait.assertEquals.

Maybe, but people shouldnt really be adding timeouts without looking at what the action-appropriate timeout and retry interval actually is.

clebertsuconic:
I am merging this now... anything extra now could be sent later.

PR #4602, opened 2023-09-01
ARTEMIS-4415 accept indeterminism of broker dispatch

gtully:


gtully:
make me wonder what is actually being verified, at the point of clientsession start, it is not possible to know the state of the broker unless we wait for some metric, maybe wait to count == 2. 
It was breaking after a few thousand iterations for me and now is not, but to some end, it is now verifying two things!

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/LVQTest.java
```diff
@@ -167,11 +167,13 @@ public void testMultipleMessages() throws Exception {
       ClientMessage m = consumer.receive(1000);
       Assert.assertNotNull(m);
       m.acknowledge();
-      Assert.assertEquals(m.getBodyBuffer().readString(), "m3");
+      String val = m.getBodyBuffer().readString();
+      Assert.assertTrue("1 or 3 =? " + val, "m1".equals(val) || "m3".equals(val));
```
gemmellr, line 171: This doesnt seem like a desirable change for an LVQ test. It would then be about as well deleted it for all the good it would really be doing in verifying the 'LV' bit.

It currently creates the consumer before sending so it seems rather pot-luck what it will see. The next test creates it after, and also checks the count as you said. Or it could use a transacted producer for the 'second round' of sends. Something synchronous between send and check, basically.

gtully: I think the crux is that the broker decides what is the LV, and it is non deterministic in this case.

gemmellr: It is currently but per your comment and other tests it could probably be made deterministic. Instead, with this change the test wouldnt really be verifying anything much at all.

clebertsuconic: @gtully can't you instead make the test deterministic? like keeping consumers and senders on a certain condition that's required for the test?


I had this issue in other tests... where I needed to verify the bindings were in place before asserting.. or the consumers were ready to receive with credits.. etc...

jbertram:
I believe this was fixed via da861bd6fd00bef0b96cc4be42c800099bcd3b4c and 73ca0021af1616509d9ac1b77c3c439839feb7ef which made the test deterministic since it now waits for the expected message count, messages added, and messages replaced before attempting to consume. I ran the test over 8,000 times without a failure.

PR #4603, opened 2023-09-05
ARTEMIS-4417: use ConcurrentHashMap for thread safety

gemmellr:
The insert + remove methods in AbstractJournalStorageManager that can be used by the connection router LocalCache are not thread safe. They operate on a 'parent' ConcurrentHashMap, but entries within it are regular HashMap values and so are not thread safe despite potentially being used concurrently.

Switch the sub-maps to ConcurrentHashMap accordingly and also reuse some identical insertion code to simplify.


rvais (commented):
Legit!

PR #4604, opened 2023-09-05
ARTEMIS-4418 limit use of openwire consumer close lastSequenceId to e…

gtully:
…xclusive queues, where order is guarenteed


gtully:
there may be another way to look at this, setting the brokerSequenceId based on a consumer delivery sequence id and tracking that. Such that the queue order is not relevant. That is what the amqp protocol head does. Will investigate. Thanks @gemmellr

PR #4605, opened 2023-09-05
ARTEMIS-4419 Add federation support to AMQP broker connections

tabish121:
Allows federation of addresses and queues over an outbound AMQP broker connection and provide configuration via XML or broker propeties.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/config/FederationConfiguration.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/config/FederationConfiguration.java
@@ -73,8 +73,7 @@ public class FederationConfiguration implements Serializable {
       return federationPolicyMap;
    }
 
-   // strange spelling!, it allows a type match for singular of correct plural Policies from properties
-   public FederationConfiguration addQueuePolicie(FederationQueuePolicyConfiguration federationPolicy) {
+   public FederationConfiguration addQueuePolicy(FederationQueuePolicyConfiguration federationPolicy) {
       federationPolicyMap.put(federationPolicy.getName(), federationPolicy);
       return this;
    }
@@ -83,7 +82,7 @@ public class FederationConfiguration implements Serializable {
       return federationPolicyMap;
    }
 
-   public FederationConfiguration addAddressPolicie(FederationAddressPolicyConfiguration federationPolicy) {
+   public FederationConfiguration addAddressPolicy(FederationAddressPolicyConfiguration federationPolicy) {
       federationPolicyMap.put(federationPolicy.getName(), federationPolicy);
       return this;
    }
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
@@ -22,12 +22,15 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Predicate;
 import java.util.stream.Stream;
 
 import org.apache.activemq.artemis.api.core.ActiveMQException;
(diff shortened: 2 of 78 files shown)


jbertram:
This looks good!

It's not immediately clear to me how this compares to the existing federation functionality. Can you clarify that? Also, when would one choose one over the other? Does this make the existing federation obsolete?

tabish121:
> This looks good!
> 
> It's not immediately clear to me how this compares to the existing federation functionality. Can you clarify that? Also, when would one choose one over the other? Does this make the existing federation obsolete?

This implementation isn't itself intended to be a replacement of the core based federation implementation as that would likely result in breakage of existing users of that federation which wouldn't fly in a minor version update.  This implementation exists on the AMQP broker connection feature to avoid breaking those existing users.  

Folks doing AMQP messaging might chose this implementation as it would then result in AMQP messages remaining in that form throughout their lifetime avoiding conversions to / from Core as they cross the core based federation consumer links.  

Another reason would be that unlike the core based implementation this implementation supports two way federation or simply "downstream" federation over the single outgoing connection avoiding the need to configure connection details for the remote to connect back to the local server which is how the core based version works for "downstream" configurations (which might not be possible in cloud based brokers, or might not be allowed due to firewall rules etc).

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
```diff
@@ -239,7 +271,44 @@ SimpleString getMirrorSNF(AMQPMirrorBrokerConnectionElement mirrorElement) {
...
+    * remote link closed handling should be ignored.
+    *
+    * @param id
+    *    A unique Id value that identifies the intercepter for later removal.
+    * @param intercepter
+    *    The predicate that will be called for any link close.
+    *
+    * @return this broker connection instance.
+    */
+   public AMQPBrokerConnection addLinkClosedIntercepter(String id, Predicate<Link> intercepter) {
```
gemmellr, line 286: Intercepter -> Interceptor for consistency with other areas?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/federation/AMQPFederationPolicySupport.java
```diff
@@ -0,0 +1,535 @@
...
+
+      if (element.getIncludes() != null && !element.getIncludes().isEmpty()) {
+         includes = new HashSet<>(element.getIncludes().size());
+
+         element.getIncludes().forEach(addressMatch -> includes.add(addressMatch.getAddressMatch()));
+      } else {
+         includes = Collections.EMPTY_SET;
+      }
+
+      if (element.getExcludes() != null && !element.getIncludes().isEmpty()) {
```
gemmellr, line 457: c&p error? null-checks excludes but then empty-checks includes.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/federation/AMQPFederationPolicySupport.java
```diff
@@ -0,0 +1,535 @@
...
+      if (element.getIncludes() != null && !element.getIncludes().isEmpty()) {
+         includes = new HashSet<>(element.getIncludes().size());
+
+         element.getIncludes().forEach(queueMatch ->
+            includes.add(new AbstractMap.SimpleImmutableEntry<String, String>(queueMatch.getAddressMatch(), queueMatch.getQueueMatch())));
+      } else {
+         includes = Collections.EMPTY_SET;
+      }
+
+      if (element.getExcludes() != null && !element.getIncludes().isEmpty()) {
```
gemmellr, line 511: ditto

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/config/FederationConfiguration.java
```diff
@@ -74,7 +74,7 @@ public Map<String, FederationPolicy> getQueuePolicies() {
    }
 
    // strange spelling!, it allows a type match for singular of correct plural Policies from properties
-   public FederationConfiguration addQueuePolicie(FederationQueuePolicyConfiguration federationPolicy) {
+   public FederationConfiguration addQueuePolicy(FederationQueuePolicyConfiguration federationPolicy) {
```
gemmellr, line 77: Comment above method is inaccurate after the name changes.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
tabish121: fixed

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/federation/AMQPFederationPolicySupport.java
tabish121: fixed

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/config/FederationConfiguration.java
tabish121: removed the old comment

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/federation/AMQPFederationPolicySupport.java
tabish121: fixed

PR #4606, opened 2023-09-06
ARTEMIS-4418 use consumer delivery sequence in place message id for o…

gtully:
…penwire broker sequence id, makes delivery ccount calculation independent of message order

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
@@ -27,6 +27,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 
@@ -81,6 +82,7 @@ public class AMQConsumer {
    private boolean internalAddress = false;
    private volatile Set<MessageReference> rolledbackMessageRefs;
    private ScheduledFuture<?> delayedDispatchPrompter;
+   private AtomicLong deliveredSequenceId = new AtomicLong(0);
 
    public AMQConsumer(AMQSession amqSession,
                       org.apache.activemq.command.ActiveMQDestination d,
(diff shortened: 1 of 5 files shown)


artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
```diff
@@ -81,6 +81,7 @@ public class AMQConsumer {
    private boolean internalAddress = false;
    private volatile Set<MessageReference> rolledbackMessageRefs;
    private ScheduledFuture<?> delayedDispatchPrompter;
+   private long deliveredSequenceId = 0;
```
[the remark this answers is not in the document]
gtully: correction, thread safe for the increment, so atomic

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
```diff
@@ -293,6 +295,7 @@ public int handleDeliver(MessageReference reference, ICoreMessage message) {
          }
          //handleDeliver is performed by an executor (see JBPAPP-6030): any AMQConsumer can share the session.wireFormat()
          dispatch = OpenWireMessageConverter.createMessageDispatch(reference, message, session.wireFormat(), this, session.getCoreServer().getNodeManager().getUUID());
+         dispatch.getMessage().getMessageId().setBrokerSequenceId(deliveredSequenceId.getAndAdd(1));
```
gemmellr, line 298: The createMessageDispatch method already looks to set this, to the existing different value...perhaps it should be changed to pass this value in there to set, helping avoid any inconsistency later.

Also getAndAdd - > getAndIncrement ?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
```diff
@@ -656,6 +656,7 @@ public void removeItself() throws Exception {
          browserDeliverer.close();
       } else {
          messageQueue.removeConsumer(this);
+         messageQueue.deliverAsync();
```
gemmellr, line 659: Kind of wonder if it shouldnt be the queue doing this instead of the consumer? Also feels like it may be a separate change really.

gtully: it could be but the test that now has an exclusive and non exclusive variant was  intermittently failing and this was the reason.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
gtully: good idea, there will be more impact by that api change but it is better. thanks

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerConsumerImpl.java
gemmellr: sure, but that seems more like a bug in the queue itself, and this change isnt specific to openwire, should have been addressed seperately from this openwire-consumer-specific JIRA.

PR #4610, opened 2023-09-11
ARTEMIS-4424 Unecessary AMQ212025 (not connected) when no nodes are c…

clebertsuconic:
…onnected

The exception thrown by serverLocator.connect() should be all you need on such case and the caller should then be responsible for taking appropriate action.


clebertsuconic:
this PR contains a trivial fix. I will merge it as soon as I have some tests done.

clebertsuconic:
It currently also contains the JDBC fix with it.. which will be removed of course before merging. (I'm keeping both commits here now just for testing).

clebertsuconic:
I'm done with tests.. just waiting CI from the PR.

PR #4613, opened 2023-09-12
ARTEMIS-4428 Expand default loggers configuration

brusdev:


gemmellr:
This says 'Expand default loggers configuration'...but it has actually significantly _changed_ the configuration, turning off all non o.a.activemq loggers entirely, as opposed to them being at INFO previously. The change or the JIRA is misleading.

gemmellr:
^ See #4616

PR #4615, opened 2023-09-12
ARTEMIS-4397 Fixing the upgrade command

clebertsuconic:


clebertsuconic:
@brusdev I need your help on this. 
I don't understand the context on why you added regular expressions to the upgrade.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/upgradeTest/UpgradeCurrentVersion.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.apache.activemq.artemis.tests.smoke.common.SmokeTestBase;
+import org.apache.activemq.artemis.tests.util.CFUtil;
+import org.apache.activemq.artemis.util.ServerUtil;
+import org.apache.activemq.artemis.utils.RandomUtil;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+
+/** This test is making sure the upgrade command would be able to upgrade a test I created with artemis 2.25.0 */
```
gemmellr, line 36: Comment is wrong, remove? (It applied to the original test, but not this one).

tests/smoke-tests/pom.xml
```diff
@@ -1318,6 +1318,31 @@
...
+                  <phase>test-compile</phase>
+                  <id>upgrade-current-version-phase2</id>
+                  <goals>
+                     <goal>upgrade</goal>
+                  </goals>
+                  <configuration>
+                     <instance>${basedir}/target/upgrade/currentVersion</instance>
+                  </configuration>
+               </execution>
+
```
gemmellr, line 1345: This is being inserted in the middle of existing executions used together for a test, separating them...moving this addition up (e.g line 1306) or down (e.g. line 1379) would make it more understandable later.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/upgradeTest/UpgradeCurrentVersion.java
```diff
@@ -0,0 +1,70 @@
...
+   @Before
+   public void beforeTest() throws Exception {
+      upgradedServer = new File(basedir + "/target/upgrade/currentVersion");
+
+      processServer = ServerUtil.startServer(upgradedServer.getAbsolutePath(), "upgradedServer", 0, 5000);
+      addProcess(processServer);
+   }
+
+   @Test
+   public void testSimpleSendReceive() throws Throwable {
```
gemmellr, line 52: I think a likely better and faster test would be to comparable to / part-of the existing CompareUpgradeTest. We already have a current-expected-instance-config for those existing tests to compare with their old-upgraded-instance output to check for match. We can compare a new current-version-upgraded instance config with that and check they match in much the same fashion, whereas this test wouldnt notice if it did change something unless that actually breaks the server.

brusdev:
> @brusdev I need your help on this. I don't understand the context on why you added regular expressions to the upgrade.

@clebertsuconic I used regex mainly because the `uri` attribute of the `binding` element depends on the `-http-host` option of the `artemis create` command, i.e. the command `artemis create --http-host test ...` generates a bootstrap.xml with the following content
```
   <web path="web" rootRedirectLocation="console">
       <binding name="artemis" uri="http://test:8161">
```

brusdev:
@clebertsuconic I have just pushed a commit to revert and fix the upgrade regex
https://github.com/apache/activemq-artemis/pull/4615/commits/76695e4b2f10f8cf1bc6cf979e6f4936974dc4b7

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/upgradeTest/UpgradeCurrentVersion.java
gemmellr: though the other test uses a split-config and so one would need a rework to use the same comparison config for both.

clebertsuconic: @gemmellr I just want this test to validate if the server works after a self upgrade. (Which could be used to change home... etc...)

clebertsuconic:
@gemmellr , @brusdev I think this is ready to be merged. We could add another test.. but if we keep doing the compare tests. 

Further tests could be added later.. but I think what we have now is enough.

brusdev (approved):

gemmellr:
I think a system test for this is unnecessary (we have lots of tests using the default config already), far slower, and wont actually catch future unexpected changes that happen to not down the broker like this one did.

clebertsuconic:
create a server, upgrade and start the server is actually how you verify this issue.

if you want to do the hard way (compare every single file) it will take longer to develop, and you still need to keep the system test.


As a matter of fact I'm almost done with a compare test.. it took me a couple hours to do it... (probably 3 or 4) which is way more than this test takes to execute (1 second each time).  and I'm probably keeping the system test as well to validate future issues during the upgrade. I think this is a valid test.

gemmellr:
> create a server, upgrade and start the server is actually how you verify this issue.
> 

The code needing tested is in neither the broker or client that test uses, the broker wasnt involved at all in the bug or the fix, so I disagree thats how it should actually be tested.

The broker could still start even if a multitude of other highly similar/essentially-identical issues had occurred in the upgrade process. Using the broker+client to 'test' this by side-effect is both slow and gives a poor result. Spending 1 second (or whatever it actually is) every time the test suite is run in future while testing this type of issue, vs the comparitively much smaller amount of closely verifying the files / cli-handling, is essentially exactly how over time the test suite ends up in the state it is in now.

> if you want to do the hard way (compare every single file) it will take longer to develop, and you still need to keep the system test.

I wouldnt call it the hard way, but test development is often a time tradeoff yes. Development time now vs output quality and maintenance burden later. More time developing a good test, that runs fast, is reliable, is targeted, gives good failure messages, catches the entire general issue rather than just 1 variant that had a no-startup side effect...vs later having years of wasting time on every execution of slower poorer tests that verify by side effect, wont catch various near-identical related issues, dont give as nice a failure if they should fail, perhaps even fail sporadically....and which actually a lot of time people dont even bother to run precisely because the complete test suite is so slow and unreliable overall, which I know happens fairly regularly from the number of times I encounter it while reviewing PRs.

> 
> As a matter of fact I'm almost done with a compare test.. it took me a couple hours to do it... (probably 3 or 4) which is way more than this test takes to execute (1 second each time). and I'm probably keeping the system test as well to validate future issues during the upgrade. I think this is a valid test.

It might catch _this_ issue as a complete side effect, but it wont catch other almost identical issues, which the other test will have caught already. Its just a significantly poorer test overall relative to the alternative of checking the actual CLI behaviour to verify the CLI behaviour. We already have other tests starting the broker up with the same 'default config' the upgrade test should be verifying is what is still present after such a 'non-upgrade'.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/FileUtil.java
```diff
@@ -72,4 +82,27 @@ public static final boolean deleteDirectory(final File directory) {
...
+
+            @Override
+            public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException {
+               Files.copy(file, targetPath.resolve(sourcePath.relativize(file)), StandardCopyOption.REPLACE_EXISTING);
+               return FileVisitResult.CONTINUE;
+            }
+         });
+      } catch (IOException e) {
+         e.printStackTrace();
+      }
```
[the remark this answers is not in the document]
clebertsuconic: Yes.... it should be taken out

PR #4617, opened 2023-09-14
ARTEMIS-4270 Allow hierarchy of wildcard addresses

essobedo:
In case the bindings "news.#" and "news.europe.#" are registered, only the first one matches with the address "news.europe" while both are supposed to match.


essobedo:
@gtully Hi 👋 since your are the code owner what I modified, maybe you could tell if it makes sense or not, thank you in advance 🙏

gtully:
That use case seems correct. I just wonder why I had that canvisit=false logic in there?

If our test suite is happy this looks good

essobedo:
TBH it was not clear to me too that's why I'm asking

essobedo:
Did I miss a step preventing the PR to be reviewed?

jbertram:
> Did I miss a step preventing the PR from being reviewed?

@essobedo, no I don't think so. Why do you ask?

I ran the test-suite and there was one failure, but it looks completely unrelated and it was failing intermittently before this change.

jbertram:
@essobedo, thanks for this fix! Very nice work!

clebertsuconic:
nice fix indeed!

PR #4619, opened 2023-09-18
ARTEMIS-4432 - respect actor and operation context for openwire conne…

gtully:
…ction failure processing

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
@@ -681,6 +681,24 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
       return this.inWireFormat;
    }
 
+   private void rollbackInProgressLocalTransactions() {
+
+      for (Transaction tx : txMap.values()) {
+         AMQSession session = (AMQSession) tx.getProtocolData();
+         if (session != null) {
+            session.getCoreSession().resetTX(tx);
+            try {
+               session.getCoreSession().rollback(false);
+            } catch (Exception expectedOnExistingOutcome) {
+            } finally {
+               session.getCoreSession().resetTX(null);
+            }
+         } else {
+            tx.tryRollback();
+         }
+      }
+   }
+
    private void shutdown(boolean fail) {
 
       try {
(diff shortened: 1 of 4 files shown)


gtully:
of note is the shutdown behaviour of the actor, on failure of the connection. We don't want any pending tasks executed that could block the cleanup task, but I want the cleanup task executed by the actor ordered executor.
What I don't know if that  behaviour is particular to openwire of can be more generally applied?

gtully:
all good with the full test run.

gtully:
removed the system.out in the test

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -832,7 +863,16 @@ public AMQConnectionContext initContext(ConnectionInfo info) throws Exception {
       createInternalSession(info);
 
       // the actor can only be used after the WireFormat has been initialized with versioning
-      this.openWireActor = new ThresholdActor<>(executor, this::act, actorThresholdBytes, OpenWireConnection::getSize, this::disableAutoRead, this::flushedActor);
+      this.openWireActor = new ThresholdActor<>(executor, this::act, actorThresholdBytes, OpenWireConnection::getSize, this::disableAutoRead, this::flushedActor) {
+         @Override
+         protected void doTask(Object task) {
```
clebertsuconic, line 868: I would place this logic on ThresholdActor.

Set something when shutdown is called, and have the doTask checking if (task==shutdownTask), it would do it right away.

The ThresholdActor is being done for OpenWire pretty much, and I think that logic would apply to the super class anyways.

clebertsuconic: basically.. just keep doTask a final, and bring that logic up there. I think it would be clearer that way!?

gtully: fair enough, as openwire is the only user of that class. on it. thanks for the review

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/PrefetchRedeliveryCountOpenwireTest.java
```diff
@@ -176,4 +192,119 @@ public void testExclusiveConsumerOrderOnReconnectionLargePrefetch() throws Excep
          }
       }
    }
+
+   @Test(timeout = 600_000)
+   public void testExclusiveConsumerTransactionalBatchOnReconnectionLargePrefetch() throws Exception {
```
brusdev, line 197: testExclusiveConsumerTransactionalBatchOnReconnectionLargePrefetch passes even without the changes

gtully: darn, good catch, I now see that the assertion is being caught! to spit out the paging data and the error is lost. Will fix. thanks!

PR #4622, opened 2023-09-20
NO-JIRA Remove Jackson from managed dependencies

amarkevich:


gemmellr:
This would need a JIRA if we were to do it, but I dont see that we would want to do it. Its there to manage the version to be consistent and often lift the version to not use ones with CVEs.  Closing.

amarkevich:
Jasckson is not a part of binary distribution. Addition effort required to manage not used dependency

PR #4627, opened 2023-09-22
ARTEMIS-4141 Update credits even for expired messages

essobedo:
Fixes https://issues.apache.org/jira/browse/ARTEMIS-4141

## Motivation

When big messages are produced if a consumer receives an expired message, the credits are not updated, so if the consumer is too slow and an expiry delay has been set, we can end up with a situation where there are no more credits which prevents the consumer from receiving any more messages.

## Modifications:

* Updates the credits even in case of an expired message to avoid running out of credits
* Add a test simulating the use case of the ticket


jbertram:
I ran the full test-suite on this and it looks good.

clebertsuconic:
this LGTM. thank you so much for this

PR #4628, opened 2023-09-22
ARTEMIS-4436 Artemis is logging warnings during clean shutdown of server in cluster.

ehsavoie:
When we know that a node leaves a clustercleanly we shouldn't log WARN messages about it.

Issue: https://issues.apache.org/jira/browse/ARTEMIS-4436

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
@@ -1602,4 +1602,10 @@ public interface ActiveMQServerLogger {
 
    @LogMessage(id = 224133, value = "{} orphaned page transactions have been removed", level = LogMessage.Level.INFO)
    void cleaningOrphanedTXCleanup(long numberOfPageTx);
+
+   @LogMessage(id = 224134, value = "Connection closed with failedOver={}", level = LogMessage.Level.INFO)
+   void bridgeConnectionClosed(Boolean failedOver);
+
+   @LogMessage(id = 224135, value = "nodeID {} is closing. Topology update ignored", level = LogMessage.Level.INFO)
+   void nodeLeavingCluster(String nodeID);
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionImpl.java
@@ -554,9 +554,13 @@ public final class ClusterConnectionImpl implements ClusterConnection, AfterConn
     * @return
     */
    @Override
-   public boolean removeMember(final long uniqueEventID, final String nodeId) {
+   public boolean removeMember(final long uniqueEventID, final String nodeId, final boolean disconnect) {
       if (nodeId.equals(nodeManager.getNodeId().toString())) {
-         ActiveMQServerLogger.LOGGER.possibleSplitBrain(nodeId);
+         if (!disconnect) {
+            ActiveMQServerLogger.LOGGER.possibleSplitBrain(nodeId);
+         } else {
+            ActiveMQServerLogger.LOGGER.nodeLeavingCluster(nodeId);
+         }
          return false;
       }
       return true;
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/warnings/ClusterCleanNodeShutdownTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/warnings/ClusterCleanNodeShutdownTest.java
@@ -1,0 +1,96 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.tests.integration.cluster.warnings;
+
+import java.lang.invoke.MethodHandles;
+import org.apache.activemq.artemis.core.server.cluster.impl.MessageLoadBalancingType;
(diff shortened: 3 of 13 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/warnings/ClusterCleanNodeShutdownTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+      printBindings(2);
+
+      sendInRange(1, "queues.testaddress", 0, 10, true, null);
+
+      try (AssertionLoggerHandler loggerHandler = new AssertionLoggerHandler(true)) {
+         logger.info("*****************************************************************************");
+         stopServers(0);
+         // Waiting some time after stopped
+         Thread.sleep(2000);
```
clebertsuconic, line 74: Can. you use a Wait Clause here?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/warnings/ClusterCleanNodeShutdownTest.java
```diff
@@ -0,0 +1,108 @@
...
+      }
+      startServers(0);
+
+      waitForBindings(0, "queues.testaddress", 1, 1, true);
+      waitForBindings(1, "queues.testaddress", 1, 0, true);
+
+      waitForBindings(0, "queues.testaddress", 1, 0, false);
+      waitForBindings(1, "queues.testaddress", 1, 1, false);
+
+      printBindings(2);
```
clebertsuconic, line 87: do you always need the printBindings? Can't you add it only when something "bad" happened?

or at the worst case... if (logger.isDebugEnabled()) ?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/warnings/ClusterCleanNodeShutdownTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+      waitForBindings(0, "queues.testaddress", 1, 1, true);
+      waitForBindings(1, "queues.testaddress", 1, 0, true);
+
+      waitForBindings(0, "queues.testaddress", 1, 0, false);
+      waitForBindings(1, "queues.testaddress", 1, 1, false);
+
+      sendInRange(1, "queues.testaddress", 0, 10, true, null);
+      try (AssertionLoggerHandler loggerHandler = new AssertionLoggerHandler(true)) {
+         logger.info("*****************************************************************************");
```
clebertsuconic, line 71: logger.debug();


for these two statements...



Everything else looks good... I run the whole test suite and it was good

clebertsuconic: I may amend myself and merge it today (later...)

gemmellr:
This looks like it aims to suppress a [warning] notice that the other broker has gone away, is that correct? Its not clear to me that entirely makes sense. I could certainly see reducing the notice to Info level, but it still seems like one should be given? I don't necessarily care that the other broker was shut down cleanly, often just that it isn't there any more and was expected to be.

(I've also seen several times where people 'cleanly' shut down things without realising they are doing it, then come asking why it has 'crashed' or 'shut down by itself')

ehsavoie:
@gemmellr Yes that's correct. Moving the message to info level is another solution. Currently people removing a node from a cluster and seeing warning is also a problem. They are doing something completely fine so WARN is a bit too high

clebertsuconic:
@gemmellr do you have any concerns on this PR? If not I think it should be merged.

gemmellr:
Yes, per my comment, I dont think it makes sense to remove notice that clients/brokers-acting-as-clients lost connection just because the other side was shut down 'cleanly'. This side still expected to be connected and no longer is, thats detail they'll often want/need to know, regardless of whether the other side went away intentionally (which, per my comment, I've seen people do unwittingly on several occasions).

I agree a warning level message is excessive, but suppressing notice entirely seems the wrong fix to that...I'd still report it happened, but at Info level.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
```diff
@@ -1593,4 +1593,10 @@ void noQueueIdDefined(org.apache.activemq.artemis.api.core.Message message,
 
    @LogMessage(id = 224130, value = "The {} value {} will override the {} value {} since both are set.", level = LogMessage.Level.INFO)
    void configParamOverride(String overridingParam, Object overridingParamValue, String overriddenParam, Object overriddenParamValue);
+
+   @LogMessage(id = 224131, value = "Connection closed with failedOver={}", level = LogMessage.Level.INFO)
+   void bridgeConnectionClosed(Boolean failedOver);
+
+   @LogMessage(id = 224132, value = "There is a possible split brain on nodeID {} closing. Topology update ignored", level = LogMessage.Level.INFO)
+   void possibleSplitBrainAfterClosing(String nodeID);
```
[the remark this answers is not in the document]
ehsavoie: Correct, this should go

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionImpl.java
```diff
@@ -554,10 +554,12 @@ public boolean updateMember(long uniqueEventID, String nodeId, TopologyMemberImp
...
    @Override
-   public boolean removeMember(final long uniqueEventID, final String nodeId) {
+   public boolean removeMember(final long uniqueEventID, final String nodeId, final boolean disconnect) {
       if (nodeId.equals(nodeManager.getNodeId().toString())) {
-         ActiveMQServerLogger.LOGGER.possibleSplitBrain(nodeId);
-         return false;
+         if (!disconnect) {
+            ActiveMQServerLogger.LOGGER.possibleSplitBrain(nodeId);
+            return false;
+         }
```
gemmellr, line 562: With my other comment about the 'info message if disconnect' you added previously, I was meaning more to leave it always logging the warning, wheres now this is back to not saying anything.

Unless we know for sure there isnt such possibility because it was a shutdown?  Its not clear to me, I dont know enough about the clustering to say either way. 

This also seems to change it from 'rejecting a topology update' to accepting it in this case, which seems a larger change than just the logging, and which I'm similarly not clear on the effect of. Need others with more knowledge here...@clebertsuconic @jbertram ?

ehsavoie: yes maybe I should leave it to return false.
Leaving it to WARN will mean having a warning when a node leaves which seems strange to me

ehsavoie:
@gemmellr @clebertsuconic  @jbertram could you please take a look now that I've updated the PR

gemmellr:
My last feedback was that I didnt know enough to say whether it was ok or not, and to ask for others to finish reviewing, so at this point I'll defer to those that do.

brusdev (approved):

PR #4632, opened 2023-09-24
Docker docs & usability

jbertram:


michaelandrepearce:
Is it possible to add back a search functionality and also ability to have collapsable nav like old version had? Rather than just top level nav?

jbertram:
@michaelandrepearce, there is no existing option to just turn on search or a collapsible ToC, but technically speaking [there are ways to customize the output](https://stackoverflow.com/questions/66651861/how-to-collapse-toctable-of-contents-in-spring-restdoc-asciidoc) that would probably allow this. I know the [AsciiDoctor docs](https://docs.asciidoctor.org/asciidoctor/latest/) (which use AsciiDoc) have search functionality and I've seen instance of collapsible ToC in other projects as well. 

That said, the previous search functionality wasn't great in my opinion, and folks might be better off using a search query like this instead:
```
site:https://activemq.apache.org/components/artemis/documentation/latest/ <searchTerms>
```
Queries like this are supported by the major search engines that I'm aware of.

docs/user-manual/docker.adoc
```diff
@@ -0,0 +1,130 @@
...
+
+== Official Images
+
+Official https://www.docker.com/[Docker] images are https://hub.docker.com/r/apache/activemq-artemis/tags[available on dockerhub].
+Images are pushed along with all the other distribution artifacts for every release.
+The fastest, simplest way to get started is with this command which will create and start a container named `mycontainer` and remove it when it terminates:
+
+[,console]
+----
+$ docker run --rm --name mycontainer -it -p 61616:61616 -p 8161:8161 apache/activemq-artemis:latest-alpine
```
clebertsuconic, line 15: Just a small tweak. I think we should "educate" users to use detached mode, and to use logs to look at the output.

I'm not strong opinionated about this though. just a suggestion.

jbertram: Done.

PR #4634, opened 2023-09-26
ARTEMIS-4443 support broker plugin config via broker properties

gtully:


gtully:
I need to revisit, registerBrokerPlugins is part of the official api and documented as the way to register/install a plugin.
I need to rework it to support a two stage add and register, but must respect that api in some way.

gtully:
by making registerBrokerPlugins check for contains on each of the typed collections, it is safe to have it called twice in the case or programmatic calls before server.start. Doing it at start allows entries added to brokerPlugins to be properly registered.

brusdev:
@gtully this is a cool improvement!

What about using an annotation to avoid the suffix `.class` and the default getName method?
```
   public @interface BrokerProperty {
      enum NameType {
         PROPERTY,
         CLASS
      }
      public NameType itemName() default NameType.PROPERTY;
   }

   @Override
   @BrokerProperty(itemName = BrokerProperty.NameType.CLASS)
   public List<ActiveMQServerBasePlugin> getBrokerPlugins() {
   ...
```

gtully:
Thanks Dom, not yet sure of the annotation gives any value, but I see that we never reuse the configuration instance on reload, and it may be that a plugin is never referenced for configuration purposes, meaning a name attribute won't be necessary, everything needs to go through init which is a single line. That can simplify the use case a lot

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
```diff
@@ -2057,6 +2061,57 @@ public void testStatusOnErrorApplyingProperties() throws Exception {
...
+      // serer initialisePart1
+      configuration.registerBrokerPlugins(configuration.getBrokerPlugins());
+
+      Assert.assertEquals(1, configuration.getBrokerPlugins().size());
+      Assert.assertEquals(1, configuration.getBrokerMessagePlugins().size());
+      Assert.assertEquals(1, configuration.getBrokerConnectionPlugins().size());
+
+      insertionOrderedProperties = new ConfigurationImpl.InsertionOrderedProperties();
+
+      // possible to change any attribute, but plugins only registered on start
```
gtully, line 2089: I think this is invalid, on reload, a new configuration impl will be created. And the pattern on a plugin is a single call to init, which is one line of properties config

brusdev (approved):

PR #4635, opened 2023-09-27
ARTEMIS-4444 Support custom directories to add runtime dependencies

brusdev:
The system property `artemis.extra.libs` is a comma separated list of directories that contains jar files, i.e.
```
-Dartemis.extra.libs=/usr/local/share/java/lib1,/usr/local/share/java/lib2
```
The environment variable `ARTEMIS_EXTRA_LIBS` is a comma separated list of directories that contains jar files and
is ignored if the system property `artemis.extra.libs` is defined, i.e.
```
export ARTEMIS_EXTRA_LIBS=/usr/local/share/java/lib1,/usr/local/share/java/lib2
```

### DIFF
--- a/docs/user-manual/using-server.adoc
+++ b/docs/user-manual/using-server.adoc
@@ -532,10 +532,22 @@ NOTE: the environment variable `JAVA_ARGS_APPEND` can be used to append or overr
 
 Runtime dependencies like diverts, transformers, broker plugins, JDBC drivers, password decoders, etc. must be accessible by the broker at runtime.
 Package the dependency in a jar, and put it on the broker's classpath.
-This can be done by placing the jar file in the `lib` directory of the broker distribution itself or in the `lib` directory of the broker instance.
+This can be done by placing the jar file in the `lib` directory of the broker distribution itself,
+by placing the jar file in the `lib` directory of the broker instance,
+by setting the system property `artemis.extra.libs` with the directory that contains the jar file, or
+by setting the environment variable `ARTEMIS_EXTRA_LIBS` with the directory that contains the jar file,
 A broker instance does not have a `lib` directory by default so it may need to be created.
 It should be on the "top" level with the `bin`, `data`, `log`, etc.
 directories.
+The system property `artemis.extra.libs` is a comma separated list of directories that contains jar files, i.e.
+```
+-Dartemis.extra.libs=/usr/local/share/java/lib1,/usr/local/share/java/lib2
+```
+The environment variable `ARTEMIS_EXTRA_LIBS` is a comma separated list of directories that contains jar files and
+is ignored if the system property `artemis.extra.libs` is defined, i.e.
+```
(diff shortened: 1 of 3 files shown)


gtully (approved):
like it!

brusdev:
I changed the implementation to allow only a comma separated list of directories.

gtully:
looks even better :-)

docs/user-manual/using-server.adoc
```diff
@@ -532,10 +532,16 @@ NOTE: the environment variable `JAVA_ARGS_APPEND` can be used to append or overr
 
 Runtime dependencies like diverts, transformers, broker plugins, JDBC drivers, password decoders, etc. must be accessible by the broker at runtime.
 Package the dependency in a jar, and put it on the broker's classpath.
-This can be done by placing the jar file in the `lib` directory of the broker distribution itself or in the `lib` directory of the broker instance.
+This can be done by placing the jar file in the `lib` directory of the broker distribution itself,
+by placing the jar file in the `lib` directory of the broker instance or
+by setting the environment variable `ARTEMIS_EXTRA_LIBS` with the directory that contain the jar file.
```
gemmellr, line 537:
```suggestion
by placing the jar file in the `lib` directory of the broker instance, or
by setting the environment variable `ARTEMIS_EXTRA_LIBS` with the directory that contains the jar file.
```

brusdev: suggestion applied, thanks

gemmellr (approved):

clebertsuconic:
I think we should use the db-tests to validate this. I am going to tweak the Derby Test to use this variable.

clebertsuconic:
@brusdev I converted the Derby Page Test to use this extension, and I used a -D definition on the library folder.

And that actually raised me an issue... you'r using System.getEnv. Shouldn't you use System.getProperty() which would work with either -D or a exported property.


I sent a commit to your branch for your evaluation.

clebertsuconic:
If you agree with the test perhaps you should squash the change.

brusdev:
@clebertsuconic thanks for the test, I like it. Maybe for a system property I should use a lowercase name, i.e. `artemis.extra.libs` WDYT?

brusdev:
Actually, @gtully pointed me to `envVarNameFromSystemPropertyName`, see https://github.com/apache/activemq-artemis/blob/2.31.0/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/DefaultSensitiveStringCodec.java#L151

I'll push a new commit to handle both env (`ARTEMIS_EXTRA_LIBS`) ans system property (`artemis.extra.libs`)

clebertsuconic:
@brusdev I manually tested and I had the impression that System.getProperty would handle the property defined on both env and -D from java.


perhaps we should also test the System.evn to make sure?

clebertsuconic:
LGTM.

clebertsuconic (approved):

PR #4636, opened 2023-09-28
ARTEMIS-4446 improve readability of code/config blocks in user manual

jbertram:


jbertram:
Here's what it looked like before:
image
Here's what it looks like after:
image

PR #4637, opened 2023-09-28
ARTEMIS-4447 Add paging prefetch parameters into address settings

clebertsuconic:
we are adding new attributes to determine how many messages (or bytes) we are reading from paging into Queue memory.


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/paging/PageFilterTest.java
```diff
@@ -0,0 +1,163 @@
...
+            }
+            oneThousandCharacters = builder.toString();
+         }
+
+         MessageProducer producer = session.createProducer(topic);
+
+
+         Session txsession = otherConnection.createSession(true, Session.SESSION_TRANSACTED);
+         MessageProducer txProducer = txsession.createProducer(topic);
+         for (int i = 0; i < 100; i++) {
```
gemmellr, line 109: hard coding 100 here when numberOfMessages is defined to the rather random looking 3 * 24 = 72 at the start just seems confusing.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/paging/PageFilterTest.java
```diff
@@ -0,0 +1,163 @@
...
+         server.stop();
+      }
+
+      server.start();
+
+      try (Connection connection = connectionFactory.createConnection()) {
+         Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+         Topic topic = session.createTopic(getName());
+         connection.start();
+         MessageConsumer consumerWhole = session.createSharedDurableConsumer(topic, "partialThing", filter);
```
gemmellr, line 141: consumerWhole using the "partialThing" sub seems confusing (as opposed to e.g the "wholeThing" sub that was created but not used...and which itself doesnt get everything anyway). Perhaps some clearer naming?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/paging/PageFilterTest.java
```diff
@@ -0,0 +1,163 @@
...
+         int receivedMessages = 0;
+         for (int i = 0; i < numberOfMessages; i++) {
+            Message message = consumerWhole.receive(1000);
+            if (message == null) {
+               break;
+            }
+            Assert.assertNotNull(message);
+            receivedMessages++;
+         }
+         logger.debug("Received {} messages", receivedMessages);
```
gemmellr, line 151: We aren't even asserting how many messages were actually received?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/paging/PageFilterTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+         connection.start();
+
+         int receivedI = 0;
+         do {
+            Message message = partialThing.receive(1000);
+            Assert.assertNotNull(message);
+            receivedI = message.getIntProperty("i");
+         }
+         while (receivedI < 10);
```
gemmellr, line 130: This doesnt even check what it receives?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/paging/PageFilterTest.java
```diff
@@ -0,0 +1,163 @@
...
+import org.apache.activemq.artemis.tests.util.ActiveMQTestBase;
+import org.apache.activemq.artemis.tests.util.CFUtil;
+import org.apache.activemq.artemis.tests.util.PrintDataUtil;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class PageFilterTest extends ActiveMQTestBase {
```
gemmellr, line 43: This class doesnt appear to even use the new config it seems. Is it actually related?

It doesnt really assert much of anything either. Is it a dev 'utility/debug test' ?

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -3946,6 +3954,14 @@
             </xsd:annotation>
          </xsd:element>
 
+         <xsd:element name="prefetch-page-bytes" type="xsd:string" default="20M" maxOccurs="1" minOccurs="0">
+            <xsd:annotation>
+               <xsd:documentation>
+                  How many bytes we are reading from paging and keeping ready in queues ready to deliver to consumers.
```
gemmellr, line 3960: The default could be spelled out clearer? It isnt really just 20M, but whatever max-read-page-bytes is (the default for which is also not just 20M but 2x page size...which similarly isnt clear). Similar to other comment, interaction of the two things isnt that clear either.

Similar for the messages config.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
```diff
@@ -1623,6 +1623,32 @@ public void testAddressSettingsPageLimitInvalidConfiguration5() throws Throwable
...
+   public void testPagePrefetch() throws Throwable {
+      ConfigurationImpl configuration = new ConfigurationImpl();
+
+      Properties properties = new Properties();
+
+      String randomString = RandomUtil.randomString();
+
+      properties.put("addressSettings.#.expiryAddress", randomString);
+      properties.put("addressSettings.#.prefetchPageMessages", "333"); // this should make the PagingStore to parse it as null
+      properties.put("addressSettings.#.prefetchPageBytes", "777"); // this should make the PagingStore to parse it as null
```
gemmellr, line 1636: Comments dont seem correct, its explicitly checking they aren't null?

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
```diff
@@ -1623,6 +1623,32 @@ public void testAddressSettingsPageLimitInvalidConfiguration5() throws Throwable
...
+      Assert.assertEquals(333, configuration.getAddressSettings().get("#").getPrefetchPageMessages());
+      Assert.assertEquals(777, configuration.getAddressSettings().get("#").getPrefetchPageBytes());
+
+      PagingStore storeImpl = new PagingStoreImpl(new SimpleString("Test"), (ScheduledExecutorService) null, 100L, Mockito.mock(PagingManager.class), Mockito.mock(StorageManager.class), Mockito.mock(SequentialFileFactory.class), Mockito.mock(PagingStoreFactory.class), new SimpleString("Test"), configuration.getAddressSettings().get("#"), null, null, true);
+
+      Assert.assertEquals(333, storeImpl.getPrefetchPageMessages());
+      Assert.assertEquals(777, storeImpl.getPrefetchPageBytes());
+   }
+
+
```
gemmellr, line 1651: A test [somewhere] checking the values are are as expected when _not_ explicitly set would seem appropriate.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/paging/PageFilterTest.java
clebertsuconic: When I started investigation this issue, I had an OME and a heap dump from a user, and it seemed the issue was on filtering. 

In the end the issue I was investigating was a result of misconfigured maxes.. and I left the test here.


There are other tests testing filtering.. I will remove that test from this PR.

clebertsuconic:
@gemmellr docs are updated...

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
clebertsuconic: I need to do this one still... doing it now.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -3937,11 +3937,29 @@
             </xsd:annotation>
          </xsd:element>
 
-         <xsd:element name="max-read-page-bytes" type="xsd:string" default="20M" maxOccurs="1" minOccurs="0">
+         <xsd:element name="prefetch-page-messages" type="xsd:int" default="20MB" maxOccurs="1" minOccurs="0">
```
gemmellr, line 3940: c&p error on the default, this is the messages one so 20MB doesnt make sense.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -3937,11 +3937,29 @@
...
+            <xsd:annotation>
+               <xsd:documentation>
+                  How many messages we are reading from paging and keeping ready in queues ready to deliver to consumers. Default: If not defined max-read-page-messages is used.
```
gemmellr, line 3943: This, coupled with max-page-read-size starting "How many messages are we allowed to read from page into the Queue each time." doesnt do that great a job of conveying the behaviour or relationship between the two.

Arguably max-read-page-messages needs updated to drop the "each time" to help clarify it being a limit on the total read in, while this one further allows limiting the amount read in at any given time to a smaller number.

```suggestion
                  How many messages the broker will read from paging and keep ready in queues ready to deliver to consumers. Default: If not defined max-read-page-messages is used.
```

docs/user-manual/paging.adoc
```diff
@@ -135,10 +141,10 @@ The configured policy will start based on the first value to reach its mark.
 
 ==== Maximum read from page
 
-`max-read-page-messages` and `max-read-page-bytes` are used to control messaging reading from paged file into the Queue.
-The broker will add messages on the Queue until either `max-read-page-meessages` or `max-read-page-bytes` reaches the limit.
+`max-read-page-messages`, `max-read-page-bytes`, `prefetch-page-messages` and `prefetch-page-bytes` are used to control reading from paged file into the Queue.
+The broker will add messages as long as all these limits are satisfied.
 
-If both values are set to -1 the broker will keep reading messages as long as the consumer is reaching for more messages.
+If all these values are set to -1 the broker will keep reading messages as long as the consumer is reaching for more messages.
```
gemmellr, line 147: Its still just if `max-read-page-messages` and `max-read-page-bytes`are set to <=0, as it was before before, so its probably clearer to just say that.
```suggestion
If `max-read-page-messages` and `max-read-page-bytes` are set to -1 the broker will keep reading messages as long as the consumer is reaching for more messages. The `prefetch-page-messages` and `prefetch-page-bytes` will not be considered.
```

artemis-server/src/main/resources/schema/artemis-configuration.xsd
gemmellr: (presumably this is what made the test runs all fail)

docs/user-manual/paging.adoc
clebertsuconic: I will actually update the code to use all these values.

clebertsuconic:
I am going to merge this now.... further fixes in documentation (or anything else) can be sent before a release.

thanks for all the help.

PR #4639, opened 2023-09-29
NO-JIRA github action JDK 21

clebertsuconic:


gemmellr:
Yeah, hadnt done this yet since Temurin 21 has still to be released (https://adoptium.net/en-GB/blog/2023/09/temurin21-delay/). Have already been checking it periodically with 21-ea elsewhere for many months though. Might actually just push a change to use the 21-ea for now, given that its based on the RC1/GA jdk tag version already.

gemmellr:
Changed it to 21-ea for now in commit https://github.com/apache/activemq-artemis/commit/3f2587c85aa727def979d9412a8b44806f1634ba

 > Downloading Java 21.0.0+35.0.ea (Temurin-Hotspot)

Its still a pre-GA build, but per previous comment it is built from what become the JDK 21 GA after the RC period (e.g see https://github.com/openjdk/jdk21/tags showing jdk-21+35 and jdk-21-ga tags both using the same commit)

PR #4640, opened 2023-10-02
ARTEMIS-4449 Fix url parameter separator in acceptor configuration

nandorsoma:
https://issues.apache.org/jira/browse/ARTEMIS-4449


nandorsoma:
As this is my first pull request in the activemq community, I'm not familiar with the processes here. Please tell me if I need to do anything differently.

PR #4641, opened 2023-10-04
ARTEMIS-4450 - Auto-deleted clustered destinations can cause message …

AntonRoskvist:
…loss

If a destination has a remote binding but not a local one, certain MessageLoadBalancingTypes can cause message loss, transparent to the producer.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/Bindings.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/Bindings.java
@@ -63,7 +63,5 @@ public interface Bindings extends UnproposalListener {
 
    int size();
 
-   default boolean contains(Class clazz) {
-      return false;
-   }
+   boolean hasLocalBinding();
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/BindingsImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/BindingsImpl.java
@@ -52,6 +52,7 @@ import org.apache.activemq.artemis.utils.CompositeAddress;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.lang.invoke.MethodHandles;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.function.BiConsumer;
 
 public final class BindingsImpl implements Bindings {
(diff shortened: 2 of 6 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1269,7 +1269,7 @@ private AddressInfo checkAddress(RoutingContext context, SimpleString address) t
 
    Bindings simpleRoute(SimpleString address, RoutingContext context, Message message, AddressInfo addressInfo) throws Exception {
       Bindings bindings = addressManager.getBindingsForRoutingAddress(address);
-      if (bindings == null && context.getServerSession() != null) {
+      if ((bindings == null || !bindings.contains(LocalQueueBinding.class)) && context.getServerSession() != null) {
```
clebertsuconic, line 1272: simpleRoute is a hot method. This would cause an iteration on bindings to verify if hasLocal or not. 

I'm amending your commit to have the attribute cached.


The only place this method was called (which was recently introduced) was during a queue creation.. and there it wouldn't be as hot as this one.

clebertsuconic:
@AntonRoskvist I used a feature from GitHub where I pushed a commit to your repo. 

if you could rebase your repo with your local fork, I added a commit, which I will run tests on it now.


if you could review it please?

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/Bindings.java
```diff
@@ -63,7 +63,5 @@ public interface Bindings extends UnproposalListener {
 
    int size();
 
-   default boolean contains(Class clazz) {
-      return false;
-   }
+   boolean hasLocalBinding();
```
clebertsuconic, line 66: I know this will have compilation failures :(... I'm fixing and I will push another commit

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/BindingsImpl.java
```diff
@@ -61,6 +61,19 @@ public final class BindingsImpl implements Bindings {
...
 
+   private boolean hasLocal;
+
+   private void checkBinding(Long bindingID, Binding binding) {
+      if (binding instanceof LocalQueueBinding) {
+         hasLocal = true;
+      }
+   }
+
+   private void checkBindinsIdMap() {
```
tabish121, line 72: Seems like a spelling error maybe "checkBindingsIdMap"

clebertsuconic: @tabish121 thanks

clebertsuconic:
@AntonRoskvist thanks for this contribution....

Please check my addition through your branch.

clebertsuconic:
It is merged.  

GitHub has a thing that if you allow. Members of the project can change your branch.   I did that right from your fork :)

clebertsuconic:
So if you do a git fetch on your fork you will see my push into your fork. 

But I have merged this already in main.

PR #4642, opened 2023-10-04
ARTEMIS-4451: fix non-SASL AMQP connection + resource audit logging

gemmellr:
Ensure the eligibility check done can succeed with the audit logging enabled, allowing non-SASL connections if broker config allows (rather than erroneously failing and indicating SASL is required when it shouldnt be).


gtully:
Nice unit test additions

PR #4643, opened 2023-10-06
ARTEMIS-4453 Lots of addresses breaks cluster bridge flow control

jbertram:


jbertram:
@clebertsuconic, one concern I have with this solution is that once `MAX_UNREFERENCED_CREDITS_CACHE_SIZE` is exceeded then every time an entry is added to `unReferencedCredits` a full scan of the `Map` will occur. If there's nothing to remove and entries keep getting added then that will be a lot of scan potentially. However, moving this to some async still have problems because everything is `synchronized`.

clebertsuconic:
@jbertram we could use a scheduled executor and scan every X seconds as opposed to on every message sent?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerCreditManagerImpl.java
```diff
@@ -162,26 +162,18 @@ private void addToUnReferencedCache(final SimpleString address, final ClientProd
...
-         // Remove the oldest entry
-
          Iterator<Map.Entry<SimpleString, ClientProducerCredits>> iter = unReferencedCredits.entrySet().iterator();
-
-         Map.Entry<SimpleString, ClientProducerCredits> oldest = iter.next();
-
-         iter.remove();
-
-         removeEntry(oldest.getKey(), oldest.getValue());
+         while (iter.hasNext()) {
```
clebertsuconic, line 166: I would use map.forEach here.


And regarding your concern with always doing this... we could do this through a schedueledExecutor? Schedule a call every X seconds.. and keep it scheduled as long as it's beyond the limit?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/SimpleSymmetricClusterTest.java
```diff
@@ -127,6 +136,80 @@ public void testSimple() throws Exception {
 
    }
 
+   @Test
+   public void testWildcardFlowControl() throws Exception {
```
clebertsuconic, line 140: @jbertram how long this test would take to run? Perhaps we should have this one test under SoakTests? (No need to make a RealServerTest.. just put it under soak *if * this is a long running test?

jbertram: The test is actually quite fast so I don't think it's necessary to move. It finishes in around 4 seconds on my laptop.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerCreditManagerImpl.java
jbertram: Regarding using `forEach`...I don't think I can do this because I need to remove an entry from the map I'm inspecting.

I went with the `ScheduledExecutorService`.

clebertsuconic:
@jbertram I'm converting this PR to draft since I sent a draft commit to your repo

clebertsuconic:
@jbertram you should/must assert the number of elements on the ClientProducerCreditManagerImpl otherwise there's no way to verify the implementation is correct.


You need to expose the private property through an accessor... and make the assertion.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerCreditManagerImpl.java
```diff
@@ -16,20 +16,29 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public class ClientProducerCreditManagerImpl implements ClientProducerCreditManager {
 
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
    public static final int MAX_UNREFERENCED_CREDITS_CACHE_SIZE = 1000;
 
+   private static final int MAX_UNREFERENCED_CREDITS_CACHE_TIME_CHECK = 500; // milliseconds
```
clebertsuconic, line 40: I set this as half second just for the test... but I'm actually thinking this should be something with a short timeout like this.. as long as the cleanup works.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientProducerCreditManagerImpl.java
```diff
@@ -168,33 +160,30 @@ public synchronized int creditsMapSize() {
 
    @Override
    public synchronized int unReferencedCreditsSize() {
```
jbertram, line 162: Rename to `getAnonymousCreditsSize()`?

PR #4645, opened 2023-10-10
ARTEMIS-4456 Register metrics plugin

brusdev:


erwindon:
great! now I can work with a ready-made build on https://github.com/rh-messaging/artemis-prometheus-metrics-plugin/issues/19...

PR #4648, opened 2023-10-12
ARTEMIS-4459 log when rejecting dupe MQTT QoS2 pub

jbertram:
In accordance with the QoS2 protocol outlined in the MQTT specification(s), once the broker receives a PUBLISH then any other PUBLISH it receives on that same session with the same packet ID must be ignored until the QoS2 protocol for that ID is completed.

The broker does this, but it doesn't log anything so it's not clear when this is actually happening.


jbertram:
The main details in the new logging are the MQTT client ID as well as the ID for the MQTT `PUBLISH` packet. Neither of these will be available in the audit log. You'd need MQTT protocol logging (i.e. `TRACE` on `org.apache.activemq.artemis.core.protocol.mqtt.MQTTUtil`) to be active. I considered putting this new logging on that same logger, but ultimately I didn't. What do you think about using the same logger as the protocol trace? Thinking about it more now it seems like a natural fit.

PR #4650, opened 2023-10-17
ARTEMIS-4461 make implicit dependencies explicit in artemis-features

jbertram:


artemis-features/pom.xml
```diff
@@ -49,6 +49,58 @@
...
+        <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>artemis-hornetq-protocol</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>artemis-openwire-protocol</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
gemmellr, line 103: I'm guessing you are adding this for build-ordering reasons? If so, it may be wise to structure things so that they apply build constraints without really altering the user-facing dependencies, e.g excluding all the deps (and perhaps making them optional also).

The fact these have never been here mean they clearly arent needed by the actual feature usage. Their new presence however could mean they are now all depended on for things touching the pom, and may need to resolve all those modules and their dependencies, even if the feature being used doesnt need them.

clebertsuconic:
@gemmellr I just pushed -f on Justin's branch to depend on the Artemis-distribution pom instead. if you could check this please?


I tested it here and it worked quite alright.

PR #4651, opened 2023-10-19
ARTEMIS-4463 fixed navigation for attributes-button in address-table

erwindon:
see also https://issues.apache.org/jira/browse/ARTEMIS-4463


erwindon:
code shamelessly copied from the similar code in queues.js, tested ok

PR #4653, opened 2023-10-23
ARTEMIS-4467 Core client code visibility change required.

ehsavoie:
* Since the constructors in RemotingConnectionImpl can be used from the server and the client. If the server calling code is in a different classloader then the constructor can't be called.
* same for the 'active' boolean property of ActiveMQChannelHandler

Issue: https://issues.apache.org/jira/browse/ARTEMIS-4467


jbertram:
This part of the commit message is not clear to me:

> If the server calling code is in a different classloader then the constructor can't be called.

Do you mean that "any code calling the `RemotingConnectionImpl` constructor in a different classloader will fail"?

ehsavoie:
@jbertram yes because the constructor is package protected but the package is different because the classloader is different.

PR #4655, opened 2023-10-26
#ARTEMIS-4470: prevent the common trap

pekuz:
so far omitting reencode results in latent error that surfaces only in AMQP-AMQP messaging


pekuz:
removed the offending trailing space, I have overlooked earlier.

jbertram:
FWIW, the commits should be squashed.

pekuz:
Squeezing these particular commits would erase some noise from history. But I'm at limit of my Git knowledge. When I try in my clone of the forked repo, GitHub desktop complaints there is a merge commit preventing squeezing. The merge commit had been indeed avoidable by reordering Git operations but once it got in I do not know how to eliminate it.

PR #4656, opened 2023-10-26
ARTEMIS-4476 Race conditions through failure detection in AMQP and Core

clebertsuconic:


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonServerSenderContext.java
```diff
@@ -980,6 +983,19 @@ class DefaultController implements SenderController {
...
+            // Avoiding possible NPEs that could happen on mock tests
+            if (connection != null &&
+               (handler = connection.getHandler()) != null && (qpidConnection = handler.getConnection()) != null) {
+               if (qpidConnection.getRemoteState() == EndpointState.CLOSED) {
+                  logger.warn("AMQP Connection creating invalid consumer for closed connection", connection);
+                  throw new IllegalStateException("AMQP connection " + connection.getRemoteAddress() + " creating invalid consumer for closed connection");
+               }
+            }
+         }
+
```
tabish121, line 998: I'm not particularly fond of this "make mocking easier approach" as it ignore what should be an invalid case of missing proton handler and connection.  If a test needs to mock this it should be able to supply mocks that answer with mocked versions of these objects.

clebertsuconic: fair enough... 

I"m in the process to making it ready to be merged now.. and I will cleanup this before I set it ready to merge


Thanks for looking into it.

tests/integration-tests-isolated/src/test/java/org/apache/activemq/artemis/tests/integration/isolated/client/ConnectionDroppedTest.java
```diff
@@ -0,0 +1,508 @@
...
+                  peer.remoteBegin().queue();
+                  peer.expectBegin();
+                  peer.remoteAttach().ofReceiver().withName(RandomUtil.randomString()).withSenderSettleModeUnsettled().withReceivervSettlesFirst().withTarget().also().withSource().withAddress("test-queue").withExpiryPolicyOnLinkDetach().withDurabilityOfNone().withCapabilities("queue").withOutcomes("amqp:accepted:list", "amqp:rejected:list").also().queue();
+                  peer.expectAttach();
+                  peer.remoteClose().queue();
+                  peer.expectClose();
+
+                  peer.connect("localhost", 61616);
+
+                  // Waits for all the commands to fire and the drop action to be run.
```
tabish121, line 144: Comment isn't correct, no drop action is queued in this test

tests/integration-tests-isolated/src/test/java/org/apache/activemq/artemis/tests/integration/isolated/client/ConnectionDroppedTest.java
```diff
@@ -0,0 +1,508 @@
...
+            try {
+               peer.queueClientSaslAnonymousConnect();
+               peer.remoteOpen().queue();
+               peer.remoteBegin().queue();
+               peer.remoteAttach().ofReceiver().withName(RandomUtil.randomString()).withSenderSettleModeUnsettled().withReceivervSettlesFirst().withTarget().also().withSource().withAddress(getName()).withExpiryPolicyOnLinkDetach().withDurabilityOfNone().withCapabilities("queue").withOutcomes("amqp:accepted:list", "amqp:rejected:list").also().queue();
+
+               peer.connect("localhost", 61616);
+
+               // Waits for all the commands to fire and the drop action to be run.
+               peer.waitForScriptToComplete();
```
tabish121, line 352: I might add a 'peer.expectOpen()' before the connect as the wait action will throw an exception if the broker emits an open before stalling on the begin processing which could lead to some confusing logging output at the least.  Although in this case you could probably also just use a different wait method 'peer.waitForScriptToCompleteIgnoreErrors()' which will just let it do what you are asking and then shutdown cleanly without any errors being thrown or logged.

tests/integration-tests-isolated/src/test/java/org/apache/activemq/artemis/tests/integration/isolated/client/ConnectionDroppedTest.java
```diff
@@ -0,0 +1,508 @@
...
+
+            try {
+               peer.queueClientSaslAnonymousConnect();
+               peer.remoteOpen().queue();
+               peer.remoteBegin().queue();
+               peer.remoteAttach().ofReceiver().withName(RandomUtil.randomString()).withSenderSettleModeUnsettled().withReceivervSettlesFirst().withTarget().also().withSource().withAddress(getName()).withExpiryPolicyOnLinkDetach().withDurabilityOfNone().withCapabilities("queue").withOutcomes("amqp:accepted:list", "amqp:rejected:list").also().queue();
+
+               peer.connect("localhost", 61616);
+
+               // Waits for all the commands to fire and the drop action to be run.
```
tabish121, line 351: Comment is not correct, seems to be cut and pasted from another test, no drop action was added in this test.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/ActiveMQProtonRemotingConnection.java
```diff
@@ -73,17 +78,54 @@ public void fail(final ActiveMQException me, String scaleDownTargetNodeID) {
...
+      if (destroyed) {
+         return;
+      }
+
+      destroyed = true;
+
+      if (logger.isInfoEnabled()) {
+         try {
+            logger.debug("Connection regular close. amqpConnection.getHandler().getConnection().getRemoteState() = {}, remoteIP={}", amqpConnection.getHandler().getConnection().getRemoteState(), amqpConnection.getConnectionCallback().getTransportConnection().getRemoteAddress());
+         } catch (Throwable e) { // just to avoid a possible NPE from the debug statement itself
```
gemmellr, line 120: gate doesnt match the actual logging

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/ActiveMQProtonRemotingConnection.java
```diff
@@ -73,17 +78,54 @@ public void fail(final ActiveMQException me, String scaleDownTargetNodeID) {
 
       destroyed = true;
 
-      //filter it like the other protocols
-      if (!(me instanceof ActiveMQRemoteDisconnectException)) {
-         ActiveMQClientLogger.LOGGER.connectionFailureDetected(amqpConnection.getConnectionCallback().getTransportConnection().getRemoteAddress(), me.getMessage(), me.getType());
+      if (logger.isInfoEnabled()) {
+         try {
+            logger.debug("Connection failure detected. amqpConnection.getHandler().getConnection().getRemoteState() = {}, remoteIP={}", amqpConnection.getHandler().getConnection().getRemoteState(), amqpConnection.getConnectionCallback().getTransportConnection().getRemoteAddress());
```
gemmellr, line 83: gate doesnt match the actual logging

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ExecutorNettyAdapter.java
```diff
@@ -0,0 +1,229 @@
...
+import io.netty.util.concurrent.ScheduledFuture;
+import org.apache.activemq.artemis.utils.actors.ArtemisExecutor;
+
+/** Test cases may supply a simple executor instead of the real Netty Executor
+ *  On that case this is a simple adapter for what's needed from these tests.
+ *  Not intended to be used in production.
+ *
+ *  TODO: This could be refactored out of the main codebase but at a high cost.
+ *        We may do it some day if we find an easy way that won't clutter the code too much.
+ *  */
```
gemmellr, line 45: This is in the main code, shouldnt be if only used for tests? I feel like we deleted something like this recently.

clebertsuconic: I meant to move the ExecutorNettyAdapter from the amqpModel as a class private under inVM. 

This is used to simulate the inVM as an eventLoop.

clebertsuconic: I will do the move

clebertsuconic: this is being done at the new PR now. This ExecutorAdapter is being removed there, and inVM will now throw an exception if inVMLoop is called (what shouldn't happen).

So, this was all dead code anyways.. we never needed to use inVM with AMQP.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
```diff
@@ -2048,6 +2048,37 @@ public void deliverScheduledMessages(String filter) throws Exception {
       }
    }
 
+   @Override
+   public String verifyConnections() throws Exception {
```
clebertsuconic, line 2052: hmmm... I need a better name for this method...

getInvalidConnections()
inspectInvalidConnections

any other ideas?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationConnectTest.java
```diff
@@ -219,6 +219,7 @@ public void testFederationCreatesControlLinkAndClosesConnectionDetachIndicatesNo
                                        .withNullTarget();
          peer.remoteDetach().withErrorCondition("amqp:unauthorized-access", "Not authroized").queue();
          peer.expectDetach().optional();
+         peer.expectClose().optional();
```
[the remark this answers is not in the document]
clebertsuconic: Moving to the Connection Failure to the Netty thread unveiled an issue with Broker Connection. By the time I issue a failure, we won't wait the flush of the connection.close to be sent to the client before we close it... we just disconnect(). If you are writing a unit test like this.. the expect close becomes optional. Fixing it would require some work, and it would be orthogonal to this.

As for now, this is what I"m supposed to do with these tests.

tests/integration-tests-isolated/src/test/java/org/apache/activemq/artemis/tests/integration/isolated/client/ConnectionDroppedTest.java
```diff
@@ -0,0 +1,506 @@
...
+      } catch (Exception ignored) {
+      }
+      if (!done.await(10, TimeUnit.SECONDS)) {
+         for (int i = 0; i < 10; i++) {
+            System.out.println("Will fail");
+            Thread.sleep(1000);
+         }
+         logger.warn(ThreadDumpUtil.threadDump("Still running"));
+         Assert.fail("Threads are still running");
+      }
```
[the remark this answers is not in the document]
clebertsuconic: this is left over debugging.. at some point I needed the wait while I was checking status of the queue.


I will just remove it.. thanks for pointing this out.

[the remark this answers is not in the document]
clebertsuconic: I should practice adding a // REMOVE-ME at the end of my temporary statements :)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationConnectTest.java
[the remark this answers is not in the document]
gemmellr: Cant say I agree with that. Doesnt seem at all orthogonal for this change to be stopping it doing what it was/should.

tests/integration-tests-isolated/src/test/java/org/apache/activemq/artemis/tests/integration/isolated/client/ConnectionDroppedTest.java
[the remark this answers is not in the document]
gemmellr: I've told you a few times thats almost exactly what I actually do: //TODO: remove :)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationConnectTest.java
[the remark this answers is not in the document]
clebertsuconic: If you ran the test Long enough before it would probably also fail.

all I did was to move the connectionFailure to the correct thread.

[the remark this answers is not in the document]
tabish121: Just quick note to say that it wasn't sending close before, and only occasionally sending the detach.  It seems now it sometimes sends both, and sometimes not.  Would be nice if it sent them both as a well behaved peer should, but for clarity the original test omitted this as it was never found to have been sent.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
clebertsuconic: I'm enhancing listConsumerAsJSON for this...

thanks to @jbertram idea

clebertsuconic:
@tabish121 , @gemmellr thanks for the reviews.. I'm merging this now..

if you guys find anything later.. let me know and we can send further commits

clebertsuconic:
actually.. there's a test failure... and I' fixing that first.

clebertsuconic:
actually the failure is unrelated to my changes.. merging it

PR #4657, opened 2023-10-26
ARTEMIS-4477: Transform shaded references in META-INF/services in artemis-commons

gastaldi:
- Fixes ARTEMIS-4477: This makes sure that the shaded classes referenced in the META-INF/services are correct

### DIFF
--- a/artemis-commons/pom.xml
+++ b/artemis-commons/pom.xml
@@ -156,6 +156,9 @@
                            <include>jakarta.json:jakarta.json-api</include>
                         </includes>
                      </artifactSet>
+                     <transformers>
+                        <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
+                     </transformers>
                   </configuration>
                </execution>
             </executions>


gemmellr:
Comparing the 2.31.0 and 2.31.1 jars, the key difference looks like being that the former didnt have any services file at all, while the latter now has an incorrect one as it references the non-shaded class that then isnt present.

Given I believe the original reason for shading it was to have an 'internal, not-javax, not-jakarta' impl, its unclear to me that it should be registering it at all, which is what this change would presumably do once again by correcting the class name...so I'm not sure its the correct change? Excluding the services file again looks like it could be preferable?

gemmellr:
I raised #4658 for the alternative idea, if you wanted to try it out and see that it resolves the issue for you.

turing85:
@gemmellr Let's say it like this: I removed the `META-INF/services` directory from the `2.31.1 commons`-jar, and everything works. So I think both soutions (yours and the one provided by @gastaldi) should fix the issue on our end.

gastaldi:
@gemmellr @turing85 removing the file will make the shaded JsonProvider.provider() fail the lookup (https://github.com/jfuerth/javax-json/blob/master/api/src/main/java/javax/json/spi/JsonProvider.java#L87) but if that's not even called why bother shading it in first place? The previous versions were not returning the shaded impl?

gemmellr:
The file was never there before, in e.g 2.31.0 and prior, so thats why I propose it. I believe the impl is used directly, as the underlying impl for an internal API aiming to be agnostic to javax or jakarta (as the server jar is used in both) and so doesnt need it registering. I didnt implement it but its been this way for at least a couple years.

gastaldi:
Just saw that the Johnzon provider is instantiated directly in  https://github.com/apache/activemq-artemis/blob/main/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/JsonLoader.java#L38, so the META-INF/services is useless. 

I'll recommend merging #4658 instead.

gemmellr:
I have merged #4658 instead. Thanks a lot for the PR and report making it easy to identify the issue/problem-change.

gemmellr:
2.31.2 is available on central now the alternative fix.

turing85:
Thanks a lot, @gemmellr, your help is much appreciated :heart:

PR #4658, opened 2023-10-26
ARTEMIS-4477: remove incorrect services file

gemmellr:
Possible alternative to fix the issue raised in #4657. Partially restores the exclusion filter that was previously removed, narrowing it to drop the services file.


turing85:
Fix works as expected:
https://github.com/quarkiverse/quarkus-artemis/pull/316
https://github.com/quarkiverse/quarkus-artemis/pull/317
https://github.com/quarkiverse/quarkus-artemis/pull/318

PR #4659, opened 2023-10-31
ARTEMIS-4480: rationalise openwire session tx usage, ensure operation…

gtully:
… context completion callbacks are sequentially ordered to support exclusive consumer isolation

### DIFF
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
@@ -682,18 +682,7 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
    private void rollbackInProgressLocalTransactions() {
 
       for (Transaction tx : txMap.values()) {
-         AMQSession session = (AMQSession) tx.getProtocolData();
-         if (session != null) {
-            session.getCoreSession().resetTX(tx);
-            try {
-               session.getCoreSession().rollback(false);
-            } catch (Exception expectedOnExistingOutcome) {
-            } finally {
-               session.getCoreSession().resetTX(null);
-            }
-         } else {
-            tx.tryRollback();
-         }
+         tx.tryRollback();
       }
    }
 
@@ -729,7 +718,7 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
          for (SessionId sessionId : sessionIdMap.values()) {
             AMQSession session = sessions.get(sessionId);
             if (session != null) {
-               session.close();
+               session.close(fail);
             }
          }
          internalSession.close(false);
@@ -782,8 +771,6 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
 
       recoverOperationContext();
 
-      rollbackInProgressLocalTransactions();
-
       if (me != null) {
          //filter it like the other protocols
          if (!(me instanceof ActiveMQRemoteDisconnectException)) {
@@ -796,6 +783,20 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
          }
       } catch (InvalidClientIDException e) {
          logger.warn("Couldn't close connection because invalid clientID", e);
+      } finally {
+         // there may be some transactions not associated with sessions
+         // deal with them after sessions are removed via connection removal
+         operationContext.executeOnCompletion(new IOCallback() {
+            @Override
+            public void done() {
+               rollbackInProgressLocalTransactions();
+            }
+
+            @Override
+            public void onError(int errorCode, String errorMessage) {
+               rollbackInProgressLocalTransactions();
+            }
+         });
       }
       shutdown(true);
    }
@@ -1338,7 +1339,11 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
 
       @Override
       public Response processRollbackTransaction(TransactionInfo info) throws Exception {
-         Transaction tx = lookupTX(info.getTransactionId(), null, true);
+         Transaction tx = lookupTX(info.getTransactionId(), null);
+
+         if (tx == null) {
+            throw new IllegalStateException("Transaction not started, " + info.getTransactionId());
+         }
 
          final AMQSession amqSession;
          if (tx != null) {
@@ -1354,11 +1359,12 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
             if (amqSession != null) {
                amqSession.getCoreSession().resetTX(tx);
 
-               try {
-                  returnReferences(tx, amqSession);
-               } finally {
-                  amqSession.getCoreSession().resetTX(null);
-               }
+               tx.addOperation(new TransactionOperationAbstract() {
+                  @Override
+                  public void beforeRollback(Transaction tx) throws Exception {
+                     returnReferences(tx, amqSession);
+                  }
+               });
             }
          }
 
@@ -1406,6 +1412,9 @@ public class OpenWireConnection extends AbstractRemotingConnection implements Se
             }
          } else {
             if (tx != null) {
+               // returnReferences() is only interested in acked messages already in the tx, hence we bypass
+               // getCoreSession().rollback() logic for CORE which would add any prefetched/delivered which have
+               // not yet been processed by the openwire client.
                tx.rollback();
             }
          }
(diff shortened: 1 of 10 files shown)


gtully:
sorted some xa related tests, tx in prepared state not being active, pointed to an unnecessary change. investigating some transacted mirror issues now related to closed sessions.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -1338,7 +1339,11 @@ public Response processRemoveSubscription(RemoveSubscriptionInfo subInfo) throws
 
       @Override
       public Response processRollbackTransaction(TransactionInfo info) throws Exception {
-         Transaction tx = lookupTX(info.getTransactionId(), null, true);
+         Transaction tx = lookupTX(info.getTransactionId(), null);
+
+         if (tx == null) {
+            throw new IllegalStateException("Transaction not started, " + info.getTransactionId());
```
clebertsuconic, line 1345: I would use a logger-id from the MessageBundle.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
```diff
@@ -517,7 +517,6 @@ public void removeRolledback(MessageReference messageReference) {
    }
 
    public void addRolledback(MessageReference messageReference) {
-      currentWindow.decrementAndGet();
```
clebertsuconic, line 520: why? were you having issues here?

AntonRoskvist: It was a while since I looked at that, but I think this change will enable the consumer to fetch messages past the prefetchLimit. It happens when doing redeliveries, such that it can theoretically prefetch twice as many messages as it's supposed to.

Again, saw this quite some time back so I could be mistaken...

gtully:
I need to go back to the drawing board, the ordered executor should keep completions sequential, there must be something else at play

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
AntonRoskvist: ...yes, that is indeed the case. Verified by adding messages to rollback in "testRedeliveredMessageNotOverflowingPrefetch" and printing deliveryCount instead of asserting.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
gtully: there is currently no use of the bundle in any of the exceptions, that can be another PR if really necessary

gtully:
part of my difficulty was misreading the clearContext at: https://github.com/apache/activemq-artemis/blob/a0784e3b8a75116424dd345037426e38a36f85d0/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/OperationContextImpl.java#L319 in terms of the clear method on the context itself, which I have renamed reset! For a future self or other. The clear is clearContext refers to the thread local only, which is ok.

clebertsuconic:
@gtully this looks good to me.

But can you rebase and run the testsuite again please? if all good I think you should merge this.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
AntonRoskvist: @gtully unless adressed elsewhere, this will still cause a regression on the openwire consumers prefetch-limit...

gtully: will investigate a bit more here, thanks

gtully: in testRedeliveredMessageNotOverflowingPrefetch: currentWindow goes negative and message 14 does not arrive! the logic that puts messaged back on the delivered list and pulls from the transaction is in the mix here. That list has 11 messages, prefetch + the one rolled back message. I think currentWindow should only be decremented by the 1 rolledback message. 
This is very tricky, but it the root problem may be that there is a single transaction now, doing the rollback of the client message and the prefetched messages.

gtully:
doing some more work to get to the bottom of the currentWindow issue

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/amq/AMQConsumer.java
gtully: @AntonRoskvist I pushed a new test that fails with a revert of line 520, the decrement. I don't yet know what is at fault, but the root cause is the combination of rolledback and delivered messages in a single transaction. 
we decrement 11 times in stead of 1 if the session transaction is in play. 
it may be that cancelRefs needs its own transaction all the time, the comments at: https://github.com/apache/activemq-artemis/blob/6e28e4683458490ab0e3990ba8f35e9c6531c461/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java#L2234
may have a hint - it speaks to the same problem.
it may be that cancleRefs should not have a tx at all, or a new one all the time that has no IO may work ok.

AntonRoskvist: Quite interesting... I have actually noticed queue counters (metrics) being incorrect from time to time on queues using OW consumers... in particular ones using XA-transactions. This type of application is getting exceedingly rare in the envs I manage, and it has not happened often enough to warrant further investigation, but it is something I have seen multiple times. Guessing it could have a common source to this as well.

For the added test you mentioned though, I found it to behave pretty much as I would expect it to, besides prefetch being of-by-one (which I guess might be expected given the rollback, not sure). 

The error I am getting is caused by a null value returned from the receive-call, which happens because some message gets moved over to DLQ after exceeding "maximumRedeliveries"... so this is the expected result given the queue is empty.

Setting `connection.getRedeliveryPolicy().setMaximumRedeliveries(-1);` makes the test pass.

Additionally I ran with `queueControl.getDeliveringCount()` printed for each loop iteration and prefetch never exceeds prefetchPolicy + 1.


I'm not sure about this next part at all: 
From what I recall from last time I looked at this, I think I interpreted the incorrect `currentWindow` and `prefetch` values to be caused by from some difference in how OW handled redeliveries compared to Core... something about the OW client holding on to the message for each redelivery but also notifying the broker each time about the redelivery, whereas for Core the message actually "bounces" back and forth for each delivery.

gtully: Thanks. There was something weird, but messages going to the DLQ was part of it.
The other part is the returnReferences that puts messages back in delivering rather than back on the queue on a tx rollback. With the core session in play, that also had prefetched messages in the tx, and that was breaking the returnReferenced logic and causing to many decrements on the currentWindow. 
Thanks for picking up on that change :-)

gtully: I played with doing the session cancelRef in a new transaction, but multiple transactions can interact with IO and become out of sync due to any operation context task clearing the current context. When the delivery count is persisted, there is IO. That was to root cause of the original investigation, so I don't want to introduce that possibility again. I think what i have now is good.

gtully:
full test run looks good

PR #4667, opened 2023-11-09
ARTEMIS-4495 Improving data retention documentation

clebertsuconic:


artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.java
```diff
@@ -1988,11 +1988,27 @@ void resetUser(@Parameter(name = "username", desc = "Name of the user") String u
    @Operation(desc = "forces the broker to reload its configuration file", impact = MBeanOperationInfo.ACTION)
    void reloadConfigurationFile() throws Exception;
 
+   /**
```
clebertsuconic, line 1991: I'm adding javadoc here, since I am adding a hyperlink from the doc towards the javadoc.

PR #4669, opened 2023-11-12
ARTEMIS-4476 Validating process to remove orphaned consumers

clebertsuconic:


artemis-hawtio/artemis-plugin/src/main/webapp/plugin/js/components/consumers.js
```diff
@@ -106,7 +106,8 @@ var Artemis;
               {name: "Messages Acknowledged", visible: false},
               {name: "Messages Acknowledged awaiting Commit", visible: false},
               {name: "Last Delivered Time", visible: false},
-              {name: "Last Acknowledged Time", visible: false}
+              {name: "Last Acknowledged Time", visible: false},
+              {name "status", visible: false}
```
jbertram, line 110: Missing a colon here. It should be `{name: "status", visible: false}`

clebertsuconic: thanks a lot.. that's why it was not working.

PR #4671, opened 2023-11-13
ARTEMIS-4499 fix ThreadCreateAction so it works properly with Securit…

ehsavoie:
…yManager

Issue: https://issues.apache.org/jira/browse/ARTEMIS-4499


jbertram:
I think this is OK for now, but pretty soon we're going to need to rip this all out since the SecurityManager has been deprecated for removal.

PR #4676, opened 2023-11-14
ARTEMIS-4504: use cli script to fix examples shutdown config issue

gemmellr:


gemmellr:
Passed the PR build a couple times on my fork, and a full test suite build plus the examples verification script.

PR #4677, opened 2023-11-15
ARTEMIS-4505 Cleanup page transactions on startup of the broker

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PageTransactionInfoImpl.java
```diff
@@ -67,11 +67,27 @@ public final class PageTransactionInfoImpl implements PageTransactionInfo {
 
    private List<LateDelivery> lateDeliveries;
 
+   /** To be used during by the RebuildManager.
+    *  When reading transactions not found transactions are marked as done. */
+   private boolean orphaned;
+
+
```
tabish121, line 74: Extra newline added, seems like some extras were removed down below so probably want to prune this one

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingManagerImpl.java
```diff
@@ -583,22 +591,65 @@ public void forEachTransaction(BiConsumer<Long, PageTransactionInfo> transaction
...
+      }
+      AtomicBoolean proceed = new AtomicBoolean(true);
+      currentStoreMap.forEach((a, b) -> {
+         if (!b.getCursorProvider().isRebuildDone()) {
+            logger.warn("cannot proceed on cleaning up page transactions as page cursor for {} is not done rebuilding it", b.getAddress());
+            proceed.set(false);
+         }
+      });
+
+      if (!proceed.get()) {
```
tabish121, line 635: Is there some additional handling to retry or is this cleanup left for another restart?  It's unclear to me what the state is after the code determines it can't proceed.

clebertsuconic: In case one of the PageStore cleanup failed for any reason, I would prefer to never do this cleanup, and only try it at a next restart.

It shouldn't really happen, but I wanted to add this check here.

tabish121: Might be worth a comment explaining the thinking as I starred at it for awhile trying to figure out where it was being restarted etc.

clebertsuconic: I just did. thanks

PR #4678, opened 2023-11-15
ARTEMIS-4506 restore MQTT sub queue name syntax

jbertram:
Starting with 2.28.0, the broker doesn't translate the character `/` to the configured wildcard delimiter (i.e. `.` by default) when creating subscription queues for MQTT clients.

This commit fixes that regression and restores the proper translation.


artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTUtil.java
```diff
@@ -137,11 +138,11 @@ public class MQTTUtil {
 
    public static final int DEFAULT_MAXIMUM_PACKET_SIZE = MAX_PACKET_SIZE;
 
-   public static String convertMqttTopicFilterToCoreAddress(String filter, WildcardConfiguration wildcardConfiguration) {
-      return convertMqttTopicFilterToCoreAddress(null, filter, wildcardConfiguration);
+   public static String convertMqttTopicFilterToCore(String filter, WildcardConfiguration wildcardConfiguration) {
+      return convertMqttTopicFilterToCore(null, filter, wildcardConfiguration);
    }
 
-   public static String convertMqttTopicFilterToCoreAddress(String prefixToAdd, String filter, WildcardConfiguration wildcardConfiguration) {
```
clebertsuconic, line 144: I liked the older name better.. I thought it was more descriptive.

Or there was another reason for the change?

PR #4681, opened 2023-11-17
ARTEMIS-4510 - Add auto-create-destination logic to diverts

AntonRoskvist:
This enables the use of dynamic routing decisions within the transformer by setting the message address. It also covers for a rare problem where if any of the forwarding addresses are removed during runtime, such as from auto-delete, the message would get silently dropped.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/DivertImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/DivertImpl.java
@@ -16,12 +16,15 @@
  */
 package org.apache.activemq.artemis.core.server.impl;
 
+import org.apache.activemq.artemis.api.core.AutoCreateResult;
 import org.apache.activemq.artemis.api.core.Message;
+import org.apache.activemq.artemis.api.core.QueueConfiguration;
 import org.apache.activemq.artemis.api.core.RoutingType;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.filter.Filter;
 import org.apache.activemq.artemis.core.persistence.StorageManager;
 import org.apache.activemq.artemis.core.postoffice.PostOffice;
+import org.apache.activemq.artemis.core.server.ActiveMQMessageBundle;
 import org.apache.activemq.artemis.core.server.ComponentConfigurationRoutingType;
 import org.apache.activemq.artemis.core.server.Divert;
 import org.apache.activemq.artemis.core.server.RoutingContext;
@@ -136,6 +139,21 @@ public class DivertImpl implements Divert {
 
             // We call reencode at the end only, in a single call.
             copy.reencode();
+
+            if (postOffice.getAddressSettingsMatch(copy.getAddress()).isAutoCreateDivertDestination()) {
+
+               if (copy.getRoutingType() != null) {
+                  AutoCreateResult autoCreateResult = context.getServerSession().checkAutoCreate(
+                     new QueueConfiguration(copy.getAddress())
+                        .setRoutingType(copy.getRoutingType())
+                        .setDurable(copy.isDurable()));
+
+                  if (autoCreateResult == AutoCreateResult.NOT_FOUND) {
+                     throw ActiveMQMessageBundle.BUNDLE.noSuchQueue(copy.getAddressSimpleString());
+                  }
+               }
+            }
+
          } else {
             copy = message;
          }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
@@ -147,6 +147,8 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
 
    public static final boolean DEFAULT_ENABLE_INGRESS_TIMESTAMP = false;
 
+   public static final boolean DEFAULT_AUTO_CREATE_DIVERT_DESTINATIONS = false;
+
    static {
       metaBean.add(AddressFullMessagePolicy.class, "addressFullMessagePolicy", (t, p) -> t.addressFullMessagePolicy = p, t -> t.addressFullMessagePolicy);
    }
@@ -530,6 +532,11 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
       metaBean.add(Integer.class, "idCacheSize", (t, p) -> t.idCacheSize = p, t -> t.idCacheSize);
    }
    private Integer idCacheSize = null;
+
+   static {
+      metaBean.add(Boolean.class, "autoCreateDivertDestination", (t, p) -> t.autoCreateDivertDestination = p, t -> t.autoCreateDivertDestination);
+   }
+   private Boolean autoCreateDivertDestination = null;
 
    static {
       metaBean.add(Integer.class, "queuePrefetch", (t, p) -> t.queuePrefetch = p, t -> t.queuePrefetch);
@@ -1281,6 +1288,15 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
       return this;
    }
 
+   public boolean isAutoCreateDivertDestination() {
+      return autoCreateDivertDestination != null ? autoCreateDivertDestination : AddressSettings.DEFAULT_AUTO_CREATE_DIVERT_DESTINATIONS;
+   }
+
+   public AddressSettings setAutoCreateDivertDestination(final boolean autoCreateDivertDestination) {
+      this.autoCreateDivertDestination = autoCreateDivertDestination;
+      return this;
+   }
+
    /**
     * Merge two AddressSettings instances in one instance
     *
@@ -1998,7 +2014,9 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
          return false;
       if (!Objects.equals(idCacheSize, that.idCacheSize))
          return false;
-      return Objects.equals(queuePrefetch, that.queuePrefetch);
+      if (!Objects.equals(queuePrefetch, that.queuePrefetch))
+         return false;
+      return Objects.equals(autoCreateDivertDestination, that.autoCreateDivertDestination);
    }
 
    @Override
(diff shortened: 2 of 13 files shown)


jbertram:
This is an interesting improvement which I think has some merit. However, I have two concerns:

- This changes the default behavior of every divert. This may break existing users who actually depend on the current behavior. Therefore, this functionality should be configurable.
- This will have a negative impact on performance. _Every_ time a message is diverted `checkAutoCreate` will be invoked with a new `QueueConfiguration`. This will increase CPU utilization and GC pressure. Perhaps the divert could invoke `checkAutoCreate` once when it was deployed and then also during the execution of `route` if and only if a transformer was invoked.

AntonRoskvist:
Thanks for your feedback @jbertram 

I'd argue that the behavior already is configurable though, through the "auto-create"-configs in the address-settings...

As for the performance implication, I agree that this could have some additional cost associated with it... however the same operation is carried out during normal routing within the postOffice, twice... at `PostOfficeImpl#checkAddress()` and `PostOfficeImpl#simpleRoute()`.

The reason they are not applied for the diverted messages is because the message copies does not have a reference back to their original `ServerSession`... don't know if this is safe, but perhaps the Divert could do this instead:

`postOffice.route(copy, new RoutingContextImpl(context.getTransaction()).setReusable(false).setRoutingType(copy.getRoutingType()), false);`

 --->

`postOffice.route(copy, new RoutingContextImpl(context.getTransaction()).setReusable(false).setRoutingType(copy.getRoutingType()).setServerSession(context.getServerSession()), false);`

That way the currently implemented auto-create-logic within PostOffice would apply instead. Thoughts on this?

And just to be clear, I'd be open to re-implement all together according to your suggestion if you think that makes more sense, just let me know.

jbertram:
You're right that the `auto-create-*` configs in the `address-settings` apply here. The problem is that they don't apply to this use-case _right now_, and this commit changes this fact which will potentially change behavior for folks who upgrade with no changes to their configuration. Furthermore, this commit also changes the behavior if the target does not exist. Previously no error would be thrown if the target address didn't exists, but now an error will be thrown. Such an error will not only impact the divert but the original send as well. From what I can tell **no** message will get routed if the divert's target doesn't exist and auto-creation fails. We just can't do that in a minor release.

Therefore there needs to be a new configuration element on the divert that disables this behavior by default.

There are security implications here as well. You're using `context.getServerSession()` for auto-creation which will use the original sender's credentials to auto-create the address and/or queue. This may not make sense for every use-case.

Lastly, there's a weird wrinkle with `org.apache.activemq.artemis.core.server.impl.ServerSessionImpl#checkAutoCreate` in that if the routing-type of the passed `QueueConfiguration` is `null` then it returns `AutoCreateResult.EXISTED` which can lead to some unexpected behavior with the divert's default routing-type (i.e. `STRIP`).

AntonRoskvist:
@jbertram 
Yeah that makes sense, your right.
I hope these changes are proper and address your points there. Otherwise just let me know.

Btw, I rebased against main as well.

AntonRoskvist:
This was accidentally closed it seems.

@jbertram please review the most recent changes when you have the time for it and see if they conform to what you requested.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -1679,7 +1699,8 @@ public int getEncodeSize() {
          BufferHelper.sizeOfNullableInteger(idCacheSize) +
          BufferHelper.sizeOfNullableSimpleString(pageFullMessagePolicy != null ? pageFullMessagePolicy.toString() : null) +
          BufferHelper.sizeOfNullableInteger(prefetchPageBytes) +
-         BufferHelper.sizeOfNullableInteger(prefetchPageMessages);
+         BufferHelper.sizeOfNullableInteger(prefetchPageMessages) +
+         BufferHelper.sizeOfNullableBoolean(autoCreateDivertDestination);
```
clebertsuconic, line 1703: I'm not sure this is needed any longer... we are not changing the wires for the old methods.

Preetty much just use JSON for anything new now.. older stuff should stay as is.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -1601,6 +1617,10 @@ public void decode(ActiveMQBuffer buffer, boolean tryCompatible) {
       if (buffer.readableBytes() > 0) {
          prefetchPageMessages = BufferHelper.readNullableInteger(buffer);
       }
+
+      if (buffer.readableBytes() > 0) {
```
clebertsuconic, line 1621: I don't think we should change these.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -1835,6 +1856,8 @@ public void encode(ActiveMQBuffer buffer) {
       BufferHelper.writeNullableInteger(buffer, prefetchPageBytes);
 
       BufferHelper.writeNullableInteger(buffer, prefetchPageMessages);
+
+      BufferHelper.writeNullableBoolean(buffer, autoCreateDivertDestination);
```
clebertsuconic, line 1860: don't think we should change these.

clebertsuconic: new version now is using the JSON encoded on the journal.. the older stuff is just for the journals... so these are not needed I think.

AntonRoskvist: @clebertsuconic Thanks, I have made the changes you suggested. The only thing I'm noticing now is that if I inject this setting in some current persistence tests they start failing, though I guess that might be expected with this new method?

One example of a test that would fail is for instance:
`org.apache.activemq.artemis.tests.integration.persistence.AddressSettingsConfigurationStorageTest#testStoreSecuritySettings()` where if I add some random previous address setting (like `setAutoDeleteAddresses(true)`) the test still passes, whereas if I add `setAutoCreateDivertDestination(true)` the test fails.

Is that right/expected or am I missing something?

clebertsuconic: I will run the tests and take a look... will comment back here after my findings by the end of my day .

clebertsuconic: @AntonRoskvist Can you add your change and rebase your PR on top of https://github.com/apache/activemq-artemis/pull/4854


The tests were still using the old method. I am removing *any* usage from the AddressSettings usage, and I am now making it *clearer* with a big javadoc to not have any more additions.


The reason that you needed to add a field now was a reason I added the JSON format. I got tired to see that method's wire being changed over time.  That got even worse when I played with my private branches when I was only bringing the newer field and not the previous field.

So, if I added field A and B on apache/master

and only Brough field B to a private branch, my wire would be inconsistent with apache/main.  It was being clusterFreak to manage it :)


If you could bring my commit to your PR (Rebase as I said), and bring any changes you like, it should now work.

clebertsuconic: @AntonRoskvist I have merged #4854... if you rebase against main, you can add anything to the tests you mentioned and it should work now.

I have removed any usage of the older encoder.. just keeping the decoder for compatibility with older journals.

AntonRoskvist: @clebertsuconic Yes, looking good after rebasing. Thanks!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/DivertImpl.java
```diff
@@ -136,6 +139,21 @@ public void route(final Message message, final RoutingContext context) throws Ex
 
             // We call reencode at the end only, in a single call.
             copy.reencode();
+
+            if (postOffice.getAddressSettingsMatch(copy.getAddress()).isAutoCreateDivertDestination()) {
```
clebertsuconic, line 143: Instead of getAddressSettingsMatch on every message (which will actually have a negative perf impact for any case), you should rather check once, and register a listener on the HierarchicalRepositoryChangeListener.

This is how actually page updates the maxPage data after anything is changed:

https://github.com/apache/activemq-artemis/blob/89c3a627e9ca2b1ef48a6411d372c32f90a6f9ed/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingManagerImpl.java#L192-L194


you could also have such listener done on a higher level and call every Divert to recalculate attributes (that would be less risky of leaks, say a Divert being unregistered would leak on the Listener).

clebertsuconic:
@jbertram I actually thought we would create auto-created destinations on Divert. I kind of agree with @AntonRoskvist here that we already have an auto-create attribute. 

If you're Diverting to an non existing destination, you could create it.


Although.. going a step back.. why you would create a Divert to a non existing destination? a Divert is from one destination to another, right? why so you need auto-create on diverts?

AntonRoskvist:
@clebertsuconic 
The use-case for me (and another user I spoke to on the slack channel) is to be able to compute a destination during runtime inside the transformer. The other user was after something like "If message has header X, address = somePrefix + headerX"

I want to use this for a global routing system that would look something like this:
1. Divert wildcard-matches destinations following a naming system that includes a site name (this matches a few hundred destinations, destinations are created dynamically by clients).
2. Divert adds some metadata to messages and aggregates them to a common/aggregate destination (no auto-create needed here)
3. A core bridge picks these messages up and sends them to their destination site.
4. A Divert on the destination site routes messages from the aggregate destination and back to their original destination name (or in some cases another related destination name), based on logic in a transformer (ultimately changing the messages address)

I do not want to use the __AMQ_ORIGINAL_DESTINATION headers for this as in the event of some network issues or similar messages would end up on a DLA whereby said header changes to the aggregate queue name and identification afterwards becomes... labour intensive...

AntonRoskvist:
@clebertsuconic @jbertram Let me know if you want me to either: 
Remove the condition and new address setting for auto-creation and simply auto create if the destination is configured for it
Or to add a subscribe to changes by registering a listener on the HierarchicalRepositoryChangeListener.

I can go either way but personally I prefer the first option... the risk of breaking a preexisting use-case where sending directly to a non existing destination should create it but diverting to it should silently fail seems small... perhaps a note about this change of semantics could be added to the versions doc instead?

clebertsuconic:
@AntonRoskvist being 100% open about this... I have a disagreement point with @jbertram on this. I think if you Divert to an address foo, that address should be part of the verification of the auto-create. Justin thinks that's a breaking change, and I see it as a bug.... 

Maybe you could open a discuss thread on the dev-list to talk about it and find consensus among other people?

IMO we should just document the change and tell users if they must route to nowhere they should have either a disabled auto-create on the Address-settings.. or an empty MultiCast address (eg address ReturnedItems or Orders without a subscription).



One thing I 100% agree with Justin though is that we shouldn't penalize the regular case with checking auto-create every time. We need to find an optimal way of doing it.

AntonRoskvist:
@clebertsuconic Started a thread on the dev mailing list to see if we can reach some consensus there, flagging this as draft until then.

PR #4683, opened 2023-11-17
ARTEMIS-3932 Deprecate and replace addAddressSettings with json version

clebertsuconic:
As I worked through implementing a more generic JSON marshaller, I tried using reflection through BeanUtils and other ways however the endresult was always worse as there were a few caveats that were not as easy to accomplish.

For that reason I went to a declarative appraoch where I define a meta-data object on AddressSettings and AddressSettingsInfo and reuse the metadata in a few other places.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
@@ -88,6 +88,7 @@ import org.apache.activemq.artemis.core.messagecounter.MessageCounterManager;
 import org.apache.activemq.artemis.core.messagecounter.impl.MessageCounterManagerImpl;
 import org.apache.activemq.artemis.core.persistence.StorageManager;
 import org.apache.activemq.artemis.core.persistence.config.PersistedAddressSetting;
+import org.apache.activemq.artemis.core.persistence.config.PersistedAddressSettingJSON;
 import org.apache.activemq.artemis.core.persistence.config.PersistedConnector;
 import org.apache.activemq.artemis.core.persistence.config.PersistedSecuritySetting;
 import org.apache.activemq.artemis.core.postoffice.Binding;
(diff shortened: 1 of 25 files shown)


clebertsuconic:
Notice: I created this PR with the current status of my work before a week break on American Thanksgiving. 

I won't be able to do extensive work on this PR on the week between Nov-19th and Nov-25th. but I will certainly respond to small inquiries.

clebertsuconic:
also: as I commented on the commit, the amount of work to use BeanUtils would be a lot larger and not satisfactory.

I still have my branch saved from when I was trying to use BeanUtils, but I would have a much larger set of changes for a small gain over what I'm doing here.

For instance: the getters would always go through the default values instead of getting the attributes directly. and it wouldn't be possible to generate the same object back from a "marshaller".

(well: wouldn't be possible is a strong word here, lets just say, I could do it with a lot more work involved. Hence I will use the declarative approach as I mentioned here).

clebertsuconic:
I am marking this as draft as I will refactor the journal record where I'm using a JSON object to be reused for other cases.

clebertsuconic:
hmmm... actually I will not do that..

but I will still change the Address and Queue records to use the JSON object format.

clebertsuconic:
I gave up on the idea. I was going to write a meta-record..which would simplify usage, but it would probably increase memory of the broker if those settings are over-used. so I'm giving up the idea.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -3596,6 +3534,41 @@ public void addAddressSettings(final String address,
...
+   public String addAddressSettings(String address, String addressSettingsConfigurationAsJson) throws Exception {
+      if (AuditLogger.isBaseLoggingEnabled()) {
+         AuditLogger.addAddressSettings(this.server, addressSettingsConfigurationAsJson);
+      }
+      checkStarted();
+
+      clearIO();
+
+      try {
+         // when the QueueConfiguration is passed through createQueue all of its defaults get set which we return to the caller
```
jbertram, line 3547: This comment is incorrect.

PR #4684, opened 2023-11-20
Move the queue consumer count inside the initialization block

stefanom:
Artemis currently fails to cleanup federated queues after consumers disconnect from an address.

This happens because the `remoteQueueConsumer` counter is incremented many more times than the number of consumers. This mismatch prevents proper cleanup in the `removeRemoteConsumer()` method below because the `if (remoteQueueConsumer.decrementCount() <= 0) {` is never invoked.

I have manually validated this change with a fully connected federation of 3 regions using MQTT publishers and subscribers and auto-delete enabled on the federation: when the subscriber leaves, the local queue is eliminated immediately and the federation queues (all 6) are closed and removed after a few minutes.

NOTE: this is a severe liability and consistently leads to OutOfMemory issues no matter the JVM heap size in environments that create a lot of ephemeral addresses.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedAbstract.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedAbstract.java
@@ -117,6 +117,7 @@ public abstract class FederatedAbstract implements ActiveMQServerBasePlugin {
             remoteQueueConsumer = new FederatedQueueConsumerImpl(federation, server, transformer, key, upstream, callback);
             remoteQueueConsumer.start();
             remoteQueueConsumers.put(key, remoteQueueConsumer);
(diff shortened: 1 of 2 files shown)


jbertram:
Any chance you could write a test for this? We need a way to verify this fixes the problem and to mitigate future regressions.

stefanom:
I would but I need assistance. Right now it takes several minutes for Artemis to cleanup after itself in my manual test. 

Is there a way to force Artemis to "garbage collect unused queues" so that the test can be made consistent and fast?

stefanom:
I was able to find a way to trigger the queue and address garbage collection via the `PostOffice` and now the test passes and it validates that the fix works (because without the fix the test does not pass).

tabish121:
Closing this as abandoned as comments where never responded to.  Reopen or create a new PR to address comments.

PR #4686, opened 2023-11-21
ARTEMIS-4513 HTTP request logging not working

jbertram:


artemis-unit-test-support/src/main/java/org/apache/activemq/artemis/utils/ThreadLeakCheckRule.java
```diff
@@ -295,6 +295,8 @@ private boolean isExpectedThread(Thread thread) {
          return true;
       } else if (threadName.contains("GC Daemon")) {
          return true;
+      } else if (threadName.contains("org.eclipse.jetty.util.RolloverFileOutputStream")) {
```
clebertsuconic, line 298: Isn't this hiding a thread leakage? can't we just close the thread instead?

clebertsuconic: Or move this test to the isolated tests if you can't figure out how to close the thread.

clebertsuconic: I don't mean to move the whole test into integration-tests-isolated. but rather write just this test into the isolated package.. 


This is definitely causing a thread leakage.. and this thread could be bothering other tests.

I strongly recommend against it.

jbertram: The thread is specific to the HTTP request logging from Jetty. Looking at the Jetty source code there appears to be no way to close it. I'll move the test into the isolated module.

PR #4688, opened 2023-11-23
ARTEMIS-4164: add acceptor sslAutoReload option to have the broker wa…

gtully:
…tch key and trust store paths for modifications and automatically call reload


gtully:
It is a little more complicated b/c management is optional but I am thinking it may be a good idea to first attempt reload via acceptorcontrol and if that does not exist revert to calling the acceptor directly. the benefit of going via the management control is that there is an audit log for the reload op.

gtully:
turns out our managementservice is always created, I have use AccptorControll.reload which gives us a consistent audit point. that is better.

PR #4692, opened 2023-11-28
ARTEMIS-4517 - cancel message references in queue sequence order

gtully:
this needs more eyes please


gtully:
pushed an update to rename the idComparator the queueSequenceComparator

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/MessageReference.java
```diff
@@ -142,4 +142,7 @@ default boolean skipDelivery() {
     */
    long getPersistentSize() throws ActiveMQException;
 
+   long getQueueSequence();
```
clebertsuconic, line 145: just a suggestion! I would just call it getSequence here... Like the reference Queue, since a reference will belong to a queue anyway. Nit-pick probably.. do it if you like the suggestion!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/AbstractProtocolReference.java
```diff
@@ -28,6 +28,7 @@
 public abstract class AbstractProtocolReference extends LinkedListImpl.Node<MessageReferenceImpl> implements MessageReference {
 
    private HashMap<Class, Object> protocolDataMap;
+   protected volatile long queueSequence = 0;
```
clebertsuconic, line 31: sequence, if you take my suggestion

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/MessageReferenceImpl.java
```diff
@@ -33,17 +33,21 @@
  */
 public class MessageReferenceImpl extends AbstractProtocolReference implements MessageReference, Runnable {
 
-   private static final MessageReferenceComparatorByID idComparator = new MessageReferenceComparatorByID();
+   private static final MessageReferenceComparatorByQueueSequence queueSequenceComparator = new MessageReferenceComparatorByQueueSequence();
 
-   public static Comparator<MessageReference> getIDComparator() {
-      return idComparator;
+   public static Comparator<MessageReference> getQueueSequenceComparator() {
```
clebertsuconic, line 38: sequenceComparator if you take my suggestion

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/MessageReference.java
clebertsuconic: another possible name would be order.... it doesn't matter if you take order or sequence. just putting  out an alternative.

gtully: makes sense, will fix it up. thanks

clebertsuconic:
@gtully beware I added a change to your branch. please update your local fork.. feel free to squash the change if you prefer

clebertsuconic:
I think this is ready to merge...

PR #4694, opened 2023-11-30
ARTEMIS-4476 Possible isolated consumer with OpenWire

clebertsuconic:


tests/integration-tests-isolated/src/test/java/org/apache/activemq/artemis/tests/integration/isolated/client/ConnectionDroppedTest.java
```diff
@@ -498,4 +500,64 @@ public void beforeCreateSession(String name,
...
+            logger.warn(e.getMessage(), e);
+         } finally {
+            done.countDown();
+         }
+      });
+
+      Assert.assertTrue(beforeCreateCalled.await(5, TimeUnit.MINUTES));
+
+      server.getRemotingService().getConnections().forEach(r -> {
+         r.fail(new ActiveMQException("this is a simulation"));
```
clebertsuconic, line 551: this would block with the current version as this is calling shutdown. 

I'm blocking the thread on purpose there.


I could not reproduce the leakage here. But I believe the change I made would be better in this case.


I'm open to suggestions.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -761,7 +761,11 @@ public void fail(ActiveMQException me, String message) {
 
       final ThresholdActor<Command> localVisibleActor = openWireActor;
       if (localVisibleActor != null) {
-         localVisibleActor.shutdown(() -> doFail(me, message));
+         localVisibleActor.requestShutdown();
+      }
+
+      if (executor != null) {
+         executor.execute(() -> doFail(me, message));
```
clebertsuconic, line 768: @gtully notice that the executor is used by the Actor. Meaning it will always execute after sthudown is called, using the same thread it would have used with the actor.

This is just a safer option.

clebertsuconic:
replaced by https://github.com/apache/activemq-artemis/pull/4700

PR #4695, opened 2023-11-30
fast reconnect openwire test that fails

gtully:


gtully:
the mods to logging are not necessary, just keep it cleaner.
this now needs investigation to find the root cause.

clebertsuconic:
replaced by https://github.com/apache/activemq-artemis/pull/4700

@gtully take a look please?

PR #4696, opened 2023-11-30
ARTEMIS-4520 JMSContext.acknowledge() doesn't work if last message received is null

jbertram:


tabish121:
This looks good assuming all the current tests still pass

jbertram:
Test-suite looks good. There were a few unrelated failures that all succeeded locally when retried.

PR #4699, opened 2023-12-01
ARTEMIS-4522 Dedicated thread pool for flow-control-executor

MrEasy:
Adds a dedicated thread pool for flow-control-executor to resolve slow-consumer-handling not being able to run in good time if common thread-pool is exhausted

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
@@ -137,6 +137,7 @@ public final class ServerLocatorImpl implements ServerLocatorInternal, Discovery
 
    private transient Executor threadPool;
 
+   private transient Executor flowControlPool;
    private transient ScheduledExecutorService scheduledThreadPool;
 
    private transient DiscoveryGroup discoveryGroup;
@@ -188,11 +189,14 @@ public final class ServerLocatorImpl implements ServerLocatorInternal, Discovery
       ActiveMQClient.clearThreadPools();
    }
 
+
    private synchronized void setThreadPools() {
       if (threadPool != null) {
          return;
       } else if (config.useGlobalPools) {
          threadPool = ActiveMQClient.getGlobalThreadPool();
+
+         flowControlPool = ActiveMQClient.getFlowControlThreadPool(); //TODO  add option for config
 
          scheduledThreadPool = ActiveMQClient.getGlobalScheduledThreadPool();
       } else {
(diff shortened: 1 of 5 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
```diff
@@ -188,36 +189,49 @@ public static synchronized void clearThreadPools() {
...
+   private synchronized void setThreadPools()
+   {
       if (threadPool != null) {
          return;
-      } else if (config.useGlobalPools) {
+      }
+      else if (config.useGlobalPools) {
          threadPool = ActiveMQClient.getGlobalThreadPool();
 
+         flowControlPool = ActiveMQClient.getFlowControlThreadPool(); //TODO  add option for config
```
MrEasy, line 201: Any help/comment rg. this would be good - or deemed unneccessary?

clebertsuconic:
If you set a smaller thread pool you could get other issues...

you fixed one case by adding a new thread pool for flow control. 

When you start having issues with other pool saturated are you going to add another pool to that feature also?


Having a warn that the pool is saturated and suggestion to increase would probably be a better feature IMO.

MrEasy:
Hi,

> When you hit another feature not working because the thread pool is saturated, are you going to add another pool to that feature also?
If you reach the configured limit, I think everyone is clear that you will get delays until resources are freed. Our goal here was to address the issue that was more unexpected to us, which lead to the big 10 seconds delays based on the flow-control explicitly.
You could argue, that if going into slow-consumer-handling is not aiming for highest throughput anyhow, but having consumer-window-size>0 this has another disadvantage for us (see lengthy explanatin by Sascha in slack).
So goal for us here was to stick to the slow-consumer-handling, but avoiding the big 10 seconds influence of the full pool on it.

> Having a warn that the pool is saturated and suggestion to increase would probably be a better feature IMO.
I think that would also be an excellent improvement, since now it is not easily visible when the client thread-pool reaches its limit.

jbertram:
I've talked with @MrEasy (and one of his colleague's) in Slack about his use-case. This solution seems reasonable to me. If you want to constrain the "normal" thread pool but still ensure that flow-control works without extreme latency then we need something like this.

In the future we could probably ditch these kinds of things using Virtual Threads.

MrEasy:
> @MrEasy, you've got a handful of checkstyle violations:
> ```
> Also, your commits should be squashed.

Reduced violence and squashed

jbertram:
@MrEasy, I followed-up on this with #4708. The new PR has an additional commit which drastically simplifies the test and finishes the implementation with configuration properties, etc.

PR #4700, opened 2023-12-01
ARTEMIS-4523 Openwire leaving consumers isolated after reconnects

clebertsuconic:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/openwire/FastReconnectOpenWireTest.java
```diff
@@ -0,0 +1,281 @@
...
+      executor.shutdown();
+      assertTrue(executor.awaitTermination(60, TimeUnit.SECONDS));
+
+      Wait.assertEquals(0, () -> server.locateQueue(durableQueue).getConsumerCount());
+      assertTrue(errors.isEmpty());
+
+   }
+
+   @Test(timeout = 60_000)
+   public void testFastReconnectCreateConsumerNoErrorsNoClientIdA() throws Exception {
```
gtully, line 132: this test is now a duplicate of the one that comes after, it can safely be deleted.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -761,7 +760,11 @@ public void fail(ActiveMQException me, String message) {
 
       final ThresholdActor<Command> localVisibleActor = openWireActor;
       if (localVisibleActor != null) {
-         localVisibleActor.shutdown(() -> doFail(me, message));
```
gtully, line 764: afaik, this is the only call to shutdown(runnnable), if this pattern is better then I guess lets remove that method.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -779,11 +782,18 @@ private void doFail(ActiveMQException me, String message) {
...
          if (this.getConnectionInfo() != null) {
-            protocolManager.removeConnection(this.getConnectionInfo(), me);
+            protocolManager.removeConnection(this.getConnectionInfo().getClientId(), this);
          }
-      } catch (InvalidClientIDException e) {
-         logger.warn("Couldn't close connection because invalid clientID", e);
       } finally {
+         try {
+            // doing the disconnect on the finally for extra caution.
+            // if removeConnection throws an exception of any kind, this should still happen
```
gtully, line 790: I think delete this comment, it is clear by the use of finally

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -779,11 +782,18 @@ private void doFail(ActiveMQException me, String message) {
...
+            disconnect(false);
+         } catch (Throwable e) {
+            // it should never happen, but never say never
+            logger.warn("OpenWireConnection::disconnect failure", e);
```
gtully, line 794: also here, I think, Throwable ignored and logger.debug. If this warn appears in a log, there is nothing a user can do. it will only be noise.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireConnection.java
```diff
@@ -1817,14 +1804,18 @@ public Response processRecoverTransactions(TransactionInfo info) throws Exceptio
...
-               propagateLastSequenceId(sessionState, lastDeliveredSequenceId);
-            }
-            protocolManager.removeConnection(state.getInfo(), null);
-         } catch (Throwable e) {
-            // log
+         for (SessionState sessionState : state.getSessionStates()) {
+            propagateLastSequenceId(sessionState, lastDeliveredSequenceId);
+         }
+         if (state.getInfo() == null || state.getInfo().getClientId() == null) {
+            // I don't know how we could get to this state, I think it's impossible
```
gtully, line 1811: this is just some sensible defensive programming, I would drop the comment. for me it is debatable as to whether there is any value in a warn, possibly debug. What could a user do in this case? And what is the problem, we have not state to whack!

clebertsuconic: this used to throw a warning. I would rather have it logged and know about it.

PR #4670, opened 2023-11-13
ARTEMIS-4498 - Expose internal queues for management and observability

AntonRoskvist:
This was enabled by default in previous versions of the broker and where quite good for troubleshooting and observability purposes. Adding it here as a configurable option defaulting to the current behavior

PR #4702, opened 2023-12-04
ARTEMIS-4498 enable management for internal addresses & queues

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -1881,7 +1904,8 @@ public int getEncodeSize() {
          BufferHelper.sizeOfNullableInteger(idCacheSize) +
          BufferHelper.sizeOfNullableSimpleString(pageFullMessagePolicy != null ? pageFullMessagePolicy.toString() : null) +
          BufferHelper.sizeOfNullableInteger(prefetchPageBytes) +
-         BufferHelper.sizeOfNullableInteger(prefetchPageMessages);
```
clebertsuconic, line 1884: why? I added a new record for AddressSettings, right?

clebertsuconic: I wouldn't bother about this.. it's not going to be used.. the new record would be used.


Although shouldn't we always enable management for internal addresses?

clebertsuconic:
Can't we just always allow management of internal queues? that was my point on the other PR.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
jbertram: Indeed. I'll remove those changes.

jbertram:
After thinking about this more I think we should go with #4670.

PR #4704, opened 2023-12-06
NO-JIRA Adding log.info back into test

clebertsuconic:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/MQTT5Test.java
```diff
@@ -70,7 +75,7 @@ public void testSimpleSendReceive() throws Exception {
       subscriber.setCallback(new DefaultMqttCallback() {
          @Override
          public void messageArrived(String topic, MqttMessage message) {
-            System.out.println("Message received from " + topic + ": " + message);
+            logger.info("Message received from topic {}, message={}", topic, message);
```
clebertsuconic, line 78: @jbertram I think you meant this as the compilation fix?

PR #4705, opened 2023-12-06
ARTEMIS-4527 - Redistributor race when consumerCount reaches 0 in clu…

AntonRoskvist:
…ster

This is a very rare bug but when triggered, messages in the queue with 0 consumers will have the redistributors loop messages between some or all brokers in a cluster as fast as they can manage, until either some system resource or the clusterBridges producerFlowControl is reached. Will keep happening until consumers are added or cluster bridges are restarted.

I don't have a test for this but instead added a reproducer that works with a considerable amount of tweaks. Comments in the reproducer explains how to run it. The reproducer is _not_ a valid or reasonable use case... it builds on some unrelated work I did that accidentally triggered this. I have seen this multiple times in a production environment over the course of several years though, I've just been unable to reproduce it outside of production before accidentally stumbling on it recently.

Problem occurs when CONSUMER_CREATED notification arrive before the BINDING_ADDED notification.
When that happens the consumerCount for RemoteBinding is incorrect (something like 1-2 lower than actual consumerCount value).

Then when consumers disconnect, all are registered properly and RemoteBinding gets a negative consumerCount. The `isHighAcceptPriority` used by the redistributor checks for consumerCount == 0 but since count is now negative it returns as a valid destination.

Fix adds synchronization on the postoffice when processing createConsumer so then the previously issued addBinding for sure is done before continuing.

I also added double safety in the RemoteQueueBinding by not lowering consumerCount below 0 and also checking for consumerCount <= 0 instead of consumerCount == 0, though neither of these should really be necessary if the cluster notifications always arrive in the correct order.

If anyone can figure out a consistent way to trigger this issue I'd be happy to add it. Regardless, if the changes look good otherwise I think the reproducer should be removed rather than merged, leaving it here for verification purposes.

One final though is that perhaps all of the sort of create/add/remove operations in the `org.apache.activemq.artemis.core.protocol.core.ServerSessionPacketHandler` should be synchronized?
Something building on the current pattern of:
`onMessagePacket()`
```
switch
   fast1:
      fast1Stuff();
   fast2:
     fast2Stuff();
  default:
     slow()

slow:
switch
   slow1:
      slow1Stuff();
   slow2:
      slow2Stuff();
   default:
     synchronizedStuff()
     
synchronizedStuff:
switch
  ...
  ...
```


jbertram:
Checkstyle violations:
```
[INFO] There are 5 errors reported by Checkstyle 10.11.0 with /home/runner/work/activemq-artemis/activemq-artemis/tests/integration-tests/../../etc/checkstyle.xml ruleset.
Error:  src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/UnsynchronizedClusterReproducer.java:[123,49] (whitespace) WhitespaceAround: '{' is not followed by whitespace.
Error:  src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/UnsynchronizedClusterReproducer.java:[137,77] (whitespace) WhitespaceAround: '{' is not followed by whitespace.
Error:  src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/UnsynchronizedClusterReproducer.java:[153,40] (whitespace) WhitespaceAround: '{' is not followed by whitespace.
Error:  src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/UnsynchronizedClusterReproducer.java:[223,37] (whitespace) WhitespaceAround: '{' is not followed by whitespace.
Error:  src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/UnsynchronizedClusterReproducer.java:[235,37] (whitespace) WhitespaceAround: '{' is not followed by whitespace.
```

AntonRoskvist:
Fixed the checkstyle violations... though even with them fixed, I do not intend to keep the` src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/UnsynchronizedClusterReproducer.java` in this commit unless someone really wants it. I'm hoping that someone can use it and the other changes here to help me figure out a more consistent and proper way to demonstrating this bug...

jbertram:
Do you have any insight into how/why the `CONSUMER_CREATED` notification is being sent before the `BINDING_ADDED`? Theoretically the `BINDING_ADDED` should always be sent before the `CONSUMER_CREATED` even if they are done in different threads since any operation should block waiting for the `createQueue` to return before continuing on to create the consumer.

AntonRoskvist:
No, I have only been able to get an idea of what happens after the fact... the window of opportunity for this to happen is really slim... In fact, early in my troubleshooting I tried to add logging in Postoffice and the Clusterconnection for the notifications but doing so seemingly added enough of a delay to not trigger the issue (at least in the setup i used to reproduce, it's possible it would happen given different run values for the reproducer).

From what I can gather at least, locally everything happens in the correct order. Local counters have always been correct.

My **guess** would be that in some circumstance the servers `createQueue()` can take some time to finish, such that it allows a binding to get added, but before the BINDING_ADDED notification is sent, a call to the ServerConsumers `createConsumer()` is issued... this call requires no synchronization on Postoffice (as far as I can tell) and so its able to finish (and send its notification) before the servers `createQueue()` finishes all the way and sends its own notification.

So... my assumption is that something along those lines are causing this, which is why I added synchronization on Postoffice before allowing `createConsumer()` to send its notification (since it's `addBinding()` in Postoffice that sends the BINDING_ADDED notification). After making that change I have been unable to reproduce the issue again.

If it where to happen again though, the changes made in `RemoteQueueBindingImpl` should stop the redistributor from causing any issues regardless, but I'd much rather understand everything that's going on here for sure, if nothing else to be able to write a better reproducer for this...

AntonRoskvist:
Hello @jbertram 

I had some time to look further into this and came up with another fix and reproducer/test which seem to work better.

The main issue is that in some very race conditions the broker will send out it's notification for an added consumer before sending the binding_added notification for the queue the consumer is bound to.

From my testing this seems to happen when `postOfficeImpl#addBinding()` has _just_ added the actual binding to addressManager, but not yet called `managementService.sendNotification()` so that the unsynchronized `postOfficeImpl#getBinding()` will return it for ServerSessionImpl#createConsumer() and enabling it to lock the `managementService` before postOffice is able to.

I have not been able to point out _exactly_ what conditions has to be met for this to occur, but the new test included in this change works reliably to triggering the first issue in the chain leading up the "redistributor race" i.e getting the clusters remoteBindings out of sync with regards to their consumer count.

I'm removing the "Draft" status as of now but please let me know if anything looks off about these changes.

jbertram:
You added `addressLock` and use it inside `removeBinding`, `addBinding`, and `getBinding`. However, both `removeBinding` and `addBinding` are already `synchronized`. Therefore, you can just make `getBinding` `synchronized` as well. This would simplify your PR significantly.

AntonRoskvist:
@jbertram 

You are correct. I actually tried this at first but got issues with deadlocking because of calls to addBinding() from `PostOffice#onNotification`, then I did not try it again after having those calls go directly to the `addressManager`. Thanks!

During testing of this though, I saw the problem yet again... it's much less of an issue after the synchronization but can still happen apparently. I narrowed it down to varying delays in `PostOffice#processRoute()`... don't know if my fix for that is okay though but I figured if all references are nonDurables the call to `storageManager.afterCompleteOperations()` would not be necessary, is that right? Have not seen the issue since at least.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionBridge.java
```diff
@@ -278,7 +279,12 @@ private void setupNotificationConsumer() throws Exception {
                                                    createPermissiveManagementNotificationToFilter() +
                                                    ")");
 
-         sessionConsumer.createQueue(new QueueConfiguration(notifQueueName).setAddress(managementNotificationAddress).setFilterString(filter).setDurable(false).setTemporary(true));
+         sessionConsumer.createQueue(new QueueConfiguration(notifQueueName)
+                                        .setAddress(managementNotificationAddress)
+                                        .setFilterString(filter)
+                                        .setDurable(false)
+                                        .setTemporary(true)
+                                        .setRoutingType(RoutingType.MULTICAST));
```
clebertsuconic, line 287: was this an actual issue? or you are just making sure?

the address being a Multicast, was it creating ANYCAST queues in some cases?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/ClusterNotificationsTest.java
```diff
@@ -0,0 +1,124 @@
...
+            container.setConcurrentConsumers(20);
+            container.setConnectionFactory(new CachingConnectionFactory(factory));
+            container.setDestinationName(queue.getQueueName());
+            container.setBackOff(new FixedBackOff(1000, 2));
+            container.setReceiveTimeout(100);
+            container.setMessageListener((MessageListener) msg -> {
+               latch.countDown();
+            });
+
+            container.initialize();
```
clebertsuconic, line 109: This test is failing on my laptop. not being able to create native threads...

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/RemoteQueueBindingImpl.java
```diff
@@ -242,7 +243,10 @@ public synchronized void removeConsumer(final SimpleString filterString) throws
          }
       }
 
-      consumerCount--;
+      if (--consumerCount < 0) {
+         ActiveMQServerLogger.LOGGER.consumerCountError("Tried to decrement consumer count below 0: " + this);
```
clebertsuconic, line 247: Is this new error necessary?


I have been trying to use more log.debug for things not really pertaining to the user.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/ClusterNotificationsTest.java
clebertsuconic, line 109: This test is always failing for me:


```

java.lang.OutOfMemoryError: unable to create native thread: possibly out of memory or process/resource limits reached

	at java.base/java.lang.Thread.start0(Native Method)
	at java.base/java.lang.Thread.start(Thread.java:802)
	at org.springframework.core.task.SimpleAsyncTaskExecutor.doExecute(SimpleAsyncTaskExecutor.java:248)
	at org.springframework.core.task.SimpleAsyncTaskExecutor.execute(SimpleAsyncTaskExecutor.java:202)
	at org.springframework.core.task.SimpleAsyncTaskExecutor.execute(SimpleAsyncTaskExecutor.java:180)
	at org.springframework.jms.listener.DefaultMessageListenerContainer.doRescheduleTask(DefaultMessageListenerContainer.java:780)
	at org.springframework.jms.listener.AbstractJmsListeningContainer.resumePausedTasks(AbstractJmsListeningContainer.java:545)
	at org.springframework.jms.listener.AbstractJmsListeningContainer.doStart(AbstractJmsListeningContainer.java:299)
	at org.springframework.jms.listener.AbstractJmsListeningContainer.start(AbstractJmsListeningContainer.java:277)
	at org.springframework.jms.listener.DefaultMessageListenerContainer.start(DefaultMessageListenerContainer.java:657)
	at org.apache.activemq.artemis.tests.integration.cluster.distribution.ClusterNotificationsTest.testClusterNotificationsContinuity(ClusterNotificationsTest.java:110)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater$1.execute(IdeaTestRunner.java:38)
	at com.intellij.rt.execution.junit.TestsRepeater.repeat(TestsRepeater.java:11)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:35)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:232)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:55)

```


Can we do this test without spring though? it gets easier to debug tests. I'm not aware of any other tests actually using spring for consuming messages and asserting.


I reckon it removes a lot of boilerplate code from the consumer.. but I would prefer having more control on what's going on... 


like now: the tests Is failing with lack of threads and I have no idea how to fix it.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/ClusterConnectionBridge.java
AntonRoskvist: Semi related and making sure. I saw this issue presenting very consistently on a broker cluster in a TEST environment. Turn out they had configured default-routing-type = ANYCAST as an address setting and a match on "#" so that the notifications got load balanced between brokers, sort of mimicking the issue I'm going after with this PR.

I figured there ought to be no use case for setting up cluster notifications with ANYCAST semantics so I added this.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/ClusterNotificationsTest.java
AntonRoskvist: The problem here is (I guess) that in order for the test to trigger the issue this PR is about, underlying hardware most likely has to be working close to capacity. (This is not the case with an "actual" cluster though for some reason).
If you are interested in just trying this out I'd advice lowering queue count and consumer count on the Spring DMLC (while I see if I can find some formula to scale this based on available system resources or something).

Change would be:
:52   final int queueCount = 500; --> 100?

:100 container.setConcurrentConsumers(20); --> 5?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/RemoteQueueBindingImpl.java
AntonRoskvist: It is not strictly necessary no, I can remove it or add some debug logging

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/ClusterNotificationsTest.java
AntonRoskvist: Yes, I would also _really_ prefer to do this without Spring but so far I have been unable to reproduce this behavior without Spring (outside of production of course).

Closest thing I got to working involved "rolling restarts" of all nodes in a cluster with the connected clients running failover-semantics... this caused the issue a few times but not consistently _at all_ and the overall setup was to chaotic to make anything useful of it regardless.

clebertsuconic: on the case of the resources.. I think this test should be moved to tests/soak-tests... 

some tests I have are using parameters.sh... and we could then increase the sizing...

We should be able to run the tests on our envs.. and that kind of variable be applied to make it use a bigger box

AntonRoskvist: Yeah, that makes sense, I will move the test tomorrow.

Where you able to run the current test to trigger the issue by reducing concurrentConsumers and queueCount? Wondering if it's strictly timing plus resources that makes the difference or something more...

jbertram: FWIW, the test runs fine on my machine. I was able to back out the fix, reproduce the issue, change some stuff, retest, etc.

clebertsuconic: Ok.  Let's make it a soak test. And merge this PR. 

@AntonRoskvist let me know if you have any issues doing it.  If you can't let me know.

AntonRoskvist:
@jbertram 
@clebertsuconic

I have moved the test over to soak-tests and it is configurable through the parameters-script. Hope this is looking good and that the test is easier to set-up in this format.

clebertsuconic:
@AntonRoskvist You didn't need to use a real server in the soak tests.. you could still have used the embedded... 

(not asking you to change it.. if that's what you want to do.. just saying you didn't need to do it)

AntonRoskvist:
@clebertsuconic Oh... well thats certainly good to know. I just looked at a couple of other soak-tests and incidentally all of them used real servers so I just figured that's how those tests are supposed to be.

If it's all the same I'd like to keep the test as is, if nothing else to have as a starting point for any future tests that actually require real servers.

clebertsuconic:
@AntonRoskvist can you try my branch with a commit I added?


I can't make these tests to run.  Even on my biggest CI.


I tried adapting your test on my branch, to not require UDP.

And one of the issues I see is that the threads keep being created on Spring. So I used an Executor to limit some of that.

But still the container won't be stopped in case of any issues in the test (leaking threads onward).


to try my branch:


```
git remote add clebert https://github.com/clebertsuconic/activemq-artemis.git
git fetch clebert
git checkout clebert/ARTEMIS-4527 -B any-local-name-you-like
```


on this commit I'm also not using your template, and using broker-properties to set the address-settings.


Let me know if you can help me figure out this one please? you can reach me on slack if you prefer.

clebertsuconic:
I added commit efe450298dde7f9be0676495907f2746072a68d8

clebertsuconic:
commit efe450298dde7f9be0676495907f2746072a68d8 added on main branch... closing this PR as it's been "merged"

PR #4708, opened 2023-12-07
ARTEMIS-4522

jbertram:
This is a follow-up to #4699. I was going to just modify the original commit, but there were more changes necessary than I originally thought so created an independent commit.


artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/client/ActiveMQClient.java
```diff
@@ -309,7 +306,7 @@ public static int getGlobalScheduledThreadPoolSize() {
     */
    public static void initializeGlobalThreadPoolProperties() {
 
-      setGlobalThreadPoolProperties(Integer.valueOf(System.getProperty(ActiveMQClient.THREAD_POOL_MAX_SIZE_PROPERTY_KEY, "" + ActiveMQClient.DEFAULT_GLOBAL_THREAD_POOL_MAX_SIZE)), Integer.valueOf(System.getProperty(ActiveMQClient.SCHEDULED_THREAD_POOL_SIZE_PROPERTY_KEY, "" + ActiveMQClient.DEFAULT_SCHEDULED_THREAD_POOL_MAX_SIZE)));
+      setGlobalThreadPoolProperties(Integer.valueOf(System.getProperty(ActiveMQClient.THREAD_POOL_MAX_SIZE_PROPERTY_KEY, "" + ActiveMQClient.DEFAULT_GLOBAL_THREAD_POOL_MAX_SIZE)), Integer.valueOf(System.getProperty(ActiveMQClient.SCHEDULED_THREAD_POOL_SIZE_PROPERTY_KEY, "" + ActiveMQClient.DEFAULT_SCHEDULED_THREAD_POOL_MAX_SIZE)), Integer.valueOf(System.getProperty(ActiveMQClient.FLOW_CONTROL_THREAD_POOL_SIZE_PROPERTY_KEY, "" + ActiveMQClient.DEFAULT_FLOW_CONTROL_THREAD_POOL_MAX_SIZE)));
```
MrEasy, line 309: Integer.parseInt would avoid the boxing

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/ActiveMQClientLogger.java
```diff
@@ -347,4 +347,7 @@ public interface ActiveMQClientLogger {
 
    @LogMessage(id = 214034, value = "{} has negative counts {}\n{}", level = LogMessage.Level.ERROR)
    void negativeRefCount(String message, String count, String debugString);
+
+   @LogMessage(id = 214035, value = "Couldn't finish the client globalFlowControlThreadPool in less than 10 seconds, interrupting it now", level = LogMessage.Level.WARN)
```
MrEasy, line 351: thanks for the added warning.
Would recommend "Could not" instead of "Couldn't" like in other messages

MrEasy (approved):
Thanks for the improvements - added 2 minor comments

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/client/ActiveMQClient.java
jbertram: My change follows the established pattern of using `Integer.valueOf`. I would be fine with change this to use `Integer.parseInt`, but it would need to be in another commit that changes the pattern for every use here.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/ActiveMQClientLogger.java
jbertram: My change follows the established pattern of using "Couldn't finish the..." in the other messages related to stopping the other global thread pools. I would be fine with changing this to use `Could not`, but it would need to be in another commit that changes the pattern for every use here.

PR #4710, opened 2023-12-13
ARTEMIS-4532 MQTT-to-core wildcard conversion is broken

jbertram:
Currently when an MQTT topic filter contains characters from the configured wildcard syntax the conversion to/from this syntax breaks.

For example, when using the default wildcard syntax if an MQTT topic filter contains a . the conversion from the MQTT wildcard syntax to the core wildcard syntax and back will result in the `.` being replaced with a `/.`.

This commit fixes that plus a few other things...

 - Implements proper conversions to/from one WildcardConfiguration to another.
 - Refactors the MQTT code which invokes these conversion methods. This includes simplifying a lot of test code.
 - Adds lots of tests for everything.
 - Clarifies some variable naming to better distinguish between core and MQTT.

### DIFF
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTUtil.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTUtil.java
@@ -53,6 +53,7 @@ import org.apache.commons.text.CaseUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import java.lang.invoke.MethodHandles;
+import java.util.Objects;
 
 import static io.netty.handler.codec.mqtt.MqttProperties.MqttPropertyType.CONTENT_TYPE;
 import static io.netty.handler.codec.mqtt.MqttProperties.MqttPropertyType.CORRELATION_DATA;
@@ -123,7 +124,7 @@ public class MQTTUtil {
 
    public static final int TWO_BYTE_INT_MAX = Integer.decode("0xFFFF"); // 65_535
 
-    // https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901011
+   // https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901011
    public static final int VARIABLE_BYTE_INT_MAX = 268_435_455;
 
    public static final int MAX_PACKET_SIZE = VARIABLE_BYTE_INT_MAX;
@@ -138,26 +139,70 @@ public class MQTTUtil {
 
    public static final int DEFAULT_MAXIMUM_PACKET_SIZE = MAX_PACKET_SIZE;
 
-   public static String convertMqttTopicFilterToCore(String filter, WildcardConfiguration wildcardConfiguration) {
-      return convertMqttTopicFilterToCore(null, filter, wildcardConfiguration);
+   public static final WildcardConfiguration MQTT_WILDCARD = new WildcardConfiguration().setDelimiter(SLASH).setAnyWords(HASH).setSingleWord(PLUS);
+
+   /**
+    * This method takes the MQTT-related input and translates it into the proper name for a core subscription queue. The
+    * {@code topicFilter} may be either for a shared subscription in the format {@code $share/<shareName>/<topicFilter>}
+    * or a normal MQTT topic filter (e.g. {@code a/b/#}, {@code a/+/c}, {@code a/b/c}, etc.).
+    *
+    * @param topicFilter the MQTT topic filter
+    * @param clientId the MQTT client ID, used for normal (i.e. non-shared) subscriptions
+    * @param wildcardConfiguration the {@code WildcardConfiguration} governing the core broker
+    * @return the name of the core subscription queue based on the input
+    */
+   public static String getCoreQueueFromMqttTopic(String topicFilter, String clientId, WildcardConfiguration wildcardConfiguration) {
+      Objects.requireNonNull(topicFilter, "MQTT topic filter must not be null");
+      Objects.requireNonNull(wildcardConfiguration, "Broker wildcard configuration must not be null");
+
+      if (isSharedSubscription(topicFilter)) {
+         Pair<String, String> decomposed = decomposeSharedSubscriptionTopicFilter(topicFilter);
+         return new StringBuilder().append(decomposed.getA()).append(".").append(getCoreAddressFromMqttTopic(decomposed.getB(), wildcardConfiguration)).toString();
+      } else {
+         Objects.requireNonNull(clientId, "MQTT client ID must not be null");
+         return new StringBuilder().append(clientId).append(".").append(getCoreAddressFromMqttTopic(topicFilter, wildcardConfiguration)).toString();
+      }
    }
 
-   public static String convertMqttTopicFilterToCore(String prefixToAdd, String filter, WildcardConfiguration wildcardConfiguration) {
-      if (filter == null) {
-         return "";
-      }
+   /**
+    * This method takes the MQTT-related input and translates it into the proper name for a core address. The
+    * {@code topicFilter} must be normal (i.e. non-shared). It should not be in the format
+    * {@code $share/<shareName>/<topicFilter>}.
+    *
+    * @param topicFilter the MQTT topic filter
+    * @param wildcardConfiguration the {@code WildcardConfiguration} governing the core broker
+    * @return the name of the core addres based on the input
+    */
+   public static String getCoreAddressFromMqttTopic(String topicFilter, WildcardConfiguration wildcardConfiguration) {
+      Objects.requireNonNull(topicFilter, "MQTT topic filter must not be null");
+      Objects.requireNonNull(wildcardConfiguration, "Broker wildcard configuration must not be null");
 
-      String converted = MQTT_WILDCARD.convert(filter, wildcardConfiguration);
-      if (prefixToAdd != null) {
-         converted = prefixToAdd + converted;
-      }
-      return converted;
+      return MQTT_WILDCARD.convert(topicFilter, wildcardConfiguration);
    }
 
-   public static String convertCoreAddressToMqttTopicFilter(String address, WildcardConfiguration wildcardConfiguration) {
-      if (address == null) {
-         return "";
-      }
+   /**
+    * This is exactly the same as {@link #getCoreAddressFromMqttTopic(String, WildcardConfiguration)} except that it
+    * also prefixes the return with
+    * {@link org.apache.activemq.artemis.core.protocol.mqtt.MQTTUtil#MQTT_RETAIN_ADDRESS_PREFIX}
+    *
+    * @param topicFilter the MQTT topic filter
+    * @param wildcardConfiguration the {@code WildcardConfiguration} governing the core broker
+    * @return the name of the core address based on the input, stripping
+    *         {@link org.apache.activemq.artemis.core.protocol.mqtt.MQTTUtil#MQTT_RETAIN_ADDRESS_PREFIX} if it exists
+    */
+   public static String getCoreRetainAddressFromMqttTopic(String topicFilter, WildcardConfiguration wildcardConfiguration) {
+      return MQTT_RETAIN_ADDRESS_PREFIX + getCoreAddressFromMqttTopic(topicFilter, wildcardConfiguration);
+   }
+
+   /**
+    *
+    * @param address the core address
+    * @param wildcardConfiguration the {@code WildcardConfiguration} governing the core broker
+    * @return the name of the MQTT topic based on the input
+    */
+   public static String getMqttTopicFromCoreAddress(String address, WildcardConfiguration wildcardConfiguration) {
+      Objects.requireNonNull(address, "Address must not be null");
+      Objects.requireNonNull(wildcardConfiguration, "Broker wildcard configuration must not be null");
 
       if (address.startsWith(MQTT_RETAIN_ADDRESS_PREFIX)) {
          address = address.substring(MQTT_RETAIN_ADDRESS_PREFIX.length());
@@ -165,16 +210,6 @@ public class MQTTUtil {
 
       return wildcardConfiguration.convert(address, MQTT_WILDCARD);
    }
-
-   public static class MQTTWildcardConfiguration extends WildcardConfiguration {
-      public MQTTWildcardConfiguration() {
-         setDelimiter(SLASH);
-         setSingleWord(PLUS);
-         setAnyWords(HASH);
-      }
-   }
-
-   public static final WildcardConfiguration MQTT_WILDCARD = new MQTTWildcardConfiguration();
 
    private static ICoreMessage createServerMessage(MQTTSession session, SimpleString address, MqttPublishMessage mqttPublishMessage) {
       long id = session.getServer().getStorageManager().generateID();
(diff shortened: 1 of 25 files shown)


docs/user-manual/versions.adoc
```diff
@@ -31,6 +31,13 @@ These settings are now no longer required.
 +
 Configure `defaultMqttSessionExpiryInterval` instead.
 
+* Due to https://issues.apache.org/jira/browse/ARTEMIS-4532[ARTEMIS-4532] the names of addresses and queues related to MQTT topics and subscriptions respectively may change. This will only impact you:
++
+  . If the broker is configured to use a xref:wildcard-syntax.adoc[wildcard syntax] which doesn't match the xref:mqtt.adoc#wildcard-syntax[MQTT wildcard syntax].
+  . If you are using characters from the broker's wildcard syntax in your MQTT topic names or topic filters.
```
gemmellr, line 37: It could be clearer from the text both items in the list need apply, just to drive it home

docs/user-manual/versions.adoc
```diff
@@ -31,6 +31,13 @@ These settings are now no longer required.
...
++
+In this case the characters from the broker's wildcard syntax that do not match the characters in the MQTT wildcard syntax will be escaped with a backslash (i.e. `\`).
```
gemmellr, line 39: Suggestion for what a user might need to do to adapt? E.g rename an existing queues to what they 'should be' now?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/spec/controlpackets/PublishTests.java
```diff
@@ -311,15 +312,15 @@ public void testRetainFlagFalse() throws Exception {
...
+      Wait.assertTrue(() -> getRetainedMessageQueue(TOPIC).getMessageCount() == 1, 1000, 100);
 
       // send an unretained message; should *not* remove the existing retained message
       producer.publish(TOPIC, UNRETAINED_PAYLOAD.getBytes(), 2, false);
 
       producer.disconnect();
       producer.close();
 
-      Wait.assertFalse(() -> server.locateQueue(MQTTUtil.convertMqttTopicFilterToCore(MQTTUtil.MQTT_RETAIN_ADDRESS_PREFIX, TOPIC, MQTTUtil.MQTT_WILDCARD)).getMessageCount() > 1, 1000, 100);
+      Wait.assertFalse(() -> getRetainedMessageQueue(TOPIC).getMessageCount() > 1, 1000, 100);
```
gemmellr, line 323: Is there a reason why the assert should not just be checking for a count of 2 ?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/spec/controlpackets/PublishTests.java
```diff
@@ -269,15 +270,15 @@ public void testRetainFlagWithEmptyMessage() throws Exception {
       // send first retained message
       producer.publish(TOPIC, "retain1".getBytes(), 2, true);
 
-      Wait.assertTrue(() -> server.locateQueue(MQTTUtil.convertMqttTopicFilterToCore(MQTTUtil.MQTT_RETAIN_ADDRESS_PREFIX, TOPIC, MQTTUtil.MQTT_WILDCARD)).getMessageCount() == 1, 2000, 100);
+      Wait.assertTrue(() -> getRetainedMessageQueue(TOPIC).getMessageCount() == 1, 2000, 100);
```
gemmellr, line 273: Lots of tests asserting the queue count is 1 now, but would be better if they also verified it was 0 to begin with.

gemmellr (commented):
Also needs a rebase + retest now.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/WildcardConfiguration.java
```diff
@@ -119,19 +138,94 @@ public String getSingleWordString() {
...
+         .replace(ESCAPE + getSingleWordString(), getSingleWordString())
+         .replace(ESCAPE + getAnyWordsString(), getAnyWordsString());
+   }
+
+   private boolean isEscaped(final String input) {
+      for (int i = 0; i < input.length() - 1; i++) {
+         if (input.charAt(i) == ESCAPE && (input.charAt(i + 1) == getDelimiter() || input.charAt(i + 1) == getSingleWord() || input.charAt(i + 1) == getAnyWords())) {
+            return true;
+         }
+      }
```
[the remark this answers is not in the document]
jbertram: Shouldn't the `input.length() - 1` mitigate this?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/spec/controlpackets/PublishTests.java
```diff
@@ -653,6 +654,10 @@ public void messageArrived(String topic, MqttMessage message) throws Exception {
       consumer.close();
    }
 
+   private Queue getRetainedMessageQueue(String TOPIC) {
+      return server.locateQueue(MQTTUtil.convertMqttTopicFilterToCore(MQTTUtil.MQTT_RETAIN_ADDRESS_PREFIX, TOPIC, server.getConfiguration().getWildcardConfiguration()));
+   }
```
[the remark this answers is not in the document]
jbertram: Can you give me an example of what you mean?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/spec/controlpackets/PublishTests.java
jbertram: If we just check for 2 and the count is somehow 3 or more then the check will pass even though the behavior is wrong.

docs/user-manual/versions.adoc
jbertram: Updated.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/WildcardConfigurationTest.java
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.core.config;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class WildcardConfigurationTest extends Assert {
```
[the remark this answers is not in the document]
jbertram: I added a new test.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/WildcardConfigurationTest.java
```diff
@@ -0,0 +1,40 @@
...
+
+   public static final WildcardConfiguration MQTT_WILDCARD = new WildcardConfiguration().setDelimiter('/').setAnyWords('#').setSingleWord('+');
+   public static final WildcardConfiguration DEFAULT_WILDCARD = new WildcardConfiguration();
```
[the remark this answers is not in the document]
jbertram: Done.

docs/user-manual/versions.adoc
jbertram: Updated.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/spec/controlpackets/PublishTests.java
jbertram: Fixed.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/spec/controlpackets/PublishTests.java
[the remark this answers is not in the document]
jbertram: BTW, I fixed the parameter name.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/WildcardConfiguration.java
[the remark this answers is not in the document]
gemmellr: Doesn't seem like it. That governs that the variable i cant be beyond the 0-based string index used for _charAt(i)_, but the loop then also uses _charAt(i+1)_, so for the last character that would become input.length() and be beyond the limit?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/spec/controlpackets/PublishTests.java
gemmellr: Makes sense. I think maybe I misread that when I commented, not sure I saw/considered the wait. I'd generally try to avoid burning 1 second like that by doing something else instead thats likely to be more timely but gives the same 'processing should have occurred' result , e.g publishing another message somewhere else after the one currently being sent here, and waiting for it to arrive (after processing of the earlier message completes) before then checking on the retained message queue. Its only a second, yes, but it quickly adds up...there are ~20,000 tests after all.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/spec/controlpackets/PublishTests.java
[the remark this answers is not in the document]
gemmellr: This method seems to have been the source of multiple breaks, or changes in behaviour, over the last several releases. Obviously bugs happen, but it feels like some of these would have been caught by comprehensive unit testing of the method. The broker calls the method to name its queues, but the system tests also call the method to get the queue name the broker used. If for example the method returns "wrong-value" as it effectively has been doing, then broker names things in that way, and the system tests also just use that as the expected value, and so in those cases theyll always match and wont notice if anything is wrong, e.g I go in there later and mess something up with the naming.

If on the other hand there is a direct unit testing of the method comprehensively checking the output, ensuring that the exact value expected is returned, then the behaviour of that method is then more directly tested (and a few ms should tell if it is later broken...vs running minutes of system tests that mostly wont notice because they use the same method to establish what is expected), and likely more understandable because there are now specific tests for it showing what its expected to do, and quickly pointing out if changed inappropriately.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/WildcardConfigurationTest.java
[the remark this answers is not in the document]
gemmellr: The equals method itself should be far better tested in addition to the handling in convert(). Negative tests, tests of varying each applicable component field, checking that the equals and hashCode mutual contract is respected (I caught one in a PR just last week where that was broken).

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/WildcardConfigurationTest.java
```diff
@@ -0,0 +1,60 @@
...
+      assertEquals(coreAddress, MQTT_WILDCARD.convert(mqttTopicFilter, DEFAULT_WILDCARD));
+      assertEquals(mqttTopicFilter, DEFAULT_WILDCARD.convert(coreAddress, MQTT_WILDCARD));
+   }
+
+   @Test
+   public void testEquality() {
+      WildcardConfiguration a = new WildcardConfiguration().setDelimiter('/').setAnyWords('#').setSingleWord('+');
+      WildcardConfiguration b = new WildcardConfiguration().setDelimiter('/').setAnyWords('#').setSingleWord('+');
+
+      assertEquals(a, b);
```
gemmellr, line 54: Should also check the reverse

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/WildcardConfigurationTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+      String toConvert = RandomUtil.randomString();
+      assertTrue(toConvert == a.convert(toConvert, b));
+      assertTrue(toConvert == a.convert(toConvert, a));
```
gemmellr, line 58: Using assertSame would be nicer (tooling also tends to pick out this form of == use as a likely mistake)

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/WildcardConfiguration.java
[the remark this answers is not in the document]
jbertram: I think the `i < input.length() - 1` is preventing any issue here. If it was using `<=` or just `input.length()` there would be a problem, but as it stands now you can pass a `String` where the last character is the escape character (e.g. `foo\\`) and there's no IOOBE.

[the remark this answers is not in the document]
gemmellr: Yes I think you are right, think I misread that too and believed the loop itself was looking at the last char...so it wont throw because the loop never actually looks at the last character, just the if through the charAt(i+1).

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/WildcardConfigurationTest.java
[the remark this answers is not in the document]
jbertram: New tests added.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/WildcardConfigurationTest.java
jbertram: Done.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/WildcardConfigurationTest.java
jbertram: Done.

jbertram:
@gemmellr, I just pushed a big update that hopefully addresses all the issues. Once I got into it a bit more I decided to do a more thorough refactoring of this code. There were a number of messy & inconsistent bits that I straightened out.

gemmellr:
@jbertram Great. I was going to skim it now, but its a bigger change now and harder to see what changed, so ill review it all again on Monday. For now, the CI job looks to be calling for some javadoc fixups.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/MQTT5Test.java
```diff
@@ -83,6 +83,29 @@ public void messageArrived(String topic, MqttMessage message) {
...
+      final String topic = "foo1.0/bar/baz";
+
+      CountDownLatch latch = new CountDownLatch(1);
+      MqttClient subscriber = createPahoClient("subscriber");
+      subscriber.connect();
+      subscriber.setCallback(new DefaultMqttCallback() {
+         @Override
+         public void messageArrived(String t, MqttMessage message) {
+            logger.info("Message received from topic {}, message={}", topic, message);
+            assertEquals(topic, t);
```
gemmellr, line 97: Would be better to remember the value here and then assert outside the callback after the latch trips. That way the test will fail quicker and in a more helpful manner if the value isnt as expected, having the assertion directly cause it rather than e.g just timing out after 500ms on the latch (because it wasnt tripped) and requiring someone go look at logs and hope that there is a stack logged that might explain it. It would also just be clearer what is really being tested, took me a bit to even notice what the test really cared about while doing the send+receive.

gemmellr (commented):
The latest updates seem to have caused a couple of test failures (probably around the new argument validations):
```
java.lang.IllegalArgumentException
	at org.apache.activemq.artemis.tests.integration.mqtt5.MQTT5Test.testSharedSubscriptionsWithSameName(MQTT5Test.java:359)
```

```
java.lang.IllegalArgumentException
	at org.apache.activemq.artemis.tests.integration.mqtt5.MQTT5Test.testSharedSubscriptionsWithSameName2(MQTT5Test.java:414)```

File: artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTUtil.java [hunk 12]
```diff
@@ -138,25 +138,71 @@ public class MQTTUtil {
...
+    * or a normal MQTT topic filter (e.g. {@code a/b/#}, {@code a/+/c}, {@code a/b/c}, etc.).
+    *
+    * @param topicFilter the MQTT topic filter
+    * @param clientId the MQTT client ID, used for normal (i.e. non-shared) subscriptions
+    * @param wildcardConfiguration the {@code WildcardConfiguration} governing the core broker
+    * @return the name of the core subscription queue based on the input
+    */
+   public static String getCoreQueueFromMqttTopic(String topicFilter, String clientId, WildcardConfiguration wildcardConfiguration) {
+      if (topicFilter == null || clientId == null || wildcardConfiguration == null) {
+         throw new IllegalArgumentException();
```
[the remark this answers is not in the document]
jbertram: Funny. I had already started making that change!

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt5/MQTT5Test.java
jbertram: Good call.

jbertram:
Those test failures should be resolved now.

PR #4714, opened 2023-12-15
ARTEMIS-4535 prevent big stacktrack when invalid filter given by user

erwindon:
see https://issues.apache.org/jira/browse/ARTEMIS-4535

* prevent a large stack trace when the user-supplied message filter has errors
* improve feedback to user for filter-validity
* indent the search-field in the "Browse Queue" screen same as on other screens
* fixed a spelling error in a nearby comment


jbertram:
@erwindon, thanks!

PR #4715, opened 2023-12-15
ARTEMIS-4537 Fix merge of multiple address setting matches

brusdev:
Merge multiple address setting matches on a new instance to be idempotent.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
@@ -531,6 +531,9 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
    }
    private Integer idCacheSize = null;
 
+   static {
+      metaBean.add(Integer.class, "queuePrefetch", (t, p) -> t.queuePrefetch = p, t -> t.queuePrefetch);
+   }
    //from amq5
    //make it transient
    private transient Integer queuePrefetch = null;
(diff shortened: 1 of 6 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -1284,238 +1287,19 @@ public AddressSettings setIDCacheSize(int idCacheSize) {
...
+      metaBean.forEach((type, name, setter, getter, gate) -> {
+         Object sourceValue = getter.apply(AddressSettings.this);
+         if (sourceValue != null) {
+            setter.accept(target, sourceValue);
+         } else {
+            setter.accept(target, getter.apply(merged));
+         }
+      });
+
+      return target;
```
gemmellr, line 1302: Changing from merging in the current instance, to returning a new instance and leaving the current untouched, seems a pretty big change, a breaking one, that may not be spotted because the method signature is still compatible. Do we know it isnt going to silently break users?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/scheduling/DelayedMessageTest.java
```diff
@@ -57,7 +57,7 @@ protected void initServer() throws Exception {
 
       AddressSettings qs = server.getAddressSettingsRepository().getMatch("*");
       AddressSettings newSets = new AddressSettings().setRedeliveryDelay(DelayedMessageTest.DELAY);
-      newSets.merge(qs);
+      newSets = newSets.merge(qs);
```
gemmellr, line 60: ...e.g like it broke this unrelated test?

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
brusdev: I see your point, I reverted the previous merge signature and I added the mergeCopy override.

PR #4718, opened 2023-12-19
ARTEMIS-4531: Limit logfile output by default to 5

geertschuring:
Added a DefaultRolloverStrategy with a max of 5 files to the log appenders "log_file" and "audit_log_file" to prevent the (unlikely, I admit) situation where the system runs out of disk space because of the limitless growth of the amount of log files written to disk.

### DIFF
--- a/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/log4j2.properties
+++ b/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/log4j2.properties
@@ -87,6 +87,8 @@ appender.log_file.policies.type = Policies
 appender.log_file.policies.cron.type = CronTriggeringPolicy
 appender.log_file.policies.cron.schedule = 0 0 0 * * ?
 appender.log_file.policies.cron.evaluateOnStartup = true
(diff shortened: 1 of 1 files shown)


gemmellr:
This seems likely to surprise some folks using the existing default behaviour that has been there for...ages. I dont know how long, before my time. If it was decided to do a change like this by default, I'd wonder if it shouldn't be held for e.g the next major version.

There is no perfect choice here. 5 days actually wont be _enough_ logs for some folks, who currently get more out the box. It will also still be _too much_ for some others, since its a varying amount depending in large part on broker activity. Essentially the config file is created as a base, people really should be configuring it based on their own specific expected env and usage. Though many wont bother.

geertschuring:
@gemmellr, @jbertram After some more digging around it turns out the date in the filePattern (log/artemis.log.%d{yyyy-MM-dd}) keeps the DefaultRolloverStrategy from cleaning up old log files. To enable it to work the filePattern would need to be changed as well to something like (log/artemis.log.%i)
I already changed my mind after reading your previous comments, but now I'm sure it would be better to delete this PR and close the Issue without implementing it.

PR #4719, opened 2023-12-19
ARTEMIS-4488 Merge settings with wildcards after literal matches

brusdev:
This PR is a draft only because it depends on #4715


jbertram (approved):

jbertram:
This looks good to me, but it needs to be rebased.

PR #4722, opened 2023-12-20
ARTEMIS-4546 modernize Validators, etc.

jbertram:
No semantic changes here. Just some code gardening.


clebertsuconic:
looks way better without the clutter

PR #4723, opened 2023-12-21
ARTEMIS-4547 properly display empty message

erwindon:
see https://issues.apache.org/jira/browse/ARTEMIS-4547

tested in browser -> OK


erwindon:
before:
image

erwindon:
after:
image

PR #4724, opened 2023-12-21
ARTEMIS-4542 improve MQTT state storage

jbertram:
This commit:

 - Eliminates MQTT session storage on every successful connection. Instead data is only written when subsriptions are created or destroyed.
 - Adds a configuration property for the storage timeout.
 - Updates the documentation with relevant information.
 - Refactors a few bits of code to eliminate unnecessary variables, etc.


gemmellr:
Can you give this a rebase and re-test.

clebertsuconic:
LGTM.. as long as tests are good

PR #4725, opened 2023-12-22
ARTEMIS-4540 validate MQTT session state data

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
```diff
@@ -285,7 +285,7 @@ public Acceptor createAcceptor(TransportConfiguration info) {
             managementService.registerAcceptor(acceptor, info);
          }
       } catch (Exception e) {
-         ActiveMQServerLogger.LOGGER.errorCreatingAcceptor(info.getFactoryClassName(), e);
+         ActiveMQServerLogger.LOGGER.errorCreatingAcceptor(info.toString(), e);
```
gemmellr, line 288: This seems like it could/should be a separate change.

I'm also not sure logging all the info is that necessary, given the length it could be and that it often contains credentials (though there does look to be masking involved here). The underlying cause is being being logged anyway.

The acceptors are required to have names, what about just using that rather than the factory class name?

gemmellr (commented):
Can you rebase and retest this

artemis-server/src/main/java/org/apache/activemq/artemis/core/remoting/server/impl/RemotingServiceImpl.java
jbertram: Using the name is reasonable. The problem with just using the factory class name is that it's ambiguous. I'll update and push.

clebertsuconic:
LGTM ... as long as tests (whole test suite) are good...

PR #4726, opened 2023-12-22
ARTEMIS-4548 refactor MQTT tests

jbertram:
Many MQTT tests are run twice - once using TCP and once using WebSockets. This is essentially a big waste of time since once the connection is established to the broker the tests are identical. The tests should be refactored to run just once and then there can be a small number of tests specifically for WebSockets.

This should knock several minutes off the test-suite.


gemmellr:
Not clear if you ran the full test suite on this already, but probably worth a rebase and re-run at this point either way. Havent looked at it yet but will try to get to it soon.

jbertram:
Rebased. Test-suite is good!

tabish121:
Change seems to make sense as it reduces redundant test runs, something we made the mistake of doing in the 5.x code base was have all protocol tests run for every permutation of TCP, SSL, NIO, Auto, etc transports which explodes the test time.  It is important to create tests that capture any cases where the protocol or transport do come into play though.

PR #4729, opened 2024-01-04
ARTEMIS-4552 Config all connection-router settings by using broker properties

brusdev:


artemis-server/src/main/java/org/apache/activemq/artemis/core/config/routing/ConnectionRouterConfiguration.java
```diff
@@ -26,8 +26,8 @@ public class ConnectionRouterConfiguration implements Serializable {
    private KeyType keyType = KeyType.SOURCE_IP;
    private String keyFilter = null;
    private String localTargetFilter = null;
-   private CacheConfiguration cacheConfiguration = null;
-   private PoolConfiguration poolConfiguration = null;
+   private CacheConfiguration cacheConfiguration = new CacheConfiguration().setEnabled(false);
+   private PoolConfiguration poolConfiguration = new PoolConfiguration().setEnabled(false);
```
tabish121, line 30: I found this bit particularly confusing as it defaults the value in the constituent configuration objects to true but then forces them to false here.  My assumption is that this is somehow tied to trying to make this work both in the XML configuration and the broker properties bits but I'm not 100% its consistent between the two.  There doesn't appear to be an enable option in the XML but instead seems to imply that the presence alone is enabling but the properties would require the option to be set true even if any of the other configuration is present.  

In general I'd prefer if the code used a simpler default to false approach and the configuration options where consistent or could be changed such that the user gets a default enabled from XML via the file parser using a default and / or making it configurable in XML.  Then it would be nice also doing something consistent in the broker properties.  Just looking at the code changes to this point set of red flags for me and I had to chase down why it did this.

brusdev: @tabish121 I like your solution, it is more consistent. I changed the implementation.

gemmellr: I had the same reaction as Tim to the original code but hadnt got to thinking of an improvement.

In some ways this is definitely better, the code is a little more consistent with itself and other previous code. In others its much the same as before; the 'enabled' bit can now be controlled in both configs rather than just one, but there is still effectively 3 different/conflicting 'default' behaviours for it between XML and broker-properties and both, which is confusing. But I dont necessarily have an idea thats nicer to resolve that while there are actually 3 different ways of configuring things.

brusdev: I reverted the default value of the `enabled` bit for cache and pool configurations to true but I keep the xml attribute `enabled` suggested by @tabish121. The 3 different ways of configuring things seem more consistent:

- xml: the `connection-router` element has no cache or pool enabled by default but if you create a cache or pool element they are enabled by default (this is important for backward compatibility)
- programmatically: the `ConnectionRouterConfiguration` class has no cache or pool enabled by default but if you create a CacheConfiguration or PoolConfiguration class they are enabled by default (this is important for backward compatibility)
- broker properties: the `connectionRouters` items have no cache or pool enabled by default but you can set enabled to true to enable them

brusdev: I completely removed the `enabled` attribute.

gemmellr:
LGTM from my minimal understanding of the properties magic.

@gtully thoughts?

tabish121:
> LGTM from my minimal understanding of the properties magic.
> 
> @gtully thoughts?

I like the idea, although I remain mostly befuddled by the broker properties code.

gtully:
that looks good to me. it makes sense that the properties applier will auto create default instances for any intermediary that it encounters while trying to apply nested properties.
What we may need is a test for the case where w.x.y.z=1 is set and both getX and getY can return null.

gtully (approved):

PR #4731, opened 2024-01-05
ARTEMIS-4554 upgrade to latest Java for Docker images

jbertram:


artemis-docker/Dockerfile-centos7-19
```diff
@@ -17,7 +17,7 @@
 
 # ActiveMQ Artemis
 
-FROM eclipse-temurin:11-centos7
+FROM eclipse-temurin:19-jdk-centos7
```
gemmellr, line 20: This change probably doesnt make sense. Java 19 isnt really supported anymore, so similarly the Temurin Java 19 based images [are not either](https://github.com/docker-library/docs/blob/master/eclipse-temurin/README.md#supported-tags-and-respective-dockerfile-links). I'd either leave it at 11 until centos7 itself soon isnt (EOL in June) or use the Java 17 image variant, since both those JDKs/images will already be more up to date in terms of Java security fixes than the Java 19 one ever will be.

jbertram: Fixed.

artemis-docker/prepare-docker.sh
```diff
@@ -53,23 +53,23 @@ Well done! Now you can continue with building the Docker image:
...
+  $ docker build -f ./docker/Dockerfile-alpine-21 -t artemis-alpine .
 
-  # For Alpine with just JRE 11
-  $ docker build -f ./docker/Dockerfile-alpine-11-jre -t artemis-alpine-jre .
+  # For Alpine with just JRE 21
+  $ docker build -f ./docker/Dockerfile-alpine-21-jre -t artemis-alpine-jre .
 
   # For Ubuntu on Linux ARMv7/ARM64 with full JDK
-  $ docker buildx build --platform linux/arm64,linux/arm/v7 --push -t {your-repository}/apache-artemis:2.17.0-SNAPSHOT -f ./docker/Dockerfile-ubuntu-11 .
+  $ docker buildx build --platform linux/arm64,linux/arm/v8 --push -t {your-repository}/apache-artemis:2.16.0 -f ./docker/Dockerfile-ubuntu-21 .
```
gemmellr, line 72: Didnt notice this one before...not sure it makes sense since ARM v8 essentially is arm64, so its not clear thats command is something people should use (also be inconsistent with the comment above it still noting ARMv7). Could be worth just switching it to only linux/arm64 though at this point.

Also, should the "2.16.0" not just be replaced with a version placeholder given its usually going to be incorrect (assuming they arent actually using 2.16.0).

jbertram: Fixed.

artemis-docker/prepare-docker.sh
```diff
@@ -53,23 +53,23 @@ Well done! Now you can continue with building the Docker image:
...
 
-  # For Alpine with just JRE 11
-  $ docker build -f ./docker/Dockerfile-alpine-11-jre -t artemis-alpine-jre .
+  # For Alpine with just JRE 21
+  $ docker build -f ./docker/Dockerfile-alpine-21-jre -t artemis-alpine-jre .
 
-  # For Ubuntu on Linux ARMv7/ARM64 with full JDK
-  $ docker buildx build --platform linux/arm64,linux/arm/v7 --push -t {your-repository}/apache-artemis:2.17.0-SNAPSHOT -f ./docker/Dockerfile-ubuntu-11 .
+  # Multi-platform for Ubuntu on Linux AMD64 & ARM64/V8 with full JDK
+  $ docker buildx build --platform linux/amd64,linux/arm64/v8 --push -t {your-repository}/apache-artemis:{your-version} -f ./docker/Dockerfile-ubuntu-21 .
```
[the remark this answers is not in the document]
jbertram: The reason I chose `linux/arm64/v8` instead of `linux/arm64` is because that's what is reported for the `eclipse-temurin:21-jdk` [base image](https://hub.docker.com/layers/library/eclipse-temurin/21-jdk/images/sha256-dd7c71ee2e3564d88178b9cdd8dbe83c337d13bf19d0c3038c6f6b3b33186488?context=explore) being used in the example. I've also used `--platform linux/amd64,linux/arm64` in my work on [ARTEMIS-4556](https://issues.apache.org/jira/browse/ARTEMIS-4556) and it works fine although `linux/arm64` works fine as well. I'm honestly not sure which one should be preferred.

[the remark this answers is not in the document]
gemmellr: My understanding was they are essentially interchangeable, especially in this case with cross compiling against the bare 'linux/arm64' saying much the same thing as 'linux/arm/v8' because v8 was the original arm64. There could be more specific variants being produced in cases it might need to be explicitly distingish for a particular image, especially e.g if there are to be different supported variants of the image at the same time for different instruction classes and using a specific one is required in certain cases.

For comparison, we also only tend to use 'linux/amd64' for images but there are also multiple variants of AMD64/x86-64 as well now to distinguish different instruction additions over time through feature levels, e.g the [docs I referenced for docker buildx](https://docs.docker.com/build/building/multi-platform/#cross-compilation) demonstrates those too such as linux/amd64/v2 and linux/amd64/v3. In the past, most stuff was built for just AMD64 'v1' but that increasingly isnt true anymore.

As a related example to both of these, RHEL9 binaries are actually compiled for AMD64v2 feature level however the UBI9 images dont use linux/amd64/v2 for the os/arch name but [simply linux/amd64](https://hub.docker.com/layers/redhat/ubi9/latest/images/sha256-3b696988fc62a88631898d24bde1f2bfb703fe6530f36ad9b98baff9ad4c73fc?context=explore). The UBI9 image ARM bits are also published for simply linux/arm64 as another comparison.

(Aside, its still being explored [whether RHEL10 will be compiled for AMD64/v3](https://developers.redhat.com/articles/2024/01/02/exploring-x86-64-v3-red-hat-enterprise-linux-10)...I would guess the images will just continue to use simply linux/amd64 for the os/arch name even if they are for/require v3).

So e.g perhaps in future we could want/need to add linux/arm64/v9 in addition to linux/arm64 (or linux/arm64/v8) if we ever wanted to publish e.g v8 and v9 aimed images with different content at the same time to support some specific native instruction usages (dont see that happening for a while, personally), but otherwise we would likely just have one version and folks would all still just get the 'linux/arm64' image.

Actually, doing some more searching, this issue around the docker buildx stuff suggests that in some cases at least ['linux/arm64/v8' could just result in 'linux/arm64' output without specified variant since its not actually to expectation](https://github.com/moby/buildkit/issues/4039) and comments indicating instead it expects  just linux/arm64 or its said it could have been linux/arm/v8 if specifying v8 (i.e no 64 in that syntax). Though another comment there then also says 'linux/arm64' and 'linux/arm64/v8' are equal and neither is incorrect (in other contexts), which probably explains why the Temurin stuff is indeed using it just fine hehe.

The original example here has long used linux/arm64 for the V8 bits, as do all the examples in docs for the docker build stuff by default, as do the UBI images. I think linux/arm64 is the nicer/neater/simpler option until we should actually need to do something more complex. I think I'd go with it as it was unless there is compelling obviously-correct reason we should change it now, which I'm not really seeing.

PR #4732, opened 2024-01-08
ARTEMIS-4557 expose producer window size in clusterconnection JMX

andytaylor:
https://issues.apache.org/jira/browse/ARTEMIS-4557

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ClusterConnectionControlImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ClusterConnectionControlImpl.java
@@ -321,5 +321,18 @@ public class ClusterConnectionControlImpl extends AbstractControl implements Clu
 
    }
 
+   @Override
+   public long getProducerWindowSize() {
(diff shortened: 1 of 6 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ClusterConnectionControlImpl.java
```diff
@@ -321,5 +321,18 @@ public Map<String, Object> getBridgeMetrics(String nodeId) {
 
    }
 
+   @Override
+   public long getProducerWindowSize() {
+      if (AuditLogger.isBaseLoggingEnabled()) {
+         AuditLogger.getTopology(this.clusterConnection);
```
gemmellr, line 327: This seems wrong, surely getProducerWindowSize() shouldn't be audit logging about getting topology?

jbertram: Nice catch. I skimmed right over that. I fixed it with 42ddf99ca8846218ee7b49d97a4f5087318c15e3.

PR #4735, opened 2024-01-09
ARTEMIS-4559 refactor HA docs & code/module naming 

jbertram:

 - Updates HA docs including the chapter on network isolation (i.e. split brain). The network isolation chapter is now more about high-level explanation and the HA doc now has all the configuration parameters.
 - Changes references to "pluggable quorum voting" to "pluggable lock manager." The pluggable functionality really isn't about voting. Conceptually is much more like the functionality you'd get from a distributed lock so this naming is more clear. Both the docs and the code have been changed.
 - Reorganize lock manager modules as sub-modules. The API and RI modules are renamed, but that should be OK based on the "experimental" tag that's been on this feature up to this point.
 - Remove the "experimental" tag from the lock manager.

These changes will not break folks using the standalone broker. However, they will break folks embedding the broker *if* they are using the artemis-quorum-ri or artemis-quorum-api modules or the o.a.a.a.c.c.h.DistributedPrimitiveManagerConfiguration class.

There are no functional changes here. Renaming these modules is more a conceptual change to facilitate better documentation and increased adoption.


jbertram:
In my opinion, inspecting the documentation from the diff is difficult. I'd recommend pulling the PR and [building the docs](https://activemq.apache.org/components/artemis/documentation/hacking-guide/#only-docs). Then you can really see the new content and flow.

clebertsuconic:
This is a big beyond refactoring just docs, right? you're also renaming the package and project name?


not complaining about the change.. just noticing it's a bit more than just doc changes?

jbertram:
@clebertsuconic, that's a fair point. I updated the Jira and commit message to reflect this.

jbertram:
@ehsavoie, is merging this going to cause any problems for WildFly? Are you guys using `org.apache.activemq.artemis.core.config.ha.DistributedPrimitiveManagerConfiguration`. I'm guessing you don't since you don't ship either of the current `artemis-quorum-api` or `artemis-quorum-ri` jars.

ehsavoie:
@jbertram I don't think so but I'll do a quick run tomorrow

ehsavoie:
@jbertram LGTM

jbertram:
@clebertsuconic, I think this is ready to go.

PR #4736, opened 2024-01-10
ARTEMIS-4561 expose store type on web component tls binding config

gtully:


artemis-web/src/test/java/org/apache/activemq/cli/test/WebServerComponentTest.java
```diff
@@ -338,6 +345,21 @@ public void simpleSecureServer() throws Exception {
       Assert.assertFalse(webServerComponent.isStarted());
    }
 
+   @Test
+   public void testStoreTypeConfigAndProviderRegistration() throws Exception {
+
+      BindingDTO bindingDTO = new BindingDTO();
+      bindingDTO.setKeyStorePath(PEM_KEY_STORE_PATH);
+      bindingDTO.setKeyStoreType("PEMCFG");
```
jbertram, line 353: What is this `PEMCFG` keystore type? I don't see it documented anywhere? Is this something a user would use?

gtully: It is a convience when cert and private key are in two separate pem files. The path is to a property file that references both.

gtully: References https://github.com/ctron/pem-keystore#reading-keycert-from-two-files

PR #4738, opened 2024-01-10
ARTEMIS-4556 support release of multi-platform Docker images

jbertram:


.gitignore
```diff
@@ -33,3 +33,5 @@ artemis-native/src/main/c/org_apache_activemq_artemis_jlibaio_LibaioContext.h
 
 # overlay outpit
 **/overlays/**/*
+
+artemis-docker/_TMP_
```
gemmellr, line 37: Could you use the target directory instead of _TMP_? That way it would be more consistent with other modules and not need a separate exclusion here? (...and perhaps elsewhere.. e.g doing a source assembly creation would presumably still pick this stuff up, but wouldn't need special handling to skip use of target).

jbertram: Using `target` instead of `_TMP_` makes sense. That said, it's worth noting that `artemis-docker` isn't actually a Maven module so I'm not sure the source assembly example would be a concern.

gemmellr: Yep I know it isnt a module (could make it one for target dir cleanup at some point..), but I think it would be picked up by the source assembly stuff as thats mainly just done with a fileset on the build base dir.

jbertram: That makes sense.

PR #4741, opened 2024-01-15
ARTEMIS-4567 support OpenWire clients for broker embedded in Jakarta environment

jbertram:
If the broker is embedded into a Jakarta environment then the existing artemis-openwire-protocol module won't work because it uses javax classes.

This commit adds a new Jakarta-specific module that can be used to support OpenWire clients in Jakarta environments (e.g. Spring Boot 3). Users will simply need to include this version on their classpath to enable support.

### DIFF
--- a/artemis-protocols/artemis-jakarta-openwire-protocol/pom.xml
+++ b/artemis-protocols/artemis-jakarta-openwire-protocol/pom.xml
@@ -1,0 +1,152 @@
+<!--
+  Licensed to the Apache Software Foundation (ASF) under one or more
+  contributor license agreements.  See the NOTICE file distributed with
+  this work for additional information regarding copyright ownership.
+  The ASF licenses this file to You under the Apache License, Version 2.0
+  (the "License"); you may not use this file except in compliance with
+  the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+   <modelVersion>4.0.0</modelVersion>
+
+   <parent>
+      <artifactId>artemis-protocols</artifactId>
+      <groupId>org.apache.activemq</groupId>
+      <version>2.32.0-SNAPSHOT</version>
+   </parent>
+
+   <artifactId>artemis-jakarta-openwire-protocol</artifactId>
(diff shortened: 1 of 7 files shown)


gemmellr (commented):
Might adding a compatibility test make sense..or expanding the one that already exists for the other jakarta bits? As-is this would seem to be entirely untested.

artemis-protocols/artemis-jakarta-openwire-protocol/pom.xml
```diff
@@ -0,0 +1,162 @@
...
+         <exclusions>
+            <exclusion>
+               <groupId>org.apache.geronimo.specs</groupId>
+               <artifactId>geronimo-jms_1.1_spec</artifactId>
+            </exclusion>
+            <exclusion>
+               <groupId>org.apache.geronimo.specs</groupId>
+               <artifactId>geronimo-j2ee-management_1.1_spec</artifactId>
+            </exclusion>
+         </exclusions>
```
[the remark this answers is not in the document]
jbertram: Fixed.

pom.xml
```diff
@@ -779,6 +779,22 @@
...
+            <exclusions>
+               <exclusion>
+                  <groupId>org.apache.geronimo.specs</groupId>
+                  <artifactId>geronimo-jms_1.1_spec</artifactId>
+               </exclusion>
+               <exclusion>
+                  <groupId>org.apache.geronimo.specs</groupId>
+                  <artifactId>geronimo-j2ee-management_1.1_spec</artifactId>
+               </exclusion>
+            </exclusions>
```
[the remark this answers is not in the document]
jbertram: ditto

jbertram:
@gemmellr, I added a simple compatibility test to ensure OpenWire support is enabled when Artemis is in a Jakarta environment.

tests/compatibility-tests/pom.xml
```diff
@@ -337,6 +337,8 @@
                               <arg>org.apache.activemq:artemis-server:${project.version}</arg>
                               <arg>org.apache.activemq:artemis-journal:${project.version}</arg>
                               <arg>org.apache.activemq:artemis-hornetq-protocol:${project.version}</arg>
+                              <arg>org.apache.activemq:artemis-jakarta-openwire-protocol:${project.version}</arg>
+                              <arg>org.apache.logging.log4j:log4j-slf4j2-impl:2.21.1</arg>
```
gemmellr, line 341: This should use the version property to pick up the appropriate version the codebase is using.

tests/compatibility-tests/pom.xml
```diff
@@ -686,6 +688,7 @@
                            <optional>true</optional>
                            <libListWithDeps>
                               <arg>org.apache.activemq:activemq-client:5.11.0</arg>
+                              <arg>org.apache.logging.log4j:log4j-slf4j2-impl:2.21.1</arg>
```
gemmellr, line 691: Ditto (for the log4j bit)

gemmellr (commented):
LGTM other than using the appropriate version prop in a couple places.

tests/compatibility-tests/pom.xml
jbertram: Yes, indeed.

tests/compatibility-tests/pom.xml
jbertram: Ditto

PR #4743, opened 2024-01-16
ARTEMIS-4569 Blocked Producers will hold runnables until messages are…

clebertsuconic:
… consumed.

When initially developed the expectation was that no more producers would keep connecting but in a scenario like this the consumers could actually give up and things will just accumulate on the server.

We should cleanup these upon disconnect.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/runnables/RunnableList.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/runnables/RunnableList.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.activemq.artemis.utils.runnables;
+
+import java.util.HashSet;
+import java.util.function.Consumer;
+
+public class RunnableList {
+
+   private final HashSet<AtomicRunnable> list = new HashSet<>();
+
+   public RunnableList() {
+   }
+
+   public synchronized void add(AtomicRunnable runnable) {
+      runnable.setAcceptedList(this);
+      list.add(runnable);
+   }
+
+   public int size() {
+      return list.size();
+   }
+
+   public synchronized void remove(AtomicRunnable runnable) {
+      list.remove(runnable);
+   }
+
+   public synchronized void cancel() {
+      list.forEach(this::cancel);
+      list.clear();
+   }
+
+   private void cancel(AtomicRunnable atomicRunnable) {
+      atomicRunnable.cancel();
+   }
+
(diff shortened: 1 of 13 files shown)


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/runnables/RunnableList.java
```diff
@@ -0,0 +1,60 @@
...
+
+   public int size() {
+      return list.size();
+   }
+
+   public synchronized void remove(AtomicRunnable runnable) {
+      list.remove(runnable);
+   }
+
+   public void afterRun(AtomicRunnable runnable) {
```
tabish121, line 43: What is the purpose of this method as opposed to just calling the remove method directly as this method does, would seem to be more descriptive for the callers to invoke remove plus remove is directly declared as synchronized which seems the point of the atomic types being implemented.

clebertsuconic: @tabish121 I just thought it would be more descriptive to a dedicated operation to be called after the execution.. Future extension points.. etc.


I will change it to call remove directly though. I will mark this as a resolve when I push the commit.


Thanks a lot

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/runnables/RunnableList.java
```diff
@@ -0,0 +1,56 @@
...
+   public synchronized void cancel() {
+      list.forEach(this::cancel);
+      list.clear();
+   }
+
+   private void cancel(AtomicRunnable atomicRunnable) {
+      atomicRunnable.cancel();
+   }
+
+   public void forEach(Consumer<AtomicRunnable> runnable) {
```
clebertsuconic, line 52: my own review to myself here: wrong name for this variable.. runnable.. consumerRunnable perhaps.

tests/leak-tests/src/test/java/org/apache/activemq/artemis/tests/leak/ProducerBlockedLeakTest.java
```diff
@@ -0,0 +1,162 @@
...
+         server.addAddressInfo(new AddressInfo(QUEUE_NAME).addRoutingType(RoutingType.ANYCAST));
+         server.createQueue(new QueueConfiguration(QUEUE_NAME).setAddress(QUEUE_NAME).setRoutingType(RoutingType.ANYCAST).setDurable(true));
+
+         // clients need to be disconnected while blocked. For that reason a new VM is being spawned
+         Process process = SpawnedVMSupport.spawnVM(ProducerBlockedLeakTest.class.getName(), protocol, "10");
+
+         Wait.assertTrue(() -> loggerHandler.findText("AMQ222183"), 5000, 10); //unblock
+
+         process.destroyForcibly();
+         Assert.assertTrue(process.waitFor(10, TimeUnit.SECONDS));
```
clebertsuconic, line 116: Wrong comment here. It's actually blocked...

tests/leak-tests/src/test/java/org/apache/activemq/artemis/tests/leak/ProducerBlockedLeakTest.java
```diff
@@ -0,0 +1,162 @@
...
+      try (AssertionLoggerHandler loggerHandler = new AssertionLoggerHandler()) {
+         AtomicInteger messagesSent = new AtomicInteger(0);
+
+         server.addAddressInfo(new AddressInfo(QUEUE_NAME).addRoutingType(RoutingType.ANYCAST));
+         server.createQueue(new QueueConfiguration(QUEUE_NAME).setAddress(QUEUE_NAME).setRoutingType(RoutingType.ANYCAST).setDurable(true));
+
+         // clients need to be disconnected while blocked. For that reason a new VM is being spawned
+         Process process = SpawnedVMSupport.spawnVM(ProducerBlockedLeakTest.class.getName(), protocol, "10");
+
+         Wait.assertTrue(() -> loggerHandler.findText("AMQ222183"), 5000, 10); //unblock
```
clebertsuconic, line 113: wrong comment here.. it's blocked

PR #4744, opened 2024-01-17
ARTEMIS-4570 filter not applied to all brokers in cluster

gaohoward:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -4322,7 +4322,9 @@ public QueueConfiguration getQueueConfiguration() {
          .setTemporary(temporary)
          .setInternal(internalQueue)
          .setTransient(refCountForConsumers instanceof TransientQueueManagerImpl)
-         .setAutoCreated(autoCreated);
+         .setAutoCreated(autoCreated)
```
gtully, line 4325: this change does not seem related, what is going on here?

gtully (commented):
a cluster is expected to be symmetrical, but if for some reason it is not, then there will be an error logged on an unrecognised notification type, which is fine.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gaohoward: This is related to the test, where I get the QueueConfiguration from the queue and use it as the base, change the filter and call PostOffice to updateQueue with it. If I don't add them the test throws exception, like so:
`java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because the return value of "org.apache.activemq.artemis.api.core.QueueConfiguration.isGroupRebalancePauseDispatch()" is null

	at org.apache.activemq.artemis.core.postoffice.impl.PostOfficeImpl.updateQueue(PostOfficeImpl.java:776)
	at org.apache.activemq.artemis.tests.integration.cluster.distribution.MessageLoadBalancingTest.testMessageLoadBalancingWithFiltersUpdate(MessageLoadBalancingTest.java:128)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
	at java.base/java.lang.reflect.Method.invoke(Method.java:578)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:54)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at org.junit.rules.TestWatcher$1.evaluate(TestWatcher.java:61)
	at com.intellij.rt.junit.IdeaTestRunner$Repeater.startRunnerWithArgs(IdeaTestRunner.java:33)
	at com.intellij.rt.junit.JUnitStarter.prepareStreamsAndStart(JUnitStarter.java:235)
	at com.intellij.rt.junit.JUnitStarter.main(JUnitStarter.java:54)
`

gtully: hmm. I don't know whether the update needs to be smarter or every QueueConfiguration must be fully defined. I think the update needs to be smarter b/c there is a semantic meaning to null afaik.

I see another partial QueueConfiguration at https://github.com/apache/activemq-artemis/blob/8c41fe67db3462790aed38f62fb4e4b1e55c3d49/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java#L695

there are a few in PostOfficeImpl. This suggests that the update logic needs to check for the null. I think.

jbertram: Both things are true. The update needs to be smarter to not throw an NPE, but `getQueueConfiguration` should also be setting all the relevant properties on the `QueueConfiguration` it returns. In short, I think this is fine and we can follow-up with another PR to deal with the update.

gaohoward: Yeah I agree, the update needs to be smarter and also the returned queue config should be complete. I'll do update PR quickly. Thanks guys!

gaohoward:
> a cluster is expected to be symmetrical, but if for some reason it is not, then there will be an error logged on an unrecognised notification type, which is fine.

yeah that's true.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gaohoward: On second thought it would be better to do it in a separate commit/PR as @jbertram said. :)

clebertsuconic: You could do it similarly to how I did AddessSettings and that Metadata thingy, on a separate PR.

gtully:
@clebertsuconic had an interesting observation that changing a filter was previously not allowed, a queue needed to be deleted and recreated with a new filter and those events would be propagated across a cluster. I wonder if there is any potential for having stuck messages with dynamic filters, when the filter becomes more restrictive that would result on unmatched messages that are already on the queue? I guess we could recommend a purge before mod if that is a problem. The bigger question is if a filter should be something that is dynamic?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gaohoward: @clebertsuconic Hi Clebert can you point me to where you did it? Thanks.

gaohoward:
I think for topic subscriptions (jms) if you changed the filter the effect is the sub-queue will be recreated (unsubscribed).
if the filter is on the queue side, that's something different.

clebertsuconic:
Queues are not supposed to be changed... that's something recently added. I don't think your user (I know your user) is actually hitting an update queue.. Did you confirm that?

gtully:
it can happen from config reload, it seems a valid use case.

gaohoward:
yep we should either 1) forbid filter update on both local binding and remote binding, or 1) allow both.
Currently we allow local update but not remote, that make the policy inconsistent.

PR #4746, opened 2024-01-17
ARTEMIS-4572 upgrade to Apache parent 31

jbertram:


pom.xml
```diff
@@ -1750,10 +1749,16 @@
                <artifactId>jacoco-maven-plugin</artifactId>
                <version>${version.org.jacoco.plugin}</version>
             </plugin>
+            <plugin>
+               <!-- This is needed until https://issues.apache.org/jira/browse/MSHADE-462 is resolved. Once that is
+                    done then we can inherit the version from the Apache parent POM. -->
+               <groupId>org.apache.maven.plugins</groupId>
+               <artifactId>maven-shade-plugin</artifactId>
+               <version>3.4.1</version>
+            </plugin>
```
gemmellr, line 1758: You can just set (i.e override) the property they used in the parent dependencyManagement to the desired version, rather than add the full config and hard-coded version.

etc/checkstyle.xml
```diff
@@ -100,14 +100,15 @@ under the License.
...
-      <module name="DiamondOperatorForVariableDefinition"/>
       <module name="RequiredParameterForAnnotation">
          <property name="annotationName" value="Parameterized.Parameters"/>
          <property name="requiredParameters" value="name"/>
       </module>
 
       <module name="UnnecessarySemicolonInTryWithResources"/>
 
+      <!-- Sevntu checks, http://sevntu-checkstyle.github.io/sevntu.checkstyle/ -->
+      <module name="DiamondOperatorForVariableDefinition"/>
```
gemmellr, line 111: The comment seems potentially out of place now since e.g RequiredParameterForAnnotation also looks to be a Sevntu check? Did this need moved?

pom.xml
jbertram: Good idea!

etc/checkstyle.xml
jbertram: I looked in the Sevntu Git repo for `RequiredParameterForAnnotation` and I couldn't find it, but I see now that I was looking in the wrong place (i.e. `coding` package instead of `annotation`). My goal here was to clarify which specific checks were from Sevntu.

PR #4749, opened 2024-01-18
ARTEMIS-4576 ServerSessionImpl#updateProducerMetrics access large messages after being routed

gaohoward:
ServerSessionImpl#updateProducerMetrics method is being called after a message is routed. If the message is a large message, it can be acked quickly and it's backing file will be closed before the method is being called.
This bug causes random failure in test org.apache.activemq.artemis.tests.integration.paging.MessagesExpiredPagingTest#testSendReceiveCORELarge

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
@@ -2357,11 +2357,15 @@ public class ServerSessionImpl implements ServerSession, FailureListener {
          routingContext.setAddress(art.getName());
          routingContext.setRoutingType(art.getRoutingType());
 
+         // Retrieve message size for metrics update before routing,
+         // since large message backing files may be closed once routing completes
+         int mSize = msg instanceof LargeServerMessageImpl ? ((LargeServerMessageImpl)msg).getBodyBufferSize() : msg.getEncodeSize();
+
          result = postOffice.route(msg, routingContext, direct);
 
          logger.debug("Routing result for {} = {}", msg, result);
 
-         updateProducerMetrics(msg, senderName);
+         updateProducerMetrics(msg, senderName, mSize);
       } finally {
          if (!routingContext.isReusable()) {
             routingContext.clear();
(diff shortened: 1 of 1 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -2357,11 +2357,20 @@ public synchronized RoutingStatus doSend(final Transaction tx,
...
 
+         int lmSize = 0;
+         if (msg instanceof LargeServerMessageImpl) {
+            // before routing large messages,
+            // prefetch its body size for updating
+            // producer metrics after message being routed
+            // because after route the large messages backing file
+            // may be closed quickly with acks from clients
+            lmSize = ((LargeServerMessageImpl)msg).getBodyBufferSize();
+         }
```
gemmellr, line 2368: Rather than only passing the size for large messages, seems like it would be nicer if it assessed the size for either type here and then always pass it to the update. That way the existing check that was previously done later can just be moved here (i.e do same work, just in different place) rather than needing to repeat the check.

The comment can then perhaps also just be summarised a bit, e.g "Retrieve message size for metrics update before routing, since large message backing files may be closed once routing completes"

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -2357,11 +2357,19 @@ public synchronized RoutingStatus doSend(final Transaction tx,
          routingContext.setAddress(art.getName());
          routingContext.setRoutingType(art.getRoutingType());
 
+         // Retrieve message size for metrics update before routing,
+         // since large message backing files may be closed once routing completes
+         int lmSize = 0;
```
gemmellr, line 2362: variable name needs updated (here and in updateProducerMetrics), lmSize doesnt make sense now

gaohoward: ha, yes. Thanks Robbie. I think I can also use ?: form to combine the statements into one line.

gaohoward: Done!

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ServerSessionImpl.java
```diff
@@ -2520,10 +2524,10 @@ public String toManagementString() {
       return "ServerSession [id=" + getConnectionID() + ":" + getName() + "]";
    }
 
-   private void updateProducerMetrics(Message msg, String senderName) {
+   private void updateProducerMetrics(Message msg, String senderName, int lmSize) {
```
gemmellr, line 2527: lmSize still needs updated here

gemmellr:
Once the variable is renamed I think this looks good.

Though I think someone more familiar with the routing and large message stuff should also give it a once over. @clebertsuconic @jbertram

clebertsuconic:
@howardgao I don't actually understand what's being fixed.. it seems the same semantic.. replacing 6 by half dozen?


Perhaps you meant to check this for AMQP, in which case you would have to also check for AMQPLargeServerMessage... or maybe add another method to an upper level?



if you actually create a test to validate what's wrong, it would be check the right fix.

clebertsuconic:
Ohhh.. I see now...

clebertsuconic:
I pushed this directly... hence I'm closing this PR as I already "merged it"

PR #4752, opened 2024-01-19
ARTEMIS-4579 Add the *FirstMessage* API for scheduled messages

jsmucr:
Alerting on issues with messages not being received properly for a period of time is an uneasy task. We use the `getFirstMessageAge()` command to trigger alerts in Zabbix, and it works as long as there are no consumers.
But this approach fails when there are consumers repeatedly failing to receive a message. That message is getting scheduled for redelivery over and over, and even though there still is an old message in the queue to be reported, it's no longer visible via `getFirstMessage*()` API.

The goal here is to add a set of functions working with messages scheduled for delivery:
```java
getFirstScheduledMessageAsJSON()
getFirstScheduledMessageTimestamp()
getFirstScheduledMessageAge()
```
It may be not the most effective approach but it's quite a convenient one, especially when monitoring a wide set of queues, each with its own set of alerts.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerImpl.java
@@ -51,9 +51,13 @@ public class ScheduledDeliveryHandlerImpl implements ScheduledDeliveryHandler {
 
    // This contains RefSchedules which are delegates to the real references
    // just adding some information to keep it in order accordingly to the initial operations
+   // Do not forget to call notifyScheduledReferencesUpdated() when updating the set.
    private final TreeSet<RefScheduled> scheduledReferences = new TreeSet<>(new MessageReferenceComparator());
 
    private final QueueMessageMetrics metrics;
+
+   // Oldest by timestamp, not by scheduled delivery time
+   private MessageReference oldestMessage = null;
 
    public ScheduledDeliveryHandlerImpl(final ScheduledExecutorService scheduledExecutor,
          final Queue queue) {
@@ -82,6 +86,7 @@ public class ScheduledDeliveryHandlerImpl implements ScheduledDeliveryHandler {
    public void addInPlace(final long deliveryTime, final MessageReference ref, final boolean tail) {
       synchronized (scheduledReferences) {
          scheduledReferences.add(new RefScheduled(ref, tail));
+         notifyScheduledReferencesUpdated();
       }
       metrics.incrementMetrics(ref);
    }
@@ -129,6 +134,7 @@ public class ScheduledDeliveryHandlerImpl implements ScheduledDeliveryHandler {
             MessageReference ref = iter.next().getRef();
             if (predicate.test(ref)) {
                iter.remove();
+               notifyScheduledReferencesUpdated();
                refs.add(ref);
                metrics.decrementMetrics(ref);
             }
@@ -151,6 +157,7 @@ public class ScheduledDeliveryHandlerImpl implements ScheduledDeliveryHandler {
             if (ref.getMessage().getMessageID() == id) {
                ref.acknowledge(tx);
                iter.remove();
+               notifyScheduledReferencesUpdated();
                metrics.decrementMetrics(ref);
                return ref;
             }
@@ -185,6 +192,34 @@ public class ScheduledDeliveryHandlerImpl implements ScheduledDeliveryHandler {
          if (logger.isTraceEnabled()) {
             logger.trace("Couldn't make another scheduler as {} is already set, now is {}", deliveryTime, now);
          }
+      }
+   }
+
+   protected void notifyScheduledReferencesUpdated() {
+      oldestMessage = null;
+   }
+
+   @Override
+   public MessageReference peekFirstScheduledMessage() {
+      synchronized (scheduledReferences) {
+         if (scheduledReferences.isEmpty()) {
+            return null;
+         }
+         if (oldestMessage != null) {
+            return oldestMessage;
+         }
+         MessageReference result = null;
+         long oldestTimestamp = Long.MAX_VALUE;
+         for (RefScheduled scheduledReference : scheduledReferences) {
+            MessageReference ref = scheduledReference.getRef();
+            long refTimestamp = ref.getMessage().getTimestamp();
+            if (refTimestamp < oldestTimestamp) {
+               oldestTimestamp = refTimestamp;
+               result = ref;
+            }
+         }
+         oldestMessage = result;
+         return result;
       }
    }
 
@@ -232,6 +267,7 @@ public class ScheduledDeliveryHandlerImpl implements ScheduledDeliveryHandler {
                }
 
                iter.remove();
+               notifyScheduledReferencesUpdated();
                metrics.decrementMetrics(reference);
 
                reference.setScheduledDeliveryTime(0);
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/logs/AuditLogger.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/logs/AuditLogger.java
@@ -2682,4 +2682,32 @@ public interface AuditLogger {
    @LogMessage(id = 601772, value = "User {} is getting producerWindowSize on target resource: {}", level = LogMessage.Level.INFO)
    void getProducerWindowSize(String user, Object source);
 
+   static void peekFirstScheduledMessage(Object source) {
+      BASE_LOGGER.peekFirstScheduledMessage(getCaller(), source);
+   }
+
+   @LogMessage(id = 601773, value = "User {} is getting first scheduled message on target resource: {}", level = LogMessage.Level.INFO)
+   void peekFirstScheduledMessage(String user, Object source);
+
+   static void peekFirstScheduledMessageAsJSON(Object source) {
+      BASE_LOGGER.peekFirstScheduledMessageAsJSON(getCaller(), source);
+   }
+
+   @LogMessage(id = 601774, value = "User {} is getting first scheduled message as json on target resource: {}", level = LogMessage.Level.INFO)
+   void peekFirstScheduledMessageAsJSON(String user, Object source);
+
+   static void peekFirstMessage(Object source) {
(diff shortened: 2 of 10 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerImpl.java
```diff
@@ -188,6 +194,34 @@ private void scheduleDelivery(final long deliveryTime) {
...
+            long refTimestamp = ref.getMessage().getTimestamp();
+            if (refTimestamp < oldestTimestamp) {
+               oldestTimestamp = refTimestamp;
+               result = ref;
+            }
+         }
+         oldestMessage = result;
+         return result;
+      }
+   }
```
clebertsuconic, line 223:
```suggestion
   @Override
   public MessageReference peekFirstScheduledMessage() {
      synchronized (scheduledReferences) {
         if (scheduledReferences.isEmpty()) {
            return null;
         }

         RefScheduled refScheduled = scheduledReferences.first();
         return refScheduled != null ? refScheduled.getRef() : null;
      }
   }    
```

Why not just use scheduledReferences.first()?


I have been dealing with memory leaks in the past, especially the ones that I tested with the leak-tests I wrote, and this would actually eventually leak one message. Besides I think it's more efficient and less error prone to just call getfirst().


in My suggestion I'm still checking for isEmpty(), which I'm not sure it's needed.. something to verify.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerImpl.java
```diff
@@ -51,10 +51,13 @@ public class ScheduledDeliveryHandlerImpl implements ScheduledDeliveryHandler {
 
    // This contains RefSchedules which are delegates to the real references
    // just adding some information to keep it in order accordingly to the initial operations
+   // Do not forget to call notifyScheduledReferencesUpdated() when updating the set.
```
clebertsuconic, line 54: if you apply my suggestion please remove these calls..

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerImpl.java
```diff
@@ -51,10 +51,13 @@ public class ScheduledDeliveryHandlerImpl implements ScheduledDeliveryHandler {
...
    private final TreeSet<RefScheduled> scheduledReferences = new TreeSet<>(new MessageReferenceComparator());
 
    private final QueueMessageMetrics metrics;
 
+   private MessageReference oldestMessage = null;
```
clebertsuconic, line 59: remove this upon accepting my suggestion.. (and please squash the change as yours)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerImpl.java
clebertsuconic: when / if you apply my suggestion, please squash the commit as yours.

clebertsuconic:
my only concern is storing a reference outside the data structure for the first message. treeMap.first() is there for right that purpose.


on the implementation for first(), this is just returning:


   public E first() {
        return m.firstKey();
    }
    
    
    
Which means you can always just return first().



I'm not sure if you need to check for empty or not.. as you would need to double check on that.

jsmucr:
Thanks for the review @clebertsuconic, I'll process your suggestions as the first thing tomorrow morning.

jbertram:
For awhile I've actually discouraged folks from using the "first message" metrics. [I discussed this on the ActiveMQ users list](https://lists.apache.org/thread/d79n3kbb28k2v4pm7y0kywb5xpvrmpf4) not long ago:

> The `getFirstMessageAge` operation is actually fairly "heavy" and not generally recommended. Furthermore, the age of the first message isn't meaningful in and of itself in this scenario because if the `consumerCount` is 0 then by definition no messages can be stuck. A robust stuck-message detection mechanism must, at the very least, verify that `consumerCount` > 0. Also, instead of using the age of the first message I recommend inspecting `messagesAcknowledged` over time. For example, if the `consumerCount` > 0 and `messagesAcknowledged` remains unchanged for 60 seconds then messages (or more likely *consumers*) may be stuck. If you're using Prometheus then I believe you can use a [range vector selector](https://prometheus.io/docs/prometheus/latest/querying/basics/#range-vector-selectors) for this kind of operation.

At this point I'm against adding "first message" metrics for scheduled messages because it will be a relatively "heavy" operation due to the `synchronized` block. A lot of JMX monitoring tools will simply poll queue MBeans which means this new management method may be invoked *a lot*, especially on a broker with lots of queues. Over the last few years we've seen an increasing number of deployments with many thousands of queues. This is one reason we implemented (and generally recommend using) [pluggable metrics](https://activemq.apache.org/components/artemis/documentation/latest/metrics.html#metrics) which should provide a lighter footprint than JMX and open the door for easier integration with tools that specialize in graphing and alerting (e.g. Prometheus & Grafana).

Would it be possible for you to use existing metrics to solve your problem rather than implementing this new management method?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ScheduledDeliveryHandlerImpl.java
jsmucr: The `MessageReferenceComparator` uses scheduled delivery time as a measure, so if I'm not mistaken, `scheduledReferences.first()` returns the first message to be delivered, not the oldest message in the queue.

jsmucr:
@jbertram Only by **very complicated** workarounds. I need to find out if there's an old message in a queue that's been there for a while now, and I'm not allowed to use DLQs for that, because it means a manual intervention on the broker if processing fails too many times.
I know this method is heavy but so is our workflow here, which is pretty much based on wasting resources in exchange for usability and easy maintenance. :-)
I understand your point of view, I really do, but I also find Artemis to be quite an advanced tool all by itself, and that if I slow it down by using a method that is heavy on resources, it's entirely my fault as an operator. Perhaps a page of documentation would be enough to get rid of responsibility here.

jbertram:
> I need to find out if there's an old message in a queue that's been there for longer than a specified interval, and I'm not allowed to use DLQs for that, because it means a manual intervention on the broker if processing fails too many times.

I'd really like to focus on exactly why you need to know if there's an old message in a queue that's been there for longer than a specified interval?

As noted in my previous comment, in order to have a robust detection of stalled consumers (a.k.a. stuck messages) you _already_ need to look at multiple metrics. If you're relying solely on a metric like `firstMessageAge` you're liable to get false positives. It's just not a good solution. I've long considered deprecating these metrics because folks tend to misunderstand and misuse them.

Keep in mind that we have [slow consumer detection](https://activemq.apache.org/components/artemis/documentation/latest/slow-consumers.html#detecting-slow-consumers) built into the broker.

> I know this method may be heavy but so is our workflow here, which is pretty much based on wasting resources in exchange for usability and easy maintenance.

The problem, as I see it, is that by implementing this method you're going to force this wasting of resources on other users who likely don't want it. As noted previously, it's very common for JMX monitoring tools to fetch the values of every attribute on a given MBean. By adding this method you're implicitly impacting that use-case. Lots of folks use JMX monitoring tools that scan MBeans more than once per minute.

clebertsuconic:
I'm also concerned about adding weight here. 


I would be okay on adding a method to peek at the first message.. but adding extra processing weight for those who don't need this.. I'm kind of concerned also.


Can't you get this by listScheduledMessages on the control?


or... also I would be okay on adding a metric for number of scheduled messages... (I think it would make sense).. as that would just get the counter directly. and then you can do your processing on listing all the messages if certain metrics apply.


Also, you could tell a lot by the number of redeliveries on the system. you should probably add a warning on your system for the redeliveries counter.

jbertram:
> or... also I would be okay on adding a metric for number of scheduled messages...

We already have that. It's called `scheduledCount`.

clebertsuconic:
Right! I looked at the callers on the TreeMap.. I forgot we have the metrics in place.

jsmucr:
And as I was composing my previous post, I realized that maybe it's not the idea but the approach that we couldn't agree upon!

> Can't you get this by listScheduledMessages on the control?

That's pretty close actually. In order to be able to monitor the entire cluster via Zabbix I've created sort of a JMX proxy which collects and exposes data from all nodes at once. That proxy is also capable of calling methods, and exposing their results as custom JMX attributes.

---

So basically, the issue you have with the `*FirstMessage*` API is that it's a set of attributes, not a set of operations, right?

clebertsuconic:
@jbertram if this was an operation instead of an attribute, then you wouldn't have the issue of scans from JMX providers, right?

jsmucr:
I've replaced the attributes with this new `peekFirstScheduledMessage()` operation. Maybe I could add `peekFirstMessage()` too, or perhaps rename both to avoid confusion: `peekOldestMessage()`, `peekOldestScheduledMessage()`

jbertram:
> Currently, this basically means to pause the queue, because the GUI doesn't display messages-in-flight. 

There **is** a management method which shows details of in-flight messages. It's called `listDeliveringMessagesAsJSON`. That operation, combined with `listScheduledMessagesAsJSON`, should give you everything you need, right?

jbertram:
> ...if this was an operation instead of an attribute, then you wouldn't have the issue of scans from JMX providers, right?

@clebertsuconic, that would be _much_ better, but I still feel like there's a better solution here.

jsmucr:
> There **is** a management method which shows details of in-flight messages. It's called `listDeliveringMessagesAsJSON`. That operation, combined with `listScheduledMessagesAsJSON`, should give you everything you need, right?

It should but there are two things I don't like about this approach:
* The output requires some postprocessing, and therefore it is fairly unusable from the GUI.
* There's nothing that stops the broker from transforming and exporting significant amount of data when all I need is one number.

> I still feel like there's a better solution here
Again, I'm all ears. The bigger picture is that fast response time is a must, or I'm stopping a car assembly line (literally), so every help is much appreciated. :-)

jbertram:
After thinking about it more I haven't come up with any good alternatives. I'm OK with merging this now.

PR #4753, opened 2024-01-19
ARTEMIS-4575 Start/stop consumers on status change

joshb1050:
Currently, with many consumers tracked on a single ServerSession, each time a new consumer is added, we iterate through all the existing consumers to set their status to "started". This means that if there are k existing consumers on a session, adding a new consumer is O(k).

However, we can optimize this by using an AtomicBoolean to ensure that we only set `started=true` on existing consumers when it was previously false, or vice versa.


tabish121:
I've just sent my own PR for this issue.  I am skeptical of this change as it omits starting the actual added consumers after the first one which omits and calls to prompt delivery on the consumers as they are added. It also changes broker code that might impact other implementations that rely on the current behavior of those APIs which gives me pause.  I think the changes should be isolated to the openwire module if possible.

joshb1050:
Closing in favor of approach in #4754

PR #4754, opened 2024-01-19
ARTEMIS-4575 Only start the consumers that were added

tabish121:
Change from forcing a session start cycle on each consumer add event and start only those consumers that were added which will trigger a prompt delivery action on each. The session should be marked started on create to account for the remove of the start on each consumer add event.


tabish121:
I've run the full test suite on this and all the tests passed.  I don't see any simple unit test for this as the desired outcome is that there should be no change other than a reduction in resource utilization on consumer adds to an existing session.

joshb1050:
Agreed this makes more sense than #4754, especially isolating impact to OpenWire.

joshb1050:
@tabish121 Thanks for the quick PR and merge on this—is there an estimate for when a new release will be cut with this change?

clebertsuconic:
I'm actually planning to cut a release this week. I will send a heads up to the dev-list...

PR #4755, opened 2024-01-19
ARTEMIS-4580 security settings for FQQN not enforced on send

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/FQQNSendSecurityTest.java
```diff
@@ -0,0 +1,77 @@
...
+      SecurityConfiguration securityConfiguration = new SecurityConfiguration();
+      securityConfiguration.addUser(USER, PASS);
+      securityConfiguration.addRole(USER, ROLE);
+      securityManager.setConfiguration(securityConfiguration);
+      server.setSecurityManager(securityManager);
+
+      server.start();
+   }
+
+   @Test
```
tabish121, line 67: Shouldn't there also be a negative test to check that when not able to send to the FQQN an exception is thrown from the sender?

jbertram: Fixed.

tabish121:
LGTM, looks to need a rebase before merging

PR #4756, opened 2024-01-22
NO-JIRA Fix PagingStoreImplTest#testBlockUnblock random failure

gaohoward:
PagingStroeImpl.checkReleasedMemory() will kick off executor.execute(this::memoryReleased) to pull from the queue onMemoryFreedRunnables asynchronously. If the executor fires the task too late, it can pick up one of the late trackMemoryChecks runnable and increase its calls, making assertion fail. need to flush the executors to make sure it doesn't happen.

### DIFF
--- a/tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/paging/impl/PagingStoreImplTest.java
+++ b/tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/paging/impl/PagingStoreImplTest.java
@@ -1252,6 +1252,7 @@ public class PagingStoreImplTest extends ActiveMQTestBase {
          };
          store.applySetting(new AddressSettings().setMaxSizeBytes(1000).setAddressFullMessagePolicy(AddressFullMessagePolicy.BLOCK));
          store.addSize(100);
+         store.flushExecutors();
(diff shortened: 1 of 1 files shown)


clebertsuconic:
LGTM

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/paging/impl/PagingStoreImplTest.java
```diff
@@ -1252,6 +1252,7 @@ public void run() {
          };
          store.applySetting(new AddressSettings().setMaxSizeBytes(1000).setAddressFullMessagePolicy(AddressFullMessagePolicy.BLOCK));
          store.addSize(100);
+         store.flushExecutors();
          store.checkMemory(trackMemoryChecks, null);
          assertEquals(1, calls.get());
```
clebertsuconic, line 1257: another possible fix would probably be Wait.assertEquals

gaohoward: The wait may not work well for example here waiting for zero "0" would not prevent the task being fired later as it's init state is 0.

PR #4758, opened 2024-01-25
ARTEMIS-4583: remove the artemis-server test-jar, do related cleanup

gemmellr:


gemmellr:
I ran the full test suite on this a couple times (and several times before that prior to latest rebase).

tabish121:
Looks good, nice to start removing the test-jar chains.

brusdev:
LGTM, nice work!

PR #4760, opened 2024-01-30
ARTEMIS-4586 Auto reload web binding SSL stores on change

brusdev:


jbertram:
LGTM

artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
```diff
@@ -258,6 +277,10 @@ private ServerConnector createServerConnector(HttpConfiguration httpConfiguratio
                }
             }
          }
+         if (binding.getSslAutoReload() != null && binding.getSslAutoReload() == true) {
```
gemmellr, line 280: Perhaps replace the 2 checks with simply Boolean.TRUE.equals(binding.getSslAutoReload()) ?

docs/user-manual/web-server.adoc
```diff
@@ -43,6 +43,8 @@ The minimum number of threads the embedded web server will hold to service HTTP
 Default is `8` or the value of `maxThreads` if it is lower.
 idleThreadTimeout::
 The time to wait before terminating an idle thread from the embedded web server. Measured in milliseconds. Default is `60000`.
+scanPeriod::
+How often to scan for file changes. Measured in seconds. Default is `5`.
```
gemmellr, line 47: Think this could be clearer. The TLS reload it only happens if the other setting is enabled, which it isnt by default, but this reads like scanning is always happening...though it doesnt say what is being scanned. I think its likely to be interpreted as being other files to do with the web stuff, so I think it needs spelled out here if TLS reload is all this setting currently governs, and also mention to the other config option given the 2 are documented so far apart yet it is really the other setting that is more important (and the other setting does reference this one).

jbertram:
@gemmellr, this OK to merge?

gemmellr:
didnt get notified it had been updated, so havent looked yet

gemmellr:
Yep update looks good, merged.

PR #4773, opened 2024-01-30
ARTEMIS-4601 remove OpenTelemetry deps from main pom

jbertram:


brusdev:
io.opentelemetry:opentelemetry-semconv 1.34.1-alpha doesn't exist, the latest version is 1.30.1-alpha
https://mvnrepository.com/artifact/io.opentelemetry/opentelemetry-semconv

it was moved to io.opentelemetry.semconv:opentelemetry-semconv
https://github.com/micrometer-metrics/tracing/issues/343

the latest version of io.opentelemetry.semconv:opentelemetry-semconv is  1.23.1-alpha
https://mvnrepository.com/artifact/io.opentelemetry.semconv/opentelemetry-semconv

jbertram:
Looking at this a bit more closely...I think we can remove the OpenTelemetry dependencies from the main `pom.xml` and isolate to the `opentelemetry` example itself.

clebertsuconic:
It makes even more sense now that the examples are separated from the main repo.

PR #4782, opened 2024-01-31
ARTEMIS-1230 Added artemis-bom

amarkevich:
Rework https://github.com/apache/activemq-artemis/pull/2615
- excluded test dependencies;
- introduce parent submodule


amarkevich:
> @amarkevich, this needs to be rebased since there are conflicts now.

done

artemis-parent/pom.xml
```diff
@@ -0,0 +1,2074 @@
...
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+   <modelVersion>4.0.0</modelVersion>
+   <parent>
+      <groupId>org.apache.activemq</groupId>
+      <artifactId>artemis-pom</artifactId>
+      <version>2.33.0-SNAPSHOT</version>
+   </parent>
+
+   <artifactId>artemis-parent</artifactId>
```
gemmellr, line 26: Something else to consider might be whether we instead introduced a new name at the root pom (e.g artemis-project, artemis-parent...could still use e.g build-parent here later) and leave this as artemis-pom for now, perhaps avoiding the change of effective content breaking anyone...e.g some might be [ab]using it as an import due to lack of specific bom.

(or also like it breaks the now-standalone Examples build, which currently still uses it as a parent as it always had...in part as that was the minimal change during their recent move, but also that there was no bom yet and I didnt want to import the parent during the move or duplicate all the management).

artemis-parent/pom.xml
```diff
@@ -0,0 +1,2074 @@
...
+                     <artifactId>activemq-artemis-native</artifactId>
+                     <version>${activemq-artemis-native-version}</version>
+                     <type>jar</type>
+                     <overWrite>false</overWrite>
+                     <outputDirectory>${project.build.directory}/bin</outputDirectory>
+                     <includes>**/*.so</includes>
+                  </artifactItem>
+               </artifactItems>
+            </configuration>
+         </plugin>
```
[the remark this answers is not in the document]
amarkevich: also changed property to `-Djava.library.path="${activemq.basedir}/target/bin"`

artemis-parent/pom.xml
```diff
@@ -0,0 +1,2074 @@
...
+
+   <dependencyManagement>
+      <dependencies>
+         <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>artemis-bom</artifactId>
+            <version>${project.version}</version>
+            <type>pom</type>
+            <scope>import</scope>
+         </dependency>
```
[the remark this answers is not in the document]
amarkevich: in general rest of transitive dependencies are here but not a top level one

artemis-parent/pom.xml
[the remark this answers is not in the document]
gemmellr: I'm not sure changing that is correct, the files will go in subdirs which is what the property contents pointed to I believe. It was just that you moved the extraction config and so changed the effect of where they were being extracted to that was the issue.

artemis-parent/pom.xml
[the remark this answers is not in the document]
gemmellr: right, I'm saying we should decide what we want to do with them...though I guess we can do that later.

artemis-pom/pom.xml
```diff
@@ -0,0 +1,2063 @@
...
+                        <goals>
+                           <!-- combine results for all modules into a single report -->
+                           <goal>aggregate</goal>
+                        </goals>
+                     </execution>
+                  </executions>
+               </plugin>
+            </plugins>
+         </build>
+      </profile>
```
gemmellr, line 1174: Seems like an aggregrate goal usage should be be in the root?

artemis-pom/pom.xml
```diff
@@ -0,0 +1,2063 @@
...
+                        <phase>compile</phase>
+                        <goals>
+                           <goal>check</goal>
+                        </goals>
+                     </execution>
+                  </executions>
+               </plugin>
+            </plugins>
+         </build>
+      </profile>
```
gemmellr, line 1259: Feels like the enforcement should be at the root at least to fail before doing anything.

artemis-pom/pom.xml
```diff
@@ -0,0 +1,2063 @@
...
+                        <id>source-release-assembly</id>
+                        <configuration>
+                           <skipAssembly>true</skipAssembly>
+                        </configuration>
+                     </execution>
+                  </executions>
+               </plugin>
+            </plugins>
+         </build>
+      </profile>
```
gemmellr, line 1279: This needs to be at the root to continue having the desired effect, since it was to suppress activity at the root.

artemis-pom/pom.xml
```diff
@@ -0,0 +1,2063 @@
...
+            </plugin>
+            <plugin>
+               <groupId>org.apache.maven.plugins</groupId>
+               <artifactId>maven-release-plugin</artifactId>
+               <configuration>
+                  <autoVersionSubmodules>true</autoVersionSubmodules>
+                  <tagNameFormat>@{project.version}</tagNameFormat>
+                  <arguments>-DskipTests</arguments>
+               </configuration>
+            </plugin>
```
gemmellr, line 1778: Feels like this one should be at the root also

artemis-pom/pom.xml
```diff
@@ -0,0 +1,2063 @@
...
+                     <rules>
+                        <requireMavenVersion>
+                           <version>3.5.0</version>
+                           <message>You must use Maven 3.5.0+ to build</message>
+                        </requireMavenVersion>
+                     </rules>
+                  </configuration>
+               </execution>
+            </executions>
+         </plugin>
```
gemmellr, line 1816: Feels like these should be at the root as they are overrides

artemis-pom/pom.xml
```diff
@@ -0,0 +1,2063 @@
...
+            <artifactId>dependency-check-maven</artifactId>
+            <version>${owasp.version}</version>
+            <reportSets>
+               <reportSet>
+                  <reports>
+                     <report>aggregate</report>
+                  </reports>
+               </reportSet>
+            </reportSets>
+         </plugin>
```
gemmellr, line 2046: another aggregate, does it need to be at root?

artemis-bom/pom.xml
```diff
@@ -0,0 +1,243 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gemmellr, line 1: This new file is seemingly using Windows line endings according to the git diff, can you convert it to Unix line endings to match the rest.

artemis-pom/pom.xml
```diff
@@ -0,0 +1,870 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
gemmellr, line 1: Similarly, this new file has Windows line endings, switch it to Unix for consistency with others.

gemmellr (approved):
LGTM. Verified with a full test suite run and verified the examples against it.

I'll give others some more time to review first before I merge.

jbertram:
LGTM.

tabish121:
Changes look okay to me

PR #4797, opened 2024-01-31
ARTEMIS-4626 Improve demand tracking for federation consumers

tabish121:
Implement idempotent tracking for tracking demand at start while also tracking demand as consumer or bindings are added and removed.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/federation/internal/FederationAddressEntry.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/federation/internal/FederationAddressEntry.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.activemq.artemis.protocol.amqp.federation.internal;
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.apache.activemq.artemis.core.postoffice.Binding;
+
+/**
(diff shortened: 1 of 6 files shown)


tabish121:
I've run through the tests several times now in CI and no longer see any random failures from the tests that were exhibiting the over counting behaviors

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/federation/internal/FederationQueueEntry.java
```diff
@@ -28,19 +33,19 @@
  * on a federation resource such that it is not torn down until all demand has been
  * removed from the local resource.
  */
-public class FederationConsumerEntry {
+public class FederationQueueEntry {
 
    private final FederationConsumerInternal consumer;
 
-   private int references = 1;
+   private final Set<String> consumerDemand = new HashSet<>();
```
clebertsuconic, line 40: I was wondering why you didn't just use a HashSet of ServerConsumers here.

But I guess this is better since it decreases chances of memory leaking? if you forgot an entry here at least you wouldn't leak the consumer->connection and everything associated with it?

clebertsuconic:
this looks good to me. 


I have one question about the HashSet used but it's purely a question for my understanding.. not asking any changes.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/federation/internal/FederationQueueEntry.java
tabish121: I wanted to avoid any chance of leaking the ServerConsumer instance which have lead to holding onto other resources in the past.  If the notification isn't sent and we don't remove the demand at least we aren't holding onto the actual consumer.

clebertsuconic: @tabish121 that's what I thought why you did it.. thanks

clebertsuconic:
@tabish121 this looks good to me.. I'm merging it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/federation/internal/FederationAddressEntry.java
```diff
@@ -0,0 +1,92 @@
...
+      demandBindings.add(binding);
+      return this;
+   }
+
+   /**
+    * Reduce demand on this federation address consumer from the given binding.
+    *
+    * @return this federation address consumer entry.
+    */
+   public FederationAddressEntry removeDenamd(Binding binding) {
```
gemmellr, line 88: Typo, Denamd - > Demand

EDIT:  Also, the method on FederationQueueEntry is called _reduce_Demand rather than _remove_Demand, possibly they should align.

PR #4790, opened 2024-01-31
ARTEMIS-4617 upgrade JLine to 3.24.1

jbertram:

PR #4800, opened 2024-02-01
ARTEMIS-4617 upgrade JLine to 3.25.1

clebertsuconic:
I had to remove the indirect dependency between the maven plugin and jline to avoid the maven plugin to parse some classes in jline that require experimental features on the JDK even when they are not in use.

### DIFF
--- a/artemis-boot/src/main/java/org/apache/activemq/artemis/boot/Artemis.java
+++ b/artemis-boot/src/main/java/org/apache/activemq/artemis/boot/Artemis.java
@@ -61,7 +61,7 @@ public class Artemis {
 
 
 
-      Object result = execute(fileHome, fileInstance, fileBrokerETC, true, args);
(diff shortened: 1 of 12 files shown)


jbertram:
According to [this](https://github.com/jline/jline3?tab=readme-ov-file#jansi) Jansi should be integrated into 3.25.x so we should be able to get ride of the Jansi dependency now.

clebertsuconic:
@jbertram ohh.. jansi.. I think I reverted one change by accident... I will try to remove it now

artemis-distribution/pom.xml
```diff
@@ -157,6 +157,16 @@
          <version>${project.version}</version>
       </dependency>
 
+      <!-- shell integration -->
```
clebertsuconic, line 160: I don't need this any more.

clebertsuconic:
this was previously discussed here: https://github.com/apache/activemq-artemis/pull/4790

this PR superseded #4790

PR #4801, opened 2024-02-02
ARTEMIS-4630 Validate if the pid is really from Artemis

angelogalvao:
In rare cases the status is returning 'running' to a process that is not Artemis. It was create a extra level of verification by validating if the running process working directory is the
same of ${ARTEMIS_INSTANCE}.

### DIFF
--- a/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service
+++ b/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service
@@ -59,9 +59,13 @@ fi
 status() {
   if [ -f "${PID_FILE}" ] ; then
     pid=`cat "${PID_FILE}"`
+    pid_workdir=$ARTEMIS_INSTANCE
+    if [ `command -v lsof &> /dev/null` ]; then
+      pid_workdir="`lsof -a -d cwd -p ${pid} | awk '{print $9}' | tail -1`"
+    if
     # check to see if it's gone...
     ps -p ${pid} > /dev/null
-    if [ $? -eq 0 ] ; then
+    if [ $? -eq 0 ] && [ "$pid_workdir" = "$ARTEMIS_INSTANCE" ]; then
       return 0
     else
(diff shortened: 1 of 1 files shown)


jbertram:
Do you have any idea what leads to this situation with the pid in the first place?

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service
```diff
@@ -61,7 +61,7 @@ status() {
     pid=`cat "${PID_FILE}"`
     # check to see if it's gone...
     ps -p ${pid} > /dev/null
-    if [ $? -eq 0 ] ; then
+    if [ $? -eq 0 ] && [ `pwdx ${pid} | awk '{print $2}'` == "$ARTEMIS_INSTANCE" ]; then
```
jbertram, line 64: I couldn't get this to work in my testing. I kept getting this:
```
./artemis-service: 64: [: /home/jbertram: unexpected operator
```
However, this worked:
```
if [ $? -eq 0 ] && [ "`pwdx ${pid} | awk '{print $2}'`" = "$ARTEMIS_INSTANCE" ] ; then
```

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service
```diff
@@ -61,7 +61,7 @@ status() {
     pid=`cat "${PID_FILE}"`
     # check to see if it's gone...
     ps -p ${pid} > /dev/null
-    if [ $? -eq 0 ] ; then
+    if [ $? -eq 0 ] && [ "`pwdx ${pid} | awk '{print $2}'`" == "$ARTEMIS_INSTANCE" ]; then
```
jbertram, line 64: Unfortunately it's still broken. I still get the same error as before. You're using `==` when I believe you should be using `=`.

jbertram:
@angelogalvao, nice work. Thanks for the contribution!

PR #4810, opened 2024-02-06
ARTEMIS-4637 - Allow unordered xml conf elements for clusters and bri…

AntonRoskvist:
…dges

I hope this change makes sense as it's a little bit outside of my comfort zone... it's purpose is to fix a minor annoyance of mine and something Ive seen confuse some other users as well, where moving or adding a configuration element in the wrong place causes the config to be invalidated.

A side effect of this change is that the cluster property/attribute `allow-direct-connections-only` in `static-connectors` would pass validation even on core-bridges... setting this property has no effect on the core-bridges either way so I'm hoping it's not a cause for concern. I made it that way to have current configurations validate as before this change instead if adding it as a separate element within the cluster config or similar. Ideas on how to make this better if it's an issue are welcome.

Ideally this would apply to all configuration at some point but I believe that would require moving to xsd version 1.1 which is not scoped in this change.

### DIFF
--- a/artemis-server/src/main/resources/schema/artemis-configuration.xsd
+++ b/artemis-server/src/main/resources/schema/artemis-configuration.xsd
@@ -1054,6 +1054,37 @@
       </xsd:annotation>
    </xsd:element>
 
+   <xsd:element name="discovery-type" abstract="true"/>
+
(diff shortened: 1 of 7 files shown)


clebertsuconic:
this LGTM.


Thanks a lot for this!

PR #4811, opened 2024-02-06
ARTEMIS-4630 Validate if the pid is really from Artemis

angelogalvao:
Fix a typo in the orignial commit.

In rare cases the status is returning 'running' to a process
that is not Artemis. It was create a extra level of verification
by validating if the running process working directory is the
same of ${ARTEMIS_INSTANCE}.

### DIFF
--- a/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service
+++ b/artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/bin/artemis-service
@@ -59,9 +59,13 @@ fi
 status() {
   if [ -f "${PID_FILE}" ] ; then
     pid=`cat "${PID_FILE}"`
+    pid_workdir=$ARTEMIS_INSTANCE
+    if [ `command -v lsof &> /dev/null` ]; then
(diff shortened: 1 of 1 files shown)


gemmellr:
I cherry-picked the fix and pushed it since the PR had conflicts.

jbertram:
After further investigation I don't think this solution is valid. The problem is that `lsof -a -d cwd -p <pid>` will report the directory from which the `artemis-service` script _was executed_. However, this script can be invoked from *any* directory. Therefore, it's very easy to break the script with this change such that it actually starts the broker but reports that it didn't and then continues to report the wrong status.

I will revert these commits soon.

gemmellr:
(I actually only cherry-picked the second commit earlier, I thought the first was the same as yesterdays.)

Reverting seems reasonable based on what you just said. FWIW past 'check the process' stuff I have seen tended to just add something as a marker to the command run originally, so it could be looked for later.

jbertram:
These commits have been reverted.

@angelogalvao, I think if we want to address this problem we need to determine _how_ the wrong pid made it into `artemis.pid` in the first place.

PR #4815, opened 2024-02-07
ARTEMIS-4588 30 second pause for large msgs + federation

jbertram:
Large message support was added to
o.a.a.a.c.s.f.FederatedQueueConsumerImpl#onMessage via cf85d35 for ARTEMIS-3308. The problem with that change is that when onMessage returns o.a.a.a.c.c.i.ClientConsumerImpl#callOnMessage will eventually call o.a.a.a.c.c.i.ClientLargeMessageImpl#discardBody which eventually ends up in o.a.a.a.c.c.i.LargeMessageControllerImpl#popPacket waiting 30 seconds (i.e. the default readTimeout) for more packets to arrive (which never do). This happens because the FederatedQueueConsumer short-cuts the "normal" process by using LargeMessageControllerImpl#take.

This commit fixes that by tracking the number of bytes "taken" and then looking at that value later when discarding the body effectively skipping the 30 second wait.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/LargeMessageControllerImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/LargeMessageControllerImpl.java
@@ -83,6 +83,8 @@ public class LargeMessageControllerImpl implements LargeMessageController {
 
    private long packetLastPosition = -1;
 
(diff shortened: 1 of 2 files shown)


gtully (approved):
Looks sensible. I recall a comment that server side large message apis may be a better match but never got to investigate. I wonder if those would be relevant here? From what I recall I reused that take pattern from an example!

PR #4817, opened 2024-02-09
ARTEMIS-4641 Support events for add of missing federation resources

tabish121:
When an AMQP federation instance attempts to federate an address or queue it can fail if the remote address or queue is not present or cannot be created based on broker policy. A federation link can also closed if the federated resource is removed from the remote broker by management etc. In those cases the remote broker should note the resources that were targets of federation and send alerts to the source federation broker to notify it that these resources become available for federation and the source should attempt again to create federation links if demand still exists. This allows an AMQP federation instance to heal itself based on updates from the remote.


tabish121:
I've run the full test suite with these changes.

tabish121:
Made the cleanups requested and rebased.  Ran all AMQP federation tests again and all passed.

PR #4819, opened 2024-02-14
ARTEMIS-4642 Fix tracked federated demand when links are rejected or …

tabish121:
…blocked

Under some scenarios federation demand tracking is losing track of total demand for a federated resource leading to teardown of federated links before all local demand has been removed from the resource. This occurs most often if the attempts to establish a federation link are refused because the resource hasn't yet been created and an eventual attach succeeds, but can also occur in combination with a plugin blocking or not blocking federation link creation in some cases.


tabish121:
Ran the tests in CI and looks good

PR #4821, opened 2024-02-15
ARTEMIS-4644: convert broker-connection tests to use test peer

gemmellr:


gemmellr:
I ran the full test suite on the changes. The effective changes are really just in the 2 classes changed though.

tabish121:
LGTM

PR #4823, opened 2024-02-16
ARTEMIS-4646 Reuse only acknowledged MQTT IDs

thezbyg:
Generate MQTT message IDs from full allowed range of 1-65535 and skip currently used values. Do not use atomic integer for current ID, because all accesses and modifications are performed in synchronized context.

### DIFF
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionState.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionState.java
@@ -25,7 +25,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.atomic.AtomicInteger;
 import java.util.regex.Pattern;
 
 import io.netty.buffer.ByteBuf;
@@ -444,14 +443,13 @@ public class MQTTSessionState {
    }
 
    public class OutboundStore {
-
       private HashMap<Pair<Long, Long>, Integer> artemisToMqttMessageMap = new HashMap<>();
 
       private HashMap<Integer, Pair<Long, Long>> mqttToServerIds = new HashMap<>();
 
       private final Object dataStoreLock = new Object();
 
-      private final AtomicInteger ids = new AtomicInteger(0);
+      private int currentId = 0;
 
       private Pair<Long, Long> generateKey(long messageId, long consumerID) {
          return new Pair<>(messageId, consumerID);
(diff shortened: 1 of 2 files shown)


artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTSessionState.java
```diff
@@ -461,8 +459,14 @@ public int generateMqttId(long messageId, long consumerId) {
          synchronized (dataStoreLock) {
             Integer id = artemisToMqttMessageMap.get(generateKey(messageId, consumerId));
             if (id == null) {
-               ids.compareAndSet(Short.MAX_VALUE, 1);
-               id = ids.addAndGet(1);
+               do {
+                  if (currentId == 0xffff) {
```
jbertram, line 463: You can use `org.apache.activemq.artemis.core.protocol.mqtt.MQTTUtil#TWO_BYTE_INT_MAX` here instead of `0xffff`.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/MQTTTest.java
```diff
@@ -447,6 +447,59 @@ public void testSendMoreThanUniqueId() throws Exception {
       publisher.disconnect();
    }
 
+   @Ignore
```
jbertram, line 450: How come this test is ignored?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/MQTTTest.java
```diff
@@ -447,6 +447,59 @@ public void testSendMoreThanUniqueId() throws Exception {
...
+   @Test(timeout = 600 * 1000)
+   public void testNoMessageIdReuseBeforeAcknowledgment() throws Exception {
```
jbertram, line 452: This test doesn't fail without your fix. In order to demonstrate the problem and mitigate regressions the test should fail without the fix and succeed with it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/MQTTTest.java
thezbyg: I noticed a similar ignored test **testSendMoreThanUniqueId** and assumed that this new test should also be ignored because it also takes a long time run.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/mqtt/MQTTTest.java
thezbyg: Fixed test by using two initial messages. Previously it would not fail as first message had ID 1 and first ID after wraparound was 2.

jbertram:
@thezbyg, nice work. Thanks for the contribution!

PR #4824, opened 2024-02-16
ARTEMIS-4648 support typed properties from CLI producer

jbertram:


artemis-cli/src/test/java/org/apache/activemq/cli/test/CliProducerTest.java
```diff
@@ -54,9 +54,11 @@ public void tearDown() throws Exception {
       super.tearDown();
    }
 
-   private void produceMessages(String address, String message, int msgCount) throws Exception {
-      new Producer()
+   private Object produceMessages(String address, String message, int msgCount, String properties) throws Exception {
+      System.out.println(properties);
```
clebertsuconic, line 58: I wouldn't use System.out.println here.. would keep logger.debug

artemis-cli/src/test/java/org/apache/activemq/cli/test/CliProducerTest.java
```diff
@@ -88,6 +90,124 @@ public void testSendMessage() throws Exception {
       checkSentMessages(session, address, null);
    }
 
+   @Test
+   public void testBooleanMessageProperties() throws Exception {
```
clebertsuconic, line 94: can't you make a single test testMessageProperties... have multiple messages (or maybe just one) and validate them all in a single test?

gemmellr:
Related question. Does adding the ability to send properties, perhaps also suggest there would be utility in having the option for the CLI consumer to print such properties? E.g either print all, or specifically specified ones.

artemis-cli/src/test/java/org/apache/activemq/cli/test/CliProducerTest.java
jbertram: Yes. I could combine all the independent property tests into a single test. However, it's not clear to me what benefit that would provide.

jbertram:
@gemmellr, yes, I do think that's a worthy improvement for the consumer. I opened [ARTEMIS-4659](https://issues.apache.org/jira/browse/ARTEMIS-4659) for this.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/ProducerThread.java
```diff
@@ -190,6 +195,43 @@ protected Message createMessage(long i, String threadName) throws Exception {
...
+               case "float":
+                  answer.setFloatProperty(key, Float.parseFloat(value));
+                  break;
+               case "double":
+                  answer.setDoubleProperty(key, Double.parseDouble(value));
+                  break;
+               case "string":
+                  answer.setStringProperty(key, value);
+                  break;
+            }
```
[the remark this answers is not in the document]
jbertram: Done. I also added a new test to make sure incorrect types are dealt with properly.

artemis-cli/src/test/java/org/apache/activemq/cli/test/CliProducerTest.java
jbertram: I forgot to remove that.

artemis-cli/src/test/java/org/apache/activemq/cli/test/CliProducerTest.java
gemmellr: I'd guess that Clebert was just thinking about it starting X brokers and making X connections, sessions, producers currently to test something carried on the message alone that could still be done with 1 of each.

I tend to prefer more granular tests. Although, I'd personally go further and think the CLI bits could probably have been unit tested without any broker etc since its just a case of verifying the arg is decoded as expected, and ensuring things are set as expected on the message before it is sent...the clients sending of a message with props is covered elsewhere already)

artemis-cli/src/test/java/org/apache/activemq/cli/test/CliProducerTest.java
```diff
@@ -88,6 +91,138 @@ public void testSendMessage() throws Exception {
...
+      PrintStream originalErr = System.err;
+      ByteArrayOutputStream baos = new ByteArrayOutputStream();
+      PrintStream newErr = new PrintStream(baos);
+      System.setErr(newErr);
+      try {
+         produceMessages("test", null, 1, "[{'type':'myType','key':'myKey','value':'myValue'}]".replaceAll("'", "\""));
+         assertEquals("Unable to set property: myKey. Did not recognize type: myType. Supported types are: boolean, int, long, byte, short, float, double, string.\n", baos.toString());
+      } finally {
+         System.setErr(originalErr);
+      }
```
gemmellr, line 223: Ick on messing with the System. I believe most of the regular artemis CLI commands have a 'context' of some kind thats used for the output, which allows testing the output without this kind of hack. Doesnt the Producer?

jbertram: The `ProducerThread` logs directly using `System`. I sent #4846 to fix this. Once that's merged I'll update this test.

artemis-cli/src/test/java/org/apache/activemq/cli/test/CliProducerTest.java
jbertram: I consolidated all the individual tests into one.

jbertram: I like the idea of unit tests vs. integration tests so I implemented that.

jbertram: I actually attempted to use a mock to begin with but wasn't familiar enough (or patient enough) to get it working. Thanks for the commit. I've squashed and pushed.

gemmellr: I actually meant push to main since I already reviewed them, should have been more specific given the context hehe, woops. Merged now :)

PR #4825, opened 2024-02-17
ARTEMIS-4649 Use unique STOMP message ID per subscription

thezbyg:
Use combined consumer and message ID as STOMP message ID.

### DIFF
--- a/artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/VersionedStompFrameHandler.java
+++ b/artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/VersionedStompFrameHandler.java
@@ -29,6 +29,7 @@ import org.apache.activemq.artemis.core.protocol.stomp.Stomp.Headers;
 import org.apache.activemq.artemis.core.protocol.stomp.v10.StompFrameHandlerV10;
 import org.apache.activemq.artemis.core.protocol.stomp.v11.StompFrameHandlerV11;
 import org.apache.activemq.artemis.core.protocol.stomp.v12.StompFrameHandlerV12;
+import org.apache.activemq.artemis.core.server.ServerConsumer;
 import org.apache.activemq.artemis.utils.ExecutorFactory;
 
 import static org.apache.activemq.artemis.core.protocol.stomp.ActiveMQStompProtocolMessageBundle.BUNDLE;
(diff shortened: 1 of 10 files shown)


jbertram:
The tests you added don't actually fail without the fix. This `WARN` message is logged:
```
WARN  [org.apache.activemq.artemis.core.protocol.stomp] AMQ332069: Sent ERROR frame to STOMP client /127.0.0.1:43064: AMQ339026 subscription id sub2 does not match sub1
```
However, the test succeeds. The tests should fail without the fix.

jbertram:
Can you squash your commits?

artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/VersionedStompFrameHandler.java
```diff
@@ -344,7 +348,8 @@ public StompFrame createMessageFrame(ICoreMessage serverMessage, StompSubscripti
       }
       frame.setByteBody(data);
 
-      StompUtils.copyStandardHeadersFromMessageToFrame((serverMessage), frame, deliveryCount);
+      frame.addHeader(Stomp.Headers.Message.MESSAGE_ID, String.valueOf(consumer.getID()) + StompSession.MESSAGE_ID_SEPARATOR + String.valueOf(serverMessage.getMessageID()));
```
jbertram, line 351: This is going to be done for every message dispatched from the broker so it would be worth making this is as fast as possible. Perhaps something like:
```java
frame.addHeader(Stomp.Headers.Message.MESSAGE_ID, new StringBuilder(32).append(consumer.getID()).append(StompSession.MESSAGE_ID_SEPARATOR).append(serverMessage.getMessageID()).toString());
```

thezbyg: Done, but I set initial StringBuilder capacity to 41 to fit two long values and separator.

jbertram: How did you arrive at the `41` value?

thezbyg: Minimum long value is 20 characters (-9223372036854775808), so two long values and separator could require 41 characters total.

jbertram: Message and consumer IDs should never be < 0 so the potential values are 0 (1 character) through 9223372036854775807 (19 characters). Since these characters are just digits (i.e. part of ISO-8859-1) then they will be represented by a single byte in the `byte[]` that backs the `String`. We want to avoid expanding the backing array but we also want to avoid over-allocating (and therefore wasting) space. Ten characters for each long seems like a fair middle ground.

jbertram:
@thezbyg, thanks for the contribution!

PR #4828, opened 2024-02-20
ARTEMIS-4652 rollback of XAResource impl shouldn't return XA_RETRY

jbertram:
This fix was based on static analysis of the code and inspection of the XA specification. There is no test associated with it due to the difficult nature in reproducing the failure. This code has been essentially the same for a decade and only now have there been any reports of it actually sending back the wrong XA code.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
@@ -1901,18 +1901,10 @@ public final class ClientSessionImpl implements ClientSessionInternal, FailureLi
          workDone = false;
       } catch (XAException xae) {
          throw xae;
-      } catch (ActiveMQException e) {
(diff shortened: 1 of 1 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionImpl.java
```diff
@@ -1901,18 +1901,10 @@ public void rollback(final Xid xid) throws XAException {
...
-         // This should never occur
-         XAException xaException = new XAException(XAException.XAER_RMFAIL);
-         xaException.initCause(e);
-         throw xaException;
       } catch (Throwable t) {
-         // This could occur if the TM interrupts the thread
+         if (logger.isTraceEnabled()) {
+            logger.trace("Rollback failed:: {}", convert(xid), t);
+         }
          XAException xaException = new XAException(XAException.XAER_RMFAIL);
```
clebertsuconic, line 1908: would the TM retry the rollback in such event?

Did this request come from Wildly / @ehsavoie ?

My biggest concern in such changes is the integration with MDBs as I remember having massive headaches with MDB / XA tests back then which I don't want to go back :)

clebertsuconic: @jbertram @ehsavoie I merged this, but this was of course intentional as there was a test validating it. I sent a further commit to fix the test.


So, if you decide against this @ehsavoie please send a revert.

PR #4833, opened 2024-02-27
ARTEMIS-4657 support byte[] correlation ID compat b/w core & AMQP JMS

jbertram:
This is a prelude to implementing the correlation ID support for https://issues.apache.org/jira/browse/ARTEMIS-4264 (i.e. interop between MQTT 5 & JMS clients).

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
@@ -20,6 +20,7 @@ import javax.management.openmbean.CompositeData;
 import javax.management.openmbean.CompositeDataSupport;
 import javax.management.openmbean.OpenDataException;
 import javax.management.openmbean.SimpleType;
+import java.lang.invoke.MethodHandles;
 import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.Collections;
@@ -80,7 +81,6 @@ import org.apache.qpid.proton.message.Message;
 import org.apache.qpid.proton.message.impl.MessageImpl;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import java.lang.invoke.MethodHandles;
 
 import static org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport.getCharsetForTextualContent;
 
@@ -1562,7 +1562,7 @@ public abstract class AMQPMessage extends RefCountMessage implements org.apache.
             return getAMQPUserID();
          case MessageUtil.CORRELATIONID_HEADER_NAME_STRING:
             if (properties != null && properties.getCorrelationId() != null) {
-               return AMQPMessageIdHelper.INSTANCE.toCorrelationIdString(properties.getCorrelationId());
+               return AMQPMessageIdHelper.INSTANCE.toCorrelationIdStringOrBytes(properties.getCorrelationId());
             }
             return null;
          default:
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageIdHelper.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageIdHelper.java
@@ -113,7 +113,7 @@ public class AMQPMessageIdHelper {
       }
    }
 
-   public String toCorrelationIdString(Object idObject) {
+   public Object toCorrelationIdStringOrBytes(Object idObject) {
       if (idObject instanceof String) {
          final String stringId = (String) idObject;
 
@@ -130,6 +130,11 @@ public class AMQPMessageIdHelper {
             // It has "ID:" prefix and doesn't have encoding prefix, use it as-is.
             return stringId;
          }
+      } else if (idObject instanceof Binary) {
+         ByteBuffer dup = ((Binary) idObject).asByteBuffer();
+         byte[] bytes = new byte[dup.remaining()];
+         dup.get(bytes);
+         return bytes;
       } else {
          // Not a string, convert it
          return convertToIdString(idObject);
--- a/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
+++ b/artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
@@ -28,6 +28,9 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.lang.invoke.MethodHandles;
+import java.nio.ByteBuffer;
+import java.nio.charset.MalformedInputException;
+import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -159,7 +162,7 @@ public final class OpenWireMessageConverter {
       coreMessage.putIntProperty(OpenWireConstants.AMQ_MSG_COMMAND_ID, messageSend.getCommandId());
       final String corrId = messageSend.getCorrelationId();
       if (corrId != null) {
-         coreMessage.putStringProperty(OpenWireConstants.JMS_CORRELATION_ID_PROPERTY, new SimpleString(corrId));
+         coreMessage.setCorrelationID(corrId);
       }
       final DataStructure ds = messageSend.getDataStructure();
       if (ds != null) {
@@ -590,9 +593,15 @@ public final class OpenWireMessageConverter {
       }
       amqMsg.setCommandId(commandId);
 
-      final SimpleString corrId = getObjectProperty(coreMessage, SimpleString.class, OpenWireConstants.JMS_CORRELATION_ID_PROPERTY);
-      if (corrId != null) {
-         amqMsg.setCorrelationId(corrId.toString());
+      final Object correlationID = coreMessage.getCorrelationID();
+      if (correlationID instanceof String || correlationID instanceof SimpleString) {
+         amqMsg.setCorrelationId(correlationID.toString());
+      } else if (correlationID instanceof byte[]) {
+         try {
+            amqMsg.setCorrelationId(StandardCharsets.UTF_8.newDecoder().decode(ByteBuffer.wrap((byte[]) correlationID)).toString());
+         } catch (MalformedInputException e) {
+            ActiveMQServerLogger.LOGGER.unableToDecodeCorrelationId(e.getMessage());
+         }
       }
 
       final byte[] dsBytes = getObjectProperty(coreMessage, byte[].class, OpenWireConstants.AMQ_MSG_DATASTRUCTURE);
@@ -943,6 +952,8 @@ public final class OpenWireMessageConverter {
             continue;
          }
          if (!coreMessage.containsProperty(ManagementHelper.HDR_NOTIFICATION_TYPE) && (keyStr.startsWith("_AMQ") || keyStr.startsWith("__HDR_"))) {
+            continue;
+         } else if (s.equals(OpenWireConstants.JMS_CORRELATION_ID_PROPERTY)) {
             continue;
          }
          final Object prop = coreMessage.getObjectProperty(s);
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSCorrelationIDTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSCorrelationIDTest.java
@@ -1,0 +1,184 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.tests.integration.jms.multiprotocol;
+
+import javax.jms.Connection;
+import javax.jms.Message;
+import javax.jms.MessageConsumer;
+import javax.jms.MessageProducer;
+import javax.jms.Queue;
+import javax.jms.Session;
+
+import org.apache.activemq.artemis.tests.util.RandomUtil;
+import org.junit.Assert;
+import org.junit.Ignore;
+import org.junit.Test;
+
+/*
+ * JMS supports setting the correlation ID as a String or a byte[]. However, OpenWire only supports correlation ID as
+ * a String. When it is set as a byte[] the OpenWire JMS client just converts it to a UTF-8 encoded String, and
+ * therefore when it sends a JMS message with a correlation ID the broker can't tell if the value was set as a String
+ * or a byte[]. Due to this ambiguity the broker is hard-coded to treat the incoming OpenWire value as a String. This
+ * doesn't cause any problems if the consumer is also OpenWire, but if the consumer is Core or AMQP (which both
+ * differentiate between String and binary values) then retrieving the correlation ID as a byte[] will fail and nothing
+ * can be done about it aside from updating the OpenWire protocol.
+ *
+ * Therefore, all the tests which involve the OpenWire JMS client using Message.setJMSCorrelationIDAsBytes() on a
+ * message sent to a different JMS implementation are ignored. The test are ignored rather that being completely
+ * removed to make clear this was an explicit decision not to test & support this use-case.
+ */
+public class JMSCorrelationIDTest extends MultiprotocolJMSClientTestSupport {
+
+   private void testCorrelationIDAsBytesSendReceive(Connection producerConnection, Connection consumerConnection) throws Throwable {
+      Session session = producerConnection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+      Queue queue = session.createQueue(getQueueName());
+
+      byte[] bytes = new byte[0xf + 1];
+      for (int i = 0; i <= 0xf; i++) {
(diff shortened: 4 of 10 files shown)


artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -590,9 +593,14 @@ private static ActiveMQMessage toAMQMessage(MessageReference reference,
       }
       amqMsg.setCommandId(commandId);
 
-      final SimpleString corrId = getObjectProperty(coreMessage, SimpleString.class, OpenWireConstants.JMS_CORRELATION_ID_PROPERTY);
-      if (corrId != null) {
-         amqMsg.setCorrelationId(corrId.toString());
+      final Object correlationID = coreMessage.getCorrelationID();
+      if (correlationID != null) {
+         if (correlationID instanceof String || correlationID instanceof SimpleString) {
```
tabish121, line 598: Can the core message carry something else besides SimpleString, String or byte[] and if so should it be ignored as it is now?  If not should the if be inverted to simply check for byte[] and just toString anything else?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageIdHelper.java
```diff
@@ -130,6 +130,8 @@ public String toCorrelationIdString(Object idObject) {
             // It has "ID:" prefix and doesn't have encoding prefix, use it as-is.
             return stringId;
          }
+      } else if (idObject instanceof Binary) {
+         return ((Binary)idObject).getArray();
```
gemmellr, line 134: Strictly speaking, its possible the array isnt just the id...the Binary should be checked that it doesnt have an array offset and is the same length as the array.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSCorrelationIDTest.java
```diff
@@ -0,0 +1,186 @@
...
+    * will fail.
+    *
+    * JMS means for the correlation ID as a byte[] to be used for "native" clients which makes it a good candidate for
+    * interoperability between other protocols like MQTT 5 which *only* supports correlation ID as byte[].
+    */
+   @Ignore
+   @Test(timeout = 60000)
+   public void testCorrelationIDAsStringSendReceiveFromOpenWireToAMQP() throws Throwable {
+      testCorrelationIDAsStringSendReceive(createOpenWireConnection(), createConnection());
+   }
```
gemmellr, line 179: Seems that you actually hit the interop issues I commented on from looking at the code.

Why is the broker 'hard coded to byte[]' for Openwire when this comment explicitly notes it effectively only does String? Why isnt it hard coded to using String...like it was before?

If MQTT only supports byte[] then it seem like it is the MQTT stuff that should be jumping through hoops such as converting to/from a UTF-8 bytes, not the Openwire bits, especially as doing it this way breaks the typical+existing Openwire<->AMQP/Core interop.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -1562,7 +1562,7 @@ public final Object getObjectProperty(String key) {
             return getAMQPUserID();
          case MessageUtil.CORRELATIONID_HEADER_NAME_STRING:
             if (properties != null && properties.getCorrelationId() != null) {
-               return AMQPMessageIdHelper.INSTANCE.toCorrelationIdString(properties.getCorrelationId());
+               return AMQPMessageIdHelper.INSTANCE.toCorrelationIdStringOrBytes(properties.getCorrelationId());
```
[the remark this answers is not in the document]
jbertram: This does change the semantics of the original `toCorrelationIdString`. Now if the ID is a `Binary` it will return a `byte[]` instead of a `String`. If we want to support compatibility between JMS clients as well as other protocols which use `byte[]` correlation IDs I don't see a way around a change like this. Currently no JMS client is compatible with another using `setJMSCorrelationIDAsBytes`.

jbertram:
If you executed the `JMSCorrelationIDTest` I added in d38b5a7 without the other changes this would be the result:

- Correlation ID as `byte[]`
    - :white_check_mark: AMQP to AMQP
    - :x: AMQP to Core
    - :x: AMQP to OpenWire
    - :white_check_mark: Core to Core
    - :x: Core to AMQP
    - :x: Core to OpenWire
    - :white_check_mark: OpenWire to OpenWire
    - :x: OpenWire to AMQP
    - :x: OpenWire to Core
- Correlation ID as `String`
    - :white_check_mark: AMQP to AMQP
    - :white_check_mark: AMQP to Core
    - :white_check_mark: AMQP to OpenWire
    - :white_check_mark: Core to Core
    - :white_check_mark: Core to AMQP
    - :white_check_mark: Core to OpenWire
    - :white_check_mark: OpenWire to OpenWire
    - :white_check_mark: OpenWire to AMQP
    - :white_check_mark: OpenWire to Core

So...right now there is full compatibility with `String` among JMS clients and no compatibility with `byte[]`.

If you executed the `JMSCorrelationIDTest` I added in d38b5a7 along with the other changes this would be the result:

- Correlation ID as `byte[]`
    - :white_check_mark: AMQP to AMQP
    - :white_check_mark: AMQP to Core
    - :white_check_mark: AMQP to OpenWire
    - :white_check_mark: Core to Core
    - :white_check_mark: Core to AMQP
    - :white_check_mark: Core to OpenWire
    - :white_check_mark: OpenWire to OpenWire
    - :white_check_mark: OpenWire to AMQP
    - :white_check_mark: OpenWire to Core
- Correlation ID as `String`
    - :white_check_mark: AMQP to AMQP
    - :white_check_mark: AMQP to Core
    - :white_check_mark: AMQP to OpenWire
    - :white_check_mark: Core to Core
    - :white_check_mark: Core to AMQP
    - :white_check_mark: Core to OpenWire
    - :white_check_mark: OpenWire to OpenWire
    - :x: OpenWire to AMQP
    - :x: OpenWire to Core

So...the changes in d38b5a7 are sacrificing 2 OpenWire `String` use-cases for full compatibility with `byte[]`. Whether or not this is OK is the subject of this PR.

If I hard-code the broker to deal with the OpenWire correlation ID as a **`String`** these are the results:

- Correlation ID as `byte[]`
    - :white_check_mark: AMQP to AMQP
    - :white_check_mark: AMQP to Core
    - :x: AMQP to OpenWire
    - :white_check_mark: Core to Core
    - :white_check_mark: Core to AMQP
    - :x: Core to OpenWire
    - :white_check_mark: OpenWire to OpenWire
    - :x: OpenWire to AMQP
    - :x: OpenWire to Core
- Correlation ID as `String`
    - :white_check_mark: AMQP to AMQP
    - :white_check_mark: AMQP to Core
    - :white_check_mark: AMQP to OpenWire
    - :white_check_mark: Core to Core
    - :white_check_mark: Core to AMQP
    - :white_check_mark: Core to OpenWire
    - :white_check_mark: OpenWire to OpenWire
    - :white_check_mark: OpenWire to AMQP
    - :white_check_mark: OpenWire to Core

In short, full compatibility with `String` and no compatibility with OpenWire with `byte[]`.

Perhaps our hands are tied here and we just can't support OpenWire correlation ID as `byte[]` since it will break existing use-cases. I'm running the full test-suite now, and if there are existing test-cases that break then that will help convince me. However, at this point it seems to me like using `byte[]` to gain compatibility between JMS and a "native" format (e.g. core) is actually the right thing to do so supporting `byte[]` across all protocols at the sacrifice of `String` is appropriate especially since this will open the door to compatibility with MQTT 5 (although I grant that's likely an edge case).

I'm, of course, open to being convinced otherwise. I'd love your thoughts.

tabish121:
> Perhaps our hands are tied here and we just can't support OpenWire correlation ID as `byte[]` since it will break existing use-cases. I'm running the full test-suite now, and if there are existing test-cases that break then that will help convince me. However, at this point it seems to me like using `byte[]` to gain compatibility between JMS and a "native" format (e.g. core) is actually the right thing to do so supporting `byte[]` across all protocols at the sacrifice of `String` is appropriate especially since this will open the door to compatibility with MQTT 5 (although I grant that's likely an edge case).
> 
> I'm, of course, open to being convinced otherwise. I'd love your thoughts.

Given what Openwire does (treats all correlation IDs as String even the byte[] ones) I'd have chosen to keep incoming Openwire message correlation IDs as String upon conversion to Core as you have no visibility into what it started as but most likely guess is probably that it was a String (set from getJMSMessageID).  On the outbound from AMQP / Core / MQTT to Openwire you could UTF-8 encode any binary value as Openwire again is not making it clear what the source value started as and it is just going to UTF8 decode it when you call getJMSCorrelationIDAsBytes

Core and AMQP actually won't return anything (Qpid JMS will throw) from getJMSCorrelationIDAsBytes if it didn't arrive as a byte[] but that seems a semi-reasonable outcome when going from Openwire to Core / AMQP as again Openwire gives you not hint as to what it was at the start so failing on getJMSCorrelationIDAsBytes for that case is reasonable.

gemmellr:
Sacrificing any of the existing widely-used working String-based correlation ID interop in favour of the non-portable little-used byte[] versions, that dont work currently, simply does not strike me as an improvement whatsoever.

String based JMSCorrelationID is the _primary, required_ form of JMSCorrelationID. Its what JMS providers _must_ implement. Its the only way to correlate using the String-only JMSMessageID values. Its what people use in selectors. Its what applications just tend to use in general (I dont think I have ever come across a real world application that used the byte[], whereas I know I have come across lots that use the String).

Its also what frameworks tend to use, because they simply cant depend on the byte[] method and thus avoid it. The Camel repo for example is littered with String-based setJMSCorrelationID usage, but doesnt seem to have a single setJMSCorrelationIDAsBytes usage (according to a couple searches of their git repo). Ditto for Spring.

The byte[] method is optional, you do not need to implement it at all and are simply allowed to throw UnsupportedOperationException. The API expressly notes that it is non-portable to discourage its use. The TCK only has 1 test variant (only only added with JMS 2) that uses it, and it simply swallows that potential UOE and prints it is ignoring it.  The method is essentially only there to support niche cases of providers that also have non-JMS clients that require a specific 'native' correlation encoding. I think I historically saw it referenced for interop with the 'non-jms' IBM MQ clients for example, as they encoded things in a certain structure with certain hex encodings etc that need to be matched.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/AMQPMessageIdHelper.java
jbertram: I think I get what you mean. Check the latest push.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -590,9 +590,9 @@ private static ActiveMQMessage toAMQMessage(MessageReference reference,
       }
       amqMsg.setCommandId(commandId);
 
-      final SimpleString corrId = getObjectProperty(coreMessage, SimpleString.class, OpenWireConstants.JMS_CORRELATION_ID_PROPERTY);
-      if (corrId != null) {
-         amqMsg.setCorrelationId(corrId.toString());
+      final Object correlationID = coreMessage.getCorrelationID();
+      if (correlationID != null) {
+         amqMsg.setCorrelationId(correlationID.toString());
```
gemmellr, line 595: The old code looks like it would get null if the types didnt match, and so not actually add thing...this looks more like it will start adding the toString() of non-Strings. Perhaps it should still be type-checking to avoid that, given the other changes make it more likely for this to receive non-String values now?

gemmellr: (Or also, as Tim suggested, potentially try UTF8 encoding e.g bytes received at this point).

gemmellr: I think the actual-expected-behaviour should be tested. This specific behaviour wasnt noticed as there is a simply a 'disabled test as it wont send matching bytes', which is not covering what actually was/is expected to happen in this situation either before or after the changes here.

jbertram: I added a couple tests which I believe should cover the relevant possibilities.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -590,9 +590,9 @@ private static ActiveMQMessage toAMQMessage(MessageReference reference,
       }
       amqMsg.setCommandId(commandId);
 
-      final SimpleString corrId = getObjectProperty(coreMessage, SimpleString.class, OpenWireConstants.JMS_CORRELATION_ID_PROPERTY);
-      if (corrId != null) {
-         amqMsg.setCorrelationId(corrId.toString());
+      final Object correlationID = coreMessage.getCorrelationID();
```
tabish121, line 593: It seems like we could retain support for carried byte[] type correlation IDs by decoding a UTF8 string here as the receiving Openwire client would either decode it if the getter for bytes was called or would just return a string value from the non-bytes getter, which allows basic interop.  Equally there are types that could be carried in an AMQP message correlation ID like a UUID that could make sense to convert to string as well as it could allow some edge case interop for receivers wanting to sync up correlation ID values.   It might make sense to have some consideration for those cases and what we should or shouldn't be allowing to propagate to an Openwire client.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
gemmellr: So the expectation is an Openwire client will now always receive null if a non-Openwire sender sends byte[], as opposed to previously getting a string-encoding of the value if the sender was AMQP?

Seems like it we are going to change it, it could instead try sending the bytes as a string, if they are UTF-8...or keep doing what it did before for AMQP.

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
jbertram: I updated this bit to support `byte[]` conversion to UTF8 `String`. At this point `UUID` is left out. We can add that later if necessary.

tabish121: That should be fine as the current AMQP to Core converter stringifies the correlation ID so it will carry through as a somewhat identifiable version of the original type.

jbertram:
ARTEMIS-4657

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
jbertram: @gemmellr, is this ready to merge in your opinion?

gemmellr: I commented on a new aspect related to changes made from this comment, lets move over there :)

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -590,9 +591,11 @@ private static ActiveMQMessage toAMQMessage(MessageReference reference,
...
+      if (correlationID instanceof String || correlationID instanceof SimpleString) {
+         amqMsg.setCorrelationId(correlationID.toString());
+      } else if (correlationID instanceof byte[]) {
+         amqMsg.setCorrelationId(new String((byte[])correlationID, StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
jbertram: I thought about this previously, and I tried to come up with a test to break it. However, I couldn't come up with "not actually UTF-8 decodable content." Do you have an example?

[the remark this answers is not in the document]
tabish121: You could send bytes for a normal two byte encoding where the second byte violates the prefix value of the encoding rules.  The UTF8 spec uses upper bits of the bytes to stretch out the characters encodings across subsequent bytes, looking into the proton-j read UTF8 reading code in ReadableBuffer would likely shed a bit of light on how to corrupt a trailing encoded byte.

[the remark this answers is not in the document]
jbertram: After doing some more testing directly with `new String(byte[])` I see now why I wasn't able to generate a failure. This is what the JavaDoc for that method states (in part):

> This method always replaces malformed-input and unmappable-character sequences with this charset's default replacement string. 

So this won't blow up and tank the message conversion even with "not actually UTF-8 decodable content."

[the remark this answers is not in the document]
gemmellr: Yeah I forgot about that nugget; so instead of blowing up or sending nothing it will just send a corrupted representation of the original bytes. Arguably worse? :)

Per the next bit of that javadoc, you'd need to actually decode the bytes as UTF-8 rather than just creating a String instance with them, so that the decoder would instead blow up for garbage rather than substitute. Tim's idea for malformed bytes is simple, but as per your link the simplest is actually just a single byte that isnt an ASCII char or the multi-byte prefix.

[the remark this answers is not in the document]
jbertram: It will now return `null` for correlation ID when the bytes are invalid UTF-8 and log a `WARN`.

PR #4834, opened 2024-02-28
ARTEMIS-4658 Prevent reflections when using dual address federation

tabish121:
When federation is configured in two directions between nodes for an address the message can reflect from one node to another if max hops is not set or not set correctly and in some federation topologies the max hops value can't solve the issue and still result in a working configuration. This reflection should be prevented at the federation consumer level for address consumers.


tabish121:
I've done a few full runs of the tests to check for regressions

PR #4836, opened 2024-02-29
ARTEMIS-4664 - autoCreatedResource can get removed while receiving ba…

AntonRoskvist:
…tch of messages

This is probably not all that likely to ever happen to anyone. I stumbled on this issue while working on something else but I figured it would be better to handle it.

From what I can tell the auto created resource would have to have been empty for one pass of the "AddressQueueReaper" to be marked as "swept", then a large batch of messages would get sent to the resource just as the next pass of  "AddressQueueReaper" begins and finishes before the message batch finishes.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
@@ -4090,6 +4090,9 @@ public class ActiveMQServerImpl implements ActiveMQServer {
             }
             final QueueBinding queueBinding = (QueueBinding) rawBinding;
             if (ignoreIfExists) {
+               //Reset potentially ongoing auto-delete status of queue
+               queueBinding.getQueue().setSwept(false);
+
                return queueBinding.getQueue();
             } else {
                throw ActiveMQMessageBundle.BUNDLE.queueAlreadyExists(queueConfiguration.getName(), queueBinding.getAddress());
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/AutoCreateDeadLetterResourcesTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/AutoCreateDeadLetterResourcesTest.java
@@ -27,6 +27,8 @@ import org.apache.activemq.artemis.api.core.client.ClientProducer;
 import org.apache.activemq.artemis.api.core.client.ClientSession;
 import org.apache.activemq.artemis.api.core.client.ClientSessionFactory;
 import org.apache.activemq.artemis.api.core.client.ServerLocator;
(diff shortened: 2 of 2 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3919,7 +3919,10 @@ private void createResources(boolean isAutoCreate, SimpleString destinationAddre
             SimpleString destinationQueueName = prefix.concat(getAddress()).concat(suffix);
             SimpleString filter = new SimpleString(String.format("%s = '%s'", Message.HDR_ORIGINAL_ADDRESS, getAddress()));
             try {
-               server.createQueue(new QueueConfiguration(destinationQueueName).setAddress(destinationAddress).setFilterString(filter).setAutoCreated(true).setAutoCreateAddress(true), true);
+               Queue destinationQueue = server.createQueue(new QueueConfiguration(destinationQueueName).setAddress(destinationAddress).setFilterString(filter).setAutoCreated(true).setAutoCreateAddress(true), true);
+               if (destinationQueue != null && destinationQueue.isSwept()) {
+                  destinationQueue.setSwept(false);
```
clebertsuconic, line 3924: shouldn't this be done here instead?

https://github.com/apache/activemq-artemis/blob/5ce70f9e37ba1bb7c56372c2e79f9e9d8c9d1fa5/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java#L4093

clebertsuconic: Meaning.. if anyone called createQueue, and it already existed.. the swept should be set to false I think. that means the fix would apply to any other case where createQueue was called.

AntonRoskvist: Sure, that's no problem. The intent being to also catch some similar circumstances outside of the auto created resources, right?

clebertsuconic: The sweep is only used for the reaper.  So it wouldn't hurt to set it to false in there. 

I could say set it if it was auto created.  But it shouldn't matter.

clebertsuconic: I run my complete CI with your changes (before my requested change) and they were all good.

I will merge as soon as you amend the change.

AntonRoskvist: Should be done now, hope it looks good... only thing I'm considering is if the `setSwept` should be done before the `ignoreIfExists`-check instead of after? Decided against it for now but I can change again if that makes sense.

PR #4838, opened 2024-03-01
ARTEMIS-4665 Fix AMQP federation pull consumer test failures

tabish121:
Fix intermittent test failures in pull consumer test by asserting that there are the expected number of message on the queue before running the JMS consume cycle to consume credit and trigger federation credit to flow.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationQueuePolicyTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationQueuePolicyTest.java
@@ -103,6 +103,7 @@ import org.apache.activemq.artemis.protocol.amqp.federation.FederationReceiveFro
 import org.apache.activemq.artemis.protocol.amqp.proton.AmqpSupport;
 import org.apache.activemq.artemis.tests.integration.amqp.AmqpClientTestSupport;
 import org.apache.activemq.artemis.tests.util.CFUtil;
+import org.apache.activemq.artemis.utils.Wait;
 import org.apache.qpid.proton.amqp.transport.AmqpError;
 import org.apache.qpid.proton.amqp.transport.LinkError;
 import org.apache.qpid.protonj2.test.driver.ProtonTestClient;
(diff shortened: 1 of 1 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationQueuePolicyTest.java
```diff
@@ -2782,6 +2785,8 @@ public void testPullQueueConsumerBatchCreditTopUpAfterEachBacklogDrain() throws
                                     .now();
             }
 
+            Wait.assertTrue(() -> server.queueQuery(queueName).getMessageCount() == DEFAULT_PULL_CREDIT_BATCH_SIZE);
```
clebertsuconic, line 2788: shouldn't it be >=?



Also, I have been using timeouts on all the assertEquals and assertTrues I have been adding.


I have been procrastinating removing the default method and always requiring a timeout set. as an eventual failure here would take 30seconds to happen.

tabish121: I suppose it could be although the test only sends DEFAULT_PULL_CREDIT_BATCH_SIZE message so being larger than that would indicate an intrinsic failure on the part of the broker to queue the messages sent in a rational manner.

tabish121: Setting a different timeout assumes one can guess how slow / fast the CI machine is which seems tough given the failures I've been seeing.  Personally I tend to be more concerned with the lengthy amount of time it takes to run the existing tests that do pass rather than the occasional failures due to timeout.  

I'd recommend lowering the default timeout value if that is of a major concern, drop it to 5 or 10 seconds as a baseline.

clebertsuconic: @tabish121 I tried that and it wasn't simple for some reason I don't remember now.. I think some tests were failing. 

I talked to Robbie about that and we thought about removing the default timeout method.


this is a different discussion of course.

PR #4840, opened 2024-03-04
ARTEMIS-4668 Moving AMQP Large Message file handling away from Netty thread

clebertsuconic:


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonDeliveryHandler.java
```diff
@@ -16,6 +16,7 @@
  */
 package org.apache.activemq.artemis.protocol.amqp.proton;
 
+import org.apache.activemq.artemis.api.core.Message;
```
tabish121, line 19: Checkstyle violations causes build errors

tests/artemis-test-support/src/main/java/org/apache/activemq/transport/amqp/client/AmqpMessage.java
```diff
@@ -640,6 +640,11 @@ public String getText() throws NoSuchElementException {
       throw new NoSuchElementException("Message does not contain a String body");
    }
 
+   public byte[] getBytes() {
+      Data body = (Data)getWrappedMessage().getBody();
+      return body.getValue().getArray();
```
tabish121, line 645: The byte array contained within is not guaranteed to be zero indexed or span the full array, you need to check and account for that if you are going to expose it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageWriter.java
```diff
@@ -81,11 +84,20 @@ public boolean isWriting() {
    public void close() {
       if (!closed) {
          try {
+            try {
+               if (largeBodyReader != null) {
+                  largeBodyReader.close();
+               }
+               largeBodyReader = null;
```
tabish121, line 91: Should be nulled in a finally block, which would happen in the resetClosed method called in the main finally block if you added a null there for the large body reader as you should.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageWriter.java
```diff
@@ -96,18 +108,36 @@ public AMQPLargeMessageWriter open() {
...
+
+      try {
+         largeBodyReader = message.getLargeBodyReader();
+         largeBodyReader.open();
+      } catch (Exception e) {
+         serverSender.reportDeliveryError(this, reference, e);
+      }
+   }
+
+   private void resetClosed() {
```
tabish121, line 128: added member large body reader is no null as it should be here

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
```diff
@@ -77,39 +74,68 @@ public AMQPLargeMessageReader open() {
...
    @Override
-   public Message readBytes(Delivery delivery) throws Exception {
+   public boolean readBytes(Delivery delivery) throws Exception {
       if (closed) {
          throw new IllegalStateException("AMQP Large Message Reader is closed and read cannot proceed");
       }
 
+      serverReceiver.connection.requireInHandler();
+
+      logger.trace("Reading {}", delivery);
```
tabish121, line 84: This trace isn't likely to add much useful context just more log noise

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
```diff
@@ -77,39 +74,68 @@ public AMQPLargeMessageReader open() {
...
+
+         if (!isPartial) {
+            final AMQPLargeMessage message = currentMessage;
+            message.releaseResources(serverReceiver.getConnection().isLargeMessageSync(), true);
+            logger.trace("finishing {} on {}", currentMessage, this);
+            // We don't want a close to delete the file now, we've released the resources.
+            currentMessage = null;
+            close();
+            serverReceiver.connection.runNow(() -> {
+               DeliveryAnnotations deliveryAnnotations = message.getDeliveryAnnotations();
```
tabish121, line 128: Can skip temp var creation and just call message.getDeliveryAnnotations directly.

tabish121 (commented):
There are a number of unused imports left scattered about now that need cleaned up besides the items listed in the review.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
```diff
@@ -847,6 +847,7 @@ public MessageWriter selectOutgoingMessageWriter(ProtonServerSenderContext sende
             if (message.isLargeMessage()) {
                selected = largeMessageWriter != null ? largeMessageWriter :
                   (largeMessageWriter = new AMQPLargeMessageWriter(sender));
+               largeMessageWriter.openContext(reference);
```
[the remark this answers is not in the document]
clebertsuconic: thing is.. it's always asynchronous... except to large messages... (since the previous interaction already you had the current large message returning something to make the handler return BUSY).

So.. in all the other cases it should be treating through the acceptor.. except to this large message case here.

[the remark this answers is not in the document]
tabish121: So are you saying a handler can now accept an additional message while it's trying to process the previously passed large message ?  That seem rife with errors. Otherwise I don't follow what you are saying.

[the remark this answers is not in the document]
clebertsuconic: this method only makes sense in the context of the AMQP Large Message. It doesn't make sense in any other writers (not yet at least).  

I've tried to make them all use this method, but the other writers were writing it asynchronously as I mentioned earlier.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
```diff
@@ -77,39 +74,68 @@ public AMQPLargeMessageReader open() {
...
          final long id = sessionSPI.getStorageManager().generateID();
-         currentMessage = new AMQPLargeMessage(id, delivery.getMessageFormat(), null,
-                                               sessionSPI.getCoreMessageObjectPools(),
-                                               sessionSPI.getStorageManager());
+         currentMessage = new AMQPLargeMessage(id, delivery.getMessageFormat(), null, sessionSPI.getCoreMessageObjectPools(), sessionSPI.getStorageManager());
          currentMessage.parseHeader(dataBuffer);
-
+         logger.trace("Initializing current message {} on {}", currentMessage, this);
          sessionSPI.getStorageManager().largeMessageCreated(id, currentMessage);
+         sessionSPI.execute(currentMessage::validateFile);
```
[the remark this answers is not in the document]
clebertsuconic: I was swallowing on purpose, as any subsequent reads would endue on an exception through addBytes calls (which is calling AMQPConnectionContext::exception


I'm adding the validation now just in case (amended the commit)

https://github.com/clebertsuconic/activemq-artemis/blob/1bae8665458e53a8c1793330147ee6b389e3b586/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java#L392-L400

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
```diff
@@ -77,39 +74,68 @@ public AMQPLargeMessageReader open() {
...
       }
 
-      currentMessage.addBytes(dataBuffer);
+      logger.trace("Disable autoread on {}", this);
```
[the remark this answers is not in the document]
clebertsuconic: @tabish121 I had a toString() method added when I was debugging this... let me think about it tomorrow if I should add the toString or remove this trace.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
```diff
@@ -77,39 +74,68 @@ public AMQPLargeMessageReader open() {
...
+      serverReceiver.getConnection().disableAutoRead();
 
-      final AMQPLargeMessage result;
+      byte[] bytes = new byte[dataBuffer.remaining()];
+      dataBuffer.get(bytes);
+      logger.trace("Getting {}", bytes.length);
```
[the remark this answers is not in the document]
clebertsuconic: I had one specific case I needed... it doesn't make sense to keep it now.. will remove it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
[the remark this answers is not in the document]
tabish121: I've made the change here locally and I don't see any reason it doesn't work to just make the open method accept the message reference and do the initialization there as opposed to adding a second "special" open method the select of the writer is done in the same thread as the sender will be calling open prior to handling the write into the connection thread.

artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPTunneledCoreLargeMessageReaderTest.java
```diff
@@ -159,19 +169,18 @@ public void testReadMessageByteByByteFromDeliveryBuffer() throws Exception {
...
       }
 
-      assertNotNull(reader.getDeliveryAnnotations());
-
-      final DeliveryAnnotations annotations = reader.getDeliveryAnnotations();
-
-      assertTrue(annotations.getValue().get(Symbol.valueOf("a")).equals("a"));
-      assertTrue(annotations.getValue().get(Symbol.valueOf("b")).equals("b"));
-      assertTrue(annotations.getValue().get(Symbol.valueOf("c")).equals("c"));
+      DeliveryAnnotations currentAnnotations = reader.getDeliveryAnnotations();
```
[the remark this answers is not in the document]
clebertsuconic: The readMethod now calls close when there's no more pending. that is the onMessageComplete is called and the reader is closed. Meaning the test here would call getDeliveryAnnotations and it wouldn't be here.


the only case where getDeliveryAnnotations will return anything is if the read it still partial now.

artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPTunneledCoreLargeMessageReaderTest.java
```diff
@@ -116,21 +128,19 @@ public void testReadDeliveryAnnotationsFromDeliveryBuffer() throws Exception {
...
 
-      assertNotNull(reader.getDeliveryAnnotations());
-
-      final DeliveryAnnotations annotations = reader.getDeliveryAnnotations();
-
-      assertTrue(annotations.getValue().get(Symbol.valueOf("a")).equals("a"));
-      assertTrue(annotations.getValue().get(Symbol.valueOf("b")).equals("b"));
-      assertTrue(annotations.getValue().get(Symbol.valueOf("c")).equals("c"));
+      // peek current delivery annotations
+      DeliveryAnnotations currentAnnotations = reader.getDeliveryAnnotations();
```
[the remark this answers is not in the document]
clebertsuconic: The readMethod now calls close when there's no more pending. that is the onMessageComplete is called and the reader is closed. Meaning the test here would call getDeliveryAnnotations and it wouldn't be here.

the only case where getDeliveryAnnotations will return anything is if the read it still partial now.

artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPTunneledCoreLargeMessageReaderTest.java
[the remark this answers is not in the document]
clebertsuconic: @tabish121 I remembered what happened.. I tried removing the open method by just the accept.. and I had made a mess back and forth until I stabilized on what you saw... 


thanks for the review.. I made the change.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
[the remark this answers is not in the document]
clebertsuconic: I made the change.

I had done some back and forth changes and I couldn't see it now.. thanks for the review.

tests/artemis-test-support/src/main/java/org/apache/activemq/transport/amqp/client/AmqpMessage.java
clebertsuconic: I just wanted to assert a few bytes[] that were set with setBytes([]).

I have no idea on what I'm supposed to do here. I will just revert the test addition.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
```diff
@@ -77,39 +73,73 @@ public AMQPLargeMessageReader open() {
...
+         logger.trace("Adding {} bytes on currentMessage={}, this={}", dataBuffer.remaining(), currentMessage, this);
+         currentMessage.addBytes(dataBuffer);
+
+         if (!isPartial) {
+            final AMQPLargeMessage message = currentMessage;
+            message.releaseResources(serverReceiver.getConnection().isLargeMessageSync(), true);
+            logger.trace("finishing {} on {}", currentMessage, this);
+            // We don't want a close to delete the file now, we've released the resources.
+            currentMessage = null;
+            close();
```
tabish121, line 133: This seems to be incorrect as closing here is wrong as a general rule as the resource that opened this should be closing it but also this is occurring in a different thread than the connection thread where the reader was opened which could lead to error on inconsistent state.  The receiver should be closing this in the connection thread.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
```diff
@@ -77,39 +73,73 @@ public AMQPLargeMessageReader open() {
...
-         final AMQPSessionCallback sessionSPI = serverReceiver.getSessionContext().getSessionSPI();
          final long id = sessionSPI.getStorageManager().generateID();
          currentMessage = new AMQPLargeMessage(id, delivery.getMessageFormat(), null,
                                                sessionSPI.getCoreMessageObjectPools(),
                                                sessionSPI.getStorageManager());
          currentMessage.parseHeader(dataBuffer);
-
+         logger.trace("Initializing current message {} on {}", currentMessage, this);
          sessionSPI.getStorageManager().largeMessageCreated(id, currentMessage);
+         sessionSPI.execute(() -> validateFile(currentMessage));
```
tabish121, line 96: This check seems to be mostly pointless now as the code doesn't use a continuation to resume delivery after validation it just throws it into another thread and then continues on as if it was successful

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPTunneledCoreMessageReader.java
```diff
@@ -155,6 +149,10 @@ public Message readBytes(Delivery delivery) {
       coreMessage.reloadPersistence(buffer, sessionSPI.getCoreMessageObjectPools());
       coreMessage.setMessageID(sessionSPI.getStorageManager().generateID());
 
-      return coreMessage;
+      serverReceiver.onMessageComplete(delivery, coreMessage, deliveryAnnotations);
+
+      close();
```
tabish121, line 154: The receiver should be closing this in onMessageComplete, a reader shouldn't close itself.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
clebertsuconic: It still needs to create the file. which should happen away from the Netty Thread. 

in case of an exception it will initialize the exception on the session... The current writes are still going through though but the exception would be initialized. I could interrupt the addBytes if there was an exception, but the possibility of having an exception there is actually so low that if any issues are happening with the storage, other things will happen to put the server down anyways through IOExceptions.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPTunneledCoreMessageReader.java
clebertsuconic: I changed the contract there. so the thing is closing itself before calling the onComplete.

I was going to use that to release the resources.. but it's not needed any more.

tabish121: I'm saying its wrong and I'm saying its thread unsafe.  And if I need to I'm saying -1 until I'm sure it is

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
```diff
@@ -77,39 +73,73 @@ public AMQPLargeMessageReader open() {
...
+            logger.trace("finishing {} on {}", currentMessage, this);
+            // We don't want a close to delete the file now, we've released the resources.
+            currentMessage = null;
+            close();
+            serverReceiver.connection.runNow(() -> serverReceiver.onMessageComplete(delivery, message, message.getDeliveryAnnotations()));
+         }
+      } catch (Throwable e) {
+         logger.warn(e.getMessage(), e);
+         close();
+         serverReceiver.connection.exception(e);
```
tabish121, line 139: The addBytes is not running on the connection thread so this call will end up operating on the proton bits outside the connection thread leading to possible inconsistent state when handling this.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPTunneledCoreLargeMessageReader.java
```diff
@@ -209,13 +208,15 @@ public Message readBytes(Delivery delivery) throws Exception {
 
             state = State.DONE;
 
-            return result;
-         }
+            serverReceiver.onMessageComplete(delivery, result, deliveryAnnotations);
 
-         return null;
+            close();
```
gemmellr, line 213: Doesnt feel like a read should be closing itself. Was meant to be managed by the interested party using it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPTunneledCoreMessageReader.java
gemmellr: ..commented the same on another file before I got to this one :)

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/MessageReader.java
```diff
@@ -51,16 +49,7 @@ public interface MessageReader {
     *
     * @param delivery
     *    The delivery that has pending incoming bytes.
+    * @return true if the delivery was partial and more reads are required to complete the operation
```
gemmellr, line 52: Rest of the method javadoc needs updating too, its now totally wrong

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/LargeServerMessageImpl.java
```diff
@@ -363,13 +363,17 @@ public String toString() {
...
 
-   public void ensureFileExists(boolean toOpen) throws ActiveMQException {
+   public void ensureFileExists(boolean toOpen) {
       synchronized (largeBody) {
-         largeBody.ensureFileExists(toOpen);
+         try {
+            largeBody.ensureFileExists(toOpen);
+         } catch (Exception e) {
+            logger.debug(e.getMessage(), e);
+         }
```
gemmellr, line 376: Seems odd to swallow this...so ensureFileExists() no longer ensures the file exists? Meaning validateFile() isnt validating much?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageWriter.java
```diff
@@ -81,33 +84,54 @@ public boolean isWriting() {
...
       }
    }
 
    @Override
-   public AMQPLargeMessageWriter open() {
-      if (!closed) {
-         throw new IllegalStateException("Trying to open an AMQP Large Message writer that was not closed");
+   public AMQPLargeMessageWriter open(MessageReference reference) {
+      this.reference = reference;
+      this.message = (AMQPLargeMessage) reference.getMessage();
```
tabish121, line 107: This is storing the message which means a call to close() will call usageDown on the reference but the way the code is now structured there isn't a 100% certainly that usageUp will always be called on the reference.  It needs to be called here likely so that a close will indeed not accidentally lower usage when it hadn't been increased.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
clebertsuconic: I added a commit to be squashed addressing this

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPTunneledCoreMessageReader.java
clebertsuconic: I added a commit to be squashed addressing this

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/NettyConnection.java
```diff
@@ -126,6 +126,9 @@ public final Channel getNettyChannel() {
    @Override
    public final void setAutoRead(boolean autoRead) {
       channel.config().setAutoRead(autoRead);
+      if (autoRead) {
+         channel.read();
+      }
```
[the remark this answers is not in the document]
clebertsuconic: I added a commit to be squashed addressing this

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -378,6 +389,16 @@ public AMQPConnectionCallback getConnectionCallback() {
...
 
+   public void exception(Throwable e) {
+      logger.warn(e.getMessage(), e);
+      ErrorCondition error = new ErrorCondition();
+      error.setCondition(AmqpError.INTERNAL_ERROR);
+      error.setDescription("Unrecoverable error: " + (e.getMessage() == null ? e.getClass().getSimpleName() : e.getMessage()));
+      getHandler().getConnection().setCondition(error);
+      getHandler().getConnection().close();
+      flush();
+   }
```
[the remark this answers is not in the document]
clebertsuconic: I added a commit to be squashed addressing this

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPConnectionContext.java
```diff
@@ -94,6 +94,17 @@ public class AMQPConnectionContext extends ProtonInitializable implements EventH
...
    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
+   public void disableAutoRead() {
+      connectionCallback.getTransportConnection().setAutoRead(false);
+      handler.setReadable(false);
+   }
+
+   public void enableAutoRead() {
+      connectionCallback.getTransportConnection().setAutoRead(true);
+      getHandler().setReadable(true);
```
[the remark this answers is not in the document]
clebertsuconic: enableAutoRead is now called from the Netty Thread with a runNow call 


this was going to call flush which would transfer the execution anyways.


my commit to be squashed is addressing this

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageWriter.java
```diff
@@ -144,21 +159,28 @@ private void resume() {
...
    private void tryDelivering() {
       // This is discounting some bytes due to Transfer payload
-      final int frameSize = protonSender.getSession().getConnection().getTransport().getOutboundFrameSizeLimit() - 50 - (delivery.getTag() != null ? delivery.getTag().length : 0);
 
       try {
+         if (this.delivery == null) {
+            this.delivery = serverSender.createDelivery(reference, (int) this.message.getMessageFormat());
+            if (sessionSPI.invokeOutgoing(message, (ActiveMQProtonRemotingConnection) sessionSPI.getTransportConnection().getProtocolConnection()) != null) {
+               return;
+            }
```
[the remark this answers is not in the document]
clebertsuconic: will work on it..

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPTunneledCoreLargeMessageReader.java
clebertsuconic: to be squashed commit handling it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageWriter.java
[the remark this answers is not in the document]
clebertsuconic: commit to be squashed 55ce243a690a1dcae26587ea6e962effb896f039 is addressing this

clebertsuconic:
@gemmellr , @tabish121 I added 3 temporary commits to this PR. Let me know if you have any questions and I will squash them before setting it ready to merge.

clebertsuconic:
@tabish121 , @gemmellr  I tried simplifying the changes as much as possible here...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/ProtonHandler.java
```diff
@@ -213,9 +224,7 @@ public boolean isHandler() {
    }
 
    public void requireHandler() {
-      if (!workerExecutor.inEventLoop()) {
-         throw new IllegalStateException("this method requires to be called within the handler, use the executor");
-      }
+      assert workerExecutor.inEventLoop();
```
gemmellr, line 227: Is the trade-off worth it? If it goes wrong in live usage then previously we would always find out. Now we probably wont, except any weird consequences of illegal use.

Also feels like a separate change either way.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageWriter.java
```diff
@@ -81,33 +84,55 @@ public boolean isWriting() {
...
-            reset(true);
+            resetClosed();
          }
       }
    }
 
    @Override
-   public AMQPLargeMessageWriter open() {
-      if (!closed) {
-         throw new IllegalStateException("Trying to open an AMQP Large Message writer that was not closed");
```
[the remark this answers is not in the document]
clebertsuconic: it was an honest mistake.. I didn't mean to.. will place it back.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonAbstractReceiver.java
```diff
@@ -320,28 +325,45 @@ public final void onMessage(Delivery delivery) throws ActiveMQAMQPException {
             return;
          }
 
-         final Message message = messageReader.readBytes(delivery);
+         {
+            Message completeMessage;
+            if ((completeMessage = messageReader.readBytes(delivery)) != null) {
+               // notice the AMQP Large Message Reader will always return false
+               // and call the onMessageComplete directly
+               // since that happens asynchronously
```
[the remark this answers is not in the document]
clebertsuconic: thanks

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonAbstractReceiver.java
```diff
@@ -320,28 +325,45 @@ public final void onMessage(Delivery delivery) throws ActiveMQAMQPException {
             return;
          }
 
-         final Message message = messageReader.readBytes(delivery);
+         {
```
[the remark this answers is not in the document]
clebertsuconic: I added it to make explicit only.. removing it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonAbstractReceiver.java
```diff
@@ -320,28 +325,45 @@ public final void onMessage(Delivery delivery) throws ActiveMQAMQPException {
...
-            // delivery annotations.
-            final DeliveryAnnotations deliveryAnnotations = messageReader.getDeliveryAnnotations();
+   public void onMessageComplete(Delivery delivery,
+                          Message message, DeliveryAnnotations deliveryAnnotations) {
+      connection.requireInHandler();
 
-            this.messageReader.close();
-            this.messageReader = null;
+      try {
+         final Receiver receiver = ((Receiver) delivery.getLink());
```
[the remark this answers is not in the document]
clebertsuconic: @tabish121 I am just moving code from here:

https://github.com/apache/activemq-artemis/blob/93a74dc00c5885c238e59e8b612aec47c657f541/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonAbstractReceiver.java#L305


I assumed it was needed.. and I changed as minimal as possible.


Should I use the receiver already stored instead?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageWriter.java
```diff
@@ -81,33 +84,55 @@ public boolean isWriting() {
...
          } finally {
-            reset(true);
+            resetClosed();
          }
       }
    }
 
    @Override
-   public AMQPLargeMessageWriter open() {
-      if (!closed) {
```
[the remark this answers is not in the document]
clebertsuconic: was an accident I placed it back already

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/handler/ProtonHandler.java
clebertsuconic: I placed it back...

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonAbstractReceiver.java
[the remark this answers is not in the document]
tabish121: I realized it afterwards, it isn't really needed as it shouldn't ever be a receiver that isn't the one we have tracked but guess its fine either way, removing unneeded code is usually a nice thing.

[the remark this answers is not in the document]
clebertsuconic: @tabish121 sure.. will remove it

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
```diff
@@ -34,7 +34,7 @@ public class AMQPLargeMessageReader implements MessageReader {
 
    private final ProtonAbstractReceiver serverReceiver;
 
-   private AMQPLargeMessage currentMessage;
+   private volatile AMQPLargeMessage currentMessage;
```
clebertsuconic, line 37: @tabish121 , @gemmellr currentMessage is set to null as part of the Session Executor. 
We shouldn't have multiple threads messing with the variable at the same time as we set readable to false and transfer the execution, but I think it would be safer to set it as volatile.


Let me know if you have a concern here.

tabish121: I'd go with abundance of caution since we are crossing threads where we weren't before.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
```diff
@@ -50,14 +50,15 @@ public DeliveryAnnotations getDeliveryAnnotations() {
...
       if (!closed) {
-         if (currentMessage != null) {
-            try {
-               currentMessage.deleteFile();
-            } catch (Throwable error) {
-               ActiveMQServerLogger.LOGGER.errorDeletingLargeMessageFile(error);
-            } finally {
-               currentMessage = null;
+         try {
+            AMQPLargeMessage localCurrentMessage = currentMessage;
```
gemmellr, line 54: There are various other usages of currentMessage that could probably benefit from similar update.

clebertsuconic: There was actually only addBytes... as the initialization should stay as is.

I amended the PR with the addBytes change.

clebertsuconic: @gemmellr actually... I am doing another change instead.


The AMQPLargeMessage should be created within the session executor.

Right now that's only working because I'm not actually creating a file there yet.. but this is a bit fragile.. as anyone could break it..


I will move the initialization of the new Object within the addBytes..

gemmellr: > There was actually only addBytes... as the initialization should stay as is.
> 
> I amended the PR with the addBytes change.

readBytes also reads (and writes) the variable multiple times.

clebertsuconic: not any more after my push -f

PR #4841, opened 2024-03-05
ARTEMIS-4669 Clarify Storage Manager usage around large messages

clebertsuconic:


clebertsuconic:
this is a simple renaming of methods... 


if anyone has any concerns about these names I used.. please send a  new commit...

all I cared here is to not have createLargeMessage and largeMessageCreated on the API

(which is something I did in the past.. but it was way confusing for me :) )


I am merging as my next PR is based no this.

PR #4843, opened 2024-03-06
ARTEMIS-4672 - Upgrade Netty to 4.107.Final

ehsavoie:


gemmellr:
Should probably also update the netty-tcnative test dep.

PR #4844, opened 2024-03-07
ARTEMIS-4675 - Add replication status metrics

alvinkwekel:


alvinkwekel:
Hi @jbertram, thanks for your response. 

I've added all cluster/replication related metrics that could to help with debugging later. Looking at the activation sequence over time would give an indication on restarts which could help to detect issues. Although I'm also not too sure if it will fit in well with the other metrics.

Are there additional metrics you think that we should add for the replication monitor use case? The "synchronizing" is not working as I expected. What would be the way to get insight in the current state of synchronization?

jbertram:
Generally speaking, exported metrics are not meant for debugging, per se. Exported metrics are really designed for runtime monitoring which is why we don't export everything but only the stuff that really matters for understanding how the broker is functioning in real time. This keeps the metrics' footprint relatively low and performance relatively high.

Metrics are often consumed by external systems that are not only used to visualize the data but are also used to alert on specific conditions. With that in mind, metrics should be something that can either be meaningfully visualized over time (i.e. the size of a queue) or whose state might indicate a problem (e.g. a passive backup broker that becomes active might trigger an alert).

It's not clear to me where the activation sequence fits in to this framework. It seems like it would be better to log activation sequence changes and if you need to debug something then you can take a look at the log. At this point, I don't see how it would be specifically useful to expose this as a metric. Do you have a particular use-case in mind?

Regarding the "synchronizing" metric you added, please elaborate on how it's functioning now and how you expect it to function.

Also, it's worth noting that the "synchronizing" and "replica.sync" metrics are looking at the _same_ underlying data so one is redundant and should be removed.

alvinkwekel:
Your feedback is much appreciated. 

I absolutely follow your reasoning on the purpose of metrics. I'll remove the activation sequence metric as its has no direct purpose in our current use case.

Currently we check the replica sync with a jolokia query before we restart the servers for maintenance. When its not synced we halt the process and manually correct. This happend a few times. With these metrics we want to detect these issues sooner, so we can correct them when they happen.

With the current state of this PR the synchronizing is returning 0 and the replica sync is returning 1. Based on the description of the replica sync I understand this is the _initial_ replication synchronization. I was hoping to get more insight on the running/current synchronization with the synchronizing metric. But if its basically returning the same underlying data I will remove the "synchronizing" metric. 

Could you elaborate, is the replica sync indeed fixed to either true or false after the initial synchronization of will it reflect the current state? Or can we use an other method to get the current replication state - after the initial synchronization?

jbertram:
> With the current state of this PR the synchronizing is returning 0 and the replica sync is returning 1.

I see the replica sync metric, but I don't see the "synchronizing" metric you're referring to. Can you clarify which metrics you're inspecting?

The replica sync metric indicates whether the initial synchronization has completed. Once the initial synchronization is done the primary is guaranteed to be in sync with the backup unless the replication connection between the brokers fails for some reason. It's not clear to me at this point if the replication manager actually exposes metrics to track the current status properly.

jbertram:
@alvinkwekel, I believe it would. I'm running the PR checks on your branch. If they pass I'll merge this.

PR #4847, opened 2024-03-08
ARTEMIS-4678 JDBC User and password not used by CLI

clebertsuconic:


clebertsuconic:
not sure how to test this..

clebertsuconic:
I will write a test before merging this...

Although the only thing I will be able to do is to verify the DBOptions gets the user and password on the properties.

Otherwise I would need a Database configured just for the test itself. which gets a bit fragile.

PR #4849, opened 2024-03-11
ARTEMIS-4682 Improving performance of reloading JDBC storage

clebertsuconic:
- fixing leaks
- page.exists() would lead to a select(all) and find.


clebertsuconic:
this is a simple bug fix... I'm merging it... we can always review it further after merged.
if anyone sees any issue please ping me in any way you can :)

PR #4852, opened 2024-03-13
ARTEMIS-4687 concurrent use of DocumentBuilder breaking Xpath filtering

jbertram:


artemis-selector/src/main/java/org/apache/activemq/artemis/selector/filter/JAXPXPathEvaluator.java
```diff
@@ -56,8 +60,11 @@ protected boolean evaluate(String text) {
 
    protected boolean evaluate(InputSource inputSource) {
       try {
-         return ((Boolean)xpath.evaluate(xpathExpression, builder.parse(inputSource), XPathConstants.BOOLEAN)).booleanValue();
+         synchronized (builder) {
```
clebertsuconic, line 63: shouldn't there be multiple instances created instead?  I mean... it seems the factory always returns the same instance?

jbertram: I'm not exactly sure if there _should_ be multiple instances. The `builder` is created in a `static` block in `org.apache.activemq.artemis.selector.filter.XPathExpression`.

PR #4855, opened 2024-03-14
ARTEMIS-4689 import command should accept URL

jbertram:


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/xml/XmlDataImporter.java
```diff
@@ -90,10 +91,12 @@ public final class XmlDataImporter extends ActionAbstract {
    private ClientSession session;
    private ClientProducer producer;
 
-   @Option(names = "--host", description = "The host used to import the data. Default: localhost.")
-   public String host = "localhost";
+   @Deprecated(forRemoval = true)
+   @Option(names = "--host", description = "The host used to import the data. Default: null. Deprecated! Use 'url' instead.")
```
clebertsuconic, line 95: just use hidden=true, no need to deprecate things in the CLI.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/xml/XmlDataImporter.java
```diff
@@ -90,10 +91,12 @@ public final class XmlDataImporter extends ActionAbstract {
...
+   public String host = null;
 
-   @Option(names = "--port", description = "The port used to import the data. Default: 61616.")
+   @Deprecated(forRemoval = true)
+   @Option(names = "--port", description = "The port used to import the data. Default: 61616. Deprecated! Use 'url' instead.")
```
clebertsuconic, line 99: just use hidden

clebertsuconic: eg: https://github.com/apache/activemq-artemis/blob/7742936583d48e179f67c7d73b7b49cbe83731d5/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/DestAbstract.java#L41-L42


if you have a script showing the older parameters, it should work fine, but it will prevent users from discovering the deprecated term.


it is the best approach on the CLI.. hidden=true

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/xml/XmlDataImporter.java
jbertram: I added `hidden = true`, but I left `@Deprecated` so that we will remember to remove it at the next major version.

clebertsuconic:
hmmmm... maybe we need to tweak the Abstract a bit.


does it make sense to specify --protocol on the import?

what about clientID?

```
Usage: artemis data imp [--legacy-prefixes] [--silent] [--sort] [--verbose]
                        [--acceptor=<acceptor>] [--clientID=<clientID>]
                        --input=<input> [--password=<password>]
                        [--protocol=<protocol>] [--url=<brokerURL>]
                        [--user=<user>]
Import all message-data using an XML that could be interpreted by any system.
      --acceptor=<acceptor>
                          Name used to find the default connection URL on the
                            acceptor list. If an acceptor with that name cannot
                            be found the CLI will look for a connector with the
                            same name.
      --clientID=<clientID>
                          ClientID set on the connection.
      --input=<input>     The input file name. Default: exp.dmp.
      --legacy-prefixes   Do not remove prefixes from legacy imports.
      --password=<password>
                          Password used to connect.
      --protocol=<protocol>
                          Protocol used. Valid values are AMQP, CORE
      --silent            Disable all the inputs, and make a best guess for any
                            required input.
      --sort              Sort the messages from the input (used for older
                            versions that won't sort messages).
      --url=<brokerURL>   Connection URL. Default: build URL from the 'artemis'
                            acceptor defined in the broker.xml or tcp:
                            //localhost:61616 if the acceptor cannot be parsed.
      --user=<user>       User used to connect.
      --verbose           Print additional information.

```

jbertram:
No, it doesn't really make sense to specify `protocol` or `clientID`, but it doesn't really hurt either as far as I can tell. It would be worth adjusting the abstract to have one that's more generic for basic connectivity.

clebertsuconic:
@jbertram I think it gets confusing. I think we should have an upper class that does not ask for those. 

This is actually probably the reason why it didn't extend the class and made a copy of the properties instead of reusing stuff.

jbertram:
@clebertsuconic, I separated out the basic connection stuff. Let me know what you think.

PR #4856, opened 2024-03-15
ARTEMIS-4498 - Expose internal queues for management and observability

clebertsuconic:


clebertsuconic:
@AntonRoskvist this is your PR. I simplified it by removing the isManaged.. 

I have been having issues myself when I was in phone calls with customers.. and I strongly believe we should keep everything manageable. The only intent for the internal queue was about redistribution and load balancing... 

Having those managed should always be preferable. and I have a strong opinion about this since it bit myself in production.  (Just a heads up in case anyone wants to argue that :) )... (kind of half joking here, but It's true).

I want to have this merged before the release on monday.

clebertsuconic:
Hmmm... I think I figured out what's going on.

When we implemented openWire advisors, we used an internal queue that was meant to be hidden.

I'm not sure it will be sanitary to have open wire connections to a server and showing all the advisors on the admin console.


It's different though from SNFs from Mirroring and Clustering.



Perhaps we would need a separation from those, but I would rather add an attribute for hidden=true, rather than manageable. And have the code somewhat simpler.


I will leave this PR open for now and not merge it before the release. Perhaps I will wait the new admin console merged before moving into this since we will probably want to show the advisors in certain cases.

AntonRoskvist:
@clebertsuconic Perhaps a simpler interim solution could be used in the mean-time, doing something like what I've added here:
https://github.com/AntonRoskvist/activemq-artemis/tree/interim_ARTEMIS-4498

If you'd rather wait for the proper fix then that's fine too of course.

clebertsuconic:
I don't understand your interim solution? you're forcing the registration if address =management address?



I think everything should be manageable. (I have been bitten by that before)


you may want to hidden certain address from the console, that could reappear after a toggle is selected.



(maybe we hide internal addresses?)


functionally I think we should have a flag to hide certain control queues only... (that I don't have a name for them yet).


MQTT and OpenWire protocol managers have a few... even those should be able to reappear after a select toggle on the console.

AntonRoskvist:
@clebertsuconic I agree that everything should be manageable, except perhaps for some protocol specific queues as you stated.

Since the property "internal" is used to filter weather or not an address or queue should get exposed for management or not ( as of now ) and said protocol managers rely on this to filter Advisories for instance, a new property might be needed to decide if the destination should be manageable or not, instead of the current "internal" that has other uses, besides deciding manageability.

Since that change would take some more work to implement it is out of scope for this release, which I 100% agree with.

My interim solution/idea/example is to leave all this logic as is, with the only change being to manually register management for addresses and queues under "Notifications, SnF and AMQPMirror(which already ran this for the queues)". i.e. get management/metrics for the most critical parts for the upcoming release. This would later be replaced by the aforementioned "proper" change. 

(I don't like doing things that way but I'm weighing it against the usefulness of these destinations metrics I thought I'd ask at least)

clebertsuconic:
@AntonRoskvist I would rather delay this fix and do a proper fix in the next release.


I'm looking to fix this before the next release, which may be a shorter release (timewise on when we are releasing it).

clebertsuconic:
@AntonRoskvist , @andytaylor can we settle with these changes?


I added the "internal attribute" to the AddressControl as well.

@andytaylor if you could take a look at the changes made to the web console here please?

AntonRoskvist:
@clebertsuconic this doesn't quite do it for me as the change still leaves out management and metrics for the internal queues. In my use case I'm primarily looking for metrics on the clusters SnF queues as well as the notification queues.

Perhaps the servers `createQueue` could get overloaded with a variant that inhibits creation of management objects that can be used for the relevant MQTT and OpenWire destinations?

clebertsuconic:
As far as I know I changed the SNF to always register management.  What am I missing ?

clebertsuconic:
I should be basically always registering any queue. Did I miss that ?

clebertsuconic:
@AntonRoskvist we have management enabled on all the queues:

basically your test is still here:

https://github.com/clebertsuconic/activemq-artemis/blob/56e8fbb50934cb8c451e2ae5b764df64da06481f/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/SimpleSymmetricClusterTest.java#L369-L416


what I did was to remove all the extra code around specifying types, and I'm just making everything manageable.


if you're missing any parts, please let me know.

clebertsuconic:
(I closed it by accident)


@AntonRoskvist if you could take a look please? management on SNF still enabled per your tests.

AntonRoskvist:
@clebertsuconic Yes the test is still passing so I suppose they are manageable, however at least from my own testing I have been unable to collect any metrics for both the SnF queues and the notifications queues. Removing the check for `isInternal()` in `org.apache.activemq.artemis.core.server.management.impl.ManagementServiceImpl#registerQueue()` seems to fix that though but then I guess it will also display all of the OpenWire Advisories as well.

I tried the web console on this PR as well and in there I can see that the filters added for `internal` are working but looking at the tree I see that though addresses are visible for these destinations, their queues are not.

PR #4857, opened 2024-03-18
ARTEMIS-4691 log.warn on inVM closing

clebertsuconic:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/SimpleTest.java
jbertram, line 1: This should be in a new test rather than modifying `SimpleTest`.

clebertsuconic: the test still simple ;)

clebertsuconic: I don't see any reference from the doc as the javadoc in the test states.

jbertram: As the JavaDoc states, "A simple test-case used for documentation purposes." This test is referenced in the Hacking Guide as a guide to help folks get started writing tests. Adding anything superfluous to basic testing is going to diminish the effectiveness of the test as a generic example.

PR #4861, opened 2024-03-20
ARTEMIS-4696 Deprecate queue prefetch address setting

brusdev:


jbertram:
LGTM

brusdev:
I updated the PR to only add deprecated attributes

PR #4867, opened 2024-03-28
ARTEMIS-4510 - Add auto-create-destination logic to diverts

AntonRoskvist:
This is the second try of #4681 after a round of feedback on the dev mailing list.
@jbertram @brusdev @clebertsuconic Please let me know if you agree to these changes or if I should change anything else.

An additional thing of note in this PR is added handling for a fringe issue I found while working on this PR pertaining to redistributed messages arriving on the divert. There is an added test for this as well. I can break this part out into a separate PR if preferred.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/config/ActiveMQDefaultConfiguration.java
@@ -698,6 +698,8 @@ public final class ActiveMQDefaultConfiguration {
 
    private static final boolean DEFAULT_MANAGEMENT_MESSAGE_RBAC = false;
 
+   private static final boolean DEFAULT_DIVERT_REUSE_USER_SESSION = false;
+
    /**
     * If true then the ActiveMQ Artemis Server will make use of any Protocol Managers that are in available on the classpath. If false then only the core protocol will be available, unless in Embedded mode where users can inject their own Protocol Managers.
     */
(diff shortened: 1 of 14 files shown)


clebertsuconic:
@AntonRoskvist I'm thinking we should just always reuse the session... meaning I think that's the missing link :)

clebertsuconic:
I think the remove attributes from cluster becomes a separate issue indeed.

i don't mind about a separate PR.. but definitely a separate commit.


Can you separate that change?


but let me do some testing with removing the configuration options and just always reusing the session.

AntonRoskvist:
@clebertsuconic Thanks for your feedback!
I agree, however that means potentially breaking some use-case currently depending on those messages getting "discarded" . I feel a note in the upcoming versions doc should suffice but just making sure everyone's in agreement about this.

I will separate the changes as well. I will probably send that by Monday

clebertsuconic:
@AntonRoskvist some time ago I had already made routing able to auto-create destinations as long as the session from the client is used.

Since you are passing the session now, I think we should just use that functionality as is... We wouldn't break anythiing.


This is different than having Divert performing the auto-creation.  this was the missing link.. it's a lot cleaner!!!

AntonRoskvist:
@clebertsuconic Sounds great, thanks!

I went ahead and separated the changes.

Let me know when/if you are happy with your testing (removing the config option and always reusing the session) and I'll go ahead and remove the config option as well as some documentation from this PR.

clebertsuconic:
@AntonRoskvist I pulled your changes, removed all the docs and stuff and pushed on main (I pushed as you).

now you have to give a JIRA to your other change.

it's probably easier if you cherry-pick that issue on main than rebasing as you now have conflicts. I don't mind if you keep this very same PR but you will need to push -f with just that change now.


if you can't do it I will do those things myself.. let me know if you want / can do it.


Thank you.

clebertsuconic:
actually.. I pushed a commit

AntonRoskvist:
@clebertsuconic Excellent, thanks for your fixes!

PR #4870, opened 2024-04-02
ARTEMIS-4710 Keep pem-keystore dependencies as is; make it really optional

amarkevich:
The latest version already uses bcprov-jdk18on


gemmellr:
PR clearly needs a JIRA. Particularly as it seems to have an unrelated code change to boot. Is that so you can exclude the deps if not using the functionality? I'd guess thats how you even noticed the earlier required exclusions+additions that lead to the other changes.

PR #4871, opened 2024-04-02
ARTEMIS-4709 plugin to enforce connection periodic expiry per acceptor

gtully:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/plugin/impl/ConnectionPeriodicExpiryPlugin.java
```diff
@@ -0,0 +1,130 @@
...
+
+         final long currentTime = System.currentTimeMillis();
+         for (Acceptor acceptor : remotingService.getAcceptors().values()) {
+            if (matchPattern.matcher(acceptor.getName()).matches()) {
+               if (acceptor instanceof NettyAcceptor) {
+                  NettyAcceptor nettyAcceptor = (NettyAcceptor) acceptor;
+
+                  for (NettyServerConnection nettyServerConnection : nettyAcceptor.getConnections().values()) {
+                     RemotingConnection remotingConnection  = remotingService.getConnection(nettyServerConnection.getID());
+                     if  (currentTime > remotingConnection.getCreationTime() + periodSeconds ) {
```
gemmellr, line 70: Seems like it needs a validity check, connection could have already gone away by other means while this checking was happening, so perhaps it could NPE here.

Related...as is, if this task ever throws for any reason (e.g above potential NPE) the expiry process will then simply stop working as the task wont be rescheduled.

docs/user-manual/broker-plugins.adoc
```diff
@@ -178,3 +178,30 @@ In the example below `ROLE_PROPERTY` is set to `permissions` when that property
...
+
+This plugin can be useful when credential rotation or credential validation must be enforced at regular intervals as authentication will be enforced on reconnect.
+
+The plugin requires the configuration of the `acceptorMatchRegex` to determine the acceptors to monitor. It is typical to separate client acceptors and federation or cluster acceptors such that only client connections will be subject to periodic expiry. The `acceptorMatchRegex` must be configured to match the name of the acceptor(s) whose connections will be subject to periodic expiry.
+
+|===
+| Property | Property Description | Default Value
+
+|`acceptorMatchRegex`|the regular expression used to match against the names of acceptors to monitor | ""
+|`periodSeconds`|the max duration or period, in seconds, that a connection can last | 15 minutes (as seconds)
```
gemmellr, line 194: the default would be more obvious if offered in the way it is to actually be used, e.g:  "900 (i.e 15 minutes)"

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/plugin/impl/ConnectionPeriodicExpiryPlugin.java
```diff
@@ -0,0 +1,130 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
```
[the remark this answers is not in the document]
gtully: thanks

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/plugin/impl/ConnectionPeriodicExpiryPlugin.java
gtully: good catch, thanks. fixed and some more tests added

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/plugin/impl/ConnectionPeriodicExpiryPlugin.java
```diff
@@ -0,0 +1,130 @@
...
+
+   @Override
+   public void init(Map<String, String> properties) {
+      name = properties.getOrDefault("name", name);
+      periodSeconds = Long.parseLong(properties.getOrDefault("periodSeconds", Long.toString(periodSeconds)));
+      accuracyWindowSeconds = Integer.parseInt(properties.getOrDefault("accuracyWindowSeconds", Long.toString(accuracyWindowSeconds)));
+      if (accuracyWindowSeconds <= 0) {
+         throw new IllegalArgumentException("accuracyWindowSeconds must be > 0");
+      }
+      acceptorMatchRegex = properties.getOrDefault("acceptorMatchRegex", acceptorMatchRegex);
```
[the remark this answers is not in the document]
gtully: The value must be configured so a non null default does not make sense. Have added a check and throw if it is null. Is it ok to configure something that won't match, I guess so.

docs/user-manual/broker-plugins.adoc
gtully: fair point, thanks!

jbertram:
Is there a particular reason this needs to be a plugin vs. just a normal bit of functionality in the core server? Plugins are nice for some things, but they're kind of clunky to configure. This seems like something that would fit well on an `acceptor` URL.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/plugin/impl/ConnectionPeriodicExpiryPlugin.java
```diff
@@ -0,0 +1,150 @@
...
+   @Override
+   public void registered(ActiveMQServer server) {
+
+      sanityCheckConfig();
+
+      executor = server.getScheduledPool();
+      remotingService = server.getRemotingService();
+      matchPattern = Pattern.compile(acceptorMatchRegex);
+
+      task = executor.scheduleWithFixedDelay(() -> {
```
clebertsuconic, line 66: Optional: I have been using ActiveMQScheduledComponent to all the scheduled services we have. it supports having a separate Executor for the task itself.

gtully: I don't know if it is necessary or better, but consistency is good. I was trying to keep it as simple as possible, the task should be fast to complete, and it delegates a new task for the actual disconnect so I did not thing it warranted a separate executor.

jbertram:
I've gone back and forth on this in my mind a few times. 

I see your points, but if this functionality "becomes part of lots of deployments" then changing the way it's configured will be a breaking change for users which won't be a good experience. If the configuration is part of the core schema then they can get auto-completion (depending on the config editor they use) and validation.

The code will be shipped with the broker either way. Whether it is optional can depend on configuration regardless of it it is a plugin.

Generally speaking, I see plugins as a way for _users_ to add functionality to the broker.

Personally I think this makes sense in the core broker.

gtully:
on the config change, if a similar feature became available in the core, it would never be a breaking change, it would require a config change, but not a break. new config would be needed and the plugin removed. 

But more generally, on plugins for "users", I think it is an extension point, but we are all users!

jbertram (approved):

PR #4875, opened 2024-04-04
ARTEMIS-4714 mitigate NPE in FederatedQueueConsumerImpl MessageListener

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/federation/FederatedQueueConsumerImpl.java
```diff
@@ -325,7 +325,9 @@ public void onMessage(ClientMessage clientMessage) {
       } catch (Exception e) {
          ActiveMQServerLogger.LOGGER.federationDispatchError(clientMessage.toString(), e);
          try {
-            clientSession.rollback();
+            if (clientSession != null) {
+               clientSession.rollback();
```
clebertsuconic, line 329: in theory the NPE here could still happen.

you check if clientSession != null

a connectionFailed happened on a different thread.


you call rollback on the current thread.

Bang!!!! NPE!!!



Of course the JDK could be using non volatile references.. etc...


But it would be better if you cached the reference locally


Such as 

```ClientSession localSession = clientSesssion;
if (localSession != null) {
    localSession.rollback();
}
```


I agree the probably of the NPE happening is lower with your changes.. but they can still happen.

jbertram: I had considered that, but I thought the only way to address it was with synchronization. Your solution is a good one. I implemented it.

PR #4876, opened 2024-04-05
ARTEMIS-4666 Correctly set queue match when parsing XML confing

tabish121:
Correct the XML parser for core federation queue match policy loading to call the setQueueMatch instead of setAddressMatch when reating the queue match element.


tabish121:
I've run the full test suite following this change

clebertsuconic:
commit message says when "reating" queue element.. I think there's a typo in that

clebertsuconic:
LGTM.. only thing I see is the typo on the commit message.

Feel free to merge yourself (if up to me) after you amend the commit message.

PR #4880, opened 2024-04-11
ARTEMIS-4723 Avoid objects left on ThreadLocal from OrderedExecutorFactory

clebertsuconic:
co-authored: Clebert Suconic

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/HandlerBase.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/HandlerBase.java
@@ -26,22 +26,39 @@ package org.apache.activemq.artemis.utils.actors;
 public abstract class HandlerBase {
 
    private static class Counter {
-      int count = 0;
+      int count = 1;
    }
 
    /** an actor could be used within an OrderedExecutor. So we need this counter to decide if there's a Handler anywhere in the stack trace */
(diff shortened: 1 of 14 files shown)


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/actors/ProcessorBase.java
```diff
@@ -117,13 +112,6 @@ public int shutdownNow(Consumer<? super T> onPendingItem, int timeout, TimeUnit
       requestedShutdown = true;
       yielded = false;
 
-      if (!inHandler()) {
-         // We don't have an option where we could do an immediate timeout
-         // I just need to make one roundtrip to make sure there's no pending tasks on the loop
-         // for that I ellected one second
-         flush(timeout, unit);
```
clebertsuconic, line 124: This was for tests and it was still causing issues anyway...


so, time to go !!! more issues than benefits

clebertsuconic:
@cardamon notice I'm keeping you as the author on the commit. as you discovered the issue.


I went further and just removed the usage. if all my tests are good I will merge this.

cardamon:
> @cardamon notice I'm keeping you as the author on the commit. as you discovered the issue.
> 
> I went further and just removed the usage. if all my tests are good I will merge this.

Thanks!

clebertsuconic:
even thought I wanted to get rid of the shutdown.. that represented some semantic change.. and it's impossible to measure consequences. I will have to keep the behaviour to avoid breaking use cases and causing unexpected issues.

PR #4885, opened 2024-04-12
ARTEMIS-4702 only run command needs custom JVM settings

jbertram:


artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/artemis.profile
```diff
@@ -36,7 +36,7 @@ ARTEMIS_INSTANCE_ETC_URI='${artemis.instance.etc.uri}'
 HAWTIO_ROLE='${role}'
 
 # Java Opts
-if [ -z "$JAVA_ARGS" ]; then
+if [ -z "$JAVA_ARGS" ] && [ "$1" = "run" ]; then
```
clebertsuconic, line 39: the export might need special memory to load large journals as well.

Also, if you don't pass an argument you may enter the shell where you can do export, data print, and also.. run from the shell. (not very common, but possible).

brusdev:
Replaced by #5136

PR #4888, opened 2024-04-16
ARTEMIS-4725 AMQP Delivery reencoding before getDeliverBuffer

clebertsuconic:
Issue manifested during Mirror delivery and full brokers restart

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
@@ -700,7 +700,7 @@ public abstract class AMQPMessage extends RefCountMessage implements org.apache.
             if (Header.class.equals(constructor.getTypeClass())) {
                header = (Header) constructor.readValue();
                headerPosition = constructorPos;
-               encodedHeaderSize = data.position();
+               encodedHeaderSize = data.position() - constructorPos;
                if (header.getTtl() != null) {
                   if (!expirationReload) {
                      expiration = System.currentTimeMillis() + header.getTtl().intValue();
@@ -778,6 +778,7 @@ public abstract class AMQPMessage extends RefCountMessage implements org.apache.
     * @return a Netty ByteBuf containing the encoded bytes of this Message instance.
     */
    public ReadableBuffer getSendBuffer(int deliveryCount, MessageReference reference) {
+      ensureMessageDataScanned();
       ensureDataIsValid();
 
       DeliveryAnnotations daToWrite = reference != null ? reference.getProtocolData(DeliveryAnnotations.class) : null;
@@ -825,6 +826,7 @@ public abstract class AMQPMessage extends RefCountMessage implements org.apache.
       }
 
       writeDeliveryAnnotationsForSendBuffer(result, deliveryAnnotations);
+
       // skip existing delivery annotations of the original message
       duplicate.position(encodedHeaderSize + encodedDeliveryAnnotationsSize);
       result.writeBytes(duplicate.byteBuffer());
--- a/artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessageTest.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessageTest.java
@@ -48,6 +48,7 @@ import org.apache.activemq.artemis.api.core.ICoreMessage;
 import org.apache.activemq.artemis.api.core.RoutingType;
 import org.apache.activemq.artemis.api.core.SimpleString;
 import org.apache.activemq.artemis.core.message.openmbean.CompositeDataConstants;
+import org.apache.activemq.artemis.core.server.MessageReference;
 import org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageIdHelper;
 import org.apache.activemq.artemis.protocol.amqp.converter.AMQPMessageSupport;
 import org.apache.activemq.artemis.protocol.amqp.util.NettyReadable;
@@ -73,14 +74,18 @@ import org.apache.qpid.proton.amqp.messaging.Header;
 import org.apache.qpid.proton.amqp.messaging.MessageAnnotations;
 import org.apache.qpid.proton.amqp.messaging.Properties;
 import org.apache.qpid.proton.amqp.messaging.Section;
+import org.apache.qpid.proton.codec.AMQPDefinedTypes;
+import org.apache.qpid.proton.codec.DecoderImpl;
 import org.apache.qpid.proton.codec.EncoderImpl;
 import org.apache.qpid.proton.codec.EncodingCodes;
 import org.apache.qpid.proton.codec.ReadableBuffer;
+import org.apache.qpid.proton.codec.WritableBuffer;
 import org.apache.qpid.proton.message.Message;
 import org.apache.qpid.proton.message.impl.MessageImpl;
 import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
+import org.mockito.Mockito;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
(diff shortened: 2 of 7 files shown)


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -825,8 +826,14 @@ protected ReadableBuffer createDeliveryCopy(int deliveryCount, DeliveryAnnotatio
       }
 
       writeDeliveryAnnotationsForSendBuffer(result, deliveryAnnotations);
-      // skip existing delivery annotations of the original message
-      duplicate.position(encodedHeaderSize + encodedDeliveryAnnotationsSize);
+
+      if (headerPosition > deliveryAnnotationsPosition && headerPosition != VALUE_NOT_PRESENT && deliveryAnnotationsPosition != VALUE_NOT_PRESENT) {
+         // this is for a case where delivery annotations was swiched wrongly in a previous version
```
erwindon, line 831: swiched --> switched?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -825,8 +826,14 @@ protected ReadableBuffer createDeliveryCopy(int deliveryCount, DeliveryAnnotatio
...
+         duplicate.position(deliveryAnnotationsPosition + encodedDeliveryAnnotationsSize);
```
gemmellr, line 832: I dont really understand this. Is it trying to _handle_ the delivery-annotations having been written in the wrong place, in front of the header, in an old stored message?

If so, how does setting the position to the end of the delivery annotations do that? Wouldnt that mean it was now positioned at the start of the illegally-positioned header...which we have already just written [if needed] previously a few lines earlier in this method?

clebertsuconic: I tried using headerPosition+headerSize, and didn't work... I ended up thinking it was correct, but now that you raised this I don't know either.. :)


let me review this.. I might need your help

clebertsuconic: basically the test works with what I have now, and it failed with setting the position after the header.

gemmellr: I dont understand the issue to the extent you do, but if from the comment in the code you are saying the sections were reversed, then the code as-is simply does not seem correct. If they were something more than reversed, its not clear to me it could be handled. (Its not clear to me that it should be trying at all, its basically a corrupted message).

The tests also passed before, so that may just mean you dont have a focused enough test yet. I probably wouldnt be using Qpid JMS to ultimately check this either, it should be verifying/decoding the payload more closely. Arguably this should be more easily unit testable given its in the message object.

clebertsuconic: @gemmellr the issue manifested by stopping all brokers with messages accumulated. and this is what the test is doing.

It's also mixing versions so that they can be swapped.


I will add more tests but I will keep this test I wrote.

clebertsuconic: in regard to this calculation, the headerPosition was calculated wrongly when the header was switched.... I'm fixing it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPMessage.java
```diff
@@ -825,8 +826,14 @@ protected ReadableBuffer createDeliveryCopy(int deliveryCount, DeliveryAnnotatio
       }
 
       writeDeliveryAnnotationsForSendBuffer(result, deliveryAnnotations);
-      // skip existing delivery annotations of the original message
-      duplicate.position(encodedHeaderSize + encodedDeliveryAnnotationsSize);
+
```
tabish121, line 829: I'm not convinced you need to change this after having adjusted the encodedHeraderSize value to account for it not being at the start of the buffer.  I must say that it feels a bit odd to be allowing wrongly encoded AMQP message format zero messages through in the first place, we probably should have been catching that and rejecting the messages really.

tabish121: I created a couple simple tests for this and just the one line fix [here](https://github.com/tabish121/activemq-artemis/commit/7f19cfc965d06c08217967fcf5bfb6d28d5058da) and that seems to be all that's needed:

clebertsuconic: @tabish121 this is working around wrongly encoded messages from the older version.

in case an user has messages wrongly encoded, they could just switch to the newer version that would know how to deal with it.

What is the alternative?

clebertsuconic: and this is the reason I expanded the compatibility-tests.. if you generated the messages in 2.33, stopped both brokers with accumulated messages, the clients won't know how to parse them because they are wrongly encoded.

tabish121: The compatibly tests are fairly incomprehensible so I can't really tell what it is doing or if it is or isn't reproducing anything.  I added a link to commit that tests the basic handling of message encoded with delivery annotations ahead of headers and showed that the one line change to fix the encoded header size sorts that scenario.

clebertsuconic: @tabish121 the compatibility tests is creating the situation with 2.33.. and consuming the message in HEADER/main


basically creating the real condition where it happened.

clebertsuconic: and the messages are not consumed properly with your fix.. I just tested it.

tabish121: Then it would be good to create a test case in the AMQPMessageTest that is reproducing the formatting and verifying a fix, you fix doesn't look correct to me but without knowing what you are actually reproducing its hard to judge

clebertsuconic: The test is creating the wrong encoding in 2.33 and consuming it in main using real bits from 2.33, just like the customer case.

I will work on the AMQPMessageTest.

clebertsuconic: @tabish121 my fix is only applied when the header is switched with the delivery annotation, an aberration that shouldn't actually ever happen, but it did.

clebertsuconic: @tabish121 the test you wrote is validating if the parsing would work correctly.. however with your fix the client is not being to actually receive the message.

You would need to write a test with the wrong encoding and have an actual client receiving the message.

tabish121: There should be no need for a client to test this, just the test decodes the message based on what you said is the issue, without knowing more I can't write a better test but that doesn't mean it isn't testable without a client.

PR #4894, opened 2024-04-17
ARTEMIS-4726 removing scheduled msg from q via mngmnt can cause negative msg count

jbertram:


jbertram:
@clebertsuconic, can you review and merge for 2.34.0?

PR #4897, opened 2024-04-20
ARTEMIS-4420 user auth leaks into non-Artemis servlets

jbertram:


jbertram:
@clebertsuconic, let's get this into 2.34.0. Can you review and merge? Thanks!

gtully (commented):
my understanding is the problem is with the filters setting thread locals and not clearing them,  so even if they filters are only on specific contexts, the threads won't respect that, and the data will leak.

gtully:
Using a thread local to propagate the session subject is fine, but it needs to be scoped to the user of that thread for the request, and cleared on response. so set every time.

jbertram:
@gtully, point taken. I've updated the PR with what I believe will address the `ThreadLocal` issue. I wasn't able to come up with a way to test it automatically, but manual tests (e.g. the use-case outlined in the Jira) is working fine now.

artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
```diff
@@ -219,6 +239,18 @@ public synchronized void start() throws Exception {
       cleanupTmp();
       server.start();
 
+      /*
```
gtully, line 242: I would revert this change. 
the metrics call to jmx which is instrumented to audit, with out this filter the audit on metrics will be less informative.
in addition, there are cases where access to metrics will need authentication, especially when there is RBAC on JMX mbeans.

artemis-web/src/main/java/org/apache/activemq/artemis/component/WebServerComponent.java
```diff
@@ -166,6 +173,19 @@ public synchronized void start() throws Exception {
...
                webContext.setInitParameter(DIR_ALLOWED, "false");
                webContext.getSessionHandler().getSessionCookieConfig().setComment("__SAME_SITE_STRICT__");
+               webContext.addEventListener(new ServletContextListener() {
+                  @Override
+                  public void contextInitialized(ServletContextEvent sce) {
+                     sce.getServletContext().addListener(new ServletRequestListener() {
+                        @Override
+                        public void requestDestroyed(ServletRequestEvent sre) {
+                           ServletRequestListener.super.requestDestroyed(sre);
+                           AuditLogger.currentCaller.remove();
```
gtully, line 183: that looks right

gtully (changes requested):

gtully (approved):
that looks good to me.

PR #4902, opened 2024-04-22
ARTEMIS-4741 use specific assertion methods

jbertram:


gemmellr:
I started doing equivalent updates and though the assertNull changes definitely make sense, its not clear a lot of the assertTrue/assertFalse changes necessarily do. They are mostly checking Boolean (rather than boolean) return values, so by changing them to assertTrue/assertFalse they are potentially just going to throw an NPE before the assertion actually happens. I'd be inclined to leave those ones as-is.

gemmellr:
> I started doing equivalent updates and though the assertNull changes definitely make sense, its not clear a lot of the assertTrue/assertFalse changes necessarily do. They are mostly checking Boolean (rather than boolean) return values, so by changing them to assertTrue/assertFalse they are potentially just going to throw an NPE before the assertion actually happens. I'd be inclined to leave those ones as-is.

I went through them all, less were Boolean than it initially seemed. I converted almost all of them (left the 10 or so where the type wasnt even known to be Boolean, since the assertEquals handles both the null and the type) and added assertNotNulls beforehand for the Boolean ones. 

In addition to these original changes, also converted an assertNotEquals(null, foo) into assertNotNull(foo).

PR #4904, opened 2024-04-22
ARTEMIS-4742 decoding PersistedSecuritySetting fails after upgrade

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/security/PersistedSecuritySettingTest.java
```diff
@@ -43,4 +48,31 @@ public void testNPE() {
       persistedSecuritySetting.getViewRoles();
       persistedSecuritySetting.getEditRoles();
    }
+
+   @Test
+   public void testUpgradeAfterARTEMIS_4582() {
```
clebertsuconic, line 53: did you think about writing a compatibility test, perhaps using the functionality?

jbertram: I didn't. The stack-trace on the exception was pretty clear, and since we've dealt with this kind of issue before a quick unit test seemed adequate.

artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/config/PersistedSecuritySetting.java
```diff
@@ -252,8 +252,12 @@ public void decode(final ActiveMQBuffer buffer) {
       browseRoles = buffer.readNullableSimpleString();
       createAddressRoles = buffer.readNullableSimpleString();
       deleteAddressRoles = buffer.readNullableSimpleString();
-      viewRoles = buffer.readNullableSimpleString();
```
gtully, line 255: darn, I read the method readNullableSimpleString as doing the check for an non existent value, my bad,

PR #4906, opened 2024-04-25
ARTEMIS-4744 Fully suppoert multple host broker connections URIs

tabish121:
Create a new NettyConnector for each connection attempt that is configured from distinct broker connection URIs which allows for differing TLS configuration per remote connection configuration.


jbertram:
Nitpick - there's a typo in the commit message (i.e. "suppoert").

tabish121:
I've run this through the full test suite a few times.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPConnectSaslTest.java
```diff
@@ -214,4 +216,186 @@ private void doConnectWithExternalTestImpl(boolean requireClientCert) throws Exc
...
+         final URI secondPeerURI = secondPeer.getServerURI();
+         logger.debug("Connect test started, second peer listening on: {}", secondPeerURI);
+
+         // First connection fails because we use a server certificate with whose common name
+         // doesn't match the host, second connection should work as we disable host verification
+         String amqpServerConnectionURI =
+            "(tcp://localhost:" + firstPeerURI.getPort() + "?verifyHost=true" +
+            ",tcp://localhost:" + secondPeerURI.getPort() + "?verifyHost=false)" +
+               "?sslEnabled=true;trustStorePath=" + SERVER_TRUSTSTORE_NAME +
+               ";trustStorePassword=" + SERVER_TRUSTSTORE_PASSWORD;
```
gemmellr, line 263: The docs around this indicate a different syntax, not having "(...)", and saying you are meant to use # for the first separator: https://github.com/apache/activemq-artemis/blob/2.33.0/docs/user-manual/amqp-broker-connections.adoc?plain=1#L87-L97

Though I dont really understand why given it says what is after is to be comma separated. Its not clear if you were meant to be able to specify different options for the CSV elements, or just host:port values as it indicates, since it doesnt say either way.

The existing test I found around this (that doesnt use TLS and so didnt notice it didnt work so much with that) uses the same basic config from the doc: https://github.com/apache/activemq-artemis/blob/2.33.0/tests/e2e-tests/src/main/resources/servers/brokerConnect/replicaMainServerA/broker.xml#L59

We should probably be testing whatever the doc says...and/or updating the docs to clarify what is actually expected to work.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPConnectSaslTest.java
```diff
@@ -214,4 +216,186 @@ private void doConnectWithExternalTestImpl(boolean requireClientCert) throws Exc
...
+
+         AMQPBrokerConnectConfiguration amqpConnection =
+            new AMQPBrokerConnectConfiguration(getTestName(), amqpServerConnectionURI);
+         amqpConnection.setRetryInterval(100); // Allow reconnects
```
gemmellr, line 267: Maybe put a limit on attempts, just in case?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPConnectSaslTest.java
tabish121: I cannot make it work using a '#' and I've based what I've used on other code I cam across elsewhere.  If someone that can understand how the parser is supposed to work wants to document it further that'd be great but I am not that person.

tabish121: I think I have figured out the fragment style URI it wants and have updated the tests so it should match the broker connection documentation (although that is the least obvious format).

gemmellr: I have been fine with updating the docs to match the previous tests. Just seemed rather off for the 2 to dramatically differ is all :)

PR #4907, opened 2024-04-26
ARTEMIS-4745 Allow configuration of the pull consumer batch size

tabish121:
Allow for configuration of the batch size granted to the remote when an AMQP federation queue receiver is pulling messages only when there is local capacity to handle them. Some code housekeeping is done here to make adding future properties a bit simpler and require fewer changes.


tabish121:
I've run the full test suite with these changes

PR #4909, opened 2024-04-30
ARTEMIS-3622 MQTT can deadlock on client cxn/discxn

jbertram:
This commit fixes the deadlock described on ARTEMIS-3622 by moving the synchronization "up" a level from the MQTTSession to the MQTTConnectionManager. It also eliminates the synchronization on the MQTTSessionState in the MQTTConnectionManager because it's no longer needed. This change should not only eliminate the deadlock, but improve performance relatively as well.

There is no test associated with this commit as I wasn't able to reproduce the deadlock with any kind of straight-forward test. There was a test linked on the Jira, but it involved intrusive and fragile scaffolding and wasn't ultimately tenable. That said, I did test this fix with that test and it was successful. In any case, I think static analysis should be sufficient here as the changes are pretty straight-forward.

### DIFF
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
@@ -51,7 +51,7 @@ public class MQTTConnectionManager {
       session.getConnection().addFailureListener(failureListener);
    }
 
-   void connect(MqttConnectMessage connect, String validatedUser, String username, String password) throws Exception {
+   synchronized void connect(MqttConnectMessage connect, String validatedUser, String username, String password) throws Exception {
       if (session.getVersion() == MQTTVersion.MQTT_5) {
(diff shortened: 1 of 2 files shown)


artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
```diff
@@ -176,33 +174,27 @@ ServerSessionImpl createServerSession(String username, String password, String v
       return (ServerSessionImpl) serverSession;
    }
 
-   void disconnect(boolean failure) {
+   synchronized void disconnect(boolean failure) {
       if (session == null || session.getStopped()) {
          return;
       }
 
-      synchronized (session.getState()) {
```
gemmellr, line 184: This seems fine in terms of protecting the connect+disconnect calls similar to before...but, are there other areas synchronising on and accessing or manipulating the session state that might no longer be protected as before that still need to be?

jbertram: I've looked through it carefully and I don't see any potential problems. The relevant changes to the state are performed by `connect` and `disconnect` which are themselves now use `synchronized`.

PR #4911, opened 2024-05-01
ARTEMIS-4751 upgrade to Apache parent 32

jbertram:
Due to https://github.com/apache/maven-apache-parent/pull/188 the property `maven.compiler.release` is now being set which precludes exporting and using any internal Java classes. Therefore this commit removed references to `--add-exports` from the build, switches to reflection, and adds `--add-opens` to the runtime JVM parameters.


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/perf/MicrosClock.java
```diff
@@ -43,23 +45,28 @@ public static boolean isAvailable() {
    }
 
    public static long now() {
-      long epochSecond = offset;
-      long nanoAdjustment = jdk.internal.misc.VM.getNanoTimeAdjustment(epochSecond);
+      try {
+         long epochSecond = offset;
+         Class vm = Class.forName("jdk.internal.misc.VM");
+         Method getNanoTimeAdjustment = vm.getMethod("getNanoTimeAdjustment", long.class);
+         long nanoAdjustment = (long) getNanoTimeAdjustment.invoke(getNanoTimeAdjustment, epochSecond);
```
gemmellr, line 52: This might be better done once, e.g during the initial availability checks, and then the Method reused here rather than doing lookups every time, given the aimed usage ?

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/artemis.profile
```diff
@@ -37,7 +37,7 @@ HAWTIO_ROLE='${role}'
 
 # Java Opts
 if [ -z "$JAVA_ARGS" ]; then
-    JAVA_ARGS="-XX:AutoBoxCacheMax=20000 -XX:+PrintClassHistogram -XX:+UseG1GC -XX:+UseStringDeduplication -Xms512M -Xmx${java-memory} -Dhawtio.disableProxy=true -Dhawtio.realm=activemq -Dhawtio.offline=true -Dhawtio.rolePrincipalClasses=org.apache.activemq.artemis.spi.core.security.jaas.RolePrincipal -Dhawtio.http.strictTransportSecurity=max-age=31536000;includeSubDomains;preload -Djolokia.policyLocation=${ARTEMIS_INSTANCE_ETC_URI}jolokia-access.xml -Dlog4j2.disableJmx=true ${java-opts}"
+    JAVA_ARGS="-XX:AutoBoxCacheMax=20000 -XX:+PrintClassHistogram -XX:+UseG1GC -XX:+UseStringDeduplication -Xms512M -Xmx${java-memory} -Dhawtio.disableProxy=true -Dhawtio.realm=activemq -Dhawtio.offline=true -Dhawtio.rolePrincipalClasses=org.apache.activemq.artemis.spi.core.security.jaas.RolePrincipal -Dhawtio.http.strictTransportSecurity=max-age=31536000;includeSubDomains;preload -Djolokia.policyLocation=${ARTEMIS_INSTANCE_ETC_URI}jolokia-access.xml -Dlog4j2.disableJmx=true --add-opens java.base/jdk.internal.misc=ALL-UNNAMED ${java-opts}"
```
gemmellr, line 40: Hadnt actually considered this possibility...but if doing it, is there a windows equivalent needed somewhere?

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/messages/perf/MicrosClock.java
jbertram: Done.

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/artemis.profile
jbertram: Good call. Done.

PR #4912, opened 2024-05-01
ARTEMIS-4752 upgrade ErrorProne to 2.27.0

jbertram:


jbertram:
@gemmellr, I don't meant to step on your toes with your work on the JUnit 5 upgrade, but these test changes were necessary for the upgrade. If necessary we can certainly wait until your done if you prefer.

gemmellr:
I only asked not to merge the other one and have me roll related changes into my work since it was quite big but wasnt actually 'fixing' anything that was 'broken', so it would have been annoying to deal with the conflicts given that...this is fixing stuff, so it goes in and I deal with it along with the other rebase stuff needed. (Happily its also smaller).

PR #4916, opened 2024-05-01
ARTEMIS-4754 Structure the names used for federation internal queues

tabish121:
When creating internal temporary queues for the federation control links and the events link we should use a structured naming convention to ease in configuring security for the federation user where all internal names fall under a root prefix which can be used to grant read and write access for the federation user. This change allows security on the wildcarded address "$ACTIVEMQ_ARTEMIS_FEDERATION.#". This change also includes some further restrictions added to federation resources and adds support for wildcarding '$' prefixed addresses.


gtully (approved):
this looks great. 
I note that we may have a general problem with match addresses that contain regular expression constructs, would a user ever use | in their address name, I guess never say never. But that can be for another day.

gemmellr:
LGTM too. Might be best for e.g @jbertram to look before merge too, given the match change.

tabish121:
> LGTM too. Might be best for e.g @jbertram to look before merge too, given the match change.

@jbertram actually helped me figure out why the wildcards weren't working on the internal addresses. We both thought it should be working already before tracking down the change here.

PR #4918, opened 2024-05-02
ARTEMIS-4758 Hardening Mirroring

clebertsuconic:
This is a list of improvements done as part of this commit / task:

* Page Transactions on mirror target are now optional.

If you had an interrupt mirror while the target destination was paging, duplicate detection would be ineffective unless you used paged transactions Users can now configure the ack manager retries intervals. Say you need some time to remove a consumer from a target mirror. The delivering references would prevent acks from happening. You can allow bigger retry intervals and number of retries by tinkiering with ack manager retry parameters.

* AckManager restarted independent of incoming acks

The ackManager was only restarted when new acks were coming in. If you stopped receiving acks on a target server and restarted that server with pending acks, those acks would never be exercised. The AckManager retries are now restarted as soon as the server is started.


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/MirrorTransaction.java
```diff
@@ -49,4 +51,13 @@ protected synchronized void afterCommit(List<TransactionOperation> operationsToC
...
 
+   @Override
+   public boolean isAllowPageTransaction() {
+      return ignorePageTransaction;
+   }
+
+   public MirrorTransaction setIgnorePageTransaction(boolean ignorePageTransaction) {
+      this.ignorePageTransaction = ignorePageTransaction;
+      return this;
+   }
```
gemmellr, line 62: The naming of the methods, and the variable feels off. The variable defaults to false, so ignorePageTransaction=false, meaning the return of 'isAllowPageTransaction()' is false. So it 'isnt ignored', but it also 'isnt allowed'? Confusing. Is there a better name here for one of them to make clearer?

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/Configuration.java
```diff
@@ -1509,4 +1509,32 @@ default String resolvePropertiesSources(String propertiesFileUrl) {
...
+    *  It is not intended to be configured through the XML.
+    *  The default value here is 100, and this is in milliseconds. */
+   int getMirrorAckManagerRetryDelay();
+
+   Configuration setMirrorAckManagerRetryDelay(int delay);
+
+   /** Should Mirror use Page Transactions When target destinations is paging?
+    *  When a target queue on the mirror is paged, the mirror will not record a page transaction for every message.
+    *  The default is false, and the overhead of paged messages will be smaller, but there is a possibility of eventual duplicates in case of interrupted communication between the mirror source and target.
+    *  If you set this to false there will be extra syncs and a record stored on the journal for the page-transaction additionally to the record in the page store. */
```
gemmellr, line 1536: The 'default is false, and the overhead will be smaller'....but then, 'if you set this false, there will be \<extra overhead\>' ?    Which is it?

artemis-server/src/main/java/org/apache/activemq/artemis/core/transaction/Transaction.java
```diff
@@ -109,4 +109,8 @@ enum State {
 
    /** To be used on control transactions that are meant as internal and don't really require a hard sync. */
    Transaction setAsync(boolean async);
+
+   default boolean isAllowPageTransaction() {
+      return true;
+   }
```
gemmellr, line 115: Similar confusion that this code default is true, but the overriding method looks to get a field value thats false by default?

gemmellr (commented):
I didnt really follow much of this, but I noticed one bug, and then some especially confusing bits that may also be.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/MirrorTransaction.java
clebertsuconic: ohhh.... I renamed the method and forgot to rename that one.. thanks

artemis-commons/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQScheduledComponent.java
```diff
@@ -212,15 +212,19 @@ public long getPeriod() {
...
+         restartIfNeeded();
+      }
       return this;
    }
 
    public synchronized ActiveMQScheduledComponent setPeriod(long period, TimeUnit unit) {
-      this.period = period;
-      this.timeUnit = unit;
-      restartIfNeeded();
+      if (this.period != period && this.timeUnit != unit) {
```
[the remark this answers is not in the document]
clebertsuconic: Yeah... there's a bug in here... 


which means I'm not testing this :)


Will add a test

artemis-server/src/main/java/org/apache/activemq/artemis/core/transaction/Transaction.java
clebertsuconic: MirrorTransactions will disable page transaction by default. 

I have corrected the names

clebertsuconic:
I need to merge this.. 

We can always follow up with further changes and PRs if needed.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/ProtonProtocolManagerFactory.java
```diff
@@ -87,7 +87,8 @@ public String getModuleName() {
    @Override
    public void loadProtocolServices(ActiveMQServer server, List<ActiveMQComponent> services) {
       try {
-         AckManager ackManager = AckManagerProvider.getManager(server, false);
+         AckManager ackManager = AckManagerProvider.getManager(server);
+         services.add(ackManager);
```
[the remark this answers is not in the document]
clebertsuconic: I didn't think so. 

If you confirm it to be an issue it wouldn't be critical but it should be fixed.

PR #4921, opened 2024-05-02
ARTEMIS-4743 Improve Queue Stat output by allowing --loop as an option

clebertsuconic:


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/queue/StatQueue.java
```diff
@@ -99,6 +100,12 @@ public enum OPERATION {
    @Option(names = "--include-management", description = "Include queues created for notification management in the output")
    private boolean includeManagement = false;
 
+   @Option(names = "--loop", description = "Keep Queue Stat in a forever loop, that you can interrupt with Ctrl-C, sleeping for --loop-time between each iteration.")
```
brusdev, line 103: In kubectl the name of a similar option is `watch`

PR #4925, opened 2024-05-07
ARTEMIS-4764 increase automation for dependecy updates

jbertram:


gemmellr:
Seems reasonable, but is presumably going to cause it to spam a bunch of PRs (10 at a time) initially for various stuff we arent on the latest for, e.g especially test deps, so I was thinking we should wait until after the already-delayed release.

jbertram:
I'm going to merge this and start dealing with the influx of PRs.

PR #4928, opened 2024-05-08
ARTEMIS-4765 DuplicateIDCache on Mirror Target is using 20K elements instead of amqpCredits

clebertsuconic:
in this commit I'm storing a binding record with the address-settings for the correct size this is also validating eventual merges of the AddressSettings in the same namespace.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -60,6 +60,8 @@ import org.apache.activemq.artemis.core.message.impl.CoreMessage;
 import org.apache.activemq.artemis.core.paging.PagingManager;
 import org.apache.activemq.artemis.core.paging.PagingStore;
 import org.apache.activemq.artemis.core.persistence.StorageManager;
+import org.apache.activemq.artemis.core.persistence.config.AbstractPersistedAddressSetting;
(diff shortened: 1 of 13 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1467,6 +1476,22 @@ public DuplicateIDCache getDuplicateIDCache(final SimpleString address, int cach
...
+      if (recordedSetting == null || recordedSetting.getSetting().getIDCacheSize() == null || recordedSetting.getSetting().getIDCacheSize().intValue() != cacheSizeToUse) {
+         AddressSettings settings = recordedSetting != null ? recordedSetting.getSetting() : new AddressSettings();
+         settings.setIDCacheSize(cacheSizeToUse);
+         server.getAddressSettingsRepository().addMatch(address.toString(), settings);
+         try {
+            storageManager.storeAddressSetting(new PersistedAddressSettingJSON(address, settings, settings.toJSON()));
+         } catch (Exception e) {
+            // nothing could be done here, we just log
+            // if an exception is happening, if IO is compromised the server will eventually shutdown
+            logger.warn(e.getMessage(), e);
```
gtully, line 1490: better say what the context here is, "error storing an address setting, reason: " + e.getMessage()

clebertsuconic: I'm adding a loggerID.. thanks

PR #4929, opened 2024-05-08
ARTEMIS-4768 _AMQ_SCHED_DELIVERY msg prop lost after broker restart

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/PostOfficeJournalLoader.java
```diff
@@ -243,10 +243,6 @@ public void handleAddMessage(Map<Long, Map<Long, AddMessageRecord>> queueMap) th
                MessageReference ref = postOffice.reload(record.getMessage(), queue, null);
 
                ref.setDeliveryCount(record.getDeliveryCount());
-
-               if (scheduledDeliveryTime != 0) {
-                  record.getMessage().setScheduledDeliveryTime(0L);
```
clebertsuconic, line 248: I think this was on purpose. If you are reloading it it shouldn't be scheduled.

jbertram: If you look a few lines up from the code I removed you'll see this:
```java
if (scheduledDeliveryTime != 0) {
   record.getMessage().setScheduledDeliveryTime(scheduledDeliveryTime);
}
```
So first we set it to `scheduledDeliveryTime` and then we set it to `0`. One of these must be wrong. I believe the one I removed in the wrong one.

There is another block still further up that does this:
```java
if (scheduledDeliveryTime != 0 && scheduledDeliveryTime <= currentTime) {
   scheduledDeliveryTime = 0;
   record.getMessage().setScheduledDeliveryTime(0L);
}
```
Reloading a message _whose scheduled delivery time is already passed_ should not be scheduled. Is this what you were referring to?

jbertram: @clebertsuconic, can you take another look at this before 2.34.0?

PR #4930, opened 2024-05-10
NO-JIRA fix intermittent failure, this test failed on local repeat ge…

gtully:
…tting a null from immediate


clebertsuconic:
I bump into things like this all the time :) 

not sure if something we recently changed made it happen.. or it's just another one you found..


I'm merging it though.

PR #4934, opened 2024-05-15
ARTEMIS-4745 Expose registered broker plugin class names as JMX Attribute

mnapierajAvSystem:


tabish121:
Seems as though you should be referencing [ARTEMIS-4772](https://issues.apache.org/jira/browse/ARTEMIS-4772) in this commit.

mnapierajAvSystem:
you are absolutely right. Updating commit message will be enough or create a branch with proper name and new PR is required?

PR #4935, opened 2024-05-15
ARTEMIS-4772 Expose registered broker plugin class names as JMX Attribute

mnapierajAvSystem:


jbertram:
Thanks for the contribution! I squashed the commits and cleaned up a few things during the merging process. Nice work!

PR #4936, opened 2024-05-15
NO-JIRA: fix docs locator.createClientSessionFactory should be locator.createSessionFactory

coursar:
[`ServerLocator`](https://github.com/apache/activemq-artemis/blob/ae5d98337f00d5a123d95994b8437cdca1d319c4/artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/client/ServerLocator.java#L41) interface doesn't have `createClientSessionFactory` method -> should be `createSessionFactory`


clebertsuconic:
thank you

PR #4937, opened 2024-05-16
ARTEMIS-4774 Fixing PageCounters out of sync after AckMnager retries

clebertsuconic:


clebertsuconic:
I need to merge this PR. I have hours and hours of testing on this where I tweaked parameters.sh in multiple ways  and the counters are always accurate.


If there are minor issues or even tweaks that are needed to the code those could be addressed in a later PR.. .or feel free to send further commits on the branch.

PR #4941, opened 2024-05-22
ARTEMIS-4777 Add additional configuration for broker connection

tabish121:
Add additional configuration for the broker connection senders and receivers that allows for setting the source or target address value and optional source and target capabilities to allow providing address mapping on some remote peers.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
@@ -17,6 +17,7 @@
 package org.apache.activemq.artemis.protocol.amqp.connect;
 
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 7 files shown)


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
```diff
@@ -701,7 +745,11 @@ private void connectSender(Queue queue,
             sender.setSenderSettleMode(SenderSettleMode.UNSETTLED);
             sender.setReceiverSettleMode(ReceiverSettleMode.FIRST);
             Target target = new Target();
-            target.setAddress(targetName);
+            if (targetAddress != null && !targetAddress.isBlank()) {
+               target.setAddress(targetAddress);
```
gemmellr, line 749: This targetName / targetAddress distinction is a bit awkard with both around given that 'target address' is usually the normal terminology and what the value actually sets. Perhaps change the prior one to targetAddress instead of targetName, then use targetAddressOverride for the new 'use this instead' value to clarify things?

Similarly for the receiver bit above.

EDIT: or see other comment, simplify this and move decision to the call site where config is being pulled originally.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
```diff
@@ -282,18 +285,52 @@ public void validateMatching(Queue queue, AMQPBrokerConnectionElement connection
...
+         if (connectionElement instanceof AMQPSenderBrokerConnectionElement) {
+            final AMQPSenderBrokerConnectionElement sender = (AMQPSenderBrokerConnectionElement) connectionElement;
+
+            targetAddress = sender.getTargetAddress();
+            targetCapabilities = parseTerminusCapabilities(sender.getTargetCapabilities());
          }
-         if (connectionElement.getType() == AMQPBrokerConnectionAddressType.RECEIVER) {
-            connectReceiver(protonRemotingConnection, session, sessionContext, queue);
+
+         connectSender(queue, queue.getAddress().toString(), null, null, null, null, null, targetAddress, targetCapabilities);
```
gemmellr, line 301: Could we actually just move the other 'is an explicit override address defined' bit here, and then simply pass 'targetName' as before (perhaps renamed to targetAddress though?), but which is then either the queues address as originally, or the new configured 'override' target if defined? Would save passing around 2 competing values for 1 thing and leave the other code simpler / as-is.

Similarly for the receiver bit below.

PR #4942, opened 2024-05-23
ARTEMIS-4771 AMQP Large Message Writer no-op staged write if already …

tabish121:
…closed

If the writer is closed while flow controlled write attempts are pending in the connection executor an NPE is thrown which should be avoided as the closed state should trigger the writes to stop. Add fix and test that checks that this works as it should.


artemis-protocols/artemis-amqp-protocol/src/test/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageWriterTest.java
```diff
@@ -0,0 +1,255 @@
...
+         Runnable runnable = invocation.getArgument(0);
+
+         try {
+            runnable.run();
+         } catch (Exception e) {
+            fail("Queued invocation of writer tasks should not throw.");
+         }
+
+         return null;
+     }).when(connectionContext).runLater(any(Runnable.class));
```
gemmellr, line 231: Checkstyle failed here, missing 1 space.

PR #4944, opened 2024-05-27
ARTEMIS-4306 add authn/z cache metrics

jbertram:
This commit includes the following changes:

 - Management operations to get sucess & failure counts for authn and authz along with the corresponding audit logging.
 - Export the aforementioned authn & authz metrics.
 - Export metrics for the underlying authn & authz caches including the ability to enable/disable them.
 - Update metrics tests to validate tags in addition to keys and values.
 - Update documentation to explain new functionality and clarify existing metric tags.

### DIFF
--- a/artemis-server/src/main/resources/schema/artemis-configuration.xsd
+++ b/artemis-server/src/main/resources/schema/artemis-configuration.xsd
@@ -4913,6 +4913,14 @@
                </xsd:annotation>
             </xsd:element>
 
+            <xsd:element name="security-caches" type="xsd:boolean" default="false" maxOccurs="1" minOccurs="0">
+               <xsd:annotation>
+                  <xsd:documentation>
+                     whether to report metrics for the authentication and authorization caches
+                  </xsd:documentation>
+               </xsd:annotation>
+            </xsd:element>
+
             <xsd:element name="plugin" maxOccurs="1" minOccurs="0">
                <xsd:complexType>
                   <xsd:annotation>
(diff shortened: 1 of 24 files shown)


artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -4913,6 +4913,14 @@
                </xsd:annotation>
             </xsd:element>
 
+            <xsd:element name="security-caches" type="xsd:boolean" default="false" maxOccurs="1" minOccurs="0">
```
clebertsuconic, line 4916: shouldn't this be called security-events, perhaps security-attempts?

jbertram: The instances of Caffeine caches which are related to security is what is being instrumented here. Metrics that are made available when this is set to `true` are things like:

- `cache.size`
- `cache.puts`
- `cache.gets` tagged by `result` - either `hit` or `miss`
- `cache.evictions`
- `cache.eviction.weight`

Therefore, calling it `security-caches` seems appropriate.

clebertsuconic:
@jbertram did you run the testsuite on this?
because it broke quite a few tests.

jbertram:
It only really broke 2 tests. It's just those two tests are parameterized and inherited so the failures multiply. I verified the fixes locally and pushed. I'm running the test-suite now to fully verify.

jbertram:
Test-suite looks good with regard to this PR now. Thank for the heads up!

PR #4948, opened 2024-05-29
ARTEMIS-4165 Do delete messages on queue delete

clebertsuconic:


jbertram:
Looking at the solution I'm not sure how it actually addresses the problem. Can you provide a short description of why it works?

clebertsuconic:
Before this change, Upon a restart the broker would check for orphaned Page Transactions and not right away as the queue is removed.

With this change in place the broker will now just remove the messages from paging as it would on regular messages, and they would be acked accordingly.


This is probably also good for mirroring... although I didn't check the result of such event on mirror.

clebertsuconic:
@jbertram I was just making the PR for the temporary branch @iliya-gr had created.. I just realized it was a few years ago.

I will update the PR to remove the page transactions without using ACK records.

PR #4949, opened 2024-05-30
ARTEMIS-4788 Fix a rare race on broker shutdown in AMQP federation

tabish121:
Race on consumer create and broker shutdown could lead to a deadlocak trying to access configuration from the policy manager while the federation instance is trying to shutdown the policy manager.


clebertsuconic:
LGTM

PR #4950, opened 2024-05-31
ARTEMIS-4789 Page.destroy race with cleanup

clebertsuconic:
this is fixing PagingTest::testPagingStoreDestroyed(db=derby) (or any other DB).

This could eventually also fail on journal.

The cleanup could act on a destroyed folder, and issue an IOException stopping the server with this race.

You would need the server active cleaning messages while the queue is being removed.


clebertsuconic:
I will squash this change after some tests.

I'm using a valid commit message in case someone hit merge before I finish the test, but I intend to squash it.

PR #4953, opened 2024-06-03
ARTEMIS-4790: use JUnit 5 for the tests

gemmellr:
Use JUnit 5 for the test suite; primarily the Jupiter API and junit-jupiter-engine, but junit-vintage-engine for a few select modules or test classes that rely on runners or deps that still need the earlier APIs.

Given the changes are test-only and the significant amount of time I have already spent rebasing them to this point, I'd like to merge this quickly and then address tweaking as followups.


jbertram:
After a skim I'm proceeding with a commit-then-review approach to avoid having to rebase this more than you've already had to.

PR #4956, opened 2024-06-03
ARTEMIS-4792 Allow consumer priority to be added to receiver link add…

tabish121:
…ress

Allow the Source address to provide consumer priority on the address using the same option value as a core consumer '?consumer-priority=X'. The change parses any query string appended to an address and uses the address portion as the actual receiver address and currently only looks at consumer priority values in the extracted address query parameters and ignores any other options found. The existing consumer priority taken from link properties takes precedence over the value placed on the address query options if both are present.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ParameterisedAddress.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ParameterisedAddress.java
@@ -19,6 +19,7 @@ package org.apache.activemq.artemis.api.core;
 import static org.apache.activemq.artemis.utils.uri.URISupport.appendParameters;
 import static org.apache.activemq.artemis.utils.uri.URISupport.parseQuery;
 
+import java.util.Collections;
 import java.util.Map;
 
 import org.apache.activemq.artemis.utils.uri.URISupport;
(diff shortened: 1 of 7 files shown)


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/DefaultSenderController.java
```diff
@@ -409,16 +419,38 @@ public Consumer init(ProtonServerSenderContext senderContext) throws Exception {
...
-      return (Consumer) sessionSPI.createSender(senderContext, queue, multicast ? null : selector, browseOnly);
+      return null;
    }
 
    @Override
    public void close() throws Exception {
       Source source = (Source) protonSender.getSource();
-      if (source != null && source.getAddress() != null && multicast) {
-         SimpleString queueName = SimpleString.toSimpleString(source.getAddress());
+      if (source != null && getSourceAddress(source) != null && multicast) {
```
gemmellr, line 452: Given that getSourceAddress(source) handles null Source values, would seem like this could be put in a variable here rather than potentially doing the extraction twice as it is.

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ParameterisedAddress.java
```diff
@@ -106,4 +107,46 @@ public static boolean isParameterised(SimpleString address) {
...
+    */
+   public static String extractAddress(String address) {
+      final int index = address.indexOf('?');
+
+      if (index == -1) {
+         return address;
+      } else {
+         return address.substring(0, index);
+      }
+   }
```
gemmellr, line 151: The methods cant handle a null but dont say that and the 'otherwise return provided' bit might imply otherwise. Should either say so or handle it.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/DefaultSenderController.java
```diff
@@ -207,6 +212,10 @@ public Consumer init(ProtonServerSenderContext senderContext) throws Exception {
          }
          source.setAddress(queue.toString());
       } else {
+         final String sourceAddress = ParameterisedAddress.extractAddress(source.getAddress());
+
+         addressParameters = ParameterisedAddress.extractParameters(source.getAddress());
```
gemmellr, line 217: The method cant handle being passed a null, but its possible this value is null.

I'd guess that might even explain the unit test failure (from an exception type having changed, to an internal exception)

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/DefaultSenderController.java
```diff
@@ -440,7 +474,7 @@ public void close() throws Exception {
          }
       } else if (source != null && source.getDynamic() && (source.getExpiryPolicy() == TerminusExpiryPolicy.LINK_DETACH || source.getExpiryPolicy() == TerminusExpiryPolicy.SESSION_END)) {
          try {
-            sessionSPI.removeTemporaryQueue(SimpleString.toSimpleString(source.getAddress()));
+            sessionSPI.removeTemporaryQueue(SimpleString.toSimpleString(getSourceAddress(source)));
```
gemmellr, line 477: Can this one use the new variable too?

PR #4959, opened 2024-06-04
ARTEMIS-4796 simplify SimpleString API

jbertram:

 - deprecate the verbosely named `toSimpleString` static factory methods
 - add `of` static factory methods for all the ctors
 - replace any uses of the normal ctors with the `of` counterparts

This makes the code more concise and readable.


artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/SimpleString.java
```diff
@@ -79,6 +96,8 @@ public static SimpleString toSimpleString(final String string, StringSimpleStrin
    public SimpleString(final String string) {
       int len = string.length();
 
+      String.valueOf(1);
```
tabish121, line 99: What is this for in this context?

jbertram: Whoops. That was left in by mistake.

gemmellr:
Bigger 'SimpleString API simplication': deletion? :)

clebertsuconic:
I wouldn't mind seeing SimpleString gone. Change the packets in Core to send char[] ... and be gone!

But that would require a lot of testing.. especially around the pools on reading?

clebertsuconic:
I don't really like the name "of" but that's just my personal taste probably... (of kind of kicks my OCD.. being a preposition, I feel like of what? :) 


but as I said.. that's my personal preference only...


but I don't want to get in the way.. I will get used to it :) and I'm +1 to merge it..  and you should probably merge it soon before you have to rebase it.

+1 from me.

gemmellr:
> I don't really like the name "of" but that's just my personal taste probably... (of kind of kicks my OCD.. being a preposition, I feel like of what? :)
> 
> but as I said.. that's my personal preference only...
> 
> but I don't want to get in the way.. I will get used to it :) and I'm +1 to merge it.. and you should probably merge it soon before you have to rebase it.
> 
> +1 from me.

I presume it comes from the trend of Map.of(..) etc factory helpers. I dont mind it, though I personally wouldnt have changed it here given I'd rather SimpleString vanish, along with all the realted conversions and the byte doubling etc etc that come with it.

tabish121:
> > I don't really like the name "of" but that's just my personal taste probably... (of kind of kicks my OCD.. being a preposition, I feel like of what? :)
> > but as I said.. that's my personal preference only...
> > but I don't want to get in the way.. I will get used to it :) and I'm +1 to merge it.. and you should probably merge it soon before you have to rebase it.
> > +1 from me.
> 
> I presume it comes from the trend of Map.of(..) etc factory helpers. I dont mind it, though I personally wouldnt have changed it here given I'd rather SimpleString vanish, along with all the realted conversions and the byte doubling etc etc that come with it.

Agreed with Robbie on the name, it matches some Collections APIs etc and isn't hard to grok plus its just less verbose.  But I'd also like to see an effort towards removing the SimpleString to simplify working with the broker internals.

jbertram:
As far as I'm aware, the "of" naming convention was introduced in 2008 in the second edition of Joshua Block's _Effective Java_. He identified it as a way to draw attention to static factories and adhere to common naming conventions. It is a more concise version of `valueOf` which is also used prolifically throughout the Java API. It's an industry standard at this point.

jbertram:
That said, if we can remove `SimpleString` completely I'm good with that too. Of course, this PR is not mutually exclusive with that objective. Consider it a stop-gap on the way.

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/SimpleString.java
```diff
@@ -56,20 +56,37 @@ public final class SimpleString implements CharSequence, Serializable, Comparabl
...
+
+   @Deprecated(forRemoval = true)
+   public static SimpleString toSimpleString(final String string) {
+      return of(string);
+   }
+
+   @Deprecated(forRemoval = true)
+   public static SimpleString toSimpleString(final String string, StringSimpleStringPool pool) {
+      return of(string, pool);
+   }
```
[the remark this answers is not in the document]
jbertram: I fleshed out the JavaDoc on both the old and new methods as well as deprecating the constructors.

clebertsuconic:
We can merge it I think.

gemmellr:
The change should probably be rebased to current main, which has already added new use of the now-deprecated bits that this misses switching over.

From doing a grep, there also looks to remain a number of other existing uses of toSimpleString that perhaps should be updated also.

Feels like we should be unit testing any otherwise-unused deprecated bits, but it doesnt appear like either of the two SimpleStringTest classes do so.

jbertram:
I rebased and changed as many of the remaining references to deprecated methods that I could. There are a handful of references to deprecated methods in the compatibility tests, but those must remain due to the fact that those tests are actually using older versions that don't have the new methods.

clebertsuconic:
lets merge and review? otherwise you will keep rebasing forever

gemmellr:
Same general review comment as last time. There are still several references to toSimpleString (outside compatibility tests). There is now-unused deprecated stuff that isn't tested.

PR #4960, opened 2024-06-05
ARTEMIS-4797 Destroy Netty connection if exception and inactive

joshb1050:
If an exception is caught before the channel is deemed inactive, the reference to the Netty connection will not get destroyed and will leak memory in `NettyAcceptor`. An exception is normal in the case of failover.

This commit adds a flag when an exception is encountered so that the connection is still destroyed in this case when the channel becomes inactive.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/ActiveMQChannelHandler.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/ActiveMQChannelHandler.java
@@ -41,6 +41,7 @@ public class ActiveMQChannelHandler extends ChannelDuplexHandler {
    private final BaseConnectionLifeCycleListener<?> listener;
 
    protected volatile boolean active;
+   protected volatile boolean exception;
 
    private final Executor listenerExecutor;
 
@@ -85,7 +86,7 @@ public class ActiveMQChannelHandler extends ChannelDuplexHandler {
    @Override
    public void channelInactive(final ChannelHandlerContext ctx) throws Exception {
       synchronized (this) {
-         if (active) {
+         if (active || exception) {
             listenerExecutor.execute(() -> listener.connectionDestroyed(channelId(ctx.channel()), true));
 
             active = false;
@@ -104,6 +105,7 @@ public class ActiveMQChannelHandler extends ChannelDuplexHandler {
       // and we don't want to spew out stack traces in that event
       // The user has access to this exeception anyway via the ActiveMQException initial cause
 
+      exception = true;
       ActiveMQException me = new ActiveMQException(cause.getMessage());
       me.initCause(cause);
 


joshb1050:
I am running this swapped into the environment in which this was originally detected and it has not appeared to leak any memory so far.

gemmellr:
Your Jira and PR comments are aimed at Acceptor behaviour, and you note on the Jira that you are using Openwire clients...however the changed bits are also used by the Artemis Core client. I'm not familiar with this code, but it seems to me like the change could result in the connectionDestroyed listener callback being called for clients when it wouldnt currently be, after the connectionException listener callback could already have been called, potentially causing a double-failover or other weirdness?

joshb1050:
@gemmellr That's possible...though I don't know how else to handle this. The `connectionDestroyed` listener needs to be called for these but they aren't, and we leak memory in the case where the exception callback is called first.

josh-byster:
Hi @gemmellr, any additional feedback here?

joshb1050:
Hi @gemmellr, any feedback here?

clebertsuconic:
Did you check against latest versions ?  Because I fixed issues around that. 


Also. A leak test would be good.

clebertsuconic:
@joshb1050 do you think you could modify this test to reproduce the same condition you have?

https://github.com/apache/activemq-artemis/blob/14c564a48144b0e052c6e0eeee64ba0f94744335/tests/leak-tests/src/test/java/org/apache/activemq/artemis/tests/leak/ConnectionDroppedLeakTest.java



with CheckLeak you can count the number of instances in memory and can verify if the leak still happening with a basic assertion.

clebertsuconic:
As Robbie mentioned, the same issue should apply to all the protocols. 

if you reproduce this issue for one, it would reproduce for all of them.


A memory leak test would verify the issue... it would be great to have it.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/ActiveMQChannelHandler.java
```diff
@@ -104,6 +105,7 @@ public void exceptionCaught(final ChannelHandlerContext ctx, final Throwable cau
       // and we don't want to spew out stack traces in that event
       // The user has access to this exeception anyway via the ActiveMQException initial cause
 
+      exception = true;
```
clebertsuconic, line 108: I downloaded the branch, and this is basically not effective.

exceptionCaught (the method where this is being set) will return if !active. Which means this change here is pointless.


that's why I think it's important to reproduce the issue on a Leak test.

First: because I had fixed a few open wire leaks.. if this is another leak, it would be better to first reproduce the issue.

clebertsuconic:
Perhaps you were facing ARTEMIS-4731 /  6fe53b813a0e958b73ea3e703bad7b0c892e6044 and thought your fix would address it?


I was looking at the code now and I'm not sure how an exception would be thrown before the activation of the connection.

We can try adding a Mock test

clebertsuconic:
@joshb1050  The more I think about this, the more I think you were just seeing ARTEMIS-4731. I will need you to confirm.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/netty/ActiveMQChannelHandler.java
gemmellr: I believe the point of the report was, this being callled while it _is_ still active, which further down this method then sets active=false, which then meant by the time it later called channelInactive that it didnt fire the connectionDestroyed listener. So this adds the exception=true flag for that case so that exceptionCaught firing whilst active records that, and then when channelInactive fired later it could still fire connectionDestroyed from that.

My earlier feedback was that while that seems reasonable enough at the server side, the Core client side also uses the same handler methods ,and there it looks less appropriate as those are both triggers for failover. Its probably the client usage that influenced the current setup where either exceptionCaught or channelInactive gets to handle 'problem while active' then setting active=false so the other one cant do anything.

clebertsuconic:
@joshb1050 ???

joshb1050:
This was happening on v2.32.0—do you expect the issue to have surfaced there? I can try it with v2.34.0 to see if it is resolved.

clebertsuconic:
Try 2.35

clebertsuconic:
@joshb1050 but yeah.. it was fixed into 2.34.0



I will close this for now.. if you still see an issue please reopen it? 

I will need a test for whatever is going on if you still see it.. The fix into 2.34.0 involved not ignoring an exception listener.. hence why I think this was the fix.

All other protocols were not ignoring the exception.

PR #4961, opened 2024-06-05
ARTEMIS-4798 Use original message to calculate the Mirror SNF size in calculations

clebertsuconic:


clebertsuconic:
there is a test failure I need to investigate.. a negative counter in the address size on BrokerInSyncTest::testLVQ

there could be something else in there..

PR #4962, opened 2024-06-06
NO-JIRA Implement CALL expression in filters

Hathoute:
An attempt to implement custom filters by using keyword CALL, as discussed in the [users mailbox](https://lists.apache.org/thread/8nfk874zf87jg1hrjn8m78x1kdt0j58n)

This PR is not intended to be merged, and only serves to discuss this feature request and potentially identify implementation details. For now I only implemented the parsing in StrictParser.

### DIFF
--- a/artemis-selector/src/main/java/org/apache/activemq/artemis/selector/filter/CallExpression.java
+++ b/artemis-selector/src/main/java/org/apache/activemq/artemis/selector/filter/CallExpression.java
@@ -1,0 +1,112 @@
+package org.apache.activemq.artemis.selector.filter;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class CallExpression implements BooleanExpression {
+   private final String functionName;
+   private final List<Object> parameters;
+   private final BooleanExpression expression;
+
+   CallExpression(final String functionName, final List<Object> parameters) {
+      this.functionName = functionName;
+      this.parameters = parameters;
+
+      try {
+         final var function = findFunction(functionName);
+         expression = extractExpression(function, parameters);
+      } catch (final Exception e) {
+         throw new IllegalStateException("Failed to create CallableExpression filter", e);
+      }
+   }
+
+   @Override
+   public Object evaluate(final Filterable message) throws FilterException {
+      return expression.evaluate(message);
+   }
+
+   @Override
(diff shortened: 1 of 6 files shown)


gemmellr:
It doesnt seem like a good idea to me to add something like this to the broker, it is a runtime-user-input-configurable RCE mechanism.

Hathoute:
Can't we mitigate this risk by having a config where we map a function name to the factory method, and validate methods at broker startup? I'm thinking of some sort of whitelist.

gemmellr:
Possibly, though then another concern might be around which users are allowed to use which methods, depending on what they do...then also what are allowed/safe arguments to those methods, depending on what they do. In all I still don't think this is a great idea to add, as presented.

clebertsuconic:
This is definitely not a good thing IMO.

it's already difficult to manage situations where the broker by itself has had issues in the past, now having the user being able to call stuff.. makes things even worse.

Say you filter Queue.messageCounter() on the filter...now you have a deadlock, starvation , performance issue... and there's no way to control or debug what is going on.


Even worse this is difficult to support and has security concerns.


I feel bad you made a very complex implementation on this.. but testing something like this gets even bigger and opens space for more problems in the end. I'm not sure how we would manage this.

Hathoute:
I do agree that this could be a bad idea, that's why I tried discussing it in the users mailbox. This PR is just an attempt I did before starting the discussion.

I believe it is the responsibility of the broker admin to verify that a custom filter is trusted before adding it, just like plugins, with the only difference that filters can receive external user input. Now if the filter does not correctly validate its input, that's another issue...

Also, existing filters (XPath for example) can also have a performance impact if not used correctly, yet any user is allowed to use them.

Again, I agree it might not be a good idea to move forward if we are not being extremely careful with the implementation, and perfectly implementing this will probably take too much effort, which in the end won't be worth it.

Hathoute:
For future people interested in this feature, you can hack? your way using just plugins.

I'll close the MR for now, but I will be happy to continue discussing this.

jbertram:
Generally speaking, I'm not opposed to having some sort of pluggable server-side filtering for queues, but I think it would be better to avoid reflection and instead create an interface that is implemented (e.g. as with existing plugins, transformers, etc.).

Also, it would be good to have a canonical use-case to help users (and developers) conceptualize how this ability might be used.

Hathoute:
Yes, the loading mecanism should be the same as what's been done for plugins. I just assumed it used reflection to instantiate plugin classes.

As for how this should be used, I would say also something similar to plugins; having jars in classpath and a broker config that declares the class names to load.

PR #4964, opened 2024-06-06
ARTEMIS-4800 simplify QueueConfiguration API

jbertram:

 - deprecate all QueueConfiguration ctors
 - add `of` static factory methods for all the deprecated ctors
 - replace any uses of the normal ctors with the `of` counterparts

This makes the code more concise and readable.


tabish121:
This looks good,I looked for awhile and didn't see any missed references.

PR #4966, opened 2024-06-07
ARTEMIS-4809 Allow configuring initial queue buffer size

joshb1050:
In some setups, there could be a few hundred thousand queues that are created due to many consumers that are connecting. However, most of these are empty and stay empty for the entire day since there aren't necessarily messages to be sent.  The 8K `intermediateMessageReferences` instantiates an `64KB` buffer (`Object[]`). This means we have large allocation and live heap that ultimately remains empty for almost the entire day.

In this commit, we introduce `initial-queue-buffer-size`, which defaults to the current value of `8192`. It can be set programmatically via
`QueueConfiguration#setInitialQueueBufferSize(int)`.

Note that this must be a power of 2.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQMessageBundle.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQMessageBundle.java
@@ -559,4 +559,6 @@ public interface ActiveMQMessageBundle {
    @Message(id = 229255, value = "Bridge {} cannot be {}. Current state: {}")
    ActiveMQIllegalStateException bridgeOperationCannotBeExecuted(String bridgeName, String failedOp, BridgeImpl.State currentState);
 
+   @Message(id = 229256, value = "{} must be a positive power of 2 (actual value: {})")
+   IllegalArgumentException positivePowerOfTwo(String name, Number val);
 }
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
@@ -533,6 +533,12 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
    static {
       metaBean.add(Integer.class, "queuePrefetch", (t, p) -> t.queuePrefetch = p, t -> t.queuePrefetch);
    }
+
+   static {
+      metaBean.add(Integer.class, "initialQueueBufferSize", (t, p) -> t.initialQueueBufferSize = p, t -> t.initialQueueBufferSize);
+   }
+   private Integer initialQueueBufferSize = null;
+
    //from amq5
    //make it transient
    @Deprecated
@@ -1283,6 +1289,15 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
       return this;
    }
 
+   public Integer getInitialQueueBufferSize() {
+      return initialQueueBufferSize;
+   }
+
+   public AddressSettings setInitialQueueBufferSize(int initialQueueBufferSize) {
+      this.initialQueueBufferSize = initialQueueBufferSize;
+      return this;
+   }
+
    /**
     * Merge two AddressSettings instances in one instance
     *
@@ -1783,6 +1798,9 @@ public class AddressSettings implements Mergeable<AddressSettings>, Serializable
          return false;
       if (!Objects.equals(idCacheSize, that.idCacheSize))
          return false;
+      if (!Objects.equals(initialQueueBufferSize, that.initialQueueBufferSize)) {
+         return false;
+      }
       return Objects.equals(queuePrefetch, that.queuePrefetch);
    }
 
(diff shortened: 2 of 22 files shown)


jbertram:
Thanks for the PR!

I think this would be better as an address-setting so you could easily configure large swaths of queues. Right now there's no way to configure this via `broker.xml`. It can only be configured if you're directly creating the queue(s) programmatically.

Lastly, there's no tests or documentation for this new setting. Tests are mandatory to ensure the functionality works properly and to mitigate future regressions.

joshb1050:
Thanks @jbertram! I have updated my PR to make this configurable on a per-address basis and via XML as well. I have also added tests for the functionality. Please let me know if there is a different approach I should be taking here.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -1604,6 +1619,10 @@ public void decode(ActiveMQBuffer buffer, boolean tryCompatible) {
          prefetchPageMessages = BufferHelper.readNullableInteger(buffer);
       }
 
+      if (buffer.readableBytes() > 0) {
+         intermediateMessageBufferInitialSize = BufferHelper.readNullableInteger(buffer);
+      }
+
```
jbertram, line 1625: Notice the comment below. This code should be removed.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -4457,6 +4457,15 @@
             </xsd:annotation>
          </xsd:element>
 
+         <xsd:element name="intermediate-message-buffer-initial-size" type="xsd:int" maxOccurs="1" minOccurs="0">
```
jbertram, line 4460: Specify the default value here as with the other elements.

joshb1050: This is done.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
joshb1050: Missed that—done now.

joshb1050:
Checkstyle is fixed now as well.

jbertram:
For what it's worth you can run checkstyle locally using:
```
mvn -Pdev install
```
See the [hacking guide](https://activemq.apache.org/components/artemis/documentation/hacking-guide/#using-the-dev-profile) for more details.

joshb1050:
@jbertram thank you—it's strange since I did run `mvn checkstyle:check` and `mvn -Pdev install` and both passed with the checkstyle error. It's also unclear given that `check` passed for 2 of the 3 failures. I have corrected the latest one that I see on CI though.

joshb1050:
Hi @jbertram, any feedback on this?

gemmellr:
> @jbertram thank you—it's strange since I did run `mvn checkstyle:check` and `mvn -Pdev install` and both passed even with the checkstyle error locally. It's also unclear given that `check` passed for 2 of the 3 failures. I have corrected the latest one that I see on CI though.

The last failure you had on 1 of the 3 'check' jobs, was actually from ErrorProne rather than Checkstyle, due to missing an @ Override. There is a different ErrorProne profile used on JDK 11+ builds and JDK16+ (i.e the JDK 17 and 21 CI runs), and that latter profile _also_ does not check the @ Override condition, because a Java 15 class signature change meant that one (or more, I forget) of the classes in the codebase actually violates that on Java 15+ but not on earlier versions (aside: I deliberately left 15 hanging when adapting the other profile for 16/17+, since it was out of support anyway).

This came up because in the past, the JDK > 11 runs were all using their own newer compile signatures, because we couldnt set the maven.compiler.release flag at the time. That is actually recently in place, so theyll all now be using the JDK11 method signatures now, hopefully meaning that issue shouldnt be present any more and I can adjust it to error on missing @ Override again in the JDK16+ profile too for consistency.

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
```diff
@@ -1604,6 +1619,7 @@ public void decode(ActiveMQBuffer buffer, boolean tryCompatible) {
          prefetchPageMessages = BufferHelper.readNullableInteger(buffer);
       }
 
+
```
gemmellr, line 1622: Superfluous newline

docs/user-manual/address-settings.adoc
```diff
@@ -394,6 +394,10 @@ that helps to detect and prevent the processing of duplicate messages based on t
 By default, the `id-cache-size` setting inherits from the global `id-cache-size`, with a default of `20000`
 elements if not explicitly configured. Read more about xref:duplicate-detection.adoc#configuring-the-duplicate-id-cache[duplicate id cache sizes].
 
+intermediate-message-buffer-initial-size::
+defines the initial number of elements of the intermediate message reference buffer (an `Object[]`), which is instantiated for each queue.
+By default, this value is `8192` bytes if not explicitly configured. This must be a positive power of 2 (i.e. `0` is not an option).
```
gemmellr, line 399: The units should be removed since it isnt in bytes: "value is `8192` _bytes_"

Dont think the array type needs listed, doesn't seem needed for the usage and just seems likely to go stale at some point.

docs/user-manual/configuration-index.adoc
```diff
@@ -798,6 +798,11 @@ see `auto-create-queues` & `auto-create-addresses`
 | Number of messages a management resource can browse
 | 200
 
+
+| xref:address-settings.adoc#address-settings[intermediate-message-buffer-initial-size]
+| The number of elements in the intermediate message buffer (`Object[]`) allocated for each queue
```
gemmellr, line 803: Ditto on listing the type.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQMessageBundle.java
```diff
@@ -551,4 +551,7 @@ IllegalStateException invalidRoutingTypeUpdate(String queueName,
 
    @Message(id = 229253, value = "Unable to acquire OperationContext when replicating packet: {}. ExecutorFactory: {}")
    IllegalStateException replicationFailureRepliTokenNull(String packet, String executorFactory);
+
+   @Message(id = 229254, value = "{} must be a power of 2 (actual value: {})")
```
gemmellr, line 555: This class is conflicting, presumably some other new message/id addition was made here so this one will need bumped accordingly.

gemmellr:
ErrorProne check on JDK16+ adjusted in https://github.com/apache/activemq-artemis/commit/6584220f27c51b00a591b9973544d9eefba7e628

artemis-server/src/main/java/org/apache/activemq/artemis/core/settings/impl/AddressSettings.java
joshb1050: Done

docs/user-manual/address-settings.adoc
joshb1050: Done

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQMessageBundle.java
joshb1050: Done

joshb1050:
Hi @gemmellr @jbertram can I please get a re-review on this?

gemmellr:
Spotted a few small niggles, nothing serious.

Would like Justin or someone else more familiar to also review.

jbertram:
Generally speaking I think this looks good. However, there's a couple of things I would revisit:

1. The name `intermediate-message-buffer-initial-size`, while technically accurate, is probably not terribly clear to users. I think something like `initial-queue-buffer-size` is more clear.
2. The documentation defines this as the " initial number of elements of the intermediate message reference buffer." Again, this is technically accurate, but it doesn't give users a clear reason _why_ they might want to configure it. The docs should make clear that this is tied directly to the number of _bytes_ from the JVM's heap each queue will consume initially. The docs might also mention the canonical use-case for this (i.e. lots of queues which might remain empty or have a small number of messages).

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/RandomUtil.java
```diff
@@ -60,6 +60,10 @@ public static int randomPositiveInt() {
       return Math.abs(RandomUtil.randomInt());
    }
 
+   public static int randomPowerOfTwo() {
+      return (int) Math.pow(2, RandomUtil.randomInterval(1, 10));
+   }
+
```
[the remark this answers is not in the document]
joshb1050: That makes sense to me—I have just inlined the one use case.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ValidatorsTest.java
```diff
@@ -152,4 +152,14 @@ public void testTWO_CHARACTERS() {
...
    }
 
+   @Test
+   public void testPOSITIVE_POWER_OF_TWO() {
+      ValidatorsTest.failure(Validators.POSITIVE_POWER_OF_TWO, 0);
+      ValidatorsTest.failure(Validators.POSITIVE_POWER_OF_TWO, -10);
+      ValidatorsTest.failure(Validators.POSITIVE_POWER_OF_TWO, 127);
+
+      ValidatorsTest.success(Validators.POSITIVE_POWER_OF_TWO, 2);
+      ValidatorsTest.success(Validators.POSITIVE_POWER_OF_TWO, 64);
```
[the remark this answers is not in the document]
joshb1050: Have added a few more tests here.

joshb1050:
@jbertram I agree, that makes sense. I have revised to `initial-queue-buffer-size` and changed usages, and also added additional context on when one might want to configure.

gemmellr:
LGTM

docs/user-manual/address-settings.adoc
```diff
@@ -394,6 +394,11 @@ that helps to detect and prevent the processing of duplicate messages based on t
 By default, the `id-cache-size` setting inherits from the global `id-cache-size`, with a default of `20000` elements if not explicitly configured.
 Read more about xref:duplicate-detection.adoc#configuring-the-duplicate-id-cache[duplicate id cache sizes].
 
+initial-queue-buffer-size::
```
jbertram, line 397: There should be an entry for `initial-queue-bufffer-size` in the example `address-setting` above. The value used in the example should be the default.

jbertram: I made this change myself during the merge.

jbertram:
@joshb1050, thanks for the contribution!

PR #4967, opened 2024-06-07
ARTEMIS-4796 add tests & remove refs

jbertram:


jbertram:
This is a follow-up from #4959 to address some comments from @gemmellr.

PR #4969, opened 2024-06-10
ARTEMIS-4811 upgrade Netty to 4.1.111.Final

ehsavoie:


jbertram:
Looks like we may need to wait for 4.1.111.Final in order to get https://github.com/netty/netty/issues/14080.

jbertram:
Actually 4.1.111.Final has already been released so just update your PR and this problem should go away.

PR #4972, opened 2024-06-12
ARTEMIS-4814 Make direct binding lookup time constant

joshb1050:
Currently, with 500K+ queues, the cleanup step of `TempQueueCleanerUpper` requires invoking `SimpleAddressManager#getDirectBindings`, which is O(k) in the number of queues `k`.

From method profiling, this can consume up to 95% of our CPU time when needing to clean up many of these.

Add a new map to keep track of the direct bindings, and add a test assertion that fails if we don't properly remove it.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SimpleAddressManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SimpleAddressManager.java
@@ -68,6 +68,8 @@ public class SimpleAddressManager implements AddressManager {
 
    private final ConcurrentMap<SimpleString, Pair<Binding, Address>> nameMap = new ConcurrentHashMap<>();
 
+   private final ConcurrentMap<SimpleString, Collection<Binding>> directBindingMap = new ConcurrentHashMap<>();
+
    private final BindingsFactory bindingsFactory;
 
    protected final MetricsManager metricsManager;
@@ -100,6 +102,11 @@ public class SimpleAddressManager implements AddressManager {
       if (nameMap.putIfAbsent(binding.getUniqueName(), bindingAddressPair) != null) {
          throw ActiveMQMessageBundle.BUNDLE.bindingAlreadyExists(binding);
       }
+      directBindingMap.compute(binding.getAddress(), (key, value) -> {
+         Collection<Binding> bindingList = value == null ? new ArrayList<>() : value;
+         bindingList.add(binding);
+         return bindingList;
+      });
 
       if (logger.isTraceEnabled()) {
          logger.trace("Adding binding {} with address = {}", binding, binding.getUniqueName(), new Exception("trace"));
@@ -116,7 +123,18 @@ public class SimpleAddressManager implements AddressManager {
          return null;
       }
 
-      removeBindingInternal(binding.getA().getAddress(), uniqueName);
+      SimpleString address = binding.getA().getAddress();
+      removeBindingInternal(address, uniqueName);
+      directBindingMap.compute(address, (key, value) -> {
+         if (value == null) {
+            return null;
+         }
+         value.remove(binding.getA());
+         if (value.isEmpty()) {
+            return null;
+         }
+         return value;
+      });
 
       return binding.getA();
    }
@@ -160,15 +178,7 @@ public class SimpleAddressManager implements AddressManager {
    @Override
    public Collection<Binding> getDirectBindings(final SimpleString address) throws Exception {
       SimpleString realAddress = CompositeAddress.extractAddressName(address);
-      Collection<Binding> bindings = new ArrayList<>();
-
-      nameMap.forEach((bindingUniqueName, bindingAddressPair) -> {
-         if (bindingAddressPair.getA().getAddress().equals(realAddress)) {
-            bindings.add(bindingAddressPair.getA());
-         }
-      });
-
-      return bindings;
+      return new ArrayList<>(directBindingMap.getOrDefault(realAddress, Collections.emptyList()));
    }
 
    @Override
--- a/tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/postoffice/impl/WildcardAddressManagerUnitTest.java
+++ b/tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/postoffice/impl/WildcardAddressManagerUnitTest.java
@@ -196,6 +196,7 @@ public class WildcardAddressManagerUnitTest extends ActiveMQTestBase {
 
       assertEquals(0, ad.getAddresses().size());
       assertEquals(0, ad.getBindings().count());
+      assertEquals(0, ad.getDirectBindings(SimpleString.of("Topic1.>")).size());
    }
 
    @Test
@@ -367,6 +368,8 @@ public class WildcardAddressManagerUnitTest extends ActiveMQTestBase {
 
                // publish again, read only
                ad.getBindingsForRoutingAddress(pubAddr);
+
+               ad.getDirectBindings(pubAddr);
 
             } catch (Exception e) {
                e.printStackTrace();


clebertsuconic:
there are a few test failures:

Test / org.apache.activemq.artemis.tests.integration.cli.AddressCommandTest.testForceDeleteAddressWhenExistsQueues
Test / org.apache.activemq.artemis.tests.integration.addressing.TwoWaysRemoveAddressTest.testDeadLock
Test / org.apache.activemq.artemis.tests.integration.ra.ActiveMQMessageHandlerTest.testSimpleMessageReceivedOnQueueManyMessagesAndInterruptTimeout
Test / org.objectweb.jtests.jms.conform.session.TopicSessionTest.testUnsubscribe
Test / org.objectweb.jtests.jms.conform.session.TopicSessionTest.testCreateSubscriber_1
Test / org.objectweb.jtests.jms.conform.session.TopicSessionTest.testCreateSubscriber_2
Test / org.objectweb.jtests.jms.conform.session.TopicSessionTest.testDurableSubscriber
Test / org.objectweb.jtests.jms.conform.session.TopicSessionTest.testCreateDurableSubscriber_1
Test / org.objectweb.jtests.jms.conform.session.TopicSessionTest.testCreateDurableSubscriber_2
Test / org.objectweb.jtests.jms.conform.session.UnifiedSessionTest.testCreateTemporaryTopicOnQueueSession
Test / org.objectweb.jtests.jms.conform.session.UnifiedSessionTest.testCreateQueueOnTopicSession
Test / org.objectweb.jtests.jms.conform.session.UnifiedSessionTest.testCreateBrowserOnTopicSession
Test / org.objectweb.jtests.jms.conform.session.UnifiedSessionTest.testUnsubscribeOnQueueSession
Test / org.objectweb.jtests.jms.conform.session.UnifiedSessionTest.testCreateTemporaryQueueOnTopicSession
Test / org.objectweb.jtests.jms.conform.session.UnifiedSessionTest.testCreateDurableSubscriberOnQueueSession
Test / org.objectweb.jtests.jms.conform.session.UnifiedSessionTest.testCreateDurableConnectionConsumerOnQueueConnection
Test / org.objectweb.jtests.jms.conform.session.UnifiedSessionTest.testCreateTopicOnQueueSession
Test / org.objectweb.jtests.jms.conform.topic.TemporaryTopicTest.testTemporaryTopic
Test / org.apache.activemq.artemis.tests.integration.amqp.connect.AMQPFederationAddressPolicyTest.testFederationHandlesAddressDeletedAndConsumerRecreates
Test / org.apache.activemq.artemis.tests.integration.amqp.connect.AMQPFederationAddressPolicyTest.testRemoteBrokerClosesFederationReceiverAfterAddressRemoved
Test / org.apache.activemq.artemis.tests.integration.cluster.distribution.SimpleSymmetricClusterTest.testDeleteAddress
Test / org.apache.activemq.artemis.tests.integration.cluster.distribution.SymmetricClusterWithDiscoveryTest.testStartStopServers


Why you need this optimization? Are you constantly querying for the destination on a producer or consumer?

clebertsuconic:
Why you need this optimization? are you querying the queue every time you produce?

joshb1050:
@clebertsuconic I believe I have fixed, though I wasn't able to see all of those test failures when running locally (and it shows me as the CI passing as well).

This is needed because each time one connection disconnects, say, with a few thousand consumers, the `TempQueueCleanerUpper` needs to clean each of these up, and it invokes `getDirectBindings` for each—so then this ends up being `O(k * n)` where `k` is the number of queues on the server and `n` is the number of consumers on the connection. It completely stops the server from functioning, and seems to be a blocker for any setup that has a few hundred thousand plus temp queues.

clebertsuconic:
@joshb1050 there's a complete test suite that you can run with -Ptests.

The CI in Jenkins runs a limited version of the test suite for basic validation. 

For more through verification I run the entire test suite but I wouldn't have reousrces for such a long process in the public CI from GitHub.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SimpleAddressManager.java
```diff
@@ -160,15 +172,7 @@ public Collection<Binding> getMatchingBindings(final SimpleString address) throw
...
-      Collection<Binding> bindings = new ArrayList<>();
-
-      nameMap.forEach((bindingUniqueName, bindingAddressPair) -> {
-         if (bindingAddressPair.getA().getAddress().equals(realAddress)) {
-            bindings.add(bindingAddressPair.getA());
-         }
-      });
-
-      return bindings;
+      return new ArrayList<>(directBindingMap.getOrDefault(realAddress, Collections.emptyList()));
```
tabish121, line 175: It seems rather odd to return a new ArrayList that's wrapping a potentially immutable empty list.  This should probably be made a bit smarter.

joshb1050: It's a copy operation, and works efficiently internally (at least in JDK 17). We could do a null check instead and return a new list if null, but this has to be a defensive copy since the tests will otherwise fail. 
```
    public ArrayList(Collection<? extends E> c) {
        Object[] a = c.toArray();
        if ((size = a.length) != 0) {
            if (c.getClass() == ArrayList.class) {
                elementData = a;
            } else {
                elementData = Arrays.copyOf(a, size, Object[].class);
            }
        } else {
            // replace with empty array.
            elementData = EMPTY_ELEMENTDATA;
        }
    }
```

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SimpleAddressManager.java
```diff
@@ -100,6 +102,8 @@ public boolean addBinding(final Binding binding) throws Exception {
       if (nameMap.putIfAbsent(binding.getUniqueName(), bindingAddressPair) != null) {
          throw ActiveMQMessageBundle.BUNDLE.bindingAlreadyExists(binding);
       }
+      directBindingMap.computeIfAbsent(binding.getAddress(), (unused) -> new ArrayList<>())
+              .add(binding);
```
clebertsuconic, line 106: you are using ConcurrentHashMap on  the directBindingsMap, however you're using a non Concurrent list on directBindingsMap. 

I'm wondering if we shouldn't use something more protective in multi-thread here. 

I understand it's unlikely the broker to remove or add a queue on the same address from multiple threads.. but it's not impossible. (say an application with a MultiCast address and applications creating and deleting subscriptions from multiple places).


it would be a rare race, but difficult to track if it ever happened. Shouldn't we use something more protective here.. perhaps CopyOnWrite or anything else?

joshb1050: Good point, definitely should be thread-safe. Changed to a COW ArrayList.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SimpleAddressManager.java
```diff
@@ -116,7 +121,15 @@ public Binding removeBinding(final SimpleString uniqueName, Transaction tx) thro
          return null;
       }
 
-      removeBindingInternal(binding.getA().getAddress(), uniqueName);
+      SimpleString address = binding.getA().getAddress();
+      removeBindingInternal(address, uniqueName);
+      Collection<Binding> directBindings = directBindingMap.get(address);
+      if (directBindings != null) {
+         directBindings.remove(binding.getA());
+         if (directBindings.isEmpty()) {
```
tabish121, line 129: Isn't there a race condition here in that the directBindings collection might report empty here but a new binding could be added between this check and the next call to remove the bindings entry from the Map meaning that the tracked state is now inconsistent with reality?

joshb1050: Should we perhaps put this in a `ConcurrentHashMap#compute` lambda instead?

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SimpleAddressManager.java
tabish121: Using a copy on write structure would seem to be adding a significant amount of GC overhead as each new binding will copy the array and toss out the old instance.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SimpleAddressManager.java
joshb1050: I have now used `compute` which should take advantage of the striped locking nature of the map, and should remove the race condition mentioned.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SimpleAddressManager.java
joshb1050: Now using an `ArrayList` given that the modifications are happening in a (synchronized) `compute` block.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SimpleAddressManager.java
```diff
@@ -100,6 +104,11 @@ public boolean addBinding(final Binding binding) throws Exception {
       if (nameMap.putIfAbsent(binding.getUniqueName(), bindingAddressPair) != null) {
          throw ActiveMQMessageBundle.BUNDLE.bindingAlreadyExists(binding);
       }
+      directBindingMap.compute(binding.getAddress(), (key, value) -> {
+         Collection<Binding> bindingList = value == null ? new ArrayList<>() : value;
+         bindingList.add(binding);
```
clebertsuconic, line 109: Maybe I'm being naive... but as far as I remember.. the ArrayList itself is not supposed to be used from multiple threads.

The compute will do a replay, but you could still call the compute from multiple threads.

I feel like you should have:

```
    synchronized(bindingList) {
        bindingsList.add(binding);
    }
```


Or another data structure to be used here.


similarly on the remove the queue, it should use a synchronized or use a concurrent equivalent.

clebertsuconic: say you call this from different threads... 


perhaps we should add a test having mulitple threads on this call here, and make sure the outcome is as expected.

joshb1050: The `directBindingMap` is a `ConcurrentHashMap` so it will guarantee thread safety on the `compute` by locking the node. https://stackoverflow.com/a/59601136

Also, this should be tested already via `WildcardAddressManagerUnitTest#testConcurrentCalls`.

tabish121:
There are several checkstyle violations that needs addressed here as well.

joshb1050:
> There are several checkstyle violations that needs addressed here as well.

Sorry, Checkstyle not working locally for me. Think I have fixed the issues now.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/SimpleAddressManager.java
clebertsuconic: Yeah.. I did some testing.. I thought the compute would do a loop retry only (I read the wrong code).


I added some testing.. and I changed your code with adding sleeps and everything worked.. I could see the wait to lock threads.


If you could rebase against main now that I added a test to validate this?


also, you can verify check style by doing this:

mvn -Pdev install -DskipTests=true -T2

clebertsuconic:
I think this can be merged.. I will leave it open for a 1 or 2 days if anyone else wants to give some feedback, but this is +1 from me.

clebertsuconic:
I'm actually going to merge this now... 

Thanks for this.

gemmellr:
The addition to and removals from the concurrent _map_ are clearly safe from each other, locked on the node of the map via the _compute_ calls.

However, the copying of the non-concurrent _list_ (in _getDirectBindings_) contained inside that map node, and the additions to / removal from that list done inside those compute calls, still seem quite able to happen concurrently with each other? The compute calls only protect the map directly, they dont fully protect the list inside it unless _all_ list usage occurs solely within those locked compute blocks (or other similarly protected map methods). In this case that isnt true, there is list usage outside of the compute blocks. The list copy would also need to be inside a similarly protected compute block to be protected by the other uses of compute during the add/remove updates.

joshb1050:
That's a good point @gemmellr, you could have the defensive copy operation throw a CME if it's happening during the removal. Perhaps we can make this a `ConcurrentLinkedQueue` instead? I would like to avoid a COW list based on the concern [above](https://github.com/apache/activemq-artemis/pull/4972#discussion_r1639034848). If so, I can separately create an issue and PR to change the data structure here.

clebertsuconic:
I'm adding https://github.com/apache/activemq-artemis/pull/4990

although my test on synchronization is not failing on this.

PR #4975, opened 2024-06-13
Supress the login credentials during pod start-up

StandByStormMetaMaster:
For troubleshooting the output of our pods are fed to kibana. People analyzing these logs should not have the ability to see these credentials hence the proposal to add a SUPPRESS_CREDENTIALS environment variable to suppress this.

### DIFF
--- a/artemis-docker/docker-run.sh
+++ b/artemis-docker/docker-run.sh
@@ -33,7 +33,11 @@ fi
 
 if ! [ -f ./etc/broker.xml ]; then
     CREATE_ARGUMENTS="--user ${ARTEMIS_USER} --password ${ARTEMIS_PASSWORD} --silent ${LOGIN_OPTION} ${EXTRA_ARGS}"
-    echo CREATE_ARGUMENTS=${CREATE_ARGUMENTS}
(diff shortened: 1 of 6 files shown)


artemis-docker/docker-run.sh
```diff
@@ -33,7 +33,11 @@ fi
 
 if ! [ -f ./etc/broker.xml ]; then
     CREATE_ARGUMENTS="--user ${ARTEMIS_USER} --password ${ARTEMIS_PASSWORD} --silent ${LOGIN_OPTION} ${EXTRA_ARGS}"
-    echo CREATE_ARGUMENTS=${CREATE_ARGUMENTS}
+    if [[ ${SUPPRESS_CREDENTIALS,,} == "true" ]]; then
+      echo CREATE_ARGUMENTS="--user XXXXX --password XXXXX --silent ${LOGIN_OPTION} ${EXTRA_ARGS}"
```
clebertsuconic, line 37: wouldn't be simpler to just not echo CREATE_ARGUMENTS?

just open a JIRA, and remove this line... no need to print that.

StandByStormMetaMaster: Yes, it would 👍

PR #4979, opened 2024-06-14
ARTEMIS-4818 Improve support for multiple broker plugins from broker properties

brusdev:


artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -958,6 +958,15 @@ public <T> T convert(Class<T> type, Object value) {
...
    }
 
+   private static boolean isClassProperty(String property) {
+      return property.endsWith(PROPERTY_CLASS_SUFFIX);
+   }
+
+   private static String extractPropertyClassName(String property) {
+      int propertyClassSuffixIndex = property.indexOf(PROPERTY_CLASS_SUFFIX);
+      return property.substring(0, propertyClassSuffixIndex);
+   }
```
gemmellr, line 968: These two are a bit inconsistent, in that the first one only checks for _endsWith_, but the second extracts only up to the _first_ instance found, which could be different. Would be better if they both operated on the end basis for consistency (also saves looking for the location index of a value you can already calculate).

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -3489,10 +3498,19 @@ private Object trackCollectionOrMap(String name, Object resolved, Object bean) {
 
       private Object findByNameProperty(String key, Collection collection) throws InvocationTargetException, IllegalAccessException, NoSuchMethodException {
          // locate on name property, may be a SimpleString
-         for (Object candidate : collection) {
-            Object candidateName = getProperty(candidate, "name");
-            if (candidateName != null && key.equals(candidateName.toString())) {
-               return candidate;
+         if (isClassProperty(key)) {
```
gemmellr, line 3501: The comment on the line above this should either be adjusted, or move down toward the code it applied to originally.

PR #4980, opened 2024-06-15
ARTEMIS-4819 Improve user list output in CLI

ViliusS:
Previously it was not so easy to parse `artemis user list` output. If one would want to do something in a bash script with every user he had to do something like this:

```
while read -r user; do something with `echo $user | sed 's/^"\(.*\)".*/\1/'`; done < ./artemis user list --url tcp://localhost:61616 --user admin --password pass
```

This patch splits user and roles with a space and removes quotes around the user. So now one could just do:
```
while read -r user roles; do something with $user; done < ./artemis user list --url tcp://localhost:61616 --user admin --password pass
```

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/user/ListUser.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/user/ListUser.java
@@ -45,7 +45,7 @@ public class ListUser extends UserAction {
     * @throws Exception if communication with the broker fails
     */
    private void list() throws Exception {
-      StringBuilder logMessage = new StringBuilder("--- \"user\"(roles) ---\n");
+      StringBuilder logMessage = new StringBuilder("--- user (roles) ---\n");
       int userCount = 0;
       final String[] result = new String[1];
 
@@ -62,7 +62,7 @@ public class ListUser extends UserAction {
       JsonArray array = JsonUtil.readJsonArray(result[0]);
       for (int arrayIndex = 0; arrayIndex < array.size(); arrayIndex++) {
          JsonObject object = array.getJsonObject(arrayIndex);
-         logMessage.append("\"").append(object.getString("username")).append("\"").append("(");
+         logMessage.append(object.getString("username")).append(" ").append("(");
          JsonArray roles = object.getJsonArray("roles");
          for (int i = 0; i < roles.size(); i++) {
             logMessage.append(roles.getString(i));
--- a/artemis-cli/src/test/java/org/apache/activemq/cli/test/ArtemisTest.java
+++ b/artemis-cli/src/test/java/org/apache/activemq/cli/test/ArtemisTest.java
@@ -469,7 +469,7 @@ public class ArtemisTest extends CliTestBase {
          logger.debug("output1:\n{}", result);
 
          //default only one user admin with role amq
-         assertTrue(result.contains("\"admin\"(amq)"));
+         assertTrue(result.contains("admin (amq)"));
          checkRole("admin", roleFile, basic, "amq");
 
          //add a simple user
@@ -487,8 +487,8 @@ public class ArtemisTest extends CliTestBase {
          result = context.getStdout();
          logger.debug("output2:\n{}", result);
 
-         assertTrue(result.contains("\"admin\"(amq)"));
-         assertTrue(result.contains("\"guest\"(admin)"));
+         assertTrue(result.contains("admin (amq)"));
+         assertTrue(result.contains("guest (admin)"));
 
          checkRole("guest", roleFile, basic, "admin");
          assertTrue(checkPassword("guest", "guest123", userFile, basic));
@@ -508,9 +508,9 @@ public class ArtemisTest extends CliTestBase {
          result = context.getStdout();
          logger.debug("output3:\n{}", result);
 
-         assertTrue(result.contains("\"admin\"(amq)"));
-         assertTrue(result.contains("\"guest\"(admin)"));
-         assertTrue(result.contains("\"scott\"(admin,operator)"));
+         assertTrue(result.contains("admin (amq)"));
+         assertTrue(result.contains("guest (admin)"));
+         assertTrue(result.contains("scott (admin,operator)"));
 
          checkRole("scott", roleFile, basic, "admin", "operator");
          assertTrue(checkPassword("scott", "tiger", userFile, basic));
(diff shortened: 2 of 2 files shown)


jbertram:
I'm not in favor of this change. If you want output that's formatted clearly and in a way that can be parsed by scripts then I recommend the [management API](https://activemq.apache.org/components/artemis/documentation/latest/management.html#management). In this can you could invoke the [`listUser` operation on `ActiveMQServerControl`](https://activemq.apache.org/components/artemis/documentation/javadocs/javadoc-latest/org/apache/activemq/artemis/api/core/management/ActiveMQServerControl.html#listUser(java.lang.String)), e.g. this command:
```
curl -s -H "Origin: http://localhost" -u myUser:myPass --header "Content-type: application/json" --request POST --data '{"operation": "listUser", "mbean": "org.apache.activemq.artemis:broker=\"0.0.0.0\"", "type": "exec", "arguments": [""] }' http://localhost:8161/console/jolokia | jq '.value | fromjson'
```
Might provide output like this:
```json
[
  {
    "username": "myUser1",
    "roles": [
      "myRole1"
    ]
  },
  {
    "username": "myUser2",
    "roles": [
      "myRole2",
      "myRole3"
    ]
  }
]
```

ViliusS:
While I agree that JSON could be suitable for programming languages which natively support it, I don't really think that it is in a spirit of Bash shell scripting. Parsing JSON would involve much more complicated steps, first parsing the output, then somehow storing parsed data in a temporary variable, and only then doing needed actions on users. It also assumes that jq or any other JSON tool is available, which is not necessary true.

Linux already has all the tool it needs for such trivial task and all of them use spaces as parsing markers, does parsing automatically, effectively directly from SDIN.

jbertram:
> I don't really think that it is in a spirit of Bash shell scripting.

I'm not heavy into Bash shell scripting so I may not be the best judge of it's "spirit," but the fact that the [`jq`](https://jqlang.github.io/jq/) command exists and appears to be popular indicates that lots of folks are dealing with JSON from the shell.

> It also assumes that jq or any other JSON tool is available, which is not necessary true.

That's true, but given the popularity of JSON I think that's a relatively safe assumption.

> If you are worried that proposed output is not clear or pretty, another way to go could be to add additional flag to output the list without decorations.

My chief concern is that we don't needlessly complicate something that was designed to be very simple. If you want to treat the output of the CLI as an API (i.e. to be used in scripts and other commands) then you can't simply change it between minor versions because you'll break anybody who is already relying on the existing output format. If you want to eventually replace the existing output you must first deprecate it and provide a new flag or command. Then when a new major version is released we can remove the deprecated item. You can see how a small change like this can grow into something much larger. Of course, that's not to mention the subjective evaluation of what is _clear_ or _pretty_.

Ultimately I think we should encourage folks to use tools the way they were designed to be used. In short, use the management API if you need an API and use the CLI for simple, human tasks. I recently [clarified the documentation](https://github.com/apache/activemq-artemis/blob/main/docs/user-manual/using-cli.adoc#command-line-interface) to this end.

clebertsuconic:
I just had an issue with ./artemis queue stat where some users complained to me about parsing the output. so I added --json


if json is no good for scripting, we could add some other format output for things like that.


We should do it in a way we are consistent throughout the CLI. doing it now in one way here, and on a different way on queue stat makes it confusing.

ViliusS:
> > I don't really think that it is in a spirit of Bash shell scripting.
> 
> I'm not heavy into Bash shell scripting so I may not be the best judge of it's "spirit," but the fact that the [`jq`](https://jqlang.github.io/jq/) command exists and appears to be popular indicates that lots of folks are dealing with JSON from the shell.
> 
> > It also assumes that jq or any other JSON tool is available, which is not necessary true.
> 
> That's true, but given the popularity of JSON I think that's a relatively safe assumption.
> 

I just checked and jq is not installed in RPM based and DEB based distributions by default. It's true that you can install it on standard VM (assuming you have root access), but it could be complicated to get these tools in let's say Kubernetes Job scripts which usually run on minimal busybox images, or even in Artemis docker images if it is not run with root permissions under Kubernetes.

> > If you are worried that proposed output is not clear or pretty, another way to go could be to add additional flag to output the list without decorations.
> 
> My chief concern is that we don't needlessly complicate something that was designed to be very simple. If you want to treat the output of the CLI as an API (i.e. to be used in scripts and other commands) then you can't simply change it between minor versions because you'll break anybody who is already relying on the existing output format. If you want to eventually replace the existing output you must first deprecate it and provide a new flag or command. Then when a new major version is released we can remove the deprecated item. You can see how a small change like this can grow into something much larger. Of course, that's not to mention the subjective evaluation of what is _clear_ or _pretty_.
> 
> Ultimately I think we should encourage folks to use tools the way they were designed to be used. In short, use the management API if you need an API and use the CLI for simple, human tasks. I recently [clarified the documentation](https://github.com/apache/activemq-artemis/blob/main/docs/user-manual/using-cli.adoc#command-line-interface) to this end.

I completely support that CLI should not be used as API. My own use cases were just a temporary cleanup scripts which I used once or twice on our testing environment.  The pros of the CLI is exactly that - it should be easy to put a one-liner fast for simple tasks without going into data structures, regular expressions and similar stuff. IMHO, my change makes it even easier to do that. 

Though I agree that it could break bash script of others which rely on the current output. I'm not sure what is the best way to avoid that though, and I see this has happened in the past couple of times so it won't be unexpected. I can add changes to Version upgrade documentation if that would help.

ViliusS:
> I just had an issue with ./artemis queue stat where some users complained to me about parsing the output. so I added --json
> 
> if json is no good for scripting, we could add some other format output for things like that.
> 
> We should do it in a way we are consistent throughout the CLI. doing it now in one way here, and on a different way on queue stat makes it confusing.

I completely support that CLI should consistent. IMHO, my changes brings `./artemis user list` output parsable the same way to what we already have in `./artemis address show`.

jbertram:
> I just checked and jq is not installed in RPM based and DEB based distributions by default.

FWIW, `jq` is provided by default in Ubuntu starting in 23.10. It's also available as a package in Alpine Linux so it should be relatively easy to install there as well.

That said, it's not clear to me that the availability of `jq` _on the broker's host_ is particularly important when most use-cases will involve management from a _remote_ machine. In any event, I don't see how getting `jq` in such environments will be any more difficult than getting ActiveMQ Artemis itself.

> Though I agree that it could break bash script of others which rely on the current output. I'm not sure what is the best way to avoid that though...

The best way to avoid that is to not change the output.

Ultimately this change is about a few characters which makes a _small_ difference in how one-line commands are run (per your example). This change is not significant enough to risk breaking existing users or adding the complexity of multiple output formats.

At this point I feel we're in danger of hitting the [law of triviality](https://en.wikipedia.org/wiki/Law_of_triviality) (a.k.a. bike shedding). It's probably best if we move forward with other work.

PR #4981, opened 2024-06-17
Update zookeeper to 3.9.2 to fix CVE-2024-23944

fredbalves:
Update zookeeper to 3.9.2 to fix CVE-2024-23944


jbertram:
Merged via https://issues.apache.org/jira/browse/ARTEMIS-4849.

PR #4988, opened 2024-06-19
ARTEMIS-4827 remove unreachable or redundant code

jbertram:


clebertsuconic:
This is following up #4972

PR #4989, opened 2024-06-20
ARTEMIS-4831: use surefire default behaviour around test failure

gemmellr:
Always use surefure default behaviour by default: fail module build when a test fails.

Anyone wanting to ignore failures so as to run all modules, can use the standard surefire prop to request that, e.g:
  mvn test -Dmaven.test.failure.ignore

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -219,6 +219,7 @@
       <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
       <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
+      <maven.test.failure.ignore>false</maven.test.failure.ignore>
       <maven.test.redirectTestOutputToFile>false</maven.test.redirectTestOutputToFile>
 
       <!--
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -861,9 +858,7 @@
                <configuration>
                   <forkCount>1</forkCount>
                   <reuseForks>true</reuseForks>
-                  <testFailureIgnore>${testFailureIgnore}</testFailureIgnore>
                   <runOrder>alphabetical</runOrder>
-                  <redirectTestOutputToFile>${maven.test.redirectTestOutputToFile}</redirectTestOutputToFile>
```
gemmellr, line 866: Not strictly related, but since both are using the standard surefire properties it is superfluous, so when removing the other configuration element I took this out too.

gtully:
sensible

pom.xml
```diff
@@ -219,6 +219,7 @@
       <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
       <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
+      <maven.test.failure.ignore>false</maven.test.failure.ignore>
       <maven.test.redirectTestOutputToFile>false</maven.test.redirectTestOutputToFile>
```
clebertsuconic, line 223: I would make this one true.. and anyone else actually requiring to not generate it, should make it false.

The other day I was having a hard time figuring out how to get the output of my tests... and I didn't know why it was not generated..


If I, working on this codebase daily, had a difficult time... imaging other people.

clebertsuconic: not related to your changes I know, but since you are there.. can you make it true please?

gemmellr: I can, and have always wanted to...that prop is only even there as I wanted to be able to, so I changed it so allow configuring with the standard prop because it was hard coded and thus not configurable at all originally: https://github.com/apache/activemq-artemis/commit/0cdd0c8bd9050e203799cc0cbfa3f1d055fb7b26

However, not in this PR, since it also needs adjustments made to CI jobs to either flip it back, or begin actually archiving the results files so you can ever see the output at all.

PR #4990, opened 2024-06-20
ARTEMIS-4814 Improve synchronization on SimpleAddressManager::getDirectBindings

clebertsuconic:


clebertsuconic:
This is following up https://github.com/apache/activemq-artemis/pull/4972

clebertsuconic:
@joshb1050 , @gemmellr  ^^

tabish121:
This looks more correct in terms of the concurrency issue the referenced commit adds.  The test would likely work unless it managed to do a get at the same moment the array list was resizing internally.  As an aside it feels like methods like this should be returning the result in immutable collection wrappers to enforce the idea that the result is a view only.

clebertsuconic:
@tabish121 I wrapped the result with   this:

` return Collections.unmodifiableCollection(outputList);`

gemmellr:
Looks good in terms of resolving the concurrency issue I mentioned on the earlier PR.

PR #4992, opened 2024-06-20
ARTEMIS-4829 use lambdas consistently

jbertram:
This commit uses lambdas or method references wherever possible. There are still a handful of places that appear like they could be changed but couldn't mainly because they use "this" and the meaning of "this" changes when using a lambda.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/process/ProcessBuilder.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/process/ProcessBuilder.java
@@ -29,17 +29,14 @@ public class ProcessBuilder {
    static ConcurrentHashSet<Process> processes = new ConcurrentHashSet<>();
 
    static {
-      Runtime.getRuntime().addShutdownHook(new Thread() {
-         @Override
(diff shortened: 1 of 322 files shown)


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/process/ProcessBuilder.java
```diff
@@ -29,17 +29,14 @@ public class ProcessBuilder {
...
-               //               if (p.isAlive())
-               {
-                  p.destroy();
-               }
+      Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+         for (Process p : processes) {
+            //               if (p.isAlive())
+            {
+               p.destroy();
             }
```
clebertsuconic, line 37: Since you are already there.. can you clean this up? just call p.destroy() and remove the if up there?

clebertsuconic: I actually already amended the commit, and pushed it.

so I'm closing this as I already sent... nice once @jbertram

PR #4996, opened 2024-06-26
ARTEMIS-4794 configure pending ack behavior for bridge

jbertram:
When a bridge is stopped it doesn't wait for pending send acknowledgements to arrive. However, when a bridge is paused it does wait. The behavior should be consistent and more importantly configurable. This commit implements these improvements and generally refactors BridgeImpl to clarify and simplify the code. In total, this commit includes the follow changes:

 - Removes the hard-coded 60-second timeout for pending acks when pausing the bridge and adds a new config parameter (i.e. "pending-ack-timeout").
 - Applies the new pending-ack-timeout when the bridge is stopped.
 - Updates existing and adds new logging messages for clarity.
 - De-duplicates code for sending bridge-related notifications.
 - Avoids converting bridge name to/from SimpleString.
 - Removes unnecessary comments.
 - Renames variables & functions for clarity.
 - Replaces the `started`, `stopping`, & `active` booleans with a single `state` variable which is an enum.
 - Adds `final` to a few variables that were functionally final.
 - Synchronizes `stop` & `pause` methods to add safety when invoked concurrently with `handle` (since both deal with `state` and execute runnables on the ordered executor).
 - Reorganizes and removes a few methods for clarity.
 - Relocates `connect` method directly into `ConnectRunnable` (mirroring the structure of the `StopRunnable` and `PauseRunnable`).
 - Eliminates unnecessary variables in `ConnectRunnable` and `ScheduledConnectRunnable`.
 - Adds test to verify pending ack timeout works as expected with both `stop` & `pause` with both regular and large messages.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
@@ -16,6 +16,7 @@
  */
 package org.apache.activemq.artemis.core.server.cluster.impl;
 
+import java.lang.invoke.MethodHandles;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
@@ -28,6 +29,7 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.activemq.artemis.api.core.ActiveMQAddressFullException;
+import org.apache.activemq.artemis.api.core.ActiveMQDisconnectedException;
 import org.apache.activemq.artemis.api.core.ActiveMQException;
 import org.apache.activemq.artemis.api.core.ActiveMQExceptionType;
 import org.apache.activemq.artemis.api.core.ActiveMQInterruptedException;
@@ -47,13 +49,13 @@ import org.apache.activemq.artemis.api.core.client.TopologyMember;
 import org.apache.activemq.artemis.api.core.management.CoreNotificationType;
 import org.apache.activemq.artemis.core.client.impl.ClientProducerCredits;
 import org.apache.activemq.artemis.core.client.impl.ClientProducerFlowCallback;
-import org.apache.activemq.artemis.core.client.impl.ClientSessionFactoryImpl;
 import org.apache.activemq.artemis.core.client.impl.ClientSessionFactoryInternal;
 import org.apache.activemq.artemis.core.client.impl.ClientSessionInternal;
 import org.apache.activemq.artemis.core.client.impl.ServerLocatorInternal;
 import org.apache.activemq.artemis.core.config.BridgeConfiguration;
 import org.apache.activemq.artemis.core.filter.Filter;
 import org.apache.activemq.artemis.core.filter.impl.FilterImpl;
+import org.apache.activemq.artemis.core.server.ActiveMQMessageBundle;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
 import org.apache.activemq.artemis.core.server.ActiveMQServerLogger;
 import org.apache.activemq.artemis.core.server.HandleStatus;
@@ -74,14 +76,12 @@ import org.apache.activemq.artemis.utils.UUID;
 import org.apache.activemq.artemis.utils.collections.TypedProperties;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import java.lang.invoke.MethodHandles;
-import org.apache.activemq.artemis.api.core.ActiveMQDisconnectedException;
-
-/**
- * A Core BridgeImpl
- */
 
 public class BridgeImpl implements Bridge, SessionFailureListener, SendAcknowledgementHandler, ReadyListener, ClientProducerFlowCallback {
+
+   public enum State {
+      STARTING, STARTED, PAUSING, PAUSED, STOPPING, STOPPED
+   }
 
    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
 
@@ -109,16 +109,12 @@ public class BridgeImpl implements Bridge, SessionFailureListener, SendAcknowled
 
    private boolean blockedOnFlowControl;
 
-   /**
-    * Used when there's a scheduled reconnection
-    */
-   protected ScheduledFuture<?> futureScheduledReconnection;
+   protected ScheduledFuture<?> scheduledReconnection;
 
    protected volatile ClientSessionInternal session;
 
    // on cases where sub-classes need a consumer
    protected volatile ClientSessionInternal sessionConsumer;
-
 
    // this will happen if a disconnect happened
    // upon reconnection we need to send the nodeUP back into the topology
@@ -132,11 +128,7 @@ public class BridgeImpl implements Bridge, SessionFailureListener, SendAcknowled
 
    private volatile ClientProducer producer;
 
-   private volatile boolean started;
-
-   private volatile boolean stopping = false;
-
-   private volatile boolean active;
+   private volatile State state = State.STOPPED;
 
    private boolean deliveringLargeMessage;
 
@@ -148,11 +140,11 @@ public class BridgeImpl implements Bridge, SessionFailureListener, SendAcknowled
 
    private boolean keepConnecting = true;
 
-   private ActiveMQServer server;
+   private final ActiveMQServer server;
 
    private final BridgeMetrics metrics = new BridgeMetrics();
 
-   private BridgeConfiguration configuration;
+   private final BridgeConfiguration configuration;
 
    public BridgeImpl(final ServerLocatorInternal serverLocator,
                      final BridgeConfiguration configuration,
@@ -185,11 +177,6 @@ public class BridgeImpl implements Bridge, SessionFailureListener, SendAcknowled
       this.server = server;
    }
 
-   /** For tests mainly */
-   public boolean isBlockedOnFlowControl() {
-      return blockedOnFlowControl;
-   }
-
    public static final byte[] getDuplicateBytes(final UUID nodeUUID, final long messageID) {
       byte[] bytes = new byte[24];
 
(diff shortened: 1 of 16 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
```diff
@@ -363,49 +350,35 @@ public Executor getExecutor() {
    }
 
    @Override
-   public void stop() throws Exception {
-      if (stopping) {
+   public synchronized void stop() throws Exception {
+      if (state == State.STOPPING || state == State.STOPPED || state == State.PAUSING) {
```
gemmellr, line 354: Seems like it would previosuly have tried to stop while already pausing (not saying it would have worked), while this rules out it trying...silently. Feels like throwing e.g ISE to alert of the issue (/ need to try again) might be better if its not actually going to do anything?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
```diff
@@ -467,7 +433,7 @@ public void sendFailed(Message message, Exception e) {
    public void sendAcknowledged(final Message message) {
       logger.trace("BridgeImpl::sendAcknowledged received confirmation for message {}", message);
 
-      if (active) {
+      if (state == State.STARTED || state == State.STOPPING || state == State.PAUSING) {
```
gemmellr, line 436: Given the variable is now volatile, might be nicer to read the value once to a local variable and operate on that in this likely active section?

I also wonder if, given the logging above this line, whether there should be some more logging to note when it entirely skips doing anything here.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
```diff
@@ -185,12 +176,7 @@ public BridgeImpl(final ServerLocatorInternal serverLocator,
       this.server = server;
    }
 
-   /** For tests mainly */
-   public boolean isBlockedOnFlowControl() {
-      return blockedOnFlowControl;
-   }
-
-   public static final byte[] getDuplicateBytes(final UUID nodeUUID, final long messageID) {
+   public static byte[] getDuplicateBytes(final UUID nodeUUID, final long messageID) {
```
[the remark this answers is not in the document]
jbertram: Nothing specific, just the general assumption that since `static` methods can't be overridden and since there's no apparent reason to prevent it from being hidden by a subclass then `final` is unnecessary.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
jbertram: Agreed. Will fix.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
jbertram: Agreed on both points.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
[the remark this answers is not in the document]
gemmellr: Preventing subclasses hiding statics (ick) doesnt actually seem like a bad thing which is why I wondered if there was a reason to alter the existing code when one of the comments was about adding final to other things.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
gemmellr: Looks like your changes went quite a bit further, also throwing if you start when starting/started, stop when stopping/stopped, or pause when pausing/paused.

Its not really _wrong_, but seems like it could lead to quite a few more exceptions being thrown, and perhaps confusion as to why they were vs previous. Personally I think its ok if those cases nooped, as its already in or heading toward the state being requested. Thats entirely different than what I commented on here, where it was silently ignoring a call to stop due to already pausing (rather than e.g queuing up the stop after the pausing operation) and thus leaving it silently reaching an entirely different state than requested.

jbertram: I made calls to start, stop, & pause idempotent if the state already matches (or will match). If the state doesn't match then an exception is thrown.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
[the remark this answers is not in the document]
jbertram: Restored `final`.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
```diff
@@ -251,22 +243,22 @@ public long sequentialID() {
...
    public synchronized void start() throws Exception {
-      if (started) {
-         return;
+      if (state == State.STARTING || state == State.STARTED || state == State.STOPPING || state == State.PAUSING) {
+         logger.debug("Bridge {} state is {}. Ignoring call to start.", configuration.getName(), state);
+         if (state == State.STOPPING || state == State.PAUSING) {
+            throw ActiveMQMessageBundle.BUNDLE.bridgeOperationCannotBeExecuted(configuration.getName(), "started", state);
+         } else {
+            return;
+         }
```
gemmellr, line 252: Probably better to read the state once and check, otherwise it could change between each of the checks and being read for the exception creation method reporting the state that caused it, e.g if it was pausing it might reach paused during the check+throw process and throw 'cant be started because paused'.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
```diff
@@ -363,49 +355,43 @@ public Executor getExecutor() {
...
-      if (stopping) {
-         return;
+   public synchronized void stop() throws Exception {
+      if (state == State.STOPPING || state == State.STOPPED || state == State.PAUSING) {
+         logger.debug("Bridge {} state is {}. Ignoring call to stop.", configuration.getName(), state);
+         if (state == State.PAUSING) {
+            throw ActiveMQMessageBundle.BUNDLE.bridgeOperationCannotBeExecuted(configuration.getName(), "stopped", state);
+         } else {
+            return;
+         }
```
gemmellr, line 365: Similar comment to start() around state check

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
```diff
@@ -363,49 +355,43 @@ public Executor getExecutor() {
...
-         } catch (Exception e) {
-            ActiveMQServerLogger.LOGGER.broadcastBridgeStoppedError(e);
+   @Override
+   public synchronized void pause() throws Exception {
+      if (state == State.STOPPING || state == State.STOPPED || state == State.PAUSING || state == State.PAUSED) {
+         logger.debug("Bridge {} state is {}. Ignoring call to pause.", configuration.getName(), state);
+         if (state == State.STOPPING || state == State.STOPPED) {
+            throw ActiveMQMessageBundle.BUNDLE.bridgeOperationCannotBeExecuted(configuration.getName(), "paused", state);
+         } else {
+            return;
```
gemmellr, line 386: Similar comment to start() around state check.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/bridge/BridgeTest.java
```diff
@@ -2084,6 +2085,108 @@ public void testManagementLeak() throws Exception {
...
+         ClientMessage message = session0.createMessage(true);
+         message.putIntProperty(propKey, i);
+         message.getBodyBuffer().writeBytes(bytes);
+         producer0.send(message);
+      }
+
+      session0.close();
+      sf0.close();
+
+      Wait.assertEquals(10L, () -> bridge.getMetrics().getMessagesPendingAcknowledgement(), 2000, 100);
```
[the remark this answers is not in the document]
jbertram: Done.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/bridge/BridgeTest.java
```diff
@@ -2084,6 +2085,108 @@ public void testManagementLeak() throws Exception {
...
+   public void testWaitForPendingAcksOnStopWithLargeMessages() throws Exception {
+      testPendingAcks(true, true);
+   }
+
+   @TestTemplate
+   public void testWaitForPendingAcksOnPauseWithLargeMessages() throws Exception {
+      testPendingAcks(false, true);
+   }
+
+   private void testPendingAcks(boolean stop, boolean large) throws Exception {
```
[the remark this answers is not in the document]
jbertram: Added a handful of new tests for this use-case.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
jbertram: Done.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
jbertram: Done.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/impl/BridgeImpl.java
jbertram: Done.

PR #5075, opened 2024-07-08
ARTEMIS-4909 use names that match original xml tags

erwindon:
see https://issues.apache.org/jira/browse/ARTEMIS-4909

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
@@ -4689,7 +4689,7 @@ public class ActiveMQServerImpl implements ActiveMQServer {
 
    private void deployReloadableConfigFromConfiguration() throws Exception {
       if (configurationReloadDeployed.compareAndSet(false, true)) {
-         ActiveMQServerLogger.LOGGER.reloadingConfiguration("security");
+         ActiveMQServerLogger.LOGGER.reloadingConfiguration("security settings");
          securityRepository.swap(configuration.getSecurityRoles().entrySet());
          recoverStoredSecuritySettings();
 


gemmellr:
This all seems to assume that the config is only coming from the XML tags, which actually isnt necessarily the case, so it feels like this should stay more general as it currently is.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/ActiveMQServerImpl.java
```diff
@@ -4689,11 +4689,11 @@ private static void setUnsetQueueParamsToDefaults(QueueConfiguration c) {
 
    private void deployReloadableConfigFromConfiguration() throws Exception {
       if (configurationReloadDeployed.compareAndSet(false, true)) {
-         ActiveMQServerLogger.LOGGER.reloadingConfiguration("security");
+         ActiveMQServerLogger.LOGGER.reloadingConfiguration("security-settings");
```
gemmellr, line 4692: I could see making this "security settings", i.e adding " settings" to make it consistent with the "address settings" one below (which would remain as it was), but I dont think they should strictly match the XML tags since that may in fact not be where the configuration is coming from.

jbertram:
Some more context here...

The broker currently supports configuration via XML, [properties](https://activemq.apache.org/components/artemis/documentation/latest/configuration-index.html#broker-properties), and the management API. JSON support was also [recently added](https://github.com/apache/activemq-artemis/pull/5120). All of these use some variant of the words "security," "address," and "settings" together, and I think it's reasonable for the log to use wording consistent with this. However, the wording in the log doesn't need to be exactly the same as the XML.

I recommend you keep `address settings` as it is in the log and just change `security` to `security settings`.

erwindon:
thank for your comments. I'll update the PR with Justins suggestion.

erwindon:
PR is updated (and rebased)

jbertram:
@erwindon, thanks!

PR #5078, opened 2024-07-09
ARTEMIS-4913 provide more details on error callbacks

jbertram:


tabish121:
This one needs some touch ups now as it has come conflicts

jbertram:
@tabish121, conflicts resolved.

gemmellr:
Some off the changes add file names to the error messages...can those message ever make it out to a client? Might not want to be passing filenames if it could?

tabish121:
> Some of the changes add file names to the error messages...can those message ever make it out to a client? Might not want to be passing filenames if it could?

It looks like some of the existing loggers already do add filenames for a quick scan.

jbertram:
I don't see any places where callbacks are sending these error messages back to clients. It's all for server-side logging as far as I can tell.

PR #5079, opened 2024-07-09
ARTEMIS-4910 fix divert config encoding

jbertram:
This commit fixes 4 distinct issues with divert configuration encoding:

 - The encoding size was calculated incorrectly in three ways:
   - Using 0 instead of `DataConstants.SIZE_NULL` when no transformer is defined resulting in a calculated size discrepancy of -1.
   - Using `DataConstants.INT` instead of `DataConstants.SIZE_INT` for the number of transformer properties resulting in a calculated size discrepancy of +2 when using a configuration with a transformer.
   - Using `BufferHelper.sizeOfNullableBoolean` instead of `DataConstants.SIZE_BOOLEAN` for the `exclusive` property resulting in a calculated discrepancy of +1.
 - Encoding was using `writeString` instead of `writeNullableString` for the name of the transformer class resulting in an actual buffer size discrepancy of -1.

### DIFF
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/core/config/DivertConfigurationEncodingTest.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/core/config/DivertConfigurationEncodingTest.java
@@ -1,0 +1,217 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.core.config;
+
+import java.util.Map;
+
+import org.apache.activemq.artemis.api.core.ActiveMQBuffer;
+import org.apache.activemq.artemis.api.core.ActiveMQBuffers;
+import org.apache.activemq.artemis.core.server.ComponentConfigurationRoutingType;
+import org.apache.activemq.artemis.utils.DataConstants;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertArrayEquals;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+public class DivertConfigurationEncodingTest {
+
+   @Test
+   public void testEncodedBytes() {
+      final String name = "aa";
+      final String address = "bb";
+      final String forwardingAddress = "cc";
+      final String routingName = "dd";
+      final boolean exclusive = true;
+      final String filterString = "ee";
+      final ComponentConfigurationRoutingType routingType = ComponentConfigurationRoutingType.ANYCAST;
+      final String transformerClass = "ff";
+      final String transformerKey = "gg";
+      final String transformerValue = "hh";
+      final TransformerConfiguration myDivertTransformer = new TransformerConfiguration(transformerClass);
+
+      myDivertTransformer.getProperties().put(transformerKey, transformerValue);
+
+      DivertConfiguration configuration = new DivertConfiguration()
+         .setName(name)
+         .setAddress(address)
+         .setForwardingAddress(forwardingAddress)
+         .setFilterString(filterString)
+         .setRoutingName(routingName)
+         .setExclusive(exclusive)
+         .setRoutingType(routingType)
+         .setTransformerConfiguration(myDivertTransformer);
+
+      int encodeSize = configuration.getEncodeSize();
+      ActiveMQBuffer data = ActiveMQBuffers.fixedBuffer(encodeSize);
+      configuration.encode(data);
+      assertEquals(encodeSize, data.writerIndex());
+
+      // name
+      byte[] read = new byte[9];
+      data.readBytes(read, 0, 9);
+      assertArrayEquals(new byte[] {DataConstants.NOT_NULL, 0, 0, 0, 2, 0, 97, 0, 97}, read);
+
+      // address
+      read = new byte[9];
+      data.readBytes(read, 0, 9);
+      assertArrayEquals(new byte[] {DataConstants.NOT_NULL, 0, 0, 0, 2, 0, 98, 0, 98}, read);
+
+      // forwardingAddress
+      read = new byte[9];
+      data.readBytes(read, 0, 9);
+      assertArrayEquals(new byte[] {DataConstants.NOT_NULL, 0, 0, 0, 2, 0, 99, 0, 99}, read);
+
+      // routingName
+      read = new byte[9];
+      data.readBytes(read, 0, 9);
+      assertArrayEquals(new byte[] {DataConstants.NOT_NULL, 0, 0, 0, 2, 0, 100, 0, 100}, read);
+
+      // exclusive
+      assertEquals((byte) -1, data.readByte());
+
+      // filterString
+      read = new byte[9];
+      data.readBytes(read, 0, 9);
+      assertArrayEquals(new byte[] {DataConstants.NOT_NULL, 0, 0, 0, 2, 0, 101, 0, 101}, read);
+
+      // routingType
+      assertEquals(ComponentConfigurationRoutingType.ANYCAST.getType(), data.readByte());
+
+      // transformerClassName
+      read = new byte[9];
+      data.readBytes(read, 0, 9);
+      assertArrayEquals(new byte[] {DataConstants.NOT_NULL, 0, 0, 0, 2, 0, 102, 0, 102}, read);
+
+      // number of transformer properties
+      read = new byte[4];
+      data.readBytes(read, 0, 4);
+      assertArrayEquals(new byte[] {0, 0, 0, 1}, read);
+
+      // transformer key name
+      read = new byte[9];
+      data.readBytes(read, 0, 9);
+      assertArrayEquals(new byte[] {DataConstants.NOT_NULL, 0, 0, 0, 2, 0, 103, 0, 103}, read);
+
+      // transformer key value
+      read = new byte[9];
+      data.readBytes(read, 0, 9);
+      assertArrayEquals(new byte[] {DataConstants.NOT_NULL, 0, 0, 0, 2, 0, 104, 0, 104}, read);
+
+      assertEquals(0, data.readableBytes());
+   }
+
+   @Test
+   public void testEncodeDecodeWithTransformer() {
+      TransformerConfiguration mytransformer = new TransformerConfiguration("myDivertTransformer");
+      mytransformer.getProperties().put("key1", "prop1");
+      mytransformer.getProperties().put("key2", "prop2");
+      mytransformer.getProperties().put("key3", "prop3");
+
+      DivertConfiguration configuration = new DivertConfiguration()
+         .setName("name")
+         .setAddress("address")
+         .setForwardingAddress("forward")
+         .setFilterString("foo='foo'")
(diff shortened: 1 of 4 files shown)


artemis-server/src/test/java/org/apache/activemq/artemis/core/config/DivertConfigurationEncodingTest.java
```diff
@@ -0,0 +1,47 @@
...
+      configuration.setExclusive(false);
+      configuration.setRoutingType(ComponentConfigurationRoutingType.ANYCAST);
+      TransformerConfiguration myDivertTransformer = new TransformerConfiguration("myDivertTransformer");
+      myDivertTransformer.getProperties().put("foo", "foo");
+      configuration.setTransformerConfiguration(myDivertTransformer);
+
+      int encodeSize = configuration.getEncodeSize();
+      ActiveMQBuffer data = ActiveMQBuffers.fixedBuffer(encodeSize);
+      configuration.encode(data);
+      assertEquals(encodeSize, data.writerIndex());
```
gemmellr, line 45: Asserting the actual encoding size is as expected, and the actual bytes written, would seem like a good idea for an encoding test. Had that been done previously the bugs fixed should have been caught. Meanwhile this test would still let similar problems be introduced, e.g if it suddenly didn't write the correct things, or anything at all, but said that, this test would still pass.

gemmellr:
Are there any compatibility implications to consider and either handle or document?

The JIRA looks to cover to an over-the-wire issue...any issues if e.g a rolling upgrade is tried with the different encoding behaviours? Should a particular rollout order be used?

Also, are these values persisted anywhere? If so is there any issue when an updated broker starts with an existing stored configuration value from an older version?

jbertram:
I don't believe there will be any compatibility issues. To be clear, the changes only impact how a divert configuration is _encoded_. Nothing is changing with how it is _decoded_. This will mainly impact new brokers encoding new configurations. Any old configuration being read from disk or from the wire by a new broker will continue to work as it always had. Furthermore, a new broker sending a config over the wire to an old broker will also work properly not least because now since it's being encoded correctly.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/DivertConfigurationEncodingTest.java
jbertram: I added some more tests. :+1:

gemmellr:
I find it awkward to reconcile the encoding having changed, with there being no consideration needed at all in terms of upgrade behaviour. Either during decoding of old persisted data, or data generated afresh by an old broker.  By 'continue to work as it always had', do you for example mean 'continue to behave incorrectly, until only data generated by new brokers exists'? If so, that would still seems worth documenting somewhere, e.g jsut the Jira.

E.g if the calculated size has changed, which apparently is enough to fix a bug over the wire for live brokers, couldnt that have similar effect during the access of old persisted data with the wrong encoded size by a newer broker? Or does it somehow not use that incorrect size of persisted data to decode it correctly, or refreshes it so it has the correct encoded size persisted, when it is a new broker reading it?

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/DivertConfigurationEncodingTest.java
gemmellr: The new tests are better, but still dont really cover the original comment.

The bug here was apparently that the value of getEncodeSize() was wrong. For a comprehensive encoder test I'd expect an assertion on the exact actual value for given inputs to check it is as expected, but there still isnt such an assertion. All the tests do is verify the number of bytes written equals that value returned, which it always should, but that doesnt actually check the value was correct to expectations.

Similarly with the actual encoded bytes, they are not directly asserted, only indirectly verified by using the decoder. The new tests decode those output bytes, and assert the resulting components are as expected, which is certainly much better than before. However that also doesnt strictly check that they were encoded properly to begin with; the encoder and decoder could have compensating bugs and both be incorrect. Or the wrong methods could be used to encode a value giving it an unexpected (but perhaps decodable as equivalent) value in the buffer.. You also fixed such an issue in the encoder in the PR, though in this case it seems like it would just have caused an NPE.

jbertram:
> I find it awkward to reconcile the encoding having changed, with there being no consideration needed at all in terms of upgrade behaviour.

The _main_ thing that changed here is the calculation of the encoding size, a calculation which isn't performed during the decoding process. 

The only actual encoding change was to fix a bug where encoding & decoding were using mismatched methods. Encoding was using `writeString` and decoding was using `readNullableString`. Technically speaking, this could have been fixed by changing the decoding to using `readString`, but that would have had implications for compatibility. Instead I changed the encoding to use `writeNullableString` which means that the data in the buffer is correct now and there should be no compatibility issues.

>  By 'continue to work as it always had', do you for example mean 'continue to behave incorrectly, until only data generated by new brokers exists'?

If you have an old broker sending broken data to a new broker the new broker will choke on it. I don't see a reasonable way to avoid that. To clarify, are you saying that this fact needs to be documented?

> ...if the calculated size has changed...couldnt that have similar effect during the access of old persisted data with the wrong encoded size by a newer broker?

The encoding size is calculated and written to disk (or wire) along with the record. The new calculation won't come into play when reading this old data. The new broker will read the data from disk just like the old broker did since there's been no changes to decoding.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/DivertConfigurationEncodingTest.java
jbertram: > The bug here was apparently that the value of getEncodeSize() was wrong.

That was _most_ of the bug, but not all. There was also a bug in the actual encoding (which used `writeString` instead of `writeNullableString`).

> For a comprehensive encoder test I'd expect an assertion on the exact actual value for given inputs to check it is as expected, but there still isnt such an assertion.

The test does, in fact, make an assertion on the actual value in question (i.e. the data in the buffer) to ensure it matches the calculated encoded size.

Are you suggesting the test essentially re-implement `o.a.a.a.c.c.DivertConfiguration#getEncodeSize` and compare that with the value that the "real" implementation returns?

> All the tests do is verify the number of bytes written equals that value returned, which it always should, but that doesnt actually check the value was correct to expectations.

I'm not following. If the code just tested something which "always should" be the case then it would never fail. However, the test does fail if you remove the fix(es). To be clear, the test is comparing the _calculated_ encoding size with the _actual_ encoded size of the data to ensure they are the same.

> Similarly with the actual encoded bytes, they are not directly asserted, only indirectly verified by using the decoder.

I'm not clear on how to "directly" assert the values of the bytes without essentially re-implementing the decoder or using static values which themselves are generated from existing code. Do you have any specific suggestions on how to do this?

FWIW, my goal for this PR was not to provide an exhaustive set of tests for divert encoding/decoding. I simply wanted to fix and verify the specific use-case from ARTEMIS-4910.

gemmellr:
> > I find it awkward to reconcile the encoding having changed, with there being no consideration needed at all in terms of upgrade behaviour.
> 
> The _main_ thing that changed here is the calculation of the encoding size, a calculation which isn't performed during the decoding process.
> 
> The only actual encoding change was to fix a bug where encoding & decoding were using mismatched methods. Encoding was using `writeString` and decoding was using `readNullableString`. Technically speaking, this could have been fixed by changing the decoding to using `readString`, but that would have had implications for compatibility. Instead I changed the encoding to use `writeNullableString` which means that the data in the buffer is correct now and there should be no compatibility issues.
> 
> > By 'continue to work as it always had', do you for example mean 'continue to behave incorrectly, until only data generated by new brokers exists'?
> 
> If you have an old broker sending broken data to a new broker the new broker will choke on it. I don't see a reasonable way to avoid that. To clarify, are you saying that this fact needs to be documented?
> 
> > ...if the calculated size has changed...couldnt that have similar effect during the access of old persisted data with the wrong encoded size by a newer broker?
> 
> The encoding size is calculated and written to disk (or wire) along with the record. The new calculation won't come into play when reading this old data. The new broker will read the data from disk just like the old broker did since there's been no changes to decoding.

Ok, so you have confirmed everything I thought would be the case. So yes, I think it is reasonable that is at least noted somewhere, even if just the JIRA. Someone who has hit this, upgrades their broker and has now a 'bug fixed' version, is likely to expect that completely resolves their issue. If it actually wont change anything until only data generated by new brokers is in play, it feels like that should be stated so the user can figure that out and learn what to do about it. E.g is it a case of deleting their store to remove the old persisted thing? Removing a specific config and recreating it? Other? I really dont know, hence the suggestion we should perhaps have a few word covering the scenario for anyone else that wouldnt immediately know.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/DivertConfigurationEncodingTest.java
gemmellr: > > The bug here was apparently that the value of getEncodeSize() was wrong.
> 
> That was _most_ of the bug, but not all. There was also a bug in the actual encoding (which used `writeString` instead of `writeNullableString`).
> 
> > For a comprehensive encoder test I'd expect an assertion on the exact actual value for given inputs to check it is as expected, but there still isnt such an assertion.
> 
> The test does, in fact, make an assertion on the actual value in question (i.e. the data in the buffer) to ensure it matches the calculated encoded size.
> 

It asserts it is the same as some other length value _it also doesnt really know is correct_. Checking the two values are equal does not validate that either value was actually correct. This PR is even fixing bugs on both sides of the comparison. 

> Are you suggesting the test essentially re-implement `o.a.a.a.c.c.DivertConfiguration#getEncodeSize` and compare that with the value that the "real" implementation returns?
> 

I am saying that for a given input we should know the exact encoded size and output expected and assert it in a test, but none of these tests really do that. Dont need to implement anything to do that (but could), just verify it.

The new additional tests at least check round-tripping gets back to the expected values, which is much improved, but still doesnt check the encoder and decoder are actually doing the correct thing, just that they 'both agree'. I could unwittingly change them in future in a way that 'breaks' them (/the output) but the tests wont detect because they dont really verify the output itself. I think it should be the first test that notices something changed unexpectedly.

> > All the tests do is verify the number of bytes written equals that value returned, which it always should, but that doesnt actually check the value was correct to expectations.
> 
> I'm not following. If the code just tested something which "always should" be the case then it would never fail. However, the test does fail if you remove the fix(es). To be clear, the test is comparing the _calculated_ encoding size with the _actual_ encoded size of the data to ensure they are the same.
> 

I meant it 'always should' in the sense that _if the size calculation and encoding code is all correct_, those values clearly should match. The code wasnt all correct (both sides being broken in some way), so unwinding the fix would be expected to _potentially_ make it fail. A mismatch between those two proves at least one side is incorrect, which is the minimal value in the assertion, but them being the same size does not prove either is correct. That needs verified separately.

> > Similarly with the actual encoded bytes, they are not directly asserted, only indirectly verified by using the decoder.
> 
> I'm not clear on how to "directly" assert the values of the bytes without essentially re-implementing the decoder or using static values which themselves are generated from existing code. Do you have any specific suggestions on how to do this?
> 

In this case I meant literally compare the bytes with what they should be, as we can tell what they should be.

One case would as you said be taking the output from the encoder, verify manually they are correct as-is, and then codifying the expected output into the test via assertion(s). Then in future if it the output changes unexpectedly, it would fail, and indicate exactly where and how the output changed.

Another route would be essentially 'reimplementing' as you say. We effectively do both for proton and qpid-jms. There are tests that verify known bytes for encoders and decoders testing using fixed output assertions or input bytes. There are also simpler alternative protocol implementations forming the test peers that can be run against the full client / engine output to exercise them and/or validate the bytes on the wire as changes are being made to the 'production' implementations.

In this case, verifying 'known expected bytes' seems the more suitable and simple approach.

> FWIW, my goal for this PR was not to provide an exhaustive set of tests for divert encoding/decoding. I simply wanted to fix and verify the specific use-case from [ARTEMIS-4910](https://issues.apache.org/jira/browse/ARTEMIS-4910).

A simple encoder that spits out expected bytes from a given input is some of the most easily comprehensively testable code we ever come across. Not adding appropriate testing when creating the code, or later fixing the code and clearly seeing it is missing such testing, is pretty much exactly how to repeatedly end up with situations like this one.

jbertram: I updated the PR. Let me know what you think.

gemmellr: Looks good. I expanded the newest test in https://github.com/apache/activemq-artemis/commit/59e7a3c01f to also cover the other 2 cases that had roundtrip-variant checks.

PR #5084, opened 2024-07-11
ARTEMIS-4920 strip leading forward slash from IP address

jbertram:


tabish121:
This seems to break a couple tests in the SlowConsumerTest where it appears they are still expecting the slash in front of the IP address in the test assertions.

jbertram:
@tabish121, should be fixed now. Thanks for the heads-up!

PR #5086, opened 2024-07-11
ARTEMIS-4922 remove spurious warn msg when OpenWire client disconnects

jbertram:


tabish121:
I tested this locally and the errors are gone now

PR #5087, opened 2024-07-12
ARTEMIS-4923 reduce synchronization in ManagementServiceImpl

jbertram:
The `ManagementService` is used by the broker to register and unregister components for management as well as send notifications. When the broker is busy dealing with new sessions and auto-creating queues, addresses, etc. there is a lot of contention.

To reduce synchronization and improve the service overall this commit does the following:

 - Remove `synchronized` from most methods. In most cases it's completely unnecessary because the methods are already using a thread-safe data-structure (e.g. `ConcurrentHashMap`) or more specific synchronization is already in place (e.g. on `mbeanServer`).
 - Adds new & clarifies existing logging.
 - Synchronizes `start` & `stop` methods and adds gates via `started`.
 - Simplifies the `sendNotification` method by synchronizing once rather than twice and performing legitimacy checks sooner.
 - Removing an unnecessary overload of the `registereQueue` method.

To be clear, there are no tests included with this commit as there should be no semantic changes. Existing tests should be sufficient to identify any regressions.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
@@ -26,7 +26,6 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
@@ -152,7 +151,7 @@ public class ManagementServiceImpl implements ManagementService {
 
    private final SimpleString managementAddress;
 
-   private boolean started = false;
+   private volatile boolean started = false;
 
    private final boolean messageCounterEnabled;
 
@@ -166,6 +165,7 @@ public class ManagementServiceImpl implements ManagementService {
 
    private final Pattern viewPermissionMatcher;
 
+   private final Set<ObjectName> registeredNames = new ConcurrentHashSet<>();
 
    public ManagementServiceImpl(final MBeanServer mbeanServer, final Configuration configuration) {
       this.mbeanServer = mbeanServer;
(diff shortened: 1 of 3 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
```diff
@@ -692,16 +644,16 @@ public SimpleString getManagementNotificationAddress() {
       return managementNotificationAddress;
    }
 
-   // ActiveMQComponent implementation -----------------------------
-
    @Override
-   public void start() throws Exception {
+   public synchronized void start() throws Exception {
+      if (started) {
```
tabish121, line 649: This is a lot less protection than before for concurrent calls to start since you aren't using an atomic operation to compare and set so you can get two concurrent start calls competing in here.  Would depend a lot on whether the call is idempotent.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
```diff
@@ -719,10 +671,18 @@ public void activated() {
             }
          }
       });
+
+      started = true;
    }
 
    @Override
    public synchronized void stop() throws Exception {
+      if (!started) {
```
tabish121, line 680: Similar to above this doesn't really prevent concurrent calls to stop for overlapping so this also needs to be idempotent.  There is also the concern that start / stop calls can run concurrently now which could leave this in an odd state if not handled.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/impl/ManagementServiceImpl.java
jbertram: Won't the fact that `start` is `synchronized` stop concurrent calls?

tabish121: You are right, I got glassy eyed reading all the other removes of synchronized and read that backwards.  Ignore me.

clebertsuconic:
@jbertram can you fix conflicts please? there's a conflict with another PR merged from you.

I wasn't 100% sure how to fix it.. if you could do it please?

jbertram:
@clebertsuconic, this should be ready to go.

PR #5092, opened 2024-07-16
ARTEMIS-4925 Fixing issue with mirroring and replication

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/replication/ReplicationEndpoint.java
```diff
@@ -831,8 +830,8 @@ private void handlePageEvent(final ReplicationPageEventMessage packet) throws Ex
    private void handlePageWrite(final ReplicationPageWriteMessage packet) throws Exception {
       PagedMessage pgdMessage = packet.getPagedMessage();
       pgdMessage.initMessage(storageManager);
-      Message msg = pgdMessage.getMessage();
-      Page page = getPage(msg.getAddressSimpleString(), packet.getPageNumber());
+      SimpleString address = packet.getAddress();
+      Page page = getPage(address, packet.getPageNumber());
```
jbertram, line 834: nitpick...could be just:
```java
Page page = getPage(packet.getAddress(), packet.getPageNumber());
```

clebertsuconic: thanks

PR #5102, opened 2024-07-19
ARTEMIS-4936 verify response correlationId when using Core client

jbertram:
If, for whatever reason, the response for a packet sent with blocking semantics is never returned it's possible that an async response received in the interventing time will be interpreted as the current response. This is because ChannelImpl does not verify the correlation ID set on the response packet when it is received.


jbertram:
FYI - the full test-suite looks good on this.

PR #5103, opened 2024-07-19
ARTEMIS-4939 allow configuring request header size for embedded web server

jbertram:


artemis-dto/src/main/java/org/apache/activemq/artemis/dto/WebServerDTO.java
```diff
@@ -107,6 +107,9 @@ public class WebServerDTO extends ComponentDTO {
    @XmlAttribute
    public Integer scanPeriod;
 
+   @XmlAttribute
+   public Integer requestHeaderSize;
```
tabish121, line 111: My only comment here is that this attribute name, while matching the HttpConfiguration type's value isn't exactly self documenting as it doesn't really convey it is a limit unlike the other options in this DTO.  I'd have gone with something like requestHeaderSizeLimit or maxRequestHeaderSize just so it conveys the intent more clearly.

jbertram: I actually had the same thought.

jbertram:
Don't merge this just yet. I think I'm going to add the ability to configure the max _response_ header size as well since large request headers may result in large response headers.

PR #5108, opened 2024-07-22
ARTEMIS-4941 Remove lazy update after application properties as it's no longer needed

clebertsuconic:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/paging/AmqpPagingTest.java
```diff
@@ -121,87 +121,4 @@ public void testPaging() throws Exception {
       connection.close();
    }
 
-
-   @TestTemplate
-   @Timeout(60)
-   public void testSizeCalculationsForApplicationProperties() throws Exception {
```
gtully, line 127: how does this test behave after the revert? I recall the original issue was negative memory usage on an address. Do we now avoid that by taking a guestimate of the size of unmarshalled content and just work with that?

PR #5109, opened 2024-07-22
ARTEMIS-4930 & ARTEMIS-4931

jbertram:


artemis-server/src/test/java/org/apache/activemq/artemis/core/config/BridgeConfigurationEncodingTest.java
```diff
@@ -0,0 +1,522 @@
...
+
+      // staticConnector count
+      read = new byte[4];
+      data.readBytes(read, 0, 4);
+      assertArrayEquals(new byte[] {0, 0, 0, 1}, read);
+
+      // static connector
+      read = new byte[9];
+      data.readBytes(read, 0, 9);
+      assertArrayEquals(new byte[] {DataConstants.NOT_NULL, 0, 0, 0, 2, 0, 105, 0, 105}, read);
```
gemmellr, line 294: Should probably also have test(s) without staticConnector, especially seeing as its size was incorrectly handled previously and is being fixed.

gemmellr (commented):
Committing a test 'refactor' separately which then breaks the tests by exposing a bug, then fixing it seperately, means the build will be broken forever at that point. That can be annoying when bisecting etc at a later date.

I'd suggest swapping the commits. Or else disabling the affected tests and re enabling them in the fixing commit.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/BridgeConfigurationEncodingTest.java
jbertram: Tests added.

jbertram:
I disabled the failing test in the first commit and enabled it again in the following commit.

PR #5110, opened 2024-07-23
ARTEMIS-4949 TimedBuffer.checkSync could fail on ClosedChannelException

clebertsuconic:


artemis-journal/src/main/java/org/apache/activemq/artemis/core/io/nio/NIOSequentialFile.java
```diff
@@ -387,7 +396,7 @@ public void blockingWriteDirect(ByteBuffer bytes, boolean sync, boolean releaseB
       internalWrite(bytes, sync, null, releaseBuffer);
    }
 
-   private synchronized void internalWrite(final ByteBuffer bytes,
+   private void internalWrite(final ByteBuffer bytes,
```
clebertsuconic, line 399: I didn't mean to remove this synchronized. adding it back.

PR #5113, opened 2024-07-24
ARTEMIS-4941 Keep AMQP memory estimate more consistent

clebertsuconic:
After the change I made on fb2a57f3ed25895681d3636e3f5fb9d9b9d0a053, I decided to keep the calculations simpler and more consistent.

For that reason I completely removed the application properties encoding and decoding out of the picture and I raised the memory estimate a little bit.

Notice that as I was looking to write a test validating if the encoding would be the same after decoding I found an AMQPMessageTest doing exactly that but asserting the opposite. Since the code change I am performing here I am changing that test.


artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
```diff
@@ -80,6 +80,8 @@ public interface Message {
    // The value is somewhat higher on 64 bit architectures, probably due to different alignment
    int memoryOffset = 352;
 
+   int ESTIMATE_DATA_WEIGHT = 2;
```
tabish121, line 83: I generally like to see values like this carry some amount of documentation about what they do and why they have the given value just so future maintainers don't have to spend time to come up with an educated guess.

clebertsuconic: sure.. thanks!

I also moved it to AMQPMessage.

clebertsuconic:
I'm gonig to revert to previous calculations (updating after decoding) however stop doing it for paged messages.. which is the issue I had at hand.


It will be better to manage user's expectations, especially about when the system enters into page mode.

PR #5117, opened 2024-07-25
ARTEMIS-4952 Use getObjectPropertyForFilter when applying filters

tabish121:
When counting messages with provided filters and grouping use the API getObjectPropertyForFilter which translates values from AMQP message annotations that are otherwise missed by the filters.


tabish121:
Ran this through the full test suite without any new failures

PR #5118, opened 2024-07-29
ARTEMIS-4959 moveMessages operation can move more messages than max messageCount

gaohoward:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -4484,6 +4484,10 @@ public Integer expectedHits() {
        * @throws Exception
        */
       public abstract boolean actMessage(Transaction tx, MessageReference ref) throws Exception;
+
+      public boolean reachExpectedHits(int currentHits) {
```
tabish121, line 4488: I would name it 'reachedExpectedHits' or my personal preference 'expectedHitsReached' to make the meaning more clear

gaohoward: Yeah I think expectedHitsReached is more consistent one. I'll update. Thanks Tim.

gemmellr:
LGTM

PR #5120, opened 2024-07-31
ARTEMIS-4955 Support broker properties from JSON files

brusdev:


gemmellr:
I think it could be nicer to use an extension=type setup to convey what type of config is being used, rather than using e.g broker.properties files for different types of config. Seems more obvious for now, and more open to addition later for other specific types at some point? E.g think of the log4j config as example.

brusdev:
@gemmellr your suggestion sounds good to me. I have just pushed a commit to implement that.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -559,38 +562,41 @@ public Configuration parseProperties(String fileUrlToProperties) throws Exceptio
...
+         if (file.getName().endsWith(".json")) {
+            brokerProperties.loadJson(reader);
+         } else {
+            brokerProperties.load(reader);
+         }
+      }
+
+      parsePrefixedProperties(this, file.getName(), brokerProperties, null);
+   }
+
```
lavocatt, line 599: @brusdev we don't have an openapi definition for the broker properties atm, but when we start to having one, we could easily add a validation step upon reading the file to ensure that any properties we're loading is valid according to the spec.

PR #5125, opened 2024-08-02
ARTEMIS-4969 Ensure AMQP senders can target FQQN if authorized

tabish121:
When using targeted FQQN permissions the AMQP sender needs to check that it can access not only the address but also the queue if sent an FQQN so that the security can validate if the sender has been granted directed access to the FQQN as a whole.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpTargetedFQQNSecurityTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpTargetedFQQNSecurityTest.java
@@ -1,0 +1,362 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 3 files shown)


jbertram:
* authorized

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpTargetedFQQNSecurityTest.java
```diff
@@ -0,0 +1,362 @@
...
+   }
+
+   @Test
+   public void testReceiver1CannotAttachAsSenderToEitherFQQN() throws Exception {
+      doTestReceiverCannotAttachAsSenderToEitherFQQN(FQQN_RECEIVER_1);
+   }
+
+   @Test
+   public void testReceiver2CannotAttachAsSenderToEitherFQQN() throws Exception {
+      doTestReceiverCannotAttachAsSenderToEitherFQQN(FQQN_RECEIVER_1);
```
gemmellr, line 306: Arg should be FQQN_RECEIVER_2

gemmellr (commented):
Looks good other than the test c&p error, feel free to update and merge.

PR #5130, opened 2024-08-05
ARTEMIS-4974 Dot not allow run in the Shell execution

clebertsuconic:


gemmellr:
Could it not just omit the _run_ command when in shell mode, then no warning would be needed? Much like it doesnt include the _shell_ command in shell mode.

clebertsuconic:
@gemmellr I was debating with myself between that and what I did... 

I decided for what I did as I thought it would be clearer for someone that used to do this. I was trying to be more compatible with the previous semantic, giving the user an alternative rather than just failing. 


If you are strong opinioned about omitting I can definitely do that, no big deal.

gemmellr:
I expect as many people will be confused by it being there, with no possibility of it working, as would be confused by it just not being there. Its a little awkward. 

Can commands be hidden like options can? Like when options are renamed. Might give a middle ground...seems to be not there but gives the warning if someone tries it anyway? Or can arbitrary non-command text perhaps be added to the help output, to note you need to use 'artemis run' to start a broker?

PR #5132, opened 2024-08-06
ARTEMIS-4974 Removing Run from Shell

clebertsuconic:


clebertsuconic:
I created a GitHub issue into pico cli for this:

https://github.com/remkop/picocli/issues/2331

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Run.java
```diff
@@ -45,7 +43,7 @@
 import picocli.CommandLine.Command;
 import picocli.CommandLine.Option;
 
-@Command(name = "run", description = "Run the broker.")
+@Command(name = "run", description = "Run the broker.", hidden = true)
```
gemmellr, line 46: If you are just removing run from the shell use case for now, shouldn't this hidden=true be omitted? Otherwise wont this annotation still take effect for the regular non-shell case?

PR #5136, opened 2024-08-08
ARTEMIS-4785 ARTEMIS-4702 Add profile and log4j2 files for non-run CLI commands

brusdev:
The run command uses the artemis.profile and log4j2.properties files while all other CLI commands use the artemis-utility.profile and log4j2-default.properties files.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Upgrade.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Upgrade.java
@@ -83,6 +83,9 @@ public class Upgrade extends InstallAbstract {
 
       final File bin = new File(directory, "bin");
       File etcFolder = new File(directory, etc);
+      File dataFolder = new File(directory, data);
+      File logFolder = new File(directory, LOG_DIRNAME);
+      File oomeDumpFile = new File(logFolder, OOM_DUMP_FILENAME);
 
       final File artemisCmdScript = new File(bin, Create.ARTEMIS_CMD);
       final File artemisScript = new File(bin, Create.ARTEMIS);
@@ -106,7 +109,7 @@ public class Upgrade extends InstallAbstract {
       }
 
       HashMap<String, String> filters = new HashMap<>();
-      Create.addScriptFilters(filters, getHome(), getInstance(), etcFolder, new File(getInstance(), "notUsed"), new File(getInstance(), "om-not-used.dmp"), javaMemory, getJavaOptions(), "NA");
+      Create.addScriptFilters(filters, getHome(), getInstance(), etcFolder, dataFolder, oomeDumpFile, javaMemory, getJavaOptions(), getJavaUtilityOptions(), "NA");
 
       if (IS_WINDOWS) {
          // recreating the service.exe and config in case we ever upgrade it
@@ -154,6 +157,24 @@ public class Upgrade extends InstallAbstract {
          write("etc/" + Create.ETC_ARTEMIS_PROFILE_CMD, artemisProfileCmdTmp, filters, false, false);
          upgradeJDK(context, JDK_PREFIX_WINDOWS, "", KEEPING_JVM_ARGUMENTS, artemisProfileCmdTmp, artemisProfileCmd, artemisProfileCmdBkp,
                     "set ARTEMIS_INSTANCE=\"", "set ARTEMIS_DATA_DIR=", "set ARTEMIS_ETC_DIR=", "set ARTEMIS_OOME_DUMP=", "set ARTEMIS_INSTANCE_URI=", "set ARTEMIS_INSTANCE_ETC_URI=");
+
+         File artemisUtilityProfileCmd = new File(etcFolder, Create.ETC_ARTEMIS_UTILITY_PROFILE_CMD);
+         File artemisUtilityProfileCmdTmp = new File(tmp, Create.ETC_ARTEMIS_UTILITY_PROFILE_CMD);
+         File artemisUtilityProfileCmdBkp = new File(etcBkp, Create.ETC_ARTEMIS_UTILITY_PROFILE_CMD);
+         if (artemisUtilityProfileCmd.exists()) {
+            write("etc/" + Create.ETC_ARTEMIS_UTILITY_PROFILE_CMD, artemisUtilityProfileCmdTmp, filters, false, false);
+            upgradeJDK(context, JDK_PREFIX_WINDOWS, "", KEEPING_JVM_ARGUMENTS, artemisUtilityProfileCmdTmp, artemisUtilityProfileCmd, artemisUtilityProfileCmdBkp,
+                       "set ARTEMIS_INSTANCE=\"", "set ARTEMIS_DATA_DIR=", "set ARTEMIS_ETC_DIR=", "set ARTEMIS_OOME_DUMP=", "set ARTEMIS_INSTANCE_URI=", "set ARTEMIS_INSTANCE_ETC_URI=");
+         } else {
+            if (data == null || data.equals("data")) {
+               dataFolder = getDATA(context, dataFolder, artemisProfileCmd, "set ARTEMIS_DATA_DIR=");
+
+               Create.addScriptFilters(filters, getHome(), getInstance(), etcFolder, dataFolder, oomeDumpFile, javaMemory, getJavaOptions(), getJavaUtilityOptions(), "NA");
+            }
+
+            context.out.println("Creating " + artemisUtilityProfileCmd);
+            write("etc/" + Create.ETC_ARTEMIS_UTILITY_PROFILE_CMD, artemisUtilityProfileCmd, filters, false, false);
+         }
       }
 
       if (IS_NIX) {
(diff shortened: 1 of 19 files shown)


docs/migration-guide/_configuration.adoc
```diff
@@ -24,11 +24,13 @@ The main configuration file is `etc/broker.xml`.
...
-The `etc/artemis.profile` file is similar to the `bin/env` file in ActiveMQ.
-Here you can configure environment variables for the broker, mostly regular JVM args related to SSL context, debugging, etc.
+The `etc/artemis.profile` and `etc/artemis-utility.profile` files are similar to the `bin/env` file in ActiveMQ.
+In the `etc/artemis.profile` you can configure environment variables for the broker started by executing the `run` command, mostly regular JVM args related to SSL context, debugging, etc.
+In the `etc/artemis-utility.profile` file you can configure environment variables for all CLI commands other than run, mostly regular JVM args related to SSL context, debugging, etc.
 
 There's not much difference in logging configuration between two brokers, so anyone familiar with Java logging systems in general will find herself at home here.
-The `etc/log4j-config.properties` file is where it's all configured.
+The `etc/log4j.properties` file is where it's all configured for the broker.
+The `etc/log4j-utility.properties` file is where it's all configured for all CLI commands other than run.
```
gemmellr, line 33: file names are slightly wrong (original already was), should be log4j**2**.properties and log4j**2**-utility.properties

docs/user-manual/logging.adoc
```diff
@@ -3,7 +3,7 @@
 :idseparator: -
 
 Apache ActiveMQ Artemis uses the https://www.slf4j.org/[SLF4J] logging facade for logging, with the broker assembly providing https://logging.apache.org/log4j/2.x/manual/[Log4J 2] as the logging implementation.
-This is configurable via the `log4j2.properties` file found in the broker instance `etc` directory, which is configured by default to log to both the console and to a file.
+When the broker is started by executing the `run` command, this is configurable via the `log4j2.properties` file found in the broker instance `etc` directory, which is configured by default to log to both the console and to a file. For the other CLI commands, this is configurable via the `log4j2-utility.properties` file found in the broker instance `etc` directory, which is configured by default to log only errors to the console.
```
gemmellr, line 6:
```suggestion
When the broker is started by executing the `run` command, this is configurable via the `log4j2.properties` file found in the broker instance `etc` directory, which is configured by default to log to both the console and to a file. For the other CLI commands, this is configurable via the `log4j2-utility.properties` file found in the broker instance `etc` directory, which is configured by default to log only errors to the console (in addition to the usual command output).
```

docs/user-manual/upgrading.adoc
```diff
@@ -55,8 +55,8 @@ cd $NEW_ARTEMIS_DOWNLOAD/bin/
 ./artemis upgrade PATH_TO_UPGRADING_INSTANCE
 ----
 
-The broker instance `bin/artemis` script and `etc/artemis.profile`(`artemis.cmd` and `artemis.cmd.profile` on Windows) will be updated to the new versions, setting its ARTEMIS_HOME to refer to the new broker version home path.
-The tool will also create the new `<instance>/etc/log4j2.properties` configuration file if needed (e.g if you are migrating from a version prior to 2.27.0), and remove the old `<instance>/etc/logging.properties` file if present.
+The broker instance scripts `bin/artemis`, `etc/artemis.profile` and `etc/artemis-utility.profile` (`artemis.cmd`, and `artemis.cmd.profile` on Windows) will be updated to the new versions, setting its ARTEMIS_HOME to refer to the new broker version home path.
```
gemmellr, line 58:
```suggestion
The broker instance script `bin/artemis` plus profiles `etc/artemis.profile` and `etc/artemis-utility.profile` (`artemis.cmd`, `artemis.cmd.profile`, and `artemis-utility.cmd.profile` on Windows) will be updated to the new versions, setting its ARTEMIS_HOME to refer to the new broker version home path.
```

docs/user-manual/versions.adoc
```diff
@@ -309,7 +324,7 @@ Most existing customisations to the old configuration files and scripts will be
 As such you should compare the old configuration files with the refreshed ones and then port any missing customisations you may have made as necessary.
 The upgrade command itself will copy the older files it changes to an `old-config-bkp.` folder within the instance directory.
 
-Similarly, if you had customised the old `logging.properties` file you may need to prepare analogous changes for the new `log4j2.properties` file.
+Similarly, if you had customised the old `logging.properties` file you may need to prepare analogous changes for the new `log4j2.properties` and `log4j2-utility.properties` files.
```
gemmellr, line 327: Not sure if this addition makes sense, as this section is about 2.27.0 which wouldnt have that file.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Upgrade.java
```diff
@@ -154,6 +154,17 @@ public Object run(ActionContext context) throws Exception {
          write("etc/" + Create.ETC_ARTEMIS_PROFILE_CMD, artemisProfileCmdTmp, filters, false, false);
          upgradeJDK(context, JDK_PREFIX_WINDOWS, "", KEEPING_JVM_ARGUMENTS, artemisProfileCmdTmp, artemisProfileCmd, artemisProfileCmdBkp,
                     "set ARTEMIS_INSTANCE=\"", "set ARTEMIS_DATA_DIR=", "set ARTEMIS_ETC_DIR=", "set ARTEMIS_OOME_DUMP=", "set ARTEMIS_INSTANCE_URI=", "set ARTEMIS_INSTANCE_ETC_URI=");
+
+         File artemisUtilityProfileCmd = new File(etcFolder, Create.ETC_ARTEMIS_UTILITY_PROFILE);
+         File artemisUtilityProfileCmdTmp = new File(tmp, Create.ETC_ARTEMIS_UTILITY_PROFILE);
+         File artemisUtilityProfileCmdBkp = new File(etcBkp, Create.ETC_ARTEMIS_UTILITY_PROFILE);
+         if (artemisUtilityProfileCmd.exists()) {
```
gemmellr, line 161: Wrong constant being used, this is the Windows segment but these are all for the Linux file

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Upgrade.java
```diff
@@ -348,6 +370,15 @@ private void upgradeLogging(ActionContext context, File etcFolder, File bkpFolde
             }
          }
       }
+
+      File newDefaultLogging = new File(etcFolder, Create.ETC_LOG4J2_UTILITY_PROPERTIES);
```
gemmellr, line 374:
```suggestion
      File newUtilityLogging = new File(etcFolder, Create.ETC_LOG4J2_UTILITY_PROPERTIES);
```

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/artemis-utility.profile
```diff
@@ -0,0 +1,33 @@
...
+# The logging config will need an URI
+# this will be encoded in case you use spaces or special characters
+# on your directory structure
+ARTEMIS_INSTANCE_URI='${artemis.instance.uri}'
+ARTEMIS_INSTANCE_ETC_URI='${artemis.instance.etc.uri}'
+
+# Java Opts
+if [ -z "$JAVA_ARGS" ]; then
+    JAVA_ARGS="-Dlog4j2.configurationFile=${ARTEMIS_INSTANCE_ETC_URI}log4j2-utility.properties ${java-opts}"
+fi
```
gemmellr, line 33: The trouble with using JAVA_ARGS for this directly is, it then wont have any effect for anyone that is already setting JAVA_ARGS now, unless they notice this was added and adjust their JAVA_ARGS to include it. It also means anyone setting JAVA_ARGS later has to set this too, when they probably dont want to influence the logging.

Between that, and people perhaps thinking they need to replace JAVA_ARGS to change this (if they dont know later definitions take priority and e.g use JAVA_ARGS_APPEND to set it 'again'), I think it would be worth considering having a specific variable for the logging config file selection to use (if set; the main profile wouldnt set it, this profile would) such that people could use it to override/set what the profile chooses to pass on to the main script. Much like the new profile picker variable. People could then still use JAVA_ARGS or JAVA_ARGS_APPEND to override it too if they want.

(As example, part of the docs you updated in this PR, suggested people to set JAVA_ARGS for a command, but doesnt include setting this, so it wouldnt use this new logging config in that case)

docs/user-manual/versions.adoc
brusdev: I agree with you, I'll revert this change.

docs/migration-guide/_configuration.adoc
brusdev: Good catch!

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Upgrade.java
brusdev: Good catch!

artemis-cli/src/main/resources/org/apache/activemq/artemis/cli/commands/etc/artemis-utility.profile
clebertsuconic: +1

brusdev: It sounds good, I'll add the variable `LOGGING_ARGS` before `JAVA_ARGS`

gemmellr:
Looks good overall, but I did spot an issue with handling of the data dir during upgrade. I've worked up a fix but still need to look into some testing changes, as I think the Upgrade tests should have some additional checks for the new additions, and in particular definitely need updated to ensure they cover the problematic path I found/fixed which they dont yet.

Will convert to draft for now.

gemmellr:
I added https://github.com/apache/activemq-artemis/pull/5136/commits/68be64fdb8b56f92378c86ce17d380082a6c6882 with additional tests to cover the issue I noticed, and a fix, such that upgrades that didnt have a utility profile previously, will create it referencing the data dir from the existing main profile rather than the default which might be different.

I'll now squash and rebase all the changes

gemmellr:
Now rebased to current main

brusdev:
@gemmellr good catch! Thanks for fixing the bug of the data dir.

PR #5138, opened 2024-08-09
ARTEMIS-4982 Cleanup AMQP large message files for rejected sends

tabish121:
When an incoming AMQP large message send is rejected the broker should delete the large message file as part of the reject handling.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpFlowControlFailDispositionTests.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpFlowControlFailDispositionTests.java
@@ -17,7 +17,6 @@
 package org.apache.activemq.artemis.tests.integration.amqp;
 
 import static org.junit.jupiter.api.Assertions.assertTrue;
-
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
 import org.apache.activemq.artemis.core.settings.impl.AddressFullMessagePolicy;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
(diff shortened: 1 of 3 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpFlowControlFailDispositionTests.java
```diff
@@ -105,4 +108,51 @@ public void testAddressFullDisposition() throws Exception {
...
+         connection.close();
+      }
+
+      validateNoFilesOnLargeDir(getLargeMessagesDir(), expectedRemainingLargeMessageFiles);
+   }
+
+   private AmqpMessage createAmqpLargeMessageWithNoBody() {
+      AmqpMessage message = new AmqpMessage();
+
+      byte[] payload = new byte[32 * 1024];
```
gemmellr, line 149: Would be good to tie this to the constant size so that it remains 'large' in the case anyone ever changed that config (e.g as you just did) without also updating this bit....test would still pass, but no longer cover what was intended.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AmqpSecurityTest.java
```diff
@@ -277,4 +286,79 @@ public void inspectOpenedResource(Sender sender) {
...
+         connection.close();
+      }
+
+      validateNoFilesOnLargeDir();
+   }
+
+   private AmqpMessage createAmqpLargeMessageWithNoBody() {
+      AmqpMessage message = new AmqpMessage();
+
+      byte[] payload = new byte[64 * 1024];
```
gemmellr, line 355: Similarly here

fortikeco:
To be honest, this code/solution software design does not look right, ownership is not clear + multiple places same code.

jbertram:
@fortikeco, this code was merged 3 weeks ago so the review is a bit late :)

If you have suggestions on to improve the code please send a PR referencing ARTEMIS-4982 in the commit message.

Thanks!

PR #5142, opened 2024-08-12
ARTEMIS-4985 message priority occasionally broken

jbertram:
The test in this commit was distilled down from a much more complex integration test that rarely reproduced the problem. It is short and sweet and reproduces the problem every time.

The solution involves additional logic on the hot path so it should be scrutinized before merging. However, similar logic already exists in the iterator's `remove()` method. I could not find a simpler way to resolve the problem.


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityLinkedListImpl.java
```diff
@@ -54,6 +54,9 @@ public PriorityLinkedListImpl(final int priorities, Comparator<E> comparator) {
    }
 
    private void checkHighest(final int priority) {
+      for (int i = levels.length - 1; i >= 0 && levels[i].size() == 0; i--) {
```
tabish121, line 57: Had a look at this and I think this loop is probably not needed if the check in the loop in the iterator remove uses the correct exit terms.  Currently in the remove it scan s for the next highest but it is checking using the wrong variable and always reading from the same collection (index) 

`         for (int i = index; i >= 0 && levels[index].size() == 0; i--) {
            highestPriority = i;
         }
`

When it should be using the loop variable 'i' so that it exists as soon as a non-empty collection is found.

jbertram: Great catch! You're absolutely right. I looked at that code a lot, and my eyes just skipped right over that. Fixed.

tabish121:
I ran the full test suite on this a couple times for a sanity check.

PR #5145, opened 2024-08-13
ARTEMIS-4986 Compatibility issue on Quorum Voting

clebertsuconic:


tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/realServer/RolledUpgradeTest.java
```diff
@@ -0,0 +1,295 @@
...
+      testRolledOnZipDistribution(TWO_THIRTY_ZIP, "apache-artemis-2.30.0");
+   }
+
+   @Test
+   public void testRolledUpgrade_2_35() throws Exception {
+      testRolledOnZipDistribution(TWO_THIRTY_FIVE_ZIP, "apache-artemis-2.35.0");
+   }
+
+   private void testRolledOnZipDistribution(String zip, String destination) throws Exception {
+      File twoThirtyZip = new File(zip);
```
gemmellr, line 176: Looks like variable name needs updated to be more general

tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/realServer/RolledUpgradeTest.java
```diff
@@ -0,0 +1,295 @@
...
+import org.apache.activemq.artemis.utils.SpawnedVMSupport;
+import org.apache.activemq.artemis.utils.Wait;
+import org.apache.activemq.artemis.utils.cli.helper.HelperBase;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Assumptions;
+import org.junit.jupiter.api.Test;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RolledUpgradeTest extends CompatBaseTest {
```
gemmellr, line 38: RollingUpgradeTest ?

tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/realServer/RolledUpgradeTest.java
```diff
@@ -0,0 +1,295 @@
...
+
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+   private static final String TWO_THIRTY_ZIP = "./target/old-releases/apache-artemis-2.30.0-bin.zip";
+   private static final String TWO_THIRTY_FIVE_ZIP = "./target/old-releases/apache-artemis-2.35.0-bin.zip";
+
+   //private static String sourceHome = HelperBase.getHome().getAbsolutePath();
```
gemmellr, line 45: Remove leftover?

tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/realServer/RolledUpgradeTest.java
```diff
@@ -0,0 +1,295 @@
...
+
+   private void testRolledOnZipDistribution(String zip, String destination) throws Exception {
+      File twoThirtyZip = new File(zip);
+      Assumptions.assumeTrue(twoThirtyZip.exists());
+      File distFolder = new File("./target/old-releases", destination);
+      unzipJava(twoThirtyZip, new File("./target/old-releases/"), distFolder);
+      testRolledUpgrade(distFolder);
+   }
+
+   public void testRolledUpgrade(File distribution) throws Exception {
```
gemmellr, line 183: private?

tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/realServer/RolledUpgradeTest.java
```diff
@@ -0,0 +1,295 @@
...
+      runAfter(managementLive1::close);
+      runAfter(managementLive2::close);
+
+      SimpleManagement managementBKP0 = new SimpleManagement("tcp://localhost:61619", null, null);
+      SimpleManagement managementBKP1 = new SimpleManagement("tcp://localhost:61620", null, null);
+      SimpleManagement managementBKP2 = new SimpleManagement("tcp://localhost:61621", null, null);
+
+      runAfter(managementLive0::close);
+      runAfter(managementLive1::close);
+      runAfter(managementLive2::close);
```
gemmellr, line 206: Should these reference the backups rather than the lives again?

tests/soak-tests/pom.xml
```diff
@@ -191,6 +191,8 @@
 
    </dependencies>
 
+
+
```
gemmellr, line 195: Seems like these additions can be omitted

tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/realServer/RolledUpgradeTest.java
```diff
@@ -0,0 +1,295 @@
...
+   }
+
+   @Test
+   public void testRolledUpgrade_2_35() throws Exception {
+      testRolledOnZipDistribution(TWO_THIRTY_FIVE_ZIP, "apache-artemis-2.35.0");
+   }
+
+   private void testRolledOnZipDistribution(String zip, String destination) throws Exception {
+      File twoThirtyZip = new File(zip);
+      Assumptions.assumeTrue(twoThirtyZip.exists());
```
gemmellr, line 177: Similar comment as before around dropping _Assumptions._ prefix

tests/compatibility-tests/pom.xml
```diff
@@ -265,6 +271,39 @@
...
+                     <execution>
+                        <id>snapshot-check</id>
+                        <phase>compile</phase>
+                        <goals>
+                           <goal>dependency-scan</goal>
+                        </goals>
+                        <configuration>
+                           <libList>
+                              <arg>org.apache.activemq:apache-artemis:zip:bin:2.30.0</arg>
+                              <arg>org.apache.activemq:apache-artemis:zip:bin:2.35.0</arg>
```
gemmellr, line 296: Out of interest, why 2.30.0 and 2.35.0 rather than 2.31.0 (one before the issue was introduced in 2.32.0) and 2.36.0 (the current) ?

tests/compatibility-tests/pom.xml
```diff
@@ -265,6 +271,39 @@
       </plugins>
    </build>
    <profiles>
+      <profile>
+         <id>compatibility-real-servers</id>
+         <properties>
+            <skipCompatibilityTests>false</skipCompatibilityTests>
```
gemmellr, line 277: This will also enable all the other tests to be picked up by surefire even though their deps may not have been prepared if the other profile isnt run...do they all handle that (I see the new test handles the reverse case with an assumptions check) or is it just the case that you _must_ run both profiles if you run the new one?

EDIT: the answer is that the existing tests do not handle it, you need both. Should we make it clearer (e.g a comment here) that they both need to be enabled? E.g instead of setting this true we could have the new profile enforce the skipCompatibilityTests prop is true, and indicate the other profile needs to be used if it isnt.

tests/compatibility-tests/pom.xml
```diff
@@ -265,6 +271,39 @@
       </plugins>
    </build>
    <profiles>
+      <profile>
+         <id>compatibility-real-servers</id>
```
gemmellr, line 275: What about _compatibility-tests-distribution_ so that it shares a nicer prefix with the other profile, and is more descriptive of the difference between them?

Also, with this having its own profile, its likely not going to actually run in any CI jobs currently, that intended?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/quorum/QuorumManager.java
```diff
@@ -348,19 +377,42 @@ public synchronized void voteComplete() {
...
+
+         // We first try the current packet with a medium timeout
+         replyMessage = (QuorumVoteReplyMessage) clusterChannel.sendBlocking(new QuorumVoteMessage(handler, vote), -1, PacketImpl.QUORUM_VOTE_REPLY, VOTE_RESPONSE_TIMEOUT, false);
+         logger.trace("This is the reply message from the current version = {}", replyMessage);
+
+         // if no response, we try the previous versions, with still a medium timeout
+         if (replyMessage == null && oldHandlerName != null) {
+            replyMessage = (QuorumVoteReplyMessage) clusterChannel.sendBlocking(new QuorumVoteMessage(oldHandlerName, vote), -1, PacketImpl.QUORUM_VOTE_REPLY, VOTE_RESPONSE_TIMEOUT, false);
+            logger.trace("This is the reply message from the older version = {}", replyMessage);
+         }
```
gemmellr, line 399: What happens if a reply actually does arrive, but only after these short timeouts (2sec vs usual 30sec)? Ignored?

Also, in future could it check for a newer wire version and avoid doing this for newer servers?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/quorum/QuorumVoteServerConnect.java
```diff
@@ -28,7 +28,12 @@
  */
 public class QuorumVoteServerConnect extends QuorumVote<ServerConnectVote, Boolean> {
 
+
+   /** NOTE: The following String is used to identify the targetNode implementation at other servers.
+    *        Renaming such string would cause incompatibility changes. */
    public static final SimpleString PRIMARY_FAILOVER_VOTE = SimpleString.of("PrimaryFailoverQuorumVote");
+   public static final SimpleString OLD_PRIMARY_FAILOVER_NOTE = SimpleString.of("LiveFailoverQuorumVote");
```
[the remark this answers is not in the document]
clebertsuconic: oops... thanks

tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/realServer/RolledUpgradeTest.java
```diff
@@ -0,0 +1,295 @@
...
+      parameters.add("--no-mqtt-acceptor");
+      parameters.add("--no-hornetq-acceptor");
+      parameters.add("--no-stomp-acceptor");
+      parameters.add(serverLocation.getAbsolutePath());
+
+      ProcessBuilder processBuilder = new ProcessBuilder();
+      processBuilder.command(parameters.toArray(new String[parameters.size()]));
+      Process process = processBuilder.start();
+      SpawnedVMSupport.spawnLoggers(null, null, "ArtemisCreate", true, true, process);
+      Assertions.assertTrue(process.waitFor(10, TimeUnit.SECONDS));
```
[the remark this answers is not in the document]
clebertsuconic: This is asserting the process waited and finished in 10 seconds.  Not sure what else I would assert for here?

tests/compatibility-tests/pom.xml
clebertsuconic: 2.30.0 is one release before the one that changed it.. 2.31.0 would work fine on the roll upgrade.


I can change it to 2.36.0... (I honestly don't know why I chose 2.35.0.. I actually meant the latest and at the time I wrote it, and I guess I didn't think well - confused numbers.)

tests/compatibility-tests/pom.xml
clebertsuconic: Yeah.. I'm removing that.. it was a mistake

tests/compatibility-tests/pom.xml
clebertsuconic: @gemmellr I certainly don't intend it to run on GitHub CI.

I plan to change our internal CI at Red Hat to use this profile.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/quorum/QuorumManager.java
clebertsuconic: We could stop looking for this on a newer wire version... for sure.

I actually thought about upgrading the wire now, and skip these short timeouts. I had it written at some point and I reverted it... 

(I thought these retries would been enough)


To have these reply Messages being delayed more than 2 seconds here it would mean things are really bad for the cluster anyway. a Retry with the same serverID seemed enough to me here.


I could upgrade the wire now and not have these checks on current versions. WDYT?

tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/realServer/RolledUpgradeTest.java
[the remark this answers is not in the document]
clebertsuconic: I guess I don't understand what you mean by prefix here?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/cluster/quorum/QuorumManager.java
clebertsuconic: I will bump the release to avoid these checks if dealing with a current release

tests/compatibility-tests/pom.xml
```diff
@@ -265,6 +271,58 @@
...
    </build>
    <profiles>
+      <profile>
+         <id>compatibility-tests-distribution</id>
+         <build>
+            <plugins>
+               <plugin>
+                  <groupId>org.apache.maven.plugins</groupId>
+                  <artifactId>maven-dependency-plugin</artifactId>
+                  <version>3.7.1</version>
```
gemmellr, line 281: This plugin version is already managed (to 3.7.1) and so should be omitted

tests/compatibility-tests/pom.xml
```diff
@@ -265,6 +271,58 @@
...
+                  <executions>
+                     <execution>
+                        <id>unpack-artemis</id>
```
gemmellr, line 284: perhaps add the version to make the ID unique (here and below)

tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/distribution/RollingUpgradeTest.java
```diff
@@ -0,0 +1,281 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
+
+public class RollingUpgradeTest extends CompatBaseTest {
+
+   private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+   private static final String TWO_THIRTY = "./target/old-releases/apache-artemis-2.30.0";
+   private static final String TWO_THIRTY_SIX = "./target/old-releases/apache-artemis-2.35.0";
```
gemmellr, line 44: The version in the value needs updated to 2.36.0 as well

tests/compatibility-tests/src/test/java/org/apache/activemq/artemis/tests/compatibility/distribution/CompatBaseTest.java
```diff
@@ -0,0 +1,75 @@
...
+import java.io.FileOutputStream;
+import java.lang.invoke.MethodHandles;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipInputStream;
+
+import org.apache.activemq.artemis.utils.RealServerTestBase;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CompatBaseTest extends RealServerTestBase {
```
gemmellr, line 31: If moving to use dependency::unpack can this be removed?

PR #5149, opened 2024-08-15
Artemis Console next gen based on hawtIO 4

andytaylor:
This is a PR to allow developers to build a variant of the latest Artemis Release with the new Artemis Console embedded, for testing. It will not currently merge and isnt intended to.

I have staged the 1st Artemis Console release candidate, 1.0.0, and this build will use the staged dependency when built.

I will make this PR a draft and update it for each Artemis Release until it is decided to merge into main.


artemis-console/pom.xml
```diff
@@ -21,40 +21,36 @@
...
-  <properties>
-    <activemq.basedir>${project.basedir}/../..</activemq.basedir>
-  </properties>
-
   <dependencies>
     <dependency>
-      <groupId>org.eclipse.jetty.toolchain</groupId>
-      <artifactId>jetty-servlet-api</artifactId>
+      <groupId>jakarta.servlet</groupId>
+      <artifactId>jakarta.servlet-api</artifactId>
```
gemmellr, line 37: This used the jetty dep before as we consolidated on it to avoid having effectively duplicate bits in the distribution and the build. This will cause that again. Not sure it needs changing other than adjusting to the newer jetty releases dep (jetty-jakarta-servlet-api).

artemis-console/pom.xml
```diff
@@ -21,40 +21,36 @@
...
-      <artifactId>hawtio-war</artifactId>
-      <version>${hawtio.version}</version>
+      <groupId>org.apache.activemq</groupId>
+      <artifactId>artemis-console-war</artifactId>
       <type>war</type>
     </dependency>
 
     <!-- lets mark dependencies from the WAR as provided to avoid jetty:run adding duplicates -->
-    <dependency>
+    <!--<dependency>
```
gemmellr, line 48: Can remove rather than comment out

artemis-pom/pom.xml
```diff
@@ -524,10 +524,10 @@
...
          </dependency>
          <dependency>
-            <groupId>org.eclipse.jetty.toolchain</groupId>
-            <artifactId>jetty-servlet-api</artifactId>
-            <version>${jetty-servlet-api.version}</version>
-            <!-- License: (Joint): Apache 2.0 or EPL 2.0 -->
+            <groupId>jakarta.servlet</groupId>
+            <artifactId>jakarta.servlet-api</artifactId>
+            <version>${jakarta.servlet-api.version}</version>
+            <!-- License: EPL 2.0 -->
```
gemmellr, line 530: Related to earlier comment, think it should be left at the jetty one (now jetty-jakarta-servlet-api to match the newer jetty).

artemis-web/pom.xml
```diff
@@ -59,8 +59,8 @@
          <artifactId>artemis-commons</artifactId>
       </dependency>
       <dependency>
-         <groupId>org.eclipse.jetty.toolchain</groupId>
-         <artifactId>jetty-servlet-api</artifactId>
+         <groupId>jakarta.servlet</groupId>
+         <artifactId>jakarta.servlet-api</artifactId>
```
gemmellr, line 63: Similarly

pom.xml
```diff
@@ -29,6 +29,14 @@
       <relativePath>org.apache:apache</relativePath>
    </parent>
 
+   <repositories>
+      <repository>
+         <name>Apache Staging Repository</name>
+         <id>apache.staging</id>
+         <url>https://repository.apache.org/content/groups/staging/</url>
+      </repository>
+   </repositories>
```
gemmellr, line 38: Would configure this for releases-only so that it isnt ever checked for snapshots (should anyone use this overall change against development commits at a later point, it likely will be)

Maybe add a TODO:delete comment to make super clear it is temp and easy to find later.

pom.xml
```diff
@@ -105,10 +114,9 @@
       <fuse.mqtt.client.version>1.16</fuse.mqtt.client.version>
       <caffeine.version>3.1.8</caffeine.version>
       <guava.version>33.2.1-jre</guava.version>
-      <hawtio.version>2.17.7</hawtio.version>
+      <hawtio.version>4.0.0-RC1</hawtio.version>
```
gemmellr, line 117: The console build is using 4.0.0 it appears, so seems like this should be updated to at least that? (There is also a 4.1.0 now)

tests/smoke-tests/src/main/filtered-resources/servers/linuxUpgradeETC/artemis.profile
```diff
@@ -33,7 +33,7 @@ ARTEMIS_INSTANCE_ETC_URI='file:${project.basedir}/target/classes/servers/linuxUp
 # Hawtio Properties
 # HAWTIO_ROLE define the user role or roles required to be able to login to the console. Multiple roles to allow can
 # be separated by a comma. Set to '*' or an empty value to disable role checking when Hawtio authenticates a user.
-HAWTIO_ROLE='amq'
+HAWTIO_ROLES='amq'
```
gemmellr, line 36: This (and other nearby configs) shouldnt change in this way, the aim is to check the upgrade takes this old config and updates it to output the new config, then check it for expectations directly and through comparing with a freshly created broker instance config. These changes just put the new config into the not-upgraded files.

gemmellr (commented):
Looked over some of it (not the tests) and noted some small things

jbertram:
Couple of things I noticed right off the bat...

1. When the broker starts with the new console it logs this `WARN`:
    ```
    WARN  [io.hawt.web.ServletHelpers] Couldn't find file on location: file:///home/jbertram/jboss/src/activemq-artemis/artemis-distribution/target/apache-artemis-2.36.0-bin/apache-artemis-2.36.0/bin/test/etc/hawtio-oidc.properties
    ```
2. It looks like there's supposed to be an image on the log-in screen, but it just displays the text "Artemis Console", i.e.:
    
    image
3. It appears there's a similar image rendering problem once you log-in, i.e.:

    image
4. The broker used to log the following when starting up with the previous console. It's worth keeping these, I think:
    ````
    INFO  [org.apache.activemq.artemis] AMQ241002: Artemis Jolokia REST API available at http://localhost:8161/console/jolokia
    INFO  [org.apache.activemq.artemis] AMQ241004: Artemis Console available at http://localhost:8161/console
    ```

gemmellr:
I dont recall either way about the first issue you listed (didnt spot it though if it was there), but I didnt notice the 2nd or 3rd. Seems like one of us possibly had something cached. I'll give it another build tomorrow and check. I did notice the 4th about the log messages but forgot to post it in my final summary comment.

jbertram:
@gemmellr, so I restarted the broker, opened the console again, and forced a complete reload and the images appeared properly. That takes care of issues 2, 3, & 5!

gemmellr:
> Couple of things I noticed right off the bat...
> 
>     1. When the broker starts with the new console it logs this `WARN`:
>        ```
>        WARN  [io.hawt.web.ServletHelpers] Couldn't find file on location: file:///path/to/activemq-artemis/artemis-distribution/target/apache-artemis-2.36.0-bin/apache-artemis-2.36.0/bin/test/etc/hawtio-oidc.properties
>        ```

I do see this one too. It is logging from a hawtio class, and seems like it will go away with hawtio 4.1.0 as they switched it to debug:
https://github.com/hawtio/hawtio/commit/406c6812f1ccb5555d0eab7ed932ea2b6d7b3189#diff-7dbf56158197c9c42c5cda82f9ce400be6b592378d509def40c0c98e7c823325

pom.xml
gemmellr: Or actually this seems to not be needed at all. It only appears used in a bit of commented out dep in artemis-console that I previously commented on, and in leftover old module poms that are no longer being built and should be deleted: `artemis-hawtio/pom.xml`, `artemis-hawtio/activemq-branding/pom.xml`, `artemis-hawtio/artemis-plugin/pom.xml`

clebertsuconic:
I saw the exception.. I couldn't get the web console to run.. was I supposed to do anything to be able to run it?

gemmellr:
> I saw the exception.. 

The warning log messaged from hawtio discussed before? Or Something else?

>I couldn't get the web console to run.. was I supposed to do anything to be able to run it?

Just build + start the broker. When you say you couldnt get it to run, what exactly did you encounter? An error loading the console page, or did you just not try it because it didnt log its URL (which hasnt changed) at startup? Per earlier discussion, it wasnt logging the messages as before, but was still there.

artemis-distribution/pom.xml
```diff
@@ -117,16 +117,6 @@
...
-      <dependency>
-         <groupId>org.apache.activemq</groupId>
-         <artifactId>activemq-branding</artifactId>
-         <type>war</type>
-      </dependency>
-      <dependency>
-         <groupId>org.apache.activemq</groupId>
-         <artifactId>artemis-plugin</artifactId>
-         <type>war</type>
-      </dependency>
```
gemmellr, line 129: This is as much a comment on the unaltered bit remaining, as the removals.

The binary distribution assembly has a number of entries in its LICENCE file (and files it references) to do with covering the console bits. Those files haven't been changed here, but it seems almost certain they will need to be. That is, `artemis-distribution/src/main/resources/licenses/bin/LICENSE` and files under dir `artemis-distribution/src/main/resources/licenses/bin/licenses/`.

gemmellr:
I added some commits fixing many of the code items I commented on earlier. I also switched it to the current SNAPSHOT since the staging repo was dropped when the vote was cancelled, so the build wouldn't work at all otherwise.

Remaining open points are the licence stuff and other things discussed in the comments, e.g startup log messages / warnings.

andytaylor:
which references do you refer to @jbertram ?
> 6. Can you change references strictly to "Artemis" to be "ActiveMQ Artemis"?

andytaylor:
closing and replaced by https://github.com/apache/activemq-artemis/pull/5245

PR #5152, opened 2024-08-16
ARTEMIS-4972 - Use Lock in start() for Binding Variables

Yashashree-Chopada:
The start() method should use a lock to access the expectedBindings variable to ensure thread safety, since other methods use the lock and interact with expectedBindings at the same time.

Using the lock in the start() method ensures that access to the expectedBindings variable is safe in a multi-threaded environment, preventing potential thread contention issues.

[ARTEMIS-4972](https://issues.apache.org/jira/browse/ARTEMIS-4972)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/group/impl/LocalGroupingHandler.java
```diff
@@ -322,11 +322,19 @@ public synchronized void start() throws Exception {
       if (started)
          return;
 
-      if (expectedBindings == null) {
-         // just in case the component is restarted
-         expectedBindings = new LinkedList<>();
+      try
+      {
+         lock.lock();
```
tabish121, line 327: The attempt to lock is normally placed outside the try as you don't want the finally to attempt to unlock a lock if the caller fails to acquire the lock in the first place.

jbertram:
There are also 6 checkstyle violations:
```
[INFO] There are 6 errors reported by Checkstyle 10.17.0 with /home/runner/work/activemq-artemis/activemq-artemis/activemq-artemis/artemis-server/../etc/checkstyle.xml ruleset.
Error:  src/main/java/org/apache/activemq/artemis/core/server/group/impl/LocalGroupingHandler.java:[326,7] (blocks) LeftCurly: '{' at column 7 should be on the previous line.
Error:  src/main/java/org/apache/activemq/artemis/core/server/group/impl/LocalGroupingHandler.java:[332] (regexp) RegexpSingleline: Line has trailing spaces.
Error:  src/main/java/org/apache/activemq/artemis/core/server/group/impl/LocalGroupingHandler.java:[332,7] (blocks) RightCurly: '}' at column 7 should be on the same line as the next part of a multi-block statement (one that directly contains multiple blocks: if/else-if/else, do/while or try/catch/finally).
Error:  src/main/java/org/apache/activemq/artemis/core/server/group/impl/LocalGroupingHandler.java:[333] (regexp) RegexpSingleline: Line has trailing spaces.
Error:  src/main/java/org/apache/activemq/artemis/core/server/group/impl/LocalGroupingHandler.java:[334,7] (blocks) LeftCurly: '{' at column 7 should be on the previous line.
Error:  src/main/java/org/apache/activemq/artemis/core/server/group/impl/LocalGroupingHandler.java:[337] (regexp) RegexpSingleline: Line has trailing spaces.
```
You can [run checkstyle yourself](https://activemq.apache.org/components/artemis/documentation/hacking-guide/#using-the-dev-profile) locally using:
```
mvn -Pdev install
```

PR #5156, opened 2024-08-19
ARTEMIS-4998 Fix Federation link close wrongly closing the connection

tabish121:
The federation sender links can react incorrectly when the source broker closes a receiver because the demand is gone and they can result in the remote broker closing its side of the connection causing the whole federation to need to be rebuilt. Handle the closure events correctly to prevent an unexpected close and rebuild.


tabish121:
I've run this through a few rounds of the full test suite in CI now

PR #5157, opened 2024-08-19
ARTEMIS-5000 stop running most STOMP tests twice

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/v12/StompPropertiesInterceptorTest.java
```diff
@@ -14,40 +14,30 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.activemq.artemis.tests.integration.stomp;
+package org.apache.activemq.artemis.tests.integration.stomp.v12;
```
gemmellr, line 17: Changing the test package has caused test failure due to not finding a nested interceptor class being configured with a string literal that wasnt similarly updated to adjust the package.

PR #5160, opened 2024-08-20
ARTEMIS-5002 AMQP producer not unblock if the disk space is freed

gaohoward:
(test adapted from the one on the Jira)

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
@@ -1096,6 +1096,7 @@ public class PagingStoreImpl implements PagingStore {
          if (isFull()) {
             if (runOnFailure && runWhenAvailable != null) {
                addToBlockList(runWhenAvailable, blockedCallback);
+               pagingManager.addBlockedStore(this);
             }
             return false;
          }
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/GlobalDiskFullTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/GlobalDiskFullTest.java
@@ -19,22 +19,54 @@ package org.apache.activemq.artemis.tests.integration.amqp;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.lang.invoke.MethodHandles;
 import java.net.URI;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.activemq.artemis.core.config.Configuration;
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
 import org.apache.activemq.artemis.core.server.files.FileStoreMonitor;
 import org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl;
+import org.apache.activemq.artemis.core.settings.impl.AddressFullMessagePolicy;
+import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
+import org.apache.activemq.artemis.tests.extensions.parameterized.Parameter;
+import org.apache.activemq.artemis.tests.extensions.parameterized.ParameterizedTestExtension;
+import org.apache.activemq.artemis.tests.extensions.parameterized.Parameters;
 import org.apache.activemq.transport.amqp.client.AmqpClient;
 import org.apache.activemq.transport.amqp.client.AmqpConnection;
 import org.apache.activemq.transport.amqp.client.AmqpMessage;
 import org.apache.activemq.transport.amqp.client.AmqpSender;
 import org.apache.activemq.transport.amqp.client.AmqpSession;
-import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestTemplate;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+@ExtendWith(ParameterizedTestExtension.class)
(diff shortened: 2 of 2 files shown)


gaohoward:
@clebertsuconic can you take a look of this PR see if the fix makes sense? Thanks!

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AMQPBlockingTest.java
```diff
@@ -0,0 +1,128 @@
...
+      ActiveMQServer server = this.createServer(true, true);
+      server.getConfiguration().getAcceptorConfigurations().clear();
+      server.getConfiguration().getAcceptorConfigurations().add(this.addAcceptorConfiguration(server, port));
+      server.getConfiguration().setName("localhost");
+      server.getConfiguration().setJournalDirectory(server.getConfiguration().getJournalDirectory() + port);
+      server.getConfiguration().setBindingsDirectory(server.getConfiguration().getBindingsDirectory() + port);
+      server.getConfiguration().setPagingDirectory(server.getConfiguration().getPagingDirectory() + port);
+      server.getConfiguration().setMessageExpiryScanPeriod(5000L);
+      AddressSettings addressSettings = new AddressSettings();
+      addressSettings.setAddressFullMessagePolicy(AddressFullMessagePolicy.FAIL);
```
gemmellr, line 47: The Jira indicates an issue with either policy of either FAIL or DROP...but the code change seems like it could only affect FAIL, because its behind a gate checking for FAIL? Is another test and possibly fix needed?

The test method code itself below for FAIL looks essentially identical to the existing GlobalDiskFullTest, which I presume is covering PAGE, so if the same works for DROP then perhaps they could all share the same overall base test, setting the specific mode covered via the constructor (either via ParameterizedTestExtension usage, or actual trivial subclasses of a base class)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AMQPBlockingTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+         Thread threadWithAnon = new Thread(() -> {
+            try {
+               final AmqpMessage message = new AmqpMessage();
+               message.setBytes(payload);
+               anonSender.setSendTimeout(-1);
+               message.setAddress(getQueueName());
+               anonSender.send(message);
+            } catch (Exception e) {
+               e.printStackTrace();
```
clebertsuconic, line 105: logger please

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AMQPBlockingTest.java
```diff
@@ -0,0 +1,128 @@
...
+            } finally {
+               sentAnon.countDown();
+            }
+         });
+
+         threadWithAnon.start();
```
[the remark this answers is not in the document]
clebertsuconic: +1 on the executor

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/AMQPBlockingTest.java
gaohoward: I'll add a DROP test and see how it behaves.

gaohoward: actually the test pass with DROP. The fix only affects FAIL. Looks like the DROP was handled correctly.

gaohoward:
@clebertsuconic @gemmellr I've updated the PR. Thanks!

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
```diff
@@ -1096,6 +1096,7 @@ public boolean checkMemory(boolean runOnFailure, Runnable runWhenAvailableParame
          if (isFull()) {
             if (runOnFailure && runWhenAvailable != null) {
                addToBlockList(runWhenAvailable, blockedCallback);
+               pagingManager.addBlockedStore(this);
```
gemmellr, line 1099: The use of this same call lower down the class is gated by `if (usingGlobalMaxSize || pagingManager.isDiskFull()) {``, does this need to be?

It is also surrounded by various other double-checks around racing being not-full again, it seems hard to reconcile that path needing it but this one not.

I cant quite wrap my head around this so think @clebertsuconic might need to chip in here.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/GlobalDiskFullFailPolicyTest.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+@ExtendWith(ParameterizedTestExtension.class)
+public class GlobalDiskFullFailPolicyTest extends GlobalDiskFullTest {
```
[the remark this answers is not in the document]
gaohoward: ah ok. will update. Thanks!

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagingStoreImpl.java
gaohoward: me neither. Need @clebertsuconic's expertise. :)

gaohoward: PR updated. The code path is really processing many conditions, not sure what's impact of that.

clebertsuconic:
actually, I sent a new PR for this: 

https://github.com/apache/activemq-artemis/pull/5502

PR #5161, opened 2024-08-20
ARTEMIS-5004 Clean up federation address consumer bindings proactively

tabish121:
When an address consumer explicitly closes or is closed we should remove the address binding for that consumer right away instead of waiting for possible configured auto delete as the demand is gone and we don't need to binding to stick around any longer.

### DIFF
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
@@ -96,6 +96,7 @@ import org.apache.activemq.artemis.core.config.amqpBrokerConnectivity.AMQPFedera
 import org.apache.activemq.artemis.core.server.ActiveMQServer;
 import org.apache.activemq.artemis.core.server.ComponentConfigurationRoutingType;
 import org.apache.activemq.artemis.core.server.Divert;
+import org.apache.activemq.artemis.core.server.QueueQueryResult;
 import org.apache.activemq.artemis.core.server.impl.AddressInfo;
 import org.apache.activemq.artemis.core.server.transformer.Transformer;
 import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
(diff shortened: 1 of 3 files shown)


tabish121:
I've run this through a few test iterations without issue

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
```diff
@@ -3882,6 +3883,300 @@ public void testRemoteFederationReceiverCloseWithErrorTerminateRemoteConnection(
       }
    }
 
+   @Test
+   @Timeout(20)
+   public void testRemoteReceiverClosedWhenDemandRemovedCleansUpAddressBinding() throws Exception {
+      server.start();
+      server.addAddressInfo(new AddressInfo(SimpleString.of("test"), RoutingType.MULTICAST));
```
gemmellr, line 3890: Rather than literal "test" could it use e.g getTestName()? Perhaps assigned to a variable to make following the address usage simpler.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
```diff
@@ -3882,6 +3883,300 @@ public void testRemoteFederationReceiverCloseWithErrorTerminateRemoteConnection(
...
+
+      final Map<String, Object> remoteSourceProperties = new HashMap<>();
+      remoteSourceProperties.put(ADDRESS_AUTO_DELETE, true);
+      remoteSourceProperties.put(ADDRESS_AUTO_DELETE_DELAY, 1_000L);
+      remoteSourceProperties.put(ADDRESS_AUTO_DELETE_MSG_COUNT, -1L);
```
gemmellr, line 3895: If the test is checking the binding is removed proactively...should it actually be setting it to auto-delete at all?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
```diff
@@ -3882,6 +3883,300 @@ public void testRemoteFederationReceiverCloseWithErrorTerminateRemoteConnection(
...
+
+      // Send a message to check that the federation binding holds onto sends while the remote is offline
+      // due to connectivity issues.
+      final ConnectionFactory factory = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + AMQP_PORT);
+
+      try (Connection connection = factory.createConnection()) {
+         final Session session = connection.createSession(Session.AUTO_ACKNOWLEDGE);
+         final MessageProducer producer = session.createProducer(session.createTopic("test"));
+
+         producer.send(session.createMessage());
```
gemmellr, line 4033: Would be good to make the payload specific/different than before, e.g "offline", so it can be asserted.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
```diff
@@ -3882,6 +3883,300 @@ public void testRemoteFederationReceiverCloseWithErrorTerminateRemoteConnection(
...
+         }
+
+         peer.waitForScriptToComplete(5, TimeUnit.SECONDS);
+         peer.close();
+
+         if (autoDelete) {
+            // Queue binding should eventually be auto deleted based on configuration
+            Wait.assertTrue(() -> server.bindingQuery(SimpleString.of("test")).getQueueNames().size() == 0, 5_000, 100);
+         } else {
+            // Should still be there as it wasn't marched as auto delete as previously validated.
```
gemmellr, line 4172: marked

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
```diff
@@ -3882,6 +3883,300 @@ public void testRemoteFederationReceiverCloseWithErrorTerminateRemoteConnection(
...
+         peer.waitForScriptToComplete(5, TimeUnit.SECONDS);
+         peer.close();
+
+         server.stop();
+      }
+   }
+
+   @Test
+   @Timeout(20)
+   public void testRemoteConnectionSuddenCloseLeaveAddressBindingIntact() throws Exception {
```
gemmellr, line 3964: Maybe SuddenDrop (more in line with the later tests) instead of SuddenClose, to help distinguish it wasnt closed in the protocol sense?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
tabish121: I prefer it the way it is, plus it falls inline with the remainder of the tests.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
tabish121: sure, will fix.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
tabish121: This is already tested by the wait for transfer and accept of the previously sent message

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
tabish121: sure will change

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPFederationAddressPolicyTest.java
gemmellr: I prefer tests use specific names as it makes debugging them later amongst a sea of logs easier, and finding cross-test pollutions easier, and helps avoid them having an impact to begin with. I prefer variables to repeated literals because you can follow their usage with IDEs and update them more easily if needed later.

PR #5179, opened 2024-08-27
ARTEMIS-4906 CentOS 7 is EOL so remove Docker file

jbertram:


gemmellr:
A grep for "centos" shows some remaining references to this (or older Java 11 or generic variants), e.g in the script for preparing the images, and some tests.

Seems like they should be going at the same time.

jbertram:
@gemmellr, those issues should be resolved now. Thanks for the review!

PR #5181, opened 2024-08-27
ARTEMIS-4834 support consuming messages forever with CLI

jbertram:
The test here is just a sanity check to ensure an input of `-1` is parsed correctly. There's no real way to functionally test that the consumer will wait forever.

I'm open to ideas about improving the test so please give feedback.


brusdev:
You could use the AssertionLoggerHandler after adding a logging line with the receiveTimeOut.

clebertsuconic:
@brusdev good idea.. the CLI does not use the loggers though.. but you could still capture System.out somehow.

jbertram:
I updated the consumer to log the timeout and updated the test to verify what's logged.

brusdev (approved):

PR #5193, opened 2024-08-30
ARTEMIS-4962 mitigate NPE in FilterImpl

jbertram:


gemmellr:
Change seemed ok but needed testing. I pushed another commit adding one, see what you think.

gemmellr:
OK, my crime against spacing should hopefully be resolved now.

jbertram:
You want me to squash these commits or just merge them as-is?

gemmellr:
I'm fine either way, I just left it separate so it was easy to spot.

PR #5194, opened 2024-08-30
ARTEMIS-4965 mitigate NPE when missing SQL property

jbertram:


artemis-jdbc-store/src/test/java/org/apache/activemq/artemis/jdbc/store/sql/PropertySQLProviderTest.java
```diff
@@ -0,0 +1,59 @@
...
+         }
+      }
+   }
+
+   @AfterEach
+   public void shutdownDerby() {
+      try {
+         DriverManager.getConnection("jdbc:derby:;shutdown=true");
+      } catch (Exception ignored) {
+      }
```
gemmellr, line 56: Probably nicer just to call DBSupportUtil._shutdownDerby_, which also e.g verifies the shutdown exception is as expected.

artemis-jdbc-store/src/test/java/org/apache/activemq/artemis/jdbc/store/sql/PropertySQLProviderTest.java
```diff
@@ -0,0 +1,59 @@
...
+         PropertySQLProvider factory = (PropertySQLProvider) new PropertySQLProvider.Factory(DERBY).create("who-cares", storeType);
+         factory.sql("create-file-table");
+      }
+   }
+
+   @Test
+   public void testGetMissingProperty() {
+      for (SQLProvider.DatabaseStoreType storeType : SQLProvider.DatabaseStoreType.values()) {
+         try {
+            PropertySQLProvider factory = (PropertySQLProvider) new PropertySQLProvider.Factory(DERBY).create("who-cares", storeType);
```
gemmellr, line 42: Shouldnt this be outside the try?

artemis-jdbc-store/src/test/java/org/apache/activemq/artemis/jdbc/store/sql/PropertySQLProviderTest.java
```diff
@@ -0,0 +1,59 @@
...
+            factory.sql("foo");
```
gemmellr, line 43: Your first attempt failed some tests because it didnt allow for returning null as expected in a specific case. You have updated things to fix that by allowing the behaviour to be specified, but dont seem to be testing the new mechanism. Seems like there should be a related variant of this test to cover that.

Also, perhaps use e.g "missing-property" as the key name for later clarity?

artemis-jdbc-store/src/test/java/org/apache/activemq/artemis/jdbc/store/sql/PropertySQLProviderTest.java
jbertram: Hmmm...For some reason this change succeeding locally but blowing up on the PR build.

gemmellr: Interesting. Doesnt the failure suggest the driver was never registered in the test to begin with (I just assumed it was, given you added the shutdown step)? If so maybe it passing for you but failing here means its dependent on the behaviour of other tests classes, or even the order of tests being run? When you say 'locally' do you mean in-IDE, or on CLI (where order should be the same, unless particular profiles influence the tests run)? I do see the earlier tests in the PR build all look to have their own bespoke Derby shutdown handling (some similarly swallowing all exceptions and thus not knowing if they got the shutdown-succeeded one)

Presumably also means your shutdown attempt was also failing originally, just being swallowed? Which was partly what made me think of using the utility method to ensure it wasnt failing. And also as I thought it had more options to stop the shutdown deregistering the driver, in case other tests use it later.

Depending on the situation it might be a case of having to swallow the failure though, to cover off the alternative of never shutting it down if it is later actually registered/used due to code changes?

jbertram: > Doesnt the failure suggest the driver was never registered in the test to begin with (I just assumed it was, given you added the shutdown step)?

Yes. I originally copied this test from `org.apache.activemq.artemis.jdbc.store.journal.JDBCJournalLoaderCallbackTest` which had the `shutdownDerby` method. Apparently this isn't needed at all and in that test it swallows a similar exception every time. I changed my tests to use a `null` SQL dialect to make it more clear that Derby isn't required.

tests/artemis-test-support/src/main/java/org/apache/activemq/artemis/tests/util/ActiveMQTestBase.java
```diff
@@ -484,6 +484,7 @@ protected DatabaseStorageConfiguration createDefaultDatabaseStorageConfiguration
          // and we should only schedule one task
          derbyDropped = true;
          runAfterEx(this::dropDerby);
+         System.out.println("!!!!!!!!!!!!!!!!!!!!!!!!");
```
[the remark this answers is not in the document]
jbertram: :man_facepalming:

PR #5198, opened 2024-08-30
Mitigate NPEs in ScaleDownHandler

jbertram:
Two commits for two Jiras for two potential NPEs in the same method. :dizzy_face:


gemmellr:
The API doc for the createSession method(s) does not say null return is even a possibility. If it is then it should really be mentioned as expected behaviour, especially as it is kinda weird in this case; create methods like that usually throw (which is what is implied by the doc if anything...though with the wiggle room that an exception 'did not occur' here...but only because the code chose not to throw when deciding it failed).

Given the cause and effect and fix is the same and its all in the same method, this seems like one change to me rather than two separate changes. I'd either reference both Jiras in one commit, or update the description of one Jira and drop the other.

jbertram:
@gemmellr, I think that's fair.

PR #5205, opened 2024-09-03
ARTEMIS-4782 ActiveMQConnectionFactory - Add getBrokerURL method

jbertram:


jbertram:
This is turning into a bit of a head-ache. It's definitely not as straight-forward as it first appeared due to URL parsing, conversion, etc. I'm shelving this for now.

PR #5208, opened 2024-09-04
ARTEMIS-4182 specify client ID on bridge as node ID

jbertram:
Automatically fill in the client ID on bridge connections with the node ID of the server where the bridge is configured. This makes the bridge connection easier to identify on the target broker and also mirrors what is already done for both Core and AMQP federation.


tabish121:
I ran a CI job on this and saw several new failures I hadn't seen before, some examples:


> Test and Examples / Test / 2 / org.apache.activemq.artemis.tests.integration.routing.AutoClientIDShardClusterTest.testWithConsistentHashClientIDModTwo() protocol: AMQP
> Test and Examples / Test / 2 / org.apache.activemq.artemis.tests.integration.routing.AutoClientIDShardClusterTest.testWithConsistentHashClientIDModTwo() protocol: CORE
> 

Might be worth a look

jbertram:
@tabish121, thanks for the heads-up. I see what's happening now after much head-scratching. I'll need to rework this PR so that the client ID is configurable because having it always-on may not fit every use-case.

gemmellr:
Configurable sounds like it would be needed in case of ClientID, since isn't it possible to have multiple bridge connections going to the same broker? If so, they cant use the same ClientID.  (Aside: the AMQP Federation bit sets a container-id, but it doesnt add the mechanism to request it be enforced as unique, i.e to have it treated like a ClientID)

jbertram:
For core connections client ID doesn't _have_ to be unique.

gemmellr:
Ah, didnt realise the Core client ClientID, and Core JMS client ClientID, aren't the same thing.

erwindon:
:-) :-)

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -2598,6 +2606,14 @@
             </xsd:annotation>
          </xsd:element>
 
+         <xsd:element name="client-id" type="xsd:string" maxOccurs="1" minOccurs="0">
+            <xsd:annotation>
+               <xsd:documentation>
+                  the identifier to use for the bridge connection; helps with identifying the connection on the remote broker
```
gemmellr, line 2612: bridge -> cluster?

docs/user-manual/clusters.adoc
```diff
@@ -582,6 +583,11 @@ Here we have defined 2 servers that we know for sure will that at least one will
 There may be many more servers in the cluster but these will;
 be discovered via one of these connectors once an initial connection has been made.
 
+client-id::
+An optional identifier to use for the bridge connection.
```
gemmellr, line 587: bridge -> cluster ?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/bridge/BridgeTest.java
```diff
@@ -554,6 +559,44 @@ public void testClientSessionFactoryLeak() throws Exception {
...
+      final String queueName1 = "queue1";
+
+      TransportConfiguration server1tc = new TransportConfiguration(getConnector(), server1Params);
+
+      HashMap<String, TransportConfiguration> connectors = new HashMap<>();
+      connectors.put(server1tc.getName(), server1tc);
+      server0.getConfiguration().setConnectorConfigurations(connectors);
+
+      List<String> connectorConfig = List.of(server1tc.getName());
+      // intentionally configure a bridge that will fail to connect and attempt to reconnect multiple times quickly
```
gemmellr, line 584: Is this true? The test seems to verify a successfull connection

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/bridge/BridgeTest.java
```diff
@@ -554,6 +559,44 @@ public void testClientSessionFactoryLeak() throws Exception {
...
+      server0.getConfiguration().setBridgeConfigurations(List.of(new BridgeConfiguration().setName("bridge1").setQueueName(queueName0).setForwardingAddress(forwardAddress).setRetryInterval(0).setReconnectAttempts(-1).setStaticConnectors(connectorConfig).setClientId(clientId)));
```
gemmellr, line 585: We should avoid setting infinite reconnects in tests, it just makes it more likely for bad things to happen later, e.g. cross-test contamination when something unexpected isnt handled nicely.

The test only seems to allow 2 seconds for connection, so the attempts + intervals only need to span that.

(I'd also name the bridge based on the test name, helps identify them in logs later in debug, and also useful for finding contaminations)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/distribution/ClusterConnectionConfigTest.java
```diff
@@ -103,4 +109,30 @@ private void makeSureForwardingFlowControl(int producerWindow, int... indices) t
       }
    }
 
+   @Test
+   public void testBridgeClientId() throws Exception {
```
gemmellr, line 113: Bridge -> Cluster?

PR #5216, opened 2024-09-05
ARTEMIS-5032 Ensure AMQP message priority is honored after restart

tabish121:
Ensure that on server restart the original priority value assigned to an AMQP message is used when dispatching durable messages from the store. The AMQP Header section is scanned if present and the priority value is recovered in an efficient manner.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPStandardMessage.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPStandardMessage.java
@@ -39,8 +39,12 @@ import org.apache.qpid.proton.amqp.messaging.Header;
 import org.apache.qpid.proton.amqp.messaging.MessageAnnotations;
 import org.apache.qpid.proton.amqp.messaging.Properties;
 import org.apache.qpid.proton.amqp.messaging.Section;
+import org.apache.qpid.proton.codec.DecodeException;
+import org.apache.qpid.proton.codec.DecoderImpl;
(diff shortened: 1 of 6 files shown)


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPStandardMessage.java
```diff
@@ -235,10 +239,87 @@ public void reloadPersistence(ActiveMQBuffer record, CoreMessageObjectPools pool
...
       resetMessageData();
+      recoverHeaderDataFromEncoding();
+
       modified = false;
       messageDataScanned = MessageDataScanningStatus.RELOAD_PERSISTENCE.code;
    }
 
+   private void recoverHeaderDataFromEncoding() {
+      final DecoderImpl decoder = TLSEncode.getDecoder();
+      decoder.setBuffer(data);
```
clebertsuconic, line 250: I'm confused.. wasn't this done before?


Why this new code here?

clebertsuconic: I mean.. I remember we would parse the header in out of order scenarios.. and we even had code.. I'm just wondering why you needed to add new code for that?

tabish121: The recover from persistence isn't reloading Header, or anything else for that matter before adding it to the Queue as the supplied test shows.  Messages read after restart all get added as if priority was set to default as I state in the issue if you happen to have read it.

PR #5220, opened 2024-09-06
ARTEMIS-5037: option to limit mirror propagation

lavocatt:
Add a new option in the Mirror setting to prevent a broker from propagating messages.

When working with a topology where 4 nodes form a square and each node mirrors to its two neighbors, a message leaving a corner can reach the opposite corner of the square by two different routes. This is causing the ordering of message to get broken.

example:
1 <-> 2
^     ^
|     |
v     v
4 <-> 3

A message from a will reach 3 by 2 and 4. Message duplication checks will prevent the message from being duplicated but won't help regarding the order of the messages.

Using the new option to not forward messages coming from a link, we break the possibilities to have two routes to reach the opposite corner.

On the example above, we ask 4 to not forward any messages coming from 1 if 1 is the primary sender of the messages.

This means that now, a message sent to 1 will reach 3 only by 2 and not by 4.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
@@ -432,11 +432,29 @@ public class AMQPBrokerConnection implements ClientConnectionLifeCycleListener, 
                   final boolean coreTunnelingEnabled = isCoreMessageTunnelingEnabled(replica);
                   final Symbol[] desiredCapabilities;
 
-                  if (coreTunnelingEnabled) {
-                     desiredCapabilities = new Symbol[] {AMQPMirrorControllerSource.MIRROR_CAPABILITY,
-                                                         AmqpSupport.CORE_MESSAGE_TUNNELING_SUPPORT};
-                  } else {
-                     desiredCapabilities = new Symbol[] {AMQPMirrorControllerSource.MIRROR_CAPABILITY};
+                  if (!replica.isForward()){
+                     if (coreTunnelingEnabled) {
+                        desiredCapabilities = new Symbol[] {
+                           AMQPMirrorControllerSource.MIRROR_CAPABILITY,
+                           AmqpSupport.CORE_MESSAGE_TUNNELING_SUPPORT,
+                           AMQPMirrorControllerSource.NO_FORWARD
+                        };
+                     } else {
+                        desiredCapabilities = new Symbol[] {
+                           AMQPMirrorControllerSource.MIRROR_CAPABILITY,
+                           AMQPMirrorControllerSource.NO_FORWARD
+                        };
+                     }
+                  }else {
+                     if (coreTunnelingEnabled) {
+                        desiredCapabilities = new Symbol[] {
+                           AMQPMirrorControllerSource.MIRROR_CAPABILITY,
+                           AmqpSupport.CORE_MESSAGE_TUNNELING_SUPPORT};
+                     } else {
+                        desiredCapabilities = new Symbol[] {
+                           AMQPMirrorControllerSource.MIRROR_CAPABILITY
+                        };
+                     }
                   }
 
                   final Symbol[] requiredOfferedCapabilities = new Symbol[] {AMQPMirrorControllerSource.MIRROR_CAPABILITY};
@@ -447,7 +465,7 @@ public class AMQPBrokerConnection implements ClientConnectionLifeCycleListener, 
                                 (r) -> AMQPMirrorControllerSource.validateProtocolData(protonProtocolManager.getReferenceIDSupplier(), r, getMirrorSNF(replica)),
                                 server.getNodeID().toString(),
                                 desiredCapabilities,
-                                null,
+                                replica.isForward() ? null : new Symbol[] {AMQPMirrorControllerSource.NO_FORWARD},
                                 requiredOfferedCapabilities);
                } else if (connectionElement.getType() == AMQPBrokerConnectionAddressType.FEDERATION) {
                   // Starting the Federation triggers rebuild of federation links
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
@@ -92,6 +92,7 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
 
    public static final SimpleString INTERNAL_ID_EXTRA_PROPERTY = SimpleString.of(INTERNAL_ID.toString());
    public static final SimpleString INTERNAL_BROKER_ID_EXTRA_PROPERTY = SimpleString.of(BROKER_ID.toString());
+   public static final Symbol NO_FORWARD = Symbol.getSymbol("amq.no.forward");
 
    private static final ThreadLocal<RoutingContext> mirrorControlRouting = ThreadLocal.withInitial(() -> new RoutingContextImpl(null));
 
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
@@ -57,6 +57,7 @@ import org.apache.activemq.artemis.protocol.amqp.proton.MessageReader;
 import org.apache.activemq.artemis.protocol.amqp.proton.ProtonAbstractReceiver;
 import org.apache.activemq.artemis.utils.ByteUtil;
 import org.apache.activemq.artemis.utils.pools.MpscPool;
+import org.apache.qpid.proton.amqp.Symbol;
 import org.apache.qpid.proton.amqp.messaging.Accepted;
 import org.apache.qpid.proton.amqp.messaging.AmqpValue;
 import org.apache.qpid.proton.amqp.messaging.DeliveryAnnotations;
@@ -264,6 +265,14 @@ public class AMQPMirrorControllerTarget extends ProtonAbstractReceiver implement
    protected void actualDelivery(Message message, Delivery delivery, DeliveryAnnotations deliveryAnnotations, Receiver receiver, Transaction tx) {
       OperationContext oldContext = recoverContext();
       incrementSettle();
+      boolean noForward = false;
+      if (receiver.getRemoteDesiredCapabilities() != null) {
+         for (Symbol capability : receiver.getRemoteDesiredCapabilities()) {
+            if (capability == AMQPMirrorControllerSource.NO_FORWARD) {
+               noForward = true;
+            }
+         }
+      }
 
       logger.trace("{}::actualDelivery call for {}", server, message);
       setControllerInUse(this);
@@ -320,14 +329,14 @@ public class AMQPMirrorControllerTarget extends ProtonAbstractReceiver implement
                   }
                }
             } else {
-               if (sendMessage(amqpMessage, deliveryAnnotations, messageAckOperation)) {
+               if (sendMessage(amqpMessage, deliveryAnnotations, messageAckOperation, noForward)) {
                   // since the send was successful, we give up the reference here,
                   // so there won't be any call on afterCompleteOperations
                   messageAckOperation = null;
                }
             }
          } else {
-            if (sendMessage(message, deliveryAnnotations, messageAckOperation)) {
+            if (sendMessage(message, deliveryAnnotations, messageAckOperation, noForward)) {
                // since the send was successful, we give up the reference here,
                // so there won't be any call on afterCompleteOperations
                messageAckOperation = null;
@@ -486,7 +495,7 @@ public class AMQPMirrorControllerTarget extends ProtonAbstractReceiver implement
     * as the sendMessage was successful the OperationContext of the transaction will take care of the completion.
     * The caller of this method should give up any reference to messageCompletionAck when this method returns true.
     * */
-   private boolean sendMessage(Message message, DeliveryAnnotations deliveryAnnotations, ACKMessageOperation messageCompletionAck) throws Exception {
+   private boolean sendMessage(Message message, DeliveryAnnotations deliveryAnnotations, ACKMessageOperation messageCompletionAck, boolean noForward) throws Exception {
       if (message.getMessageID() <= 0) {
          message.setMessageID(server.getStorageManager().generateID());
       }
@@ -511,6 +520,12 @@ public class AMQPMirrorControllerTarget extends ProtonAbstractReceiver implement
       }
 
       routingContext.setDuplicateDetection(false); // we do our own duplicate detection here
+
+      if (noForward) {
+         message.usageDown(); // large messages would be removed here
+         flow();
+         return false;
+      }
 
       DuplicateIDCache duplicateIDCache;
       if (lruDuplicateIDKey != null && lruDuplicateIDKey.equals(internalMirrorID)) {
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/amqpBrokerConnectivity/AMQPMirrorBrokerConnectionElement.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/amqpBrokerConnectivity/AMQPMirrorBrokerConnectionElement.java
@@ -30,6 +30,8 @@ public class AMQPMirrorBrokerConnectionElement extends AMQPBrokerConnectionEleme
 
    boolean queueCreation = true;
 
+   boolean forward = true;
+
    boolean queueRemoval = true;
 
    boolean messageAcknowledgements = true;
@@ -72,6 +74,12 @@ public class AMQPMirrorBrokerConnectionElement extends AMQPBrokerConnectionEleme
 
    public AMQPMirrorBrokerConnectionElement setQueueCreation(boolean queueCreation) {
       this.queueCreation = queueCreation;
+      return this;
+   }
+
+   public boolean isForward () { return forward; }
+   public AMQPMirrorBrokerConnectionElement setForward(boolean forward) {
+      this.forward = forward;
       return this;
    }
 
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQSquareMirroringTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQSquareMirroringTest.java
@@ -1,0 +1,247 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.tests.integration.amqp.connect;
+
+import org.apache.activemq.artemis.core.config.amqpBrokerConnectivity.AMQPBrokerConnectConfiguration;
+import org.apache.activemq.artemis.core.config.amqpBrokerConnectivity.AMQPMirrorBrokerConnectionElement;
+import org.apache.activemq.artemis.core.server.ActiveMQServer;
+import org.apache.activemq.artemis.core.server.Queue;
+import org.apache.activemq.artemis.tests.integration.amqp.AmqpClientTestSupport;
+import org.apache.activemq.artemis.tests.util.CFUtil;
+import org.apache.activemq.artemis.utils.Wait;
+import org.junit.jupiter.api.Test;
+
+import javax.jms.*;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+public class AMQSquareMirroringTest extends AmqpClientTestSupport {
+
+   protected static final int AMQP_PORT_2 = 5673;
+   protected static final int AMQP_PORT_3 = 5674;
+   protected static final int AMQP_PORT_4 = 5675;
+
+   ActiveMQServer server_2;
+   ActiveMQServer server_3;
+   ActiveMQServer server_4;
+
+   @Override
+   protected ActiveMQServer createServer() throws Exception {
+      return createServer(AMQP_PORT, false);
+   }
+
+   @Test
+   public void testSquare() throws Exception {
+      server_2 = createServer(AMQP_PORT_2, false);
+      server_3 = createServer(AMQP_PORT_3, false);
+      server_4 = createServer(AMQP_PORT_4, false);
+
+      // name the servers, for convenience during debugging
+      server.getConfiguration().setName("1");
+      server_2.getConfiguration().setName("2");
+      server_3.getConfiguration().setName("3");
+      server_4.getConfiguration().setName("4");
+
+      /**
+       * 1 <----> 2
+       * ^        ^
+       * |        |
+       * v        v
+       * 4 <----> 3
+       */
+
+      {
+         AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration("to2", "tcp://localhost:" + AMQP_PORT_2).setRetryInterval(100).setReconnectAttempts(-1);
+         amqpConnection.addElement(new AMQPMirrorBrokerConnectionElement());
+         server.getConfiguration().addAMQPConnection(amqpConnection);
+         amqpConnection = new AMQPBrokerConnectConfiguration("to4", "tcp://localhost:" + AMQP_PORT_4).setRetryInterval(100).setReconnectAttempts(-1);
+         amqpConnection.addElement(new AMQPMirrorBrokerConnectionElement().setForward(false));
+         server.getConfiguration().addAMQPConnection(amqpConnection);
+      }
+
+      {
+         AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration("to1", "tcp://localhost:" + AMQP_PORT).setRetryInterval(100).setReconnectAttempts(-1);
+         amqpConnection.addElement(new AMQPMirrorBrokerConnectionElement());
+         server_2.getConfiguration().addAMQPConnection(amqpConnection);
+         amqpConnection = new AMQPBrokerConnectConfiguration("to3", "tcp://localhost:" + AMQP_PORT_3).setRetryInterval(100).setReconnectAttempts(-1);
+         amqpConnection.addElement(new AMQPMirrorBrokerConnectionElement());
+         server_2.getConfiguration().addAMQPConnection(amqpConnection);
+      }
+
+      {
+         AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration("to2", "tcp://localhost:" + AMQP_PORT_2).setRetryInterval(100).setReconnectAttempts(-1);
+         amqpConnection.addElement(new AMQPMirrorBrokerConnectionElement());
+         server_3.getConfiguration().addAMQPConnection(amqpConnection);
+         amqpConnection = new AMQPBrokerConnectConfiguration("to4", "tcp://localhost:" + AMQP_PORT_4).setRetryInterval(100).setReconnectAttempts(-1);
+         amqpConnection.addElement(new AMQPMirrorBrokerConnectionElement());
+         server_3.getConfiguration().addAMQPConnection(amqpConnection);
+      }
+
+      {
+         AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration("to1", "tcp://localhost:" + AMQP_PORT).setRetryInterval(100).setReconnectAttempts(-1);
+         amqpConnection.addElement(new AMQPMirrorBrokerConnectionElement());
+         server_4.getConfiguration().addAMQPConnection(amqpConnection);
+         amqpConnection = new AMQPBrokerConnectConfiguration("to3", "tcp://localhost:" + AMQP_PORT_3).setRetryInterval(100).setReconnectAttempts(-1);
+         amqpConnection.addElement(new AMQPMirrorBrokerConnectionElement());
+         server_4.getConfiguration().addAMQPConnection(amqpConnection);
+      }
+
+      server.start();
+      server_2.start();
+      server_3.start();
+      server_4.start();
+
+      createAddressAndQueues(server);
+      Wait.assertTrue(() -> server.locateQueue(getQueueName()) != null);
+      Wait.assertTrue(() -> server_2.locateQueue(getQueueName()) != null);
+      Wait.assertTrue(() -> server_3.locateQueue(getQueueName()) != null);
+      Wait.assertTrue(() -> server_4.locateQueue(getQueueName()) != null);
+
+      Queue q1 = server.locateQueue(getQueueName());
+      assertNotNull(q1);
+
+      Queue q2 = server.locateQueue(getQueueName());
+      assertNotNull(q2);
+
+      Queue q3 = server.locateQueue(getQueueName());
+      assertNotNull(q3);
+
+      Queue q4 = server.locateQueue(getQueueName());
+      assertNotNull(q4);
+
+      ConnectionFactory factory = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + AMQP_PORT+"?amqp.idleTimeout=-1");
+      ConnectionFactory factory2 = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + AMQP_PORT_2+"?amqp.idleTimeout=-1");
+      ConnectionFactory factory3 = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + AMQP_PORT_3+"?amqp.idleTimeout=-1");
+      ConnectionFactory factory4 = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + AMQP_PORT_4+"?amqp.idleTimeout=-1");
+
+      try (Connection conn = factory.createConnection()) {
+         Session session = conn.createSession();
+         MessageProducer producer = session.createProducer(session.createQueue(getQueueName()));
+         for (int i = 0; i < 40; i++) {
+            producer.send(session.createTextMessage("message " + i));
+         }
+      }
+
+      Thread.sleep(1_000); // some time to allow eventual loops
+
+      Wait.assertEquals(40L, q1::getMessageCount, 1000, 100);
+      Wait.assertEquals(40L, q2::getMessageCount, 1000, 100);
+      Wait.assertEquals(40L, q3::getMessageCount, 1000, 100);
+      Wait.assertEquals(40L, q4::getMessageCount, 1000, 100);
+
+      try (Connection conn = factory.createConnection()) {
+         Session session = conn.createSession();
+         conn.start();
+         MessageConsumer consumer = session.createConsumer(session.createQueue(getQueueName()));
+         for (int i = 0; i < 10; i++) {
+            TextMessage message = (TextMessage) consumer.receive(1000);
+            assertNotNull(message);
+            assertEquals("message " + i, message.getText());
+         }
+         consumer.close();
+      }
+
+      Wait.assertEquals(30L, q1::getMessageCount, 1000, 100);
+      Wait.assertEquals(30L, q2::getMessageCount, 1000, 100);
+      Wait.assertEquals(30L, q3::getMessageCount, 1000, 100);
+      Wait.assertEquals(30L, q4::getMessageCount, 1000, 100);
+
+      try (Connection conn = factory2.createConnection()) {
+         Session session = conn.createSession();
+         conn.start();
+         MessageConsumer consumer = session.createConsumer(session.createQueue(getQueueName()));
+         for (int i = 10; i < 20; i++) {
+            TextMessage message = (TextMessage) consumer.receive(1000);
+            assertNotNull(message);
+            assertEquals("message " + i, message.getText());
+         }
+         consumer.close();
+      }
+
+      Wait.assertEquals(20L, q1::getMessageCount, 1000, 100);
+      Wait.assertEquals(20L, q2::getMessageCount, 1000, 100);
+      Wait.assertEquals(20L, q3::getMessageCount, 1000, 100);
+      Wait.assertEquals(20L, q4::getMessageCount, 1000, 100);
+
+      try (Connection conn = factory3.createConnection()) {
+         Session session = conn.createSession();
+         conn.start();
+         MessageConsumer consumer = session.createConsumer(session.createQueue(getQueueName()));
+         for (int i = 20; i < 30; i++) {
+            TextMessage message = (TextMessage) consumer.receive(1000);
+            assertNotNull(message);
+            assertEquals("message " + i, message.getText());
+         }
+         consumer.close();
+      }
+
+      Wait.assertEquals(10L, q1::getMessageCount, 1000, 100);
+      Wait.assertEquals(10L, q2::getMessageCount, 1000, 100);
+      Wait.assertEquals(10L, q3::getMessageCount, 1000, 100);
+      Wait.assertEquals(10L, q4::getMessageCount, 1000, 100);
+
+      try (Connection conn = factory4.createConnection()) {
+         Session session = conn.createSession();
+         conn.start();
+         MessageConsumer consumer = session.createConsumer(session.createQueue(getQueueName()));
+         for (int i = 30; i < 40; i++) {
+            TextMessage message = (TextMessage) consumer.receive(1000);
+            assertNotNull(message);
+            assertEquals("message " + i, message.getText());
+         }
+         consumer.close();
+      }
+
+      Wait.assertEquals(0L, q1::getMessageCount, 1000, 100);
+      Wait.assertEquals(0L, q2::getMessageCount, 1000, 100);
+      Wait.assertEquals(0L, q3::getMessageCount, 1000, 100);
+      Wait.assertEquals(0L, q4::getMessageCount, 1000, 100);
+
+      try (Connection conn = factory.createConnection()) {
+         Session session = conn.createSession();
+         conn.start();
+         MessageConsumer consumer = session.createConsumer(session.createQueue(getQueueName()));
+         assertNull(consumer.receiveNoWait());
+         consumer.close();
+      }
+
+      try (Connection conn = factory2.createConnection()) {
+         Session session = conn.createSession();
+         conn.start();
+         MessageConsumer consumer = session.createConsumer(session.createQueue(getQueueName()));
+         assertNull(consumer.receiveNoWait());
+         consumer.close();
+      }
+
+      try (Connection conn = factory3.createConnection()) {
+         Session session = conn.createSession();
+         conn.start();
+         MessageConsumer consumer = session.createConsumer(session.createQueue(getQueueName()));
+         assertNull(consumer.receiveNoWait());
+         consumer.close();
+      }
+
+      try (Connection conn = factory4.createConnection()) {
+         Session session = conn.createSession();
+         conn.start();
+         MessageConsumer consumer = session.createConsumer(session.createQueue(getQueueName()));
+         assertNull(consumer.receiveNoWait());
+         consumer.close();
+      }
+
+   }
+}


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
```diff
@@ -432,11 +432,29 @@ private void doConnect() {
                   final boolean coreTunnelingEnabled = isCoreMessageTunnelingEnabled(replica);
                   final Symbol[] desiredCapabilities;
 
-                  if (coreTunnelingEnabled) {
-                     desiredCapabilities = new Symbol[] {AMQPMirrorControllerSource.MIRROR_CAPABILITY,
-                                                         AmqpSupport.CORE_MESSAGE_TUNNELING_SUPPORT};
-                  } else {
-                     desiredCapabilities = new Symbol[] {AMQPMirrorControllerSource.MIRROR_CAPABILITY};
+                  if (!replica.isForward()){
```
tabish121, line 435: Once you get to this level of complexity to manage the elements in the desired capabilities I'd suggest just switch to creating an ArrayList and adding the elements individually based on the configuration and then just doing a toArray() on that to get the final capabilities array.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/amqpBrokerConnectivity/AMQPMirrorBrokerConnectionElement.java
```diff
@@ -75,6 +77,12 @@ public AMQPMirrorBrokerConnectionElement setQueueCreation(boolean queueCreation)
       return this;
    }
 
+   public boolean isForward () { return forward; }
```
tabish121, line 80: I'd rather this use the same formatting as other accessors and not jam it all on one line, and also space things as others are done here.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
```diff
@@ -512,6 +521,12 @@ private boolean sendMessage(Message message, DeliveryAnnotations deliveryAnnotat
 
       routingContext.setDuplicateDetection(false); // we do our own duplicate detection here
 
+      if (noForward) {
```
tabish121, line 524: If getting to this point the message isn't sent then it would be good to tie a bow on the logging that has occurred to this point and log a trace level message indicating what happened.

lavocatt:
@tabish121 Thanks for the review. FYI this is in a very drafty state ATM. Many refactors might come after that, in any case I'll make sure to take your notes in.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -92,6 +92,7 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
 
    public static final SimpleString INTERNAL_ID_EXTRA_PROPERTY = SimpleString.of(INTERNAL_ID.toString());
    public static final SimpleString INTERNAL_BROKER_ID_EXTRA_PROPERTY = SimpleString.of(BROKER_ID.toString());
+   public static final Symbol NO_FORWARD = Symbol.getSymbol("amq.no.forward");
```
gemmellr, line 95: Given this will be specific to the mirroring, I'd wind that into the value so its also specific.

This should be moved 4 lines up beside the related main MIRROR_CAPABILITY.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
```diff
@@ -264,6 +265,14 @@ public void flow() {
...
       OperationContext oldContext = recoverContext();
       incrementSettle();
+      boolean noForward = false;
+      if (receiver.getRemoteDesiredCapabilities() != null) {
+         for (Symbol capability : receiver.getRemoteDesiredCapabilities()) {
+            if (capability == AMQPMirrorControllerSource.NO_FORWARD) {
+               noForward = true;
+            }
+         }
+      }
```
gemmellr, line 275: This isnt going to change for the duration of the link, so it makes sense to evaluate it once at the start when initializing the mirror target, rather than determining the same result repeatedly for each arriving message.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
gemmellr: Also, this seems to be too early to check this. It cant know whether 'mirror forwarding' will occur until during the routing stages later in this method, which is where mirroring handling occurs. Returning here seems like it would mean it drops all messages, even from _this_ broker they were being mirrored to.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/amqpBrokerConnectivity/AMQPMirrorBrokerConnectionElement.java
gemmellr: Same. Its not allowed by the checkstyle config anyways.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQSquareMirroringTest.java
```diff
@@ -0,0 +1,247 @@
...
+import org.apache.activemq.artemis.core.server.ActiveMQServer;
+import org.apache.activemq.artemis.core.server.Queue;
+import org.apache.activemq.artemis.tests.integration.amqp.AmqpClientTestSupport;
+import org.apache.activemq.artemis.tests.util.CFUtil;
+import org.apache.activemq.artemis.utils.Wait;
+import org.junit.jupiter.api.Test;
+
+import javax.jms.*;
+
+import static org.junit.jupiter.api.Assertions.*;
```
gemmellr, line 30: Lets add the actual imports rather than *

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQSquareMirroringTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+public class AMQSquareMirroringTest extends AmqpClientTestSupport {
```
gemmellr, line 32: All the existing tests in the package use AMQ\*P\*FooTest or just FooTest.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQSquareMirroringTest.java
```diff
@@ -0,0 +1,247 @@
...
+      /**
+       * 1 <----> 2
+       * ^        ^
+       * |        |
+       * v        v
+       * 4 <----> 3
+       */
+
+      {
+         AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration("to2", "tcp://localhost:" + AMQP_PORT_2).setRetryInterval(100).setReconnectAttempts(-1);
```
gemmellr, line 68: Rather than just "to2" etc, each of which you use more than once, I'd include both the source and target ID so that every name is different. Personally I'd also use e.g getTestMethodName() value as a prefix to make them even more to the point.

We shouldnt use -1 reconnect attempts, if something barfs in a way anything leaks (shouldn't..but stuff that shouldn't happen, often does..) it could sit around reconnecting and interfering with other tests later (which is when the unique names, including the test names, becomes especially helpful).

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQSquareMirroringTest.java
```diff
@@ -0,0 +1,247 @@
...
+      Queue q3 = server.locateQueue(getQueueName());
+      assertNotNull(q3);
+
+      Queue q4 = server.locateQueue(getQueueName());
+      assertNotNull(q4);
+
+      ConnectionFactory factory = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + AMQP_PORT+"?amqp.idleTimeout=-1");
+      ConnectionFactory factory2 = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + AMQP_PORT_2+"?amqp.idleTimeout=-1");
+      ConnectionFactory factory3 = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + AMQP_PORT_3+"?amqp.idleTimeout=-1");
+      ConnectionFactory factory4 = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + AMQP_PORT_4+"?amqp.idleTimeout=-1");
```
gemmellr, line 129: Should probably drop the idleTimeout disabling from the final version.

lavocatt:
Thanks @gemmellr for the review! There are still things that are not working with the current PR and some refactors in the added lines might happen. In any case I'll make sure to take your comments in.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
lavocatt: Thanks @gemmellr for pointing that out. I was having some weird behaviors with messages suddenly disappearing, clearly related to what you are mentioning.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
lavocatt: That makes a lot of sense. Thanks.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPSquareMirroringTest.java
```diff
@@ -0,0 +1,278 @@
...
+      Queue q2 = server.locateQueue(getQueueName());
+      assertNotNull(q2);
+
+      Queue q3 = server.locateQueue(getQueueName());
+      assertNotNull(q3);
+
+      Queue q4 = server.locateQueue(getQueueName());
+      assertNotNull(q4);
+
+      ConnectionFactory factory = CFUtil.createConnectionFactory("AMQP", "tcp://localhost:" + AMQP_PORT);
```
clebertsuconic, line 136: can you make sure this test works with other protocols?


you can either run all 3 protocols (AMQP, CORE, OPENWIRE) with a parameter, or make the selection random if this test is a long running one (I added recently an method on ActiveMQTestBase called randomProtocol())

lavocatt: I tried to use a `randomProtocol()`:
image

and then this error happened: 
image

Any idea on what could be going on?

lavocatt:
I've exposed the new `no-message-forwarding` parameter (with @gemmellr's help) , now I need to test that the parameter works in a test.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
```diff
@@ -534,6 +549,9 @@ private boolean sendMessage(Message message, DeliveryAnnotations deliveryAnnotat
 
       message.setBrokerProperty(INTERNAL_ID_EXTRA_PROPERTY, internalID);
       message.setBrokerProperty(INTERNAL_BROKER_ID_EXTRA_PROPERTY, internalMirrorID);
+      if (!this.canForwardMessages) {
+         message.setBrokerProperty(INTERNAL_NO_FORWARD, true);
```
gemmellr, line 553: I wondered if the RoutingContext could be used for the 'dont [re-]mirror this' handling. It already has functionality for doing that (see isMirrorDisabled() and related).

However it now also occurs to me, we possibly do need to mark the message like this simply because when that message is acknowledged later, we probably want to know then that it was 'no forward' now, so that we dont mirror acknowledgement for it either given we never mirrored the message itself to begin with.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/amqpBrokerConnectivity/AMQPMirrorBrokerConnectionElement.java
```diff
@@ -30,6 +30,8 @@ public class AMQPMirrorBrokerConnectionElement extends AMQPBrokerConnectionEleme
 
    boolean queueCreation = true;
 
+   boolean canForwardMessages = true;
```
gemmellr, line 33: similarly I would keep the field the same sense as the option, plus the XML and config option should also use the same sense so that the defaults are more describable for both the XML and broker-properties usage. Currently someone using broker-properties would need to configure 'canForwardMessages = false' whilst someone using XML would need to configure 'no-messages-forwarding = true' to flip their actually-shared behaviour.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPChainedReplicaTest.java
```diff
@@ -196,4 +196,4 @@ public void testChained() throws Exception {
       }
 
    }
-}
+}
```
gemmellr, line 199: Dont think we need to update this file for this change :)

artemis-server/src/main/resources/schema/artemis-configuration.xsd
```diff
@@ -2290,6 +2290,14 @@
...
          </xsd:annotation>
       </xsd:attribute>
+      <xsd:attribute name="no-message-forwarding" type="xsd:boolean" use="optional" default="false">
+         <xsd:annotation>
+            <xsd:documentation>
+               If this is true, the mirror at the opposite end of the link will not forward messages coming from that link to any other mirrors down the line.
+               This is false by default.
+            </xsd:documentation>
+         </xsd:annotation>
+      </xsd:attribute>
```
gemmellr, line 2300: This name+description, plus my earlier note about the acknowledgements, make me realise it is still going to re-mirror other things like address and queue creations, and more importantly deletions. Which could still be quite problematic just like the message-duping themselves, likely worse due to timing interactions of creation and deletion.

I'm not sure anything can be done about that except only using pre-created everything. I am again sceptical of this topology.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerTarget.java
```diff
@@ -248,6 +253,13 @@ public AMQPMirrorControllerTarget(AMQPSessionCallback sessionSPI,
       this.configuration = server.getConfiguration();
       this.referenceNodeStore = sessionSPI.getProtocolManager().getReferenceIDSupplier();
       mirrorContext = protonSession.getSessionSPI().getSessionContext();
+      if (receiver.getRemoteDesiredCapabilities() != null) {
+         for (Symbol capability : receiver.getRemoteDesiredCapabilities()) {
+            if (capability == NO_FORWARD) {
+               this.canForwardMessages = false;
+            }
```
gemmellr, line 260: As discussed yesterday, this should send back indicating it offers the support, and the origin should check it receives the offer and barf if not.

lavocatt: For now I haven't found how to implement that, but thanks for bringing it up again, that's the next step on the TODO list.

lavocatt: This is done now :slightly_smiling_face:

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPSquareMirroringTest.java
lavocatt: This got fixed, I'm now using the randomProtocols

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPSquareMirroringTest.java
```diff
@@ -0,0 +1,282 @@
...
+
+   protected String getConfiguredProtocols() {
+      return "AMQP,CORE,OPENWIRE";
+   }
+
+   @Test
+   public void testSquare() throws Exception {
+      server_2 = createServer(AMQP_PORT_2, false);
+      server_3 = createServer(AMQP_PORT_3, false);
+      server_4 = createServer(AMQP_PORT_4, false);
```
gemmellr, line 62: I'd also try adding some testing using the ProtonJ2 test peers, rather than all brokers, to validate and exercise the actual broker behaviour at the protocol level. It gives a better accounting of what is really going on and when it changes unexpectedly.

E.g it seems all but certain these brokers are currently forwarding around mirrored acks for noForward messages they didn't originally mirror to begin with, but this test will not notice that one way or the other. Similarly with various other potentially unexpected behaviours that might occur.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/PostOffice.java
```diff
@@ -65,10 +65,14 @@ default void reloadAddressInfo(AddressInfo addressInfo) throws Exception {
 
    AddressInfo removeAddressInfo(SimpleString address, boolean force) throws Exception;
 
+   AddressInfo removeAddressInfo(SimpleString address, boolean force, boolean noForward) throws Exception;
```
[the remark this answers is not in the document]
lavocatt: I've implemented this today.

artemis-server/src/main/resources/schema/artemis-configuration.xsd
lavocatt: Got that fixed.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -559,6 +587,10 @@ public void preAcknowledge(final Transaction tx, final MessageReference ref, fin
          return;
       }
 
+      if (isBlockedByNoForward()) {
+         return;
+      }
+
```
gemmellr, line 593: I suspect a different check will be needed here, inspecting the specific message; this is ultimately called by the queue, which I expect usually wont have a mirror target controller set. It will also want to accommodate not sending acks for no-forward messages [that it didnt send to begin with] even across broker restarts, due to expiry, etc etc.

I expect the test-peer based testing I know you are working on will demonstrate this.

lavocatt: I've implemented some tests and they seem to work despite the assumption we had that it would not. Maybe something is worth a double check?

gemmellr: I think your changes last week have actually already stopped it sending some of the stuff I previously thought it would and checked that it actually did.

There are other scenarios that are less clear though. For example, if you consume the message from the middle broker instead of the first, its going to mirror the ack onward, even though it didnt mirror the message originally, which is marked as no-forward.

lavocatt:
I have figured out something I don't know how to fix

On this topology:
```
1 - noForward -> 2 -> 3
^                |
|----------------|
```
producing and consuming on 1 works, 2 mirrors, 3 is left alone
producing on 1 and consuming on 2 breaks:
3 is left alone, which is good
1 doesn't receive the ack, which is bad
I don't know how to overcome that. Because if I decide to forward the Ack to one when I consume on two, then if the message is consumed on 1 it'll receive a duplicate Ack from 2.

gemmellr:
Yeah, if you say noForward, and dont provide a reverse mirror, you are not getting the ack. Thats basically as requested.

there is no way to fix that without making the mirroring some kind of full topology mesh handling system with 'command forwarding between nodes without acting on them'.

I'd consider that setup basically an invalid use of this already-pretty-questionable functionality. Ask for no forwarding -> get no forwarding.

gtully:
it makes sense to have some basic checks on the mirror, but it is also fine to have restrictions on the topologies that will work with a mirror and others that won't work. The most important thing imho is that it is understandable and deterministic

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
lavocatt: I believe this is addressed now, feel free to reopen if you catch a mistake.

lavocatt:
My issue from two weeks ago was resolved by annotating the Ack with the ID of the intended receiver. And when the message is in the actual process of being sent a filtering procedure occurs to check if the actual receiver matches the ID stored in the ACK's annotation.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
```diff
@@ -306,10 +307,10 @@ public void validateMatching(Queue queue, AMQPBrokerConnectionElement connection
    public void createLink(Queue queue, AMQPBrokerConnectionElement connectionElement) {
       if (connectionElement.getType() == AMQPBrokerConnectionAddressType.PEER) {
          Symbol[] dispatchCapability = new Symbol[]{AMQPMirrorControllerSource.QPID_DISPATCH_WAYPOINT_CAPABILITY};
-         connectSender(queue, queue.getAddress().toString(), null, null, null, null, dispatchCapability, null);
+         connectSender(queue, queue.getAddress().toString(), null, null,null, null, null, dispatchCapability, null);
```
gemmellr, line 310: Should be a space before the new null, like the surroundings.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -89,9 +90,15 @@ public class AMQPMirrorControllerSource extends BasicMirrorController<Sender> im
    // Capabilities
    public static final Symbol MIRROR_CAPABILITY = Symbol.getSymbol("amq.mirror");
    public static final Symbol QPID_DISPATCH_WAYPOINT_CAPABILITY = Symbol.valueOf("qd.waypoint");
+   public static final Symbol NO_FORWARD = Symbol.getSymbol("amq.no.forward");
+   public static final Symbol NO_FORWARD_SOURCE = Symbol.getSymbol("amq.no.forward.source");
+   public static final Symbol RECEIVER_ID_FILTER = Symbol.getSymbol("amq.receiver.id.filter");
```
gemmellr, line 95: NO_FORWARD_SOURCE and RECEIVER_ID_FILTER arent being used as a capability, so shouldn't be in this group of constants that are for capabilities. They also likely don't need to be of Symbol type as a result.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -53,6 +53,7 @@
 import org.apache.qpid.proton.amqp.Symbol;
 import org.apache.qpid.proton.amqp.messaging.DeliveryAnnotations;
 import org.apache.qpid.proton.amqp.messaging.Properties;
+import org.apache.qpid.proton.amqp.transport.Target;
```
gemmellr, line 56: Think this will be unused, the MirrorSource shouldnt need to be dealing with Target's

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -338,6 +367,12 @@ Message copyMessageForPaging(Message message) {
    public void sendMessage(Transaction tx, Message message, RoutingContext context) {
       SimpleString address = context.getAddress(message);
 
+      if (isBlockedByNoForward(message)) {
+         String remoteID = getRemoteMirrorId();
```
gemmellr, line 371: Variable looks unused, did you mean to add it to the logging?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -467,6 +504,28 @@ public static void validateProtocolData(ReferenceIDSupplier referenceIDSupplier,
       }
    }
 
+   /**
+    * Checks if the message ref should be filtered or not.
+    * @param ref the message to filter
+    * @return true if the INTERNAL_RECEIVER_ID_FILTER annotation of the message is set to a different value
+    * that the remoteMirrorID, false otherwise.
```
gemmellr, line 511: (No need to keep the suggestion commit, its just easier to show the suggested change this way)

```suggestion
    * @param ref the message to check
    * @return true if the INTERNAL_RECEIVER_ID_FILTER annotation of the message is set to a different value
    * than the remoteMirrorID, false otherwise.
```

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -543,6 +602,17 @@ public void preAcknowledge(final Transaction tx, final MessageReference ref, fin
          logger.trace("preAcknowledge::tx={}, ref={}, reason={}", tx, ref, reason);
       }
 
+      SimpleString noForwardSource = null;
+      if (Boolean.TRUE.equals(ref.getMessage().getBooleanProperty(INTERNAL_NO_FORWARD))) {
+         noForwardSource = (SimpleString) ref.getMessage().getBrokerProperty(INTERNAL_NO_FORWARD_SOURCE);
```
gemmellr, line 607: I dont think we should be setting a SimpleString value for this...and now looking at the other uses, it doesnt seem like it will be, its set as a String originally? (Which makes sense as its ultimately coming from an AMQP link property, as a string value).

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorConnectionTest.java
```diff
@@ -349,20 +512,41 @@ public void testProducerMessageIsMirroredWithoutCoreTunnelingUsesDefaultMessageF
...
 
+   @Test
+   @Timeout(20)
+   public void testProducerMessageIsMirroredWithNoForwardAndTunneling() throws Exception {
+      doTestProducerMessageIsMirroredWithCorrectMessageFormat(true, true);
+   }
+
+   @Test
+   @Timeout(20)
+   public void testProducerMessageIsMirroredWithNoForwardAndTunelingAndWithoutTunneling() throws Exception {
```
gemmellr, line 523: testProducerMessageIsMirroredWithNoForwardAndWithoutTunneling() ?

gemmellr (commented):
Gave it a quick skim, commented on some small niggles. Didnt fully consider if as yet, e.g if the internal-to-consumer filtering actually works ok (seems like there could be potential to throw off metrics, credit handling, acking, etc). There are some issues the checks build is pointing out (I might have commented on some of them, though expect there are others I didnt catch in the skim). I didn't get to the tests yet, will look tomorrow.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorConnectionTest.java
lavocatt: Hehe, thanks for pointing this out :sweat_smile:

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
lavocatt: It's a leftover from a debugging session I forgot to remove.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
lavocatt: The NO_FORWARD_SOURCE is used as a BrokerProperty which is and that requires to have a Symbol passed as a parameter:

image

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
lavocatt: Even though the value we set on the target is a String, it gets somehow turned into a SimpleString. If I try to cast that broker property to a string, I'm getting: `class org.apache.activemq.artemis.api.core.SimpleString cannot be cast to class java.lang.String`.

On the target side I do send a String in the property:
image

gemmellr: Seems to be the fault of setBrokerProperty() using TypedProperties for 'extra properties' (as well as setting a message annotation if AMQP message), which converts a String value to a SimpleString :/

I'd check the exact behaviour first, but I'd be inclined to flip the behaviour so it checks with a String instead, i.e that it null checks then does String.valueOf to get a String for the comparison. That way its not creating new SimpleString from the [typically same] remoteMirrorId value for every message checked, but instead just getting the String from the SimpleString (that likely already exists because it was  created from a String).

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -467,6 +500,28 @@ public static void validateProtocolData(ReferenceIDSupplier referenceIDSupplier,
       }
    }
 
+   /**
+    * Checks if the message ref should be filtered or not.
+    * @param ref the message to check
+    * @return true if the INTERNAL_RECEIVER_ID_FILTER annotation of the message is set to a different value
+    * than the remoteMirrorID, false otherwise.
+    */
+   public boolean filterMessage(MessageReference ref) {
```
gemmellr, line 509: Whilst this may stop the message actually being sent, its not clear to me it does anything around actually cleaning it up? As far as the broker is concerned it sent a message, but the other end can never acknowledge it since it never actually arrives as it wasnt sent...so is it just stuck in the structure as a delivered-but-not-acked message? Presumably the snf queue delivering metrics would show this (and and be forever incorrect) if so.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
lavocatt: Flipping the behavior works for me, but depending on the volume of changes that may or may not require, I think this could endup being off-topic for the PR.

gemmellr: When I speak of flipping the behaviour I just mean the comparison done within the if statements here, not anything else. Thats definitely on-topic for this PR given it is adding the bit that it would be good not to do.

lavocatt: Okay makes sense, that was some misunderstanding on my side, sorry about that.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
lavocatt: I added another commit trying another approach at filtering by doing it at an higher level. Let me know what you think about this.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorConnectionTest.java
```diff
@@ -229,6 +271,173 @@ public void testBrokerAddsAddressAndQueue() throws Exception {
...
+         Wait.assertTrue(() -> server.locateQueue(getQueueName()) != null);
+
+         final org.apache.activemq.artemis.core.server.Queue q1 = server.locateQueue(getQueueName());
+         assertNotNull(q1);
+
+         final org.apache.activemq.artemis.core.server.Queue q2 = server_2.locateQueue(getQueueName());
+         assertNotNull(q2);
+
+         final ConnectionFactory factory = CFUtil.createConnectionFactory("CORE", "tcp://localhost:" + BROKER_PORT_NUM);
+         final ConnectionFactory factory_2 = CFUtil.createConnectionFactory("CORE", "tcp://localhost:" + AMQP_PORT_2);
```
gemmellr, line 373: Probably worth renaming the test to make clear it is Core large messages being sent originally. Equally, if controlling the AMQP large message size (which here, will only matter for the non-tunnelled case as mirrored messages arrive at server 2 via server 1) then it probably makes sense to configure the Core client large message size (via its URI, e.g add "?minLargeMessageSize=\<bytes-size\>"

Probably worth having tests with large AMQP messages also.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorConnectionTest.java
```diff
@@ -229,6 +271,173 @@ public void testBrokerAddsAddressAndQueue() throws Exception {
...
+
+         // Then send messages on the broker directly connected to the peer, the messages should make it to the peer.
+         // Receiving these 3 messages in that order confirms that no previous data reched the Peer, therefore validating
+         // the test.
+         peer_3.expectTransfer().accept(); // Address create
+         peer_3.expectTransfer().accept(); // Queue create
+         peer_3.expectTransfer().withMessageFormat(messageFormat).accept(); // Producer Message
+
+         server_2.addAddressInfo(new AddressInfo(SimpleString.of("sometest"), RoutingType.ANYCAST));
+         server_2.createQueue(QueueConfiguration.of("sometest").setRoutingType(RoutingType.ANYCAST));
```
gemmellr, line 422: could probably use something more descriptive than "sometest". Either way, a variable would probably make it easier to follow where it is used, e.g in the mirror config for filtering, and again here/below.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorConnectionTest.java
```diff
@@ -229,6 +271,173 @@ public void testBrokerAddsAddressAndQueue() throws Exception {
...
+         // give some time to peer_3 to receive messages (if any)
+         Thread.sleep(100);
+         peer_3.waitForScriptToComplete(5, TimeUnit.SECONDS); // if messages are received here, this should error out
+
+         // Then send messages on the broker directly connected to the peer, the messages should make it to the peer.
+         // Receiving these 3 messages in that order confirms that no previous data reched the Peer, therefore validating
+         // the test.
+         peer_3.expectTransfer().accept(); // Address create
+         peer_3.expectTransfer().accept(); // Queue create
+         peer_3.expectTransfer().withMessageFormat(messageFormat).accept(); // Producer Message
```
gemmellr, line 419: Could also verify the text body of the message to better validate things.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorConnectionTest.java
```diff
@@ -229,6 +271,173 @@ public void testBrokerAddsAddressAndQueue() throws Exception {
...
+         server_2.addAddressInfo(new AddressInfo(SimpleString.of("sometest"), RoutingType.ANYCAST));
+         server_2.createQueue(QueueConfiguration.of("sometest").setRoutingType(RoutingType.ANYCAST));
+
+         try (Connection connection = factory_2.createConnection()) {
+            final Session session = connection.createSession(false, Session.CLIENT_ACKNOWLEDGE);
+            final MessageProducer producer = session.createProducer(session.createQueue("sometest"));
+            final TextMessage msg = session.createTextMessage("test");
+
+            connection.start();
+            producer.setDeliveryMode(DeliveryMode.PERSISTENT);
```
gemmellr, line 430: This is the default, can be dropped for conciseness. Dont actually need to start a producer-only connection either.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPNoForwardMirroringTest.java
```diff
@@ -0,0 +1,278 @@
...
+import org.apache.activemq.artemis.tests.integration.amqp.AmqpClientTestSupport;
+import org.apache.activemq.artemis.tests.util.CFUtil;
+import org.apache.activemq.artemis.utils.Wait;
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+public class AMQPNoForwardMirroringTest extends AmqpClientTestSupport {
```
gemmellr, line 39: Most of the tests added in the previous/existing class, would actually be obvious candidates for a class with this name. Perhaps a helpful comment here noting their presence in the other class and to also see it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPNoForwardMirroringTest.java
```diff
@@ -0,0 +1,278 @@
...
+      server.getConfiguration().setName("1");
+      server_2.getConfiguration().setName("2");
+      server_3.getConfiguration().setName("3");
+
+      /**
+       *
+       * the mirroring topology:
+       *
+       * v---------|
+       * 1 ------> 2       The link between 1 and 2 is noForward=true
```
gemmellr, line 72: Adding another linebetween these, to allow a more obvious [downward] side here would probably make this clearer. You could also make it wider and embed the (noForward) in the there like the other test class comment.

Its also inconsistent with its use of line types denoting the noForward config in the other later test class, would be good to make them the classes illustrate things the same way for clarity.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPNoForwardMirroringTest.java
```diff
@@ -0,0 +1,278 @@
...
+       *           ^
+       *           v
+       *           3
+       */
+
+      {
+         AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration(getTestMethodName() + "1to2", "tcp://localhost:" + AMQP_PORT_2).setRetryInterval(100);
```
gemmellr, line 79: Should add a gate on the retries (here and below).

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPNoForwardMirroringTest.java
```diff
@@ -0,0 +1,278 @@
...
+         consumer.close();
+      }
+
+      try (Connection conn = factory3.createConnection()) {
+         Session session = conn.createSession();
+         conn.start();
+         MessageConsumer consumer = session.createConsumer(session.createQueue(getQueueName()));
+         assertNull(consumer.receiveNoWait());
+         consumer.close();
+      }
```
gemmellr, line 275: There are a lot of repeated creating new connections to the same server. I'd be inclined to have a larger try-with-resources that created the connections and then [re]used them as needed.

Not necessarily for the whole test, but where it makes sense and makes things succinct and more efficient without any notable change in behaviour. E.g the 'consume from 1, from 2, from 3....then do the same again to check there is still nothing' all seems like it would be functionally the same without the extra connection creations

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPSquareMirroringTest.java
```diff
@@ -0,0 +1,283 @@
...
+       * |         |       link between 3 and 4 are noForward
+       * v         v       links in both directions.
+       * 4 <- - -> 3
+       */
+
+      {
+         AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration(getTestMethodName() + "1to2", "tcp://localhost:" + AMQP_PORT_2).setRetryInterval(100);
+         amqpConnection.addElement(new AMQPMirrorBrokerConnectionElement().setNoForward(true));
+         server.getConfiguration().addAMQPConnection(amqpConnection);
+         amqpConnection = new AMQPBrokerConnectConfiguration(getTestMethodName() + "1to4", "tcp://localhost:" + AMQP_PORT_4).setRetryInterval(100);
```
gemmellr, line 86: Similarly, should gate retries.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPSquareMirroringTest.java
```diff
@@ -0,0 +1,283 @@
...
+       *
+       * 1 <- - -> 2
+       * ^         ^       The link between 1 and 2 and the
+       * |         |       link between 3 and 4 are noForward
+       * v         v       links in both directions.
+       * 4 <- - -> 3
+       */
+
+      {
+         AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration(getTestMethodName() + "1to2", "tcp://localhost:" + AMQP_PORT_2).setRetryInterval(100);
```
gemmellr, line 83: It would probably be more readable with connection-specific variable names rather than reusing (here and other test classes). There are 8 different connections, with different configurations on top, all with name amqpConnection. There isnt even a space between the 2 in each block here.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPSquareMirroringTest.java
```diff
@@ -0,0 +1,283 @@
...
+      assertNotNull(q1);
+
+      Queue q2 = server.locateQueue(getQueueName());
+      assertNotNull(q2);
+
+      Queue q3 = server.locateQueue(getQueueName());
+      assertNotNull(q3);
+
+      Queue q4 = server.locateQueue(getQueueName());
+      assertNotNull(q4);
```
gemmellr, line 139: These all use the same queue, as they all call on "server", so the rest of the test isnt validating what it thinks when it checks them all.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorConnectionTest.java
```diff
@@ -229,6 +271,173 @@ public void testBrokerAddsAddressAndQueue() throws Exception {
...
+            amqpConnection.addElement(new AMQPMirrorBrokerConnectionElement().addProperty(TUNNEL_CORE_MESSAGES, Boolean.toString(tunneling)).setQueueCreation(true).setAddressFilter(getQueueName() + ",sometest"));
+            server_2.getConfiguration().addAMQPConnection(amqpConnection);
+
+            amqpConnection = new AMQPBrokerConnectConfiguration(getTestMethodName() + "toServer", "tcp://localhost:" + BROKER_PORT_NUM);
+            amqpConnection.addElement(new AMQPMirrorBrokerConnectionElement().setAddressFilter(getQueueName()));
+            server_2.getConfiguration().addAMQPConnection(amqpConnection);
+         }
+
+         {
+            AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration(getTestMethodName() + "toServer2", "tcp://localhost:" + AMQP_PORT_2).setRetryInterval(50);
```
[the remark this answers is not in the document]
lavocatt: We don't really need to set this retryInterval, I prefer to simply remove it.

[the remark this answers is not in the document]
gemmellr: That doesnt change the need for setting a retry limit though.

Its probably better to set the interval as well when setting the limit actually, so that you fully specify and thus control the resulting 'timeout' as it were.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPNoForwardMirroringTest.java
lavocatt: I'm trying to do that, but that makes the test fails. I'll show you next week.

tabish121:
Closing for now as this appears to have been abandoned

PR #5221, opened 2024-09-06
 ARTEMIS-5038 Mirrored ACKs are broken if using multiple priorities on producers

clebertsuconic:
PriorityLinkedList has multiple sub-lists, before this commit PriorityLinkedList::setNodeStore would set the same node store between all the lists.
When a removeWithID was called for an item on list[0] the remove from list[4] would always succeed first. This operation would work correctly most of the time except
when tail and head is being used. Many NullPointerExceptions would be seen while iterating on the list for remove operations, and the navigation would be completely broken.
    
A test was added to PriorityLinkedListTest to make sure the correct lists were used however I was not able to reproduce the NPE condition in that test.
AccumulatedInPageSoakTest reproduced the exact condition for the NPE when significant load is used.

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityLinkedListImpl.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityLinkedListImpl.java
@@ -20,6 +20,7 @@ import java.lang.reflect.Array;
 import java.util.Comparator;
 import java.util.NoSuchElementException;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
+import java.util.function.Supplier;
 
 /**
  * A priority linked list implementation
@@ -39,6 +40,10 @@ public class PriorityLinkedListImpl<E> implements PriorityLinkedList<E> {
    private int highestPriority = -1;
 
    private int lastPriority = -1;
+
+   protected void removed(final int level, final E element) {
+      exclusiveIncrementSize(-1);
+   }
 
    public PriorityLinkedListImpl(final int priorities) {
       this(priorities, null);
@@ -96,9 +101,10 @@ public class PriorityLinkedListImpl<E> implements PriorityLinkedList<E> {
    }
 
    @Override
-   public void setNodeStore(NodeStore<E> supplier) {
+   public void setNodeStore(Supplier<NodeStore<E>> supplier) {
       for (LinkedList<E> list : levels) {
-         list.setNodeStore(supplier);
+         NodeStore<E> nodeStore = supplier.get();
+         list.setNodeStore(nodeStore);
       }
    }
 
@@ -109,7 +115,7 @@ public class PriorityLinkedListImpl<E> implements PriorityLinkedList<E> {
          for (int l = 4; l < levels.length; l++) {
             E removed = levels[l].removeWithID(listID, id);
             if (removed != null) {
-               exclusiveIncrementSize(-1);
+               removed(l, removed);
                return removed;
             }
          }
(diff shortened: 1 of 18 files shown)


jbertram:
Wouldn't we expect problems with concurrent access given that the JavaDoc for the class states:

> This class is not thread safe.

clebertsuconic:
Yeah.. however I recently added some synchronization for the removeWithIDs..


I guess we will have to expand on this... The AckManager is having issues around this.

clebertsuconic:
although I'm kind of confused.. because all the access through the Ackmanager comes through queueImpl which is synchronized...


I will set this to Draft.. 


I may keep these synchronize blocks to be on the safe side.. but I will need to do some investigation.

clebertsuconic:
My user is hitting a scenario with concurrent modification errors. I'm still chasing after these scenarios, but I would like to keep this addition to be on the safe side.

clebertsuconic:
I have now found the exact condition that was being broken at my user.. and I updated the PR for the minimal changes that are required.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/PriorityLinkedListImpl.java
```diff
@@ -40,6 +41,10 @@ public class PriorityLinkedListImpl<E> implements PriorityLinkedList<E> {
 
    private int lastPriority = -1;
 
+   protected void removed(final int level, final E element) {
+      exclusiveIncrementSize(-1);
+   }
```
gemmellr, line 46: Is this new method just for debugging? It doesnt seem to use either of the arguments, which seems odd given in one case new field tracking was added just to pass it.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3070,6 +3053,11 @@ private void internalAddSorted(final MessageReference ref) {
    }
 
    private int getPriority(MessageReference ref) {
+      if (isInternalQueue()) {
+         // if it's an internal queue we need to send the events on their original ordering
+         // for example an ACK arriving before the send on a Mirror..
+         return 4;
+      }
```
gemmellr, line 3060: Is this true of _all_ such queues? They arent all for mirroring.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/LinkedListIterator.java
```diff
@@ -27,6 +27,9 @@ public interface LinkedListIterator<E> extends Iterator<E>, AutoCloseable {
 
    void repeat();
 
+   /** This method is doing exactly what {@link Iterator#remove()} would do, however it will return the removed element being removed. */
+   E removeLastElement();
```
gemmellr, line 31: removeLastElement() seems ambiguous. If its the same as remove(), which operates on the _current_ element in most Iterators, something like removeCurrentElement() element would probably be more straightforward for later users. Or something much more on the nose like removeAndReturn().

tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/brokerConnection/mirror/LogAssert.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.io.File;
+
+import org.apache.activemq.artemis.utils.FileUtil;
+
+public class LogAssert {
+
+   public static void assertServerLogsForMirror(File serverLocation) throws Exception {
+      File log = new File(serverLocation, "log/artemis.log");
+      FileUtil.find(log, l -> l.contains("NullPointerException") || l.contains("AMQ111010"));
```
gemmellr, line 28: This assert method seems like maybe it isnt asserting anything; it ignores the boolean return value from FindUtil.find() ?

PR #5224, opened 2024-09-09
ARTEMIS-5010 Adding Failback scenario to ReplicatedBothNodesMirrorTest

clebertsuconic:


clebertsuconic:
I was testing something...  and didn't find any issues.


However this is a good test addition IMO.. so I will keep it.

PR #5226, opened 2024-09-10
add common tags to meters instead of changing the provided registry

fogninid:
After #4487 an embedded broker is overriding the tagging configuration of all metrics exposed by the embedding application.

Only meters created by the broker should add broker tags

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/MetricsManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/MetricsManager.java
@@ -31,6 +31,7 @@ import io.micrometer.core.instrument.Meter;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Metrics;
 import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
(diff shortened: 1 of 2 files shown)


jbertram:
@fogninid, the underlying problem here is not clear to me. Do you have a test that demonstrates the problem and validates the proposed fix? 

The `io.micrometer.core.instrument.MeterRegistry.Config#commonTags(java.lang.String...)` method is being invoked on a specific instance of `MeterRegistry`. I wouldn't expect the configuration of one `MeterRegistry` instance to impact another. Is your application using this same `MeterRegistry` instance or its own?

fogninid:
See the linked repo. I am forwarding the (composite) meterRegistry created by spring-boot to the embedded broker.
I am not sure how you mean to create a separate registry only for the broker: I would not know how integrate it with the actuators of spring-boot.

jbertram:
@fogninid, thanks for the demo. I understand your use-case now. It's not something I imagined folks would be doing, but it makes sense.

I need to think about a way to write a test for this in the ActiveMQ Artemis test-suite...

jbertram:
I think your solution is solid. Check out PR #5230. I update the commit message and added another commit with a test plus a bit of polishing.

fogninid:
Thanks @jbertram for the quick feedback and polishing it!

PR #5230, opened 2024-09-11
Commits for ARTEMIS-5045

jbertram:


fogninid:
Thanks, the rationale in the commit also describes well my use-case

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/metrics/NettyPooledAllocatorMetrics.java
```diff
@@ -33,53 +34,64 @@ public final class NettyPooledAllocatorMetrics implements MeterBinder {
 
    private static final String BYTES_UNIT = "bytes";
    private final PooledByteBufAllocatorMetric metric;
+   private final Tags tags;
 
-   public NettyPooledAllocatorMetrics(final PooledByteBufAllocatorMetric pooledAllocatorMetric) {
+   public NettyPooledAllocatorMetrics(final PooledByteBufAllocatorMetric pooledAllocatorMetric, final Tags tags) {
       this.metric = pooledAllocatorMetric;
+      this.tags = tags;
```
gemmellr, line 41: Maybe _commonTags_ to make the linkage to where they came from, and how they are used here, clearer?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/plugin/LegacyConfigMetricsPluginTest.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <br>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <br>
```
gemmellr, line 10: Isnt Javadoc so shouldnt have the \<br\> elements, and the URL isn't indented correctly.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/plugin/PassthroughMetricsPluginTest.java
```diff
@@ -0,0 +1,78 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <br>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <br>
```
gemmellr, line 10: Ditto

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/plugin/LegacyConfigMetricsPluginTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+import org.apache.activemq.artemis.core.config.MetricsConfiguration;
+import org.apache.activemq.artemis.core.server.ActiveMQServer;
+import org.apache.activemq.artemis.core.server.metrics.plugins.SimpleMetricsPlugin;
+
+public class LegacyConfigMetricsPluginTest extends MetricsPluginTest {
+
+   @Override
+   protected void configureServer(ActiveMQServer server) {
+      server.getConfiguration().setMetricsConfiguration(new MetricsConfiguration().setPlugin(new SimpleMetricsPlugin().init(null)));
```
gemmellr, line 28: C&P error?

This does not align with that the other test did during its _legacyConfig_ runs (snippet below), but rather matches the prior non-legacy runs, and also exactly matches what the parent non-LegacyConfig MetricsPluginTest is already doing, i.e they are testing the same things.
```
if (legacyConfig) {
   server.getConfiguration().setMetricsPlugin(new SimpleMetricsPlugin().init(null));
} else {
   server.getConfiguration().setMetricsConfiguration(new MetricsConfiguration().setPlugin(new SimpleMetricsPlugin().init(null)));
}
```

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/plugin/PassthroughMetricsPluginTest.java
```diff
@@ -0,0 +1,78 @@
...
+         @Override
+         public MeterRegistry getRegistry() {
+            return meterRegistry;
+         }
+      }));
+   }
+
+   @Test
+   public void testPassthroughMeterRegistry() {
+      final String meterName = "myGauge";
```
gemmellr, line 60: Could perhaps use the test name? E.g getTestMethodName();

jbertram:
Full run of the test-suite looks good. There were a couple of failures, but they look completely unrelated.

PR #5232, opened 2024-09-12
ARTEMIS-4781 Clustering not-large AMQP message can leak Core large message file

jbertram:
When an AMQP message is sent over a cluster bridge it is embedded into a Core message. If the size of the AMQP message is barely beneath the minLargeMessageSize then the Core message in which the AMQP message is embedded will become a large message. The on the bridge target when the embedded AMQP message is extracted from the large Core message it will not be considered "large." In this situation the file for the large Core message will leak.

Thanks to Erwin Dondorp for the test. I renamed and refactored it a bit, but the fundamentals came from Erwin.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
@@ -1136,7 +1136,16 @@ public class ServerSessionPacketHandler implements ChannelHandler {
             currentLargeMessage = null;
 
             try {
-               session.send(session.getCurrentTransaction(), EmbedMessageUtil.extractEmbedded((ICoreMessage) message.toMessage(), storageManager), false, producers.get(senderID), false);
+               Message m = EmbedMessageUtil.extractEmbedded((ICoreMessage) message.toMessage(), storageManager);
+               session.send(session.getCurrentTransaction(), m, false, producers.get(senderID), false);
+
+               /*
+                * The message embedded in the large Core message (e.g. an AMQP message) may, in fact, not be large. If
+                * this is the case then we need to clean up the large message file that's no longer needed.
+                */
+               if (!m.isLargeMessage()) {
+                  message.deleteFile();
+               }
             } catch (Exception e) {
                message.deleteFile();
                throw e;
--- a/artemis-unit-test-support/src/main/java/org/apache/activemq/artemis/utils/Wait.java
+++ b/artemis-unit-test-support/src/main/java/org/apache/activemq/artemis/utils/Wait.java
@@ -128,6 +128,14 @@ public class Wait {
       }
    }
 
+   public static void assertEquals(int size, IntCondition condition, long timeout, long sleepMillis, Supplier<String> messageSupplier) throws Exception {
+      boolean result = waitFor(() -> condition.getCount() == size, timeout, sleepMillis);
+
+      if (!result) {
+         Assertions.assertEquals(size, condition.getCount(), messageSupplier);
+      }
+   }
+
    public static void assertTrue(Condition condition) {
       assertTrue(DEFAULT_FAILURE_MESSAGE, condition);
    }
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/largemessages/AmqpEmbeddedLargeCoreMessageClusterTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/largemessages/AmqpEmbeddedLargeCoreMessageClusterTest.java
@@ -1,0 +1,112 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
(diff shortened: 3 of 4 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/largemessages/AmqpEmbeddedLargeCoreMessageClusterTest.java
```diff
@@ -0,0 +1,86 @@
...
+   @Test
+   public void testAmqpMessageEmbeddedInLargeCoreMessageAcrossClusterBridge() throws Exception {
+      final String address = "address1";
+      for (int i = 0; i < 2; i++) {
+         // start with basic server
+         setupServer(i, true, true);
+
+         // the reproducer does not reproduce without this:
+         // the default is 102400 but this is in the same range as the message sizes here
+         getServer(i).getConfiguration().setJournalFileSize(10 * 1024 * 1024);
```
gemmellr, line 50: Comment could be a little clearer that its the overall test hardness that set 102400 already, the actual default is the same as is being set.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/largemessages/AmqpEmbeddedLargeCoreMessageClusterTest.java
```diff
@@ -0,0 +1,86 @@
...
+              MessageProducer producer = senderSession.createProducer(senderSession.createTopic(address))) {
+            TextMessage msg = senderSession.createTextMessage();
+            // this is the magic number - small enough to be small, but large enough to make the Core message which embeds it large
+            msg.setText("x".repeat(102176));
+            producer.send(msg);
+         }
+
+         receiverConnection.start();
+         assertNotNull(consumer.receive(1000));
+      }
```
gemmellr, line 82: This obviously isnt wrong, but would it not be far more readable just to have a single try-with-resources block that created both connections up front, and then contained a simple sequence of consumer+producer creation and send+receive ?

artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/ServerSessionPacketHandler.java
```diff
@@ -1136,7 +1136,16 @@ private void sendContinuations(final int packetSize,
             currentLargeMessage = null;
 
             try {
-               session.send(session.getCurrentTransaction(), EmbedMessageUtil.extractEmbedded((ICoreMessage) message.toMessage(), storageManager), false, producers.get(senderID), false);
+               Message m = EmbedMessageUtil.extractEmbedded((ICoreMessage) message.toMessage(), storageManager);
```
clebertsuconic, line 1139: shouldn't the fix be to actually keep the message as large instead?

say, if you look here: 
https://github.com/apache/activemq-artemis/blob/257dd86ae28205d8ad57314cbe1cf8b17659aea5/artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/protocol/EmbedMessageUtil.java#L78-L79

we could check on EMBEDDED_TYPE:


if (message.isLargeMessage) then perform the extractLargeMessage


or the encoded version wouldn't be compatible in that case?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/largemessages/AmqpEmbeddedLargeCoreMessageClusterTest.java
jbertram: I just removed the comment.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/largemessages/AmqpEmbeddedLargeCoreMessageClusterTest.java
jbertram: Updated.

artemis-unit-test-support/src/main/java/org/apache/activemq/artemis/utils/Wait.java
```diff
@@ -128,6 +128,14 @@ public static void assertEquals(int size, IntCondition condition, long timeout,
       }
    }
 
+   public static void assertEquals(int size, IntCondition condition, long timeout, long sleepMillis, Supplier<String> messageSupplier) throws Exception {
```
clebertsuconic, line 131: nice addition re the supplier

clebertsuconic:
LGTM

gemmellr:
I ran this through a full test run in CI and the new test failed (twice, out of five runs), on the receive assertion. One suggestion might be it needs a larger timeout, though before the client returns null it drains the link credit so as to ensure there isnt a message available or in flight, so that suggests the broker really told it there was no message to get.

jbertram:
I added a few assertions for intermediate states that should allow the process more time to complete and if it fails it should provide a more specific indication of what went wrong.

gemmellr:
On the last run 3 of the 5 attempts failed, at the first assertion on L86 around adding.

PR #5234, opened 2024-09-18
Commits for ARTEMIS-5048

jbertram:
This is the result I get from running the JMH benchmark locally:
```
Benchmark                                      (length)  Mode  Cnt      Score     Error  Units
Base64Benchmark.benchmarkCustomBase64Decoding         1  avgt   10    128.219 ±   1.955  ns/op
Base64Benchmark.benchmarkCustomBase64Decoding        10  avgt   10   1093.527 ±  23.791  ns/op
Base64Benchmark.benchmarkCustomBase64Decoding       100  avgt   10  11144.289 ± 141.078  ns/op
Base64Benchmark.benchmarkCustomBase64Encoding         1  avgt   10     67.392 ±   0.983  ns/op
Base64Benchmark.benchmarkCustomBase64Encoding        10  avgt   10    564.470 ±  11.583  ns/op
Base64Benchmark.benchmarkCustomBase64Encoding       100  avgt   10   5901.252 ± 335.920  ns/op
Base64Benchmark.benchmarkJavaBase64Decoding           1  avgt   10     34.935 ±   1.856  ns/op
Base64Benchmark.benchmarkJavaBase64Decoding          10  avgt   10     97.194 ±   6.151  ns/op
Base64Benchmark.benchmarkJavaBase64Decoding         100  avgt   10    931.782 ±  35.235  ns/op
Base64Benchmark.benchmarkJavaBase64Encoding           1  avgt   10     37.814 ±   1.403  ns/op
Base64Benchmark.benchmarkJavaBase64Encoding          10  avgt   10     91.547 ±   1.620  ns/op
Base64Benchmark.benchmarkJavaBase64Encoding         100  avgt   10    907.458 ±  38.804  ns/op
```
The raw numbers aren't as important as the relative difference between the Custom and Java implementations. The Java implementation provides better performance and more consistency plus we get to eliminate around 1,600 lines of code.

This is my first time using JMH so anybody with expertise please review and provide feedback.


jbertram:
BTW, full run of the test-suite looks good.

tabish121:
I ran though the benchmark locally and confirm similar performance difference on my test machine.  Likely makes sense as the JDK version will be the better maintained option.

clebertsuconic:
LGTM

gemmellr (commented):
Sometimes its a case of, the move pre-dates the JDK inclusion, or you cant use the JDK version as you still support older JDK versions that dont have it yet, so you put off using it for later. Then it gets forgotten about. 

Other times, the wasnt as fast in old versions but got better over time, so you skip it for a bit. Then again it gets forgotten about.

I'd guess the first one in this case.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/Base64.java
```diff
@@ -1,1619 +1,58 @@
...
-         return Base64._ORDERED_ALPHABET;
-      } else {
-         return Base64._STANDARD_ALPHABET;
+   public static String encodeObject(final Serializable serializableObject) {
+      try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
+           ObjectOutputStream oos = new ObjectOutputStream(baos)) {
+         oos.writeObject(serializableObject);
+         return java.util.Base64.getEncoder().encodeToString(baos.toByteArray());
+      } catch (Exception e) {
+         return null;
```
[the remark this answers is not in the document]
jbertram: Indeed.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/Base64.java
```diff
@@ -1,1619 +1,61 @@
-/**
- * This file is released into the public domain.
```
gemmellr, line 2: The source and binary assembly LICENCE files will both be needing an update due to the removal.

jbertram: See #5240.

PR #5235, opened 2024-09-18
ARTEMIS-5050: misc improvements to 'Broker-to-Broker Connectivity' docs/index

gemmellr:


jbertram:
Update `_book.adoc` as well to keep it in sync?

gemmellr:
I forgot that existed :)

gemmellr:
On first glance, I dont know that I can update it in quite the same way given it includes the files, and there is only 1 file (and splitting it doesnt necessarily make sense). Need to give it more thought tomorrow.

gemmellr:
I pushed a change that moves the includes in _book.adoc around in the same overall sense as the index.adoc. Still not quite exactly the same but a lot closer.

tabish121:
> I pushed a change that moves the includes in _book.adoc around in the same overall sense as the index.adoc. Still not quite exactly the same but a lot closer.

This looks to be about all you can do for that file, probably not time to split the broker connection docs apart but could do that if they grow to large in the future.

PR #5242, opened 2024-09-20
ARTEMIS-5058 Update Qpid protonj2 to v1.0.0-M22

tabish121:
Update to latest release, used in AMQP tests


tabish121:
Ran full CI without any new test failures

PR #5246, opened 2024-09-23
ARTEMIS-5042 don't throw Exception when clearing cluster-connection bindings

jbertram:
Throwing an exception when clearing the bindings when a cluster-connection is closed short-circuits the clearing (and closing) process. This commit fixes that by simply logging the failure to clear and continues on.

No new tests are added with this commit. It relies on existing tests.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
@@ -1621,4 +1621,7 @@ public interface ActiveMQServerLogger {
 
    @LogMessage(id = 224139, value = "Failed to stop bridge: {}", level = LogMessage.Level.ERROR)
    void errorStoppingBridge(String bridgeName, Exception e);
+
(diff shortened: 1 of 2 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/ActiveMQServerLogger.java
```diff
@@ -1621,4 +1621,7 @@ void noQueueIdDefined(org.apache.activemq.artemis.api.core.Message message,
 
    @LogMessage(id = 224139, value = "Failed to stop bridge: {}", level = LogMessage.Level.ERROR)
    void errorStoppingBridge(String bridgeName, Exception e);
+
+   @LogMessage(id = 224140, value = "Clearing bindings on cluster-connection {} failed to remove binding {}: {}", level = LogMessage.Level.WARN)
+   void clusterConnectionFailedToRemovingBindingOnClear(String clusterConnection, String binding, String exceptionMessage);
```
tabish121, line 1626: I'd feel clusterConnectionFailedToRemoveBindingOnClear would be more grammatically correct

jbertram: Fixed.

PR #5258, opened 2024-09-25
Doc improvements

jbertram:


docs/user-manual/large-messages.adoc
```diff
@@ -2,35 +2,68 @@
...
+By default when writing the final bytes to a large message all writes are synchronized to the storage medium.
+This can be configured via `large-message-sync`, e.g.:
 
-For the best performance we recommend using file store with large messages directory stored on a different physical volume to the message journal or paging directory.
+[,xml]
+----
+<configuration...>
+   <core...>
+      ...
+      <large-message-sync>false</large-message-sync>
```
gemmellr, line 62: In the examples above, the value being configured is the same as the noted default. The text here implies the default is true, but this example sets it false. Intended?

docs/user-manual/security.adoc
```diff
@@ -13,7 +28,6 @@ The size of the caches are controlled by the `authentication-cache-size` and `au
 Both default to `1000`.
 
 How long cache entries are valid is controlled by `security-invalidation-interval`, which is in milliseconds.
-Using `0` will disable caching.
```
gemmellr, line 16: Should this have moved up to the cache-size entries above, rather than being removed entirely?

docs/user-manual/configuration-index.adoc
```diff
@@ -379,9 +400,16 @@ If set to "true" the passwords are masked.
 | xref:paging.adoc#max-disk-usage[max-disk-usage]
 | The max percentage of data we should use from disks.
 The broker will block while the disk is full.
-Disable by setting -1.
+Disable by setting `-1`.
 | 90
 
+| xref:paging.adoc#minimum-disk-free[min-disk-free]
+| Min free bytes on disk below which the system blocks or fails clients.
+Supports byte notation like "K", "Mb", "GB", etc.
```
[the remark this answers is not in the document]
jbertram: Fixed here and in all the other places with the same reference.

docs/user-manual/security.adoc
jbertram: Done.

docs/user-manual/large-messages.adoc
jbertram: Fixed, plus added the default explicitly.

PR #5265, opened 2024-09-27
ARTEMIS-5074 Fix encoding of bytes properties as Binary in AMQPMessage

tabish121:
When a bytes property is added to an AMQPMessage and it is then reencoded it will fail without first wrapping the byte array in an AMQP binary as required by the ApplicationProperties section specification defined type allowances.


tabish121:
I've run these changes through CI a few times to validate no new failures

PR #5281, opened 2024-10-02
ARTEMIS-5085 use retry parameters on initial connection

jbertram:
When the Core client attempts to create the initial connection to a broker when initialConnectAttempts > 1 it will adhere to retryInterval, but it will ignore retryIntervalMultiplier & maxRetryInterval. This commit fixes that so that these parameters are taken into account.


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
```diff
@@ -764,7 +764,7 @@ private void failoverOrReconnect(final Object connectionID,
                      failoverRetries++;
                      if (failoverRetryPredicate.test(false, failoverRetries)) {
                         waitForRetry(failoverRetryInterval);
-                        failoverRetryInterval = getNextRetryInterval(failoverRetryInterval);
+                        failoverRetryInterval = serverLocator.getNextRetryInterval(failoverRetryInterval, retryIntervalMultiplier, retryInterval);
```
tabish121, line 767: I'm not clear on why this passes retryInterval in the function's (maxRetryInterval) argument vs passing the configured maxRetryInterval, is that intentional ?

jbertram: It's a bug. Nice catch!

PR #5285, opened 2024-10-04
ARTEMIS-2777 Add core federation and broker connections to the management console

ryeats:
This is a bit out of the blue but basically it has been bugging me how outgoing connections aren't able to be managed in Artemis for anything but Bridges. So I looked into adding something similar primarily for federation but also for the AMQP broker-connections. Simply displaying the configuration didn't seem useful so similar to Bridges i tried to pull up some attribute related to the actual state, for AMQP broker-connections, i exposed if the connection was actually open. For core federations I added visibility into the upstream consumers that are generate from the policy and a message count for messages they have transferred back.

Screenshot 2024-10-04 at 10 36 23 AM

Screenshot 2024-10-04 at 10 26 41 AM

Screenshot 2024-10-04 at 10 24 59 AM

### DIFF
--- a/artemis-commons/src/main/java/org/apache/activemq/artemis/logs/AuditLogger.java
+++ b/artemis-commons/src/main/java/org/apache/activemq/artemis/logs/AuditLogger.java
@@ -2744,4 +2744,61 @@ public interface AuditLogger {
 
    @LogMessage(id = 601781, value = "User {} is getting authorization failure count on target resource: {}", level = LogMessage.Level.INFO)
    void getAuthorizationFailureCount(String user, Object source);
+
+   static void startFederation(Object source) {
+      BASE_LOGGER.startFederation(getCaller(), source);
+   }
+
+   @LogMessage(id = 601782, value = "User {} is starting a federation on target resource: {}", level = LogMessage.Level.INFO)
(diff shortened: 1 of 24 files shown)


jbertram:
I was _just_ thinking about this. This looks like a great contribution.

@tabish121 & @clebertsuconic, is there anything else you think needs to be included in this?

tabish121:
Core Federation is essentially unmaintained (and under tested) and should probably be marked as deprecated for removal in a future release.  Given that I'd say adding management APIs and code for it seems ill advised.

michaelandrepearce:
Core Federation is essentially unmaintained (and under tested)  ?? Why do you say this? I mean i don't see any large bug tickets, and its used in a number of commercial bank systems.

ryeats:
Should I remove the core federation code and mark the core federation API's deprecated and point developers to the AMQP federation then?

ryeats:
Thanks @tabish121 for reimplementing this for broker-connection in https://github.com/apache/activemq-artemis/pull/5347.
 I have tried it out on the latest 2.39 release and am happy with it so I will close this change and the related ticket since i am okay with switching to broker-connection federation for my use cases.

PR #5288, opened 2024-10-08
ARTEMIS-5090 MirrorSNF is not supposed to expire messages

clebertsuconic:


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2550,6 +2552,11 @@ public boolean isExpirationRedundant() {
          return true;
       }
 
+      if (isInternalQueue() && name.toString().startsWith(MIRROR_ADDRESS)) {
+         logger.trace("Mirrored queues are not supposed to expiry. Address={}, Queue={}", address, name);
```
gemmellr, line 2556: Message is confusing, it rather reads like the origin queues shouldnt expire message, not just the mirror SnF queue.

Something like "Mirror SnF queues are not supposed to expire messages. Address={}, Queue={}" would be clearer.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorExpiryQueueTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+      server.getConfiguration().getAcceptorConfigurations().clear();
+      server.getConfiguration().getAcceptorConfigurations().add(newAcceptorConfig(port, "netty-acceptor"));
+      server.getConfiguration().setName(brokerName);
+      server.getConfiguration().setJournalDirectory(server.getConfiguration().getJournalDirectory() + port);
+      server.getConfiguration().setBindingsDirectory(server.getConfiguration().getBindingsDirectory() + port);
+      server.getConfiguration().setPagingDirectory(server.getConfiguration().getPagingDirectory() + port);
+      server.getConfiguration().setJMXManagementEnabled(true);
+      server.getConfiguration().setMessageExpiryScanPeriod(100);
+      server.getConfiguration().setMessageExpiryScanPeriod(5);
```
gemmellr, line 72: double-configuration

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorExpiryQueueTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+   @Test
+   public void testExpiryOnMirrorSNF() throws Exception {
+
+      final long numberOfMessages = 10;
+      ActiveMQServer serverA = createServer(5671, "serverA");
+      ActiveMQServer serverB = createServer(6671, "serverB");
+
+      {
+         AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration("A_to_B", "tcp://localhost:6671").setReconnectAttempts(-1).setRetryInterval(10);
```
gemmellr, line 86: Should now be using infinite reconnects. Set a reasonable limit based on test expectations and the configured interval.

Adding the test name to the broker connection name makes it easier to track which test it is for. Especially if an issue causes leakage to other tests (this has happened several times which is why I'm commenting on these 2 things).

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorExpiryQueueTest.java
```diff
@@ -0,0 +1,140 @@
...
+         AMQPMirrorBrokerConnectionElement replica = new AMQPMirrorBrokerConnectionElement().setType(AMQPBrokerConnectionAddressType.MIRROR).setDurable(true);
+         amqpConnection.addElement(replica);
+         serverA.getConfiguration().addAMQPConnection(amqpConnection);
+      }
+
+      {
+         AMQPBrokerConnectConfiguration amqpConnection = new AMQPBrokerConnectConfiguration("B_to_A", "tcp://localhost:5671").setReconnectAttempts(-1).setRetryInterval(10);
```
gemmellr, line 93: Ditto

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorExpiryQueueTest.java
```diff
@@ -0,0 +1,140 @@
...
+         AMQPMirrorBrokerConnectionElement replica = new AMQPMirrorBrokerConnectionElement().setType(AMQPBrokerConnectionAddressType.MIRROR).setDurable(true);
+         amqpConnection.addElement(replica);
+         serverB.getConfiguration().addAMQPConnection(amqpConnection);
+      }
+
+      String queueName = "queue" + RandomUtil.randomString();
```
gemmellr, line 99: getTestMethodName() would be a nicer base than just "queue".

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorExpiryQueueTest.java
```diff
@@ -0,0 +1,140 @@
...
+      }
+
+      Thread.sleep(10); // waiting a little more than the expiry scan period
+
+      Wait.assertEquals(numberOfMessages, expiryA::getMessageCount, 5000, 100);
+
+      serverB.start();
+
+      Wait.assertEquals(0L, serverAMirrorSNF::getMessageCount, 5000, 100);
+      Wait.assertEquals(0L, serverBMirrorSNF::getMessageCount, 5000, 100);
```
gemmellr, line 137: This test would still pass supposing the messages did expire in the SnF queue, and were never sent to the mirror, but just didnt get added to the expiry queue. It never verifies the second broker actually gets any of the messages.

I think a far better test would be to use a test peer as substitute for the second broker, not give enough credit for the messages to be sent initially, then verify they remain in the SnF queue whilst the ones on the origin queue expire, and then grant credit and verify they are actually received by the peer.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
clebertsuconic: I was doing it for everything internal.. but a test failed.. and a previous JIRA asking exactly that behavior from the internal queue.. so I changed the implementation to deal with only Mirror. 

I just changed the title and commit name.. thanks

clebertsuconic:
@gemmellr I went a further step and I'm never starting a target serverB.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2550,6 +2552,11 @@ public boolean isExpirationRedundant() {
          return true;
       }
 
+      if (isInternalQueue() && name.toString().startsWith(MIRROR_ADDRESS)) {
+         logger.trace("Queue {} is not supposed to expiry as it is a Mirror Store and Forward Queue", this.name);
```
gemmellr, line 2556: The "not supposed to expiry" does not really make sense, which is partly why I changed it to "expire messages" in my previous updated suggestion.

gemmellr:
> @gemmellr I went a further step and I'm never starting a target serverB.

I wouldnt call that further personally. It proves the prior test wasnt vetting the mirror behaviour sufficiently, but still means the test still doesnt really verify the messages that are meant to be mirrored, are. It never even inspects the SnF queue now at all. For all it knows, the messages could have been discarded at expiration rather than routed to the expiry queue, and just as with the previous version of the test it wouldnt notice they are not mirrored like they should be.

I'm assuming it did fail before the fix, so it was useful to verify the fix made a difference, but that doenst really make it a good test of the key behaviour such that the issue doesnt come back in some form. Verifying the actual mirror behaviour with the peer would.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorExpiryQueueTest.java
gemmellr: not sure why you resolve this?

gemmellr:
I don't see it touching the SnF queue at all, can you point me to it? I do see it looking at the expiry queue and checking X messages expire. X messages that were in the origin queue.

My point is that for all the test knows, I break the mirroring again 5 minutes after it i merged, such that we do expire SnF queue messages again, but only such that they get _discarded_ rather than routed to the expiry queue. That test will not notice that and so will still pass, but as a test specifically verifying the mirroring isnt expiring SnF queue messages it really should fail. Setting up a peer to verify the actual mirroring, verifies that it actually does what we expect. That will absolutely notice when it doesnt do it.

clebertsuconic:
@gemmellr verifying that the messages are in the SNF should be enough.. I added an assert for that.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2550,6 +2552,11 @@ public boolean isExpirationRedundant() {
          return true;
       }
 
+      if (isInternalQueue() && name.toString().startsWith(MIRROR_ADDRESS)) {
+         logger.trace("Mirror SNF queues are not suppsoed to expiry messages. Address={}, Queue={}", address, name);
```
gemmellr, line 2556:
```suggestion
         logger.trace("Mirror SNF queues are not supposed to expire messages. Address={}, Queue={}", address, name);
```

clebertsuconic: oops.. thanks

gemmellr:
> @gemmellr verifying that the messages are in the SNF should be enough.. I added an assert for that.

Its certainly infinitely better than not checking at all as previously, and should show the issue.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/connect/AMQPMirrorExpiryQueueTest.java
clebertsuconic: I actually already pushed this.

PR #5289, opened 2024-10-10
ARTEMIS-5096 exclude unnecessary jars from console war

jbertram:
This particular change will be moot when we integrate the new console, but we should do something similar at that point.

### DIFF
--- a/artemis-distribution/src/main/assembly/dep.xml
+++ b/artemis-distribution/src/main/assembly/dep.xml
@@ -72,9 +72,18 @@
          <includes>
             <include>org.apache.activemq:artemis-console:war</include>
          </includes>
-         <outputDirectory>web</outputDirectory>
-         <unpack>false</unpack>
-         <outputFileNameMapping>console.war</outputFileNameMapping>
+         <outputDirectory>web/console.war</outputDirectory>
+         <unpack>true</unpack>
+         <!-- Exclude these jar files from console since they are already provided in the broker distribution.
+              If any of these jars is dropped from the broker distribution they should be added back to the console. -->
+         <unpackOptions>
+            <excludes>
+               <exclude>**/guava*.jar</exclude>
+               <exclude>**/commons-io*.jar</exclude>
+               <exclude>**/commons-logging*.jar</exclude>
+               <exclude>**/failureaccess*.jar</exclude>
+            </excludes>
+         </unpackOptions>
          <useProjectArtifact>false</useProjectArtifact>
          <directoryMode>0755</directoryMode>
          <fileMode>0644</fileMode>
--- a/docs/user-manual/versions.adoc
+++ b/docs/user-manual/versions.adoc
@@ -20,6 +20,12 @@ https://issues.apache.org/jira/secure/ReleaseNote.jspa?projectId=12315920&versio
 
 * The `two-way` algorithm of the default codec is deprecated, use a custom codec instead.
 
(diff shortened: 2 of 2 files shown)


artemis-hawtio/artemis-console/pom.xml
```diff
@@ -76,9 +76,14 @@
...
           <useCache>true</useCache>
+          <!-- exclude these jars as they are already shipped with the broker -->
           <packagingExcludes>
             **/lib/slf4j*.jar,
-            **/lib/log4j*.jar
+            **/lib/log4j*.jar,
+            **/lib/guava*.jar,
+            **/lib/commons-io*.jar,
+            **/lib/commons-logging*.jar,
+            **/lib/failureaccess*.jar
```
gemmellr, line 86: The comment added above this is "\<!-- exclude these jars as they are already shipped with the broker --\>"

I think we need to be very clear what is meant by 'shipped with the broker' here, and in turn what it is is being expected/required that a user might need to provide to make the resulting war usable in their situation.

The broker `artemis-server` module does not depend on or provide guava or commons-io, and neither does the `artemis-web` module this war is typically used via. The assembly distribution does include those things, but only due to _other_ included bits bringing them in, not the broker. So if we completely remove those deps, they wont in fact be provided necessarily, i.e this actually is likely to actually break non-assembly use cases.

Previously we have entirely stripped some logging deps, but the _required_ API dep was certain to provided, either by the modules the war is most likely to be used with, or by the application itself (as is a logging implementation unless they have none configured). That isnt necessarily true for the other deps like these.

jbertram: I changed up the strategy. Instead of dealing with this when building `console.war` I am dealing with it when creating the broker distribution. That should eliminate any problems with non-assembly use-cases.

artemis-distribution/src/main/assembly/dep.xml
```diff
@@ -72,9 +72,18 @@
...
+         <!-- Exclude these jar files from console since they are already provided in the broker distribution.
+              If any of these jars is dropped from the broker distribution they should be added back to the console. -->
+         <unpackOptions>
+            <excludes>
+               <exclude>**/guava*.jar</exclude>
+               <exclude>**/commons-io*.jar</exclude>
+               <exclude>**/commons-logging*.jar</exclude>
+               <exclude>**/failureaccess*.jar</exclude>
+            </excludes>
+         </unpackOptions>
```
gemmellr, line 86: So if I understand, this effectively removes the console.war file from the assembly, and extracts its [trimmed] contents instead, i.e puts a directory called console.war in its place? (EDIT: Or I could have read further before commenting, and saw the doc explaining that is what it does, hehe :D)

Do we need to do anything with the config for that changing? Its geared around war files from memory?

@brusdev any thoughts around the switch?

jbertram: To be clear, there are some minor differences under the covers, but unexploded and exploded war files are functionally equivalent in Jetty.

brusdev: I'm testing it locally to check what is the impact for users that want to replace console.war with the hawtio.war.

brusdev: It LGTM and the exploded war provides an easy way to customize files like web.xml

brusdev (approved):

PR #5290, opened 2024-10-10
ARTEMIS-5097 Allow export and import data of undefined queues

clebertsuconic:
This will allow users eventually undoing mistakes after removing a queue by accident.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/xml/XmlDataExporter.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/xml/XmlDataExporter.java
@@ -80,6 +80,11 @@ public final class XmlDataExporter extends DBOption {
 
    private XMLStreamWriter xmlWriter;
 
+   private Throwable lastError;
+
+   @Option(names = "undefined-prefix", description = "In case a queue does not exist, this will define the prefix to be used on the message export. Default: 'UndefinedQueue_'")
+   private String undefinedPrefix = "UndefinedQueue_";
+
    // an inner map of message refs hashed by the queue ID to which they belong and then hashed by their record ID
    private final Map<Long, HashMap<Long, ReferenceDescribe>> messageRefs = new HashMap<>();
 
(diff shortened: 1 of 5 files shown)


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/tools/xml/XmlDataExporter.java
```diff
@@ -80,6 +80,12 @@ public final class XmlDataExporter extends DBOption {
 
    private XMLStreamWriter xmlWriter;
 
+   private Throwable lastError;
+
+   // I am creating a hidden feature now, in case users ever ask to expose this, it would be already here.
+   @Option(names = "undefined-prefix", hidden = true, description = "In case a queue does not exist, this will define the prefix to be used on the message export.")
```
jbertram, line 86: Why make it hidden? Seems like a nice feature that users would appreciate.

clebertsuconic: @jbertram I wasn't sure if it would make sense.



I was also thinking.. if there's just one queue missing, I would use just the prefix without concatenating anything

jbertram: Well, I think it makes sense, and I also think it's worth a full explanation of the use-case in the documentation (which I plan on writing after this is merged).

I wouldn't change the handling if there's just one queue missing. That's just more complexity to explain & maintain without much real benefit.

clebertsuconic: @jbertram I don't think we should make this an use-case.. users are not really supposed to remove the queues...


what I'm doing this is just not failing an export/import in case it ever happened.. say after a recovered data for instance.


BTW: I'm not going to change the handling of the prefix... a manual undelete will be good enough I think.

PR #5291, opened 2024-10-10
ARTEMIS-5093 support configurable onMessage timeout w/closing consumer

jbertram:
There wasn't a clear place to add documentation about this so I'm relying on JavaDoc.


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
```diff
@@ -1298,6 +1298,18 @@ public ServerLocatorImpl setInitialMessagePacketSize(final int size) {
       return this;
    }
 
+   @Override
+   public int getOnMessageCloseTimeout() {
+      return config.onMessageCloseTimeout;
+   }
+
+   @Override
```
tabish121, line 1306: The API docs specify the value set must be greater than zero, but there isn't any validation of that, should there be?

jbertram: Hmmm...That comment was left-over from a copy & paste. I went looking through the existing code to see how validation is done for other parameters, and I'm not actually seeing any. I'm continuing my investigation.

tabish121: I did look into what it does with that configuration value and since it is passing into a CountdownLatch is seems like it will not trigger an exception but will return immediately so from the standpoint of being able to be set to zero or negative it works, although seems an absurd value to assign.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/MessageHandlerTest.java
```diff
@@ -117,6 +118,42 @@ public void onMessage(final ClientMessage message) {
       session.close();
    }
 
+   @Test
+   public void testMessageHandlerCloseTimeout() throws Exception {
+      // create Netty acceptor so client can use new onMessageCloseTimeout URL parameter
+      server.getRemotingService().createAcceptor("netty", "tcp://127.0.0.1:61616").start();
+      final int timeout = 1000;
```
brusdev, line 125: does the test work with a lower timeout to reduce the test duration?

coocoo1112:
Hey sorry, I was just wondering if there was any update on this or plans to get it into a new version?

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
gemmellr: It would seem weird to configure those if actually knowing thats what it then does, but I'd guess some might think one of those would simply mean _keep waiting_ (e.g as is the described behaviour in the JMS spec, albeit that also originally described to effectively deadlock yourself as result hehe), especially if not reading the javadoc (should there also be actual-docs addition?).

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/MessageHandlerTest.java
```diff
@@ -117,6 +118,42 @@ public void onMessage(final ClientMessage message) {
...
+            } catch (InterruptedException e) {
+               // ignore
+            }
+         }
+      });
+      latch.await();
+      long start = System.currentTimeMillis();
+      consumer.close();
+      long end = System.currentTimeMillis();
+      assertTrue( (end - start >= timeout) && (end - start <= 2000), "Closing consumer took " + (end - start) + "ms");
```
gemmellr, line 154: I'd be inclined to use a shorter timeout for the close as Dom suggests, thus reducing the test duration, but also use a longer/indefinite 'dont complete yet' guard in the handler to be sure it is always still running after/when the timeout expires and can then 'immediately complete' on request, e.g instead arrange its waiting/exit-gating on an [atomic]boolean and other latch to wait on the rest of the test then complete quickly on request rather than the current time loop.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientConsumerImpl.java
```diff
@@ -911,10 +916,10 @@ private void waitForOnMessageToComplete(boolean waitForOnMessage) {
 
       sessionExecutor.execute(future);
 
-      boolean ok = future.await(ClientConsumerImpl.CLOSE_TIMEOUT_MILLISECONDS);
+      boolean ok = future.await(onMessageCloseTimeout);
```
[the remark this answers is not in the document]
jbertram: No, it is not still required.

brusdev (approved):

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
jbertram: I changed the JavaDoc to be more clear. I didn't want to add validation as it wouldn't be consistent with the rest of the method and adding validation to all the places that could realistically use it is well beyond the scope of this PR.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/MessageHandlerTest.java
jbertram: Done.

gemmellr: This actually seems more brittle now than it was when I commented on it originally.

Faster than before, yes (though could still be faster, if not just relying on friendly scheduling), but also more brittle as a result of now needing separately scheduled things to always happen <600ms apart, rather than <1000ms before.

Governing the exit point of the handler rather than just the time would entirely avoid the brittle reliance on nice scheduling.

jbertram: > Governing the exit point of the handler rather than just the time would entirely avoid the brittle reliance on nice scheduling.

I'm not following here. Can you elaborate?

jbertram: As far as I can tell the test is essentially relying on `java.util.concurrent.CountDownLatch#await(long, java.util.concurrent.TimeUnit)` (eventually called from `org.apache.activemq.artemis.api.core.client.ClientConsumer#close`) for its accuracy/reliability. Is that really reliant on nice scheduling?

jbertram: I push a change. Let me know what you think.

gemmellr: You stopped the thread hanging around much longer than the checks, which is good...however the checks still remain as brittle to scheduling as they were before the change.

The main point is surely that it did not exit during the close timeout, and that the close takes at least the timeout. So you control the handler thread cant exit while those checks are happening, but then exits immediately as soon as they have completed, once you allow it to. The 800ms magic window doesnt need to be there.

Is it likely to take that long? No. Though could it happen now and then, on e.g a slow CI env with low thread count and shared resources? Yes, when someone then needs to go look at the sporadic failure. Or it adds to the many that have sporadically failed and encouraged people not to look.

If you want to use a timeout in the handler that fair enough, but I'd make it a lot longer so it can never reasonably happen unless the test has hung elsewhere.

It also needs add a try-finally to ensure the gating boolean is tripped to have it exit immediately even if the asserts somehow failed.

jbertram: I think I've got it now.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/MessageHandlerTest.java
```diff
@@ -117,6 +119,49 @@ public void onMessage(final ClientMessage message) {
...
+         latch.countDown();
+         // don't just Thread.sleep() here because it will be interrupted on ClientConsumer.close()
+         while (!messageHandlerFinished.get()) {
+            try {
+               Thread.sleep(10);
+            } catch (InterruptedException e) {
+               // ignore
+            }
+         }
+         messageHandlerFinished.set(true);
```
gemmellr, line 150: I would use a different boolean for the looping and the completion checking, it would be much clearer, and also means the set(true) doesnt then seem like dead-code (it can only ever execute here if it is already true).

The messageHandlerFinished.set(true); should also go in a try-finally so that the later assertFalse on it is guarding against unexpected-exits as well as the sole expected one being set up by the test.

You could then also wait for it to trip before the test method exits. Maybe using a CDL instead of an AtomicBoolean actually to make that succinct and efficient.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/client/MessageHandlerTest.java
gemmellr: Thats better, though caused me to realise something about it not guarding agains unexpected-exits. I added a new comment on that and readability, so I'll resolve this one.

PR #5294, opened 2024-10-11
ARTEMIS-5101 Deprecate default codec two-way algorithm

brusdev:
Throughout the years, the standard mechanism for storing passwords has evolved. In the beginning, passwords were stored in plaintext. Developers are now encouraged to leverage adaptive one-way functions to store a password. Using a two-way function by default for storing passwords without a warning could lead users to a false sense of security.


clebertsuconic:
LGTM

PR #5307, opened 2024-10-18
ARTEMIS-4167 Enhanced deserialization filter

swerner0:
Now that Artemis is Java 11+ compatible, there is now the ability to set an ObjectInputFilter on an ObjectInputStream. This allows us to write more advanced filter patterns or plug-in our own object filter implementation.

### DIFF
--- a/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ObjectInputFilterFactory.java
+++ b/artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ObjectInputFilterFactory.java
@@ -1,0 +1,143 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements. See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.jms.client;
+
+import java.io.ObjectInputFilter;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class ObjectInputFilterFactory {
+
+   public static final String SERIAL_FILTER_PROPERTY            = "org.apache.activemq.artemis.jms.serialFilter";
+   public static final String SERIAL_FILTER_CLASS_NAME_PROPERTY = "org.apache.activemq.artemis.jms.serialFilterClassName";
+
+   private static Map<Object, ObjectInputFilter> filterCache = new ConcurrentHashMap<>();
+
+   public static ObjectInputFilter getObjectInputFilter(ConnectionFactoryOptions options) {
+      String className = getFilterClassName(options);
+      if (className != null) {
+         return getObjectInputFilterForClassName(className);
+      }
+
+      String pattern = getFilterPattern(options);
+      if (pattern != null) {
+         return getObjectInputFilterForPattern(pattern);
+      }
+
+      return null;
+   }
+
+   public static ObjectInputFilter getObjectInputFilterForPattern(String pattern) {
+      if (pattern == null) {
+         return null;
+      }
+
+      return filterCache.computeIfAbsent(new PatternKey(pattern),
+                                         k -> ObjectInputFilter.Config.createFilter(pattern));
+   }
+
+   public static ObjectInputFilter getObjectInputFilterForClassName(String className) {
+      if (className == null) {
+         return null;
+      }
+
+      return filterCache.computeIfAbsent(new ClassNameKey(className), k -> {
+         try {
+            return (ObjectInputFilter)Class.forName(className).newInstance();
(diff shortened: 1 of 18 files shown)


docs/user-manual/security.adoc
```diff
@@ -1431,6 +1431,16 @@ comma separated values for allow list
 
 These properties, once specified, are eventually set on the corresponding internal factories.
 
+=== Filtering using built-in JVM support
+
+Now that Apache ActiveMQ Artemis requires a minimum JVM version of 11, built-in Java serialization filtering mechanisms can be utilized.
+Instead of providing an `allow list` or `deny list`, you can specify either a `serialFilter` or `serialFilterClassName`.
+
+* `serialFilter` - A pattern based filter that allows you to define allow/deny lists and constraints limiting graph complexity and size. https://docs.oracle.com/en/java/javase/17/core/serialization-filtering1.html#JSCOR-GUID-8296D8E8-2B93-4B9A-856E-0A65AF9B8C66[Filter Syntax]
+* `serialFilterClassName` - For those who need a custom filtering solution, you can supply an implementation of https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/ObjectInputFilter.html[ObjectInputFilter]
```
jbertram, line 1440: The links here should likely refer to the Java 11 documentation (e.g. https://docs.oracle.com/en/java/javase/11/core/serialization-filtering1.html).

jbertram: A simple example and a blurb to compare & contrast this approach against the existing approach would help users choose what's best for their use-case.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ObjectInputFilterFactory.java
```diff
@@ -0,0 +1,143 @@
...
+
+import java.io.ObjectInputFilter;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class ObjectInputFilterFactory {
+
+   public static final String SERIAL_FILTER_PROPERTY            = "org.apache.activemq.artemis.jms.serialFilter";
+   public static final String SERIAL_FILTER_CLASS_NAME_PROPERTY = "org.apache.activemq.artemis.jms.serialFilterClassName";
```
jbertram, line 27: Neither of these system properties are documented as far as I can tell.

docs/user-manual/security.adoc
swerner0: Sorry, running 17 locally so is the one I'm always referencing. Will update the link.

swerner0: I can add a compare and contrast

docs/user-manual/security.adoc
```diff
@@ -1431,6 +1431,16 @@ comma separated values for allow list
 
 These properties, once specified, are eventually set on the corresponding internal factories.
 
+=== Filtering using built-in JVM support
+
+Now that Apache ActiveMQ Artemis requires a minimum JVM version of 11, built-in Java serialization filtering mechanisms can be utilized.
+Instead of providing an `allow list` or `deny list`, you can specify either a `serialFilter` or `serialFilterClassName`.
```
gemmellr, line 1437: I dont really see a need to include 'Now that Apache ActiveMQ Artemis requires a minimum JVM version of 11'. It conveys nothing particularly useful or necessary about the functionality.  Its also just going to go stale. 

Simply stating that  you can pass a filter string / class name to leverage the built in ObjectInputFilter support is all thats needed.

docs/user-manual/security.adoc
```diff
@@ -1431,6 +1431,16 @@ comma separated values for allow list
...
+
+* `serialFilter` - A pattern based filter that allows you to define allow/deny lists and constraints limiting graph complexity and size. https://docs.oracle.com/en/java/javase/17/core/serialization-filtering1.html#JSCOR-GUID-8296D8E8-2B93-4B9A-856E-0A65AF9B8C66[Filter Syntax]
```
gemmellr, line 1439: Just as Justin didnt really like "serialFilter" originally on the earlier PR (#4368), I still cant say I am a fan of it.

The more fully elaborated deserializationFilter name would seem more obvious to me.

docs/user-manual/security.adoc
gemmellr: Including any details around the impact (e.g precedence) if more than one of the now-several options is specified would also be good. Or else saying, and enforcing, that you cant specify the different approaches at the same time.

docs/user-manual/security.adoc
swerner0: I agree, but I was trying to stay consistent with the Oracle system wide property of `jdk.serialFilter`

docs/user-manual/security.adoc
swerner0: Do you want to continue to support the allow/denyList and custom inputstream? We can say they are deprecated in favor of the new at some point, and list an order of execution if both are specified, although that would be frowned upon

gemmellr: Regardless how quickly we were to remove the old one, both will be there until then and so what it does or allows / not should be clear and documented. Its not even clear currently what would happen between the 2 new options, or their system properties, let alone once you consider any mix with the older existing stuff.

Even if we deprecate the old, I dont see us removing the old bits particularly quickly (or ever for 2.x) given it is all anyone has ever used, and in many cases will likely continue to use right up until it goes away. We also just deprecated the original names to replace with allow/deny, so would seems especially unfair to then quickly _require_ another update for the same functionality for anyone that did already adapt to that.

docs/user-manual/security.adoc
gemmellr: First time that has been mentioned in >18 months, even when you agreed with Justin and said you were renaming it hehe. 

I'm not sure that should be a goal personally. Switching to deserializationFilter from deserializationAllowList and deserializationDenyList seems more obvious, and would be more consistent between themselves whilst both are around.

Thoughts @jbertram ?

PR #5316, opened 2024-10-23
ARTEMIS-5131 Add A Copy message button to console

andytaylor:
This exposes a copyMessage method that simply copies a message to a different queue so the new console can add a copy button

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
@@ -1362,6 +1362,32 @@ public class QueueControlImpl extends AbstractControl implements QueueControl {
    }
 
    @Override
+   public boolean copyMessage(final long messageID,
+                              final String targetQueue) throws Exception {
+      // this is a critical task, we need to prevent parallel tasks running
+      try (AutoCloseable lock = server.managementLock()) {
+         if (AuditLogger.isBaseLoggingEnabled()) {
+            AuditLogger.copyMessage(queue, messageID, targetQueue);
+         }
+         checkStarted();
+
+         clearIO();
+         try {
+            Binding binding = server.getPostOffice().getBinding(SimpleString.of(targetQueue));
+
+            if (binding == null) {
+               throw ActiveMQMessageBundle.BUNDLE.noQueueFound(targetQueue);
+            }
+
+            return queue.copyReference(messageID, binding.getAddress(), binding);
+         } finally {
+            blockOnIO();
+         }
+      }
+
+   }
+
+   @Override
    public int moveMessages(final String filterStr, final String otherQueueName) throws Exception {
       return moveMessages(filterStr, otherQueueName, false);
    }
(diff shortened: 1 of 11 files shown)


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/QueueControlTest.java
```diff
@@ -2618,6 +2618,60 @@ public void testMoveMessage() throws Exception {
       session.deleteQueue(otherQueue);
    }
 
+   @TestTemplate
```
tabish121, line 2621: For adding a feature like this you should ideally write a test that covers more of the protocols and messages (normal or large) as well as times when paging is happening vs not as there have been a number of cases lately where a breakage or misbehavior has been found when an AMQP or a Large message was involved in internal management code such as this. 

Refer to test: tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/plugin/BeforeSendPluginTest.java for some inspiration

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
```diff
@@ -1361,6 +1361,32 @@ public boolean moveMessage(final long messageID,
...
+
+         clearIO();
+         try {
+            Binding binding = server.getPostOffice().getBinding(SimpleString.of(targetQueue));
+
+            if (binding == null) {
+               throw ActiveMQMessageBundle.BUNDLE.noQueueFound(targetQueue);
+            }
+
+            return queue.copyReference(messageID, binding.getAddress(), binding);
```
clebertsuconic, line 1382: The implementation details here should be part of the Queue::copyReference.


Also... what if the targetQueue was passed a topic.

Shouldn't this call PostOffice::route instead?...

on that case targetQueue here should be targetAddress...




I started writing a test to validate this test and I bumped into that

clebertsuconic:
I sent a PR for a test that's not working at the moment, but it would give you an idea of a test to be written...


as I said before I think the method should be updated.

clebertsuconic:
@andytaylor also.. I just thought... how security handled on the security settings on the copy?


I don't think this is checking security.. is it?


and I still think it should go through the PostOffice  so you can copy for a topic for instance?

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
andytaylor: As with the move method this is adding directly to a queue, rerouting it would send it to all queus on an address which is wrong. I could do use route and add rout-to-id of the queue but would there be any benefit to doing that?

andytaylor:
> @andytaylor also.. I just thought... how security handled on the security settings on the copy?
> 
> I don't think this is checking security.. is it?
> 
> and I still think it should go through the PostOffice so you can copy for a topic for instance?

I don't sww how the topic instance makes any difference as it is just being moved between queues?

andytaylor:
I have added tests for each protocol, large messages and some around paging. one paging test copies to a queue on the same address and makes sure it is also paged

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/QueueControlImpl.java
andytaylor: actually ignore, if you follow the code thru it does use rout on the binding itself so it only goes to that queue

andytaylor:
rebased against main to get back on the radar

clebertsuconic:
LGTM

PR #5320, opened 2024-10-28
ARTEMIS-5110: mark IDs retired that were removed previously

gemmellr:
Marks various additional IDs retired to prevent their reuse, based on previous removals/reassignments.

Tweaks the regex from ActiveMQAMQPProtocolMessageBundle instead to prevent using 200+ IDs that have already been duplicated or reused historically and would be again.

We can squash the changes on push but I found it helpful having specific commits, in turn referencing the original commits they related to.


gemmellr:
Added 4 more commits, retiring some other previosuly-used IDs I missed from not initially looking prior to file renames done after the HQ donation in 2014.

As before, we can push the changes in a single squashed commit (e.g https://github.com/gemmellr/activemq-artemis/commit/4af3f5e3bbde8fd5bdb4d954cc47cc3b58b9c13c is currently equivalent), but for review and when making the changes I found the individual commits nicer to work with.

gemmellr:
I pushed the single-change update in https://github.com/apache/activemq-artemis/commit/8a912ed209d7c952994908ad5df3ea22ebfa844e 

(Same as earlier noted commit, just rebased and with the PR ref added to the commit log to close the PR out)

PR #5327, opened 2024-10-30
ARTEMIS-5119 Expired Messages on Cluster SNF should to to the original Expiry Queue

clebertsuconic:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/expiry/ClusteredExpiryTest.java
```diff
@@ -0,0 +1,115 @@
...
+         MessageProducer producer = session1.createProducer(session1.createQueue("queues." + getName()));
+         producer.setTimeToLive(2_000);
+         for (int i = 0; i < NUMBER_OF_MESSAGES; i++) {
+            producer.send(session1.createTextMessage("hello"));
+         }
+         session1.commit();
+      }
+      Wait.assertEquals(0L, serverQueue0::getMessageCount, 50_000, 100);
+      Wait.assertEquals(0L, snfPaused::getMessageCount, 50_000, 100);
+      Queue expiryQueue = servers[0].locateQueue("Expiry", "EXP.queues." + getName() + ".Expiry");
```
gemmellr, line 100: It would be a lot easier to follow this test if there were variables for the prefix, suffix, and 'main queue name' values that were then used to set up the configuration and composed for the other usages such as this (maybe via another variable).  Especially given the slightly overloaded use of "Expiry" in many of them.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/expiry/ClusteredExpiryTest.java
```diff
@@ -0,0 +1,115 @@
...
+      waitForBindings(1, "queues." + getName(), 1, 0, false);
+
+      // pausing the SNF queue to keep messages stuck on the queue
+      servers[0].getPostOffice().getAllBindings().filter(f -> f.getUniqueName().toString().startsWith("$.artemis.internal.sf")).forEach(this::pauseQueue);
+      assertNotNull(snfPaused);
+
+      long NUMBER_OF_MESSAGES = 100;
+
+      ConnectionFactory factory = CFUtil.createConnectionFactory("CORE", "tcp://localhost:61616");
+      ConnectionFactory factory2 = CFUtil.createConnectionFactory("CORE", "tcp://localhost:61617");
```
gemmellr, line 88: Is factory2 unused as it looks?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/ExpireQueueSuffixTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+         producer = session.createProducer(session.createQueue(queueB.toString()));
+         producer.setTimeToLive(100);
+         for (int i = 0; i < sendB; i++) {
+            producer.send(session.createTextMessage("queueB"));
+         }
+         session.commit();
+      }
+
+      Wait.waitFor(() -> server.locateQueue(expiryAddress.toString(), "EXP." + queueA + expirySuffix) != null, 5000);
```
gemmellr, line 89: Hehe, yes, on my earlier comment about using variables for clarity...something a lot like this. Could have saved myself some typing by reading ahead and referencing this hehe.

(A variable could perhaps be added for this whole name to make it easier to follow)

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2147,10 +2151,71 @@ public void expire(final MessageReference ref, final ServerConsumer consumer, bo
...
 
+   private void expire(final Transaction tx, final MessageReference ref, boolean delivering) throws Exception {
+      if (logger.isDebugEnabled()) {
+         logger.debug("Expiry on {}, expiryAddress={}", this.address, addressSettings.getExpiryAddress());
+      }
+
+      AddressSettings settingsToUse = getMessageAddressSettings(ref.getMessage());
+      SimpleString expiryAddress = settingsToUse.getExpiryAddress();
+
+      if (expiryAddress != null && expiryAddress.length() != 0) {
```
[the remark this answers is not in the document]
clebertsuconic: It was not part of my change.. it was already there before.. I only moved the expire method closer to the other expire method to make it easier to follow (we used to section private and public... and in this case I thought they should be together).

But I will make the change on the other method.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/expiry/ClusteredExpiryTest.java
```diff
@@ -0,0 +1,117 @@
...
+         Session session1 = connection.createSession(true, Session.SESSION_TRANSACTED);
+         MessageProducer producer = session1.createProducer(session1.createQueue(queuesPrefix + getName()));
+         producer.setTimeToLive(500);
+         for (int i = 0; i < NUMBER_OF_MESSAGES; i++) {
+            producer.send(session1.createTextMessage("hello"));
+         }
+         session1.commit();
+      }
+      Wait.assertEquals(0L, serverQueue0::getMessageCount, 50_000, 100);
+      Wait.assertEquals(0L, snfPaused::getMessageCount, 50_000, 100);
```
gemmellr, line 101: Do we need 50 seconds timeout, with the expiry reaper set at 10ms ?

clebertsuconic: I needed while debugging and forgot to revert it... will make it 5 seconds

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2129,14 +2130,26 @@ public void expire(final MessageReference ref) throws Exception {
     *           hence no information about delivering statistics should be updated. */
    @Override
    public void expire(final MessageReference ref, final ServerConsumer consumer, boolean delivering) throws Exception {
-      if (addressSettings.getExpiryAddress() != null) {
-         createExpiryResources();
+      AddressSettings settingsToUse = getMessageAddressSettings(ref.getMessage());
```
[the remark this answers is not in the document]
gemmellr: Yeah I wasn't suggesting to combine them as they did seem sufficiently different on the backend to justify the two methods. Its just the initial bits were very similar so it would have been nice for them to be set out the same way for readability.

Combining them seems to have rearranged at least the order of some stuff for one method, or added some stuff around bindings that other method didnt do before...I'd need to take a much fuller look to see what I think of those changes, but I need to finish up just now.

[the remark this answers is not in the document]
clebertsuconic: @gemmellr since you mentioned you are looking to review this.. I will leave this PR open until next week.. no rush about merging this yet.. although it would be nice before next release.. but it's not happening next week anyways... so I will leave it open for now.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2129,36 +2130,85 @@ public void expire(final MessageReference ref) throws Exception {
...
 
-         move(null, addressSettings.getExpiryAddress(), null, ref, false, AckReason.EXPIRED, consumer, null, delivering);
+         Bindings bindingList = postOffice.lookupBindingsForAddress(expiryAddress);
+
+         if (bindingList == null || bindingList.getBindings().isEmpty()) {
+            ActiveMQServerLogger.LOGGER.errorExpiringReferencesNoBindings(expiryAddress);
+            acknowledge(tx, ref, AckReason.EXPIRED, null, delivering);
+         } else {
+            move(tx, expiryAddress, null, ref, false, AckReason.EXPIRED, consumer, null, delivering);
+         }
```
gemmellr, line 2161: Until the most recent commit updates, before you combined the existing 'expire' method with your then-added new 'expireFromScan' method, the existing 'expire' did not do this check and would simply try the move. (Or if there was no expireAddress defined at all, it would log _once_ that there wasnt any.) This seems like it changes things by introducing a potential per-message warning log message about the bindings. Was that desired, given you didn't add it originally when the methods were seperate?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -2129,36 +2130,85 @@ public void expire(final MessageReference ref) throws Exception {
...
+            // print this only once
+            ActiveMQServerLogger.LOGGER.errorExpiringReferencesNoAddress(name);
+         }
 
-         acknowledge(null, ref, AckReason.EXPIRED, consumer, delivering);
+         acknowledge(tx, ref, AckReason.EXPIRED, consumer, delivering);
       }
 
-      // potentially auto-delete this queue if this expired the last message
-      refCountForConsumers.check();
```
gemmellr, line 2147: The existing 'expire' method with consumer always did this, before checking the broker plugins. It still did this when you had added the 'expireWithScan' method. Now that you combined the methods, this is only done if there is a tx, and only after checking the broker plugins (and only if there is a message plugin, even though this is a queue-level check). These both seem strange differences from just consolidating some methods. Was one of them wrong before?

clebertsuconic:
@gemmellr I added a commit to be squashed with a change...

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
gemmellr: You updated the behaviour a bit, bit it still differs from before in the sense that for the combined expire method it only runs the refCountForConsumers.check(); if there is a broker message plugin, rather than always doing it before for this old expire method (the other old expire method only did it if there is a plugin). Was it wrong to always check it in this method before, i.e was the other method correct not to do it without a plugin? Essentially, why should it need a broker message plugin to do it?

gemmellr:
Left a new comment in one of the previous comment 'threads' https://github.com/apache/activemq-artemis/pull/5327#discussion_r1872954781

clebertsuconic:
I compared the older methods. and I think they are ok now.


Testsuite is green... I will merge this.. if there is any issues let me know and I will address it then

PR #5333, opened 2024-11-01
ARTEMIS-5141 Add or remove plugins on configuration reload

AntonRoskvist:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/RedeployTest.java
```diff
@@ -233,6 +235,64 @@ public void testRedeployConnector() throws Exception {
...
+                       .anyMatch(plugin ->
+                                    plugin instanceof LoggingActiveMQServerPlugin));
+
+         Files.copy(url1.openStream(), brokerXML, StandardCopyOption.REPLACE_EXISTING);
+         brokerXML.toFile().setLastModified(System.currentTimeMillis() + 1000);
+         latch.setCount(1);
+         embeddedActiveMQ.getActiveMQServer().getReloadManager().setTick(tick);
+         latch.await(10, TimeUnit.SECONDS);
+
+         assertEquals(1, embeddedActiveMQ.getActiveMQServer().getBrokerPlugins().size());
```
brusdev, line 287: Why is `embeddedActiveMQ.getActiveMQServer().getBrokerPlugins().size() == 1` after the configuration reload?

brusdev (commented):
The method `registered` of the reload broker plugins is not invoked.

The existing broker plugins in the config are unregistered and registered on the config reload. Is that on purpose?

AntonRoskvist:
> The method `registered` of the reloaded broker plugins is not invoked.
> 
> The existing broker plugins in the config are unregistered and registered on the config reload. Is that on purpose?

That is not intentional, thanks for catching that @brusdev

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/RedeployTest.java
AntonRoskvist: I added some comments clarifying this.

I start the broker without any plugins.
I add one plugin programmatically (which should not be affected by following configuration reloads).
I add one plugin through configuration, then remove the same plugin through configuration.

The final assert is then to verify that the programmatically added plugin stayed registered.

brusdev (commented):
The existing broker plugins in the config are unregistered and registered each time the the config is reloaded. This could be a breaking change because the current version only load broker plugins at startup.

You can check this behavior adding the same plugin to both reload-plugin.xml and reload-plugin-updated.xml files, i.e. https://github.com/brusdev/activemq-artemis/commit/3115925478c64d0c3e65d3837a5c44eab3a7faeb

AntonRoskvist:
Yes I see it, I'll have to take a closer look at this. Thanks @brusdev

jbertram:
@AntonRoskvist, do you plan to continue working on this or should I close it?

AntonRoskvist:
@jbertram I've been meaning to, but I'm currently knee deep in other work and have been for some time. While I'm expecting this to get better at some point it might be a while still... I think this can be closed and I can revisit it at a later time instead.

PR #5340, opened 2024-11-12
ARTEMIS-5150: ActiveMQServerControlImpl.getHAPolicy() gets NPE

gaohoward:
When you start the broker without configuring HAPolicy, if clients call this management api it'll get NPE because it doesn't check if the HAPolicyConfiguration is null.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
@@ -75,6 +75,7 @@ import org.apache.activemq.artemis.core.config.BridgeConfiguration;
 import org.apache.activemq.artemis.core.config.Configuration;
 import org.apache.activemq.artemis.core.config.ConnectorServiceConfiguration;
 import org.apache.activemq.artemis.core.config.DivertConfiguration;
+import org.apache.activemq.artemis.core.config.HAPolicyConfiguration;
 import org.apache.activemq.artemis.core.config.TransformerConfiguration;
 import org.apache.activemq.artemis.core.filter.Filter;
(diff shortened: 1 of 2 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -834,7 +835,12 @@ public int  getAddressMemoryUsagePercentage() {
 
    @Override
    public String getHAPolicy() {
-      return configuration.getHAPolicyConfiguration().getType().getName();
+      HAPolicyConfiguration haConfig = configuration.getHAPolicyConfiguration();
```
gtully, line 838: should a default not be in the configuration rather than in the server control.
I think primary only is a sensible default, but I would also expect an embedded broker to provide configuration values that it cares about rather than depend on defaults. Defaults will only cover some subset at some time.

gtully: It should avoid the NPE, and maybe return null if the policy is null.

gaohoward: +1, returning null would reflect the truth (without depending on default behavior)

gtully (commented):
With just the change to check for null, would a junit test for ...ServerControlImpl make sense rather than the embedded use case test?

gaohoward:
> With just the change to check for null, would a junit test for ActiveMQServerControlImpl make sense rather than the embedded use case test?

Makes sense. I think I can move the test to org.apache.activemq.artemis.tests.integration.management.ActiveMQServerControlTest

gtully (approved):
looks good

PR #5344, opened 2024-11-13
ARTEMIS-5110: tests for ID validations in log annotation processor

gemmellr:
Add testing of various validations/suggestions introduced in previous changes for the Jira. Also improves an error message to report the LogBundle class like others do.


tabish121:
I ran the tests after reviewing, all looks good.

PR #5346, opened 2024-11-14
ARTEMIS-5155 Send of AMQP large messages should close the LargeMessageReader before proceeding with the message delivery

clebertsuconic:
This is fixing LargeMessageInterruptTest

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonAbstractReceiver.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonAbstractReceiver.java
@@ -345,25 +345,23 @@ public abstract class ProtonAbstractReceiver extends ProtonInitializable impleme
                           Message message, DeliveryAnnotations deliveryAnnotations) {
       connection.requireInHandler();
 
-      try {
-         receiver.advance();
+      receiver.advance();
 
-         Transaction tx = null;
-         if (delivery.getRemoteState() instanceof TransactionalState) {
-            TransactionalState txState = (TransactionalState) delivery.getRemoteState();
-            try {
-               tx = this.sessionSPI.getTransaction(txState.getTxnId(), false);
-            } catch (Exception e) {
-               this.onExceptionWhileReading(e);
-            }
+      Transaction tx = null;
+      if (delivery.getRemoteState() instanceof TransactionalState) {
+         TransactionalState txState = (TransactionalState) delivery.getRemoteState();
+         try {
+            tx = this.sessionSPI.getTransaction(txState.getTxnId(), false);
+         } catch (Exception e) {
+            this.onExceptionWhileReading(e);
          }
-
-         actualDelivery(message, delivery, deliveryAnnotations, receiver, tx);
-      } finally {
-         // reader is complete, we give it up now
-         this.messageReader.close();
-         this.messageReader = null;
       }
+
+      //  messageReader.close should be called before the delivery, to signal the reader is complete.
+      this.messageReader.close();
+      this.messageReader = null;
+
+      actualDelivery(message, delivery, deliveryAnnotations, receiver, tx);
    }
 
    @Override


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonAbstractReceiver.java
```diff
@@ -346,6 +346,7 @@ public void onMessageComplete(Delivery delivery,
       connection.requireInHandler();
 
       try {
+         this.messageReader.close();
```
gemmellr, line 349: Can you elaborate the actual issue and how this close moving here makes a difference? Looking at the large message reader, its close method, and how/when the reader calls this onMessageComplete() method, I'm not really seeing how this would alter the behaviour to fix something?

clebertsuconic: @gemmellr leaving the close to happen after the advance would allow a situation where we confirm the client the message has been accepted. Later the server crashed, even after a commit (depending on how fast it happened) and the file wasn't yet closed.. leaving a truncated file in the file system...


After a restart the file still zeroed because the buffer that contained the file before the close is gone.

clebertsuconic: 2 days to find this... not a very fun experience..


I thought the BytesMessage was something else leaking to the client.

clebertsuconic: Actually.. I'm wrong...


I found this based on the bisect.. since the problem is the ingestion this is the only part that changed between the commit that introduced this and after...


I'm puzzled now after I read the close method... 


it was the only change between the offending commit and after (on the producer's server's side).

tabish121: That doesn't make a ton of sense though as the onMessageCompleted is call only after the large message release resources has been called meaning the large message should have already been "closed" unless you mean something other than closed when you are saying closed.  Also the close in the try / finally block is a fail safe to ensure that if an error occurs on that path then things actually do get closed, which again should have already been done.  We need a bit more detail on this as currently it simply doesn't seem right as I'd have expected that close call to be a no-op as AMQPLargeMessageReader would check state of 'currentMessage' and see that it was already null as it is nulled before onMessageCompleted is ever called.

clebertsuconic: the test started to faill after c83ed8957d9d7f06bb29c0fce563fe2e3462993e

tabish121: Are you sure the test has truly stopped failing? Perhaps moving it is simply affecting timing as there would be a few volatile reads that could alter minor timing sensitive cases.  Really understanding what is going on here is key because any looking though the code so far leads me to understand that the close() call will simply check and see that the large message was already released and nulled before the code that calls onMessageComplete in AMQPLargeMessageReader::addBytes

clebertsuconic: It did stop failing.  Many repetitions. 


I think there is a failure check where I delete the file.    Moving the close reader before the advance stopped it from happening.

clebertsuconic: I will confirm is the failure check later today (I'm out of my desk at the moment)

tabish121: First thing I would do is confirm that the close call in the onMessageComplete is actually doing anything as by default it would not be doing anything other than toggling the boolean 'closed' as the currentMessage would be null and no interactions with the large message would occur there.

clebertsuconic: @tabish121 @gemmellr the issue is indeed the close after a failure removing the file.


this method here gets closed on a network error, disconnect... etc:

https://github.com/apache/activemq-artemis/blob/ffafbf350af6f92f7984cbc251be8f2d5dfcf2d4/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonAbstractReceiver.java#L369-L374



If I don't close the reader before the advance I could get to a race where the message was already confirmed and this close could eventually happen.


Closing the reader before the advance fixed the issue.



I will add a big // COMMENT block explaining the situation here.

clebertsuconic: it wasn't very pleasant to find such tiny thing to be honest :)  (just blaming myself.. lol as I made the change)

clebertsuconic: The outcome is a test failure on InterruptLargeMessageTest. This commit https://github.com/apache/activemq-artemis/commit/c83ed8957d9d7f06bb29c0fce563fe2e3462993e  broke the test as it is not setting the AMQPMessageReader as closed when it's supposed. as a consequence dropped connections will make the failureCallbacks to remove a "pending" large message which is not actually pending on this case.

clebertsuconic: the issue is actually the call to actualDelivery ...if we perform actualDelivery before calling messageReader.close() you could have the message being part of the queue while the reader will be marked as open. a connection failure happens and the body of the message removed. as a consequence the client will receive an empty body message which the server is translating as a bytesMessage.

clebertsuconic: 200 passes with testInterruptLargeMessageAMQPTXPaging with this change on the reader.close();


reverted the fix... the test failed within 2 retries.

clebertsuconic:
this change is reverting the offending change from c83ed8957d9d7f06bb29c0fce563fe2e3462993e

As I said before, marking the LargeMessageReader as complete after the delivery of the message of the queue could cause the message body to go truncated, if a network failure happened between the delivery and the actual close of the message (aka race).


I'm merging this PR now as this is clearly an issue and this fixes that issue.

gemmellr:
I changed it to draft, as we have been looking into why this change might have any effect, as it didnt seem to either myself or Tim that it should have the desired impact, and possibly pointed to a potential concurrency problem. Whilst moving this might indeed help adjust the behaviour to stop a test failing, it still isnt clear to either of us it is actually the correct change given the point the close is done. Tim may have found a related issue around the delivery handling offloading large messages to another thread.

In short, we still need to look at this a bit closer before merging.

clebertsuconic:
@gemmellr I don't think this is a draft.. it's definitely fixing the issue.. it's a small issue..



last packet arrived for the message
deliverMessage
close reader

another thread detects a failure... removed the file


consequence is an empty / truncated file

clebertsuconic:
@gemmellr  all I'm fixing now is.. when closeCurrentReader is called, it's deleting the file...


which is what I'm fixing.. if you want to improve other things in large message.. it's totally orthogonal to this.. this change is fixing the issue.

gemmellr:
We are saying it may have been wrong the way it was as well, so just putting it back that way is not necessarily correct. Still looking at it further. In the meantime, this has been this way for 8 months, lets look at it before merging rather than after.

clebertsuconic:
I think the close should happen before onMessageComplete, on the caller thread.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/ProtonAbstractReceiver.java
```diff
@@ -345,25 +345,23 @@ public void onMessageComplete(Delivery delivery,
...
+         try {
+            tx = this.sessionSPI.getTransaction(txState.getTxnId(), false);
+         } catch (Exception e) {
+            this.onExceptionWhileReading(e);
+         }
       }
+
+      //  messageReader.close should be called before the delivery, to signal the reader is complete.
+      this.messageReader.close();
+      this.messageReader = null;
```
tabish121, line 362: This null assignment should remain in a finally block as it ensures that the reader is always nulled when onMessageComplete gets called.

clebertsuconic:
I have closed this in favor of Tim's PR.

PR #5347, opened 2024-11-14
ARTEMIS-5157 Add management for AMQP broker connections with federation

tabish121:
Add management views for AMQP broker connections and implement control types for AMQP federation features along with the broker connection management. Some initial work also to provide support for other broker connection features to add management control and also plan for future views of incoming broker connections.

A sample view of the new elements in the console are shown in the image below. A broker connection entry appears in the tree and under it the federation and its associated address and queue policies are visible along with any active remote consumers managed by the policies.
Screenshot from 2024-11-14 16-37-51


artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/BrokerConnectionControlImpl.java
```diff
@@ -0,0 +1,177 @@
...
+   }
+
+   @Override
+   public boolean isStarted() {
+      if (AuditLogger.isBaseLoggingEnabled()) {
+         AuditLogger.isStarted(this.brokerConnection);
+      }
+      clearIO();
+      try {
+         return brokerConnection.isStarted();
```
gemmellr, line 45: The mismatch of "this.brokerConnection" and then just "brokerConnection" seems weird when reading, be nicer if consistent.

(here and below)

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/BrokerConnectionControlImpl.java
```diff
@@ -0,0 +1,177 @@
...
+         brokerConnection.start();
+      } finally {
+         blockOnIO();
+      }
+   }
+
+   @Override
+   public void stop() throws Exception {
+      if (AuditLogger.isBaseLoggingEnabled()) {
+         AuditLogger.startBrokerConnection(this.brokerConnection.getName());
```
gemmellr, line 80: c&p error, this is calling startBrokerConnection but is in the stop() method.

tabish121:
Fixed those issues in latest push

PR #5355, opened 2024-11-18
ARTEMIS-5155 Race on AMQP large message read and close

tabish121:
When the final frame of a large message is being written to the file in the session thread and an IO error occurs such that that connection is torn down, the large message reader can be closed before the message is fully processed resulting in corruption. The large message file close logic needs to occur on the session thread so that the processing of the bytes can finish and the message gets added to the Queue and the close can react by not deleting the file when it runs following the read task.

### DIFF
--- a/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
+++ b/artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/proton/AMQPLargeMessageReader.java
@@ -17,6 +17,8 @@
 
 package org.apache.activemq.artemis.protocol.amqp.proton;
 
+import java.lang.invoke.MethodHandles;
+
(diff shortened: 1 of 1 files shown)


tabish121:
Creating a test that reliably reproduced this wasn't possible but it does express itself in occasional failures in the tests in 'LargeMessageInterruptTest' which has not been seen after this change in CI jobs which have been executed several times.

clebertsuconic:
I think we should still merge https://github.com/apache/activemq-artemis/pull/5346


the close of the reader should be done before delivering the message, regardless. 

when I moved the close to finalize I thought it was about closing the file... as it stands that method should still be called before the actualDelivery.

tabish121:
> I think we should still merge #5346
> 
> the close of the reader should be done before delivering the message, regardless.
> 
> when I moved the close to finalize I thought it was about closing the file... as it stands that method should still be called before the actualDelivery.

Since that change is essentially just a cosmetic change it is orthogonal to this and can be done if you want but will have no real effect on the underlying threading race that is occurring here as I've outlined previously.

PR #5360, opened 2024-11-20
NO-JIRA: have dependabot ignore jgroups until JGRP-2794 is resolved

AntonRoskvist:
I have seen some compatibility issues with rolling upgrades between the past few broker releases because of JGRP-2794.


jbertram:
I think this deserves a Jira.

AntonRoskvist:
I'll make a Jira, thanks @jbertram

PR #5363, opened 2024-11-21
ARTEMIS-5170 Fix AutoCreateExpiryResourcesTest and document expiry of expiry

clebertsuconic:
…

### DIFF
--- a/docs/user-manual/message-expiry.adoc
+++ b/docs/user-manual/message-expiry.adoc
@@ -76,6 +76,13 @@ If `min-expiry-delay` is not defined then the message will not be changed.
 
 The default for both `min-expiry-delay` and `max-expiry-delay` is `-1` (i.e. disabled).
 
+[WARNING]
+====
+**If you set expiry-delay, or min/max-expiry-delay, on the expiration target address beware of the following:**
+
+* Messages will get a new expiration when moved to the expiry queue, rather than being set to 0 as usual, and so may disappear after the new expiration.
+====
+
 == Configuring Expiry Addresses
 
 Expiry address are defined in the address-setting configuration:
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/AutoCreateExpiryResourcesTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/AutoCreateExpiryResourcesTest.java
@@ -60,6 +60,9 @@ public class AutoCreateExpiryResourcesTest extends ActiveMQTestBase {
       // set common address settings needed for all tests; make sure to use getMatch instead of addMatch in invidual tests or these will be overwritten
       server.getAddressSettingsRepository().addMatch("#", new AddressSettings().setAutoCreateExpiryResources(true).setExpiryAddress(expiryAddress).setExpiryDelay(EXPIRY_DELAY));
 
+      // avoiding expiration of the expiration
+      server.getAddressSettingsRepository().addMatch(String.valueOf(expiryAddress), new AddressSettings().setExpiryDelay(-1L));
+
       server.start();
    }
 
@@ -129,7 +132,7 @@ public class AutoCreateExpiryResourcesTest extends ActiveMQTestBase {
    }
 
    private void testAutoCreatedExpiryFilter(RoutingType routingType) throws Exception {
-      final int ITERATIONS = 50;
+      final int ITERATIONS = 5;
       final int MESSAGE_COUNT = 5;
 
       for (int i = 0; i < ITERATIONS; i++) {
@@ -149,6 +152,11 @@ public class AutoCreateExpiryResourcesTest extends ActiveMQTestBase {
          locator.close();
          Wait.assertTrue(() -> server.locateQueue(getDefaultExpiryQueueName(address)) != null, 2000, 10);
          Queue expiry = server.locateQueue(AddressSettings.DEFAULT_EXPIRY_QUEUE_PREFIX.concat(address).concat(AddressSettings.DEFAULT_EXPIRY_QUEUE_SUFFIX));
+
+         // This sleep is to make the message would survive in the Expiry Address after the default delay
+         // to catch any eventual misconfiguration
+         Thread.sleep(EXPIRY_DELAY);
+
          Wait.assertEquals(MESSAGE_COUNT, expiry::getMessageCount);
       }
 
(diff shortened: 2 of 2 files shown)


clebertsuconic:
I really liked this...


can I ask a small change.. if you can't do it.. I will do it later...


if the user is null.. just don't set the user


If I have a broker without an user (anonymous user).. it's currently showing null@tcp://localhost:61616 > 


I think it should use just tcp://localhost:61616 >


although the check that you have on the Exception would need to be different.. perhaps just use a boolean to connected instead.

clebertsuconic:
It's currently showing this, when null:


null@tcp://localhost:61616 >

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/AutoCreateExpiryResourcesTest.java
```diff
@@ -59,6 +59,7 @@ public void setUp() throws Exception {
 
       // set common address settings needed for all tests; make sure to use getMatch instead of addMatch in invidual tests or these will be overwritten
       server.getAddressSettingsRepository().addMatch("#", new AddressSettings().setAutoCreateExpiryResources(true).setExpiryAddress(expiryAddress).setExpiryDelay(EXPIRY_DELAY));
+      server.getAddressSettingsRepository().addMatch(String.valueOf(expiryAddress), new AddressSettings().setExpiryDelay(-1L));
```
gemmellr, line 62: I'm actually curious why it is setting expiry delay on the regular address settings to begin with, and whether that should instead of adding this? 

Most of the tests look to send messages that have expiration and are already-expired, so the setting shouldnt affect them after this change.

The exceptions seem to be testAutoCreatedExpiryFilterAnycast / testAutoCreatedExpiryFilterMulticast that seem to send messages without expiration. Could it just change to setting one, and then remove the ExpiryDelay setting here?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/AutoCreateExpiryResourcesTest.java
```diff
@@ -194,6 +195,8 @@ public void testWithJMSFQQN() throws Exception {
 
       triggerExpiration();
 
+      Thread.sleep(EXPIRY_DELAY);
+
```
gemmellr, line 199: Is this actually needed? If so maybe a comment why given the other tests dont do it? Given the triggerExpiration method before seems to send an already-expired message, and the later receive here already has a timeout.

docs/user-manual/message-expiry.adoc
```diff
@@ -76,6 +76,14 @@ If `min-expiry-delay` is not defined then the message will not be changed.
 
 The default for both `min-expiry-delay` and `max-expiry-delay` is `-1` (i.e. disabled).
 
+[IMPORTANT]
+.Warning
+====
+**If you set expiry-delay on the target expiration target address beware of the following:**
+
+* Messages will get a new expiration when moved, and may disappear after the new expiration.
+====
```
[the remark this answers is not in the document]
clebertsuconic: this was my very first asciidoc :) thanks!


I had the impression I needed to do it that way (the WARNING thing)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/AutoCreateExpiryResourcesTest.java
clebertsuconic: There are other tests in this test class that will use this:

https://github.com/apache/activemq-artemis/blob/14c564a48144b0e052c6e0eeee64ba0f94744335/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/AutoCreateExpiryResourcesTest.java#L136-L138


I will use a prefix rather than everything.

clebertsuconic: It would be a complete rewrite of the test.. I will just go with the extra line.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/AutoCreateExpiryResourcesTest.java
clebertsuconic: I actually added a comment why

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/AutoCreateExpiryResourcesTest.java
```diff
@@ -129,7 +130,7 @@ public void testAutoCreatedExpiryFilterMulticast() throws Exception {
    }
 
    private void testAutoCreatedExpiryFilter(RoutingType routingType) throws Exception {
-      final int ITERATIONS = 50;
+      final int ITERATIONS = 5;
```
clebertsuconic, line 133: no need for 50 iterations.. I'm lowering it down to save a few seconds.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/server/AutoCreateExpiryResourcesTest.java
clebertsuconic: other tests are failing in the same situation if I don't set this..


I had added more loops and more sleeps to other tests and they also failed.


I would rather keep this rather to rewrite this test.

clebertsuconic: I could also keep running them in loop to add individually on every tests needed.. so I would rather just keep it.

clebertsuconic: I will mark it as resolved.. please unmark it if you disagree.

gemmellr: Fair enough, you had only mentioned the FQQN one so I thought it was the only one actually needing this (which is why I didnt like changing it for all tests).

clebertsuconic:
I will merge this... we can do further improvements on following up PRs if needed.

PR #5376, opened 2024-11-28
ARTEMIS-5187 Fix management authorization checks after authentication failures

brusdev:
When the ArtemisRbacMBeanServerBuilder class is used for the RBAC management, a clash of authentication cache keys between clients failing authentication and web console authenticated users can cause web console authenticated users to receive authorization errors, blank screens and similar issues after successful login to the console.


gtully (approved):
it makes sense to avoid the cache when using the current subject.

PR #5384, opened 2024-12-05
ARTEMIS-4173 Improving reliability on RandomReattachTest

clebertsuconic:
-  making the test to run faster and removing the soak and stress versions as they are not really needed
- allowing expected exceptions (cannot find ACK, and making it retry the operation)


clebertsuconic:
I'm merging this as soon as my tests are done... tests are showing extremely good results..

if anyone has any issues with this merge, please let me know and we can review it

jbertram:
It's hard to tell what semantic change you actually made. :)

clebertsuconic:
@jbertram I removed the testBase and I made it a test.. I guess it makes difficult to see the diffs...


the main diff is this:

https://github.com/apache/activemq-artemis/blob/3d196ad70c0a9db145fae003877950fc4e144077/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/reattach/RandomReattachIntegrationTest.java#L228-L253



there's a check of expectedException.... if it's an expected exception it will just skip and increase the numIts.



I also closed the SessionFactory, and locators and stop the server since there was a leak between the tests.




We could go further and just ignore any exception to an individualACK however that's not really wise.. since not that many people really use reattachment, and if this happens users should still treat it.

PR #5390, opened 2024-12-10
ARTEMIS-5142 support never expiring incoming messages

jbertram:
@gemmellr, take a look at this as a potential replacement for #5334. I feel like this is much simpler and more intuitive.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
@@ -1355,26 +1355,42 @@ public class PostOfficeImpl implements PostOffice, NotificationListener, Binding
       return status;
    }
 
-   // HORNETQ-1029
-   private static void applyExpiryDelay(Message message, AddressSettings settings) {
+   protected static void applyExpiryDelay(Message message, AddressSettings settings) {
       long expirationOverride = settings.getExpiryDelay();
 
-      // A -1 <expiry-delay> means don't do anything
-      if (expirationOverride >= 0) {
-         // only override the expiration on messages where the expiration hasn't been set by the user
+      if (settings.isNoExpiry()) {
+         if (message.getExpiration() != 0) {
+            message.setExpiration(0);
+            message.reencode();
+         }
+      } else if (expirationOverride >= 0) {
+         // A -1 <expiry-delay> means don't do anything
          if (message.getExpiration() == 0) {
-            message.setExpiration(System.currentTimeMillis() + expirationOverride);
+            // only override the expiration on messages where the expiration hasn't been set by the user
+            setExpiration(message, expirationOverride);
          }
       } else {
          long minExpiration = settings.getMinExpiryDelay();
          long maxExpiration = settings.getMaxExpiryDelay();
 
-         if (maxExpiration != AddressSettings.DEFAULT_MAX_EXPIRY_DELAY && (message.getExpiration() == 0 || message.getExpiration() > (System.currentTimeMillis() + maxExpiration))) {
-            message.setExpiration(System.currentTimeMillis() + maxExpiration);
+         if (message.getExpiration() == 0) {
+            // if the incoming message has NO expiration then apply the max if set and if not set then apply the min if set
+            if (maxExpiration != AddressSettings.DEFAULT_MAX_EXPIRY_DELAY) {
+               setExpiration(message, maxExpiration);
+            } else if (minExpiration != AddressSettings.DEFAULT_MIN_EXPIRY_DELAY) {
+               setExpiration(message, minExpiration);
+            }
+         } else if (maxExpiration != AddressSettings.DEFAULT_MAX_EXPIRY_DELAY && message.getExpiration() > (System.currentTimeMillis() + maxExpiration)) {
+            setExpiration(message, maxExpiration);
          } else if (minExpiration != AddressSettings.DEFAULT_MIN_EXPIRY_DELAY && message.getExpiration() < (System.currentTimeMillis() + minExpiration)) {
-            message.setExpiration(System.currentTimeMillis() + minExpiration);
+            setExpiration(message, minExpiration);
          }
       }
+   }
+
+   private static void setExpiration(Message m, long expiration) {
+      m.setExpiration(System.currentTimeMillis() + expiration);
+      m.reencode();
    }
 
    @Override
--- a/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
+++ b/tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
@@ -1,0 +1,275 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.artemis.tests.integration.jms.multiprotocol;
+
+import javax.jms.BytesMessage;
+import javax.jms.Connection;
+import javax.jms.MessageConsumer;
+import javax.jms.MessageProducer;
+import javax.jms.Queue;
+import javax.jms.Session;
+import java.lang.invoke.MethodHandles;
+
+import org.apache.activemq.artemis.core.settings.impl.AddressSettings;
+import org.apache.activemq.artemis.utils.RandomUtil;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.Timeout;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class JMSMessageExpiryTest extends MultiprotocolJMSClientTestSupport {
+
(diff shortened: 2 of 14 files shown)


jbertram:
@gemmellr, if you're good with this then I'll add more tests for the new `never-expire` address-setting.

artemis-server/src/main/java/org/apache/activemq/artemis/core/postoffice/impl/PostOfficeImpl.java
```diff
@@ -1355,28 +1355,42 @@ private RoutingStatus maybeSendToDLA(final Message message,
...
-         if (maxExpiration != AddressSettings.DEFAULT_MAX_EXPIRY_DELAY && (message.getExpiration() == 0 || message.getExpiration() > (System.currentTimeMillis() + maxExpiration))) {
-            message.setExpiration(System.currentTimeMillis() + maxExpiration);
+         if (message.getExpiration() == 0) {
+            // if the incoming message has NO expiration then apply the max if set and if not set then apply the min if set
+            if (maxExpiration != AddressSettings.DEFAULT_MAX_EXPIRY_DELAY) {
+               message.setExpiration(getExpirationToSet(maxExpiration));
+            } else if (minExpiration != AddressSettings.DEFAULT_MIN_EXPIRY_DELAY) {
+               message.setExpiration(getExpirationToSet(minExpiration));
+            }
+         } else if (maxExpiration != AddressSettings.DEFAULT_MAX_EXPIRY_DELAY && message.getExpiration() >= (System.currentTimeMillis() + maxExpiration)) {
```
gemmellr, line 1382: If not special casing 0 to unset expiration, can go back to  > rather than >= here. No need to change it if its equal.

gemmellr:
You mentioned adding more tests...this is one I actually do think should have integration tests as well. 

In particular, as I think about it more, I guess its likely there is a gotcha in there for AMQP messages (both for this and the existing min/max etc). I would not be surprised to see the broker apply the new expirations 'locally' but then send the original expiration to the consumer...so e.g if you _increase_ or _remove_ an original expiration, things may not turn out at all as intended.

jbertram:
@gemmellr, I added some integration tests and problem which I fixed. Let me know what you think.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
```diff
@@ -0,0 +1,197 @@
...
+         if (delayType == DelayType.MIN) {
+            p.setTimeToLive(EXPIRY_DELAY / 2);
+         } else if (delayType == DelayType.MAX) {
+            p.setTimeToLive(EXPIRY_DELAY * 2);
+         } else if (delayType == DelayType.NEVER) {
+            p.setTimeToLive(EXPIRY_DELAY);
+         }
+      }
+      long start = System.currentTimeMillis();
+      p.send(s.createMessage());
```
gemmellr, line 178: Would be better to create the message then record start, since the period of interest begins with the send itself.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
```diff
@@ -0,0 +1,197 @@
...
+      p.close();
+      MessageConsumer c = s.createConsumer(q);
+      connection.start();
+      Message m = c.receive(100);
+      assertNotNull(m);
+      if (delayType == DelayType.NEVER) {
+         assertEquals(0, m.getJMSExpiration());
+      } else {
+         long duration = System.currentTimeMillis() - start;
+         long delayOnMessage = m.getJMSExpiration() - System.currentTimeMillis();
```
gemmellr, line 188: Could also just call System.currentTimeMillis() once after the receive and record it so both duration and delayOnMessage are more precisely calculated from the same reference.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
```diff
@@ -0,0 +1,197 @@
...
+         assertTrue(delayOnMessage >= (EXPIRY_DELAY - duration - WINDOW));
```
gemmellr, line 189: Feels like WINDOW shouldnt really be doing anything (at least after using same reference for both calcs above) since you are already deducting a 'duration' which is including the consumer receive (and an assertNotNull) which necessarily makes it >= the broker-side elapsed time which could have influenced the actual expiration stamped by the broker, i.e you should already be deducting more than enough with just that?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
```diff
@@ -0,0 +1,197 @@
...
+   @Timeout(30)
+   public void testAmqpMessageExpiryDelay() throws Exception {
+      testExpiry(createConnection(), DelayType.NORMAL, false);
+   }
+
+   @Test
+   @Timeout(30)
+   public void testOpenWireMessageExpiryDelay() throws Exception {
+      testExpiry(createOpenWireConnection(), DelayType.NORMAL, false);
+   }
```
gemmellr, line 60: Given the differences in handling of 'large' and 'normal' messages I think both need covered in at least 1 test. Also, given there is explicit handling of expiration in the persisting and reload of durable messages, I think a test which restarts the broker before consumption is probably needed for both also.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class JMSMessageExpiryTest extends MultiprotocolJMSClientTestSupport {
+
+   protected static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+   private final long EXPIRY_DELAY = 9999999L;
```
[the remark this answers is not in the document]
jbertram: Fixed.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
jbertram: Fixed.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
jbertram: I recall that being necessary for tests to pass at one point. I guess I changed something along the way that eliminated the need for it.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
jbertram: Added.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
```diff
@@ -0,0 +1,197 @@
...
+         } else if (delayType == DelayType.NEVER) {
+            p.setTimeToLive(EXPIRY_DELAY);
+         }
+      }
+      long start = System.currentTimeMillis();
+      p.send(s.createMessage());
+      p.close();
+      MessageConsumer c = s.createConsumer(q);
+      connection.start();
+      Message m = c.receive(100);
```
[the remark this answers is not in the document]
jbertram: Fixed.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
gemmellr: Looks like the additions only cover Large messages and a restart together. Given the runtime and persistence/reload handling of the messages differs significantly, specifically around expiration, we should have a restart test for a 'normal' message as well, and non-restarting test with a 'large' message.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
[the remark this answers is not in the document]
gemmellr: 500ms still seems quite short, with potential to be brittle in slow shared CI envs...an unfortunately timed GC, scheduling lottery, disk slowness, or combinations thereoff, seem like they could blow that limit. Its typical to use at least 1-2 seconds, often more, in cases we absolutely expect a message to arrive.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
jbertram: In the most recent push I added three new tests using a large message (one for each protocol) and three new tests restarting the broker. The large message tests don't restart the broker and the tests restarting the broker don't use a large message.

gemmellr: Yeah, I got it reversed when reading then later commenting and shoudl ahve said 'seperately', so also reverse the resulting points that there are still gaps in the combination testing (as you just outlined). Ones a bus is likely to fit through given the near completely seperate handling of these things for the two message types.

jbertram: I added 3 more tests that use a large message and restart the broker. That should cover all the permutations.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/jms/multiprotocol/JMSMessageExpiryTest.java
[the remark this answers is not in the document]
jbertram: Test is using `1500` now.

PR #5395, opened 2024-12-11
ARTEMIS-5210 bump com.google.errorprone:error_prone_core from 2.31.0 to 2.36.0

jbertram:


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/bean/MetaBean.java
```diff
@@ -123,10 +123,9 @@ public void parseToJSON(T object, JsonObjectBuilder builder, boolean ignoreNullA
                }
             } else if (type == Boolean.class) {
                builder.add(name, (Boolean) value);
+            /* This is the same as the default, but keeping this in case we have to deal with Enums in the future
             } else if (Enum.class.isAssignableFrom(type)) {
-               // I know this is the same as the default else clause further down
-               // but i wanted to have a separate branch in case we have to deal with it later
-               builder.add(name, String.valueOf(value));
+               builder.add(name, String.valueOf(value));*/
```
gemmellr, line 128: Ick. Maybe better just deleted at this point?

clebertsuconic:
What these other changes have to do with the component upgrade?


Can they be separate commits with a better description?

jbertram:
They are required because running with `-Perrorprone` will fail without them since the duplicate branches are marked as errors.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/bean/MetaBean.java
jbertram: I'm fine with deleting this, but I left it in to be conservative.

jbertram: @clebertsuconic, are you OK with deleting this?

gemmellr:
I'm fine with the other changes, they make sense overall, but are only being done due to the upgrade so it seems reaosonable enough to bundle them. I just think the commented-out-code shouldnt be there, better nothing.

I did also type and delete 'maybe we could disable/downgrade-to-warning that new check?' before suggesting just to delete the comment since I'm fine with the changes. I'm also fine with duplicate branches where its clearer though, so I'd be fine without the check also.

clebertsuconic:
@jbertram sure.. delete it.

jbertram:
@clebertsuconic, done!

PR #5401, opened 2024-12-13
ARTEMIS-5215 AMQP federation consumer close can allow duplicates on r…

tabish121:
…emote

Messages that are in-flight on a link when the federation consumer is closed due to removal of local demand might get enqueued on the local broker but the disposition might not get sent to the remote leading to a release of the message back to the source queue leading to a duplicate being left on the remote. The AMQP receiver links need to be drained and any pending settlements need to be sent before the link can be allowed to be detached.


clebertsuconic:
this is so major work.. nice!!!


if tests are good... LGTM

tabish121:
I've run the tests on a CI box as well as running them repeatedly locally with varying loads on the host to check that no new test failures have been introduced.

PR #5404, opened 2024-12-17
ARTEMIS-5218 StompHandler HeartBeat restarted after being stopped

clebertsuconic:


clebertsuconic:
This is fixing StompV11Test::testHeartBeat4

clebertsuconic:
@jbertram can you check this please?

clebertsuconic:
I manage to reproduce it once... then after adding a few loggers it would fail 1 in 100... 

i have a branch called itFails, that I just log.info into a few places and that's enough to make it fail more often.


The fix here will prevent the HeartBeat from being restarted if it's been shutdown. In a certain case a failure is sent to the connection from stomp and then a frame comes and restarts it.. so if it's been stopped it should be kept stopped.

PR #5408, opened 2024-12-17
ARTEMIS-5182 STOMP subscriber w/noLocal + selector missing messages

jbertram:


artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompConnection.java
```diff
@@ -566,7 +566,7 @@ StompPostReceiptFunction subscribe(String destination,
          if (selector == null) {
             selector = noLocalFilter;
          } else {
-            selector += " AND " + noLocalFilter;
+            selector = "(" + selector + ") AND " + noLocalFilter;
```
brusdev, line 569: could selector be an empty string?

jbertram: Yes, I believe so. I'll add some validation.

jbertram: Validation added.

brusdev (approved):
Full test-suite and example passed

PR #5410, opened 2024-12-18
ARTEMIS-5183 STOMP noLocal filter ignores legitimate messages

jbertram:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/StompTest.java
```diff
@@ -1501,6 +1501,32 @@ public void testSubscribeToTopicWithNoLocal() throws Exception {
       conn.disconnect();
    }
 
+   @Test
+   public void testSubscribeToTopicWithNoLocalSendWithStomp() throws Exception {
+      conn.connect(defUser, defPass);
+      subscribeTopic(conn, null, null, null, true, true);
+
+      // send a message on the same connection => it should not be received is noLocal = true on subscribe
```
gemmellr, line 1509:
```suggestion
      // send a message on the same connection => it should not be received as noLocal = true on subscribe
```
(and place c&p from)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/StompTest.java
```diff
@@ -1501,6 +1501,32 @@ public void testSubscribeToTopicWithNoLocal() throws Exception {
...
+      send(conn, getTopicPrefix() + getTopicName(), null, "Hello World");
+
+      ClientStompFrame frame = conn.receiveFrame(100);
+      assertNull(frame, "No message should have been received since subscription was removed");
```
gemmellr, line 1513:
```suggestion
      assertNull(frame, "No message should have been received since subscription specified noLocal");
```
(and place c&p from)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/StompTest.java
```diff
@@ -1501,6 +1501,32 @@ public void testSubscribeToTopicWithNoLocal() throws Exception {
...
+
+      // send message on another STOMP connection => it should be received
+      StompClientConnection conn2 = StompClientConnectionFactory.createClientConnection(uri);
+      conn2.connect(defUser, defPass);
+      // a STOMP message does *not* automatically get a property named __AMQ_CID
```
gemmellr, line 1518:
```suggestion
      // a STOMP message does *not* automatically get a property named __AMQ_CID if there arent any noLocal subscriptions on the connection
```

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/StompTest.java
```diff
@@ -1501,6 +1501,32 @@ public void testSubscribeToTopicWithNoLocal() throws Exception {
...
+      send(conn2, getTopicPrefix() + getTopicName(), null, getName());
+      frame = conn.receiveFrame(10000);
+      assertNotNull(frame);
+      assertEquals(Stomp.Responses.MESSAGE, frame.getCommand());
+      assertEquals(getTopicPrefix() + getTopicName(), frame.getHeader(Stomp.Headers.Send.DESTINATION));
+      assertEquals(getName(), frame.getBody());
+
+      conn.disconnect();
+      conn2.disconnect();
```
gemmellr, line 1527: Though there is fallback handling for _conn_, there isnt for _conn2_, so the conn2.disconnect() might not happen if anything fails, might be worth adding a try-finally, or try-with-resources if simple, or maybe use runAfter (could e.g extract a utility method from the existing _conn_ fallback cleanup and pass it _conn2_)

PR #5414, opened 2024-12-20
ARTEMIS-5184 STOMP noLocal is scoped to session not subscription

jbertram:


artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
```diff
@@ -231,6 +232,9 @@ public void closed() {
    public void disconnect(ServerConsumer consumerId, String errorDescription) {
       StompSubscription stompSubscription = subscriptions.remove(consumerId.getID());
       if (stompSubscription != null) {
+         if (stompSubscription.isNoLocal()) {
+            noLocalSubscriptionCount.decrementAndGet();
+         }
```
gemmellr, line 237: This disconnect method looks to be the only place a decrement is done...what if the subscription is explicitly unsubscribed and so this doesnt occur?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/StompTest.java
```diff
@@ -1501,6 +1501,26 @@ public void testSubscribeToTopicWithNoLocal() throws Exception {
...
+         if (binding.getFilter() != null && binding.getFilter().getFilterString().toString().contains("__AMQ_CID")) {
+            noLocalCount++;
+         } else {
+            normalCount++;
+         }
+      }
+      assertEquals(1, noLocalCount);
+      assertEquals(1, normalCount);
+
+      conn.disconnect();
```
gemmellr, line 1521: Probably worth verifying removals behave are as expected also, and perhaps sending a message after and checking the expected behaviours, e.g around the header.

artemis-protocols/artemis-stomp-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/stomp/StompSession.java
jbertram: Fixed.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/stomp/StompTest.java
jbertram: I think I got all the relevant cases.

jbertram:
@gemmellr, thanks!

PR #5426, opened 2025-01-08
ARTEMIS-5233 make style for commas consistent

jbertram:
This commit enforces a consistent style for commas, namely that a space should follow a comma. Futhermore, it update all the code that violates this styling up to date so that the code is styled consistently across the entire code-base.


jbertram:
I was shocked this rule wasn't already in place.

clebertsuconic:
whoa.. I'm also shocked...

I need to merge this right away.. otherwise we will get out of sync pretty fast as I'm sending a PR myself

PR #5431, opened 2025-01-10
ARTEMIS-5232 Speed up RedirectTest

brusdev:


tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/routing/RedirectTest.java
```diff
@@ -45,37 +45,30 @@
...
-      final String[] protocols = new String[] {AMQP_PROTOCOL, CORE_PROTOCOL, OPENWIRE_PROTOCOL};
-      final String[] pools = new String[] {CLUSTER_POOL, DISCOVERY_POOL, STATIC_POOL};
       Collection<Object[]> data = new ArrayList<>();
 
-      for (String protocol : Arrays.asList(protocols)) {
-         for (String pool : Arrays.asList(pools)) {
-            data.add(new Object[] {protocol, pool});
-         }
+      for (String pool : Arrays.asList(pools)) {
+         data.add(new Object[] {AMQP_PROTOCOL, pool});
```
gemmellr, line 61: The protocol parameter isnt being used, the field it populated was removed, so seems like it can just go away rather than be passed a fixed unused value of AMQP_PROTOCOL, which will make for misleading test display names in results.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/routing/RedirectTest.java
```diff
@@ -105,39 +98,41 @@ public void testSimpleRedirect() throws Exception {
       QueueControl queueControl1 = (QueueControl)getServer(1).getManagementService()
          .getResource(ResourceNames.QUEUE + queueName);
 
-      assertEquals(0, queueControl0.countMessages());
-      assertEquals(0, queueControl1.countMessages());
+      for (String protocol : Arrays.asList(protocols)) {
```
gemmellr, line 101: As set out here, many failure scenarios will be difficult to distinguish from each other, e.g any of the assertions failing will give the exact same stack+message for all of the protocols.

Originally I was suggesting just extracting a method, and explicitly calling it in turn for each protocol of interest...in that case at least the stacks would differ for each failure.

In this style of repetition, I'd suggest adding messages to the assertions which denote the protocol, thena failure will be more immediately understandable. Also, the body of TextMessage sent could also indicate the protocol (and the test name?) rather than just "TEST", helping ensure the truly correct matching message is actually the one that was received, and not another of the other identical messages sent during the test.

( + Much the same for the other tests below)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/routing/RedirectTest.java
clebertsuconic, line 61: no need for AMQP_PROTOCOL here, it's not being used.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/routing/RedirectTest.java
```diff
@@ -45,37 +45,30 @@
...
       return data;
    }
 
 
-   private final String protocol;
-
    private final String pool;
 
 
    public RedirectTest(String protocol, String pool) {
```
clebertsuconic, line 71: and you need to remove the String protococol parameter from here.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/routing/RedirectTest.java
```diff
@@ -45,37 +45,30 @@
...
-import org.apache.activemq.artemis.tests.util.Wait;
 import org.junit.jupiter.api.TestTemplate;
 import org.junit.jupiter.api.extension.ExtendWith;
 
 @ExtendWith(ParameterizedTestExtension.class)
 public class RedirectTest extends RoutingTestBase {
+   static final String[] protocols = new String[] {AMQP_PROTOCOL, CORE_PROTOCOL, OPENWIRE_PROTOCOL};
+   static final String[] pools = new String[] {CLUSTER_POOL, DISCOVERY_POOL, STATIC_POOL};
 
    @Parameters(name = "protocol: {0}, pool: {1}")
```
clebertsuconic, line 56: you need to remove the protocol from the name.

clebertsuconic (changes requested):

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/routing/RedirectTest.java
clebertsuconic: hmmm... I made the same note.. didn't realize you had already done it.

gemmellr: The comment timestamp is from when I originally made the comment, but I only actually completed and posted my review moments before you did yours, i.e we posted the reviews at essentially the same time (its only shown to the minute, so both have same timestamp value).

brusdev: I'll remove the protocol parameter.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/routing/RedirectTest.java
brusdev: Now I see your point, extracting a method is better for analyzing a failure.

PR #5448, opened 2025-01-15
ARTEMIS-5099 fix impossible cast in AMQPLargeMessage

jbertram:
I believe this was likely a copy-and-paste error involving o.a.a.a.c.p.i.j.LargeServerMessageImpl#referenceOriginalMessage.


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/AMQPLargeMessage.java
```diff
@@ -660,7 +659,7 @@ public void referenceOriginalMessage(final Message original, final SimpleString
 
       super.referenceOriginalMessage(original, originalQueue);
 
-      if (original instanceof LargeServerMessageImpl) {
+      if (original instanceof AMQPLargeMessage) {
```
tabish121, line 662: My main question would be what is the impact of adding this call into the mix given that it wasn't happening before and I guess didn't cause anything to break?  Would something break now?

jbertram: When you say, "it wasn't happening before," what exactly do you mean? Do you mean that `org.apache.activemq.artemis.core.persistence.impl.journal.LargeBody#referenceOriginalMessage` wasn't being invoked since this `instanceof` would never succeed?

tabish121: I'm not sure what else I could have meant but let me be more clear.  What will happen now that an AMQPLargeMessage cast would succeed when this method is called?  Should something have been testing this?  From a brief look at the broker code calling this it seems like the source message would never be anything other than the AMQPLargeMessage itself so I doubt the cast to ServerLargeMessageImpl would have succeeded in the past.

jbertram: These two lines break the existing code:
```java
AMQPLargeMessage m = new AMQPLargeMessage(0, 0, null, null, null);
m.referenceOriginalMessage(new LargeServerMessageImpl((byte) 0, 0, null, null), SimpleString.of("foo"));
```
Causing this to be thrown:
```
java.lang.ClassCastException: class org.apache.activemq.artemis.core.persistence.impl.journal.LargeServerMessageImpl cannot be cast to class org.apache.activemq.artemis.protocol.amqp.broker.AMQPLargeMessage (org.apache.activemq.artemis.core.persistence.impl.journal.LargeServerMessageImpl and org.apache.activemq.artemis.protocol.amqp.broker.AMQPLargeMessage are in unnamed module of loader 'app')
```
It's not clear to me that this code is directly related to any conversion. If it was I would expect `org.apache.activemq.artemis.core.persistence.impl.journal.LargeServerMessageImpl#referenceOriginalMessage` to behave in a similar way.

clebertsuconic:
I removed the code, and pushed the commit directly

PR #5449, opened 2025-01-15
ARTEMIS-5094 potential null dereference in QueueImpl

jbertram:
The `server` variable is used liberally through this `Queue` implementation. It makes sense to require it to be non-null.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
@@ -19,6 +19,7 @@ package org.apache.activemq.artemis.core.server.impl;
 import javax.security.auth.Subject;
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.lang.invoke.MethodHandles;
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -658,6 +658,8 @@ public QueueImpl(final QueueConfiguration queueConfiguration,
                     final QueueFactory factory) {
       super(server == null ? EmptyCriticalAnalyzer.getInstance() : server.getCriticalAnalyzer(), CRITICAL_PATHS);
 
+      Objects.requireNonNull(server);
```
gemmellr, line 661: This makes the line above it a bit nonsensical and changing that might make this line  redundant, depending on the approach taken.

jbertram: Agreed. Since Java < 22 forces `super` to be the first call it makes this kind of clumsy. Maybe I should just do:
```
super(server.getCriticalAnalyzer(), CRITICAL_PATHS);
```
And if `server` is `null` then it will throw an NPE instead of `requireNonNull`.

gemmellr: Thats probably the option I'd pick this time.

PR #5450, opened 2025-01-15
ARTEMIS-5115 remove needless null checks in ChannelImpl

jbertram:
The `connection` variable is used liberally throughout `ChannelImpl` and it is never set to `null`. Therefore, it should never be `null` and we can dispense with unnecessary checks for `null`.

### DIFF
--- a/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
+++ b/artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
@@ -18,6 +18,7 @@ package org.apache.activemq.artemis.core.protocol.core.impl;
 
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentLinkedQueue;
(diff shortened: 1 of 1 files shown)


artemis-core-client/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/ChannelImpl.java
```diff
@@ -698,7 +698,7 @@ public void transferConnection(final CoreRemotingConnection newConnection) {
 
          if (logger.isTraceEnabled()) {
             logger.trace("RemotingConnectionID={} transferConnection to new RemotingConnectionID={}",
-                         (connection == null ? "NULL" : connection.getID()),
+                         connection.getID(),
```
gemmellr, line 701: It would make sense to remove the check on the related line immediately following this too, especially as _newConnection_ replaces _connection_, but would be NPEd on in the next bit of code after this logging call before it can happen. Perhaps could also add a newConnection null check at the start of the method to make it more explicit?

jbertram: Done. I think. Take a look at the recent push.

PR #5451, opened 2025-01-15
ARTEMIS-5107 using wrong value in ReplicationStartSyncMessage ctor

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/protocol/core/impl/wireformat/ReplicationStartSyncMessage.java
```diff
@@ -76,7 +76,7 @@ public static SyncDataType getDataType(byte code) {
 
    public ReplicationStartSyncMessage(boolean beforeTwoEighteen) {
       super(REPLICATION_START_FINISH_SYNC);
-      this.beforeTwoEighteen = synchronizationIsFinished;
+      this.beforeTwoEighteen = beforeTwoEighteen;
```
gemmellr, line 79: If _this.beforeTwoEighteen_ was always being set false before, due to the _synchronizationIsFinished_ field being false at this point, what are the implications of changing this now such that it might now become true?

This seems to point to a test either not existing or not actually being effective in checking this, or perhaps just not actually needing this/related code if it never actually did anything. It doesnt seem like this change should be happening without some other matching change.

jbertram: The main fix for the original issue (i.e. ARTEMIS-3767) was in `ReplicationAddMessage` and `ReplicationAddTXMessage`. The only important part of the fix in `ReplicationStartSyncMessage` is in `decodeRest` where versions before 2.18.0 will send less data than newer versions. The `beforeTwoEighteen` variable is not actually needed because sending _more_ data to an older version is not a problem. The older version will simply ignore it. I'm updating the fix to take this into account.

gemmellr: The original original PR and email discussion had suggested there was an issue with the newer servers sending the additional data to the old broker, but it would certainly seem like thats what it has always continued doing even after the subsequent PR was applied, suggesting that isnt the case (or that in the end noone hit it again).

In general I do think it makes more sense to prevent sending such new data to old brokers when making such changes, but it would appear that has always been the case here, so if it hasnt caused issue then at this stage years later perhaps just removing the gating check is the way to go at this point.

gemmellr: What are your thoughts on this @clebertsuconic ?

jbertram: As I understand it, the problem wasn't so much that the newer server was sending _additional_ data so much as it was sending _different_ data that the older server was choking on.

PR #5452, opened 2025-01-15
ARTEMIS-4926 throw checked URISyntaxException on bad URL

jbertram:


artemis-commons/src/main/java/org/apache/activemq/artemis/utils/uri/URISupport.java
```diff
@@ -140,12 +141,24 @@ public static boolean containsQuery(SimpleString uri) {
...
          if (p >= 0) {
-            String name = URLDecoder.decode(parameter.substring(0, p), StandardCharsets.UTF_8);
-            String value = URLDecoder.decode(parameter.substring(p + 1), StandardCharsets.UTF_8);
+            String name;
+            String value;
+            try {
+               name = URLDecoder.decode(parameter.substring(0, p), StandardCharsets.UTF_8);
+            } catch (IllegalArgumentException e) {
+               ActiveMQUtilLogger.LOGGER.unableToParseURLParameterName(parameter.substring(0, p), e);
+               continue;
```
gemmellr, line 154: Catching and continuing does not actually seem like an improvement to me at all. Personally, I would want it to throw rather than just continue and perhaps do something quite different than I had perhaps intended.

The original complaint seems to be more that this sub-method can throw IAE and the calling methods dont handle that explicitly, or declare so. Feels like they could as easily just declare it throws IAE, or even perhaps wrap with URISyntaxException.

If taking that approach, this could just call _parseParameters(URI uri)_ per the original report and check behviour, rather than exposing this private method.

jbertram: That's fair. See my latest push.

tabish121: This looks good to me.  I had similar thoughts on the original changes as Robbie did, I would generally expect things to fail than to suppress errors and lead people to think configuration was applied when it was not.

PR #5453, opened 2025-01-16
ARTEMIS-5248 add all roles in an ordered list to the audit log user s…

gtully:
…tring


brusdev (approved):
This seems more a bug fix, well done!

PR #5455, opened 2025-01-17
ARTEMIS-5100 support journal-max-io on 'create' CLI command

jbertram:
I struggled to decide the name of the input, ultimately settling on `--aio-journal-max-io` to clarify that it only applies to `journal-max-io` when using aio.

### DIFF
--- a/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
+++ b/artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
@@ -298,6 +298,9 @@ public class Create extends InstallAbstract {
    @Option(names = "--security-manager", description = "Which security manager to use - jaas or basic. Default: jaas.")
    private String securityManager = "jaas";
 
+   @Option(names = "--journal-max-io", description = "The journal-max-io value to use when also using the ASYNCIO journal-type. When using NIO or MAPPED this value is always '1'. Default: 4096")
+   private int journalMaxIo = ActiveMQDefaultConfiguration.getDefaultJournalMaxIoAio();
+
    @Option(names = "--jdbc-bindings-table-name", description = "Name of the jdbc bindings table.")
    private String jdbcBindings = ActiveMQDefaultConfiguration.getDefaultBindingsTableName();
 
@@ -500,6 +503,30 @@ public class Create extends InstallAbstract {
 
    public void setRole(String role) {
       this.role = role;
+   }
+
+   public int getJournalMaxIo() {
+      return journalMaxIo;
+   }
+
+   public void setJournalMaxIo(int journalMaxIo) {
+      this.journalMaxIo = journalMaxIo;
+   }
+
+   public boolean isAio() {
+      return aio;
+   }
+
+   public void setAio(boolean aio) {
+      this.aio = aio;
+   }
+
+   public boolean isNio() {
+      return nio;
+   }
+
+   public void setNio(boolean nio) {
+      this.nio = nio;
    }
 
    private boolean isBackup() {
@@ -862,7 +889,7 @@ public class Create extends InstallAbstract {
       context.out.println(String.format("   \"%s\" run", path(new File(directory, "bin/artemis"))));
 
       File service = new File(directory, BIN_ARTEMIS_SERVICE);
-      context.out.println("");
+      context.out.println();
 
       if (IS_NIX) {
          context.out.println("Or you can run the broker in the background using:");
@@ -1049,7 +1076,7 @@ public class Create extends InstallAbstract {
                Map<String, String> syncFilter = new HashMap<>();
                syncFilter.put("${nanoseconds}", "0");
                syncFilter.put("${writesPerMillisecond}", "0");
-               syncFilter.put("${maxaio}", journalType == JournalType.ASYNCIO ? "" + ActiveMQDefaultConfiguration.getDefaultJournalMaxIoAio() : "1");
+               syncFilter.put("${maxaio}", "1");
 
                getActionContext().out.println("...Since you disabled sync and are using MAPPED journal, we are diabling buffer times");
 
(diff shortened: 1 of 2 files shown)


artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -327,6 +327,9 @@ public String[] getStaticNodes() {
    @Option(names = "--jdbc-lock-expiration", description = "Lock expiration (in milliseconds).")
    long jdbcLockExpiration = ActiveMQDefaultConfiguration.getDefaultJdbcLockExpirationMillis();
 
+   @Option(names = "--aio-journal-max-io", description = "The journal-max-io value to use when also using the ASYNCIO journal-type. When using NIO or MAPPED this value is always '1'. Default: 4096")
```
clebertsuconic, line 330: what about --journal-max-aio


(journal max asynchronous io).

jbertram: That's not bad!

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -327,6 +327,9 @@ public String[] getStaticNodes() {
    @Option(names = "--jdbc-lock-expiration", description = "Lock expiration (in milliseconds).")
    long jdbcLockExpiration = ActiveMQDefaultConfiguration.getDefaultJdbcLockExpirationMillis();
 
+   @Option(names = "--journal-max-aio", description = "The journal-max-io value to use when also using the ASYNCIO journal-type. When using NIO or MAPPED this value is always '1'. Default: 4096")
```
gemmellr, line 330: Regardless what it is called, I would move this up beside the other journal related options rather than adding it after all the jdbc options (if there isnt some constraint preventing it).

The configuration impl class uses a suffix for its fields based on the journal type, JournalMaxIO_NIO and JournalMaxIO_AIO, so I'd wonder if such suffix naming convention transfers over to broker-properties usage somehow and if so might --journal-max-io-aio fit with it for alignment?

jbertram: Using the same pattern as broker properties would yield something like `--journal-max-io_aio` or maybe `--journal-max-io-aio`. I think the latter would be OK. It's one of the options I originally considered.

gemmellr: Yeah, I think `journal-max-io-aio` would be fine. I think I'd at least try to make those two use cases align if possible, if it it is not just going to be using --journal-max-io to match the xml (which I also think would be fine, I dont think its any more confusing than the XML element always being named and being present even if its not doing anything; the CLI description makes clear enough when the option applies, and it wont have any effect in non-applicable cases regardless what it is named).

clebertsuconic: I prefer journal-max-aio


io-aio sounds redundant to me.

jbertram: > io-aio sounds redundant to me.

It is a little redundant, but it's no more redundant than `Configuration.setJournalMaxIO_AIO`, `Configuration.setJournalMaxIO_NIO`, etc. which is also reflected in broker properties.

gemmellr: It is a little redundant, but also more consistent and less likely to confuse.

For me --journal-max-aio is a bit _off_ because it is so similar to, but still not the same as, the actual journal-max-io XML element setting it is actually for controlling. I expect that would trip some folks up, both in finding it, and in using it. I would just use --journal-max-io in preference to --journal-max-aio personally. The different ways of configuring journal-max-io would then at least all include journal-max-io in their naming in some form, either partially or as the whole name.

If we use journal-max-aio we will then have 3 different names for configuration for the same thing, one of which doesnt even contain the name journal-max-io despite directly existing to govern that XML element.

clebertsuconic: I don't think there's a direct reference between the Create command and their XML elements... these are distinct namespaces to me.


--journal-max-io-aio won't make it much better... if we are concerned we would have to find a totally different name such as --journal-limit-aio...


i think --journal-max-aio for this case is fine.. since there are no other similar parameters on the Create


that's my opinion on this, but I'm not strong opinionated ... if you guys want to go against what I think... I wouldn't fight over :)

jbertram: I settled on `--journal-max-io` just like we already have in the XML. Although it's not perfectly descriptive at least it's consistent, and the help text should provide enough information to clarify how to use it.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
```diff
@@ -298,6 +298,9 @@ public String[] getStaticNodes() {
    @Option(names = "--security-manager", description = "Which security manager to use - jaas or basic. Default: jaas.")
    private String securityManager = "jaas";
 
+   @Option(names = "--journal-max-io", description = "The journal-max-io value to use when also using the ASYNCIO journal-type. When using NIO or MAPPED this value is always '1'. Default: 4096")
+   private long journalMaxIo = ActiveMQDefaultConfiguration.getDefaultJournalMaxIoAio();
```
gemmellr, line 302: I just twigged with your most recent updates that this is a long...which on reflection seemed a bit large given the default, and this option being added because that is actually _too big_ in some envs.

On checking, journal-max-io is also defined as an _int_ for all the rest of the configuration bits, so this should match.

jbertram: Fixed.

PR #5457, opened 2025-01-17
ARTEMIS-2490 mitigate NumberFormatException when parsing large message

jbertram:
No specific test here as static analysis is sufficient in my opinion. This is really just an abundance of caution since there's been a report of this in the community.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
@@ -354,12 +354,18 @@ public class JournalStorageManager extends AbstractJournalStorageManager {
       largeMessage.setStorageManager(this);
 
       if (largeMessage.toMessage().containsProperty(Message.HDR_ORIG_MESSAGE_ID)) {
-         // for compatibility: couple with old behaviour, copying the old file to avoid message loss
-         long originalMessageID = largeMessage.toMessage().getLongProperty(Message.HDR_ORIG_MESSAGE_ID);
-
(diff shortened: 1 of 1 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/persistence/impl/journal/JournalStorageManager.java
```diff
@@ -354,12 +354,18 @@ protected LargeServerMessage parseLargeMessage(final ActiveMQBuffer buff) throws
...
       if (largeMessage.toMessage().containsProperty(Message.HDR_ORIG_MESSAGE_ID)) {
-         // for compatibility: couple with old behaviour, copying the old file to avoid message loss
-         long originalMessageID = largeMessage.toMessage().getLongProperty(Message.HDR_ORIG_MESSAGE_ID);
-
          SequentialFile currentFile = createFileForLargeMessage(largeMessage.toMessage().getMessageID(), true);
 
          if (!currentFile.exists()) {
+            // for compatibility: couple with old behaviour, copying the old file to avoid message loss
+            String originalMessageIDString = largeMessage.toMessage().getStringProperty(Message.HDR_ORIG_MESSAGE_ID);
+            // could have a leading "ID:", like ID:c6e2e367-d77e-11e9-9471-005056b7cdce
```
tabish121, line 362: The comment seems to acknowledge that what follows the prefix for many of the clients is not going to be a long but likely a UUID or something UUID like for say OpenWire which in that case is going to throw a NumberFormatException again only the exception is from a String without the "ID:" prefix so I'm unclear how this really improves things?

jbertram: You're right, @tabish121. That's a fatal flaw.

PR #5459, opened 2025-01-18
[ARTEMIS-5209] add test and docs for AMQP duplicate Message

guissalustiano:
I suggest add duplicate detection id header following the x-opt-delivery-time header,
but after digging around some code I notice this was already implemented, it was just
missing the docs to clarify how to use it.

### DIFF
--- a/docs/user-manual/amqp.adoc
+++ b/docs/user-manual/amqp.adoc
@@ -107,6 +107,12 @@ x-opt-delivery-delay::
 
 If both annotations are present in the same message then the broker will prefer the more specific `x-opt-delivery-time` value.
 
+== AMQP duplicate Message Detection
+
+An AMQP message can provide automatic duplicate message detection.
+The duplication id is provided by adding a application property given by the value of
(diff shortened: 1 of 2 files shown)


gemmellr:
As you noticed, the broker already supports duplicate detection with AMQP messages, in the same general way it does for e.g Core and other protocol usages. This was added in some time ago, in ARTEMIS-2294. There are existing tests from that time, though I did just make and improvement to them.

There isnt a specific note on the AMQP page for duplicate detection as it really notes behaviours specific to the AMQP support, whilst for duplicate detection the same general docs on duplicate detection apply to all the protocols and are a better fit. I did reorder and tweak the doc to emphasize the more general details first, before a note of the detail for just the Core Client.

Both the tweaks I made are in https://github.com/apache/activemq-artemis/commit/89b3ebb077b203dd0c9ae3b85983d6b2ac1231e9 and I updated the Jira to reflect that, so I will close this PR out, but thanks for the contribution!

PR #5460, opened 2025-01-18
ARTEMIS-5163 Broker fails to send MQTT LWT using certificate-based mutual TLS

JeanLucGraphalo:
Implements a fix of bug [ARTEMIS-5163](https://issues.apache.org/jira/browse/ARTEMIS-5163).

The changes follow the proposals as described in the ticket comments.
A lot of code refactoring was necessary to fix the problem, which could make the review challenging.

### DIFF
--- a/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
+++ b/artemis-protocols/artemis-mqtt-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/mqtt/MQTTConnectionManager.java
@@ -16,6 +16,8 @@
  */
 package org.apache.activemq.artemis.core.protocol.mqtt;
 
+import javax.security.auth.Subject;
+import java.lang.invoke.MethodHandles;
 import java.util.List;
 
 import io.netty.buffer.ByteBufAllocator;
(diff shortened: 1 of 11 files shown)


jbertram:
First off, thanks for sending a PR! There's some good work here, but I have a few concerns.

1. There are a number of changes that aren't strictly required to fix the issue. For example, just in the first commit there's a block of code that is refactored into its own method as well as changes to a JavaDoc comment. Other commits change the names of variables, finalize variables, change the format of boolean expressions, etc. These unnecessary changes make it difficult to see where the actual semantic changes are.
2. At the very least, the first commit doesn't build properly because it contains a Checkstyle problem.
3. Code is added in one commit and then removed in another commit. For example, 
4. Conceptually speaking this change violates the auth caching model used by the broker. Instead of having all the `Subject` values centralized in a single place (e.g. the cache) they also exist in the MQTT session state object. This increases the size of the heap makes the code more complex. It also subverts the expectation that auth entries will expire and be renewed.
5. Typically a PR is _single_ commit which includes all necessary code changes, tests, and doc updates. This simplifies the process of tracking changes for a specific bug or feature. You have 11 commits here which are all very directly related and could probably be squashed down to just 1. If the commits contains multiple notable changes then simply enumerate them in the commit message (e.g. like [this](https://github.com/apache/activemq-artemis/commit/10adca5479c7d0aa518c0ffce44008e95a268287)). Once you remove all the unnecessary changes that will simplify the commit substantially.

jbertram:
I think a really simple solution here would be to cache the `X509Certificate[]` on the `RemotingConnection` so that when the underlying Netty channel is unavailable (e.g. closed) auth will still work.

jbertram:
@JeanLucGraphalo, do you plan on addressing my concerns or should I just close this?

JeanLucGraphalo:
@jbertram, your comment has been added to the jira. I close the PR.

PR #5471, opened 2025-01-26
ARTEMIS-5263 Remove unused Docker directive

ViliusS:
JMX Exporter is long gone and pipes are not used in the scripts anymore. Plus, some other minor adjustments.

P.S. Next bash could probably also be removed from Alpine images to keep them small as the whole point of Alpine images are small size. I didn't do tests at the moment, but quickly looking at docker-run.sh and prepare-docker.sh didn't find anything requiring bash.


gemmellr:
I'm not sure about the other changes, but on the port removal you will need to rebase your PR now, as I just merged an earlier PR #5468 that did that in isolation and hadnt seen your PR or that overlapped yet.

ViliusS:
Done.

jbertram:
The commit message isn't accurate since you removed the bits related to the JMX Exporter. Once you update that I'll merge.

ViliusS:
Should be OK now.

jbertram:
@ViliusS, thanks!

PR #5477, opened 2025-01-28
ARTEMIS-5266 code gardening after move to Java 17

jbertram:


artemis-commons/src/test/java/org/apache/activemq/artemis/utils/collections/LongHashSetTest.java
```diff
@@ -327,7 +326,7 @@ public void containsEmptySet() {
       final LongHashSet other = new LongHashSet(100);
 
       assertTrue(testSet.containsAll(other));
-      assertTrue(testSet.containsAll((Collection<?>) other));
+      assertTrue(testSet.containsAll(other));
```
gemmellr, line 329: These aren't redundant casts. The class has its own containsAll(LongHashSet) but really has 2 methods also inherits containsAll(Collection) (unclear if it shouldnt be overriding it).

Notice all the changed lines result in duplicate assertions; this cast is the test trying to force which method is used in order to check both methods.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/largemessages/InterruptedAMQPLargeMessage.java
```diff
@@ -147,7 +147,7 @@ public void testInterruptedLargeMessage() throws Exception {
       AmqpReceiver receiver = session.createReceiver(getQueueName());
 
       int received = 0;
-      receiver.flow((int) (messageCount + 10));
+      receiver.flow((messageCount + 10));
```
gemmellr, line 150: Can also lose the second set of parentheses that were needed for the cast that was removed.

tests/activemq5-unit-tests/src/test/java/org/apache/activemq/openwire/v1/ActiveMQBytesMessageTest.java
```diff
@@ -40,7 +40,6 @@ public Object createObject() throws Exception {
    @Override
    protected void populateObject(Object object) throws Exception {
       super.populateObject(object);
-      ActiveMQBytesMessage info = (ActiveMQBytesMessage) object;
```
gemmellr, line 43: This removal then also renders the entire method superflous since all it does is call super.

I'd either leave it as it is, since these are originally generated files from ported tests (so the more we change them, the further from the originals they go)...or delete the methods instead of just the variable.

Same goes for the likely tens of other similar cases

tests/db-tests/src/test/java/org/apache/activemq/artemis/tests/db/paging/PagingTest.java
```diff
@@ -424,8 +425,6 @@ public void testSimpleCursorIteratorLargeMessage() throws Exception {
       server = createServer(true, config, PAGE_SIZE, PAGE_MAX);
       server.start();
 
-      final int numberOfBytes = 200 * 1024;
```
gemmellr, line 427: The method says it is for testing LargeMessage, yet it isnt using the numberOfBytes variable, and doesnt obliviously appear to be using large messages.

Rather than needing removal, this seems more like its indicating a test bug that needs fixed?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/failover/BackupSyncJournalTest.java
```diff
@@ -154,14 +154,9 @@ public void testReserveFileIdValuesOnBackup() throws Exception {
 
       assertEquals(size, backupMsgJournal.getFileSize(), "file sizes must be the same");
       Set<Pair<Long, Integer>> backupIds = getFileIds(backupMsgJournal);
-
-      int total = 0;
       for (Pair<Long, Integer> pair : primaryIds) {
-         total += pair.getB();
```
gemmellr, line 160: Not sure we need empty loop this leaves, both here and the similar one that follows immediately.

I would instead wonder what they originally existed to do and whether the test simply isnt/wasnt checking something that it was actually meant to.

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/journal/impl/JournaHistorylBackupTest.java
```diff
@@ -125,11 +126,11 @@ public void removeBKPExtension() throws Throwable {
       String withBKP = withoutBkp + ".bkp";
       // I was actually born at 4:30 :) but I need all numbers lower than 2 digits on the test
       assertEquals(withoutBkp, journal.removeBackupExtension(withBKP));
-      assertEquals(withoutBkp, journal.removeBackupExtension(withoutBkp)); // it should be possible to do it
-
-      String withoutBKP = "jrn-1.data";
+      assertEquals(withoutBkp, journal.removeBackupExtension(withoutBkp));
    }
 
+   // it should be possible to do it
```
gemmellr, line 132: Doesnt seem like this should have moved here

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
gemmellr, line 1: Most if not all of the changes in this file appear to be in the wrong commit(s), they are in e95f2ac4124d2bd1f8abecf65e919faac7a67ed6 for ARTEMIS 5272 "eliminate unused private fields" but none of the changes in the file do that, they are a mix of using interface type and text block strings.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
```diff
@@ -94,16 +94,11 @@ private enum STATE {
...
 
-   private static final long serialVersionUID = -1615857864410205260L;
-
    // This is the default value
    private ClientProtocolManagerFactory protocolManagerFactory = new ActiveMQClientProtocolManagerFactory().setLocator(this);
 
    private final boolean ha;
 
-   // this is not used... I'm only keeping it here because of Serialization compatibility and Wildfly usage on JNDI.
-   private boolean finalizeCheck;
```
gemmellr, line 105: This one is strange since, it was only made this way in recent years, but the class wasnt Serializable at the time. The history of the file is littered with references to serialization...but the file was seemingly never Serializable in the time since Artemis existed. There is a nested class in the file that _is_ Serializable though.

Is there perhaps another type of serializing the locator that the earlier change here thought needed to be concious of not breaking?

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQConnectionFactory.java
```diff
@@ -96,9 +94,6 @@ public class ActiveMQConnectionFactory extends JNDIStorable implements Connectio
 
    private boolean cacheDestinations;
 
-   // keeping this field for serialization compatibility only. do not use it
-   private boolean finalizeChecks;
```
gemmellr, line 100: Similar pondering to https://github.com/apache/activemq-artemis/pull/5477#discussion_r1933742419, with difference that I can see that the relation to JNDIBaseStorable here probably means this really is actually needed for compatibility and should stay.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jndi/ActiveMQInitialContextFactory.java
gemmellr, line 1: I would probably be inclined not to change the environment type in this file, or ReadOnlyContext, they both use Hashtable because Context / InitialContext always use Hashtable for the environment, including in the getters.

artemis-core-client/src/test/java/org/apache/activemq/artemis/api/core/management/AddressSettingsInfoTest.java
```diff
@@ -30,65 +30,67 @@ public class AddressSettingsInfoTest {
...
+         "deadLetterQueueSuffix":".FOO",
+         "autoCreateExpiryResources":true,
+         "expiryQueuePrefix":"BAR.",
+         "expiryQueueSuffix":".BAR",
+         "expiryDelay":404,
+         "minExpiryDelay":40,
+         "maxExpiryDelay":4004,
+         "enableMetrics":false
+         }
+         """;
```
gemmellr, line 93: Not actually an issue, especially for the usage here, but just noting for awareness...the text-block will subtly differ from the original string concats in having a newline at the end, since the closing delimiter is on its own line rather than the same line [and the block also doesnt use the trailing \\<line-terminator\> escape to suppress a line ending being added when the block is processed].

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/logger/ActiveMQAMQPProtocolLogger.java
```diff
@@ -32,19 +32,25 @@ public interface ActiveMQAMQPProtocolLogger {
    @LogMessage(id = 111000, value = "Scheduled task can't be removed from scheduledPool.", level = LogMessage.Level.WARN)
    void cantRemovingScheduledTask();
 
-   @LogMessage(id = 111001, value = "\n*******************************************************************************************************************************" +
-      "\nCould not re-establish AMQP Server Connection {} on {} after {} retries" +
-      "\n*******************************************************************************************************************************\n", level = LogMessage.Level.WARN)
+   @LogMessage(id = 111001, value = """
+      *******************************************************************************************************************************
```
gemmellr, line 36: Did you mean to remove the deliberate newline at the start?
(same for others)

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/brokerConnection/MirrorInfiniteRetryReplicaTest.java
```diff
@@ -153,18 +153,18 @@ private static void createMirroredServer(String serverName,
...
+
+         logger.endpoint.name=org.apache.activemq.artemis.core.replication.ReplicationEndpoint
+         logger.endpoint.level=DEBUG
+         logger.ackmanager.name=org.apache.activemq.artemis.protocol.amqp.connect.mirror.AckManager
+         logger.ackmanager.level=INFO
+         logger.mirrorTarget.name=org.apache.activemq.artemis.protocol.amqp.connect.mirror.AMQPMirrorControllerTarget
+         logger.mirrorTarget.level=INFO
+         appender.console.filter.threshold.type = ThresholdFilter
+         appender.console.filter.threshold.level = trace\
+         """));
```
gemmellr, line 167: Possibly simpler to add the close deliminator on the same line than use line ending suppression?

tests/soak-tests/src/test/java/org/apache/activemq/artemis/tests/soak/brokerConnection/mirror/SingleMirrorSoakTest.java
```diff
@@ -164,23 +163,30 @@ private static void createServer(String serverName,
...
+            logger.config.level=TRACE
+            logger.counter.name=org.apache.activemq.artemis.core.paging.cursor.impl.PageSubscriptionCounterImpl
+            logger.counter.level=DEBUG
+            logger.queue.name=org.apache.activemq.artemis.core.server.impl.QueueImpl
+            logger.queue.level=DEBUG
+            logger.rebuild.name=org.apache.activemq.artemis.core.paging.cursor.impl.PageCounterRebuildManager
+            logger.rebuild.level=DEBUG
+            appender.console.filter.threshold.type = ThresholdFilter
+            appender.console.filter.threshold.level = info\
+            """));
```
gemmellr, line 189: Similarly

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/TypedProperties.java
```diff
@@ -174,10 +174,10 @@ public Boolean getBooleanProperty(final SimpleString key) throws ActiveMQPropert
...
       if (value == null) {
          return Boolean.valueOf(null);
-      } else if (value instanceof Boolean) {
-         return (Boolean) value;
-      } else if (value instanceof SimpleString) {
-         return Boolean.valueOf(((SimpleString) value).toString());
+      } else if (value instanceof Boolean boolean1) {
+         return boolean1;
+      } else if (value instanceof SimpleString string) {
+         return Boolean.valueOf(string.toString());
```
gemmellr, line 180: Things like boolean1 (plus byte1, short1, float1,and double1 etc that follow below) are more than a bit ugly, probably even less readable compared to the casts.

Calling the SimpleString _string_ results in the silly reading `string.toString()`.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/runnables/AtomicRunnable.java
```diff
@@ -23,8 +23,8 @@
 public abstract class AtomicRunnable implements Runnable {
 
    public static AtomicRunnable checkAtomic(Runnable run) {
-      if (run instanceof AtomicRunnable) {
-         return (AtomicRunnable)run;
+      if (run instanceof AtomicRunnable runnable) {
+         return runnable;
```
gemmellr, line 27:
```suggestion
      if (run instanceof AtomicRunnable atomicRunnable) {
         return atomicRunnable;
```
It is already a Runnable, so calling it merely _runnable_ right after we determined it is more than just that and return it as a different type, simply reads horribly, far more than the original cast.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/JsonUtil.java
```diff
@@ -111,14 +108,13 @@ public static Object[] fromJsonArray(final JsonArray jsonArray) throws Exception
             }
 
             array[i] = map;
-         } else if (val instanceof JsonString) {
-            array[i] = ((JsonString) val).getString();
+         } else if (val instanceof JsonString string) {
+            array[i] = string.getString();
```
gemmellr, line 112:
```suggestion
         } else if (val instanceof JsonString jsonString) {
            array[i] = jsonString.getString();
```

gemmellr: I'm giving up looking at the rest of the json stuff...same type of recent comments+suggestions to make it more readable than originally, rather than arguably less readable, also apply to the rest of it as well.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/Message.java
```diff
@@ -507,8 +507,8 @@ default byte[] getDuplicateIDBytes() {
       if (duplicateID == null) {
          return null;
       } else {
-         if (duplicateID instanceof SimpleString) {
-            return ((SimpleString) duplicateID).getData();
+         if (duplicateID instanceof SimpleString string) {
+            return string.getData();
```
gemmellr, line 511:
```suggestion
         if (duplicateID instanceof SimpleString simpleString) {
            return simpleString.getData();
```

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/TransportConfiguration.java
```diff
@@ -328,18 +328,18 @@ private void encodeMap(final ActiveMQBuffer buffer, final Map<String, Object> ma
 
          Object val = entry.getValue();
 
-         if (val instanceof Boolean) {
+         if (val instanceof Boolean boolean1) {
```
gemmellr, line 331: booleanVal / longVal / etc style naming or similar , or just leaving the casts, would be more nicer to read later than than boolean1 etc

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
```diff
@@ -961,9 +960,9 @@ private int getConnectionWithRetry(final int reconnectAttempts, RemotingConnecti
          }
 
          if (getConnection() != null) {
-            if (oldConnection != null && oldConnection instanceof CoreRemotingConnection) {
+            if (oldConnection != null && oldConnection instanceof CoreRemotingConnection remotingConnection) {
```
gemmellr, line 963: **old**RemotingConnection would be more readable, given its used on the same line as the new CoreRemotingConnection _connection_

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQMessageProducer.java
```diff
@@ -549,27 +549,27 @@ public void sendAcknowledged(org.apache.activemq.artemis.api.core.Message client
...
          }
 
          try {
             producer.connection.getThreadAwareContext().setCurrentThread(true);
-            if (exception instanceof ActiveMQException) {
-               exception = JMSExceptionHelper.convertFromActiveMQException((ActiveMQException) exception);
-            } else if (exception instanceof ActiveMQInterruptedException) {
-               exception = JMSExceptionHelper.convertFromActiveMQException((ActiveMQInterruptedException) exception);
+            if (exception instanceof ActiveMQException qException) {
+               exception = JMSExceptionHelper.convertFromActiveMQException(qException);
```
gemmellr, line 570:
```suggestion
            if (exception instanceof ActiveMQException amqException) {
               exception = JMSExceptionHelper.convertFromActiveMQException(amqException);
```

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
```diff
@@ -634,8 +634,8 @@ public MessageConsumer createSharedConsumer(Topic topic, String name, String mes
       }
       checkTopic(topic);
       ActiveMQTopic localTopic;
-      if (topic instanceof ActiveMQTopic) {
-         localTopic = (ActiveMQTopic) topic;
+      if (topic instanceof ActiveMQTopic qTopic) {
```
gemmellr, line 637: qException was strange, but qTopic is about the most nonsense thing it could have picked :)

Here and all the others below

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/broker/ProtonProtocolManager.java
```diff
@@ -374,8 +371,8 @@ public String invokeIncoming(Message message, ActiveMQProtonRemotingConnection c
       // For tunneled messages we need to check the type as our interceptor only cares about
       // AMQP message right now so there's not notification point for other types that cross
       if (incomingInterceptors != null && !incomingInterceptors.isEmpty()) {
-         if (message instanceof AMQPMessage) {
-            return super.invokeInterceptors(this.incomingInterceptors, (AMQPMessage) message, connection);
+         if (message instanceof AMQPMessage pMessage) {
+            return super.invokeInterceptors(this.incomingInterceptors, pMessage, connection);
```
gemmellr, line 375:
```suggestion
         if (message instanceof AMQPMessage amqpMessage) {
            return super.invokeInterceptors(this.incomingInterceptors, amqpMessage, connection);
```

same below

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/mirror/AMQPMirrorControllerSource.java
```diff
@@ -510,8 +510,8 @@ private static void setProtocolData(MessageReference ref, String brokerID, long
    }
 
    private static Properties getProperties(Message message) {
-      if (message instanceof AMQPMessage) {
-         return AMQPMessageBrokerAccessor.getCurrentProperties((AMQPMessage)message);
+      if (message instanceof AMQPMessage pMessage) {
+         return AMQPMessageBrokerAccessor.getCurrentProperties(pMessage);
```
gemmellr, line 514:
```suggestion
      if (message instanceof AMQPMessage amqpMessage) {
         return AMQPMessageBrokerAccessor.getCurrentProperties(amqpMessage);
```

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/converter/CoreAmqpConverter.java
```diff
@@ -72,8 +72,8 @@
 public class CoreAmqpConverter {
 
    public static AMQPMessage checkAMQP(Message message, StorageManager storageManager) throws Exception {
-      if (message instanceof AMQPMessage) {
-         return (AMQPMessage)message;
+      if (message instanceof AMQPMessage pMessage) {
+         return pMessage;
```
gemmellr, line 76:
```suggestion
      if (message instanceof AMQPMessage amqpMessage) {
         return amqpMessage;
```
Plus another below

artemis-protocols/artemis-openwire-protocol/src/main/java/org/apache/activemq/artemis/core/protocol/openwire/OpenWireMessageConverter.java
```diff
@@ -216,8 +216,8 @@ public static org.apache.activemq.artemis.api.core.Message inbound(final Message
       }
 
       final Object scheduledDelay = messageSend.getProperties().get(ScheduledMessage.AMQ_SCHEDULED_DELAY);
-      if (scheduledDelay instanceof Long) {
-         coreMessage.putLongProperty(org.apache.activemq.artemis.api.core.Message.HDR_SCHEDULED_DELIVERY_TIME, System.currentTimeMillis() + ((Long) scheduledDelay));
+      if (scheduledDelay instanceof Long long1) {
```
gemmellr, line 219: Something like delayMs or delayLong would be far more readable later.

Just about every name added below is equally unhelpful, and results in less readable code than the original.

artemis-selector/src/main/java/org/apache/activemq/artemis/selector/filter/ComparisonExpression.java
```diff
@@ -229,8 +229,8 @@ public Object evaluate(Filterable message) throws FilterException {
             if (lv == rv || lv.equals(rv)) {
                return Boolean.TRUE;
             }
-            if (lv instanceof Comparable && rv instanceof Comparable) {
-               return compare((Comparable) lv, (Comparable) rv);
+            if (lv instanceof Comparable comparable && rv instanceof Comparable comparable1) {
+               return compare(comparable, comparable1);
```
gemmellr, line 233: Names need updater or its far readable as it originally was.

Same below

artemis-server/src/main/java/org/apache/activemq/artemis/core/paging/impl/PagedMessageImpl.java
```diff
@@ -176,8 +176,8 @@ public void initMessage(StorageManager storage) {
          largeMessageLazyData = null;
          checkLargeMessage();
       } else {
-         if (message != null && message instanceof LargeServerMessage) {
-            ((LargeServerMessage)message).setStorageManager(storageManager);
+         if (message != null && message instanceof LargeServerMessage serverMessage) {
+            serverMessage.setStorageManager(storageManager);
```
gemmellr, line 180:
```suggestion
         if (message != null && message instanceof LargeServerMessage largeServerMessage) {
            largeServerMessage.setStorageManager(storageManager);
```

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueImpl.java
```diff
@@ -3717,8 +3717,8 @@ private RoutingStatus move(final Transaction originalTX,
       Message copyMessage = makeCopy(ref, reason == AckReason.EXPIRED, address);
 
       Object originalRoutingType = ref.getMessage().getBrokerProperty(Message.HDR_ORIG_ROUTING_TYPE);
-      if (originalRoutingType != null && originalRoutingType instanceof Byte) {
-         copyMessage.setRoutingType(RoutingType.getType((Byte) originalRoutingType));
+      if (originalRoutingType != null && originalRoutingType instanceof Byte byte1) {
+         copyMessage.setRoutingType(RoutingType.getType(byte1));
```
gemmellr, line 3721:
```suggestion
      if (originalRoutingType != null && originalRoutingType instanceof Byte origRoutingType) {
         copyMessage.setRoutingType(RoutingType.getType(origRoutingType));
```
Similarly below

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/management/ArtemisRbacInvocationHandler.java
```diff
@@ -136,15 +136,15 @@ public boolean canInvoke(String name, String operationName) {
 
    private void initializeFromFirstServerMBeanRegistration(Method method, Object[] args) {
       if (activeMQServer == null && method.getName().equals("registerMBean")) {
-         if (args != null && args[0] instanceof ActiveMQServerControlImpl) {
-            activeMQServer = ((ActiveMQServerControlImpl) args[0]).getServer();
+         if (args != null && args[0] instanceof ActiveMQServerControlImpl impl) {
```
gemmellr, line 139: impl -> control ?

artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/protocol/EmbedMessageUtil.java
```diff
@@ -98,8 +98,8 @@ private static Message extractRegularMessage(ICoreMessage message, StorageManage
    private static Message readEncoded(ICoreMessage message, StorageManager storageManager, ActiveMQBuffer buffer) {
       try {
          Message returnMessage = MessagePersister.getInstance().decode(buffer, null, null, storageManager);
-         if (returnMessage instanceof LargeServerMessage) {
-            ((LargeServerMessage)returnMessage).setStorageManager(storageManager);
+         if (returnMessage instanceof LargeServerMessage serverMessage) {
+            serverMessage.setStorageManager(storageManager);
```
gemmellr, line 102:
```suggestion
         if (returnMessage instanceof LargeServerMessage largeServerMessage) {
            largeServerMessage.setStorageManager(storageManager);
```

artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/protocol/MessagePersister.java
```diff
@@ -119,8 +119,8 @@ public Message decode(ActiveMQBuffer buffer, Message record, CoreMessageObjectPo
          throw new NullPointerException("couldn't find factory for type=" + protocol);
       }
       Message message = persister.decode(buffer, record, pools);
-      if (message instanceof LargeServerMessage) {
-         ((LargeServerMessage) message).setStorageManager(storageManager);
+      if (message instanceof LargeServerMessage serverMessage) {
+         serverMessage.setStorageManager(storageManager);
```
gemmellr, line 123: Similarly

artemis-server/src/main/java/org/apache/activemq/artemis/spi/core/security/jaas/JaasCallbackHandler.java
```diff
@@ -77,8 +76,8 @@ public void handle(Callback[] callbacks) throws IOException, UnsupportedCallback
             if (peerPrincipalFromConnection != null) {
                principalsCallback.setPeerPrincipals(new Principal[] {peerPrincipalFromConnection});
             }
-         } else if (callback instanceof ClientIDCallback) {
-            ((ClientIDCallback) callback).setClientID(remotingConnection.getClientID());
+         } else if (callback instanceof ClientIDCallback dCallback) {
+            dCallback.setClientID(remotingConnection.getClientID());
```
gemmellr, line 80:
```suggestion
         } else if (callback instanceof ClientIDCallback clientIDCallback) {
            clientIDCallback.setClientID(remotingConnection.getClientID());
```

gemmellr:
finally done looking through the commits. Left 'a couple' of comments hehe. The last commit however, I didnt comment on lots of things that were just more of the same (I noted where in some cases).

For example there are lots of instanceof pattern match name additions like byte1/long1/\<other\>1, or \<last-letter-of-type-prefix\>Exception, that make the code far less readable than it was to begin with with the original variable name and a cast. I think the entire commit needs be swept for those and updated before its fit to go in.

artemis-commons/src/test/java/org/apache/activemq/artemis/utils/collections/LongHashSetTest.java
jbertram: Fixed.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/amqp/largemessages/InterruptedAMQPLargeMessage.java
jbertram: Fixed.

tests/activemq5-unit-tests/src/test/java/org/apache/activemq/openwire/v1/ActiveMQBytesMessageTest.java
jbertram: I reverted most of the changes in that module.

tests/db-tests/src/test/java/org/apache/activemq/artemis/tests/db/paging/PagingTest.java
jbertram: I sent a new, separate commit to deal with this.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/failover/BackupSyncJournalTest.java
jbertram: Fixed.

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/journal/impl/JournaHistorylBackupTest.java
jbertram: Fixed.

tests/unit-tests/src/test/java/org/apache/activemq/artemis/tests/unit/core/message/impl/MessageImplTest.java
```diff
@@ -345,14 +346,11 @@ public void run() {
...
          }
       }
-
-      final String bigString;
       {
          StringBuffer buffer = new StringBuffer();
          for (int i = 0; i < 500; i++) {
             buffer.append(" ");
          }
-         bigString = buffer.toString();
```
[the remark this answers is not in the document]
jbertram: Fixed.

artemis-cli/src/main/java/org/apache/activemq/artemis/cli/commands/Create.java
jbertram: This should be fixed now.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ServerLocatorImpl.java
jbertram: I just left this one alone.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jndi/ActiveMQInitialContextFactory.java
jbertram: Fixed.

artemis-log-annotation-processor/tests/src/test/java/org/apache/activemq/artemis/logs/annotation/processor/LogAnnotationProcessorTest.java
```diff
@@ -43,59 +43,59 @@ public class LogAnnotationProcessorTest {
...
    @Resource("org/apache/activemq/artemis/logs/annotation/processor/cases/LAPTCase1_InvalidIDForRegex.java")
    public void testIDInvalidForGivenRegex(Results results) {
-      final String expectedMessage = "org.apache.activemq.artemis.logs.annotation.processor.cases.LAPTCase1_InvalidIDForRegex: "
-                                    + "Code 100 does not match regular expression specified on the LogBundle: [0-9]{1}";
-
-      doCheckFailureErrorMessageTestImpl(results, expectedMessage);
+      doCheckFailureErrorMessageTestImpl(results, """
+         org.apache.activemq.artemis.logs.annotation.processor.cases.LAPTCase1_InvalidIDForRegex: \
+         Code 100 does not match regular expression specified on the LogBundle: [0-9]{1}\
+         """);
```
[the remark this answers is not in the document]
jbertram: That's fair.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/logger/ActiveMQAMQPProtocolLogger.java
jbertram: Fixed.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationParserTest.java
```diff
@@ -56,6 +56,66 @@
...
+               <max-size-bytes>10485760</max-size-bytes>
+               <message-counter-history-day-limit>10</message-counter-history-day-limit>
+               <address-full-policy>BLOCK</address-full-policy>
+            </address-setting>
+         </address-settings>
+      """;
+
+   private static String lastPart = "</core>";
+
+   private static String bridgePart = """
```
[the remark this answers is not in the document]
jbertram: Fixed.

tests/smoke-tests/src/test/java/org/apache/activemq/artemis/tests/smoke/brokerConnection/MirrorInfiniteRetryReplicaTest.java
jbertram: Fixed.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/collections/TypedProperties.java
jbertram: Fixed.

artemis-commons/src/main/java/org/apache/activemq/artemis/utils/runnables/AtomicRunnable.java
jbertram: Fixed.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/JsonUtil.java
jbertram: I think I got all of them.

artemis-core-client/src/main/java/org/apache/activemq/artemis/api/core/TransportConfiguration.java
jbertram: All should be fixed now.

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/client/impl/ClientSessionFactoryImpl.java
jbertram: Fixed.

artemis-jms-client/src/main/java/org/apache/activemq/artemis/jms/client/ActiveMQSession.java
jbertram: `qTopic`...lol

jbertram:
I think I addressed all of your comments :crossed_fingers: .

tests/db-tests/src/test/java/org/apache/activemq/artemis/tests/db/paging/PagingTest.java
```diff
@@ -449,6 +450,12 @@ public void testSimpleCursorIteratorLargeMessage() throws Exception {
 
          message.getBodyBuffer().writerIndex(0);
 
+         message.getBodyBuffer().writeBytes(new byte[numberOfBytes]);
+
+         for (int j = 1; j <= numberOfBytes; j++) {
+            message.getBodyBuffer().writeInt(j);
+         }
```
gemmellr, line 457: It definitely makes it use the unused numberOfBytes variable, but why write _both_ an array of bytes, and all the ints?

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
```diff
@@ -2715,13 +2715,13 @@ public void testListConnectionsAsJSON() throws Exception {
       assertNotNull(first);
       assertNotNull(second);
 
-      assertTrue(first.getString("connectionID").length() > 0);
-      assertTrue(first.getString("clientAddress").length() > 0);
+      assertTrue(!first.getString("connectionID").isEmpty());
+      assertTrue(!first.getString("clientAddress").isEmpty());
```
gemmellr, line 2719: I think it should actually use assertFalse(foo.isEmpty()) rather than introduce all these assertTrue( !foo.isEmpty()), or else perhaps just leave it the way it was.

Applies here and lots below in this class, along with various other test classes (a couple were updated to flip the assert rather than adding negations)

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/failover/BackupSyncJournalTest.java
gemmellr: That further change, to remove the loops, then just leaves both of the collections being looped over as unused themselves.

The test is called testReserveFileIdValuesOnBackup, and the collections are seemingly about the Ids. If its not using them then it doesnt seem like its actually testing what it meant to. If these bits are actually removed, its not clear the test shouldnt just be deleted instead.

gemmellr:
I looked over the latest changes and they all look good except for 2 small new things and 1 old thing I commented on. I resolved all the others conversations.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/failover/BackupSyncJournalTest.java
gemmellr: @clebertsuconic  are you clear on what the test was maybe originally trying to do, and if it can/should be fixed?

jbertram: The assertion related to these loops was removed in af9bd7b84aad32e4fe30f2c8909e51cf7300b475 by @clebertsuconic. The test was originally added way back in HornetQ by Francisco Borges (who implemented replication initially).

jbertram: The assertion related to these loops was removed in af9bd7b84aad32e4fe30f2c8909e51cf7300b475 by @clebertsuconic. The test was originally added way back in HornetQ by Francisco Borges (who implemented replication initially).

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/management/ActiveMQServerControlTest.java
jbertram: Should all be fixed now.

tests/integration-tests/src/test/java/org/apache/activemq/artemis/tests/integration/cluster/failover/BackupSyncJournalTest.java
gemmellr: Yeah, I don't actually know what it is validating.

gemmellr: > That further change, to remove the loops, then just leaves both of the collections being looped over as unused themselves.
> 

Also, if we remove creation of the collections and leave the rest of the test, the getFileIds and getPair methods used for creating them also become unused as well.

jbertram: I decided to keep the test and just clean it up by removing the unnecessary variables, loops, methods, etc.

PR #5479, opened 2025-01-29
ARTEMIS-5292 introduce jakarta version of artemis-cdi-client

cortlepp:
This PR introduces `artemis-jakarta-cdi-client`, a Jakarta (i.e. Jakarta EE 9+) equivalent of `artemis-cdi-client`. The new module is based on `artemis-cdi-client` with all necessary changes to make it jakarta-compatible. No other changes were made. You can review the individual steps in the commit history.

I will also post about this on the dev mailing list  and would appreciate any feedback.

### DIFF
--- a/artemis-bom/pom.xml
+++ b/artemis-bom/pom.xml
@@ -109,6 +109,11 @@
          </dependency>
          <dependency>
             <groupId>org.apache.activemq</groupId>
+            <artifactId>artemis-jakarta-cdi-client</artifactId>
+            <version>${project.version}</version>
+         </dependency>
+         <dependency>
+            <groupId>org.apache.activemq</groupId>
             <artifactId>artemis-jakarta-client</artifactId>
             <version>${project.version}</version>
          </dependency>
--- a/artemis-cdi-client/src/main/java/org/apache/artemis/client/cdi/extension/ArtemisClientConfigBean.java
+++ b/artemis-cdi-client/src/main/java/org/apache/artemis/client/cdi/extension/ArtemisClientConfigBean.java
@@ -45,7 +45,8 @@ class ArtemisClientConfigBean implements Bean<ArtemisClientConfiguration> {
       return emptySet();
    }
 
-   @Override
+   // we cannot use @Override here because otherwise the build of artemis-jakarta-cdi-client fails since the method was removed in jakarta ee 9+
+   @SuppressWarnings("MissingOverride")
    public boolean isNullable() {
       return false;
    }
--- a/artemis-jakarta-cdi-client/pom.xml
+++ b/artemis-jakarta-cdi-client/pom.xml
@@ -27,8 +27,22 @@
    </parent>
    <modelVersion>4.0.0</modelVersion>
 
-   <artifactId>artemis-cdi-client</artifactId>
-   <name>ActiveMQ Artemis CDI Integration</name>
+   <artifactId>artemis-jakarta-cdi-client</artifactId>
+   <name>ActiveMQ Artemis Jakarta CDI Integration</name>
+
+   <properties>
+      <jakarta.inject-api.version>${jakarta.inject-api.version.alt}</jakarta.inject-api.version>
+      <jakarta.annotation-api.version>${jakarta.annotation-api.version.alt}</jakarta.annotation-api.version>
+      <jakarta.enterprise.cdi-api.version>${jakarta.enterprise.cdi-api.version.alt}</jakarta.enterprise.cdi-api.version>
+      <jakarta.jms-api.version>${jakarta.jms-api.version.alt}</jakarta.jms-api.version>
+      <jakarta.transaction-api.version>${jakarta.transaction-api.version.alt}</jakarta.transaction-api.version>
+      <!-- the following is needed because of a transitive dependency of arquillian -->
+      <jakarta.ejb-api.version>${jakarta.ejb-api.version.alt}</jakarta.ejb-api.version>
+
+      <weld.version>${weld.version.alt}</weld.version>
+      <arquillian-weld-embedded.version>${arquillian-weld-embedded.version.alt}</arquillian-weld-embedded.version>
+      <owb.version>${owb.version.alt}</owb.version>
+   </properties>
 
    <dependencies>
       <!-- logging -->
@@ -53,11 +67,11 @@
       </dependency>
       <dependency>
          <groupId>org.apache.activemq</groupId>
-         <artifactId>artemis-jms-client</artifactId>
+         <artifactId>artemis-jakarta-client</artifactId>
       </dependency>
       <dependency>
          <groupId>org.apache.activemq</groupId>
-         <artifactId>artemis-server</artifactId>
+         <artifactId>artemis-jakarta-server</artifactId>
       </dependency>
       <dependency>
          <groupId>org.apache.activemq</groupId>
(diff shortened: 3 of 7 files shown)


gemmellr:
This looks like you manually copied the code to the new module and then changed it. None of the other -jakarta modules, such as artemis-jakarta-client which this relies on, were implemented that way with their own committed sources. Instead, the new module brings in the related original javax module code itself during the build and transforms it. That way, one set of code exists to update (or mostly not, in the case of artemis-cdi-client) and there is no need to synchronize changes or scope for them to diverge.

If creating this module then it should ideally be implemented the same way, especially as the original module is so rarely updated, mentioned, or seemingly used that discussing deleting even it came up last month just before you popped up asking about it (seemingly the first to do so).

cortlepp:
Thanks for the feedback! I wasn't aware of how this transformation was done in the other modules.

As far as I can tell using the transformer plugin worked, with two exceptions.
- the `javax.enterprise.inject.spi.Extension` file was just copied to the resources folder and not renamed to `jakarta.enterprise.inject.spi.Extension`. I added this file manually, thus both files are now in the resources of artemis-cdi-jakarta-client, which should not be problematic (I think).
- the `isNullable()` method was removed from `Bean` in Jakarta ee9+, so we cannot override after transforming the source. I removed the `@Override` (from artemis-cdi-client), while it is unfortunate that we can't do this in a prettier way I also don't think this is a serious problem.

I have never worked with this transformer plugin before though, so it would be great if someone could take another look, maybe there is a better solution.

gemmellr:
I have never worked with the plugin either but had a dig and found that the plugin actually should process the ServiceLoader services file, and looking at it with debug logging it actually _does_ do so and even indicates that the name _should_ be updated, but then it just copies it with the original name.

Looking closer at how the plugin works, its really just some glue and it is the Eclipse Transformer doing the main work underneath. I had a look there and found the ServiceLoader processing, and looking at the changes to that file it looks like it specifically had a fix that seems like it is around this issue (https://github.com/eclipse-transformer/transformer/commit/c2207cefa841cb4c6f37949fb79663d5bad3750d) of not updating the file because it _only_ needs renamed and has no change to the content.

Unfortunately, the plugin is a bit old and is using an older version of the transformer. Trying to have the build use newer versions with the fix above, it seems they aren't compatible.

Given it is just one file, I don't think its worth switching out to a different newer plugin just to get this working seamlessly. Your workaround of just adding the new services file directly seems fine for this specific case.

However I also don't think we should leave the incorrect services file around. Running an execution of the maven-clean-plugin during the generate-sources phase, just after the transform is done, to delete the copy is probably the simplest thing for now. E.g:

```
diff --git a/artemis-cdi-jakarta-client/pom.xml b/artemis-cdi-jakarta-client/pom.xml
index 218659cb69..f425880eec 100644
--- a/artemis-cdi-jakarta-client/pom.xml
+++ b/artemis-cdi-jakarta-client/pom.xml
@@ -157,6 +157,27 @@
                </dependency>
             </dependencies>
          </plugin>
+         <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-clean-plugin</artifactId>
+            <executions>
+               <execution>
+                  <id>remove-old-services-file</id>
+                  <phase>generate-sources</phase>
+                  <goals>
+                     <goal>clean</goal>
+                  </goals>
+                  <configuration>
+                    <filesets>
+                      <fileset>
+                        <directory>${project.build.directory}/generated-resources/transformed/META-INF/services/</directory>
+                      </fileset>
+                    </filesets>
+                    <excludeDefaultDirectories>true</excludeDefaultDirectories>
+               </execution>
+            </executions>
+         </plugin>
       </plugins>
    </build>
 
```

gemmellr:
The new module should probably be named `artemis-jakarta-cdi-client`, given the convention of all the other similar such modules.

The new module should be added to the bom.

This change needs a Jira raised: https://issues.apache.org/jira/browse/ARTEMIS which should then be referenced in the PR title (see most other PRs except dependabot's).

The commits should be squashed, and the Jira referenced in the commit message (see existing [commit history](https://github.com/apache/activemq-artemis/commits/main/))

cortlepp:
Ok, I think I have done all of the things you asked, except for the squashing since I was unsure whether GitHub allows to do that when merging or if I have to do it manually via force-pushing this branch.

artemis-bom/pom.xml
```diff
@@ -233,6 +233,11 @@
             <artifactId>artemis-website</artifactId>
             <version>${project.version}</version>
          </dependency>
+         <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>artemis-jakarta-cdi-client</artifactId>
+            <version>${project.version}</version>
+         </dependency>
```
gemmellr, line 240: Move this up with the other jakarta modules, sorted alphabetically.

artemis-cdi-client/src/main/java/org/apache/artemis/client/cdi/extension/ArtemisClientConfigBean.java
```diff
@@ -45,7 +45,8 @@ public Set<InjectionPoint> getInjectionPoints() {
       return emptySet();
    }
 
-   @Override
+   // we cannot use @Override here because otherwise the build of artemis-cdi-jakarta-client fails since the method was removed in jakarta ee 9+
```
gemmellr, line 48: Still need to update the module name following its rename.

(same in other class changed)

artemis-jakarta-cdi-client/pom.xml
```diff
@@ -0,0 +1,227 @@
...
+   <parent>
+      <groupId>org.apache.activemq</groupId>
+      <artifactId>artemis-pom</artifactId>
+      <version>2.40.0-SNAPSHOT</version>
+      <relativePath>../artemis-pom/pom.xml</relativePath>
+   </parent>
+   <modelVersion>4.0.0</modelVersion>
+
+   <artifactId>artemis-jakarta-cdi-client</artifactId>
+   <name>ActiveMQ Artemis CDI Integration</name>
```
gemmellr, line 31: Differentiate the name, e.g. ActiveMQ Artemis Jakarta CDI Integration, so its clear when building which module Maven was referring to in the summary.

artemis-jakarta-cdi-client/pom.xml
```diff
@@ -0,0 +1,227 @@
...
+                  <groupId>org.wildfly.extras.batavia</groupId>
+                  <artifactId>transformer-impl-eclipse</artifactId>
+                  <version>${version.batavia}</version>
+               </dependency>
+            </dependencies>
+         </plugin>
+         <plugin>
+               <groupId>org.apache.maven.plugins</groupId>
+               <artifactId>maven-clean-plugin</artifactId>
+               <executions>
```
gemmellr, line 163: The indentation is off here, should be 3-spaced like the rest.

artemis-pom/pom.xml
```diff
@@ -876,6 +876,11 @@
             <version>${apache.httpclient.version}</version>
          </dependency>
 
+         <dependency>
+            <groupId>org.jboss.shrinkwrap</groupId>
+            <artifactId>shrinkwrap-api</artifactId>
+            <version>${shrinkwrap-api.version}</version>
+         </dependency>
```
gemmellr, line 883: Should move this up to the related bunch of deps around weld/arquillian.

gemmellr:
> Ok, I think I have done all of the things you asked, except for the squashing since I was unsure whether GitHub allows to do that when merging or if I have to do it manually via force-pushing this branch.

We dont have the squash option enabled, so someone would have to do it manually. I was suggesting you do it and force push, updating the commit message. You can also rebase while at it.

artemis-jakarta-cdi-client/pom.xml
```diff
@@ -0,0 +1,227 @@
...
+               <artifactId>weld-se-core</artifactId>
+            </dependency>
+            <dependency>
+               <groupId>org.jboss.arquillian.container</groupId>
+               <artifactId>arquillian-weld-embedded</artifactId>
+            </dependency>
+            <dependency>
+               <groupId>jakarta.ejb</groupId>
+               <artifactId>jakarta.ejb-api</artifactId>
+            </dependency>
```
gemmellr, line 202: This ejb-api dependency listing wasn't part of the original `artemis-cdi-client` modules explicit dependencies. Though I see it used to be brought in transitively by the older cdi-api dep, but isnt anymore by the newer cdi-api version. 

However that cdi-api dep was always a _provided_ scope dep for `artemis-cdi-client` , meaning using applications never got it or the ejb-api via the module, whereas this ejb-api listing actually results in a _compile_ scope dependency for it meaning usages of `artemis-jakarta-cdi-client` now would see the dep.

Rather than changing it to _provided_, I tried just deleting it since I see the arquillian-weld bits already bring it in themselves transitively at test scope, and as expected that meant the tests still worked fine.

So, can this just be removed?

gemmellr: If it can please do so, and rebase your branch before pushing.

cortlepp: It seem to like this should work, i removed ejb. While I was at it I also removed a duplicate annotation-api dependency (it was present both in the normal dependencies and the OWB profile).

I also rebased this branch (sorry I thought I'd already done that last time, but I forgot that my fork is the upstream hence it did not have the latest commits).

jbertram:
This looks good to me.

gemmellr:
Latest changes look good here too, merged.

gemmellr:
I made a fixup in https://github.com/apache/activemq-artemis/commit/fd6c6e3ddb restoring/adding the test deps removed/omitted in https://github.com/apache/activemq-artemis/commit/a6cb1b283baf5750313e8e64924c1f6368067e44 as it turns out they really are required in cases like the accompanying comment said.

PR #5483, opened 2025-01-31
ARTEMIS-5280 metrics manager doesn't inspect temp queue namespace

jbertram:


artemis-server/src/test/java/org/apache/activemq/artemis/core/server/metrics/MetricsManagerTest.java
```diff
@@ -0,0 +1,82 @@
...
+         // this is a simplified version of org.apache.activemq.artemis.core.server.impl.ActiveMQServerImpl.getRuntimeTempQueueNamespace
+         if (temp) {
+            return tempQueueNamespace;
+         } else {
+            return "";
+         }
+      });
+
+      // test temp queue
+      CountDownLatch tempTest = new CountDownLatch(1);
```
brusdev, line 69: The builder is executed before the return of registerQueueGauge, could tempTest be replace with an AtomicBoolean or any other simpler class to track the execution without waiting for the timeout?

jbertram: Done.

PR #5484, opened 2025-02-01
ARTEMIS-5282 remove deprecated queue-related methods

jbertram:
All but one QueueImpl constructor was deprecated years ago in 2efa44daf52994790737e21ee29ae830f8f0a12c. Since QueueImpl is an internal implementation object (as the name suggests) these constructors can be completely removed and didn't technically need to be deprecated in the first place.

Furthermore, the methods from QueueFactoryImpl which use these constructors were also deprecated in that same commit. Despite the fact that there is a QueueFactory interface these objects are still for internal use/implementation. Therefore, these deprecated methods can and should be removed.


artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueFactoryImpl.java
```diff
@@ -79,9 +79,9 @@ public void setPostOffice(final PostOffice postOffice) {
...
+         queue = new LastValueQueue(QueueConfiguration.of(config.name()).setAddress(config.address()).setId(config.id()).setUser(config.user()).setDurable(config.isDurable()).setTemporary(config.isTemporary()).setAutoCreated(config.isAutoCreated()).setRoutingType(config.deliveryMode()).setMaxConsumers(config.maxConsumers()).setExclusive(config.isExclusive()).setGroupRebalance(config.isGroupRebalance()).setGroupBuckets(config.getGroupBuckets()).setGroupFirstKey(config.getGroupFirstKey()).setConsumersBeforeDispatch(config.consumersBeforeDispatch()).setDelayBeforeDispatch(config.delayBeforeDispatch()).setPurgeOnNoConsumers(config.isPurgeOnNoConsumers()).setLastValueKey(lastValueKey(config)).setNonDestructive(config.isNonDestructive()).setAutoDelete(config.isAutoDelete()).setAutoDeleteDelay(config.getAutoDeleteDelay()).setAutoDeleteMessageCount(config.getAutoDeleteMessageCount()).setConfigurationManaged(config.isConfigurationManaged()), config.filter(), config.getPagingStore(), config.pageSubscription(), scheduledExecutor, postOffice, storageManager, addressSettingsRepository, executorFactory.getExecutor(), server, this);
```
gemmellr, line 82: Since almost all of the options appear to be common, I think it would be both more readable and more maintainable if a single set of options was prepared first with all the common config, and then only the difference applied to them before creating the queue objects and passing them.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueFactoryImpl.java
```diff
@@ -120,9 +120,9 @@ public Queue createQueue(final long persistenceID,
...
...iveMQDefaultConfiguration.getDefaultMaxQueueConsumers()).setExclusive(ActiveMQDefaultConfiguration.getDefaultExclusive()).setGroupRebalance(ActiveMQDefaultConfiguration.getDefaultGroupRebalance()).setGroupBuckets(ActiveMQDefaultConfiguration.getDefaultGroupBuckets()).setGroupFirstKey(ActiveMQDefaultConfiguration.getDefaultGroupFirstKey()).setConsumersBeforeDispatch(ActiveMQDefaultConfiguration.getDefaultConsumersBeforeDispatch()).setDelayBeforeDispatch(ActiveMQDefaultConfiguration.getDefaultDelayBeforeDispatch()).setPurgeOnNoConsumers(ActiveMQDefaultConfiguration.getDefaultPurgeOnNoConsumers()).setLastValueKey(lastValueKey(addressSettings)).setNonDestructive(ActiveMQDefaultConfiguration.getDefaultNonDestructive()).setAutoDelete(ActiveMQDefaultConfiguration.getDefaultQueueAutoDelete(autoCreated)).setAutoDeleteDelay(ActiveMQDefaultConfiguration.getDefaultQueueAutoDeleteDelay()).setAutoDeleteMessageCount(ActiveMQDefaultConfiguration.getDefaultQueueAutoDeleteMessageCount()), filter, pageSubscription == null ? null : pageSubscription.getPagingStore(), pageSubscription, scheduledExecutor, postOffice, storageManager, addressSettingsRepository, executorFactory.getExecutor(), server, this);
```
gemmellr, line 123: It seems more than a little weird the LVQ needs all those default configs specified in the options, but the non-LVQ case below does not, given the LVQ is a subclass of the non-LVQ, and its _QueueConfiguration_-based constructor seemingly only calls the superclass then differs only in capture of the LVQ key value.

Seems like there must be significant room for simplification/consolidation here (either in the queues, or the options here), rather than this arguably making things even more complicated/different?

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueFactoryImpl.java
jbertram: I went back and forth on this for awhile today trying to decide if this code should be updated or just left alone. Ultimately I decided it should be removed along with the other deprecated code. I pushed a new commit with that change plus an updated commit message with an explanation. Let me know what you think.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueFactoryImpl.java
jbertram: See my comment above.

artemis-server/src/main/java/org/apache/activemq/artemis/core/server/impl/QueueFactoryImpl.java
gemmellr: I hadn't even noticed that the QueueConfig-using factory method I was commenting on _was itself also deprecated_ (since that doesnt show in the initial diff context of the change, and since the described aim was 'deprecated QueueImpl constructor removal', I simply didnt think to look further at the factory method since it was just being updated to use the new QueueImpl constructors).

Given the factory methods were deprecated in the same commit as the QueueImpl constructors nearly 5 years ago, and you say the factories are also still really internal/impl detail, just removing the deprecated factory methods rather than updating them seems fair.

gemmellr: Though I ran the full test suite on this and saw some failures that look like theyll be related.

jbertram: Full test-suite is green on this with the latest push.

jbertram: To be clear, I had to rebase this when #5504 was merged today.

PR #5486, opened 2025-02-03
ARTEMIS-5139 improve locking on TypedProperties

jbertram:
Back in 8e40b2d4f4f242271d3dfcda4f9b96d3f94cee1b thread safety was added to TypedProperties by synchronizing all relevant methods. This was simple and effective but suffers from performance issues in read-heavy use-cases.

This commit improves the performance by using a read/write locking mechanism so reads can execute concurrently.

No new tests were added since the original commit added tests to verify thread safety.


jbertram:
Full test-suite is good on this.

PR #5493, opened 2025-02-10
ARTEMIS-5304 Allow delayed close of federation consumers

tabish121:
When demand is removed from a federation consumer and the AMQP link is drained allow the consumer to idle for a configurable time period before closing the link. If demand returns before the idled link is closed we can recover the link by simply granting more credit.


jbertram:
I was expecting some documentation for this new setting, but I couldn't find it. Was that an oversight or does this not need any user-facing docs?

tabish121:
> I was expecting some documentation for this new setting, but I couldn't find it. Was that an oversight or does this not need any user-facing docs?

As follow work for this plus many other updates there will need to be rework of the federation docs but that's a bit out of scope here.  The defaults on this should suffice for the majority of users as this should be transparent in practice.

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/federation/AMQPFederationConfiguration.java
```diff
@@ -76,12 +77,20 @@ public final class AMQPFederationConfiguration {
...
     */
    public static final int DEFAULT_RECEIVER_QUIESCE_TIMEOUT = 60_000;
 
+   /**
+    * Default timeout (milliseconds) applied to federation receivers that have been stopped due to lack of
+    * local demand. The close delay prevent a link from detaching in cases where demand drops and returns
+    * in quick succession allowing for faster recovery. The idle timeout kicks in once the link has completed
+    * its drain of outstanding credit.
+    */
+   public static final int DEFAULT_RECEIVER_IDLE_TIMEOUT = 60_000;
```
gemmellr, line 92: I'd maybe be inclined to lower this given it acts as a default for both queue and address types. Its fine for queue federation, but given the scope for oddities to arise in some Address federation cases due to the federation consumer (and its related queue) sticking around between source demand, it seems like lower default would be appropriate there. Although  I realise it can of course be configured to 0 both overall and/or on a per-policy basis for anyone that needs to.

tabish121: Fair point and I thought about it but was thinking of starting simpler with one configuration option for both.  After thinking on it I will split the configuration for address and queue consumers as there is a chance of a race if federation of durable subscriptions is in play where it could still read message sent when it was removed and re-added for instance.  Queue consumers are fine with a longer timeout as the behavior is the same on the remote regardless of any timeout.

tabish121: I've added a commit which splits the timeout into two separate properties one for address and one for queues with the 60s value for queue consumer and a much reduced 5s value for address consumers. I chose not to retain a top level receiver idle timeout that applied to both at the federation configuration level as that may just make it more confusing in the end.

If this looks good I will squash the two commits but I kept it split for now to show the difference.

tabish121:
I've squashed the changes and rebased with main.

PR #5507, opened 2025-02-14
ARTEMIS-5302 extra fixes for using QueueConfiguration more

jbertram:


clebertsuconic:
can you add some context to the commit message so I don't need to open the JIRA to know what were the handful fixes? 


"A handful of fixes around QueueConfigurarion" would do :)



just because I didn't have the context on these changes until I opened JIRA.

clebertsuconic:
other than the commit message.. LGTM

PR #5510, opened 2025-02-19
ARTEMIS-5053 support commit interval for scale-down

jbertram:
This commit adds support for configuring how often the transactions governing scale-down are committed. It also does the following:

 - Updates ScaleDownTest to use inheritence instead of parameterization for repeating tests without repeating code.
 - Updates HAPolicy with default implementations which helps clarify which policies really care about scale-down.

### DIFF
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
@@ -953,6 +953,7 @@ public class ConfigurationImplTest extends AbstractConfigurationTestBase {
       properties.put("HAPolicyConfiguration.scaleDownConfiguration.groupName", "g0");
       properties.put("HAPolicyConfiguration.scaleDownConfiguration.clusterName", "c0");
       properties.put("HAPolicyConfiguration.scaleDownConfiguration.enabled", "false");
+      properties.put("HAPolicyConfiguration.scaleDownConfiguration.commitInterval", "33");
    }
 
    private void checkScaleDownConfiguration(ScaleDownConfiguration scaleDownConfiguration) {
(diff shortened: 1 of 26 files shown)


ehsavoie (approved):

jbertram:
@clebertsuconic and/or @gemmellr, I'd appreciate a review & merge here. Thanks!

clebertsuconic:
@jbertram one thing that failed very badly for us in the past was the practice of extending tests and duplicating them..


A test that was originally intended for something, then we duplicate the tests with something else.


I would prefer if we add just a single test validating the commit interval instead of repeating all the ones from the top.

I had done a lot of cleanup recently to remove things like this.. I would prefer if we don't do this kind of thing now.

jbertram:
@clebertsuconic, I changed up the tests like you asked. Give it another look. Thanks!

clebertsuconic:
LGTM

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImplTest.java
```diff
@@ -962,6 +963,7 @@ private void checkScaleDownConfiguration(ScaleDownConfiguration scaleDownConfigu
       assertEquals("g0", scaleDownConfiguration.getGroupName());
       assertEquals("c0", scaleDownConfiguration.getClusterName());
       assertFalse(scaleDownConfiguration.isEnabled());
+      assertEquals(33, scaleDownConfiguration.getCommitInterval());
```
clebertsuconic, line 966: a very tiny request.. can you use a different constant here? you used 33 on the XML... if the XML is parsed by accident in the test you wouldn't know if the value came from the XML or the properties.


If you okay with merging this now I can make this change myself with a git checkout and push.

clebertsuconic:
I pushed the commit with one slight change.

PR #5512, opened 2025-02-19
ARTEMIS-5314 overhaul JavaDoc formatting & content 

jbertram:
I apologize in advance for this PR. I started down this road, and it just kept growing. I didn't realize how many JavaDocs we actually had throughout the code-base.

Couple of things to note:

 - There are **no semantic changes** here. This is just comments.
 - I've set it to draft as there's no urgency to get this into 2.40.0.
 - All in all this PR removes right around 10,000 lines from the code-base which I think is a fairly significant improvement.


gemmellr:
Someone elses turn to chew on the massive code/javadoc twiddling, I wont be reviewing it :) 

Though on the commit message, the https://bugs.openjdk.org/browse/JDK-8256804 CSR Jira seems like it does a far better job of explaining {@return ...} than the decade old dev Jira referenced does, given the latters description doesnt actually mention the correct syntax or effect at all, and the comments don't appear to outline the final state either.

Presumably a draft because of the compilation failure on 21+

artemis-commons/src/main/java/org/apache/activemq/artemis/api/core/ActiveMQLargeMessageInterruptedException.java
```diff
@@ -16,8 +16,6 @@
  */
 package org.apache.activemq.artemis.api.core;
 
-/**
- */
 // XXX
```
tabish121, line 19: Missed some extra "comment" that could also get deleted.

jbertram: Removed.

artemis-commons/src/main/java/org/apache/activemq/artemis/json/JsonValue.java
```diff
@@ -57,10 +56,7 @@ public interface JsonValue {
    String toString();
 
    enum ValueType {
-      ARRAY,
-      OBJECT, STRING, NUMBER,
-      TRUE, FALSE,
-      NULL
+      ARRAY, OBJECT, STRING, NUMBER, TRUE, FALSE, NULL
```
tabish121, line 59: This doesn't appear to be javadoc related

clebertsuconic:
@jbertram can you rebase this so we can merge this ASAP?

jbertram:
@clebertsuconic, will do!

PR #5513, opened 2025-02-19
NO-JIRA Removing RetryRule

clebertsuconic:


jbertram:
Looks like you removed this line when you shouldn't have:
```java
import org.junit.After;
```

tests/activemq5-unit-tests/src/test/java/org/apache/activemq/transport/failover/FailoverClusterTest.java
```diff
@@ -35,18 +35,12 @@
 import org.apache.activemq.artemis.jms.server.config.impl.JMSConfigurationImpl;
 import org.apache.activemq.artemis.jms.server.embedded.EmbeddedJMS;
 import org.apache.activemq.broker.artemiswrapper.OpenwireArtemisBaseTest;
-import org.apache.activemq.broker.artemiswrapper.RetryRule;
-import org.junit.After;
```
jbertram, line 39: I think this line needs to be put back.

PR #5518, opened 2025-02-21
ARTEMIS-5310 check class of Principals in Subject

jbertram:


artemis-server/src/main/java/org/apache/activemq/artemis/core/security/impl/SecurityStoreImpl.java
```diff
@@ -258,6 +263,27 @@ public String authenticate(final String user,
...
+    * password) which should be cached.
+    */
+   private boolean validateExpectedUserPrincipal(Subject subject) throws ClassNotFoundException {
+      if (subject != null) {
+         Class expectedPrincipal = UserPrincipal.class;
+         if (securityManager instanceof ActiveMQJAASSecurityManager jaasManager) {
+            expectedPrincipal = Class.forName(jaasManager.getUserPrincipalClass());
+         }
+
+         if (subject.getPrincipals(expectedPrincipal).size() == 0) {
```
clebertsuconic, line 279: that's very specific for the implemention.. isn't?


Isn't there a way to make this more generic?

jbertram: This is the way it's been since the JAAS security manager was first implemented way back in Artemis 1.2. We just haven't done this check until now, and we're only doing it because a user complained about it.

Technically it would be possible to use the very generic `java.security.Principal` but then we wouldn't be able to distinguish from users and roles in the `Subject`. In any case, that change is beyond the scope of this PR.

PR #5523, opened 2025-02-24
Updated doc with env var substitution in config files

ncasaux:


jbertram:
FYI, I manually updated this to reference ARTEMIS-5327 in the commit message.

jbertram:
BTW, thanks for the contribution!

PR #5533, opened 2025-02-26
ARTEMIS-5332 add exportConfigAsProperties operation to server control

gtully:


artemis-commons/pom.xml
```diff
@@ -83,6 +83,10 @@
          <groupId>commons-beanutils</groupId>
          <artifactId>commons-beanutils</artifactId>
       </dependency>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-boot</artifactId>
```
jbertram, line 88: The `artemis-commons` module should essentially be at the "bottom" of the module hierarchy with no other dependencies on Artemis modules. This keeps it free to be used by any module without the risk of circular dependencies cropping up.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
```diff
@@ -4721,6 +4722,22 @@ public long getAuthorizationFailureCount() {
...
+   public void exportConfigAsProperties() throws Exception {
+      if (AuditLogger.isBaseLoggingEnabled()) {
+         AuditLogger.exportConfigAsProperties(this.server);
+      }
+
+      try (AutoCloseable lock = server.managementLock()) {
+         if (System.getProperty(TMP_DIR_SYSTEM_PROPERTY) == null) {
+            throw new IllegalStateException(TMP_DIR_SYSTEM_PROPERTY + " system property must be set");
+         } else {
+            File exportFileInTmp = new File(System.getProperty(TMP_DIR_SYSTEM_PROPERTY), CONFIG_AS_PROPERTIES_FILE);
```
jbertram, line 4735: Why not simply print the export and then users can do whatever they want with it (e.g. pipe it to a file, parse it in memory, etc.)?

gtully: Mostly for security. I don't want to leak any information, config can be sensitive.
In addition it could be quite large.
Doing transformation in place is safest

artemis-commons/pom.xml
gtully: I want to capture the sharing of the temp dir system properly. Maybe it should not be in boot but boot has no other dependencies.

jbertram: I understand you want access to the constant, but forfeiting the purity of `artemis-commons`'s dependencies seems a high price to pay. It would make more sense for `artemis-boot` to depend on `artemis-commons` instead of the other way around.

artemis-server/src/main/java/org/apache/activemq/artemis/core/management/impl/ActiveMQServerControlImpl.java
jbertram: We will be obfuscating sensitive (e.g. password) fields, no? Are there any security implications with putting the file into `/tmp`?

gtully: Yes, but I don't think that is sufficient.
instance-dir/tmp is the default,  instance-dir/etc would be best but could or should be read-only

artemis-commons/pom.xml
gtully: I don't think commons is that pure, the order of the modules has commons after boot. the boot jar is trivial, and the only thing on the classpath that knows how to exec the broker. commons has a bunch of deps in contrast. I don't want to change the run script to add an additional classpath jar to the broker start, but I do want to share the variable between the start and the runtime.

gtully: thanks for taking the time to review

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
```diff
@@ -945,6 +950,166 @@ public <T> T convert(Class<T> type, Object value) {
...
+         throw new RuntimeException("error accessing: " + nested, e);
+      }
+   }
+
+   private boolean isComplexConfigObject(Object value) {
+      return !(value instanceof SimpleString || value instanceof Enum<?>) && value.getClass().getPackage().getName().contains("artemis");
+   }
+
+   private boolean shouldRedact(String name) {
+      return name.toUpperCase(Locale.ENGLISH).contains("PASSWORD");
```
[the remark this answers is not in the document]
gtully: I don't like the idea of returning the entire content, the intent is to provide a view of the effective config, and present the properties format that is tricky to grok. It is a transform the would be best done inplace, the closest writable place is the broker tmp dir.
if the whole file is encrypted, the contents will be less useful.
redacting the passwords is a miminum, but not exporting the config from the broker is the best way to keep the config secure.

artemis-commons/pom.xml
jbertram: The `artemis-commons` module does have a handful of dependencies, but the important part is that it doesn't depend on any other Artemis modules (aside from logging which isn't really a concern). The whole notion of a "commons" module is that it allows modules to share classes when otherwise there would be problems with circular or otherwise unwanted dependencies. By making `artemis-commons` depend on `artemis-boot` you're now adding `artemis-boot` as a transitive dependency to anything that depends on `artemis-commons`. 

At the end of the day the dependency only exists for a constant defining `java.io.tmpdir` which is a [well-known Java system property](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/System.html#getProperties()). In my opinion we should just use the literal `"java.io.tmpdir"` in both places and dispense with the constant and the new dependency. We use literals like this for well-known Java system properties in a number of places in the code-base so a precedent exists.

It's also worth noting that the class in `artemis-commons` which uses the constant from `artemis-boot` (i.e. `org.apache.activemq.artemis.utils.SpawnedVMSupport`) is only used in tests. That class should probably be moved to a test-specific module (e.g. `artemis-unit-test-support`) if possible.

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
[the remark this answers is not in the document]
gtully: this is a good point, while the broker does the ENC in specific cases, broker properties will allow any value to use ENC, in that case, if it is read as ENC, it should be redacted. These will need to be tracked.

artemis-commons/pom.xml
gtully: I do want to use a constant such that the runtime and create dependency is captured. I will move that SpawnedVMSupport class, good catch! thanks!

jbertram: I just did a quick POC moving `org.apache.activemq.artemis.utils.SpawnedVMSupport` from `artemis-commons` to `artemis-unit-test-support` and everything looks good. Can you incorporate that change into your PR and then remove the new dependency from `artemis-commons` and add it to `artemis-unit-test-support` instead?

gtully: will do, thanks

artemis-server/src/main/java/org/apache/activemq/artemis/core/config/impl/ConfigurationImpl.java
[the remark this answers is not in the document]
brusdev: It sounds good

clebertsuconic:
This LGTM

artemis-server/pom.xml
```diff
@@ -156,6 +156,10 @@
          <groupId>io.micrometer</groupId>
          <artifactId>micrometer-core</artifactId>
       </dependency>
+      <dependency>
+         <groupId>org.apache.activemq</groupId>
+         <artifactId>artemis-boot</artifactId>
+      </dependency>
```
jbertram, line 162: I don't mean to sound like a broken record, but this seems inverted to me like the issue with `artemis-commons` before. The `artemis-boot` module starts the broker. It doesn't seem like the broker should depend up on it.

Have you considered moving the constant to `artemis-commons` and then making `artemis-boot` depend on `artemis-commons` but excluding all the `artemis-commons` dependencies so you don't get a bunch of transitive stuff? Then the boot classpath can safely be `artemis-boot` and `artemis-commons`.

I also still think it would make sense just to ditch the constant altogether and use the literal instead.

gtully: boot sets up and environment, and the broker makes use of that, it is communicating via system properties and the constant captures that.  I don't want that env to change without reference to the shared usage. boot is intentionally very small with no deps, it is a good thing to depend on.

gtully: I think your original argument for not putting it is commons made sense, boot is a better place.

clebertsuconic: Artemis-boot should be the loader.. nothing having a dependency towards boots.


We risk of circular dependencies I guess.

docs/user-manual/management.adoc
```diff
@@ -86,6 +86,7 @@ To do this use the `forceFailover()` operation.
 
 Since this method actually stops the server you will probably receive some sort of error depending on which management service you use to call it.
 ====
+* The effective configuration of the broker core can be exported as properties using `exportConfigAsProperties()`. This operation will write a file of the form `java.io.tmpDir/broker_config_as_properties.txt`. The content can provide an entry point for using xref:configuration-index.adoc#broker-properties[brokerProperties].
```
jbertram, line 89: The correct value here is `java.io.tmpdir` (i.e. `dir` not `Dir`). It also might be good to clarify that is value is coming from a System property and what that value is expected to be (i.e. `<artemis_instance>/tmp`).

gtully: Thanks, good catch. Have fixed that and added a Note in the configuration of the tmp dir.

PR #5535, opened 2025-02-27
ARTEMIS-5316 XOAUTH2 SASL mechanism for broker connection

bstlukasiewicz:


artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
```diff
@@ -1204,6 +1206,47 @@ public static boolean isApplicable(final NettyConnection connection) {
       }
    }
 
+   private static class XOAuth2SASLMechanism implements ClientSASL {
```
clebertsuconic, line 1209: Isn't there a way to write a Unit Test for this?

artemis-protocols/artemis-amqp-protocol/src/main/java/org/apache/activemq/artemis/protocol/amqp/connect/AMQPBrokerConnection.java
```diff
@@ -1204,6 +1206,47 @@ public static boolean isApplicable(final NettyConnection connection) {
...
+      }
+
+      @Override
+      public byte[] getResponse(byte[] challenge) {
+         return EMPTY;
+      }
+
+      public static boolean isApplicable(AMQPBrokerConnectConfiguration brokerConnectConfiguration, NettyConnection connection) {
+         Map<String, Object> params = connection.getConnectorConfig().getParams();
+         String amqpSaslMechanism = ConfigurationHelper.getStringProperty(TransportConstants.AMQP_SASL_MECHANISM, null, params);
```
tabish121, line 1237: This seems like something we can get from URI options and not scatter AMQP specific options into the netty TransportOptions bits that don't carry any other AMQP related constants.  I believe we already have configuration  on the acceptor URI for AMQP for what mechanisms the broker is allowed to offer.  This should be configured somewhat the same in that if you want to use XOAUTH2 then the connection should specify that as the only allowed option and fail if the remote doesn't offer it.

tabish121:
Closing as this appears to have been abandoned

PR #5537, opened 2025-02-27
ARTEMIS-5333 add config for global-max-size-percent-of-jvm-max-memory…

gtully:
… to simplify setting the global memory limit

### DIFF
--- a/artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationParserTest.java
+++ b/artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationParserTest.java
@@ -768,6 +768,7 @@ public class FileConfigurationParserTest extends ServerTestBase {
                            <core>
                               <global-max-size>10M</global-max-size>
                               <global-max-messages>1000</global-max-messages>
+                              <global-max-size-percent-of-jvm-max-memory>30</global-max-size-percent-of-jvm-max-memory>
                            </core>
                         </configuration>""");
 
@@ -777,6 +778,7 @@ public class FileConfigurationParserTest extends ServerTestBase {
 
       assertEquals(10 * 1024 * 1024, configuration.getGlobalMaxSize());
       assertEquals(1000, configuration.getGlobalMaxMessages());
+      assertEquals(30, configuration.getGlobalMaxSizePercentOfJvmMaxMemory());
    }
 
    @Test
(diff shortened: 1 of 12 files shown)


docs/user-manual/paging.adoc
```diff
@@ -158,7 +158,7 @@ You can also specify `global-max-messages` on the main configuration, specifying
 
 When you have more messages than what is configured `global-max-size` any new produced message will make that destination to go through its paging policy.
 
-`global-max-size` is calculated as half of the max memory available to the Java Virtual Machine, unless specified on the `broker.xml` configuration.
+`global-max-size` is calculated as `global-max-size-percent-of-jvm-max-memory` of the max memory available to the Java Virtual Machine, unless specified on the `broker.xml` configuration. The default value of `global-max-size-percent-of-jvm-max-memory` is 50% or half.
```
tabish121, line 161: Wondering if a rewording here would help as I needed to read this a few times before I figured out what it was saying. 

`global-max-size` is calculated as a percentage of the max memory available to the Java Virtual Machine, unless specified in the `broker.xml` configuration directly. The percentage value used is configurable using the option `global-max-size-percent-of-jvm-max-memory` in the XML configuration and if not specified the default value of `global-max-size-percent-of-jvm-max-memory` is 50% or half the max available memory.

gtully: thanks, yes, that is indeed better.

artemis-server/src/test/java/org/apache/activemq/artemis/core/config/impl/FileConfigurationParserTest.java
```diff
@@ -775,6 +776,7 @@ public void testMaxSize() throws Exception {
 
       assertEquals(10 * 1024 * 1024, configuration.getGlobalMaxSize());
       assertEquals(1000, configuration.getGlobalMaxMessages());
+      assertEquals(30, configuration.getGlobalMaxSizePercentOfJvmMaxMemory());
```
clebertsuconic, line 779: I would use a different number from the one you used on the XML.. in case the XML gets parsed by accident.

gtully:
> a test validating the properties (broker properties) would be nice

we get that for free with my other per that verifies export/import see: https://github.com/apache/activemq-artemis/pull/5533/files#diff-ab6dc3e25c4f07694286ccb6d128e61c311dd3ff89fdb91167923d82bc99ac15R1028

PR #5546, opened 2025-03-05
ARTEMIS-5338 AckManager flow control

clebertsuconic:
Say there are too many records in the JournalHashMap. In that case the sender should pace sending more data. The target should limit credits when the configured max pending records is reached.


clebertsuconic:
there's a bug.. I'm adding a test to verify it and fix it

jbertram:
LGTM

PR #5547, opened 2025-03-05
ARTEMIS-5340 ensure PEM provider is truly optional

jbertram:


artemis-core-client/src/test/java/org/apache/activemq/artemis/core/remoting/impl/netty/PEMProviderTest.java
```diff
@@ -0,0 +1,38 @@
...
+
+   @Test
+   public void testPEMProviderNotLoaded() throws Exception {
+      // use a method from SSLSupport to force the JVM to load it as well as any hard dependencies it has
+      SSLSupport.parseCommaSeparatedListIntoArray("");
+
+      // verify the actual PEM provider class is not loaded
+      Method findLoadedClassMethod = ClassLoader.class.getDeclaredMethod("findLoadedClass", String.class);
+      findLoadedClassMethod.setAccessible(true);
+      assertNull(findLoadedClassMethod.invoke(ClassLoader.getSystemClassLoader(), SSLSupport.PEM_PROVIDER));
```
brusdev, line 36: An alternative to avoid `--add-opens java.base/java.lang=ALL-UNNAMED` is to check for the defined packages, i.e.
```suggestion
      assertNull(ClassLoader.getSystemClassLoader().getDefinedPackage("de.dentrassi.crypto.pem"));
```

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/SSLSupport.java
```diff
@@ -352,10 +354,15 @@ public static KeyStore loadKeystore(final String keystoreProvider,
...
+   /**
+    * This method uses reflection to load the appropriate java.security.Provider for PEM use-cases. Reflection is used
+    * to avoid a hard dependency on the provider's implementation so that folks who don't use PEM don't have to include
+    * the corresponding dependency.
+    */
+   public static void checkPemProviderLoaded(String keystoreType) throws Exception {
       if (keystoreType != null && keystoreType.startsWith("PEM")) {
          if (Security.getProvider("PEM") == null) {
-            Security.insertProviderAt(new de.dentrassi.crypto.pem.PemKeyStoreProvider(),
+            Security.insertProviderAt((Provider) Class.forName(PEM_PROVIDER).getDeclaredConstructor().newInstance(),
```
gtully, line 365: it may make sense to use ClassLoadingUtils (with a type check) for consistency here. For some background see: https://issues.apache.org/jira/browse/ARTEMIS-4766

gemmellr: You could instead just move the original line (with wrapped next part) into a static method in its own sibling class, e.g PEMSupport.loadProvider() and then call out to it. That should then just work as intended, even without using reflection, because that class would not be initialized unless executing the method and so the related classes/dependency would then also not be referenced.

jbertram: I like that solution, @gemmellr.

artemis-core-client/src/test/java/org/apache/activemq/artemis/core/remoting/impl/netty/PemProviderTest.java
```diff
@@ -0,0 +1,64 @@
...
+@TestMethodOrder(OrderAnnotation.class)
+public class PemProviderTest {
+
+   static final String PEM_PROVIDER_PACKAGE = "de.dentrassi.crypto.pem";
+
+   @Test
+   @Order(1)
+   public void testPemProviderNotLoaded() {
+      // ensure the PEM provider wasn't already loaded by some other test
+      Assumptions.assumeFalse(Arrays.stream(ClassLoader.getSystemClassLoader().getDefinedPackages()).anyMatch(pkg -> PEM_PROVIDER_PACKAGE.equals(pkg.getName())));
```
gemmellr, line 45: Would be better to make it an _assertion_ and let it fail when it needs to, to serve as notice the two tests are clashing and need to be ordered or isolated from each other to fix the problem.

With an assumption, if some other new test is created and runs first, then this test will just start skipping and the 'dont load it to let it be removed if not used' effect the test verifies could be broken without noticing, unless someone spots the extra skip beginning to occur. Might as well not have a test at that point.

Its a requirement for the test to exist that it be run before the first other test in the module to use it, bit its not a reason to optionally run the test or not. We always want to run it, but should check the requirement was satisfied at the start, the same way the test checks it at the end currently.

artemis-core-client/src/test/java/org/apache/activemq/artemis/core/remoting/impl/netty/PemProviderTest.java
```diff
@@ -0,0 +1,64 @@
...
+   /**
+    * This test simply verifies that we're using the right literal for the PEM provider implementation and that it does
+    * actually get loaded when referenced.
+    */
+   @Test
+   @Order(2)
+   public void testPemProviderPackageName() {
+      assertEquals(PEM_PROVIDER_PACKAGE, PemKeyStoreProvider.class.getPackageName());
+      assertNotNull(ClassLoader.getSystemClassLoader().getDefinedPackage(PEM_PROVIDER_PACKAGE));
+   }
```
gemmellr, line 63: I think a better test would be loading the provider the way its normally to be loaded, by using SSLSupport to load a pem keystore (the other test could also load a non-pem keystore to similarly check that does not load the pem provider), instead of just directly loading the class.

The 'check the correct package is looked for' could still be done, just as the last thing the test does as a form of confirmation. Or it could be a third test.

artemis-core-client/src/test/java/org/apache/activemq/artemis/core/remoting/impl/netty/PemProviderTest.java
jbertram: Done.

artemis-core-client/src/test/java/org/apache/activemq/artemis/core/remoting/impl/netty/PemProviderTest.java
jbertram: I added a couple of new tests to verify loading keystores in addition to the static use.

clebertsuconic:
LGTM

artemis-core-client/src/main/java/org/apache/activemq/artemis/core/remoting/impl/ssl/SSLSupport.java
```diff
@@ -352,11 +352,14 @@ public static KeyStore loadKeystore(final String keystoreProvider,
       return ks;
    }
 
-   public static void checkPemProviderLoaded(String keystoreType) {
+   /**
+    * This method calls out to a separate class in order to avoid a hard dependency on the provider's implementation.
+    * This allows folks who don't use PEM to avoid using the corresponding dependency.
+    */
+   public static void checkPemProviderLoaded(String keystoreType) throws Exception {
```
gemmellr, line 359: Is the signature reallyneeded anymore, now the reflection is gone?

jbertram: I'm not following. Why wouldn't it be needed?

gemmellr: Since it wasnt needed beforehand and unlike the earlier reflection based changes adding various bits that throw, nothing seemingly now changes in the method in a way that freshly throws a checked exception?

PR #5553, opened 2025-03-11
ARTEMIS-5347 support environment variable substitution in bootstrap.xml

jbertram:


artemis-dto/src/main/java/org/apache/activemq/artemis/dto/XmlUtil.java
```diff
@@ -95,6 +95,7 @@ public static <T> T decode(Class<T> clazz,
       unmarshaller.setSchema(schema);
 
       Properties props = new Properties(System.getProperties());
+      props.putAll(System.getenv());
```
gemmellr, line 98: This would create an ordering effect with the opposite precedence than I would typically expect, both from other configuration bits I've come across elsewhere, and also the other XMLUtil from Artemis. Worth flipping?

artemis-dto/src/test/java/org/apache/activemq/artemis/dto/XmlUtilTest.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class XmlUtilTest {
+
+   @Test
+   public void testPropertySubstituion(@TempDir Path tempDir) throws Exception {
+      final String SYSTEM_PROP_NAME = "sysPropName";
+      final String SYSTEM_PROP_VALUE = "sysPropValue";
+      System.setProperty(SYSTEM_PROP_NAME, SYSTEM_PROP_VALUE);
```
gemmellr, line 35: This will be left set for later tests. Could use the CleanupSystemPropertiesExtension, or subclass ArtemisTestCase where it is used already, to clear things up. The latter also has helpers for getting e.g test class and method names trivially to name things.

artemis-dto/src/test/java/org/apache/activemq/artemis/dto/XmlUtilTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+      // since System.getenv() returns an immutable Map we rely here on an environment variable that is likely to exist
+      final String ENV_VAR_NAME = "HOME";
```
gemmellr, line 38: Could just define a test specific env variable in the modules own surefire config to be used for the test. Can then assert it is the expected value and compare that known value to the parsing result. It could also more nicely be used for checking precedence handling works in the manner expected.

E.g something like:
  
    <environmentVariables combine.children="append">
       
       <TEST_ENV_VARIABLE>TestEnvVariableValue123</TEST_ENV_VARIABLE>
    </environmentVariables>

jbertram: I was actually doing that before, but I saw this pattern used in another test and thought it was cleaner.

jbertram:
I sent 304eb7ae6d78d154fd3474b25d169bd5e40250c1 directly to deal with your comments, @gemmellr. Thanks for the retroactive review!

PR #5555, opened 2025-03-12
ARTEMIS-5337 consumer + auto-create receives from existing wildcard queue

jbertram:
Most protocols don't natively understand the address/queue model used by Artemis. When those protocols want anycast semantics the standard behavior is to create an address and a queue with the same name. However, sometimes a queue already exists on the address with that particular name in which case the broker will attach the consumer to that queue.

This is acceptable most of the time, but it is problematic when that queue uses a wild-card. In this case the consumer will potentially receive messages sent to other addresses (i.e. whatever the wild-card covers).

This commit fixes that by detecting the wild-card queue and ignoring it. Instead the broker will create a new queue specifically for the incoming consumer.

### DIFF
--- a/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/WildcardConfiguration.java
+++ b/artemis-server/src/main/java/org/apache/activemq/artemis/core/config/WildcardConfiguration.java
@@ -178,6 +178,26 @@ public class WildcardConfiguration implements Serializable {
       }
    }
 
+   /**
+    * Detect whether the input {@code CharSequence} contains any unescaped "single word" or "any words" characters.
+    *
+    * {@code CharSequence} is used here to support both {@code String} and {@code SimpleString} objects.
+    */
+   public boolean isWild(CharSequence input) {
+      if (input == null || input.isEmpty()) {
+         return false;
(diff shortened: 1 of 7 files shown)


artemis-server/src/main/java/org/apache/activemq/artemis/core/config/WildcardConfiguration.java
```diff
@@ -178,6 +180,22 @@ public String convert(final String input, final WildcardConfiguration target) {
       }
    }
 
+   /**
+    * Detect whether the input {@code String} contains the "any words" or "single word" character and is not escaped.
+    */
+   public boolean isWild(String input) {
+      return input == null ? false : ((input.contains(getAnyWordsString()) || input.contains(getSingleWordString())) && !isEscaped(input));
```
brusdev, line 187: Should `isWild` return true when it include one `any words` character escaped and one `any words` character not escaped?

jbertram: That is a good point. Let me have a think about this...

jbertram: This shouldn't be a problem anymore. :+1:

brusdev (approved):

